PR #776, opened 2021-02-16
Update the grant parser to also support managing permissions for roles.

mhaley-miovision:
Just adding support for a missing case of permissions that can be granted to roles.


alanmcruickshank (approved):
Looks great - thanks for taking the time to put this together! 😄

PR #777, opened 2021-02-16
Null Literal instead of "keyword"

alanmcruickshank:
I noticed that `null` was behaving differently to `true` and `false` and being classed as a keyword in some cases instead of a literal.

- This adds an additional `NullLiteralSegment` so that we can parse null differently in different situations.
- It also adds a new variant on the capitalisation rule (`L040`) which is a rule for capitalising literals like `true`, `false` and `null`.

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -42,17 +42,20 @@ only_aliases = True
 [sqlfluff:rules:L003]
 lint_templated_tokens = True
 
-[sqlfluff:rules:L010]
+[sqlfluff:rules:L010]  # Keywords
 capitalisation_policy = consistent
 
-[sqlfluff:rules:L014]
+[sqlfluff:rules:L014]  # Unquoted identifiers
 capitalisation_policy = consistent
 
 [sqlfluff:rules:L016]
 ignore_comment_lines = False
 
-[sqlfluff:rules:L030]
+[sqlfluff:rules:L030]  # Function names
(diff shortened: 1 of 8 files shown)


NiallRees:
This is a wider question to the rule added - but I'm curious as to the reasoning behind rules addressing 'inconsistency' vs 'incorrect'. Inconsistent is quite vague and doesn't really explain the behaviour, eg is that file level or project level. How would you feel about standardising on using incorrect, and then rather than auto detecting, we just enforce capitalisations across a project?

NiallRees (approved):

alanmcruickshank:
> reasoning behind rules addressing 'inconsistency' vs 'incorrectness'

Inconsistency aim to address incorrectness, but tries to infer _correctness_ from the rest of the file. The current implementation is a bit dumb in that respect in that it assumes that the first is correct, but for a sufficiently large file that is _mostly correct_ the chance that the minority of errors include the first one is low, so it's a reasonable heuristic. A better heuristic would be to take the most common interpretation and use that, but that would require us to do two passes through the file.

The idea of doing consistency across files is a good one - but we would end up with an ordering problem - of where to take precedence from.

In the end the real question here is whether we should take a stronger stance on case - should we take a side on either lower or upper case *by default* rather then explicitly allowing people to choose. I assume the eventual answer to that question is yes, the question is more when. Right now we're biased towards adoption, by not taking a side we draw more people to sqlfluff, but eventually we should take a side. And that option is now just a configuration file away 😄

PR #779, opened 2021-02-18
Issues 380, 796: Warn if outer query produces an unknown number of result columns; Warn on unused CTEs

### LINKED ISSUE #380: Rule: Warn if the overall query (i.e. final query, ignoring CTEs) produces an indeterminate number of columns
Use of `SELECT *` in production queries is generally considered a bad practice, e.g. [this StackOverflow post](https://stackoverflow.com/questions/3639861/why-is-select-considered-harmful). SQLFluff should warn if a query does this.

In our team, we often allow an exception, i.e.`SELECT *` is okay *if* it's reading from a CTE (i.e. another "table" defined elsewhere in the same query), since in that case, the schema of the CTE may be fixed, so `SELECT *` from a CTE wouldn't allow any new upstream columns from other tables or queries to leak through.

This mistake caused one of our production jobs to fail recently because the query was being used to add new records to an existing table, and it failed in a cryptic way because the `SELECT *` picked up a new column that wasn't present in the existing table being appended to.

### LINKED ISSUE #796: Rule: Warn if a query has unused CTEs
_Originally posted by @NiallRees in https://github.com/sqlfluff/sqlfluff/issues/380#issuecomment-703273699_

This rule can likely reuse the `SelectInfo` class added for another rule (see #779).

barrywhart:
Resolves #380
Resolves #796

NOTE: Originally this was just a PR for #380, but I had developed something for #796 as a "proof of concept" for whether some of the code for #380 was reusable, and it was! So the additional code for the second issue was pretty short, and I figure it's sensible to include it in this PR so they can be reviewed together.

- [x] Add tests with operators that combine results of two queries, e.g. `UNION ALL`
- [x] Add tests that combine multiple tables using comma rather than an explicit join, e.g. `SELECT * FROM t1, t2 WHERE t1.id = t2.t1_id;`
- [x] Add tests where the table name uses an alias, both for CTEs and for regular tables.
- [x] Add tests that use table functions (e.g. BigQuery `UNNEST()`). Should be similar to a table wrt this rule: If they reference individual columns, that's okay, but `*` is not.
- [x] Add tests that use `SELECT *` (no table specified) with multiple tables (e.g. `FROM a JOIN b JOIN c`). This is equivalent to `a.*, b.*, c.*`.
- [x] Add tests that exercise table names in different dialects, different kinds of names (like BQ project.ds.table)


pwildenhain (commented):
Quick question:
Why is this rule 99? Instead of rule 40?

barrywhart:
@pwildenhain: The number is a placeholder. I expect this work to take several days at least -- once it's nearly finished, I'll choose "current largest rule number + 1" and rename it. By the time my last new rule was finished, someone else had already "taken" the same rule number as me, and it shows up as a merge conflict, which is kind of confusing to deal with. (Cuz it's completely unrelated code, just happens to use the same filename, right?).

pwildenhain:
Makes total sense 🤓

src/sqlfluff/core/rules/std/L020.py
```diff
@@ -1,10 +1,26 @@
...
 import itertools
+from typing import List, NamedTuple, Optional
 
+from sqlfluff.core.dialects.base import Dialect
+from sqlfluff.core.dialects.dialect_ansi import AliasInfo
+from sqlfluff.core.parser.segments.base import BaseSegment
 from sqlfluff.core.rules.base import BaseCrawler, LintResult
 
 
+class SelectStatementColumnsAndTables(NamedTuple):
```
barrywhart, line 12: I refactored `L020.py` to make its `SELECT` analysis reusable without having to inherit from it. No change in functionality.

test/core/rules/test_cases/L043.yml
```diff
@@ -0,0 +1,287 @@
+rule: L043
```
barrywhart, line 1: I tried to be very thorough about test cases, since this rule has to deal with many features of SQL queries (and how they interact). As you review, it would be 💯 if you could scan over this and see if I have missed any important cases.

src/sqlfluff/core/rules/std/L043.py
```diff
@@ -0,0 +1,264 @@
...
+
+    This class is a wrapper for L020.Rule_L020.get_select_statement_info(), plus
+    some additional functionality required by this rule.
+    """
+
+    segment: BaseSegment
+    tables: List[str]
+
+
+class SelectInfo:
```
barrywhart, line 31: Someday, I could imagine this class and the stuff from L020 moving to a core "query analysis" area of SQLFluff, independent of any particular rules. No rush on this, but an interesting possibility.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -683,19 +683,25 @@ def select_children(
                 buff.append(seg)
         return buff
 
-    def recursive_crawl(self, *seg_type):
+    def recursive_crawl(self, *seg_type, recurse_into=True):
```
barrywhart, line 686: New, optional parameter lets you avoid descending inside matching segments.

src/sqlfluff/core/rules/std/L043.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+class Rule_L043(BaseCrawler):
+    """Query produces an unknown number of result columns.
+
+    | **Anti-pattern**
+    | Querying all columns using `*` produces a query result where the number
+    | or ordering of columns may vary due to schema changes in upstream data
+    | sources. This should be avoided because it is prone to breakage in
+    | production.
```
NiallRees, line 158: Is it not the case that using `*` leaves you less prone to breakages, by the nature of it being a wildcard? My motivation for using a rule like this is being explicit about the columns produced by a query, and that schema changes are caught when you might otherwise be unaware.

barrywhart: It's a good point. I wrote this text without thinking about it as deeply as I could. I'll update to address your point and try to do some internet reading to get other perspectives too.

I think the risk of breakage I mentioned is more to downstream consumers -- if they only _expect_ certain columns, but you give them more, who knows how they'll behave? It's also just untidy. For example, if you are publishing a table to a wide audience, you might "leak" additional data to them unintentionally. Real-world example: At my employer, the BigQuery data warehouse is assembled from hundreds of production MySQL servers, and some tables actually include MySQL host names. If we accidentally leaked that field to a contractor employee or the outside world, that'd be a security risk.

barrywhart: In reading, I found one case where `SELECT *` could break _your own code`: `UNION` and `DIFFERENCE` require the inputs to have the same number of columns in the same order. If new columns are added to one input and not the other, I think the query will fail at runtime. Also, `SELECT *` passes through the sequence of the upstream columns, so if that changes (even though no columns were added or deleted), bad things may happen: runtime failure or wacky results (note that SQL does not check that the column *names* match, so you could be mixing completely different fields).

I decided to create a separate issue (#810) about `UNION` and `DIFFERENCE`. It's very much the same kind of concern, but it's a change in scope, and I think the rule provided by the current PR is a useful new feature by itself.

src/sqlfluff/core/rules/std/L043.py
```diff
@@ -153,9 +153,20 @@ class Rule_L043(BaseCrawler):
...
     | **Anti-pattern**
     | Querying all columns using `*` produces a query result where the number
-    | or ordering of columns may vary due to schema changes in upstream data
-    | sources. This should be avoided because it is prone to breakage in
-    | production.
+    | or ordering of columns changes if the upstream table's schema changes.
+    | This should generally be avoided because it can cause slow performance,
+    | cause important schema changes to go undetected, or break production code.
+    | For example:
+    | * If a query is expected to return a particular set of columns `c`, they
```
NiallRees, line 160: Is this meant to say `c.*`?

NiallRees, line 160:
```suggestion
    | * If a query is expected to return a particular set of columns `c.*`, it
```

barrywhart: Good catch! I think I meant to list several column names. I'll fix this up.

barrywhart:
One of my rule tests is failing, I think after merging the latest code from `master`. I'm investigating...

barrywhart:
I think the tests are working now. It looks like something may have changed the parsing structure the new rule was expecting. Also had some issues renumbering the rules to avoid colliding with a newly added, same-numbered rule. 😬

src/sqlfluff/core/parser/segments/base.py
dmateusp: Interesting! So does that mean you're not looking for a SELECT that is a child of another SELECT ? (I imagine that would only be the case for a subquery?)

src/sqlfluff/core/rules/std/L020.py
```diff
@@ -78,58 +94,60 @@ def _get_aliases_from_select(cls, segment, dialect=None):
...
-            reference_buffer = list(sc.recursive_crawl("object_reference"))
-            # Add any wildcard references
-            reference_buffer += list(sc.recursive_crawl("wildcard_identifier"))
-            for potential_clause in (
-                "where_clause",
-                "groupby_clause",
-                "having_clause",
-                "orderby_clause",
+    @classmethod
+    def get_select_statement_info(
```
dmateusp, line 98: It doesn't feel right to me that you'd use a Rule method as a "util" or a "library", the only reason I would see for a Rule to use another rule is similarity in what it's fixing/looking for (through inheritance)

Do you think we could move this method and the _get_aliases_from_select into a `SelectCrawler` maybe? Which would inherit from `BaseCrawler` ?

dmateusp: happy to be convinced otherwise haha

dmateusp (commented):
This is so so exciting @barrywhart :D

src/sqlfluff/core/parser/segments/base.py
barrywhart: We are definitely interested in nested queries, but not initially. The overall design is:
* Find all the top-level queries, i.e. the "main" query and all the CTEs.
* Starting from the main query, follow all the `SELECT *` dependencies, which can take us in three "directions":
  * To a CTE
  * To a nested query
  * To an external table

We don't need to find the nested stuff up front because we're not trying to explore the full query. We're just searching until we hit one of:
* External table
* Query that does **not** use `*`

src/sqlfluff/core/rules/std/L020.py
barrywhart: I agree completely! I was hoping to get feedback from some folks before doing this, for suggestions on what directory and/or module to put it in.

@NiallRees, @alanmcruickshank: Any thoughts on the above? Happy to follow advice or forge my own path here. 🙂

barrywhart:
@dmateusp: Thanks! I'm excited, too. Niall had mentioned that implementing this would start to look more like a query planner.

I'm looking forward to possibly using similar techniques (maybe even this code) to do other kinds of dependency walking, e.g. which `SELECT`s are sourced from a particular input table. (This will be part of the employer-specific "privacy" plugin I've mentioned -- the goal would be: flag any query that reads "sensitive" column names like `user_id` or `email` from the data warehouse unless it also `JOIN`s or otherwise filters that data to remove those who have "opted out".)

Who knows what other possibilities people will think of? For things like this, is there value in having SQLFluff start analyzing dependencies **across** SQL files?

src/sqlfluff/core/rules/std/L020.py
barrywhart: @dmateusp and others: I'm 💯 on the suggested reorg, but if it's okay, I suggest maybe deferring that until a future PR. This could be an immediate follow-on PR, or perhaps wait until the next rule that uses this code.

My reason for suggesting this is that it's challenging to keep a "new rule" PR current. In the week or so this PR has been open, I think I've already had to renumber the rules twice. (Each time a new rule is added, I get a merge conflict and have to make a number of small code changes related to the renumbering, just to get the tests and everything working again.)

barrywhart: Since this PR has been open a bit, I will look into the reorg. I think I will look at creating a new module with functions, rather than a class deriving from `BaseCrawler`. My reasoning is, I think over time we may start to assemble a growing number of code analysis tools, and being able to (re-)use them without inheritance will be more flexible (like if one rule wants to use multiple tools).

barrywhart:
@dmateusp: Ok, reorg complete! I organized it as follows:
```
src/sqlfluff/core/rules/analysis
├── __init__.py
├── select.py
└── select_crawler.py
```

src/sqlfluff/core/rules/analysis/select_crawler.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class WildcardInfo(NamedTuple):
+    """Structure returned by SelectCrawler.get_wildcard_info()."""
+
+    segment: BaseSegment
+    tables: List[str]
+
+
+class SelectCrawler:
```
dmateusp, line 20: The new structure is great, I think it makes it a lot clearer and it sets an example for future "statement specific crawling".
However there's a "BaseCrawler" in the project so as a contributor I'd be a little confused that other crawlers do not inherit from the "Base".

I'm not sure what the right answer is however, do you think in a follow up PR it would make sense to rename BaseCrawler to BaseRule? And keep the "crawler" concept to functions that browse the parsed SQL?

barrywhart: I like that idea. I think calling them "rules" aligns better with the documentation.

I'll create a ticket for doing that.

dmateusp: did you create a ticket? just making sure it's not lost

src/sqlfluff/core/rules/std/L044.py
```diff
@@ -0,0 +1,140 @@
...
+class RuleFailure(Exception):
+    """Exception class for reporting lint failure inside deeply nested code."""
+
+    pass
+
+
+class Rule_L044(BaseCrawler):
+    """Query produces an unknown number of result columns.
+
+    | **Anti-pattern**
```
dmateusp, line 18: Those explanations are gold, sqlfluff is really becoming a repository of knowledge for SQL

src/sqlfluff/core/rules/std/L045.py
```diff
@@ -0,0 +1,73 @@
...
+from sqlfluff.core.rules.doc_decorators import document_fix_compatible
+
+
+@document_fix_compatible
+class Rule_L045(BaseCrawler):
+    """Query defines a CTE (common-table expression) but does not use it.
+
+    | **Anti-pattern**
+    | Defining a CTE that is not used by the query is harmless, but it means
+    | the code is unnecesary and could be removed.
```
dmateusp, line 16: very cool that we're able to detect that :)

dmateusp (approved):
This is super cool @barrywhart 🎉

src/sqlfluff/core/rules/analysis/select_crawler.py
barrywhart: I did. It's #840.

src/sqlfluff/core/rules/std/L044.py
barrywhart: Thanks! I recently reviewed the existing rules and found a few we may want to improve. #812

src/sqlfluff/core/rules/std/L045.py
barrywhart: It was a good exercise -- I was really happy to see the code was so simple by reusing code from the other rule. I think if we keep refactoring the reusable code, it'll be really great in a few months. For now it works, which is a good start. 😄

Issue #545, opened 2020-11-11
Prefer CTE's over subqueries


NiallRees:
https://discourse.getdbt.com/t/why-the-fishtown-sql-style-guide-uses-so-many-ctes/1091

pwildenhain:
This resonates with me, I could definitely see implementing this as a lower-level lint (shout out to #201)

alanmcruickshank:
Do you think that #782 addresses this enough. That PR is designed to address #89, but I realise comes close to this.

NiallRees:
> Do you think that #782 addresses this enough? That PR is designed to address #89, but I realise comes close to this.
> 
> I'm tempted to say yes.

Just added a comment on the PR 👍

alanmcruickshank:
I'm going to close this now that #782 is merged. I think the configurable elements of it solve this issue.

PR #782, opened 2021-02-19
L042: #89 No Subqueries in Joins

### LINKED ISSUE #89: Rule: No nested selects within join clauses.
This is a pretty general, high-level quality check. I often see this issue with SQL written by new data scientists or engineers -- they'll deliver a working query, but it's monolithic and 50+ lines long. Even with proper indentation, it's hard to grasp what going on.

When this occurs, SQLFluff should suggest breaking the query into two or more logical chunks, with some of these chunks written as (descriptively named) CTEs.

alanmcruickshank:
This resolves #89 (and also the oldest open issue right now!)

This flags up any subqueries in join clauses, indicating that the user should use CTEs instead.

Small note, I also corrected the naming of `MainTableExpressionGrammar` to `MainTableExpressionSegment` because it was a `Segment` and not a `Grammar`. 😄

### DIFF
--- a/src/sqlfluff/core/rules/std/L042.py
+++ b/src/sqlfluff/core/rules/std/L042.py
@@ -4,7 +4,11 @@ from sqlfluff.core.rules.base import BaseCrawler, LintResult
 
 
 class Rule_L042(BaseCrawler):
-    """Join clauses should not contain subqueries. Use CTEs instead.
+    """Join/From clauses should not contain subqueries. Use CTEs instead.
+
+    By default this rule is configured to allow subqueries within `FROM`
+    clauses but not within `JOIN` clauses. If you prefer a stricter lint
+    then this is configurable.
 
     NB: Some dialects don't allow CTEs, and for those dialects
     this rule makes no sense and should be disabled.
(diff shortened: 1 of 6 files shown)


src/sqlfluff/core/rules/std/L042.py
```diff
@@ -0,0 +1,64 @@
+"""Implementation of Rule L042."""
+
+from sqlfluff.core.rules.base import BaseCrawler, LintResult
+
+
+class Rule_L042(BaseCrawler):
+    """Join clauses should not contain subqueries. Use CTEs instead.
+
+    NB: Some dialects don't allow CTEs, and for those dialects
```
dmateusp, line 9: I feel like that should be an issue?

I think the base rule class "BaseCrawler" should have an `exclude_dialects = ` optional property

dmateusp (commented):
Do you think it only applies to JOIN clauses?

for example:

```
select x, y, z
from (
   select ...
)
```

Would be another one ?

src/sqlfluff/core/rules/std/L042.py
alanmcruickshank: great idea - but in this case it wouldn't help. It's only older versions of mysql which don't support it, newer ones are fine. Do you know of any other rules which might use a feature like this?

alanmcruickshank:
> Do you think it only applies to JOIN clauses?

There was a long discussion on this in #89, and we eventually settled out that this was the best balance of allowing some complexity where it's not crazy, but still putting a limit in place for the level of complexity.

NiallRees:
How would you feel about adding config to this to also include failing on subqueries in a from clause too? It definitely closes #545 in its entirety then.

src/sqlfluff/core/rules/std/L042.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def _eval(self, segment, **kwargs):
+        """Join clauses should not contain subqueries. Use CTEs instead.
+
+        NB: No fix for this routine because it would be very complex to
+        implement reliably.
+        """
+        if segment.is_type("join_clause"):
+            # Get the referenced table segment
+            table_expression = segment.get_child("table_expression")
```
barrywhart, line 46: Does this handle the case of multiple joins, e.g.:
```
SELECT c
FROM t1
JOIN t2
ON <<something>>
JOIN (
    SELECT x, z FROM t3
)
```

It may very well work, mainly I'm unsure how the grammar is structured. Worth adding a test for this?

barrywhart (approved):
One question about multiple joins. Overall, looks great!

:shipit:

src/sqlfluff/core/rules/std/L042.py
alanmcruickshank: It should cover it yes, but I'll add a test case to confirm that.

alanmcruickshank:
> How would you feel about adding config to this to also include failing on subqueries in a from clause too?

I like this idea, I'll make the default that it's only join clauses, but that it can be configured to also lint from clauses.

src/sqlfluff/core/rules/std/L042.py
```diff
@@ -0,0 +1,81 @@
+"""Implementation of Rule L042."""
+
+from sqlfluff.core.rules.base import BaseCrawler, LintResult
+
+
```
pwildenhain, line 5: @alanmcruickshank I think we're missing the `@document_configuration` decorator here. I'll make a quick PR to add it in

PR #783, opened 2021-02-19
Postgres Syntax Gaps

### LINKED ISSUE #341: [postgres] Support for CREATE TYPE
Please add support for `CREATE TYPE` in PostgreSQL dialect.
* https://www.postgresql.org/docs/current/sql-createtype.html

Simple example:
```sql
CREATE TYPE my_enum AS ENUM ('value');
```

### LINKED ISSUE #343: [postgres] Support schema-qualified column data type
All objects in PostgreSQL may be schema-qualified to ensure addressing a specific version, should an object with the same name exist in a schema that might be searched first.

In case of column data types, this fails:
```sql
CREATE TABLE counters (
    my_type public.MY_TYPE
);
```
...with:
```
== [db/test.sql] FAIL
L:   1 | P:   1 |  PRS | Found unparsable section: 'CREATE TABLE counters (\n
                       | my_type publ...'
```

alanmcruickshank:
This resolves #341 and resolves #343.

I've added both to the ansi dialect, because they're additive and don't conflict with anything already there. Plus there may be other dialects beyond postgres with a similar syntax.

### DIFF
--- a/test/fixtures/parser/postgres/postgres_create_type.sql
+++ b/test/fixtures/parser/postgres/postgres_create_type.sql
@@ -1,0 +1,6 @@
+-- https://www.postgresql.org/docs/current/sql-createtype.html
+CREATE TYPE foo;
+CREATE TYPE bar AS ENUM ('foo', 'bar');
+CREATE TYPE foobar AS RANGE (SUBTYPE = FLOAT);
+CREATE TYPE barbar AS (INPUT = foo, OUTPUT = bar);
+CREATE TYPE foofoo AS (foo varchar collate utf8);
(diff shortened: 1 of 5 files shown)


test/fixtures/parser/postgres/postgres_create_type.sql
```diff
@@ -0,0 +1,6 @@
+-- https://www.postgresql.org/docs/current/sql-createtype.html
+CREATE TYPE foo;
```
barrywhart, line 2: Since this PR adds to the ANSI grammar, would it make sense for this test to be in the ANSI directory rather than Postgres?

barrywhart (approved):
One question about whether to test in the Postgres or ANSI dialect. Looks great overall.

:shipit:

test/fixtures/parser/postgres/postgres_create_type.sql
alanmcruickshank: I thought about this, and in the end - given it's explicitly a postgres query it made sense to test there. Eventually the code in the ansi dialect may get more generic (e.g. with Sql server?) and then it might be two test cases, served by the same code. In that instance I think it makes sense to keep the test cases dialect specific.

PR #785, opened 2021-02-19
#716: Support configurable file extensions

### LINKED ISSUE #716: Add Support for Alternate File Types/Extensions
It would be awesome if sqlfluff could support configuring a list of file types/extensions to be considered by the `lint` and `fix` commands. Ideally, the configuration could be done in the `.sqlfluff` file.

One example would be Jinja2 files that are templated SQL statements. Using the convention of `.sql.j2	 for the file extension would allow these files to be targeted for formatting along side `.sql` files.

Thanks!

alanmcruickshank:
This resolves #716.

This was originally configurable through a kwarg of the `Linter()` class, but nothing ever used that kwarg. I've moved the configuration of that value to the `FluffConfig` so that it can be driven by the config files. I've also added `.dml`, `.ddl` and `.sql.j2` files to the default search.


src/sqlfluff/core/default_config.cfg
```diff
@@ -9,6 +9,9 @@ recurse = 0
 output_line_length = 80
 runaway_limit = 10
 ignore_templated_areas = True
+# Comma seperated list of file extentions to lint.
```
NiallRees, line 12:
```suggestion
# Comma separated list of file extensions to lint.

```

NiallRees (approved):
Crushing the issues!

pwildenhain (approved):
Ah well done, this definitely belongs in the `FluffConfig` class rather than `Linter`

PR #791, opened 2021-02-20
Snowflake parsing bug fixes

NiallRees:
Apologies for not keeping the PR specific in scope.

- Add parsing for Snowflake's `within group`: https://github.com/sqlfluff/sqlfluff/pull/791/files#diff-81fba339f787f85c59fc64220234a1a58bbfd13ca07b2fc45d90e5909e1f127b
- Allow lateral flattens after joins. This involved fixing the parsing structure of the `FromClauseSegment` and `TableExpressionSegment`. https://github.com/sqlfluff/sqlfluff/pull/791/files#diff-9b90add4f247b50c3d7d4bf0b683ed7b96af8a2569a87fafc61a287ca6a979da
- Add parsing for Snowflake's `first_value(... ignore nulls)` per https://docs.snowflake.com/en/sql-reference/functions/first_value.html


src/sqlfluff/core/dialects/dialect_ansi.py
```diff
@@ -1100,6 +1100,10 @@ class FromClauseSegment(BaseSegment):
         AnyNumberOf(
             Ref("JoinClauseSegment"), Ref("JoinLikeClauseGrammar"), optional=True
         ),
+        # In case of table functions e.g. lateral flatten
+        AnyNumberOf(
+            Sequence(Ref("CommaSegment"), Ref("TableExpressionSegment")), optional=True
```
NiallRees, line 1105: I'm not sure this is the best way to do this!

src/sqlfluff/core/dialects/dialect_snowflake.py
```diff
@@ -58,6 +58,7 @@
         "BLOCK",
         "HISTORY",
         "LATERAL",
+        "NETWORK",
```
NiallRees, line 61: network is not a reserved keyword.

src/sqlfluff/core/dialects/dialect_ansi.py
alanmcruickshank: This _works_, but isn't where I would put it. I would put it within the `Delimited(OneOf(...))` in the `parse_grammar` of `FromClauseSegment`. That's the same place which deals with the old skool join syntax of `SELECT * FROM a, b WHERE ...`.

alanmcruickshank (changes requested):

src/sqlfluff/core/dialects/dialect_ansi.py
NiallRees: Ah so the `TableExpressionSegment` already exists there:

https://github.com/sqlfluff/sqlfluff/pull/791/files#diff-e07ae5ab00e7f519401a3251b192139d2da4ebf08fd1d7403b795472bff4f29aR1089

But the issue is the terminator being a `JoinClauseSegment`:

https://github.com/sqlfluff/sqlfluff/pull/791/files#diff-e07ae5ab00e7f519401a3251b192139d2da4ebf08fd1d7403b795472bff4f29aR1091

When we can have lateral flattens after a join as demonstrated in the added test case https://github.com/sqlfluff/sqlfluff/pull/791/files#diff-9b90add4f247b50c3d7d4bf0b683ed7b96af8a2569a87fafc61a287ca6a979da

I'll have a think

NiallRees: I think I've fixed this by refactoring the `FromClauseSegment` and `TableExpressionSegment`. From clauses now live under `TableExpressionSegment`s, which in my head feels right and also according these docs: https://www.cockroachlabs.com/docs/v20.2/sql-grammar.htm#table_ref

What do you reckon?

src/sqlfluff/core/dialects/dialect_ansi.py
```diff
@@ -1110,7 +1102,11 @@ def get_eventual_aliases(self) -> List[Tuple[BaseSegment, AliasInfo]]:
         """
         buff = []
         direct_table_children = self.get_children("table_expression")
-        join_clauses = self.get_children("join_clause")
+        join_clauses = [
+            join_clause
+            for child in direct_table_children
+            for join_clause in child.get_children("join_clause")
+        ]
```
NiallRees, line 1109: Now that join clauses are nested underneath `table_expression` segments this had to be tweaked slightly

alanmcruickshank (approved):
Yeah I like this. Nice 👍

PR #792, opened 2021-02-20
Unparsable select in multiline comment bug

NiallRees:
Not yet investigated why this is happening. If anyone has any ideas before I dig in let me know!


alanmcruickshank:
I'll pick up this one along with the other test case you brought up. I think something messed up is going on in comment handling. We don't have many test cases with multi-line comments, and I did change the handling of them a few months ago 🤔 

Another great test case! Thanks @NiallRees

alanmcruickshank (approved):

Issue #531, opened 2020-11-05
Redundant boolean CASE WHEN


dmateusp:
I'm sure you have all seen these as I see them often!

```sql
select
    case
        when n > 0 then true else false end as is_positive
from table_a
```

I would like to add a rule to prefer

```sql
select
    n > 0 as is_positive
from table_a
```

A case when statement is redundant if it uses only one `when` which is turned into `true` or `false`

However, to add a "fix" the rule should also take in account the following:
```sql
select
    case
        when n >= 0 then false else true end as is_negative
from table_a
```

which should be turned into

```sql
select
    case
        not(n >= 0) as is_negative
from table_a
```

pwildenhain:
I actually haven't seen this 😓 but the rule definitely makes sense to me!

rbetzler:
Mind if I give this one a try? I've seen it out in the wild quite a bit lol

dmateusp:
That'd be great @rbetzler thanks :) I've seen your commit and it seems you're already on the right track, but let us know if you need any help

rbetzler:
Could I get a quick look at 0b88499d1320ea8f346a0ff015d7604bac5db556? Given this is my first attempt, I don't want to spam anyone with a bad initial PR lol

If it looks ok, or if it's easier for you all, I can open a PR against upstream. Just let me know what's the preferred process.

That said, I think I've got the basics on (a) returning an error and (b) providing fixes. I ran a few tests on my own company's (Petal) repo and my commit seemed to do the trick.

Hopefully this ain't too bad for a first go 😅

dmateusp:
@rbetzler it's not spam! Just go ahead an open the PR ^^, then we can have a discussion there.

You documented + tested the rule so I'd say it's a 🥇 first go

jankatins:
I also flag this in code reviews, but the logic is actually different if the condition can produce a NULL... Eg when in the initial examples the n is NULL.

barrywhart:
@dmateusp: Can this issue be closed? I see that a PR which mentions it (#798) was merged a few weeks ago.

dmateusp:
Yes! it should have been closed already, closing it now

hellomrStampe:
Hi,

Not sure if this is what resulted in Rule_L043 but it is given me some issues.

I have this where I only want to keep the value of B if the value of A is NULL. If the value of A is not null then B should be NULL

```
select
    case 
        when src.a is null
            then src.b
    end as b_trf,
    src.c
[...]

PR #798, opened 2021-02-23
#531: Use an explicit condition rather than case-when-bool

rbetzler:
Here's my attempt at #531. This is my first go at a SQLFluff PR, so hopefully I'm not too far off the mark 😅 

Thanks in advance for taking the time to review. Looking forward to get more involved in the project.

### DIFF
--- a/src/sqlfluff/core/rules/std/L043.py
+++ b/src/sqlfluff/core/rules/std/L043.py
@@ -6,7 +6,10 @@ from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 
 @document_fix_compatible
 class Rule_L043(BaseCrawler):
-    """Use an explicit condition rather than case when statement.
+    """Unnecessary case when statement. Use the "when" condition itself.
+
+    If a case when else statement returns booleans, we can reduce it to the
+    "when" condition.
 
     | **Anti-pattern**
 
@@ -18,11 +21,13 @@ class Rule_L043(BaseCrawler):
         from fancy_table
 
     | **Best practice**
+    |   Reduce to "when" condition. Wrap with "coalesce". If necessary, add
+    |   a "not" operator at the beginning.
 
     .. code-block:: sql
 
         select
-            fab > 0 as is_fab
+            coalesce(fab > 0, false) as is_fab
         from fancy_table
 
     """
--- a/src/sqlfluff/core/rules/std/L043.py
+++ b/src/sqlfluff/core/rules/std/L043.py
@@ -1,0 +1,104 @@
+"""Implementation of Rule L043."""
+
+from sqlfluff.core.rules.base import BaseCrawler, LintFix, LintResult
+from sqlfluff.core.rules.doc_decorators import document_fix_compatible
+
+
+@document_fix_compatible
+class Rule_L043(BaseCrawler):
+    """Use an explicit condition rather than case when statement.
+
+    | **Anti-pattern**
+
+    .. code-block:: sql
+
+        select
+            case
+                when fab > 0 then true else false end as is_fab
+        from fancy_table
+
+    | **Best practice**
+
+    .. code-block:: sql
+
+        select
+            fab > 0 as is_fab
+        from fancy_table
+
+    """
+
+    def _eval(self, segment, **kwargs):
+        """Find rule violations and provide fixes.
+
+        0. Look for a case expression
+        1. Find the first expression and "then"
+        2. Determine if "then" is followed by a boolean
+        3. If so, delete everything but the first expression
+        4a. If when-then-true, return deletions
+        4b. If when-then-false, return deletions and
+            * add a not condition
+            * wrap the expression in parenthesis
+        """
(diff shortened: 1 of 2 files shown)


src/sqlfluff/core/rules/std/L043.py
```diff
@@ -0,0 +1,104 @@
...
+        from fancy_table
+
+    """
+
+    def _eval(self, segment, **kwargs):
+        """Find rule violations and provide fixes.
+
+        0. Look for a case expression
+        1. Find the first expression and "then"
+        2. Determine if "then" is followed by a boolean
```
barrywhart, line 35: IIUC, the code checks for a Boolean after `THEN`, and if it finds one, it assumes the rest of the expression is:
```
else <<opposite value>>
```

What happens if this is not the case, e.g. if there's no else (result will be `NULL`, I think), or if it's `else <<complex expression>>>`?

In those cases, could we detect it doesn't fit the pattern and do nothing?

rbetzler: Yeah that's a good callout. Will narrow this rule down to only handle: `case when... then bool else bool end`.

Along these lines, now that I think about it a little more, we might also need to wrap the fixed expression in a `coalesce`.

Will try to make these updates later today.

barrywhart: Sounds great!

src/sqlfluff/core/rules/std/L043.py
```diff
@@ -0,0 +1,151 @@
...
+            case
+                when fab > 0 then true else false end as is_fab
+        from fancy_table
+
+    | **Best practice**
+
+    .. code-block:: sql
+
+        select
+            fab > 0 as is_fab
```
barrywhart, line 25: Should this example use `coalesce()`?

src/sqlfluff/core/rules/std/L043.py
```diff
@@ -0,0 +1,151 @@
+"""Implementation of Rule L043."""
+
+from sqlfluff.core.rules.base import BaseCrawler, LintFix, LintResult
+from sqlfluff.core.rules.doc_decorators import document_fix_compatible
+
+
+@document_fix_compatible
+class Rule_L043(BaseCrawler):
+    """Use an explicit condition rather than case when statement.
```
barrywhart, line 9: Can you think of a more precise way to explain the recommended practice (mentioning that it's not just "don't use case", but also about what the "case" statement is doing.

src/sqlfluff/core/rules/std/L043.py
```diff
@@ -0,0 +1,151 @@
...
+                else_bool_type = None
+                while segment.segments[idx].name != "END":
+                    if segment.segments[idx].raw_upper in ["TRUE", "FALSE"]:
+                        else_bool_type = segment.segments[idx].raw_upper
+                    idx += 1
+            # If then-bool-else-bool, return fixes
+            if (
+                then_bool_type is not None
+                and else_bool_type is not None
+                and then_bool_type != else_bool_type
```
barrywhart, line 77: In order to test this line, can you add a "passing" test case(s) like this?
```
case
            when fab > 0 then true else true end
```

barrywhart (changes requested):
The rule code looks 💯! Thanks for the updates. I have a few suggestions related to wording and adding one more test case. This is really close.

src/sqlfluff/core/rules/std/L043.py
rbetzler: Ah yeah that makes sense. I rewrote this a little. Hopefully it's clearer but not too declarative either.

barrywhart (approved):
Looks great!

rbetzler:
@barrywhart, thank you for the reviews! Much appreciated 😄 

Excited to be involved in the project.

barrywhart:
Thanks for the great work! Yes, it's a fun, cool, and *useful* project. 🎉

rbetzler:
@barrywhart, mind if I ask: is there anything else you need on this one from me to get it merged? Not sure if you need to coordinate other PRs or need another reviewer before this one goes in, but thought I'd ask. Thanks again!

dmateusp:
The coalesce bit is a very good catch, I actually had to test it myself in Snowflake because I didn't know about this behavior:

```sql
with

vals as (
    select * from values (1), (null), (-1) as t(n)
)

select
    n,
    n > 0,
    case when n > 0 then true else false end,
    coalesce(n > 0, false)
from vals
```

|N  |N > 0|CASE WHEN N > 0 THEN TRUE ELSE FALSE END|COALESCE(N > 0, FALSE)|
|---|-----|----------------------------------------|----------------------|
|1  |TRUE |TRUE                                    |TRUE                  |
|   |     |FALSE                                   |FALSE                 |
|-1 |FALSE|FALSE                                   |FALSE                 |

Leaving it here for reference

I'm going to go ahead and merge it, thanks @rbetzler 🎉

Issue #799, opened 2021-02-23
L019 (trailing vs leading commas) fix doesn't flow around inline comments 


pwildenhain:
## Expected Behaviour

Inline comments should be taken into consideration when fixing leading to trailing

## Observed Behaviour

The fix incorrectly places a comma _after_ the inline comment, instead of after the field and _before_ the comment

## Steps to Reproduce

`test.sql`
```sql
select
    field_1,
    field_2 -- some comment,
    field_3
from my_table
```

```
$ sqlfluff fix test.sql
==== finding fixable violations ====
== [test_sql.sql] FAIL
L:   3 | P:   5 | L019 | Found leading comma. Expected only trailing.
L:   4 | P:   5 | L019 | Found leading comma. Expected only trailing.
==== fixing violations ====
2 fixable linting violations found
Are you sure you wish to attempt to fix these? [Y/n] ...
Attempting fixes...
Persisting Changes...
== [test_sql.sql] PASS
Done. Please check your files to confirm.
```

fixed `test.sql`
```sql
select
    field_1,
    field_2 -- some comment,
    field_3
from my_table
```

## Version
Include the output of `sqlfluff --version` along with your Python version
sqlfluff, version 0.4.0
Python 3.8.6
## Configuration

None

This one is on me 🤦🏻 I didn't consider this test case. I'll submit a PR with a failing test case soon

PR #803, opened 2021-02-23
Account for inline comments in L019 fix

pwildenhain:
Related to #799 

Added two test cases, one for trailing and one for leading. Right now just the leading --> trailing fix is the one that fails.


pwildenhain:
I messed up the line endings, I'm just gonna start over

PR #806, opened 2021-02-24
Handle Recursion Error

alanmcruickshank:
I've had two separate reports of infinite recursion errors. I've not managed to recreate either on a real query, so I don't know *why* the slicer gets itself into this state, but this PR addresses the eventuality that allowed an infinite loop. If the _safety value_ is detected, the user will get an error message asking them to report their query, and the slicer will fail safe.

In the event that it is triggered, a slightly larger patch of query will be labelled as _"templated"_ that it should, but the only effect of that is that some linting errors will be unnecessarily suppressed.

@NiallRees - this should make the issue we discussed earlier safer - but it still doesn't address the root cause. I'd love to get a captive example to work with.


src/sqlfluff/core/templaters/python.py
```diff
@@ -632,44 +632,106 @@ def _split_uniques_coalesce_rest(
...
+                            )
+
+                if bookmark_idx == 0:
+                    # This is a SAFETY VALVE. In Theory we should never be here
+                    # and if we are it implies an error elsewhere. This clause
+                    # should stop any potential infinite recursion in it's tracks
+                    # by simply classifying the whole of the current block as
+                    # templated and just stopping here.
+                    # Bugs triggering this eventuality have been observed in 0.4.0.
+                    templater_logger.info(
```
NiallRees, line 719: Should this be a warn?

src/sqlfluff/core/templaters/python.py
```diff
@@ -632,44 +632,106 @@ def _split_uniques_coalesce_rest(
...
+                            yield TemplatedFileSlice(
+                                raw_slice.slice_type,
+                                slice(_starts[0], starts[0]),
+                                slice(_starts[1], starts[1]),
+                            )
+
+                if bookmark_idx == 0:
+                    # This is a SAFETY VALVE. In Theory we should never be here
+                    # and if we are it implies an error elsewhere. This clause
+                    # should stop any potential infinite recursion in it's tracks
```
NiallRees, line 715:
```suggestion
                    # should stop any potential infinite recursion in its tracks
```

src/sqlfluff/core/templaters/python.py
```diff
@@ -632,44 +632,106 @@ def _split_uniques_coalesce_rest(
...
+                                starts[1] + remnant_length,
+                            )
+                            yield TemplatedFileSlice(
+                                raw_slice.slice_type,
+                                slice(_starts[0], starts[0]),
+                                slice(_starts[1], starts[1]),
+                            )
+
+                if bookmark_idx == 0:
+                    # This is a SAFETY VALVE. In Theory we should never be here
```
NiallRees, line 713: Value/valve?

src/sqlfluff/core/templaters/python.py
```diff
@@ -632,44 +632,106 @@ def _split_uniques_coalesce_rest(
...
+                    # templated and just stopping here.
+                    # Bugs triggering this eventuality have been observed in 0.4.0.
+                    templater_logger.info(
+                        "        Safety Value Info: %s, %r",
+                        two_way_uniques,
+                        templated_str[int_file_slice.templated_slice],
+                    )
+                    templater_logger.warning(
+                        "        Python templater safety value unexpectedly triggered. "
+                        "Your query may be more conservatively sliced than expected. "
```
NiallRees, line 726: I don't think the end user will know what this means. Simply a 'report this with your query' would suffice imo.

src/sqlfluff/core/templaters/python.py
alanmcruickshank: Valve

src/sqlfluff/core/templaters/python.py
```diff
@@ -632,44 +632,107 @@ def _split_uniques_coalesce_rest(
...
+                    # by simply classifying the whole of the current block as
+                    # templated and just stopping here.
+                    # Bugs triggering this eventuality have been observed in 0.4.0.
+                    templater_logger.info(
+                        "        Safety Value Info: %s, %r",
+                        two_way_uniques,
+                        templated_str[int_file_slice.templated_slice],
+                    )
+                    templater_logger.warning(
+                        "        Python templater safety value unexpectedly triggered. "
```
NiallRees, line 725:
```suggestion
                        "        Python templater safety valve unexpectedly triggered. "
```

src/sqlfluff/core/templaters/python.py
```diff
@@ -632,44 +632,107 @@ def _split_uniques_coalesce_rest(
...
+
+                if bookmark_idx == 0:
+                    # This is a SAFETY VALVE. In Theory we should never be here
+                    # and if we are it implies an error elsewhere. This clause
+                    # should stop any potential infinite recursion in its tracks
+                    # by simply classifying the whole of the current block as
+                    # templated and just stopping here.
+                    # Bugs triggering this eventuality have been observed in 0.4.0.
+                    templater_logger.info(
+                        "        Safety Value Info: %s, %r",
```
NiallRees, line 720:
```suggestion
                        "        Safety Valve Info: %s, %r",
```

PR #809, opened 2021-02-25
Dialect and Rules api

### LINKED ISSUE #805: Expose supported dialects for import
As a developer using SQLFluff in other projects, it would be helpful to have a list of supported dialects available for import.

Currently I maintain a list of dialects inferred from the project source:

```python
supported_dialects = [
    "bigquery", "mysql", "teradata", "postgres", "snowflake", "exasol", "exasol_fs"
]
```

It would be great to replace this with:

```python
from sqlfluff import SUPPORTED_DIALECTS
```

alanmcruickshank:
This resolves #805 .

I've upgraded both the dialects and rules readouts to use NamedTuples for consistency. This also adds tests and examples.

### DIFF
--- a/test/api/info_test.py
+++ b/test/api/info_test.py
@@ -1,0 +1,19 @@
+"""Test using sqlfluff to extract elements of queries."""
+
+import pytest
+
+import sqlfluff
+
+
+def test__api__info_dialects():
+    """Basic linting of dialects."""
+    dialects = sqlfluff.dialects()
+    assert isinstance(dialects, list)
+    assert ("ansi", "ansi", "nothing") in dialects
+
+
+def test__api__info_rules():
+    """Basic linting of dialects."""
+    rules = sqlfluff.rules()
+    assert isinstance(rules, list)
+    assert ("L001", "Unnecessary trailing whitespace.") in rules
--- a/test/api/info_test.py
+++ b/test/api/info_test.py
@@ -1,7 +1,5 @@
 """Test using sqlfluff to extract elements of queries."""
 
-import pytest
-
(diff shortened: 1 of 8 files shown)


test/api/info_test.py
```diff
@@ -0,0 +1,17 @@
+"""Test using sqlfluff to extract elements of queries."""
+
+import sqlfluff
+
+
+def test__api__info_dialects():
+    """Basic linting of dialects."""
+    dialects = sqlfluff.dialects()
+    assert isinstance(dialects, list)
+    assert ("ansi", "ansi", "nothing") in dialects
```
NiallRees, line 10: nothing? Should we filter this out as it's really only an internal thing and everything inherits from ansi?

alanmcruickshank: what do you mean "filter it out"? You mean not include the inheritance of a dialect?

NiallRees: I mean just prevent `"nothing"` from being returned by `sqlfluff.dialects()`. Definitely include the inheritance

NiallRees: Or we could give it a meaningful name if it has to stay in? It's just a bit confusing atm

alanmcruickshank: ah  "nothing" isn't being included as a dialect here. `("ansi", "ansi", "nothing")` is a shorthand for ` DialectTuple(label="ansi", name="ansi", inherits_from="nothing")`. Does that make more sense?

NiallRees: It does indeed!

NiallRees (approved):

PR #817, opened 2021-02-25
Fix auto documentation for rules, broken by plugin PR

dmateusp:
Right now the auto documentation for rules is broken! I actually introduced the bug on my pluggy PR.

@NiallRees brought it to my attention on Slack and I'm adding the bug fix

(see current issue: https://docs.sqlfluff.com/en/latest/rules.html#module-sqlfluff.core.rules.std)

### DIFF
--- a/docs/source/rules.rst
+++ b/docs/source/rules.rst
@@ -12,7 +12,7 @@ for the rule itself, with all the other mechanics abstracted away.
 Specific Rules
 --------------
 
-.. automodule:: sqlfluff.core.rules.std
+.. automodule:: sqlfluff.core.rules
    :members:
    :member-order: alphabetical
 
--- a/plugins/sqlfluff-plugin-example/src/example/rules.py
+++ b/plugins/sqlfluff-plugin-example/src/example/rules.py
@@ -37,8 +37,10 @@ def get_configs_info() -> dict:
     }
 
 
-@document_fix_compatible
+# These two decorators allow plugins
+# to be displayed in the sqlfluff docs
 @document_configuration
+@document_fix_compatible
 class Rule_Example_L001(BaseCrawler):
     """ORDER BY on these columns is forbidden!
 
--- a/plugins/sqlfluff-plugin-example/src/example/rules.py
+++ b/plugins/sqlfluff-plugin-example/src/example/rules.py
@@ -39,8 +39,8 @@ def get_configs_info() -> dict:
 
 # These two decorators allow plugins
 # to be displayed in the sqlfluff docs
-@document_configuration
 @document_fix_compatible
+@document_configuration
 class Rule_Example_L001(BaseCrawler):
     """ORDER BY on these columns is forbidden!
 
--- a/src/sqlfluff/core/rules/__init__.py
+++ b/src/sqlfluff/core/rules/__init__.py
@@ -6,10 +6,24 @@ from sqlfluff.core.plugin.host import get_plugin_manager
 
 std_rule_set = RuleSet(name="standard", config_info=STANDARD_CONFIG_INFO_DICT)
 
+# Sphinx effectively runs an import * from this module in rules.rst, so initialise
+# __all__ with an empty list before we populate it with the rule names.
+__all__ = []
+
 # Iterate through the rules list and register each rule with the std_rule_set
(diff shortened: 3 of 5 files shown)


docs/source/rules.rst
```diff
@@ -12,7 +12,7 @@ for the rule itself, with all the other mechanics abstracted away.
 Specific Rules
 --------------
 
-.. automodule:: sqlfluff.core.rules.std
+.. automodule:: sqlfluff.core.rules
```
dmateusp, line 15: I moved it from that module! In the future if we decide to add some rules as plugins, they'll show up in the documentation website :)

src/sqlfluff/core/rules/__init__.py
```diff
@@ -6,11 +6,25 @@
...
+# Sphinx effectively runs an import * from this module in rules.rst, so initialise
+# __all__ with an empty list before we populate it with the rule names.
+__all__ = []
+
 # Iterate through the rules list and register each rule with the std_rule_set
 for plugin_rules in get_plugin_manager().hook.get_rules():
     for rule in plugin_rules:
         std_rule_set.register(rule)
 
+        if not rule.__name__.startswith("Rule_Example"):
```
dmateusp, line 18: Exclude the Example rules from the "Example" plugin

src/sqlfluff/core/rules/std/L039.py
```diff
@@ -18,7 +18,7 @@ class Rule_L039(BaseCrawler):
 
     | **Best practice**
     | Unless an indent or preceeding a comment, whitespace should
-    be a single space.
+    | be a single space.
```
dmateusp, line 21: This was actually complaining on linting, I imagine it wasn't caught before because of the issue

plugins/sqlfluff-plugin-example/src/example/rules.py
```diff
@@ -37,8 +37,10 @@ def get_configs_info() -> dict:
     }
 
 
-@document_fix_compatible
+# These two decorators allow plugins
+# to be displayed in the sqlfluff docs
 @document_configuration
+@document_fix_compatible
```
barrywhart, line 43: Nitpick: Is there a reason for moving `@document_fix_compatible` so it follows `@document_configuration` rather than preceding it?

src/sqlfluff/core/rules/__init__.py
barrywhart: It feels a little strange for the core code to have to do this. Is this code here mainly to fix the documentation, or does it serve some other purpose?

Is there potentially another way to do this? For example, if it's mainly for the docs, could we `pip uninstall` the example plugin before building the docs?

barrywhart (approved):
A couple of questions to consider, but good to merge as is.

NiallRees:
I'll test this in the morning on my machine before approving if ok 👍 but looks good.

src/sqlfluff/core/rules/__init__.py
alanmcruickshank: yeah I agree - I'm also not sure this is required 🤔 . readthedocs checks out the repo and just installs sqlfluff directly. It doesn't use the `tox.ini` file, and that's the only place I cans see that we install `plugins/sqlfluff-plugin-example`. My guess is that even if we remove this we still wont see the example rule in the docs.

alanmcruickshank (changes requested):
I think we can remove the filter for example rules and still be fine. Happy to just fix that problem when we get to it if I'm wrong.

plugins/sqlfluff-plugin-example/src/example/rules.py
dmateusp: I thought it did, but it didn't! I swapped it back

dmateusp:
Removed the Rule_Example filter.


@alanmcruickshank how are the docs being pushed to the website today? I'm not familiar with sphynx and I couldn't find it in the code

NiallRees (approved):
https://docs.sqlfluff.com/en/dmateusp-fix_documentation_issue_introduced_by_plugin_work/rules.html all working, thanks @dmateusp 🙌 

--edit build link removed, see https://docs.sqlfluff.com/en/latest/rules.html

PR #825, opened 2021-03-01
Dialect Inheritance + MySQL Parsing.

### LINKED ISSUE #793: More intelligent dialect inheritance
At the moment, the only way to have dialect inheritance is to _replace_ particular segments with new definitions. This _works_ but leads to a lot of boilerplate, and lots of boilerplate leads to issues like #643 .

This issue is to implement more intelligent `super()` like functionality to do better dialect inheritance.

Ideas so far (ft. @barrywhart ):
- Similar to the super() feature with classes, could we provide support for dialects to extend segments of a parent dialect rather than "replace wholesale"?
- About the "super()" idea, could it work similarly to "LintFix", where you specify what to delete or edit, or what to add (and where)? The"where" could be specified using "Ref()", e.g. create after "Ref("SelectSegment")".

### LINKED ISSUE #814: Error linting MySQL create table statements
## Expected Behaviour
The MySQL schema is parsed correctly, with lint issues pointed out where applicable.

## Observed Behaviour
Parsing fails, even with explicit `--dialect` and `--templater` settings.

### Schema
```sql
CREATE TABLE `foo` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```

### Options
```
--templater raw --dialect mysql
```

### Lint output
```
L:   1 | P:   1 |  PRS | Found unparsable section: 'CREATE TABLE `foo` (\n `id`
                       | int(11) unsi...'
```

### Parse output
```
[0](1, 1, 1)        |file:
[0](1, 1, 1)        |    statement:
[0](1, 1, 1)        |        unparsable:                                           !! Expected: 'statement'
[0](1, 1, 1)        |            raw:                                              'CREATE'
[6](1, 1, 7)        |            whitespace:                                       ' '
[7](1, 1, 8)        |            raw:                                              'TABLE'
[12](1, 1, 13)      |            whitespace:                                       ' '
[13](1, 1, 14)      |            raw:                                              '`foo`'
[18](1, 1, 19)      |            whitespace:                                       ' '
[19](1, 1, 20)      |            start_bracket:                                    '('
[20](1, 1, 21)      |            newline:                                          '\n'
[21](1, 2, 1)       |            whitespace:                                       '  '
[23](1, 2, 3)       |            raw:                                              '`id`'
[27](1, 2, 7)       |            whitespace:                                       ' '
[28](1, 2, 8)       |            raw:                                              'int'
[31](1, 2, 11)      |            start_bracket:                                    '('
[32](1, 2, 12)      |            raw:                                              '11'
[34](1, 2, 14)      |            end_bracket:                                      ')'
[35](1, 2, 15)      |            whitespace:                                       ' '
[36](1, 2, 16)      |            raw:                                              'unsigned'
[44](1, 2, 24)      |            whitespace:                                       ' '
[45](1, 2, 25)      |            raw:                                              'NOT'
[48](1, 2, 28)      |            whitespace:                                       ' '
[49](1, 2, 29)      |            raw:                                              'NULL'
[53](1, 2, 33)      |            whitespace:                                       ' '
[54](1, 2, 34)      |            raw:                                              'AUTO_INCREMENT'
[68](1, 2, 48)      |            comma:                                            ','
[69](1, 2, 49)      |            newline:                                          '\n'
[70](1, 3, 1)       |            whitespace:                                       '  '
[72](1, 3, 3)       |            raw:                                              'PRIMARY'
[79](1, 3, 10)      |            whitespace:                                       ' '
[80](1, 3, 11)      |            raw:                                              'KEY'
[83](1, 3, 14)      |            whitespace:                                       ' '
[84](1, 3, 15)      |            start_bracket:                                    '('
[85](1, 3, 16)      |            raw:                                              '`id`'
[89](1, 3, 20)      |            end_bracket:                                      ')'
[90](1, 3, 21)      |            newline:                                          '\n'
[91](1, 4, 1)       |            end_bracket:                                      ')'
[92](1, 4, 2)       |            whitespace:                                       ' '
[93](1, 4, 3)       |            raw:                                              'ENGINE'
[99](1, 4, 9)       |            raw:                                              '='
[100](1, 4, 10)     |            raw:                                              'InnoDB'
[106](1, 4, 16)     |            whitespace:                                       ' '
[107](1, 4, 17)     |            raw:                                              'DEFAULT'
[114](1, 4, 24)     |            whitespace:                                       ' '
[115](1, 4, 25)     |            raw:                                              'CHARSET'
[122](1, 4, 32)     |            raw:                                              '='
[123](1, 4, 33)     |            raw:                                              'utf8mb4'
[130](1, 4, 40)     |            whitespace:                                       ' '
[131](1, 4, 41)     |            raw:                                              'COLLATE'
[138](1, 4, 48)     |            raw:                                              '='
[139](1, 4, 49)     |            raw:                                              'utf8mb4_unicode_ci'
[157](1, 4, 67)     |    statement_terminator:                                     ';'
[158](1, 4, 68)     |    newline:                                                  '\n'

==== parsing violations ====
L:   1 | P:   1 |  PRS | Found unparsable section: 'CREATE TABLE `foo` (\n `id` int(11)
                       | unsi...'
```

### Fix output
```
```

## Version
sqlfluff, version 0.4.0

## Configuration
```
N/A, not using any configuration files
```

alanmcruickshank:
This resolves #814 and potentially resolves #793.

This opens up a few different ways for doing inheritance.
- You can still just completely replace a segment, this is good if enough is different that it would be just simpler to do this than the other options (like the biquery `DatatypeSegment`).
- You can inherit a non-parent version of a segment and drop it straight into the dialect (like the snowflake `SelectClauseModifierSegment`, in this case postgres modifies it and here we go back to the ansi version).
- You can inherit *and modify* the grammar of a previous grammar by _inserting_ or _removing_ elements (like in the snowflake `StatementSegment`)

A couple of discussion points:
1. In the issue, we discussed a `super()` _-like_ approach, and this approach is more verbose than that. Here, we have to explicitly say which grammar and segment we want to inherit from. It's more verbose, but I think the alternative, while more concise, would just be syntactic sugar on top of this and might actually be _less clear_. Explicit is better than implicit.
2. This approach currently only allows operations on the top level of the grammar referenced. If you wanted to do more complicated edits, deeper in the grammar structure, you would have to just replace the whole Grammar. This isn't a complete blocker, because it encourages us to break up large grammars into more meaningful chunks, and avoids very complex (and probably fragile) specifications. Downside is that we encourage more duplication than strictly necessary. _This could be remidied_ by allowing a more complex selector syntax in the `.copy()` method to specify exactly where the edits would happen, but I'm not sure whether it's worth the extra complexity at this stage.
3. Where should this approach and the guidance on when to use each method be documented? I feel like we need a _developer reference_ section of the docs which doesn't really exist yet. Alternative would be a `README.md` in the dialects folder.

Otherwise I'm curious for what people think to this approach. I've updated the most common dialects to make best use of the new syntax, but haven't yet touched teradata or exasol. I suggest I do them either as a separate PR or at least after a first pass of reviewing this PR first.


src/sqlfluff/core/dialects/dialect_bigquery.py
```diff
@@ -186,8 +182,6 @@ class SelectTargetElementSegment(AnsiSelectTargetElementSegment):
         type="function_name",
         _anti_template=r"STRUCT",
     ),
-    SelectTargetElementSegment=SelectTargetElementSegment,
```
barrywhart, line 189: Why were we able to remove these?

src/sqlfluff/core/dialects/dialect_bigquery.py
```diff
@@ -163,15 +165,9 @@ class SelectTargetElementSegment(AnsiSelectTargetElementSegment):
...
-    LiteralGrammar=OneOf(
-        Ref("QuotedLiteralSegment"),
-        Ref("DoubleQuotedLiteralSegment"),
-        Ref("NumericLiteralSegment"),
-        Ref("BooleanLiteralGrammar"),
-        Ref("QualifiedNumericLiteralSegment"),
-        Ref("NullKeywordSegment"),
-        Ref("LiteralCoercionSegment"),
+    # Add two elements to the ansi LiteralGrammar
+    LiteralGrammar=ansi_dialect.get("LiteralGrammar").copy(
```
barrywhart, line 169: Nice!

src/sqlfluff/core/dialects/dialect_mysql.py
```diff
@@ -4,9 +4,12 @@
 https://dev.mysql.com/doc/refman/8.0/en/differences-from-ansi.html
 """
 
-from sqlfluff.core.parser import NamedSegment, OneOf, Ref
+from sqlfluff.core.parser import NamedSegment, Ref, AnyNumberOf, Sequence, OneOf
 
-from sqlfluff.core.dialects.dialect_ansi import ansi_dialect
+from sqlfluff.core.dialects.dialect_ansi import (
+    ansi_dialect,
+    CreateTableStatementSegment as ansi_CreateTableStatementSegment,
```
barrywhart, line 11: Possible alternative for this:
```
import sqlfluff.core.dialects.dialect_ansi as ansi 
```

Then, later in the code, you could reference `ansi.CreateTableStatementSegment` (dot, rather than underscore). This is shorter and avoids the slight risk of typos when copying the name `CreateTableStatementSegment` -> `ansi_ CreateTableStatementSegment`.

src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -632,6 +632,84 @@ def __repr__(self):
             ),
         )
 
+    def __eq__(self, other):
+        """Two grammars are equal if their elements and types are equal.
+
+        NOTE: This could potentially mean that two grammars with
+        the same elements but _different configuration_ will be
```
barrywhart, line 639: Can you add an example or quick definition of what "configuration" means in this context?

src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -632,6 +632,84 @@ def __repr__(self):
...
+
+
+        Args:
+            insert (:obj:`list`, optional): Matchable elements to
+                insert. This is inserted pre-expansion so can include
+                unexpanded elements as normal.
+            at (:obj:`int`, optional): The position in the elements
+                to insert the item. Defaults to `None` which means
+                insert at the end of the elements.
+            before (optional): An alternative to _at_ to determine the
```
barrywhart, line 664: Why is this `Optional[Any]` rather than `Optional[int]` like `at`? I have an alternate idea that _may_ be clearer... Could we eliminate the `before` option and pass a _negative number_ to `at` for this purpose?

barrywhart (commented):
Great work! I love the simplicity and improved maintainability.

I made a few suggestions, some to trigger discussion; I'm not necessarily saying "this is the way". 

image

src/sqlfluff/core/dialects/dialect_bigquery.py
alanmcruickshank: the `@bigquery_dialect.segment(replace=True)` decorator automatically adds it so that you don't also have to do this.

src/sqlfluff/core/dialects/dialect_mysql.py
alanmcruickshank: Totally agree. 👍

src/sqlfluff/core/parser/grammar/base.py
alanmcruickshank: Yeah good point - it's basically anything that isn't part of the elements, e.g. `optional`

src/sqlfluff/core/parser/grammar/base.py
alanmcruickshank: On the datatype - not sure, I think that might just be an oversight. 🤔 

On the `before` option - at the moment you can use `after` with a negative number and it will insert it after _an index from the end_. I might see if I can do something similar to what you suggest though...

alanmcruickshank: Ah - no - this *is* intentional!

`before` allows you to provide a matching element to look for, in case you don't know it's index. I'll provide an example to show that.

src/sqlfluff/core/dialects/dialect_snowflake.py
```diff
@@ -144,29 +140,23 @@
...
-        Ref("CreateCloneStatementSegment"),
-        Ref("ShowStatementSegment"),
+    parse_grammar = ansi.StatementSegment.parse_grammar.copy(
+        insert=[
+            Ref("UseStatementSegment"),
+            Ref("CreateStatementSegment"),
+            Ref("CreateCloneStatementSegment"),
+            Ref("ShowStatementSegment"),
+        ],
+        remove=[
```
barrywhart, line 153: Thinking ahead a bit, places where we use `remove` might suggest things we'd want to remove from the base dialect. (I think several of these removals are Postgres-specific stuff.) Think it's worth creating an issue for this?

src/sqlfluff/core/dialects/dialect_snowflake.py
```diff
@@ -388,7 +372,7 @@ class QualifyClauseSegment(BaseSegment):
 
 
 @snowflake_dialect.segment(replace=True)
-class SelectStatementSegment(ansi_SelectClauseSegment):
+class SelectStatementSegment(ansi.SelectStatementSegment):
```
barrywhart, line 375: Why did the base class change from `SelectClauseSegment` to `SelectStatementSegment`?

src/sqlfluff/core/dialects/dialect_snowflake.py
alanmcruickshank: I'm tempted to make those issues when they come up, but in this case I think it's set up how it should be. It's frustrating that snowflake takes *most* of it's queues from postgres, but does remove some elements too. 🤷

barrywhart: I agree, low priority, longer term stuff. 👍

src/sqlfluff/core/dialects/base.py
```diff
@@ -153,6 +153,13 @@ def replace(self, **kwargs):
                 )
             self._library[n] = kwargs[n]
 
+    def get(self, name):
```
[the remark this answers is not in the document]
alanmcruickshank: Lol - Turns out this was a much harder problem to address. The issue was that this could return almost anything and validating that was a PITA.

I've split this into two methods - one for grammars and one for segments so that you can at least infer what you're getting.

src/sqlfluff/core/dialects/dialect_snowflake.py
alanmcruickshank: it was the wrong class in the first place, but because it wasn't registered properly nobody had noticed 🤷

alanmcruickshank:
I'm really stuck. Anyone got any bright ideas about why this test is failing?

barrywhart:
@alanmcruickshank: Could it be a circular chain of imports? I have often seen these odd errors in that case.

alanmcruickshank:
Ok this has turned into a much bigger job. I started to unpick the potential circular dependencies that @barrywhart mentioned and I realized with the new plugin work and the extensions to dialect inheritance I've been working on - the dependency situation for dialects had become very fragile.

My most recent commits here make some pretty big changes to the structure of dialects. Notably:
- I've extracted all the dialects from the core module.
- Dialects are now loaded at runtime (somewhat similar to the approach we took with rules).
- `Dialect.expand()` now returns a *copy* rather than modifying the dialect in place.
- All references to dialects are now via `dialect_selector` rather than through importing them directly.

While this is a massive change I think it's a direction I'd always imagined that we could go in but hadn't had the vision on how. This should open up the door to potentially having pluggable dialects in future.

I'd appreciate a sense check that this approach makes sense. I realize it's a now a `MEGACHONKER` PR.

CHANGELOG.md
```diff
@@ -6,11 +6,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 ### Changed
 
 - Renamed the BaseCrawler class to BaseRule. This is the base class for all
   rules. This is a breaking change for any custom rules that have been added
   via plugins or by forking the SQLFluff repo.
+- Renamed `sqlfluff.rules()` to `sqlfluff.list_rules()` and `sqlfluff.dialects()`
+  to `sqlfluff.list_dialects()` due to naming conflicts with the now separate
+  `sqlfluff.dialects` module.
+- Extracted dialect definitions from the `sqlfluff.core` module so that each
+  dialect is better isolated from eachother. This also allows more focussed
```
barrywhart, line 25: Nitpick/typo: eachother -> each other

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1269,7 +1260,7 @@ class CaseExpressionSegment(BaseSegment):
         ),
     ),
     # Expression_B_Grammar https://www.cockroachlabs.com/docs/v20.2/sql-grammar.htm#b_expr
-    Expression_B_Grammar=None,  # TODO
+    # Expression_B_Grammar=None,  # TODO
```
barrywhart, line 1263: Why is this commented out?

barrywhart (changes requested):

src/sqlfluff/dialects/dialect_ansi.py
alanmcruickshank: MyPy was complaining that `None` couldn't be in a dialect.

barrywhart: So the code still works okay? That was my main concern.

Re: The MyPy warning, can we (should we?) change a type hint, e.g. `Optional[Thing]` rather than just `Thing`?

pwildenhain (commented):
I didn't do an _exhaustive_ review, but overall I like this change. Definitely seems like it'll make adding new dialects easier. Seems similar to the change we made with putting all the rules in separate modules.

src/sqlfluff/dialects/dialect_ansi.py
alanmcruickshank: Yeah the code works fine. `Expression_B_Grammar` was never implemented and nothing refers to it. It's a remnant from when the expression structure was much closer to the cockroach one. We never actually needed it though. There's an argument for actually pulling it out altogether rather than just commenting it out - do you think that would be tidier?

barrywhart: Interesting... since several comments in the ANSI dialect expression grammar refer to corresponding parts of the Cockroach one, I'm in favor of keeping this line in place (but commented out as you have it here). I suggest adding a comment similar to what you said above, e.g.:
>Expression_B_Grammar is an unused remnant from CockroachDB, whose expression grammar was the original basis for SQLFluff's.

alanmcruickshank: I've updated the comment with some new wording to explain what's going on with Expression B.

> CockroachDB defines Expression_B_Grammar. The SQLFluff implementation of
> expression parsing pulls that logic into Expression_A_Grammar and so there's
> currently no need to define Expression_B.

barrywhart (approved):
:shipit:

PR #826, opened 2021-03-01
Don't build additional docs formats

alanmcruickshank:
Docs have been failing fairly regularly for the pdf and latex builds. This updates the config to just not build those versions.


NiallRees:
👍

PR #834, opened 2021-03-02
Issue 839: Lint rules can see file path; other tweaks to the example plugin

### LINKED ISSUE #839: Lint rules should have access to the file path of the file being linted
I'm working on an organization-specific rule (a plug-in) and realized that it could be very useful for rules to have access to the path of the file that is currently being linted.

Context: The SQL files in our projects fall into two categories, similar to dbt:
* Model files (These create output tables)
* Pre- or post-conditions (These enforce constraints on the data but do not create output tables.) These are typically found in a directory named `constraints/`, `postconditions/`, something like that.

The rule I'm developing should only apply to model files, so having access to the path is a pragmatic (admittedly org-specific and not 100% guaranteed) way to deduce the category.

Other rules may use this in other ways; this is just one real-world use case to justify the enhancement.

barrywhart:
Resolves #839 

I'm working on a real-world plugin for my employer. This PR addresses the noted issue. It also includes some othersmall changes I think will streamline plugin creation a bit. 

- [X] Update the `install_requires` SQLFluff version to an official (not alpha) release
- [X] Template the plugin name (`example`) to make it easier to change
- [X] Use `find_packages()` so deeper module directory structures will "just work"

### DIFF
--- a/plugins/sqlfluff-plugin-example/setup.py
+++ b/plugins/sqlfluff-plugin-example/setup.py
@@ -5,7 +5,7 @@ setup(
     name="sqlfluff-plugin-example",
     include_package_data=True,
     package_dir={"": "src"},
-    packages=find_packages(),
+    packages=find_packages(where="src"),
     install_requires="sqlfluff>=0.4.0",
     entry_points={"sqlfluff": ["example = example.rules"]},
 )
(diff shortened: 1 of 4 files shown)


dmateusp (approved):
Thanks for the real world testing on this :)

PR #836, opened 2021-03-02
Issue 835: Fix BigQuery parsing error with ">" or "<" inside parentheses

### LINKED ISSUE #835: BigQuery parsing error with "<" or ">" operator in a CTE
Query:
```
WITH filtered_ecomm_orders as (
    SELECT
        *
    FROM `mc-business-intelligence.user.ecommerce_orders`
    WHERE
        order_total > 0
)

SELECT * FROM filtered_ecomm_orders
```

Much simpler query with same issue:
```
SELECT (1 > 0)
```

I strongly believe this bug is related to the changes in #325, where we changed the scanner/parser to support BigQuery UDFs and the related angle bracket syntax. Why do I think this?
* If I pull out the CTE as a standalone query, it parses okay.
* If I change the `>` to `<`, I get a different error: `L:   1 | P:  31 |  PRS | Couldn't find closing bracket for opening bracket.`
* If I change it to `=`, the query parses okay.

Command:
```
$ sqlfluff parse --dialect bigquery test.sql
```

## Expected Behaviour

Successful parse

## Observed Behaviour

Parse error:
```
...Failed to Parse...
==== parsing violations ====
L:   7 | P:   1 |  PRS | Found unexpected end bracket!, was expecting one of: [<Ref:
                       | SemicolonSegment>], but got <class
                       | 'sqlfluff.core.parser.segments.raw.end_bracket_SymbolSegment'>
```

## Version
SQLFluff: 0.4.1
Python version: 3.7.9

## Configuration
```
[sqlfluff]
verbose = 0
nocolor = False
dialect = bigquery
templater = jinja
rules = None
exclude_rules = L029,L031,L032,L034
recurse = 0

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 120

[sqlfluff:rules:L010]

[sqlfluff:rules:L014]

[sqlfluff:rules:L030]

# Settings in this section will be provided to SQL templates during the SQL
# formatting/linting process. Generally, the exact values don't matter as long
# as they result in valid SQL syntax when substituted in the file.
```

barrywhart:
Resolves #835 

- [x] Update the parse output `.yml` file once the parse is working correctly
- [x] Add a similar test case using `<`
- [x] Add more test cases with a mixture of parentheses and comparison operators to see if we can surface any more issues.


src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -515,13 +515,53 @@ class BracketInfo(NamedTuple):
...
+                                    # no, assume it was not a bracket.
+                                    # Can we remove any brackets from the stack which aren't definites
+                                    # to resolve the issue?
+                                    for idx in range(len(bracket_stack) - 1, -1, -1):
+                                        if not bracket_stack[idx].is_definite:
+                                            del bracket_stack[idx]
+                                            # We don't change the string buffer, we assume that was ok.
+                                            break
+                                    else:
+                                        raise SQLParseError(  # pragma: no cover
```
barrywhart, line 560: This line is not covered by tests. I don't know how to create a test that hits it, but I'd feel uncomfortable removing this code because it will help with error reporting if we **do** encounter this situation someday. Any thoughts or concerns?

dmateusp: Let me try to add something to hit this, I guess we should have some test to verify that we do throw an exception when brackets aren't balanced

dmateusp: https://github.com/barrywhart/sqlfluff/pull/5 maybe that would do it?

barrywhart:
@dmateusp: Ok, the PR is updated to cover that previously untested line. Thanks for the help!!

test/core/parser/grammar_test.py
```diff
@@ -197,6 +197,38 @@ def test__parser__grammar__base__bracket_sensitive_look_ahead_match(
         assert match.matched_segments == (fs("foo", bracket_seg_list[8].pos_marker),)
 
 
+def test__parser__grammar__base__bracket_fail_with_open_paren_close_square_mismatch(
+    generate_test_segments, fresh_ansi_dialect
+):
+    """
+    Test that _bracket_sensitive_look_ahead_match fails when the type of a
+    close bracket doesn't match the type of the corresponding open bracket, but
+    both are "definite" brackets.
```
dmateusp: that was so specific haha, very good catch

dmateusp (approved):

PR #838, opened 2021-03-02
PascalCase capitalisation_policy and new unquoted_identifiers_policy

### LINKED ISSUE #820: Add PascalCase capitalisation_policy
[L014](https://docs.sqlfluff.com/en/stable/rules.html#sqlfluff.core.rules.std.Rule_L014) (Inconsistent capitalisation of unquoted identifiers) needs a PascalCase capitalisation_policy option, implemented as follows:

- Unquoted identifier does not contain non-alphanumeric characters
- Unquoted identifier begins with a capital letter

### LINKED ISSUE #821: L014 needs only_aliases configuration options like L029 has
[L014](https://docs.sqlfluff.com/en/stable/rules.html#sqlfluff.core.rules.std.Rule_L014) (Inconsistent capitalisation of unquoted identifiers) needs an `only_alises` (True/False) configuration option like [L029](https://docs.sqlfluff.com/en/stable/rules.html#sqlfluff.core.rules.std.Rule_L029) already has that would make the rule apply only to aliases.

The reason for this is because we can't always control the capitalization policies of the source databases, and would like to preserve their original capitalization where possible.

panasenco:
## PascalCase capitalisation_policy

Fixes #820. I rewrote `L010.py` from scratch to be able to handle ambiguity in identifier cases.

For example, this new test didn't pass with the previous version of L010 that assumed that a single letter uppercase identifier was necessarily uppercase:
```
test_consistency_6:
  # Single-letter ambiguity: Upper vs Capitalise
  pass_str: SELECT A, Boo
```

Now L010 uses a completely different algorithm based on **refutation of inconsistent cases** rather than guessing cases. For example, if we see an identifier named `IRB`, we know that we aren't dealing with `lower_case` or `Capitalise`, but we don't know for sure whether the identifier is `UPPER_CASE` or `PascalCase` because PascalCase could allow capitalised acronyms.

The new L010 algorithm only begins throwing linting errors when one of the following conditions is met:
  * The policy is `consistent` and all allowed cases are refuted
  * The exact policy specified is one of the refuted cases

I'm very excited about this new algorithm because it easily allows for more capitalisation policies to be implemented in the future like `camelCase` or `Pascal_Snake_Case`.

## unquoted_identifiers_policy

Fixes #821. The existing `only_aliases` policy was too inflexible for my intended use in L014. I only want to lint aliases outside of FROM expressions and leave table aliasing to other rules. Therefore I replaced `only_aliases` with `unquoted_identifiers_policy` with the options [`all`, `aliases`, `column_aliases`] for both L014 and L029. The function `unquoted_ids_policy_applicable` in `L014.py` allows other policies to be easily added in the future and allows other rules to easily take advantage of this configuration option.

## Breaking Changes - Please Review
  * The configuration option `capitalisation_policy` for rule L014 was replaced with a new configuration `extended_capitalisation_policy`. This will break existing configs. The reason for the differentiation is that we don't want to allow PascalCase to be a valid option for keywords (`SeLeCt` should not be valid).
  * The configuration option `only_aliases` was replaced with `unquoted_identifiers_policy` for reasons explained above. This will also break configs.

I'm open to suggestions about how to minimize these breaking effects.


src/sqlfluff/core/rules/std/L010.py
```diff
@@ -45,99 +47,134 @@ class Rule_L010(BaseCrawler):
...
+        ]
+        self.logger.debug(
+            f"Selected '{cap_policy_name}': '{cap_policy}' from options "
+            f"{cap_policy_opts}"
+        )
+
+        refuted_cases = memory.get("refuted_cases", set())
+
+        # Which cases are definitely inconsistent with the segment?
+        if segment.raw[0] != segment.raw[0].upper():
```
dmateusp, line 82: nice, seems like going by elimination does make this simpler :)

src/sqlfluff/core/rules/std/L010.py
```diff
@@ -45,99 +47,134 @@ class Rule_L010(BaseCrawler):
...
+        memory["refuted_cases"] = refuted_cases
+
+        self.logger.debug(
+            f"Refuted cases after segment '{segment.raw}': {refuted_cases}"
+        )
+
+        # Skip if no inconsistencies, otherwise compute a concrete policy
+        # to convert to.
+        if cap_policy == "consistent":
+            possible_cases = [c for c in cap_policy_opts if c not in refuted_cases]
```
dmateusp, line 105: nit: `set(cap_policy_opts).difference(refuted_cases)` up to you

src/sqlfluff/core/rules/std/L010.py
```diff
@@ -45,99 +47,134 @@ class Rule_L010(BaseCrawler):
...
+                    f"Setting concrete policy '{concrete_policy}' from cap_policy"
+                )
 
-        # If it's not a keyword just carry on
-        return LintResult(memory=memory)
+        # We need to change the segment to match the concrete policy
+        if concrete_policy in ["upper", "lower", "capitalise"]:
+            if "pascal" in cap_policy_opts and segment.raw[0].isupper():
+                # Insert undescores in transitions between lower and upper
+                fixed_raw = re.sub("(?<=[a-z0-9])(?=[A-Z])", "_", segment.raw)
```
dmateusp, line 139: Am I understanding this correctly? Is this adding underscores for pascal case?

src/sqlfluff/core/rules/std/L029.py
```diff
@@ -29,21 +30,17 @@ class Rule_L029(BaseCrawler):
...
-                return LintResult(anchor=segment)
+        if (
+            segment.name == "naked_identifier"
+            and unquoted_ids_policy_applicable(
+                self.unquoted_identifiers_policy, parent_stack
+            )
+            and (segment.raw.upper() in dialect.sets("unreserved_keywords"))
+        ):
+            return LintResult(anchor=segment)
+        else:
```
dmateusp, line 45: nit: this else: return None isn't needed but if you prefer being explicit it's fine for me :)

dmateusp (commented):
For 

> The configuration option capitalisation_policy for rule L014 was replaced with a new configuration extended_capitalisation_policy. This will break existing configs. The reason for the differentiation is that we don't want to allow PascalCase to be a valid option for keywords (SeLeCt should not be valid).

I think we should have a check in the rule L014 that goes like:

```python
if self.capitalisation_policy:
  raise RuntimeError("The 'capitalisation_policy' config for L014 was replaced with 'extended_capitalisation_policy', please rename this configuration key. See <link to the note in Changelog> for more details.")
```

dmateusp:
hey @panasenco this is phenomenal 🎉 , thanks for contributing! I'd like to get a few more eyes on this to discuss how to handle breaking changes and bring some awareness to this refactoring. I added the maintainers as reviewers.

src/sqlfluff/core/rules/std/L029.py
barrywhart: I tend to like explicit `return` statements. In my recent PRs, I've been adding MyPy type hints to several areas of the code, and IIRC, either PyCharm (my Python IDE) and/or MyPy complains if you don't explicitly return a value matching the function's return type.

src/sqlfluff/core/rules/std/L010.py
```diff
@@ -45,99 +47,134 @@ class Rule_L010(BaseCrawler):
...
     config_keywords = ["capitalisation_policy"]
 
-    def _eval(self, segment, memory, **kwargs):
+    def _eval(self, segment, memory, parent_stack, **kwargs):
         """Inconsistent capitalisation of keywords.
 
-        We use the `memory` feature here to keep track of
-        what we've seen in the past.
+        We use the `memory` feature here to keep track of cases known to be
+        INconsistent with what we've seen so far as well as the top choice
```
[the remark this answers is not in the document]
panasenco: I meant to highlight the "IN" in "inconsistent" but I guess it just came out bad. I'll just put asterisks around the whole word.

src/sqlfluff/core/rules/std/L010.py
panasenco: Will use that!

src/sqlfluff/core/rules/std/L010.py
panasenco: This segment is the first step in auto-converting PascalCase to one of the other cases. Only PascalCase so far has legal transitions between a lowercase and an uppercase character, so it's always safe to insert an underscore where we see such a transition and the first character of the identifier is uppercase.

This helps the below tests pass:

```
test_pascal_10:
  fail_str: SELECT PasCalCase
  fix_str: SELECT pas_cal_case
  configs:
    rules:
      L014:
        extended_capitalisation_policy: lower

test_pascal_11:
  fail_str: SELECT PascalCaseNAME
  fix_str: SELECT PASCAL_CASE_NAME
  configs:
    rules:
      L014:
        extended_capitalisation_policy: upper
```

src/sqlfluff/core/rules/std/L029.py
panasenco: `return None` in sqlfluff seems to have a meaning that's essentially equivalent to `return SkipThis` so it's nice to be explicit about it for the reader's sake.

src/sqlfluff/core/rules/std/L010.py
panasenco: Actually I instead added an additional test that wouldn't pass with your approach.

```
test_consistency_9:
  # Test that UPPER case is chosen first in case of ambiguity
  fail_str: SELECT U, what_Is_This_Case
  fix_str: SELECT U, WHAT_IS_THIS_CASE
```

Basically in case of ambiguity preference should be in the order specified in [`config_info.py`](https://github.com/sqlfluff/sqlfluff/blob/master/src/sqlfluff/core/rules/config_info.py#L53): `["consistent", "upper", "lower", "pascal", "capitalise"]`. In cases of ambiguity, `upper` should take precedence over `pascal` and `capitalise`. 

However, set differences are ordered alphabetically, so this ordering would be lost.

alanmcruickshank (approved):
Initially I wasn't sure about the change to `extended_capitalisation_policy`, but on fully reviewing everything this looks like this is a really well thought through PR. Good work 👍

alanmcruickshank:
@sqlfluff/sqlfluff-maintainers - I suggest we release these changes with 0.5.0 (albeit that it will be a smaller release than 0.4.0). That way we handle the potentially breaking changes.

SQLFluff is still a "beta" project so I'm happy to be a little more bullish with the breaking changes as long as it's not part of a minor release.

PR #860, opened 2021-03-13
Rule L047: use consistent syntax for row count

### LINKED ISSUE #853: Rule: consistent syntax for COUNT(*) / COUNT(1) -> COUNT ALL
I came across this post on LinkedIn: https://www.linkedin.com/posts/eczachly_dataengineering-activity-6775447325725769728-EhK3 advocating to always use `COUNT(1)` vs `COUNT(*)`.

My personal opinion, not that it matters in the context of implementation of the rule, is that both syntaxes are a way to say "COUNT all rows in this query". In many (most?) cases it seems like the engine optimizers just make these queries equivalent.

There are many "urban myths", for example:

1. `COUNT(*)` is worse because `*` is actually pulling the full row (which would be painful for columnar databases)

But in this [Stack Overflow answer](https://stackoverflow.com/a/1221649) you see the [ANSI-92 spec](http://msdn.microsoft.com/en-us/library/ms175997.aspx) quoted `If COUNT(*) is specified, then the result is the cardinality of T.`

2. `COUNT(1)` is worse because it evaluates the expression "1" against all of the rows in the table

In the same Stack Overflow answer: `COUNT(1) has been optimized out by RDBMS vendors because of this superstition.`

Now, the only way to know **know** would be to see the plan that each database gives you for these queries, or/and actually compare performance, since most of these engines are closed source.

I actually had a look into the Spark code, and there was [this interesting comment](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala#L1795):

> [...] while count(*) will be converted to count(1)

So my conclusion from this all:
1. An issue that people have strong opinions about.
2. Something that does not seem to actually have an impact on performance.
3. A great candidate for a rule.

I would suggest a simple rule that enforces consistency of either `COUNT(*)` or `COUNT(1)` across the code base (configurable), and `COUNT(*)` by default since it is specifically mentioned in the ANSI docs.

dmateusp:
Resolves #853


barrywhart:
I think I have a PR still open from a few weeks ago that creates rule numbers 44 and 45. Would you be okay renumbering this rule?

I've gotten a few comments on that PR, but no full reviews yet. 😢

src/sqlfluff/core/rules/std/L044.py
```diff
@@ -0,0 +1,83 @@
...
+
+    """
+
+    config_keywords = ["prefer_count_1"]
+
+    def _eval(self, segment, **kwargs):
+        """Find rule violations and provide fixes."""
+        if segment.is_type("function"):
+
+            if self.prefer_count_1 and segment.raw_upper == "COUNT(*)":
```
barrywhart, line 57: Will this work as intended if there are spaces, e.g. `COUNT( * )`?

barrywhart (changes requested):
Question about handling of whitespace

src/sqlfluff/core/rules/std/L044.py
dmateusp: Now it should handle counts that have whitespaces, thanks for catching that

test/core/rules/test_cases/L046.yml
```diff
@@ -0,0 +1,85 @@
...
+
+  fix_str: |
+    select
+        foo,
+        count(1)
+    from my_table
+    group by
+      foo
+
+  configs: *prefer_count_1
```
NiallRees, line 77: Oooo

NiallRees (approved):

src/sqlfluff/core/rules/std/L046.py
```diff
@@ -0,0 +1,85 @@
...
+
+    """
+
+    config_keywords = ["prefer_count_1"]
+
+    def _eval(self, segment, **kwargs):
+        """Find rule violations and provide fixes."""
+        if segment.is_type("function"):
+
+            raw_upper_no_whitespaces = segment.raw_upper.replace(" ", "")
```
barrywhart, line 57: Do we need to worry about other kinds of whitespace, e.g. tab characters or newlines? (Admittedly wacky cases, just trying to think of things that may break.)

barrywhart (approved):
I thought of another couple of potentially tricky whitespace cases: tabs or newlines. It's pretty obscure, so you can decide whether this is worth addressing. Overall, looks great!

src/sqlfluff/core/rules/std/L046.py
dmateusp: Maybe we do, I'll take a look

test/core/rules/test_cases/L046.yml
alanmcruickshank: > The yaml-fu is strong with this one.

src/sqlfluff/core/rules/std/L046.py
alanmcruickshank: We probably do, but we could easily get to that if we first look for `FunctionNameSegment` and see if it's `COUNT` and then iterate though the remaining children, keep everything between the brackets but filter out segments of type `whitespace` and type `newline`. That could also simplify the fixing logic because you'll already have a reference to whatever is in the bracket.

alanmcruickshank (changes requested):

PR #863, opened 2021-03-14
Fix variable introspection for test helpers in sqlfluff.testing

### LINKED ISSUE #862: No variable introspection on test failure
Since I refactored the test helper code to live in the `testing` module, we lost variable introspection. On rule test case failure we are not getting the string that failed anymore but just `AssertionError`.

The same issue as https://stackoverflow.com/questions/53001496/pytest-assertion-doesnt-show-differences-on-assertionerror

dmateusp:
Resolves #862


requirements_dev.txt
```diff
@@ -11,6 +11,7 @@ coverage
 hypothesis
 pytest
 pytest-cov
+pytest-sugar
```
dmateusp, line 14: I was having a hard time with the default pytest display, this just replaces how the test results get displayed and gives you the exact tests that failed if you want to re-run them (instead of scrolling)

barrywhart (approved):
One of the CircleCI checks is failing. Otherwise looks good.

requirements_dev.txt
barrywhart: There is another pytest plugin (maybe pytest-memory?) which adds an `--lf` command-line option. It will automatically rerun the failing tests. I find it really useful.

barrywhart: I think it's called pytest-cache. Will post a link if I find it. I have no objections to your change, just thought you may like this too.

barrywhart: I think it may be built into pytest now, rather than being a plug-in.

NiallRees (approved):

PR #867, opened 2021-03-14
Issue 846: Fix parse issue with queries that use "UNION ALL" or "UNION DISTINCT" and lack "FROM" clauses

### LINKED ISSUE #846: Parse error when using "UNION DISTINCT" or "UNION ALL"
## Expected Behaviour

Successful parse of this query:
```
SELECT 0
UNION ALL
SELECT 1
```

## Observed Behaviour

```
sqlfluff parse test.sql
...
L:   1 | P:   9 |  PRS | Found unparsable section: '\nUNION ALL\nSELECT 1']
```

NOTE: It works if I remove the `ALL`, so I think using the optional qualifiers `ALL` or `DISTINCT` is triggering the issue.
## Version
SQLFluff version 0.4.1
Python version 3.7.3

## Configuration

No `.sqlfluff` file present

barrywhart:
Resolves #846


src/sqlfluff/core/dialects/dialect_ansi.py
```diff
@@ -983,7 +983,16 @@ class SelectClauseSegment(BaseSegment):
     type = "select_clause"
     match_grammar = StartsWith(
         Sequence("SELECT", Ref("WildcardExpressionSegment", optional=True)),
-        terminator=OneOf("FROM", "LIMIT", Ref("SetOperatorSegment")),
+        terminator=OneOf(
+            "FROM",
+            "LIMIT",
+            # :HACK: This Sequence("UNION", ... stuff should not be needed,
```
barrywhart, line 989: This is a hacky fix, but it works. Thoughts?

NiallRees: Hey @barrywhart would you mind pushing this to a branch on the sqlfluff repo? I'd like to clone it and take a look

alanmcruickshank (changes requested):
Right thinking, but I think we can avoid the hack. I think the issue here is that we need to extend `SetOperatorSegment` to include distinct. *most* of what you've added is already in `SetOperatorSegment`, it's just missing `DISTINCT`.

barrywhart:
@alanmcruickshank: Can you clarify what you have in mind? AFAICT, the extra line I added to `terminator` was already present in `SetOperatorSegment` -- I simply copied and added it to `terminator`. I definitely agree that it's cleaner to avoid the hacky change I made; I'm simply not understanding the alternative you're suggesting. Thanks!

alanmcruickshank:
@barrywhart - found the bug in the parsing logic. You were right, it was a sign of something very very deep in the simple matching heuristic. I think I've found it though and it should be fixed with my most recent commit.

alanmcruickshank (approved):

PR #868, opened 2021-03-14
Issue 861: Fix dbt templater removing trailing newline in source SQL

### LINKED ISSUE #861: dbt-templater doesn't respect trailing newline when exclude_rules = L009
## Expected Behaviour
If `exclude_rules = L009` is included in the `.sqlfluff` config when using `templater = dbt` , then any trailing newlines that are found at the end of a source dbt SQL model should be left as is when running `sqlfluff fix my_dbt_model.sql`. If no trailing newline exists, do nothing.

## Observed Behaviour
When `exclude_rules = L009` is included it the `.sqlfluff` config and using `templater = dbt`, running `sqlfluff fix my_dbt_model.sql` _removes_ any trailing newlines if they exist in the source dbt SQL models.

This is an undesirable outcome for the following reasons:
1. Setting `exclude_rules = L009` in `.sqlfluff` should result in leaving any existing trailing newlines (if they exist) _as is_. Right now, it is _actually enforcing the opposite of L009_ (i.e., no trailing newlines allowed), not _excluding the rule from consideration_ as the config option suggests.
2. Including a trailing newline at the end of files is best practice. Not including these will result in pesky `No newline at end of file` warnings in GitHub. I would _guess_ most users want to keep these trailing newlines

## Steps to Reproduce
Hard to share a _minimal_ example since the dbt templater uses the dbt compiler under the covers, so any example will require a dbt_project.

But the gist of the problem would look like...
A `.sqlfluff` like (per the docs)
```
[sqlfluff]
templater = dbt
exclude_rules = L009
```
A dbt model like `my_dbt_model.sql` (note the trailing newline at end of file)
```sql
select *

from {{ ref('foo_bar_model') }}


```
Running `sqlfluff fix my_dbt_model.sql` will result in (note the trailing newline has been removed!)
```sql
select *

from {{ ref('foo_bar_model') }}
```

## Version
```bash
▶ python --version
Python 3.7.3
```
```bash
▶ sqlfluff --version
sqlfluff, version 0.4.1
```

## Configuration
```
# For SQLFluff Rules reference, see:
# https://docs.sqlfluff.com/en/stable/rules.html#rules-reference
[sqlfluff]
verbose = 0
nocolor = False
dialect = postgres
templater = dbt
rules = None
exclude_rules = L034
recurse = 0
output_line_length = 80
runaway_limit = 10
ignore = parsing
ignore_templated_areas = True

[sqlfluff:indentation]
indented_joins = False
template_blocks_indent = True

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 80
indent_unit = space
comma_style = trailing
allow_scalar = True
only_aliases = True

# All SQLFluff rules can be found at: https://docs.sqlfluff.com/en/stable/rules.html#rules-reference
# When a rule in not listed below, we inherit the default behavior from above.
[sqlfluff:rules:L003]
lint_templated_tokens = True

capitalisation_policy = lower

capitalisation_policy = lower

[sqlfluff:rules:L016]
# Setting to True allows us to copy/paste long URLs as comments
ignore_comment_lines = True

capitalisation_policy = lower

[sqlfluff:rules:L038]

capitalisation_policy = lower

[sqlfluff:rules:L042]
forbid_subquery_in = join

```

GClunies:
Resolves #861.

## This PR:
- Adds logic to `dbt.py` to improve the handling of trailing newlines in dbt models when `templater = dbt` in `.sqlfluff`.
- This means users do not have to include `exclude_rules = L009` when using the dbt-templater. It will now respect trailing newlines and `L009` can be enforced correctly.

The full mapping of outcomes concerning trailing newlines (L009) when running `sql lint example_dbt_model.sql` and `sql fix example_dbt_model.sql` is shown in the table below. 

| `.sqlfluff` includes `exclude_rules = L009` | `example_dbt_model.sql` has trailing newline(s)? | Output of<br>`sqlfluff lint example_dbt_model.sql`         | Lint Acceptable?                                                                                          | Output of<br>`sqlfluff fix example_dbt_model.sql` | Fix Acceptable? |
|:----------------------:|:-------------------:|-------------------------------------------------|:-----------------------------------------------------------------------------------------------------------:|----------------------------------------|:-----------------:|
| NO                   | NO                | L009 \| Files must end with a trailing newline. | ✅      | Single trailing newline added.  | ✅      |
| NO                   | YES               | No newline lint error.                                          | ✅       | Newline(s) are kept as is.        | ✅      |
| YES                  | NO                | No newline lint error.                                          | ✅       | No newline added.                   | ✅      |
| YES                  | YES               | No newline lint error.                                          | ✅       | Newline(s) are kept as is.        | ✅      |


src/sqlfluff/core/templaters/dbt.py
```diff
@@ -248,9 +249,20 @@ def _unsafe_process(self, fname, in_str=None, config=None):
...
             )
 
-        raw_sliced, sliced_file, templated_sql = self.slice_file(
-            node.raw_sql, compiled_sql, config=config
-        )
+        exclude_rules = self.sqlfluff_config.__dict__["_configs"]["core"][
+            "exclude_rules"
+        ]
+
+        if re.search("L009", str(exclude_rules)):
```
barrywhart, line 256: Nitpick: I think you can simplify this as `if "L009" in str(exclude_rules):`

src/sqlfluff/core/templaters/dbt.py
```diff
@@ -248,9 +249,20 @@ def _unsafe_process(self, fname, in_str=None, config=None):
...
+            raw_sliced, sliced_file, templated_sql = self.slice_file(
```
barrywhart, line 257: I'd prefer to avoid duplicating this line of code. How about we add a line to compute a value for`trailing_newline` based on `exclude_rules`, then pass that value to `self.slice_file()`.

barrywhart (commented):
A few minor comments on code style, etc. I don't use dbt, so I probably won't be much help reviewing the overall question of "does this PR solve the dbt newline issue".

GClunies:
@barrywhart fixes made per your review!

barrywhart (commented):
Should we consider adding or updating automated tests for this change?

The code looks good to me, but we should get a review from someone who knows the dbt parts of SQLFluff. I am only familiar with the simpler jinja template support. Who can we ask, @GClunies? Were you working with someone in Slack before starting this?

GClunies:
Agree some tests are likely in order @barrywhart. My experience in writing tests is limited but I can look at some of the other tests already in place and make a first effort for this additions.

I think it would be best to ask @dmateusp for additional review, he was part of the original [slack thread](https://sqlfluff.slack.com/archives/C01GX4L213J/p1615591289066100).

Looking back on this thread, and considering the fix I have made here, we could automatically disabling L009 like you suggested in the thread. This would eliminate the edge case I raised above in this PR. 

Thoughts? 🤔

src/sqlfluff/core/templaters/python.py
```diff
@@ -305,6 +305,11 @@ def slice_file(
             else:
                 # If it's not equal, loop around
                 templated_str = new_templated_str
+            if trailing_newline:
+                # Used by dbt.py to determine if trailing newline needs to be
```
NiallRees, line 309: 1. As this is dbt-specific, it should go in `templaters/dbt.py`, before calling slice_file.

2. We should ensure that rather than blindly adding an end of file newline, we add the same number as were in the raw SQL.

NiallRees (changes requested):
Thanks for this @GClunies and welcome to the project! I think we're on the right lines here, just a couple of changes needed.

src/sqlfluff/core/templaters/python.py
dmateusp: Same remark on 2., we should have tests checking that:
1. If a model does not contain a new line originally, then the rule fails
2. If a model contains a new line originally, then the rule passes

dmateusp:
For the tests, have a look at https://github.com/sqlfluff/sqlfluff/blob/master/test/core/templaters/dbt_test.py

dmateusp:
Also, this is an amazing write-up on the PR 🥇

src/sqlfluff/core/templaters/python.py
GClunies: Sorry for the late response to this review @NiallRees and @dmateusp.

@NiallRees 
1. I think this is a case of my comment just causing confusion instead of helping (I plan to remove it).<br>`if trailing_newline:` has to be here since I am simply looking that the argument has been set to `True` when `slice_file()` is called in `dbt.py`(by default I have set it to `False`). This ensures `slice_file()` only considers any newline logic when using the dbt templater and disregards when using the Jinja templater (or wherever else it may be used). You can see how this plays out in `dbt.py` [here](https://github.com/GClunies/sqlfluff/blob/ad559848ce0591d02d713b43bc854c7cb4531367/src/sqlfluff/core/templaters/dbt.py#L255-L262).

2. Completely agree - we should add the same number as were in the raw SQL.

@dmateusp 
1. Agree
2. Agree

Thanks again for the review. I will try and make changes this week.

NiallRees: Hey @GClunies no worries, let me try again! I can see that you've added the default argument of False to trailing_newline, but as the only templater that would use the argument is the dbt templater, it makes sense for this logic to go inside the dbt templater. It doesn't need to be in slice_file from what I can see. The dbt templater can just add any new lines it needs to before handing over the raw and templated sql to slice_file. Make sense?

GClunies: Got it, will try this.

GClunies: @NiallRees 

I remember why I ended up handling this issue in `python.py` and not `dbt.py`

If I change [these lines](https://github.com/sqlfluff/sqlfluff/blob/72d3aafc754bc1ecde4682a6624e16d8a97f6c01/src/sqlfluff/core/templaters/dbt.py#L245-L264) in `dbt.py` to
```python
...
...
...
if not compiled_sql:
            raise SQLTemplaterError(
                "dbt templater compilation failed silently, check your configuration "
                "by running `dbt compile` directly."
            )

        raw_sliced, sliced_file, templated_sql = self.slice_file(
            node.raw_sql, compiled_sql, config=config
        )

       templated_sql = templated_sql + "\n"  # ADDING NEWLINE IN dbt.py AS SUGGESTED

        return (
            TemplatedFile(
                source_str=node.raw_sql,
                templated_str=templated_sql,
                fname=fname,
                sliced_file=sliced_file,
                raw_sliced=raw_sliced,
            ),
            # No violations returned in this way.
            [],
        )
```

I get the following error when I run `sqlfluff fix`
```bash
▶ sqlfluff fix models/analytics/editorial/sl_web_editorial_time_on_page.sql
WARNING    Unable to lint models/analytics/editorial/sl_web_editorial_time_on_page.sql due to an internal error. Please report this as an issue with your query's contents and stacktrace below!
  File "/Users/gregclunies/Repos/sqlfluff/src/sqlfluff/core/linter.py", line 1271, in lint_path
    target_file.read(), fname=fname, fix=fix, config=config
  File "/Users/gregclunies/Repos/sqlfluff/src/sqlfluff/core/linter.py", line 1060, in lint_string
    parsed = self.parse_string(in_str=in_str, fname=fname, config=config)
  File "/Users/gregclunies/Repos/sqlfluff/src/sqlfluff/core/linter.py", line 810, in parse_string
  File "/Users/gregclunies/Repos/sqlfluff/src/sqlfluff/core/parser/lexer.py", line 349, in lex
    return self.enrich_segments(segment_buff, raw), violations
  File "/Users/gregclunies/Repos/sqlfluff/src/sqlfluff/core/parser/lexer.py", line 379, in enrich_segments
    templated_slice
  File "/Users/gregclunies/Repos/sqlfluff/src/sqlfluff/core/templaters/base.py", line 257, in templated_slice_to_source_slice
    template_slice.stop, inclusive=False
  File "/Users/gregclunies/Repos/sqlfluff/src/sqlfluff/core/templaters/base.py", line 184, in _find_slice_indices_of_templated_pos
    raise ValueError("Position Not Found")
ValueError: Position Not Found

```

I think this is because the modified templated SQL resulting from `templated_sql = templated_sql + "\n"` no longer "matches up with" `sliced_file` that is returned from...
```python 
raw_sliced, sliced_file, templated_sql = self.slice_file(
            node.raw_sql, compiled_sql, config=config
        )
```
...which is run just above it and is used by the `TemplatedFile()` class below it.

So seems to me the options are:
1. Solve it as I have in `python.py`. Not ideal in terms of "structure" as you called out, but gets the job done and seems to be the DRY solution.
2. Have an analogous `slice_file()` class method in `dbt.py` that overrides the definition in `python.py`. It would essentially be a carbon copy of what I did in `python.py`, just moved to `dbt.py`. I'll call this the WET solution.💦 
3. A more elegant solution I have not thought of! 😆

GClunies: @NiallRees @dmateusp 

I have updated (and simplified) the code in this PR so that:
1. All of the logic for handling trailing newlines is in `templaters/dbt.py`.
2. Rather than blindly adding a _single_ end of file newline, we add the same number as are in the raw SQL.

I have also updated the PR description above to reflect these changes. TL;DR - users do not have to include exclude_rules = L009 when using the dbt-templater anymore.

I'll be working on tests now that I have a solution I am satisfied with.

GClunies:
@NiallRees @dmateusp 

Can I get a review of my changes. I have:
- Updated the logic so trailing newline(s) are handled in `templaters/dbt.py`.
- Maintains the same number of trailing newlines as in the _source_ SQL.
- Users do not have to include `exclude_rules = L009` in their `.sqlfluff` when using the `templater=dbt` anymore.
- Adds tests to make sure all of the above works as intended.

test/core/templaters/dbt_test.py
```diff
@@ -64,7 +64,7 @@ def test__templater_dbt_templating_result(
         config=FluffConfig(configs=DBT_FLUFF_CONFIG),
     )
     # the dbt compiler gets rid of new lines
```
dmateusp, line 66: I think you can remove my comment here now :)

test/core/templaters/dbt_test.py
```diff
@@ -120,14 +120,82 @@ def test__templater_dbt_templating_test_lex(in_dbt_project_dir, dbt_templater):
 )
 @pytest.mark.dbt
 def test__templated_sections_do_not_raise_lint_error(in_dbt_project_dir, fname):  # noqa
-    """Test that the dbt test has only a new line lint error."""
+    """Test that templated sections in dbt model does not raise a linting error."""
```
dmateusp, line 123:
```suggestion
    """Test that templated sections in dbt model do not raise a linting error."""
```

test/core/templaters/dbt_test.py
```diff
@@ -120,14 +120,82 @@ def test__templater_dbt_templating_test_lex(in_dbt_project_dir, dbt_templater):
...
     lnt = lntr.lint_path(path="models/my_new_project/" + fname)
     violations = lnt.check_tuples()
     print(violations)
     assert len(violations) == 0
 
 
+@pytest.mark.parametrize(
+    "fname",
+    [
+        "single_trailing_newline.sql",
```
dmateusp, line 134: I think this fixture should be added to https://github.com/sqlfluff/sqlfluff/pull/868/files#diff-5d3be0ba02e85767f3906e01bdb3611ce1c9defc129ba33dde3c2ecd6e6727e0R118 rather than the code being copied

test/core/templaters/dbt_test.py
```diff
@@ -120,14 +120,82 @@ def test__templater_dbt_templating_test_lex(in_dbt_project_dir, dbt_templater):
...
+    )
+    tokens, lex_vs = lexer.lex(templated_file)
+    assert templated_file.source_str == "select a\nfrom table_a\n"
+    assert templated_file.templated_str == "select a\nfrom table_a\n"
+
+
+@pytest.mark.parametrize(
+    "fname",
+    [
+        "multiple_trailing_newline.sql",
```
dmateusp, line 168: I think this fixture should be added to https://github.com/sqlfluff/sqlfluff/pull/868/files#diff-5d3be0ba02e85767f3906e01bdb3611ce1c9defc129ba33dde3c2ecd6e6727e0R118 rather than the code being copied

test/core/templaters/dbt_test.py
```diff
@@ -120,14 +120,82 @@ def test__templater_dbt_templating_test_lex(in_dbt_project_dir, dbt_templater):
...
+    """Test that dbt model with a single trailing newline at end of file does not raise a linting error."""
+    lntr = Linter(config=FluffConfig(configs=DBT_FLUFF_CONFIG))
+    lnt = lntr.lint_path(path="models/my_new_project/" + fname)
+    violations = lnt.check_tuples()
+    print(violations)
+    assert len(violations) == 0
+
+
+@pytest.mark.dbt
+def test__dbt_templater_maintains_single_trailing_newline_at_file_end(
```
dmateusp, line 150: This is the same code as https://github.com/sqlfluff/sqlfluff/pull/868/files#diff-5d3be0ba02e85767f3906e01bdb3611ce1c9defc129ba33dde3c2ecd6e6727e0R99 so I would suggest using fixtures too

test/core/templaters/dbt_test.py
```diff
@@ -120,14 +120,82 @@ def test__templater_dbt_templating_test_lex(in_dbt_project_dir, dbt_templater):
...
+    """Test that dbt model with a multiple trailing newlines at end of file does not raise a linting error."""
+    lntr = Linter(config=FluffConfig(configs=DBT_FLUFF_CONFIG))
+    lnt = lntr.lint_path(path="models/my_new_project/" + fname)
+    violations = lnt.check_tuples()
+    print(violations)
+    assert len(violations) == 0
+
+
+@pytest.mark.dbt
+def test__dbt_templater_maintains_multiple_trailing_newlines_at_file_end(
```
dmateusp, line 184: This is the same code as https://github.com/sqlfluff/sqlfluff/pull/868/files#diff-5d3be0ba02e85767f3906e01bdb3611ce1c9defc129ba33dde3c2ecd6e6727e0R99 so I would suggest using fixtures too

dmateusp (changes requested):
Very clean @GClunies! I just had a few comments to keep the test code tidy and some typos

test/core/templaters/dbt_test.py
GClunies: I've simplified the tests to minimize repeated code and use fixtures.

GClunies:
@dmateusp Thanks for the review! Changes made. I _think_ this should do it! 🤞🏽

dmateusp (approved):
Amazing @GClunies 🎉  thanks for the contribution

NiallRees (approved):

GClunies:
Thanks for your patience @dmateusp and @NiallRees!

PR #870, opened 2021-03-16
Parse a and not b expression

### LINKED ISSUE #827: ANSI dialect can't parse expression with "AND NOT", e.g. "SELECT  a AND NOT i.b FROM i"

## Expected Behaviour
this is the part of the query it's saying is missing an AS alias
```
  i.item_upc IS NOT NULL
  AND i.item_upc NOT IN ('000000000000', '1', '')
  AND NOT i.is_excluded_upc
  AND COALESCE(i.linked_error NOT ILIKE '%there is already an item from%', TRUE) AS has_linkable_upc,
```
## Observed Behaviour
and its inserting the AS after the `AND NOT` which is not correct.
```
  i.item_upc IS NOT NULL
  AND i.item_upc NOT IN ('000000000000', '1', '')
  AND NOT AS i.is_excluded_upc
  AND COALESCE(i.linked_error NOT ILIKE '%there is already an item from%', TRUE) AS has_linkable_upc,
```
## Steps to Reproduce
sqlfluff lint and sqlfluff fix

## Version
sqlfluff, version 0.4.1
Python 2.7.16

## Configuration
```
[sqlfluff]
verbose = 0
nocolor = False
dialect = snowflake
templater = jinja
rules = L010,L012,L023,L030
exclude_rules = None
recurse = 0
output_line_length = 80
ignore = parsing,templating

[sqlfluff:indentation]
indented_joins = False

[sqlfluff:templater:jinja]
apply_dbt_builtins = True
load_macros_from_path=dbt/macros/parsing

[sqlfluff:rules]
tab_space_size = 2
max_line_length = 80
indent_unit = space
comma_style = trailing
allow_scalar = True
only_aliases = True

#inconsistent capitalization of keywords
[sqlfluff:rules:L010]
capitalisation_policy = upper

#unquoted identifiers
[sqlfluff:rules:L014]

#uppercase functions
[sqlfluff:rules:L030]
capitalisation_policy = upper

```

NiallRees:
Resolves #827 and #874

This logic change is causing a maximum recursion depth exceeded error in test cases `expression_recursion.sql` and `expression_recursion_2.sql`. I have increased the max depth to accommodate the increase.

Curious for your thoughts @barrywhart @dmateusp @alanmcruickshank


src/sqlfluff/__init__.py
```diff
@@ -20,6 +20,9 @@
 import pkg_resources
 import configparser
 
+# Increase the maximum recursion depth from default of 10^4 to 10^5, see https://github.com/sqlfluff/sqlfluff/pull/870
+sys.setrecursionlimit(10 ** 5)
```
barrywhart, line 24: TBH, I'm surprised that increasing it to such a high number:
* Is necessary
* Works

Usually, if I write code that recurses this deeply, it's because of accidental _infinite_ recursion. Do we have a sense for why the parser recurses so deeply but eventually does the right thing and stops?

barrywhart: TBH, I don't know what units this number is in. I had naively assumed it is the number of function calls. Is it possible it's something else, such as bytes, e.g. the amount of storage taken up the stack, local variables, ??

barrywhart (approved):
It works, so :shipit: . Confused and a _bit_ concerned why it's necessary, but I'm good with merging.

alanmcruickshank:
I think there should be a better way of fixing this issue. I might try a few things...

alanmcruickshank:
Take a look at my suggestion @NiallRees . I think adjusting the recursion limit is a hack around the expression parsing logic, the changes I've just made fix the ambiguity of `NOT` in the expression grammars so that it deals properly with all of the cases.

src/sqlfluff/core/dialects/dialect_ansi.py
```diff
@@ -1190,18 +1190,21 @@ class CaseExpressionSegment(BaseSegment):
...
                     OneOf(
-                        Ref("BinaryOperatorGrammar"),
                         Sequence(
                             Ref.keyword("NOT", optional=True),
                             Ref("LikeGrammar"),
-                        )
+                        ),
+                        Sequence(
+                            Ref("BinaryOperatorGrammar"),
+                            Ref.keyword("NOT", optional=True),
```
barrywhart, line 1206: Ooh, clever!

barrywhart (approved):
This looks great. Thanks for the great collaboration on this one, @NiallRees and @alanmcruickshank! This was a subtle issue with glaring symptoms, very happy that this 🐉 has been taken care of.

src/sqlfluff/core/dialects/dialect_ansi.py
```diff
@@ -1191,18 +1191,21 @@ class CaseExpressionSegment(BaseSegment):
                     "NOT",
                     "PRIOR",  # used in CONNECT BY clauses (EXASOL, Snowflake, Postgres...)
                 ),
-                Ref("Expression_A_Grammar"),
+                Ref("Expression_C_Grammar"),
```
NiallRees, line 1194: Hopefully this will reduce the incidence of those hard to reproduce recursion depth errors

NiallRees:
Excellent @alanmcruickshank thank you

PR #872, opened 2021-03-17
Document noqa

alanmcruickshank:
As discussed on slack. The `noqa` feature is poorly documented. This fixes that.


pwildenhain (approved):
🙈🎉

barrywhart (approved):
:shipit: 

Thanks for doing this! It feels like SQLFluff is growing up -- small feature but feels big in terms of maturity. 💪🏽

Issue #511, opened 2020-10-28
Rules page in documentation is broken.


a-litvinov:
I've got an impression that this page was supposed to list all implemented lint rules, but it is empty: https://docs.sqlfluff.com/en/latest/rules.html#ruleref

If my impression is incorrect, is there a page where I can see all rules description besides looking directly into the code?

alanmcruickshank:
@a-litvinov - thanks for raising. Your assumption is correct and you have found a bug with that docs page. I'm looking at why it's happening now...

alanmcruickshank:
@a-litvinov  - this is fixed! Check out the same link now.

a-litvinov:
Thanks for the quick reaction! :+1:

PR #875, opened 2021-03-19
Fix linting

NiallRees:
For some reason `pydocstyle==6.0.0` released on 18th March is being installed in circle ci but not locally when installing from requirements and requirements dev. The [6.0.0 release adds](pydocstyle==6.0.0):

> Methods, Functions and Nested functions that have a docstring now throw D418 (#511).

I've ignored this rule in tox.ini

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,14 +85,14 @@ jobs:
   linting:
     <<: *common
     docker:
-      - image: circleci/python:3.6
+      - image: circleci/python:3.9
         environment:
           TOXENV: linting
           UPLOAD_COVERAGE: 0
   doclinting:
     <<: *common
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -85,14 +85,14 @@ jobs:
   linting:
     <<: *common
     docker:
-      - image: circleci/python:3.6
+      - image: circleci/python:3.9
```
NiallRees, line 88: This was the first thing I tried, but don't see why we shouldn't leave it on 3.9

barrywhart (approved):
:shipit:

barrywhart:
I'm going to go ahead and merge this since other PR builds need this in order to work.

PR #877, opened 2021-03-20
Explicitly nest expressions when bracketed.

alanmcruickshank:
This is related to #870 .

In expression parsing, I noticed two parts not functioning correctly.

1. In `JoinOnConditionSegment` an additional `expression` layer was being added when it should just be _optionally bracketed_.
2. Nested expressions were being parsed as one big expression, because we we referring to an expression _grammar_ when brackets were present rather than an expression _segment_.

This addresses both.

### DIFF
--- a/src/sqlfluff/core/dialects/dialect_ansi.py
+++ b/src/sqlfluff/core/dialects/dialect_ansi.py
@@ -1072,10 +1072,9 @@ class JoinOnConditionSegment(BaseSegment):
     match_grammar = Sequence(
         "ON",
         Indent,
-        OneOf(
-            # This is a great case for "optionally bracketed"
-            Bracketed(Ref("ExpressionSegment", ephemeral_name="JoinCondition")),
+        OptionallyBracketed(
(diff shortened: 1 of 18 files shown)


src/sqlfluff/core/dialects/dialect_ansi.py
```diff
@@ -1073,8 +1073,9 @@ class JoinOnConditionSegment(BaseSegment):
         "ON",
         Indent,
         OneOf(
-            Ref("ExpressionSegment"),
+            # This is a great case for "optionally bracketed"
```
NiallRees, line 1076: See https://github.com/sqlfluff/sqlfluff/blob/master/src/sqlfluff/core/parser/grammar/anyof.py#L226, @barrywhart added it this week!

alanmcruickshank: Oh snap! 🎉

NiallRees (approved):

Issue #878, opened 2021-03-20
Snowflake schema object double-dot notation parsing error


DavidJohnQuinlan:
## Expected Behaviour

In Snowflake a fully-qualified schema object (table, view, file format etc.) has the form:

`<database_name>.<schema_name>.<object_name>`

However, this can be tedious to write, therefore, it is possible to exclude the `schema_name` and use the following syntax in a query:

`<database_name>..<object_name>`

The two dots indicate that the schema name is not specified. The PUBLIC default schema is always referenced.

Currently SQLFluff is unable to deal with the double-dot syntax and returns a parsing error for the following example query:

```
SELECT * FROM MY_DB..TABLE
```

## Observed Behaviour

When running the `sqlfluff lint query.sql` command on the `query.sql` script containing the above example query the following error is returned: 

`Found unparsable section: 'FROM MY_DB..TABLE'`

## Version

`sqlfluff, version 0.4.1`
`Python 3.7.5`

PR #879, opened 2021-03-20
Issue 878: Adding Snowflake schema object double-dot notation

DavidJohnQuinlan:
Resolves [issue 878](https://github.com/sqlfluff/sqlfluff/issues/878) by adding the option of a sequence of two dots to the `delimiter` argument in the `Delimited` function called in the `ObjectReferenceSegment` class. Also created an additional SQL test query containing the double-dot notation.


alanmcruickshank (approved):
I thought someone had already done this, but evidently not!

I wonder if this also works without the `Sequence(Ref("DotSegment"))`, in which case this could be simplified.

Happy to handle that elsewhere however.

DavidJohnQuinlan:
Thanks Alan, I shall investigate further

DavidJohnQuinlan:
Hey Alan, you were right it works without `Sequence(Ref("DotSegment"))`. This change has been added to the PR.

dmateusp:
Somehow the CI never triggered here, I don't think I have the rights to create a new run on CircleCI, maybe an empty commit would fix it @DavidJohnQuinlan ?

DavidJohnQuinlan:
Hey @dmateusp, the empty commit does not appear to have changed things. Is there something that I am missing on my side? These tests should trigger on push right?

alanmcruickshank:
No idea why CI doesn't seem to be triggered here, but I can always _admin-merge_ it. 🔨 

Thanks for putting this together!

PR #882, opened 2021-03-22
FROM clause structure

### LINKED ISSUE #871: CROSS JOIN confuses parser?
## Expected Behaviour
SQLFluff should recognize the alias `foo`

## Observed Behaviour

The alias `foo` isn't recognized. When I remove the `CROSS JOIN`, SQLFluff no longer emits L026 for `foo`. 

```
$ sqlfluff lint test.sql
== [test.sql] FAIL
L:   7 | P:   5 | L026 | Reference 'foo.name' refers to table/view 'foo' not
                       | found in the FROM clause or found in parent subquery.
L:   8 | P:   5 | L026 | Reference 'foo.value' refers to table/view 'foo' not
                       | found in the FROM clause or found in parent subquery.
L:  12 | P:   1 | L003 | Indent expected and not found compared to line #10
L:  12 | P:  16 | L031 | Avoid using aliases in join condition
L:  12 | P:  20 | L032 | Found USING statement. Expected only ON statements.
```

## Steps to Reproduce
`sqlfluff lint test.sql` with default config

Contents of `test.sql` --
```sql
WITH constants AS (
    SELECT 8760 AS hours_per_year
)

SELECT
    table1.name,
    foo.name,
    foo.value * constants.hours_per_year AS some_value
FROM table1
CROSS JOIN
    constants
JOIN table2 AS foo USING (id)
;
```

## Version
```
$ sqlfluff --version
sqlfluff, version 0.4.1
$ python --version
Python 3.9.2
```

## Configuration
I don't have a config file; I'm using the default config for 0.4.1.

alanmcruickshank:
Resolves #871 .

This reworks the order of precedence within `FROM` and `JOIN` clauses. I think the existing logic was based on having been built upon and extended a few times which meant we ended up with a structure which didn't quite work.

The new structure will allow this kind of structure:

```
SELECT *
FROM a JOIN b, c JOIN d
```


src/sqlfluff/core/dialects/dialect_ansi.py
```diff
@@ -1100,10 +1105,20 @@ class FromClauseSegment(BaseSegment):
...
+            Sequence(
+                Indent,
+                OneOf(
+                    # check first for MLTableExpression, because of possible FunctionSegment in MainTableExpression
+                    Ref("MLTableExpressionSegment"),
+                    Ref("TableExpressionSegment"),
+                ),
+                Dedent.when(indented_joins=False),
+                AnyNumberOf(
+                    Ref("JoinClauseSegment"),
```
NiallRees, line 1117: I don't think that the join clause belongs here 😬. I actually think it belongs in the `TableExpressionSegment` per my previous PR: https://github.com/sqlfluff/sqlfluff/pull/791

test/fixtures/parser/snowflake/snowflake_lateral_flatten_after_join.yml
```diff
@@ -1,96 +1,96 @@
...
+          alias_expression:
+            keyword: as
+            identifier: most_recent
+      from_clause:
+      - keyword: from
+      - table_expression:
+          main_table_expression:
+            table_reference:
+              identifier: a
+      - join_clause:
```
NiallRees, line 61: See this example. In my current understanding, a from clause is a comma separated list of table expressions. Each table expression can contain one or many tables, where there are joins. Crucially the join statements should be contained by a table expression.

test/fixtures/parser/ansi/select_simple_i.yml
```diff
@@ -1,87 +1,87 @@
...
-          join_clause:
-            keyword: JOIN
-            table_expression:
-              main_table_expression:
-                table_reference:
-                  identifier: aliased_1
-              alias_expression:
-                keyword: as
-                identifier: foo
-              join_clause:
```
NiallRees, line 76: This is definitely wrong in the current master though.

NiallRees (changes requested):
My above comments, take a look at the from clause grammar in cockroach: https://www.cockroachlabs.com/docs/v20.2/sql-grammar.html#from_clause

test/fixtures/parser/ansi/select_cross_join.yml
```diff
@@ -0,0 +1,72 @@
...
+            alias_expression:
+              keyword: AS
+              identifier: some_value
+        from_clause:
+        - keyword: FROM
+        - table_expression:
+            main_table_expression:
+              table_reference:
+                identifier: table1
+        - join_clause:
```
barrywhart, line 53: The parse tree looks good in the sense that the two joins are at the same level. I'll defer to you and Niall about the question he raised.

src/sqlfluff/core/dialects/dialect_ansi.py
alanmcruickshank: I hear you about making it part of table expression, _however_ we have a problem there where it becomes very recursive. Rather than replying here I'll add some more detail below.

alanmcruickshank:
> I don't think that the join clause belongs here 😬. I actually think it belongs in the TableExpressionSegment per my previous PR: #791

Let's take the example query:

```sql
SELECT *
FROM a JOIN b JOIN c JOIN d, e JOIN f JOIN g JOIN h
```

On current master (d4da92a8a2355d17e9de1c074aeaf28299818248) this parses as:

```bash
$ sqlfluff parse test.sql -f yaml -c
```

```yaml
- filepath: test.sql
  segments:
    file:
      statement:
        select_statement:
          select_clause:
            keyword: SELECT
            select_target_element:
              wildcard_expression:
                wildcard_identifier:
                  star: '*'
          from_clause:
          - keyword: FROM
          - table_expression:
              main_table_expression:
                table_reference:
                  identifier: a
              join_clause:
                keyword: JOIN
                table_expression:
                  main_table_expression:
                    table_reference:
                      identifier: b
                  join_clause:
                    keyword: JOIN
                    table_expression:
                      main_table_expression:
                        table_reference:
                          identifier: c
                      join_clause:
                        keyword: JOIN
                        table_expression:
                          main_table_expression:
                            table_reference:
                              identifier: d
          - comma: ','
          - table_expression:
              main_table_expression:
                table_reference:
                  identifier: e
              join_clause:
                keyword: JOIN
                table_expression:
                  main_table_expression:
                    table_reference:
                      identifier: f
                  join_clause:
                    keyword: JOIN
                    table_expression:
                      main_table_expression:
                        table_reference:
                          identifier: g
                      join_clause:
                        keyword: JOIN
                        table_expression:
                          main_table_expression:
                            table_reference:
                              identifier: h
```

The implementation of the existing join logic as part of `TableExpression` means that the join statements nest. If join order didn't matter, this wouldn't be a massive deal, but the problem here is that we're actually nesting in the wrong order to be reflective of how SQL joins actually work: i.e. we're parsing it as `a JOIN (b JOIN (c JOIN d))` when it's actually `(((a JOIN b) JOIN c) JOIN d)`.

The reason it parses this way is that although there is an `AnyNumberOf` around the `JoinClauseSegment`, that first `JoinClauseSegment` has as it's first element a `TableExpressionSegment` which contains a `AnyNumberOf` with a `JoinClauseSegment`, and so on. In essence the problem is that `JoinClauseSegment` and `TableExpressionSegment` currently contain eachother. We never get to the `AnyNumberOf` because `TableExpressionSegment` ends up being greedy on join clauses. The coachroach DB SQL docs are not clear on how they resolve this recursion.

Similar to the bugs with the Expression parsing code, I'm trying to both address the recursion issue and also create a parse structure which is more faithful to the meaning as per #871.

Assuming that the current parse structure is not the intended behavior (which is a separate debate we could have), that leaves us with two options:
1. Either `JoinClauseSegment` shouldn't contain `TableExpressionSegment`, but some subclass of it (like `MainTableExpressionSegment`). This is initially attractive, however whatever we keep here needs to still have `AliasExpressionSegment` and `PostTableExpressionGrammar` - which leads us to something very similar to the existing `TableExpressionSegment` (albeit sans Join conditions).
2. Or `TableExpressionSegment` shouldn't contain `JoinClauseSegment`, and we just migrate the join clause up to part of the `FROM` clause (which is the route I took).

Given the same sample query, this now parses as:

```yaml
- filepath: test.sql
  segments:
    file:
      statement:
        select_statement:
          select_clause:
            keyword: SELECT
            select_target_element:
              wildcard_expression:
                wildcard_identifier:
                  star: '*'
          from_clause:
          - keyword: FROM
          - table_expression:
              main_table_expression:
                table_reference:
                  identifier: a
          - join_clause:
              keyword: JOIN
              table_expression:
                main_table_expression:
                  table_reference:
                    identifier: b
          - join_clause:
              keyword: JOIN
              table_expression:
                main_table_expression:
                  table_reference:
                    identifier: c
          - join_clause:
              keyword: JOIN
              table_expression:
                main_table_expression:
                  table_reference:
                    identifier: d
          - comma: ','
          - table_expression:
              main_table_expression:
                table_reference:
                  identifier: e
          - join_clause:
              keyword: JOIN
              table_expression:
                main_table_expression:
                  table_reference:
                    identifier: f
          - join_clause:
              keyword: JOIN
              table_expression:
                main_table_expression:
                  table_reference:
                    identifier: g
          - join_clause:
              keyword: JOIN
              table_expression:
                main_table_expression:
                  table_reference:
                    identifier: h
```

Which I think is a better representation of the structure of the query and the logic it represents. Granted, it does mean that what we refer to as `TableExpressionSegment` is not the same as the coachroach DB docs, but if that's the snag then I'm happy to rename it.

@NiallRees, do you have a better idea on how to achieve the desired structure?

test/fixtures/parser/snowflake/snowflake_lateral_flatten_after_join.yml
alanmcruickshank: I don't think they *need* to be contained by something, but I'm definitely open to that. See my longer note below, where I think I can show the issue with the current logic. We could *also* make a container for the joins if you think that would make more sense?

alanmcruickshank:
@NiallRees in re-reading your previous comments, are you implying that the actual parse structure should be:

```yaml
- filepath: test.sql
  segments:
    file:
      statement:
        select_statement:
          select_clause:
            keyword: SELECT
            select_target_element:
              wildcard_expression:
                wildcard_identifier:
                  star: '*'
          from_clause:
          - keyword: FROM
          - super_table_expression:
            - table_expression:
                main_table_expression:
                  table_reference:
                    identifier: a
            - join_clause:
                keyword: JOIN
                table_expression:
                  main_table_expression:
                    table_reference:
                      identifier: b
            - join_clause:
                keyword: JOIN
                table_expression:
                  main_table_expression:
                    table_reference:
                      identifier: c
            - join_clause:
                keyword: JOIN
                table_expression:
                  main_table_expression:
                    table_reference:
                      identifier: d
          - comma: ','
          - super_table_expression:
            - table_expression:
                main_table_expression:
                  table_reference:
                    identifier: e
            - join_clause:
                keyword: JOIN
                table_expression:
                  main_table_expression:
                    table_reference:
                      identifier: f
            - join_clause:
                keyword: JOIN
                table_expression:
                  main_table_expression:
                    table_reference:
                      identifier: g
            - join_clause:
                keyword: JOIN
                table_expression:
                  main_table_expression:
                    table_reference:
                      identifier: h
```

I could get on board with this structure, and obviously renaming `super_table_expression` to something more intelligent if that's the sticking point.

NiallRees:
> @NiallRees in re-reading your previous comments, are you implying that the actual parse structure should be:
> 
> ```yaml
> - filepath: test.sql
>   segments:
>     file:
>       statement:
>         select_statement:
>           select_clause:
>             keyword: SELECT
>             select_target_element:
>               wildcard_expression:
>                 wildcard_identifier:
>                   star: '*'
>           from_clause:
>           - keyword: FROM
>           - super_table_expression:
>             - table_expression:
>                 main_table_expression:
>                   table_reference:
>                     identifier: a
>             - join_clause:
>                 keyword: JOIN
>                 table_expression:
>                   main_table_expression:
>                     table_reference:
>                       identifier: b
>             - join_clause:
>                 keyword: JOIN
>                 table_expression:
>                   main_table_expression:
>                     table_reference:
>                       identifier: c
>             - join_clause:
>                 keyword: JOIN
>                 table_expression:
>                   main_table_expression:
>                     table_reference:
>                       identifier: d
>           - comma: ','
>           - super_table_expression:
>             - table_expression:
>                 main_table_expression:
>                   table_reference:
>                     identifier: e
>             - join_clause:
>                 keyword: JOIN
>                 table_expression:
>                   main_table_expression:
>                     table_reference:
>                       identifier: f
>             - join_clause:
>                 keyword: JOIN
>                 table_expression:
>                   main_table_expression:
>                     table_reference:
>                       identifier: g
>             - join_clause:
>                 keyword: JOIN
>                 table_expression:
>                   main_table_expression:
>                     table_reference:
>                       identifier: h
> ```
> 
> I could get on board with this structure, and obviously renaming `super_table_expression` to something more intelligent if that's the sticking point.

Yep this is the structure I mean. If that makes sense to you too then I think we're good? Otherwise I'll have a further think about your other comments! 😄

alanmcruickshank:
I can work with that - I'll put something together 👍 .

alanmcruickshank:
@NiallRees , check out `test\fixtures\parser\ansi\select_many_join.sql` where I've tried to cover as many cases as possible:

```sql
SELECT *
FROM a as foo JOIN b JOIN c as foobar JOIN d, e as bar JOIN f JOIN g('blah') as tbl_func JOIN h, baz as buzz
```

Now parses as:

```yaml
file:
  statement:
    select_statement:
      select_clause:
        keyword: SELECT
        select_target_element:
          wildcard_expression:
            wildcard_identifier:
              star: '*'
      from_clause:
      - keyword: FROM
      - joined_table_expression:
        - table_expression:
            main_table_expression:
              table_reference:
                identifier: a
            alias_expression:
              keyword: as
              identifier: foo
        - join_clause:
            keyword: JOIN
            table_expression:
              main_table_expression:
                table_reference:
                  identifier: b
        - join_clause:
            keyword: JOIN
            table_expression:
              main_table_expression:
                table_reference:
                  identifier: c
              alias_expression:
                keyword: as
                identifier: foobar
        - join_clause:
            keyword: JOIN
            table_expression:
              main_table_expression:
                table_reference:
                  identifier: d
      - comma: ','
      - joined_table_expression:
        - table_expression:
            main_table_expression:
              table_reference:
                identifier: e
            alias_expression:
              keyword: as
              identifier: bar
        - join_clause:
            keyword: JOIN
            table_expression:
              main_table_expression:
                table_reference:
                  identifier: f
        - join_clause:
            keyword: JOIN
            table_expression:
              main_table_expression:
                function:
                  function_name: g
                  start_bracket: (
                  expression:
                    literal: "'blah'"
                  end_bracket: )
              alias_expression:
                keyword: as
                identifier: tbl_func
        - join_clause:
            keyword: JOIN
            table_expression:
              main_table_expression:
                table_reference:
                  identifier: h
      - comma: ','
      - table_expression:
          main_table_expression:
            table_reference:
              identifier: baz
          alias_expression:
            keyword: as
            identifier: buzz
```

Things to note:
- Alias expressions are part of `table_expression` but not `main_table_expression`. This means they're consistent, regardless of whether we're dealing with a tableau reference or a table function.
- Any joined tables are grouped into a `joined_table_expression`, which is then the thing which is delimited.
- Any tables which aren't part of a join can still survive on their own (e.g. `baz` in the example), in which case we get a lone `table_expression` which is not part of a `joined_table_expression`.
- The only potential wierdness is the treatment of snowflake `PIVOT` which is in as a `JoinLikeClauseGrammar` - and therefore triggers a `joined_table_expression`, even though there's no join. This doesn't feel quite right - but in the grand scheme of things I'm tempted to say it's fine and deal with it later if we need to refine that.

What do you reckon to this @NiallRees ?

barrywhart:
@alanmcruickshank: Quick question about `PIVOT()`, which you mentioned in Snowflake. Is this syntactically similar to `UNNEST()` in BigQuery or `LATERAL FLATTEN` in Snowflake?

These appear in a few existing test cases in `test/fixtures/parser/bigquery` and `test/fixtures/parser/snowflake`, respectively. I don't have a concern here, just trying to understand.

alanmcruickshank:
> @alanmcruickshank: Quick question about `PIVOT()`, which you mentioned in Snowflake. Is this syntactically similar to `UNNEST()` in BigQuery or `LATERAL FLATTEN` in Snowflake?

- `LATERAL FLATTEN` is different because there's always a comma between it and the preceding table reference. I'm pretty happy with how this parses (check out `snowflake_lateral_flatten_after_join.sql`).
- I've only ever seen `UNNEST` on it's own and not with other joins or tables, @barrywhart - do you have an example query where it's not just on it's own?

barrywhart:
@alanmcruickshank: Here's an example where `UNNEST()` is used with a table. I found it in the [BigQuery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays#zipping_arrays).

```
WITH combinations AS (
  SELECT
    ['a', 'b'] AS letters,
    [1, 2, 3] AS numbers
)
SELECT ARRAY_AGG(
  STRUCT(letter, numbers[OFFSET(letters_offset)] AS number)
) AS pairs
FROM combinations, UNNEST(letters) AS letter WITH OFFSET AS letters_offset;
```

barrywhart:
Here's another partial example of `UNNEST()` with a table (just the `FROM` clause):
```
FROM `project.dataset.web_sessions` ,unnest(customDimensions) cd, unnest(hits) hits
where date(session_start_at) > "2020-09-25"
group by fullVisitorId, cd.index, cd.value
```

alanmcruickshank:
@barrywhart - given the example of `UNNEST()` it looks like it will parse separately because there's a comma.

`PIVOT()` is different in that there is no comma: https://docs.snowflake.com/en/sql-reference/constructs/pivot.html

The docs aren't clear exactly how pivot performs when there are multiple joins, but that seems like a rare use case.

alanmcruickshank:
@NiallRees , based on our other convo I've merged your PR and then done some renaming:

- `table_expression` > `from_expression_element`
- `main_table_expression` > `table_expression`
- `select_target_element` > `select_clause_element`

Thoughts on this set of names? It feels more consistent to me. Open to other suggestions for `from_expression_element` and `select_clause_element`

NiallRees (approved):
Sounds good to me!

PR #885, opened 2021-03-23
Silence the Indent balance test failed warning

NiallRees:
This has been raised by several users now. As it isn't actionable by the end-user, it seems that it should be at the debug level.

https://sqlfluff.slack.com/archives/C01GX4L213J/p1614615362024600


alanmcruickshank (approved):
Makes sense

PR #886, opened 2021-03-23
From clause structure suggestion

NiallRees:


alanmcruickshank (approved):
I think we need to do some renaming (as we talked through on the phone), but I like this structure.
`main_table_expression` should be `table_expression`.
the current `table_expression` should be renamed. My starter for 10 would be `from_expression_element`, but I'm not totally sold on that yet.

Gut feel says we might want to rename `select_target_element` too so that it's consistent.

Issue #830, opened 2021-03-01
Not recognizing missing ";"


giselamorrone:
-- If this is a parsing or linting issue, please include a minimal SQL example which reproduces the issue, along with the `sqlfluff parse` output, `sqlfluff lint` output and `sqlfluff fix` output when relevant. Finally, please add any labels that are related to the issue before submitting so that it can be addressed easily! --

## Expected Behaviour
Get an error when there is a missing **;** 

**test.sql**
```
-- Obtaining max updated date
DROP TABLE if EXISTS max_date;
CREATE TEMP TABLE max_date AS
SELECT max(insert_date) AS max FROM data.final_table;

-- Obtaining rows modified since last update
CREATE TEMP TABLE stg_table AS
SELECT
  col1, col2
FROM staging.final_table
WHERE insert_date > (SELECT max FROM max_date )
ORDER BY blast_id, everlane_inserted_at
;

BEGIN TRANSACTION;
DELETE FROM data.final_table
USING stg_table stg
WHERE stg.col1 = final_table.col2

INSERT INTO data.final_table
    ( col1, col2)
select
  col1, col2
from stg_table
order by col1
;
end TRANSACTION;
```

## Observed Behaviour
The line with the missing ; is not identified as an incorrect line.

## Steps to Reproduce
Run: `sqlfluff lint test.sql`

## Version
`sqlfluff --version`
sqlfluff, version 0.4.0
`python --version`
Python 3.6.8

## Configuration
```
[sqlfluff]
dialect = postgres
```

## Results
`sqlfluff parse` 

```
[0](1, 1, 1)        |file:
[0](1, 1, 1)        |    comment:                                                  '-- Obtaining max updated date'
[29](1, 1, 30)      |    newline:                                                  '\n'
[30](1, 2, 1)       |    statement:
[30](1, 2, 1)       |        drop_statement:
[30](1, 2, 1)       |            keyword:                                          'DROP'
[34](1, 2, 5)       |            whitespace:                                       ' '
[35](1, 2, 6)       |            keyword:                                          'TABLE'
[40](1, 2, 11)      |            whitespace:                                       ' '
[41](1, 2, 12)      |            keyword:                                          'if'
[43](1, 2, 14)      |            whitespace:                                       ' '
[44](1, 2, 15)      |            keyword:                                          'EXISTS'
[50](1, 2, 21)      |            whitespace:                                       ' '
[51](1, 2, 22)      |            table_reference:
[...]

PR #887, opened 2021-03-24
Add BEGIN/END TRANSACTION.

alanmcruickshank:
this partially addresses #830 .


src/sqlfluff/core/dialects/dialect_ansi.py
```diff
@@ -1710,6 +1710,8 @@ class TransactionStatementSegment(BaseSegment):
             Ref.keyword("WORK", optional=True),
             Sequence("AND", Ref.keyword("NO", optional=True), "CHAIN", optional=True),
         ),
+        # BEGIN | END TRANSACTION
+        Sequence(OneOf("BEGIN", "END"), "TRANSACTION"),
```
dmateusp, line 1714: Should we make "TRANSACTION" optional? it is optional in Snowflake: https://docs.snowflake.com/en/sql-reference/sql/begin.html and in Postgres https://www.postgresql.org/docs/9.0/sql-begin.html

alanmcruickshank: good catch

Issue #831, opened 2021-03-01
Templated jinja is not ignored


mdimercurio:
When trying to use macros from a package, sqlfluff returns an error when these should be skipped because of the `ignore_templated_areas` parameter.

## Expected Behaviour

`dbt_utils` macros should be ignored in the lint command.

## Observed Behaviour

`sqlfluff lint` output

```
L:   - | P:   - |  TMP | Unrecoverable failure in Jinja templating: 'dbt_utils'
                       | is undefined. Have you configured your variables?
                       | https://docs.sqlfluff.com/en/latest/configuration.html
L:   4 | P:   4 |  TMP | Undefined jinja template variable: 'dbt_utils'
```
-- If this is a parsing or linting issue, please include a minimal SQL example which reproduces the issue, along with the `sqlfluff parse` output, `sqlfluff lint` output and `sqlfluff fix` output when relevant. Finally, please add any labels that are related to the issue before submitting so that it can be addressed easily! --

## Steps to Reproduce

Query:
```sql
select id
    , count(*)
from {{ source("snowplow", "events") }}
{{ dbt_utils.group_by(1) }}
```

Try to lint using:
> sqlfluff lint models/snowplow_events.sql
output:
```
== [models/snowplow_events.sql] FAIL
L:   - | P:   - |  TMP | Unrecoverable failure in Jinja templating: 'dbt_utils'
                       | is undefined. Have you configured your variables?
                       | https://docs.sqlfluff.com/en/latest/configuration.html
L:   4 | P:   4 |  TMP | Undefined jinja template variable: 'dbt_utils'
```

## Version
Include the output of `sqlfluff --version` along with your Python version

python 3.8.3

```
> sqlfluff --version
sqlfluff, version 0.4.1
```

## Configuration
```
[sqlfluff]
templater = jinja
ignore_templated_areas = True
```

NiallRees:
> When trying to use macros from a package, sqlfluff returns an error when these should be skipped because of the `ignore_templated_areas` parameter.
> 
> ## Expected Behaviour
> `dbt_utils` macros should be ignored in the lint command.
> 
> ## Observed Behaviour
> `sqlfluff lint` output
> 
> ```
> L:   - | P:   - |  TMP | Unrecoverable failure in Jinja templating: 'dbt_utils'
>                        | is undefined. Have you configured your variables?
>                        | https://docs.sqlfluff.com/en/latest/configuration.html
[...]

PR #892, opened 2021-03-25
Lint whitespace surrounding Jinja tags

### LINKED ISSUE #586: New rule. Jinja tag content should be surrounded by whitespace.
The basic machinery for doing this is now in place with #541, however some structural changes will probably still be required.

This would be a rule which lints jinja tags (`{{ }}`, `{# #}` & `{% %}`) in templated queries and ensures that there is at least a single whitespace between the delimiters of the tag and the content.

Bad: `{{ref()}}`, `{{my_variable}}`, `{{1+2}}`, `{%else%}`

Good: `{{ ref() }}`, `{{ my_variable }}`, `{{ 1+ 2 }}`, `{% else %}`.

Note: I don't think we should attempt to lint or parse the _content_ of the tag significantly beyond the edges as this would involve us effectively building a whole new sub-language within sqlfluff which feels like significant scope creep!

alanmcruickshank:
This resolves #586.

This should also pave the way for other linting errors which target templated sections explicitly be including `targets_templated` as an attribute of `BaseRule` which overrides the filter to ignore rules within templated sections.


src/sqlfluff/core/rules/std/L046.py
```diff
@@ -0,0 +1,64 @@
...
+        is_literal = getattr(segment.pos_marker, "is_literal", None)
+        source_str = getattr(templated_file, "source_str", None)
+
+        if source_slice and source_str and not is_literal:
+            # Does it actually look like a tag?
+            src_raw = source_str[source_slice]
+            if src_raw[0] != "{" or src_raw[-1] != "}":
+                return LintResult(memory=memory)
+
+            # Dedupe usign a memory of source indexes.
```
NiallRees, line 41:
```suggestion
            # Dedupe using a memory of source indexes.
```

src/sqlfluff/core/rules/std/L046.py
```diff
@@ -0,0 +1,64 @@
...
+            if not memory:
+                memory = set()
+            memory.add(src_idx)
+
+            # Get the inner section
+            inner_str = src_raw[2:-2]
+
+            # Do we have too little whitespace?
+            if len(inner_str) < 2 or inner_str[0] != " " or inner_str[-1] != " ":
+                print("foo")
```
NiallRees, line 57: Not sure this is meant to still be here 😛

src/sqlfluff/core/rules/std/L046.py
```diff
@@ -0,0 +1,64 @@
...
+                return LintResult(memory=memory, anchor=segment)
+
+            # Do we have too much whitespace?
+            if len(inner_str) > 4 and (inner_str[1] == " " or inner_str[-2] == " "):
```
NiallRees, line 61: I think we need a comment to explain the significance of 4 and [-2] here.

test/core/rules/test_cases/L046.yml
```diff
@@ -0,0 +1,10 @@
+rule: L046
+
+test_simple:
+  pass_str: SELECT 1 from {{ ref('foo') }}
+
+test_fail_1:
+  fail_str: SELECT 1 from {{ref('foo')}}
```
NiallRees, line 7: How fix-able do you think this rule is?

src/sqlfluff/core/linter.py
```diff
@@ -1014,7 +1018,12 @@ def remove_templated_errors(
         # Filter out any linting errors in templated sections if relevant.
         linting_errors = list(
             filter(
-                lambda e: getattr(e.segment.pos_marker, "is_literal", True),
+                lambda e: (
+                    # Is it in a literal section?
+                    getattr(e.segment.pos_marker, "is_literal", True)
+                    # Is it a rule that is designed to work on templated sections?
+                    or e.rule.targets_templated
```
NiallRees, line 1025: 👌

src/sqlfluff/core/rules/std/L046.py
alanmcruickshank: I've refactored this in the newer version.

test/core/rules/test_cases/L046.yml
alanmcruickshank: In theory easy, but I don't think we have a good mechanism for fixing the source file rather than the templated file. Could be a useful extension in future, but I'm not going to tackle it right now.

alanmcruickshank:
This PR also now fixes some of the line labelling issues from #831, which makes it much easier to debug.

Rather than some elements having EnrichedPostionMarkers and some not, this should mean we now have most of them or none of them. Particularly for indentation issues or code with lots of templating this makes bugfixing much easier but does change some of the test cases and the positions that they refer to.

PR #896, opened 2021-03-26
Move rule test cases to test/fixtures

alanmcruickshank:
I keep looking for rule test cases in the wrong place 🤦 .

My logic is that the rule test cases are in `yaml` and so don't need to be part of a python module. This means they can live in `test/fixtures` rather than being in a submodule of `test/rules`. This is where I would probably look for them, but also means we can separate the concerns of _how_ we test from _what_ we test.

The PR moves the rule test cases accordingly. I'd love a review from the other @sqlfluff/sqlfluff-maintainers to make sure this also makes sense to other people.


barrywhart (approved):
Overall, this change makes sense to me.

I notice that some of the files being moved to `test/fixtures/` by this PR are `.py` files, and there are already some `.py` files in `test/fixtures/`. Is there a clear rationale for what `.py` files belong in `test/fixtures/` versus some other location? I **think** there is and that I understand it, but I want to confirm my understanding is correct.

alanmcruickshank:
> I notice that some of the files being moved to test/fixtures/ by this PR are .py files

No files have been move *to* `test/fixtures/` in this PR, but there are some files which were already there which I have reorganized so that the structure makes sense.

For `.py` files in fixtures, they should be files which are designed to mimic usage of python files by the user. In the case it does make sense because they're `.py` files which define custom rules in the way that a user might define custom rules to import into sqlfluff.

barrywhart:
Got it, thanks! 🚀

PR #898, opened 2021-03-26
Use dbt templater outside of dbt project project root with project_dir config

pvonglehn:
This PR is my attempt to get round the linked issue - that sqlfluff with dbt templater must be run from the dbt project directory. This is makes it difficult to include sqlfluff in pre-commit hooks when the root of the git repo is not the same as the dbt project root.

This PR is still quite a rough draft but as this is my first time contributing I was keen to get some early feedback.

### Current situation
You can specify the dbt project directory in the `.sqlfluff` config file:
```
[sqlfluff:templater:dbt]
project_dir = <absolute path to dbt project directory>
```
But when running `sqlfluff` from outside this directory an error is thrown by the dbt templater module `File %s was not found in dbt project`.

The [dbt templater](https://github.com/sqlfluff/sqlfluff/blob/master/src/sqlfluff/core/templaters/dbt.py) uses the PathSelectorMethod from [dbt source code](https://github.com/fishtown-analytics/dbt/blob/develop/core/dbt/graph/selector_methods.py) to extract nodes from the dbt manifest corresponding to the current file being linted.

I believe the root cause of this present issue is that the PathSelectorMethod sets the root of the search to the current working directory.

`root = Path.cwd()`

So my proposed solution is to create a new subclass which allows the root to be set to the dbt_project directory instead.

Not sure how good an architectural decision this is creating a dbt native class here in SQLFluff - open to alternative suggestions.


joppevos:
@pvonglehn You can set the entry in pre-commit to the folder of your dbt project like this:
```
    hooks:
      - id: sqlfluff-lint
        entry: sqlfluff lint my_dbt_project
```
Then everything will work fine 👍

pvonglehn:
thanks @joppevos - that is indeed a much more straightforward solution

pvonglehn:
This PR is not ready to merge (e.g. CI tests haven't past), but I've opened for review because I'm interested in feedback as to whether this is a reasonable approach or indeed whether it is worth implementing at all. @joppevos has provided a workaround that handles the pre-commit use case I mentioned in the description. On the other hand I guess there would be some other use cases for this.

NiallRees:
Hey @pvonglehn I still think this would be a helpful contribution. It would be nice to be able to have the exact same functionality of running `sqlfluff` commands in any directory like any other templater. I realise it's been a while, but would love to get this merged if you're able to get it over the line!

pvonglehn (commented):
> Hey @pvonglehn I still think this would be a helpful contribution. It would be nice to be able to have the exact same functionality of running `sqlfluff` commands in any directory like any other templater. I realise it's been a while, but would love to get this merged if you're able to get it over the line!

Hi @NiallRees - more than happy to keep going on this. Do you think I'm heading broadly in the right direction? If so next steps I guess would be to get all CI tests to pass, write a unit test for this specific functionality and then ask for review? Appreciate any guidance as I'm new to open source contributing.

With this current change the user would still need to specify the dbt project directory in the .sqlfluff file, which I think is a good start as the functionality is partially in place but incomplete as it stands. But if we wanted sqlfluff commands to work from any directory without having to specify this we would need to come up with another strategy.
```
[sqlfluff:templater:dbt]
project_dir = <absolute path to dbt project directory>
```

NiallRees:
> > Hey @pvonglehn I still think this would be a helpful contribution. It would be nice to be able to have the exact same functionality of running `sqlfluff` commands in any directory like any other templater. I realise it's been a while, but would love to get this merged if you're able to get it over the line!
> 
> Hi @NiallRees - more than happy to keep going on this. Do you think I'm heading broadly in the right direction? If so next steps I guess would be to get all CI tests to pass, write a unit test for this specific functionality and then ask for review? Appreciate any guidance as I'm new to open source contributing.
> 
> With this current change the user would still need to specify the dbt project directory in the .sqlfluff file, which I think is a good start as the functionality is partially in place but incomplete as it stands. But if we wanted sqlfluff commands to work from any directory without having to specify this we would need to come up with another strategy.
> 
> ```
> [sqlfluff:templater:dbt]
> project_dir = <absolute path to dbt project directory>
> ```

It's absolutely reasonable to require the user to provide the path of the dbt project IMO. I wonder though if we should avoid requiring an absolute path (which would change between environments), one solution could be requiring a relative path to the dbt project root? Any other ideas would be good to hear too.

src/sqlfluff/core/templaters/dbt.py
```diff
@@ -104,7 +131,12 @@ def identity(x):
     @cached_property
     def dbt_selector_method(self):
         """Loads the dbt selector method."""
-        if "0.17" in self.dbt_version:
+        if self._get_project_dir() != os.getcwd():
+            self.dbt_selector_method = PathSelectorMethodWithRoot(
+                                                Path(self._get_project_dir()),
+                                                manifest = self.dbt_manifest, previous_state=None, arguments=[]
+                                            )
+        elif "0.17" in self.dbt_version:
```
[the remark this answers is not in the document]
NiallRees: > > Hey @pvonglehn I still think this would be a helpful contribution. It would be nice to be able to have the exact same functionality of running `sqlfluff` commands in any directory like any other templater. I realise it's been a while, but would love to get this merged if you're able to get it over the line!
> 
> Hi @NiallRees - more than happy to keep going on this. Do you think I'm heading broadly in the right direction? If so next steps I guess would be to get all CI tests to pass, write a unit test for this specific functionality and then ask for review? Appreciate any guidance as I'm new to open source contributing.
> 
> With this current change the user would still need to specify the dbt project directory in the .sqlfluff file, which I think is a good start as the functionality is partially in place but incomplete as it stands. But if we wanted sqlfluff commands to work from any directory without having to specify this we would need to come up with another strategy.
> 
> ```
> [sqlfluff:templater:dbt]
> project_dir = <absolute path to dbt project directory>
> ```

Have you considered temporarily switching the directory of the running process to that of the dbt project and then manipulating any file paths that may have been provided as an argument to `sqlfluff`? Any solution which avoids taking on even more internal code dependencies on dbt would require much less maintenance going forward.

pvonglehn:
> > > Hey @pvonglehn I still think this would be a helpful contribution. It would be nice to be able to have the exact same functionality of running `sqlfluff` commands in any directory like any other templater. I realise it's been a while, but would love to get this merged if you're able to get it over the line!
> > 
> > 
> > Hi @NiallRees - more than happy to keep going on this. Do you think I'm heading broadly in the right direction? If so next steps I guess would be to get all CI tests to pass, write a unit test for this specific functionality and then ask for review? Appreciate any guidance as I'm new to open source contributing.
> > With this current change the user would still need to specify the dbt project directory in the .sqlfluff file, which I think is a good start as the functionality is partially in place but incomplete as it stands. But if we wanted sqlfluff commands to work from any directory without having to specify this we would need to come up with another strategy.
> > ```
> > [sqlfluff:templater:dbt]
> > project_dir = <absolute path to dbt project directory>
> > ```
> 
> It's absolutely reasonable to require the user to provide the path of the dbt project IMO. I wonder though if we should avoid requiring an absolute path (which would change between environments), one solution could be requiring a relative path to the dbt project root? Any other ideas would be good to hear too.

Hey @NiallRees apologies for the delay, I got some time to take another look at this. I've given it another shot this time with changing directory approach as you suggested. The user has to specify the project_dir in the .sqlfluff file. I agree this is reasonable. It works with relative or absolute paths.

pvonglehn:
Just realised that this way of doing things results in the `target ` directory being created outside of the dbt project directory which is not desirable. Will rewrite so that we change directory before compiling the sql with dbt

NiallRees (changes requested):
This is looking really great @pvonglehn, a couple of thoughts:

- Changing directory inside of the `_unsafe_process` function means that if there's an exception raised during model compilation, the user's working directory won't be returned to their initial. What do you think about moving the `os.chdir()` call to `process()` under the `try:` block, and then adding a `finally:` block to call `os.chdir(self.working_dir)`? That way we'll always return the user to their initial working directory even if something bad happens.
- An idea on how to test this. Now that the dbt templater handles moving to the correct directory internally, the existing dbt templater tests should use that functionality, which would also give coverage on whether this new behaviour works correctly. If you see [here](https://github.com/sqlfluff/sqlfluff/blob/master/test/core/templaters/dbt_test.py#L110), there's an `in_dbt_project_dir` [pytest fixture](https://github.com/sqlfluff/sqlfluff/blob/master/test/fixtures/dbt/templater.py#L28) which temporarily moves the working directory to the directory of the project. We can get rid of this now and simply specify the new configuration parameters you've created in [here](https://github.com/sqlfluff/sqlfluff/blob/master/test/fixtures/dbt/templater.py#L8).

Let me know if I've not explained any of this well! Once we've got these two down I think we'll be good to merge. Thanks again for contributing to this - it's a really valuable addition.

pvonglehn:
> This is looking really great @pvonglehn, a couple of thoughts:
> 
> * Changing directory inside of the `_unsafe_process` function means that if there's an exception raised during model compilation, the user's working directory won't be returned to their initial. What do you think about moving the `os.chdir()` call to `process()` under the `try:` block, and then adding a `finally:` block to call `os.chdir(self.working_dir)`? That way we'll always return the user to their initial working directory even if something bad happens.
> * An idea on how to test this. Now that the dbt templater handles moving to the correct directory internally, the existing dbt templater tests should use that functionality, which would also give coverage on whether this new behaviour works correctly. If you see [here](https://github.com/sqlfluff/sqlfluff/blob/master/test/core/templaters/dbt_test.py#L110), there's an `in_dbt_project_dir` [pytest fixture](https://github.com/sqlfluff/sqlfluff/blob/master/test/fixtures/dbt/templater.py#L28) which temporarily moves the working directory to the directory of the project. We can get rid of this now and simply specify the new configuration parameters you've created in [here](https://github.com/sqlfluff/sqlfluff/blob/master/test/fixtures/dbt/templater.py#L8).
> 
> Let me know if I've not explained any of this well! Once we've got these two down I think we'll be good to merge. Thanks again for contributing to this - it's a really valuable addition.

Nice improvement putting the directory changes in the try except finally block. I've made the changes. Now working on the testing.

pvonglehn:
@NiallRees I'm having some difficulties getting `linter_test.py` and `std_test.py` to pass with these changes.

In these tests I pass ` DBT_FLUFF_CONFIG` to the `Linter ` when it gets instantiated but this config gets overwritten at some point, I think by the contents of `test/fixtures/dbt_project/.sqlfluff`. 

I could try to find a fix to get this test to run but I think it is highlighting an issue we may need to address.

I see now that the changes we made in this PR only work in the case where `project_dir` specified in the `.sqlfluff ` file is correct relative to the directory in which sqlfluff is launched. If the `.sqlfluff` file is in the dbt_project directory itself and the project_dir is set to "./" for example we wouldn't change to the correct directory to do the linting.

Any suggestions for how to proceed?

pvonglehn:
> @NiallRees I'm having some difficulties getting `linter_test.py` and `std_test.py` to pass with these changes.
> 
> In these tests I pass ` DBT_FLUFF_CONFIG` to the `Linter ` when it gets instantiated but this config gets overwritten at some point, I think by the contents of `test/fixtures/dbt_project/.sqlfluff`.
> 
> I could try to find a fix to get this test to run but I think it is highlighting an issue we may need to address.
> 
> I see now that the changes we made in this PR only work in the case where `project_dir` specified in the `.sqlfluff ` file is correct relative to the directory in which sqlfluff is launched. If the `.sqlfluff` file is in the dbt_project directory itself and the project_dir is set to "./" for example we wouldn't change to the correct directory to do the linting.
> 
> Any suggestions for how to proceed?

Maybe we could make the project_dir path relative to the path of the .sqlfluff file itself and expand it to an absolute path when parsing the .sqlfluff file.

NiallRees:
> > @NiallRees I'm having some difficulties getting `linter_test.py` and `std_test.py` to pass with these changes.
> > In these tests I pass ` DBT_FLUFF_CONFIG` to the `Linter ` when it gets instantiated but this config gets overwritten at some point, I think by the contents of `test/fixtures/dbt_project/.sqlfluff`.
> > I could try to find a fix to get this test to run but I think it is highlighting an issue we may need to address.
> > I see now that the changes we made in this PR only work in the case where `project_dir` specified in the `.sqlfluff ` file is correct relative to the directory in which sqlfluff is launched. If the `.sqlfluff` file is in the dbt_project directory itself and the project_dir is set to "./" for example we wouldn't change to the correct directory to do the linting.
> > Any suggestions for how to proceed?
> 
> Maybe we could make the project_dir path relative to the path of the .sqlfluff file itself and expand it to an absolute path when parsing the .sqlfluff file.

I think that's a good idea. CC'ing @alanmcruickshank as he might have a better idea or things to watch out for.

alanmcruickshank:
> Maybe we could make the project_dir path relative to the path of the .sqlfluff file itself and expand it to an absolute path when parsing the .sqlfluff file.

It think this is the way forward ☝️ . If you check out `config.py` you'll see that at the moment, anything ending in `_path` already does relative path resolution to the location of the `.sqlfluff` file (specifically in `_get_config_elems_from_file`). You could either change the config variable to end in `_path` or I'd be happy extending the relative path code to also cover config values ending in `_dir`.

pvonglehn:
> > Maybe we could make the project_dir path relative to the path of the .sqlfluff file itself and expand it to an absolute path when parsing the .sqlfluff file.
> 
> It think this is the way forward  . If you check out `config.py` you'll see that at the moment, anything ending in `_path` already does relative path resolution to the location of the `.sqlfluff` file (specifically in `_get_config_elems_from_file`). You could either change the config variable to end in `_path` or I'd be happy extending the relative path code to also cover config values ending in `_dir`.

OK great that there is existing relative path resolution that we can use. I should have a chance to attempt this at some point this week

pvonglehn:
> > Maybe we could make the project_dir path relative to the path of the .sqlfluff file itself and expand it to an absolute path when parsing the .sqlfluff file.
> 
> It think this is the way forward  . If you check out `config.py` you'll see that at the moment, anything ending in `_path` already does relative path resolution to the location of the `.sqlfluff` file (specifically in `_get_config_elems_from_file`). You could either change the config variable to end in `_path` or I'd be happy extending the relative path code to also cover config values ending in `_dir`.

@alanmcruickshank I extended the relative path code to also cover config values ending in `_dir`. I have run into an issue though:

The nested updating/overridering of configuration with local `.sqlfluff `file doesn't seem to update `Linter.templater`.

e.g. If the `.sqlfuff` file is in the dbt project_dir but we run sqlfluff from a different directory the templater will be set to jinja (from default) but is not updated to dbt even though the config is correctly read. I don't fully understand the program flow yet but I think this is because the templater is set by the `__init__()` constructor of the Linter class with the config from either the `.sqlfuff` file in the root directory (if present) or the default config. Additional config from nested `.sqlfluff` files seem to be read and added later so the `Linter.templater` is not update. 

Is this the correct interpretation? Any thoughts on steps forward?

There is a comment in linter.py "# Just use the local config from here:" Maybe we can use the local config from earlier on in the flow? Or we could explicitly update `Linter.templater` when local config is read in. 
https://github.com/sqlfluff/sqlfluff/blob/459f1c2c0b3d639436f29e684134d12e877e64f6/src/sqlfluff/core/linter.py#L865

NiallRees:
Looking good @pvonglehn !

NiallRees:
@pvonglehn Is this ready for review?

NiallRees:
@pvonglehn Your suggestion of warning the user when they try to change templater in a nested .sqlfluff file makes sense to me. If we decide to add that functionality we can do it separately.

barrywhart:
@pvonglehn: I'm following up on your question from a few weeks ago:
>The nested updating/overridering of configuration with local .sqlfluff file doesn't seem to update Linter.templater.

We'd be happy to look into this more closely. Can you tell us how to reproduce this behavior? I'm not very familiar with this area of the code, but if I can reproduce it, I can look around and try to help move this forward.

pvonglehn (commented):
> @pvonglehn: I'm following up on your question from a few weeks ago:
> 
> > The nested updating/overridering of configuration with local .sqlfluff file doesn't seem to update Linter.templater.
> 
> We'd be happy to look into this more closely. Can you tell us how to reproduce this behavior? I'm not very familiar with this area of the code, but if I can reproduce it, I can look around and try to help move this forward.

Hi @barrywhart you can reproduce this by setting up the following directory structure:

`.
└── subdir
    ├── .sqlfluff
    └── test.sql
`

in the .sqlfluff add
`[sqlfluff]
templater = dbt`

if you run `sqlfluff lint -v` from the root directory the file will lint (sucessfully) with the default jinja templater
if you run the same command from the `subdir` sqlfluff will try to lint with dbt (but will fail in this example because there is no dbt_project.yml)

This is with the current master sqlfluff, not this branch.

The templater is not updated with the nested config because the Linter object instance attribute `self.templater` gets set on instantiation (with whatever templater is set in the root directory .sqlfluff file or the default config) and not updated with the nested local .sqlfluff config files, which get read in after instantiation of the Linter object.

src/sqlfluff/cli/formatters.py
```diff
@@ -236,9 +236,9 @@ def _format_config(self, linter):
                 ("sqlfluff", get_package_version()),
                 ("python", get_python_version()),
                 ("implementation", get_python_implementation()),
-                ("dialect", linter.dialect.name),
+                ("dialect", linter.config.get("dialect_obj").name),
                 ("verbosity", self._verbosity),
-            ] + linter.templater.config_pairs()
+            ] + linter.config.get("templater_obj").config_pairs()
```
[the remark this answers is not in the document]
dmateusp: This might be big enough to separate into its own PR and issue wdyt? (My concern is that the conversation on this specific matter delays changes that are unrelated)

Without this change, can this PR still be used to lint a multi-dbt-project repository?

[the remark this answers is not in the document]
pvonglehn: > This might be big enough to separate into its own PR and issue wdyt? (My concern is that the conversation on this specific matter delays changes that are unrelated)
> 
> Without this change, can this PR still be used to lint a multi-dbt-project repository?

@dmateusp Good point. I can simplify this PR so that it only covers the functionality to run sqlfluff from outside the dbt directory and leave the issue of changing/not changing templaters with nested .sqlfluff files for another issue.

test/fixtures/dbt_project/.sqlfluff
```diff
@@ -5,3 +5,4 @@ dialect = postgres
 
 [sqlfluff:templater:dbt]
 profiles_dir = ../dbt
+project_dir = ./
```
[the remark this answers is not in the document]
pvonglehn: When running the existing tests we are now testing the functionality of linting while outside the project directory because we no longer change to the dbt project directory as part of the test fixture [in_dbt_project_dir()](https://github.com/sqlfluff/sqlfluff/pull/898/files#diff-02cf6c0e51d0bea24933771510432fa6d2b094d555c4ef5bfb53f65f075a1493L28). Instead we pass the fixture [project_dir()](in_dbt_project_dir) which is just the path of the dbt project and we change to that directory from outside as part on the new code int dbt.py.

Still, we're not testing the reading in the reading in of the project_dir from an .sqlfluff file so I'd be happy to add another test where we change directory to a directory with a .sqlfluff file and a subdirectory dbt_project.

src/sqlfluff/core/templaters/dbt.py
```diff
@@ -151,23 +154,27 @@ def _get_profiles_dir(self):
         """
         from dbt.config.profile import PROFILES_DIR
 
-        return os.path.expanduser(
-            self.sqlfluff_config.get_section(
-                (self.templater_selector, self.name, "profiles_dir")
+        return os.path.abspath(
```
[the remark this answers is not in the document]
pvonglehn: Yes we're changing into the project_dir to lint the models. I wanted `DbtTemplater.dbt_project_dir` and `DbtTemplater.working_dir` to have absolute paths to make sure we ended up in the correct place switching between the two.

But looking back I don't think this is actually necessary especially with the changes resolving the path of variables ending in `_dir` [here](DbtTemplater.)

docs/source/configuration.rst
```diff
@@ -319,10 +319,9 @@ In *.sqlfluffignore*:
 Known Caveats
 ^^^^^^^^^^^^^
 
-- In SQLFluff 0.4.0 dbt templating only works if SQLFluff CLI commands
-  are invoked from the dbt project's root directory (containing
-  `dbt_project.yml`). There is an issue to address this:
-  https://github.com/sqlfluff/sqlfluff/issues/601
+- To use the templater, you must set `templater = dbt` in a `.sqlfluff`
```
dmateusp: thanks that's a very useful note :+1:

src/sqlfluff/core/linter/linter.py
```diff
@@ -519,6 +519,10 @@ def render_string(
         # Start the templating timer
         t0 = time.monotonic()
 
+        if not config.get("templater_obj") == self.templater:
+            linter_logger.warn((f"Attempt to set templater to {config.get('templater_obj').name} failed. Using {self.templater.name} templater. "
```
dmateusp: awesome :+1:

docs/source/configuration.rst
```diff
@@ -346,13 +348,19 @@ In *.sqlfluffignore*:
     dbt_modules/
     macros/
 
+You can set the dbt project directory, profiles directory and profile with:
+
+[sqlfluff:templater:dbt]
+project_dir = <relative or absolute path to dbt_project directory>
+profiles_dir = <relative or absolute path to dbt_project directory>
```
dmateusp, line 355:
```suggestion
profiles_dir = <relative or absolute path to the directory that contains the profiles.yml file>
```

docs/source/configuration.rst
```diff
@@ -346,13 +348,19 @@ In *.sqlfluffignore*:
...
+profile = <dbt profile>
+
 Known Caveats
 ^^^^^^^^^^^^^
 
-- In SQLFluff 0.4.0 dbt templating only works if SQLFluff CLI commands
-  are invoked from the dbt project's root directory (containing
-  `dbt_project.yml`). There is an issue to address this:
-  https://github.com/sqlfluff/sqlfluff/issues/601
+- To use the dbt templater, you must set `templater = dbt` in the `.sqlfluff`
```
dmateusp, line 361: that note is very helpful :+1:

src/sqlfluff/core/linter/linter.py
```diff
@@ -519,6 +519,14 @@ def render_string(
         # Start the templating timer
         t0 = time.monotonic()
 
+        if not config.get("templater_obj") == self.templater:
+            linter_logger.warn(
```
dmateusp, line 523: codecov is complaining here, this is a very useful message so I'd like to keep it ^^

Could you add a test that just checks the templater not being updated by .sqlfluff in a subdir?

src/sqlfluff/core/templaters/dbt.py
```diff
@@ -151,24 +154,41 @@ def _get_profiles_dir(self):
...
+            os.path.expanduser(
+                self.sqlfluff_config.get_section(
+                    (self.templater_selector, self.name, "project_dir")
+                )
+                or os.getcwd()
             )
-            or os.getcwd()
         )
+        if not os.path.exists(dbt_project_dir):
+            templater_logger.error(
```
dmateusp, line 187: codecov complaining too ^^

Actually I wonder, should we be actually throwing an Exception here ? I don't think there's much we can do if the dbt project directory isn't found right?

pvonglehn: @dmateusp I initially tried throwing an `IOError` exception here but this was handled in [commands.py](https://github.com/sqlfluff/sqlfluff/blob/master/src/sqlfluff/cli/commands.py#L353) with a custom (not relevant) error message, so the message I wanted to show here never gets shown.

As the code is now we will get an exception later anyway when we try to change directory into a non existing directory, but I agree an exception here would be more appropriate. Any suggestions for how to get round `commands.py `obscuring the `IOError `exception? Should we change that code or add a new exception class here e.g. `ConfigParsingError`?

I've added a test to cover these lines [here](https://github.com/sqlfluff/sqlfluff/pull/898/files#diff-5d3be0ba02e85767f3906e01bdb3611ce1c9defc129ba33dde3c2ecd6e6727e0R225)

src/sqlfluff/core/linter/linter.py
pvonglehn: @dmateusp I've added a test to cover these lines [here](https://github.com/sqlfluff/sqlfluff/pull/898/files#diff-fe70b9c424984a64d62d1fd3dc9c3b3877ea0d2de3331b2f27fd56863bd57520R595)

test/core/linter_test.py
```diff
@@ -585,3 +589,16 @@ def test_delayed_exception():
...
+def test__attempt_to_change_templater_warning(caplog):
+    """Test warning if user tries to change templater in .sqlfluff file in subdirectory"""
+    initial_config = FluffConfig(configs={"core": {"templater": "jinja"}})
+    lntr = Linter(config=initial_config)
+    updated_config = FluffConfig(configs={"core": {"templater": "dbt"}})
+    with caplog.at_level(logging.WARNING):
+        rendered_string = lntr.render_string(
+            in_str="select * from table", fname="test.sql", config=updated_config
+        )
+    assert f"Attempt to set templater to {updated_config.get('templater_obj').name} failed. in caplog.text"
```
pvonglehn, line 604: @dmateusp I didn't explicitly write and use new .sqlfluff and subdirectory fixtures but this tests the same functionality. Is this OK or do you think it is worth setting up some new fixtures?

src/sqlfluff/core/linter/linter.py
pvonglehn, line 523: @NiallRees this is the warning we discussed

NiallRees: This is nice 👍

test/core/linter_test.py
```diff
@@ -585,3 +589,15 @@ def test_delayed_exception():
...
+def test__attempt_to_change_templater_warning(caplog):
+    """Test warning if user tries to change templater in .sqlfluff file in subdirectory."""
+    initial_config = FluffConfig(configs={"core": {"templater": "jinja"}})
+    lntr = Linter(config=initial_config)
+    updated_config = FluffConfig(configs={"core": {"templater": "dbt"}})
+    with caplog.at_level(logging.WARNING):
+        lntr.render_string(
+            in_str="select * from table", fname="test.sql", config=updated_config
+        )
+    assert f"Attempt to set templater to {updated_config.get('templater_obj').name} failed. in caplog.text"
```
alanmcruickshank, line 603: I think here you meant to have the `in caplog.text` _outside_ the quotes, rather than just asserting that a non-empty string is truthy?

test/core/templaters/dbt_test.py
```diff
@@ -56,15 +57,15 @@ def test__templater_dbt_profiles_dir_expanded(dbt_templater):  # noqa
 )
 @pytest.mark.dbt
 def test__templater_dbt_templating_result(
-    in_dbt_project_dir, dbt_templater, fname  # noqa
+    project_dir, dbt_templater, fname  # noqa: F811  # noqa: F811
```
alanmcruickshank, line 60: Why `noqa` twice?

alanmcruickshank (changes requested):
I've just tested this on our large dbt project and it works! Great work @pvonglehn 🎉 

A couple of small things I noticed when reviewing the tests. Only one of which is meaningful and I think it's just a typo. Otherwise this looks good. 👍

test/core/linter_test.py
pvonglehn: woops. Thanks for spotting that @alanmcruickshank

test/core/templaters/dbt_test.py
pvonglehn: no idea why that's there, thanks.

pvonglehn:
thanks for the additions @NiallRees. No idea why `test__attempt_to_change_templater_warning` is failing on CICD. The test passes locally. I'm not so familiar with `caplog`. Any other suggestions for testing this?

pvonglehn:
> thanks for the additions @NiallRees. No idea why `test__attempt_to_change_templater_warning` is failing on CICD. The test passes locally. I'm not so familiar with `caplog`. Any other suggestions for testing this?

Strange, it passes locally when I run `tox -e py38 -- test/core/linter_test.py` but not when I run `tox -e py38 -- `

NiallRees:
> > thanks for the additions @NiallRees. No idea why `test__attempt_to_change_templater_warning` is failing on CICD. The test passes locally. I'm not so familiar with `caplog`. Any other suggestions for testing this?
> 
> Strange, it passes locally when I run `tox -e py38 -- test/core/linter_test.py` but not when I run `tox -e py38 -- `

I'm not sure either. I'll take another look today

alanmcruickshank:
This is very strange:
- Running the test file directly (`linter_test.py`) (in tox and pytest natively) succeeds (and I don't think it's an issue with the test directly because the output is good).
- Running the test folder (`test\core`) succeeds in both tox and pytest.
- Running the test suite without a path fails in both tox and pytest.

My only idea is that one of the other tests is playing with the logging configs in a way that isn't safe. @barrywhart - have you come up against anything like this before?

pvonglehn:
> This is very strange:
> 
> * Running the test file directly (`linter_test.py`) (in tox and pytest natively) succeeds (and I don't think it's an issue with the test directly because the output is good).
> * Running the test folder (`test\core`) succeeds in both tox and pytest.
> * Running the test suite without a path fails in both tox and pytest.
> 
> My only idea is that one of the other tests is playing with the logging configs in a way that isn't safe. @barrywhart - have you come up against anything like this before?

Found the cause.

It's the line:
`fluff_logger.propagate = False` in `commands.py` [here](https://github.com/sqlfluff/sqlfluff/blob/2f94f9d4ffe57421bba8b8eea54ff91fcb9e7252/src/sqlfluff/cli/commands.py#L67)

barrywhart:
I've seen weird stuff over the years. When it fails, what's the symptom? 

One weird thing I've seen with Python logging: if you initialize Python logging a second time, pre-existing loggers will be disabled. This can cause expected logging messages not to appear.

I'm on PTO this week but could look into this next week if needed.

test/core/linter_test.py
```diff
@@ -585,3 +589,21 @@ def test_delayed_exception():
...
+
+
+def test__attempt_to_change_templater_warning(caplog):
+    """Test warning if user tries to change templater in .sqlfluff file in subdirectory."""
+    initial_config = FluffConfig(configs={"core": {"templater": "jinja"}})
+    lntr = Linter(config=initial_config)
+    updated_config = FluffConfig(configs={"core": {"templater": "dbt"}})
+    logger = logging.getLogger("sqlfluff")
+    original_propagate_value = logger.propagate
+    try:
```
pvonglehn, line 601: Worked around the logging issue by temporarily setting the propagate value to `True` after reading[ this](https://stackoverflow.com/questions/59875983/why-is-caplog-text-empty-even-though-the-function-im-testing-is-logging) SO post.

I was concerned that playing with the logging values like this would give false positive tests but I also manually tested that this warning appears when it is supposed to and it does.

what do you think?
@alanmcruickshank, @barrywhart, @NiallRees

NiallRees: I was able to get this test to pass on commits before your temporary propagate switch when running with the whole suite by setting https://github.com/sqlfluff/sqlfluff/blob/master/src/sqlfluff/cli/commands.py#L67 to True. The CLI tests run before `linter_test.py` which must explain why this test passes in isolation but not when running the entire suite.

I'm happy with your solution here, though I wonder whether it makes sense to just set `fluff_logger.propagate = True` in the CLI module when testing so we don't run into this again. What do you think @alanmcruickshank ?

src/sqlfluff/core/templaters/dbt.py
dmateusp: I think you could create an exception like `DbtProjectFolderDoesNotExistError` (adding a new exception class like you suggested), then the `except OSError` won't handle it anymore and you can add another except clause to catch this specific exception

but now that we have coverage and logging I'm happy either way

test/core/linter_test.py
dmateusp: btw there's more reading on the "propagate" issue: https://github.com/pytest-dev/pytest/issues/3697

seems like there isn't a universal way to deal with it :(

NiallRees: I think I'm just going to merge this as we know what the issue is, we can revisit it later if needed. I'll make an issue to keep track.

pvonglehn: @NiallRees, @alanmcruickshank, @dmateusp, @barrywhart Thanks for the support with this issue. It was my first open source contribution and a positive experience (although more work than I expected haha).

NiallRees: Congratulations! You picked a tough first contribution 😄 - nice work.

PR #899, opened 2021-03-26
Issue 654 - add file encoding detection for read/write of files

### LINKED ISSUE #654: Add a command-line option for file encoding (was: File encoding is not respected)
When fixing a file the encoding is maybe not respected and the fixed code is written with default system encoding. This leads in losing special chars etc.

E.g. the original file is a UTF-8 file, with the fix the code is written in CP1252 on Windows. 

This happens because prior Python 3.9 `open()` uses the system default codepage, e.g. windows = cp1252. With Python 3.9 it's opened with UTF-8 encoding if no encoding is set (which even could lead to false results, if the file is with CP1252 encoded).

[Python 3.8 docs](https://docs.python.org/3.8/library/functions.html#open):

> encoding is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever locale.getpreferredencoding() returns), but any text encoding supported by Python can be used. See the codecs module for the list of supported encodings.

This happens here:
https://github.com/sqlfluff/sqlfluff/blob/ce673c5c7731c3c7e0fff76f77d49bc8e362d7f2/src/sqlfluff/core/linter.py#L471

Possible solutions:
1. auto detect
2. configurable encoding
3. passing encoding as command line parameter
4. writing a new file, instead of overwritting the current file

I would prefer to make the encoding configurable (pt. 2 or 3) to take the full control over it because even auto-detection could fail.

nevado:
Initial attempt to add file encoding detection for lint, parse and fix, to fix #654

Adds a new 'encoding' config option which can either be `autodetect` or a valid encoding e.g. `utf-8`, `utf-8-sig`.

Still to do:
- add another cli test that creates an input file with a given encoding, fixes the file, and then verifies that the written file has the correct encoding
- add --encoding option to cli

Otherwise, is this broadly along the right lines? First time contributor, so any feedback welcome


src/sqlfluff/core/file_helpers.py
```diff
@@ -0,0 +1,15 @@
...
+
+from sqlfluff.core.config import FluffConfig
+
+
+def get_encoding(fname: str, config: FluffConfig) -> str:
+    """Get the encoding of the file (autodetect)."""
+    encoding_config = config.get("encoding", default="autodetect")
+
+    if encoding_config == "autodetect":
+        return chardet.detect(open(fname, "rb").read())["encoding"]
```
barrywhart, line 13: Let's make sure to close the file. While CPython will automatically close the file when the reference goes out of scope, other versions may not (e.g. [PyPy](https://www.pypy.org/)).

alanmcruickshank:
This looks great. I'd love to see the cli option too, thanks for putting this together so far!

NiallRees:
Hey @nevado I'd love to get this merged, is there anything we can do to help with this PR? It would be a really great addition.

nevado:
Hi, sorry for the delay - from my side this is good to merge (I've just fixed the test errors), the only missing thing is the command line parameter which could be added later, I just haven't had time to do it yet :(

If people are happy to merge now let's do that, and I'll try to do the command line in the next week?

alanmcruickshank:
> If people are happy to merge now let's do that, and I'll try to do the command line in the next week?

Very happy to have the command line option as a seperate PR. 👍

barrywhart:
This PR seems like it may have been forgotten? I opened a new PR that has identical functionality but addresses the merge conflicts, which were pretty tricky due to the big "parallel" and dbt-related refactorings.

barrywhart:
@nevado: I got this PR updated (see #1177) and hope we will merge it this time. I'm sorry about the delay -- the maintainers have been busy with our day jobs and forgot to merge your PR. If you are still interested in adding CLI support in a separate PR, that would be 💯.

barrywhart:
Replaced by #1177. GitHub marked it as "merged" because the new PR was based on this one (i.e. same commits, plus more).

PR #903, opened 2021-03-27
Issue 895: [Bigquery] Hyphen symbol (-) in project or table names

### LINKED ISSUE #895: BigQuery: hyphen symbol (-) in table project name in BigQuery
In Big Query, we can query a table named "project-a.database.a" in which, project-a is project name. However, sqlfluff get parse error with these tables.

```sql
SELECT * FROM project-a.database.a
```

## Expected Behaviour

No parse error

## Observed Behaviour

[sqlfluff.core.errors.SQLParseError("Found unparsable section: '-a.database.a'")]

## Steps to Reproduce
```python
query = 'SELECT * FROM project-a.database.a'
r = Linter(dialect='bigquery').parse_string(query)
print(r.violations)
```

## Version
'0.4.1'

barrywhart:
Resolves #895


test/core/generate_parse_fixture_yml.py
```diff
@@ -7,7 +7,7 @@
 from sqlfluff.core import FluffConfig
 from sqlfluff.cli.commands import quoted_presenter
 
-from dialects.parse_fixtures import get_parse_fixtures, load_file
+from test.dialects.parse_fixtures import get_parse_fixtures, load_file
```
barrywhart, line 10: This import was already broken somehow. Could be a separate PR if this fix is needed sooner.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -353,3 +368,26 @@ class LiteralCoercionSegment(BaseSegment):
         OneOf("DATE", "DATETIME", "TIME", "TIMESTAMP"),
         Ref("QuotedLiteralSegment"),
     )
+
+
+@bigquery_dialect.segment()
```
barrywhart, line 373: Hurray for the new dialect inheritance support! This was pretty straightforward and avoids a lot of copy/paste.

test/fixtures/parser/ansi/describe_table.yml
```diff
@@ -1,8 +1,8 @@
 file:
   statement:
     describe_statement:
-    - keyword: describe
-    - identifier: table
-    - object_reference:
+      keyword: describe
```
barrywhart, line 4: I don't know why this changed.

test/fixtures/parser/snowflake/snowflake_describe_user.yml
```diff
@@ -1,8 +1,8 @@
 file:
   statement:
     describe_statement:
-    - keyword: describe
```
barrywhart, line 4: I don't know why this changed, either.

barrywhart:
@NiallRees, @dmateusp, @alanmcruickshank: I thought I'd made progress on this issue, but a failed test has turned up an interesting new issue.

With this change, now if a subtraction expression does not contain spaces, e.g. `LEVEL-1` below, it does not parse, probably because it's being lexed as a single token.
```
            SELECT last_name, employee_id id, manager_id mgr_id,
               CONNECT_BY_ISLEAF leaf, LEVEL,
               LPAD(' ', 2*LEVEL-1)||SYS_CONNECT_BY_PATH(last_name, '/') "PATH"
            FROM employees
```

This makes me wonder if the current approach is workable. Is there a way at  the **lexer** level to support `-` as part of a name in some places but not others? Perhaps we need to handle this at at the parser level instead, i.e. create a patched BigQuery version of `ObjectReferenceSegment` that also allows `MinusSegment` as a delimiter. (Note the currently supported delimiters are:
```
OneOf(Ref("DotSegment"), Sequence(Ref("DotSegment")))
```

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -353,3 +352,26 @@ class LiteralCoercionSegment(BaseSegment):
...
+class HyphenatedObjectReferenceSegment(ansi_dialect.get_segment("ObjectReferenceSegment")):  # type: ignore
+    type = "hyphenated_object_reference"
+    match_grammar = ansi_dialect.get_segment(
+        "ObjectReferenceSegment"
+    ).match_grammar.copy()
+    match_grammar.delimiter = OneOf(
```
barrywhart, line 363: This is a bit of a hack, but it avoids having to copy/paste lots of code from ANSI ObjectReferenceSegment. Should we consider having the `copy()` function support this directly somehow?

barrywhart:
I reimplemented this at the parser level to avoid the issues parsing subtraction expressions without spaces. Ready for review...

src/sqlfluff/core/rules/std/L025.py
```diff
@@ -53,7 +53,7 @@ def _lint_references_and_aliases(
         for r in references:
             tbl_ref = r.extract_reference(level=2)
             if tbl_ref:
-                tbl_refs.add(tbl_ref[0])
+                tbl_refs.add(tbl_ref.part)
```
barrywhart, line 56: Tidying up the code, only _slightly_ related to the PR. Here `.part` is referring to a field of `ObjectReferencePart`, a `NamedTuple` whose structure changes slightly in this PR (`segment` becomes `segments`).

src/sqlfluff/core/rules/std/L026.py
```diff
@@ -53,10 +53,10 @@ def _lint_references_and_aliases(
 
                 violation_buff.append(
                     LintResult(
-                        # Return the segment rather than the string
-                        anchor=tbl_ref[1],
+                        # Return the first segment rather than the string
+                        anchor=tbl_ref.segments[0],
```
barrywhart, line 57: Also tidying, as with L025, but here, L026 references the `ObjectReferencePart` that was renamed, `segment` -> `segments`. So here, using the names rather than subscripts helps surface an error (what was previously a `BaseSegment` reference is now a `List[BaseSegment]`).

src/sqlfluff/dialects/dialect_bigquery.py
alanmcruickshank: Very interesting question. I'm not aware of doing this a better way 🤔 

I'm ok with this as it is but I agree it feels a little gross.

alanmcruickshank (approved):
Good work! I can see that this wasn't an easy one 👍

jlorince:
@alanmcruickshank @barrywhart Heads up, this solution does not work and still leads to a parse error if the project name is something like `foo-123`. E.g. `select * from project-1.schema.table` will fail. Will file an issue on this shortly.

PR #904, opened 2021-03-28
Issue 778: L033 should prefer explicit over implicit, add fix routine

### LINKED ISSUE #778: Update L033 to enforce explicit UNION ALL or UNION DISTINCT
Was discussing this issue with @nolanbconaway, who raised in #general in slack. [Slack Discussion](https://sqlfluff.slack.com/archives/C01GPHR8J0G/p1613511556030200)

In its current state, L033 will not allow a simple `UNION`. This is problematic as `UNION` eliminates duplicates while `UNION ALL` does not.

Instead of enforcing a `UNION ALL` in the place of `UNION`, enforce an explicit `UNION ALL` or `UNION DISTINCT`.

barrywhart:
Resolves #778


src/sqlfluff/core/rules/base.py
```diff
@@ -99,7 +99,7 @@ class LintFix:
             to be moved *after* the edit), for an `edit` it implies the segment
             to be replaced.
         edit (:obj:`BaseSegment`, optional): For `edit` and `create` fixes, this
-            hold the segment, or iterable of segments to create to replace at the
```
barrywhart, line 102: Fixing a typo

src/sqlfluff/core/rules/std/L031.py
```diff
@@ -59,6 +59,7 @@ def _eval(self, segment, **kwargs):
                 return None
 
             from_expression = from_clause_segment.get_child("from_expression")
+            from_expression_element = None
```
barrywhart, line 62: Fixing an undefined variable (unrelated to #778, but discovered while working on it)

nolanbconaway (approved):
I can't speak much to the implementation but the test case looks like it covers the issue!

PR #907, opened 2021-03-29
Issue 610: Rule L031 false positive when table appears twice in join

### LINKED ISSUE #610: Fix causes duplicate table aliases by simply removing them when table appears twice in join
> L020 | Duplicate table alias

Since after applying `sqlfluff fix` on for example:
```sql
select *
from a a1
join b on a1.i = b.j
join a a2 on a2.i = b.k
```
becomes:
```sql
select *
from a
join b on a.i = b.j
join a on a.i = b.k
```

Workaround I think is to move part of the join to a `with` CTE, e.g. in this case:
```sql
with a2 as (select i from a)
select *
from a
join b on a.i = b.j
join a2 on a2.b.k
```

barrywhart:
Resolves #610

### DIFF
--- a/src/sqlfluff/core/rules/std/L031.py
+++ b/src/sqlfluff/core/rules/std/L031.py
@@ -59,6 +59,7 @@ class Rule_L031(BaseRule):
                 return None
 
             from_expression = from_clause_segment.get_child("from_expression")
+            from_expression_element = None
             if from_expression:
                 from_expression_element = from_expression.get_child(
                     "from_expression_element"
@@ -97,13 +98,8 @@ class Rule_L031(BaseRule):
             )
         return None
 
-    def _lint_aliases_in_join(
-        self, base_table, table_expression_segments, column_reference_segments, segment
-    ):
-        """Lint and fix all aliases in joins - except for self-joins."""
-        # A buffer to keep any violations.
-        violation_buff = []
-
+    @classmethod
+    def _filter_table_expressions(cls, base_table, table_expression_segments):
         for table_exp in table_expression_segments:
             table_ref = table_exp.get_child("table_expression").get_child(
                 "object_reference"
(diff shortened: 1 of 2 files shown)


src/sqlfluff/core/rules/std/L031.py
```diff
@@ -131,38 +141,70 @@ def _lint_aliases_in_join(
...
+        table_counts = Counter(ai.table_ref.raw for ai in to_check)
+        table_aliases = defaultdict(set)
+        for ai in to_check:
+            table_aliases[ai.table_ref.raw].add(ai.alias_identifier_ref.raw)
+        for alias_info in to_check:
+            # If the same table appears more than once in the FROM clause with
+            # different alias names, do not consider removing its aliases.
+            # The aliases may have been introduced simply to make each
+            # occurrence of the table independent within the query.
+            if (
```
barrywhart, line 169: The next few lines are the core of the bug fix. I am not 100% sure what the filter should be here, but I think this is at least a good start. Thoughts?

NiallRees:
Will we ever stop finding bugs in this rule??! 😅

src/sqlfluff/core/rules/std/L031.py
NiallRees: What's the deal with `len(table_aliases[alias_info.table_ref.raw]) > 1`?

barrywhart: In the bug report, they were using two separate aliases for the same table. In SQL, two different aliases for the same table allows them to behave "independently" when both are used in the same query, e.g. table `b`, alias `b1` can refer to a different record or set of records than table `b`, alias `b2`. We don't want the lint rule to remove those two aliases, even if they don't appear in the select targets, because then they are forced to reference the same set of records, potentially producing a different result set.

So I'm checking to see if the same table appears with multiple aliases, and that prevents removing the aliases.

Does this answer your question? I know this is kind of a subtle and unusual SQL thing, so I'm happy to continue the conversation as needed.

NiallRees: Ah I see what's going on here now. I'm familiar with this practice in SQL but the defaultdict(set) part bypassed me initially. Thanks!

NiallRees (approved):
Nice work.

PR #909, opened 2021-03-29
Issue 876: Clean up AliasedTableReferenceSegment

### LINKED ISSUE #876: Clean up AliasedTableReferenceSegment grammar
This is a tech debt issue; nothing of concern to users.

In working on #865, Niall and I noticed something a bit odd about `AliasedTableReferenceSegment`. In order to add support for using table aliases in `UPDATE` statements, I had changed `UpdateStatementSegment` from:

```
Ref("TableReferenceSegment")
```

to:

```
Ref("AliasedTableReferenceSegment")
```

This worked, and all tests (new and old) passed because the match grammar for `AliasedTableReferenceSegment` specifies that the alias part is optional:

```
Ref("TableReferenceSegment"), Ref("AliasExpressionSegment", optional=True)
```

But... this resulted in the parse tree looking odd. (Note the nested `table_reference` nodes.)
```
    update_statement:
      keyword: UPDATE
      table_reference:
        table_reference:
          identifier: table_name
```

To avoid this, we updated my change to instead be:
```
OneOf(Ref("TableReferenceSegment"), Ref("AliasedTableReferenceSegment"))
```

This avoids the nesting, but it'd be better if we didn't have to specify `OneOf()` to make the parse tree look reasonable.  `AliasedTableReferenceSegment` supports the exact SQL syntax we want. Can we refactor the ANSI dialect a bit so the parse tree looks okay if we do this?

Another (more minor) question: The name `AliasedTableReferenceSegment` suggests that there's an alias present, but it's actually optional. Should we consider renaming it to make this clearer?

barrywhart:
Resolves #876 

The coverage check is failing for a bogus reason. It says the PR reduces code coverage by 0.01%, but that's because the PR actually makes the code smaller, and all the lines it removes were previously covered. There was no increase in **uncovered** code, just a decrease in the total amount of code. 😭


test/fixtures/parser/mysql/create_table.yml
```diff
@@ -1,43 +1,43 @@
-file:
```
barrywhart, line 1: I'm not sure why this file shows as "changed". I wonder if the line endings changed?

nolanbconaway (approved):
🟢

PR #918, opened 2021-04-04
Issue 916: L034 should report the warning on the earlier, more complex select target

### LINKED ISSUE #916: L034 should report the warning on the earlier, more complex select target
@pwildenhain: Thanks for looking at this. I am seeing the same behavior. I think the reason for this is that the "noqa" filter applies based on where the line rule **reports** the error. Notice that the two L034 errors are reported at "L: 7" and L: 10", the lines with `field3` and `field4`.

if you look at the implementation of L034, you'll see why -- it reports the warning when it finds a **simple** element following a more complex one, and the anchor (which determines the line and position of the warning) is the later, simpler segment:
https://github.com/sqlfluff/sqlfluff/blob/master/src/sqlfluff/core/rules/std/L034.py#L38LL42

As a test, I moved the "disable/enable L034" code around line 7 (rather than line 5), and it seems to behave as expected (the L034 report on line 7 goes away).

```
SELECT
    FIELD_1,
    FIELD_2,
    max(field_1) as field_1_max,
    -- noqa: disable=L034
    field_3,
    -- noqa: enable=L034
    -- noqa: disable=L009,L010,L014
    max(field_3) as max_field_3,
    field_4
from MY_TABLE
WHERE
    field_1 > 1 and field_2 > 2 and field_3 > 3
```

Output:
```
L:   4 | P:   9 | L014 | Inconsistent capitalisation of unquoted identifiers.
L:   4 | P:  18 | L010 | Inconsistent capitalisation of keywords.
L:   4 | P:  21 | L014 | Inconsistent capitalisation of unquoted identifiers.
L:   6 | P:   5 | L014 | Inconsistent capitalisation of unquoted identifiers.
L:  10 | P:   5 | L034 | Use wildcards then simple select targets before
                       | calculations and aggregates.
```

I think your point is a good one -- now that I'm looking at this, perhaps L034 should report the warning on the earlier, more complex select target. I think that should be a separate issue and PR from this work.

_Originally posted by @barrywhart in https://github.com/sqlfluff/sqlfluff/issues/891#issuecomment-810636517_

barrywhart:
Resolves #916

### DIFF
--- a/src/sqlfluff/testing/rules.py
+++ b/src/sqlfluff/testing/rules.py
@@ -70,10 +70,11 @@ def assert_rule_fail_in_sql(code, sql, configs=None, line_numbers=None):
     if line_numbers:
         actual_line_numbers = [e.line_no() for e in lerrs]
         if line_numbers != actual_line_numbers:
-            pytest.fail("Expected errors on lines {0}, but got errors on lines {1}".format(
-                line_numbers,
-                actual_line_numbers
-            ))
+            pytest.fail(
+                "Expected errors on lines {0}, but got errors on lines {1}".format(
+                    line_numbers, actual_line_numbers
+                )
+            )
     # The query should already have been fixed if possible so just return the raw.
     return linted.tree.raw
 
@@ -124,7 +125,8 @@ def rules__test_helper(test_case):
             test_case.rule,
             test_case.fail_str,
             configs=test_case.configs,
-            line_numbers=test_case.line_numbers)
+            line_numbers=test_case.line_numbers,
+        )
         # If a `fixed` value is provided then check it matches
         if test_case.fix_str:
             assert res == test_case.fix_str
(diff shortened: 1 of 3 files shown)


src/sqlfluff/testing/rules.py
```diff
@@ -66,6 +67,14 @@ def assert_rule_fail_in_sql(code, sql, configs=None):
             "No {0} failures found in query which should fail.".format(code),
             pytrace=False,
         )
+    if line_numbers:
```
barrywhart, line 70: Here, we add the ability to check the line number locations of the lint errors produced by a YAML test.

alanmcruickshank (approved):
Nice. I've wondered whether we should be testing not just line numbers but also positions in the `yaml` tests, and this is a good step in that direction.

PR #919, opened 2021-04-04
Add CREATE ROLE

### LINKED ISSUE #349: [postgres] Support CREATE ROLE
Please support `CREATE ROLE` in the PostgreSQL dialect:
* https://www.postgresql.org/docs/current/sql-createrole.html

Simple failing example:
```sql
CREATE ROLE web_anon nologin;
```
...results into:
```
== [db/test.sql] FAIL
L:   1 | P:   1 |  PRS | Found unparsable section: 'CREATE ROLE web_anon nologin'
```

alanmcruickshank:
Resolves #349 (plus snowflake and hooks for other dialects)


alanmcruickshank:
One thing I'm wondering with this PR is that `snowflake_dialect` already defined a more general `CREATE` statement, but that other dialects (including the `ansi` dialect) had stated defining individual `CREATE x` segments. This is an example of doing more of the latter, but it does result in two ways of parsing `CREATE ROLE` in snowflake. The precedence order means that `create_role_statement` overrides `create_statement`, but I wonder if we should normalise all dialects onto one or the other in this case? I'm pretty sure that we _should_, I'm just not sure which route I prefer.

barrywhart:
@alanmcruickshank: On the question about how to structure `CREATE` segment(s), my concerns with having one grand combined segment are:
* Does it (or will it in the future) be too "loose", allowing illegal things to parse correctly? I notice with the current Snowflake dialect, all of the `CREATE`s can end with a `SelectStatementSegment`, which seems wrong.
* Does one structure or the other make it easier to create linter rules for those areas of SQL? My intuition is that having separate segment types makes it easier, because the rule can look for the high-level segment type rather than looking at the individual keywords. Over time, I think we should favor the structure that makes it easier to create rules, because I think the dialects will eventually be pretty stable, while the set of rules will keep growing and changing (keep in mind the new support for plug-in rules, where organizations will be creating their own rules).

barrywhart (approved):
Looks good! I commented separately on your question about how we should structure the various `CREATE` segments.

NiallRees:
@alanmcruickshank this still on your radar or shall I pick it up?

NiallRees:
> CREATE

Will do this in

NiallRees:
> @NiallRees - if you're able to pick it up that would be great. I'm going to be occupied with the end of the big refactor for another few days still (although getting very close!). Given this was more than the initial 2 min job I hoped, I haven't been able to pick it up again yet.
> 
> Some rationalization of all of the different `CREATE` statements across all dialects would be very helpful. 🙏

Will create a separate issue - agree that some refactoring is in order

PR #924, opened 2021-04-05
Optional bracketing

### LINKED ISSUE #650: Are we optionally bracketing where we should be, and should we add an OptionallyBracketed segment?
From https://github.com/sqlfluff/sqlfluff/pull/648#discussion_r549768093:

> This should be a pattern which exists elsewhere, but I couldn't find it, so just went with a OneOf with one bracketed and one not bracketed. We could make an 'OptionallyBracketed' segment for this purpose to be used in group by, partition by and other clauses which I assume are also currently unparsable with brackets. Have I missed something here @alanmcruickshank ?

alanmcruickshank:
Last occurrence. Resolves #650


barrywhart (approved):
:shipit:

PR #925, opened 2021-04-05
Issue 911: [BigQuery] Using functions from other projects, quoted function names

### LINKED ISSUE #911: [BigQuery] Using functions from other projects
Hi there!

I've tested these 2 different ways of calling _remote_ functions and both worked inside the BigQuery editor.

```sql
SELECT
    `another-gcp-project`.functions.timestamp_parsing(log_tbl.first_move) AS first_move,
    `another-gcp-project.functions.timestamp_parsing`(log_tbl.orderdate) AS orderdate
FROM
    `gcp-project.data.year_2021` AS log_tbl
;
```

## Expected Behaviour

No errors.

## Observed Behaviour

```
L:   3 | P:  27 | L026 | Reference
                       | '`another-gcp-project`.functions.timestamp_parsing'
                       | refers to table/view 'functions' not found in the FROM
                       | clause or found in parent subquery.
L:   3 | P:  54 |  PRS | Found unparsable section: '(log_tbl.first_move) AS
                       | first_move'
L:   4 | P:   5 | L026 | Reference
                       | '`another-gcp-project.functions.timestamp_parsing`'
                       | refers to table/view 'functions' not found in the FROM
                       | clause or found in parent subquery.
L:   4 | P:  54 |  PRS | Found unparsable section: '(log_tbl.orderdate) AS
                       | orderdate'
L:   6 | P:  37 | L031 | Avoid using aliases in join condition
```

## Version

```
sqlfluff, version 0.4.1
```

## Configuration
```
[sqlfluff]
verbose = 0
dialect = bigquery
templater = python
exclude_rules = L009,L013

[sqlfluff:rules]
max_line_length = 120

[sqlfluff:templater:python:context]
source_table_id=my_table
start_date=456
end_date=456
fixed_carrier_id=456
fixed_destination_country_id=456
fixed_origin_country_id=456
number_of_buckets=123
bucket_id=123
fixed_carrier_ids=('1','2','3')
fixed_destination_country_ids=('1','2','3')
fixed_origin_country_ids=('1','2','3')
```

barrywhart:
Resolves #911

This PR changes the parse tree for `function_name`, hence the large number of files changed (`.yml` files for the parse test files).


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -759,13 +759,17 @@ class FunctionSegment(BaseSegment):
     type = "function"
     match_grammar = Sequence(
         Sequence(
-            Sequence(
```
barrywhart, line 762: Rather than allowing 0 or 1 names in front of the function name, allow any number. (BigQuery only uses 2, so we could limit it to 2 if that's preferable.)

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -79,8 +79,8 @@
 # Unreserved Keywords
 bigquery_dialect.sets("unreserved_keywords").add("SYSTEM_TIME")
 bigquery_dialect.sets("unreserved_keywords").remove("FOR")
-bigquery_dialect.sets("unreserved_keywords").add("STRUCT")
```
barrywhart, line 82: I had to make this change to fix an existing test that was broken by this change:
```
test/dialects/dialects_test.py::test__dialect__base_file_parse[bigquery-select_struct.sql]
```

test/fixtures/parser/ansi/double_dot.yml
```diff
@@ -16,4 +16,4 @@ file:
               - identifier: my_database
               - dot: .
               - dot: .
-              - identifier: my_table
\ No newline at end of file
+              - identifier: my_table
```
barrywhart, line 19: Why'd this change? I don't know.

test/fixtures/parser/ansi/transactions.yml
```diff
@@ -1,31 +1,31 @@
-file:
```
barrywhart, line 1: Why'd this change? I don't know.

test/fixtures/parser/mysql/force_index.yml
```diff
@@ -1,50 +1,50 @@
-file:
```
barrywhart, line 1: Why'd this change? I don't know.

test/fixtures/parser/mysql/use_index.yml
```diff
@@ -1,30 +1,30 @@
-file:
```
barrywhart, line 1: Why'd this change? I don't know.

barrywhart:
I'm putting this PR in "Draft" status because of the likely impact on lint rules that look for `function_name` in the parse tree.

src/sqlfluff/core/rules/std/L010.py
```diff
@@ -169,12 +169,13 @@ def _eval(self, segment, memory, parent_stack, **kwargs):
...
-                        "edit",
-                        segment,
-                        segment.__class__(raw=fixed_raw, pos_marker=segment.pos_marker),
-                    )
-                ],
+                fixes=[self._get_fix(segment, fixed_raw)],
                 memory=memory,
             )
+
+    def _get_fix(self, segment, fixed_raw):
```
barrywhart, line 176: Introducing this new overridable function allows rule L030, which handles function names, to continue working even though a parsed `function_name` now includes an additional level.

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -746,6 +748,15 @@ class WindowSpecificationSegment(BaseSegment):
     )
 
 
+@ansi_dialect.segment()
+class FunctionNameSegment(BaseSegment):
```
barrywhart, line 752: This new segment essentially replaces the old regex-based `FunctionNameSegment` so that rules looking for `function_name` in the parse tree are guaranteed to find it regardless of quoting.

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -251,8 +251,10 @@
     ParameterNameSegment=ReSegment.make(
         r"[A-Z][A-Z0-9_]*", name="parameter", type="parameter"
     ),
-    FunctionNameSegment=ReSegment.make(
```
barrywhart, line 254: Renaming this now that `FunctionNameSegment` is a class.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -175,11 +175,11 @@ class SelectClauseElementSegment(
         ),
         Sequence("WITH", "OFFSET", "AS", Ref("SingleIdentifierGrammar"), optional=True),
     ),
-    FunctionNameSegment=ReSegment.make(
```
barrywhart, line 178: Renaming this to stay aligned with `dialect_ansi.py`.

src/sqlfluff/core/rules/std/L010.py
alanmcruickshank: Can you add a note to that effect in the docstring for `_get_fix()`. Logic makes sense, but I assume future versions of ourselves will forget.

src/sqlfluff/core/rules/std/L030.py
```diff
@@ -38,4 +39,12 @@ class Rule_L030(Rule_L010):
 
     """
 
-    _target_elems: List[Tuple[str, str]] = [("name", "function_name")]
+    _target_elems: List[Tuple[str, str]] = [("type", "function_name")]
+
+    def _get_fix(self, segment, fixed_raw):
```
alanmcruickshank, line 44: Likewise I think this function needs some explanation in the docstring. It's not obvious why we need to to this from the code.

src/sqlfluff/dialects/dialect_ansi.py
alanmcruickshank: I think we should probably have `allow_gaps=False` no? Unless there can be spaces between the elements. Otherwise I wonder whether this could behave strangely.

test/fixtures/parser/ansi/transactions.yml
alanmcruickshank: Probably CRLF > LF

test/fixtures/parser/bigquery/select_udf_quote_nothing.yml
```diff
@@ -0,0 +1,33 @@
...
+      select_clause:
+        keyword: SELECT
+        select_clause_element:
+          function:
+          - identifier: gcpproject
+          - dot: .
+          - identifier: functions
+          - dot: .
+          - function_name:
+              function_name_identifier: timestamp_parsing
```
alanmcruickshank, line 13: Hmmmmm, I wonder whether the `identifiers` should be _part of_ `function_name`, i.e. that it should parse as:

```
function:
- function_name:
  - identifier: gcpproject
  - dot: .
  - idenfifier: functions
  - dot: .
  - function_name_identifier: timestamp_parsing
- start_bracket: (
...
```

Otherwise it feels wierd to nest `function_name_identifier` within `function_name` if there's never anything in there with it.

test/fixtures/parser/mysql/force_index.yml
alanmcruickshank: Probably CRLF > LF

test/fixtures/parser/mysql/use_index.yml
alanmcruickshank: Probably CRLF > LF

alanmcruickshank (changes requested):
Other than some documentation, there's only one thing I don't understand here...

I think either we should nest the other identifiers within `function_name` (as per my more detailed comment) _or_ we don't need to have a nested `function_name_identifier` and `function_name`. I don't think it makes sense to nest it and then never put anything in there with it. Or have I missed some thinking on this?

src/sqlfluff/core/rules/std/L010.py
barrywhart: Will do. I'm surprised the linter didn't flag this.

src/sqlfluff/dialects/dialect_ansi.py
barrywhart: I wasn't familiar with that property, but yes, that makes sense -- I assume it disallows white space? I'll add it.

test/fixtures/parser/bigquery/select_udf_quote_nothing.yml
barrywhart: I'll look into this after I make the simpler changes. This PR had involved some "trial and error" and did kind of end up in a weird place. The main challenge, as I recall, was to allow hyphens in function names at the parse level, since it apparently couldn't be done at the lexer level.

barrywhart: Never mind about the hyphens -- I'm confusing this with an earlier PR for project names in the `FROM` clause, which *did* allow non-quoted hyphens. That one was a little odd. I think a more straightforward approach should work here.

barrywhart:
@alanmcruickshank: I made the change you suggested re: `function_name` and `function_name_identifier`. I think I hadn't done this initially because I was worried about possible bad interactions with `CreateFunctionStatementSegment`, which also uses `FunctionNameSegment`, but after a second look, it should be fine to change how function names work there as well.

This change caused me to notice one remaining oddity about function names -- BigQuery allows quoting one or more dotted function name parts together, e.g.:
```
SELECT
    `another-gcp-project`.`functions.timestamp_parsing` (log_tbl.first_move) AS first_move
FROM
    `gcp-project.data.year_2021` AS log_tbl
```

In cases like this, the parse tree respects the quoting, so the quoted `.` becomes part of the identifier rather than a parse tree element. I think this is okay but may someday require helper functions in order for lint rules to properly interpret function names, since the same **logical** name can be represented in various ways.
```
[11](1, 2, 5)       |                    function:
[11](1, 2, 5)       |                        function_name:
[11](1, 2, 5)       |                            identifier:                       '`another-gcp-project`'
[32](1, 2, 26)      |                            dot:                              '.'
[33](1, 2, 27)      |                            identifier:                       '`functions.timestamp_parsing`'
```

barrywhart:
@alanmcruickshank: I made your suggested change `FunctionNameSegment`, so this is ready for another review. Thanks!

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -750,6 +752,28 @@ class WindowSpecificationSegment(BaseSegment):
     )
 
 
+@ansi_dialect.segment()
+class FunctionNameSegment(BaseSegment):
+    """Base part of a function name (excluding project, schema, etc.)."""
```
alanmcruickshank, line 757: Small nit - I think this docstring is out of date now. `FunctionNameSegment` explicitly allows the option to **include** project, schema etc, and this says it excludes it.

alanmcruickshank (changes requested):
One small nit on the docstring for FunctionNameSegment. Otherwise good to go! 👍

src/sqlfluff/dialects/dialect_ansi.py
barrywhart: Ah yes -- I'll fix that now.

barrywhart:
@alanmcruickshank: I fixed the docstring.

NiallRees (approved):

barrywhart:
@alanmcruickshank: Can you take another look at this PR? Niall approved it, but you had requested a change (which I have now made). I need your approval to merge. 🙏🏽

NiallRees (approved):

NiallRees:
I hadn't realised I'd already approved this! 🙈

barrywhart:
@NiallRees: No worries! Yes, I can't merge it until Alan approves it, since earlier he requested changes.

alanmcruickshank (approved):
👍 Love it

PR #928, opened 2021-04-06
Set statement parsing

### LINKED ISSUE #852: Grammar issue: If a query uses UNION, the ORDER BY should apply to the whole result set, not the last query
`
SELECT City FROM Customers
UNION
SELECT City FROM Suppliers
ORDER BY City
`

## Expected Behaviour
*orderby_clause* should be in the right under *set_expression*.

## Observed Behaviour
*orderby_clause* is two level under *set_expression*, under the last *select_statement*

the output should be the same as following query as mentioned here https://dev.mysql.com/doc/refman/8.0/en/union.html

`
(SELECT City FROM Customers)
UNION
(SELECT City FROM Suppliers)
ORDER BY City
`

## Steps to Reproduce
```python
query = """
SELECT City FROM Customers
UNION
SELECT City FROM Suppliers
ORDER BY City
"""
r = Linter(dialect='mysql').parse_string(query)
r.tree.to_tuple()
```

## Version
Python: 3.9.1
sqlfluff: 0.4.1

alanmcruickshank:
This resolves #852 . This works by altering the `SelectableGrammar` and distinguishing between _ordered_ and _unordered_ select statements. For ease, they both have the same name, but only the former may contain `ORDER` or `LIMIT` clauses.

In solving this I've solved two other bugs:
1. In the `_look_ahead_match` heuristic, if there were multiple simple matches for a single matcher (e.g. multiple opened and closed brackets), then only one was being considered. This PR now iterates through all of them.
2. In `_bracket_sensitive_look_ahead_match` a critical error was thrown if we found an unmatched closed bracket. We could however get into this situation during terminator matching. In this case it's not a problem and is simply the signal that we should stop matching. Rather than raising an exception this now logs and returns no match. In the event that a query should parse because of a stray bracket, it should still be flagged, but now just as an unparsable end section.

### DIFF
--- a/src/sqlfluff/core/parser/helpers.py
+++ b/src/sqlfluff/core/parser/helpers.py
@@ -63,7 +63,22 @@ def trim_non_code_segments(
 
 
 def iter_indices(seq: List, val: Any) -> Iterator[int]:
-    """Iterate all indices in a list that val occurs at."""
+    """Iterate all indices in a list that val occurs at.
+
+    Args:
+        seq (list): A list to look for indices in.
+        val: What to look for.
+
+    Yields:
+        int: The index of val in seq.
+
+    Examples:
+        The function works like str.index() but iterates all
+        the results rather than returning the first.
+
+        >>> print([i for i in iter_indices([1, 0, 2, 3, 2], 2)])
+        [2, 4]
+    """
     idx = 0
     while val in seq[idx:]:
         idx = seq.index(val, idx) + 1
--- a/src/sqlfluff/core/parser/helpers.py
+++ b/src/sqlfluff/core/parser/helpers.py
@@ -1,6 +1,6 @@
 """Helpers for the parser module."""
 
-from typing import Tuple, List, Any, TYPE_CHECKING
+from typing import Tuple, List, Any, Iterator, TYPE_CHECKING
 
 from sqlfluff.core.string_helpers import curtail_string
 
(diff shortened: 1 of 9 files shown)


src/sqlfluff/core/parser/helpers.py
```diff
@@ -60,3 +60,11 @@ def trim_non_code_segments(
             post_idx -= 1
 
     return segments[:pre_idx], segments[pre_idx:post_idx], segments[post_idx:]
+
+
+def iter_indices(seq: List, val: Any) -> int:
+    """Iterate all indices in a list that val occurs at."""
```
barrywhart, line 66: Could you provide one or two examples of a call and results of this function?

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -999,6 +1001,8 @@ class SelectClauseSegment(BaseSegment):
         terminator=OneOf(
             "FROM",
             "LIMIT",
+            "WHERE",
```
barrywhart, line 1004: Is it significant that above, we see `ORDER`, then `WHERE`, but here, they are reversed?

barrywhart (approved):
It's probably a little over my head for a quick reading, but overall looks good. One suggestion for adding more detail to a docstring.

barrywhart:
@alanmcruickshank: Do you think it'd be worth adding a few more tests? I'm thinking:
* One or two **negative** parse tests, verifying that the two failing examples from #852 report a parse error
* Adding a "successful parse" test for the final example, where the failing query now parses successfully with parentheses added

src/sqlfluff/dialects/dialect_ansi.py
alanmcruickshank: Not significant, but I'll make them the same to avoid confusion.

PR #929, opened 2021-04-06
Refactor L006 & L039, introduce L048

### LINKED ISSUE #360: Lint rule for spaces surrounding apostrophes (')
Can we have a lint rule that checks for spaces surrounding apostrophes? Following Simon Holywell's SQL Style Guide:

> Although not exhaustive always include spaces:
> * ...
> * surrounding apostrophes (') where not within parentheses or with a trailing comma or semicolon.

reference: https://www.sqlstyle.guide/#white-space

alanmcruickshank:
This resolves #360 .

It also refactors L006. Since the introduction of L039 (excessive whitespace), L006 no longer needs to lint for _too much_ whitespace. This PR applies that fix and also make the logic more reusable so that it can be used for L048 (missing whitespace around quoted literal). The refactor also improves the error message to be more explanatory (e.g. "Missing whitespace before *").

In implementing these I noted that L039 was only flagging on the first issue found and not all occurrences. This PR also extends that rule to fix that issue.

NB: L006 and L048 check for whitespace before *and* after (as before) but the anchor point for both is now the same. This means in comparing sets of check_tuples it appears as though the same error has appeared at the same position and therefore only once. This is just an effect of comparing `set` objects. In real use, the effect is still the same except that the position of the second linting issue has been moved back by one. This feels like a more appropriate positioning for the error.

### DIFF
--- a/src/sqlfluff/core/rules/std/L006.py
+++ b/src/sqlfluff/core/rules/std/L006.py
@@ -47,7 +47,6 @@ class Rule_L006(BaseRule):
 
         NOTE: We also allow bracket characters either side.
         """
-
         # Iterate through children of this segment looking for any of the
         # target types. We also check for whether any of the children start
         # or end with the targets.
@@ -116,11 +115,23 @@ class Rule_L006(BaseRule):
                         break
 
                 if (
+                    # There is a previous segment
                     prev_seg
+                    # And it's not whitespace
                     and not prev_seg.is_whitespace
+                    # And it's not an opening bracket
                     and not (
                         prev_seg.name.endswith("_bracket")
                         and prev_seg.name.startswith("start_")
+                    )
+                    # And it's not a placeholder that itself ends with whitespace.
+                    # NOTE: This feels convoluted but handles the case of '-%}' modifiers.
+                    and not (
+                        prev_seg.is_meta
+                        and (
+                            prev_seg.source_str.endswith(" ")
+                            or prev_seg.source_str.endswith("\n")
+                        )
                     )
                 ):
                     self.logger.debug(
@@ -159,11 +170,23 @@ class Rule_L006(BaseRule):
                         break
 
                 if (
+                    # There is a next segment
                     next_seg
+                    # It's not whitespace
                     and not next_seg.is_whitespace
+                    # It's not a closeing bracket
                     and not (
                         next_seg.name.endswith("_bracket")
                         and next_seg.name.startswith("end_")
+                    )
+                    # And it's not a placeholder that itself starts with whitespace.
+                    # NOTE: This feels convoluted but handles the case of '{%-' modifiers.
+                    and not (
+                        next_seg.is_meta
+                        and (
+                            next_seg.source_str.startswith(" ")
+                            or next_seg.source_str.startswith("\n")
+                        )
                     )
                 ):
                     self.logger.debug(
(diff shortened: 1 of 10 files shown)


src/sqlfluff/core/rules/std/L006.py
```diff
@@ -30,113 +31,187 @@ class Rule_L006(BaseRule):
...
+                                        raw=" ", pos_marker=sub_seg.pos_marker
+                                    ),
+                                )
+                            ],
                         )
-                        if anchor:
-                            description = "Operators should be followed by a space."
+                    )
+
+            if check_after:
```
barrywhart, line 161: Nitpick: The `after` section of this code looks like a mirror image of the `_before` section. Would it be worthwhile to try and have one version of this code that can be run in both directions (forward and backward)? (I'm torn on this myself.)

test/fixtures/rules/std_rule_cases/L048.yml
```diff
@@ -0,0 +1,17 @@
+rule: L048
+
+test_pass_1:
+  # Test that we don't fail * operators in brackets
```
barrywhart, line 4: I don't understand the comment here -- it doesn't seem to relate to the test query.

test/fixtures/rules/std_rule_cases/L048.yml
```diff
@@ -0,0 +1,17 @@
...
+  pass_str: "SELECT 'foo'"
+
+test_pass_2:
```
barrywhart, line 7: Isn't this a duplicate of `test_pass_1`?

src/sqlfluff/core/rules/std/L006.py
```diff
@@ -1,6 +1,8 @@
 """Implementation of Rule L006."""
 
 
+from typing import Tuple, List
```
barrywhart, line 4: I don't see any corresponding changes to the test file, `L006.yml`. Should there be new or updated tests to go with this?

src/sqlfluff/core/rules/std/L039.py
```diff
@@ -32,6 +32,7 @@ def _eval(self, segment, parent_stack, **kwargs):
         # For the given segment, lint whitespace directly within it.
         prev_newline = True
         prev_whitespace = None
+        violations = []
```
barrywhart, line 35: I don't see any corresponding changes to the test file, `L039.yml`. Should there be new or updated tests to go with this?

test/fixtures/rules/std_rule_cases/L048.yml
alanmcruickshank: oops yeah, that's a remnant from L006

src/sqlfluff/core/rules/std/L006.py
alanmcruickshank: All the existing tests are still totally relevant. I take your point on maybe adding some extra tests though.

src/sqlfluff/core/rules/std/L039.py
alanmcruickshank: Nope. Existing tests are fine. The new changes will just make it faster.

src/sqlfluff/core/rules/std/L006.py
alanmcruickshank: I don't know if we can have the same code, but I've separated out two slightly more generic methods to reuse a chunk of the logic. That should at least minimize issues from repeated code. What do you think of it now?

Issue #842, opened 2021-03-04
Parse error with regex (dbt)


daniel-murray:
-- If this is a parsing or linting issue, please include a minimal SQL example which reproduces the issue, along with the `sqlfluff parse` output, `sqlfluff lint` output and `sqlfluff fix` output when relevant. Finally, please add any labels that are related to the issue before submitting so that it can be addressed easily! --

I'm getting an unexpected parsing error.

Here's a snippet of the code. It's a safe case from string to float.
```sql
{%- macro number_to_float(column_name) -%}
case
    when {{ column_name }} ~ '^([0-9]){1,}(\.)([0-9]{1,})$' then {{ column_name }}::float
    when {{ column_name }} ~ '^[0-9]{1,}$' then {{ column_name }}::float
    else null
end
{%- endmacro -%}
```

## Expected Behaviour
No passing error.

## Observed Behaviour
```
L:  18 | P:   9 |  PRS | Found unparsable section: "case\n when value ~
                       | '^([0-9]){1,}(\\.)(..."
```

## Steps to Reproduce
Add this macro to a column in any dbt model using `{{ number_to_float("value") }}`, or the raw sql (error is not limited to it's use in a macro.)

## Version
sqlfluff, version 0.4.1
Python 3.7.5

## Configuration
```
[sqlfluff]
dialect=postgres
templater=dbt
exclude_rules = L009

[sqlfluff:indentation]
template_blocks_indent = False
```

barrywhart:
Could this be related to #828?

alanmcruickshank:
I thought this was the regex strings, but I think it's just that we don't have the `~` operator set up properly. PR is live. Thanks for the great bug report @daniel-murray  !

alanmcruickshank:
This was fixed with #932

PR #932, opened 2021-04-07
Like operators

### LINKED ISSUE #828: Lex error, unable to parse regex

## Expected Behaviour
no error

## Observed Behaviour
```
WHEN s.store_name LIKE 'Spec\'s%' THEN 'Spec\'s'
```
and 
```
AND si.catalog_item_id !~ '[a-z]'
```
this is the error message
```
== [dbt/models/emarsys/emarsys_price_change_program.sql] FAIL
L:  64 | P:  25 |  LXR | Unable to lex characters: '"!~ '[a-z]'"...'
== [dbt/models/reporting/precompiled/pc_store_features.sql] FAIL
L:  88 | P:  49 |  LXR | Unable to lex characters: '"\\'s'\n W"...'
```
## Steps to Reproduce
sqlfuff lint

## Version
sqlfluff, version 0.4.1
Python 2.7.16

## Configuration
```
[sqlfluff]
verbose = 0
nocolor = False
dialect = snowflake
templater = jinja
rules = L010,L012,L023,L030
exclude_rules = None
recurse = 0
output_line_length = 80
ignore = parsing,templating

[sqlfluff:indentation]
indented_joins = False

[sqlfluff:templater:jinja]
apply_dbt_builtins = True
load_macros_from_path=dbt/macros/parsing

[sqlfluff:rules]
tab_space_size = 2
max_line_length = 80
indent_unit = space
comma_style = trailing
allow_scalar = True
only_aliases = True

#inconsistent capitalization of keywords
[sqlfluff:rules:L010]
capitalisation_policy = upper

#unquoted identifiers
[sqlfluff:rules:L014]

#uppercase functions
[sqlfluff:rules:L030]
capitalisation_policy = upper
```

alanmcruickshank:
This resolves #828 and #842.

I thought the issue was with regex strings, but it's actually just a lack of the `LIKE` style operators. This adds a relatively fully featured set based on on Postgres. I dithered a bit about whether this should go in the `ansi` dialect, or just in `postgres` but for now I don't think it's doing any harm here.

In mysql this might conflict with the bitwise negation operator `~`, but we don't seem to have many mysql users so far.

### DIFF
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -88,6 +88,7 @@ ansi_dialect.set_lexer_struct(
             dict(is_code=True),
         ),
         ("not_equal", "regex", r"!=|<>", dict(is_code=True)),
+        ("like_operator", "regex", r"!?~~?\*?", dict(is_code=True)),
         ("greater_than_or_equal", "regex", r">=", dict(is_code=True)),
         ("less_than_or_equal", "regex", r"<=", dict(is_code=True)),
         ("newline", "regex", r"\r\n|\n", dict(type="newline", is_whitespace=True)),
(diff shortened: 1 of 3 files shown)


barrywhart (approved):
:shipit: (The build/tests hasn't finished yet, but the code looks 💯 .)

PR #934, opened 2021-04-07
Fix bug for fixing comma placements with inline comment

### LINKED ISSUE #799: L019 (trailing vs leading commas) fix doesn't flow around inline comments

pwildenhain:
Fixes #799

### DIFF
--- a/test/fixtures/rules/std_rule_cases/L019.yml
+++ b/test/fixtures/rules/std_rule_cases/L019.yml
@@ -12,6 +12,24 @@ leading_comma_violations:
       b
     FROM c
 
+leading_comma_violation_with_inline_comment:
+  fail_str: |
+    SELECT
+        a
+        , b -- inline comment
+        , c
+        /* non inline comment */
+        , d
+    FROM e
+  fix_str: |
+    SELECT
+        a,
+        b ,-- inline comment
+        c,
+        /* non inline comment */
+        d
+    FROM e
+
 leading_commas_allowed:
   pass_str: |
     SELECT
--- a/src/sqlfluff/core/rules/std/L019.py
+++ b/src/sqlfluff/core/rules/std/L019.py
@@ -53,6 +53,17 @@ class Rule_L019(BaseRule):
     config_keywords = ["comma_style"]
 
     @staticmethod
+    def _last_comment_seg(raw_stack):
+        """Trace the raw stack back to the most recent comment segment.
+
+        A return value of `None` indicates no code segments preceding the current position.
+        """
+        for segment in raw_stack[::-1]:
+            if segment.is_comment:
+                return segment
+        return None
+
+    @staticmethod
     def _last_code_seg(raw_stack):
         """Trace the raw stack back to the most recent code segment.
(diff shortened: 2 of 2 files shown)


test/fixtures/rules/std_rule_cases/L019.yml
```diff
@@ -12,6 +12,24 @@ leading_comma_violations:
...
+        a
+        , b -- inline comment
+        , c
+        /* non inline comment */
+        , d
+    FROM e
+  fix_str: |
+    SELECT
+        a,
+        b ,-- inline comment
```
barrywhart, line 27: The formatting looks odd here -- space before the comma, no space between the comma and the comment. Will the final output look like this, or do other lint rules/fixes typically handle this?

pwildenhain: Yeah, I said more about this in 3bdf0583fe78cedec1b5623ab562a43542231c36 but basically I had a hard time figuring out the formatting. The final output looks normal though, as other rules get applied. If you have some suggestions though I'm all ears 👂

barrywhart: No suggestions, really, although I'm sure it's doable with more tinkering. We were discussing a similar "rule interdependencies" question on my open PR #905. I think I'm okay having other rules help out with whitespace formatting, but I'm not sure Niall and Alan are convinced. 😜

pwildenhain: Thanks for the heads up, let's see what they have to say 🤔

https://github.com/sqlfluff/sqlfluff/pull/905#discussion_r609607938

alanmcruickshank: Until we resolve the other conversation I'm fine with this. This isn't the only example of where there's some odd formatting around, so I don't feel the need to hold this back until that's resolved. 👍

alanmcruickshank (approved):
Aside from resolving he formatting issue (which we should come back to later), this looks good, 👍

PR #935, opened 2021-04-07
Adding Petal to in the wild

rbetzler:
Saw a few ppl doing this, so I thought your friends at Petal ought to join in the fun 🎉


alanmcruickshank (approved):
Great to have you on board!

PR #938, opened 2021-04-08
Issue 930: L044 -- Max Recursion Depth

### LINKED ISSUE #930: L044: Max Recursion Depth
Came across this `CRITICAL` error when linting a few models in our dbt project using sqlfluff `0.5.0`:

#### Steps to reproduce:
```
$ sqlfluff lint models/sales/hubspot_engagement_calls.sql
CRITICAL   [L044] Applying rule L044 threw an Exception: maximum recursion depth exceeded while calling a Python object 
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/rules/base.py", line 290, in crawl
    res = self._eval(
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/rules/std/L044.py", line 138, in _eval
    return self._analyze_result_columns(select_info, dialect, queries)
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/rules/std/L044.py", line 106, in _analyze_result_columns
    self._analyze_result_columns(
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/rules/std/L044.py", line 106, in _analyze_result_columns
    self._analyze_result_columns(
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/rules/std/L044.py", line 106, in _analyze_result_columns
    self._analyze_result_columns(
  [Previous line repeated 976 more times]
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/rules/std/L044.py", line 90, in _analyze_result_columns
    for wildcard in select_info.get_wildcard_info():
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/rules/analysis/select_crawler.py", line 119, in get_wildcard_info
    WildcardInfo(
  File "<string>", line 1, in __new__
RecursionError: maximum recursion depth exceeded while calling a Python object
```

#### SQLFluff Verbose Output
[sqlfluff_verbose_out.txt](https://github.com/sqlfluff/sqlfluff/files/6268014/sqlfluff_verbose_out.txt)

#### Version

```sh
$ sqlfluff --version
sqlfluff, version 0.5.0
```

#### Configuration
```
[sqlfluff]
dialect = snowflake
templater = dbt
exclude_rules = L007,L009,L031,L033,L034

[sqlfluff:rules]
max_line_length = 120
comma_style = trailing

[sqlfluff:rules:L010]
capitalisation_policy = upper
```

#### SQL

```sql
{{
    config(
        materialized = 'table'
    )
}}

with
    hubspot__engagement_calls as (
        select * from {{ ref('hubspot__engagement_calls') }}
    )

select * from hubspot__engagement_calls
```

#### Compiled SQL

```sql

with
    hubspot__engagement_calls as (
        select * from ANALYTICS_DWALLACE.PUBLIC_intermediate.hubspot__engagement_calls
    )

select * from hubspot__engagement_calls
```

barrywhart:
Resolves #930 

The coverage report is failing, but only because I added code to detect when linter tests cause unexpected exceptions -- I think the PR is ready for review and potential merge, unless we feel strongly about adding tests for the test framework.

### DIFF
--- a/src/sqlfluff/core/rules/analysis/select_crawler.py
+++ b/src/sqlfluff/core/rules/analysis/select_crawler.py
@@ -120,6 +120,8 @@ class SelectCrawler:
                             seg,
                             [
                                 alias_info.ref_str
+                                if alias_info.aliased
+                                else alias_info.from_expression_element.raw
                                 for alias_info in self.select_info.table_aliases
                             ],
                         )
--- a/src/sqlfluff/core/rules/std/L044.py
+++ b/src/sqlfluff/core/rules/std/L044.py
@@ -104,7 +104,7 @@ class Rule_L044(BaseRule):
                             if wildcard_table in queries:
                                 # Wildcard refers to a CTE. Analyze it.
                                 self._analyze_result_columns(
-                                    queries[wildcard_table], dialect, queries
+                                    queries.pop(wildcard_table), dialect, queries
                                 )
                             else:
                                 # Not CTE, not table alias. Presumably an
--- a/src/sqlfluff/testing/rules.py
+++ b/src/sqlfluff/testing/rules.py
@@ -59,6 +59,9 @@ def assert_rule_fail_in_sql(code, sql, configs=None, line_numbers=None):
     linted = Linter(config=cfg).lint_string(sql, fix=True)
     lerrs = linted.get_violations()
     print("Errors Found: {0}".format(lerrs))
+    for e in lerrs:
+        if e.desc().startswith("Unexpected exception"):
+            pytest.fail(f"Linter failed with {e.desc()}")
     parse_errors = list(filter(lambda v: type(v) == SQLParseError, lerrs))
     if parse_errors:
         pytest.fail(f"Found the following parse errors in test case: {parse_errors}")
--- a/test/fixtures/rules/std_rule_cases/L044.yml
+++ b/test/fixtures/rules/std_rule_cases/L044.yml
@@ -292,9 +292,10 @@ test_31:
 
 test_32:
   # Issue 930: Infinite recursion if CTE queries itself.
-  fail_str: with
-    hubspot__engagement_calls as (
-      select * from hubspot__engagement_calls
-    )
+  fail_str: |
+    with
+      hubspot__engagement_calls as (
+        select * from hubspot__engagement_calls
+      )
 
-    select * from hubspot__engagement_calls
+      select * from hubspot__engagement_calls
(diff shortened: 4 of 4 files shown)


src/sqlfluff/testing/rules.py
```diff
@@ -59,6 +59,9 @@ def assert_rule_fail_in_sql(code, sql, configs=None, line_numbers=None):
     linted = Linter(config=cfg).lint_string(sql, fix=True)
     lerrs = linted.get_violations()
     print("Errors Found: {0}".format(lerrs))
+    for e in lerrs:
```
barrywhart, line 62: In reproducing the bug and testing the fix, I noticed that the tests were "eating" the infinite recursion error, i.e. if a test case had a `fail_str`, the infinite recursion was caught and treated as an expected lint error. With this change, any **unexpected** exceptions will automatically cause the test to fail. I think this is the behavior we want.

src/sqlfluff/testing/rules.py
```diff
@@ -59,6 +59,9 @@ def assert_rule_fail_in_sql(code, sql, configs=None, line_numbers=None):
...
+        if e.desc().startswith("Unexpected exception"):
+            pytest.fail(f"Linter failed with {e.desc()}")
```
barrywhart, line 64: This is the line causing the coverage check to complain -- if we were hitting this line, that'd mean we have a failing test. I think having this line uncovered is fine.

src/sqlfluff/core/rules/analysis/select_crawler.py
```diff
@@ -120,6 +120,8 @@ def get_wildcard_info(self) -> List[WildcardInfo]:
                             seg,
                             [
                                 alias_info.ref_str
+                                if alias_info.aliased
```
barrywhart, line 123: This fixes the real issue: The rule was confusing two things:
* A table whose base name was "blah"
* A CTE named "blah"

src/sqlfluff/core/rules/std/L044.py
```diff
@@ -104,7 +104,7 @@ def _analyze_result_columns(
                             if wildcard_table in queries:
                                 # Wildcard refers to a CTE. Analyze it.
                                 self._analyze_result_columns(
-                                    queries[wildcard_table], dialect, queries
```
barrywhart, line 107: This fixes another issue I discovered in tracking down the issue: A CTE that queries itself was also causing infinite recursion.

alanmcruickshank: NIce 👍 

Initially I didn't get this, but I think it's actually quite an elegant solution.

src/sqlfluff/testing/rules.py
alanmcruickshank: Good catch

alanmcruickshank (approved):
Nice.

PR #945, opened 2021-04-10
Issue 944: L045 false positive when CTE used in WHERE clause

### LINKED ISSUE #944: L045 false positive when CTE used in WHERE clause
Rule [`L045`](https://github.com/sqlfluff/sqlfluff/blob/master/src/sqlfluff/core/rules/std/L045.py) incorrectly raises an exception when a CTE is used in a correlated subquery in the `WHERE` clause. Here's an example of one such query that results in this false positive:

```sql
WITH max_date_cte AS (
    SELECT MAX(row_updated_date) AS max_date
    FROM warehouse.loaded_monthly
)

SELECT stuff
FROM warehouse.updated_weekly
WHERE row_updated_date <= (SELECT max_date FROM max_date_cte);
```

## Expected Behaviour

I expect running `sqlfluff lint` on the query above to return with exit code 0 (no linting failure).

## Observed Behaviour

Running `sqlfluff lint` on the query above results in a failure state, failing rule L045.

## Steps to Reproduce

To quickly reproduce the error in bash:

```bash
q=$'WITH mycte AS (\n    SELECT MAX(col1) AS maxdate\n    FROM loaded_monthly\n)\n\nSELECT stuff\nFROM loaded_weekly\nWHERE row_updated_date <= (SELECT maxdate FROM mycte);'
echo "$q" | sqlfluff lint - --rules L045
```

```
== [stdin] FAIL
L:   1 | P:   1 | L045 | Query defines a CTE (common-table expression) but does
                       | not use it
```

## Version

```bash
sqlfluff --version
# sqlfluff, version 0.5.1

python3 --version
# Python 3.6.8
```
## Configuration

None (default)

barrywhart:
Resolves #944

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Changed
+- Fix false positive when CTE used in WHERE clause ([#944](https://github.com/sqlfluff/sqlfluff/issues/944))
+
 ## [0.5.1] - 2021-04-09
 ### Changed
 - Parsing improvements around optional brackets.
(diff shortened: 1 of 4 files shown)


alanmcruickshank (approved):
👍

PR #947, opened 2021-04-10
Fix #943

### LINKED ISSUE #943: L048 applies to IN list
Rule [`L048`](https://github.com/sqlfluff/sqlfluff/blob/master/src/sqlfluff/core/rules/std/L048.py) results in a linting failure on the following query, which is (in my opinion) formatted correctly:

```sql
SELECT some_column
FROM some_table
WHERE some_other_column IN ('some literal', 'some other literal');
```

Here's a quick way to reproduce the error in bash<sup>[1](#footnote1)</sup>.

```bash
q=$'SELECT some_column\nFROM some_table\nWHERE some_other_column IN (\'some literal\', \'some other literal\');'
echo "$q" | sqlfluff lint - --rules L048
```

```
== [stdin] FAIL
L:   3 | P:  29 | L048 | Missing whitespace after e_literal'
```

## Expected behavior

I expect running `sqlfluff lint` on the query above to return with exit code 0 (no linting failure).

## Observed behavior

Running `sqlfluff lint` on the query above results in a failure state, failing rule L048.

## Version

```bash
$ sqlfluff --version
sqlfluff, version 0.5.1

$ python3 --version
Python 3.6.8
```

## Configuration

Default (no config files).

## Extra info

Here's a slightly more complex query that would make a good test case for rule L048:

```sql
SELECT
    col1,
    'string literal' AS new_column_literal,
    CASE WHEN col2 IN ('a', 'b') THEN 'Y' ELSE 'N' END AS new_column_case
FROM some_table
WHERE col2 IN ('a', 'b', 'c', 'd');
```
<a name="footnote1">1</a>: Please excuse the weird backslash escape sequences - [StackOverflow told me this was how to get a newline in a string](https://stackoverflow.com/a/3182519).

alanmcruickshank:
Hotfix for the new L048 to avoid flagging commas unnecessarily

### DIFF
--- a/src/sqlfluff/core/rules/std/L048.py
+++ b/src/sqlfluff/core/rules/std/L048.py
@@ -35,3 +35,15 @@ class Rule_L048(Rule_L006):
     _target_elems: List[Tuple[str, str]] = [
         ("name", "quoted_literal"),
     ]
+
+    @staticmethod
+    def _missing_whitespace(seg, before=True):
+        """Check whether we're missing whitespace given an adjoining segment.
+
+        This avoids flagging for commas after quoted strings.
+        https://github.com/sqlfluff/sqlfluff/issues/943
+        """
+        simple_res = Rule_L006._missing_whitespace(seg, before=before)
+        if not before and seg and seg.is_type("comma"):
+            return False
+        return simple_res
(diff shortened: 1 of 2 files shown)


barrywhart (approved):
:shipit:

PR #955, opened 2021-04-11
#889 Use linux style xdg on mac

### LINKED ISSUE #889: On macos: Respect XDG base dirs.
Apropos to #405, a large number of developer tools (especially those with a cli) respect XDG base dirs in both macos and linux. Those that insist on using `~/Library` on macos are outliers, as it creates unnecessary headaches for anyone that (a) maintains their own user configurations and / or (b) frequently switches between macos and linux. 

Can `sqlfluff` at least first look for configuration in `${XDG_CONFIG_HOME:-~/.config}` first on macos? Obviously `~/Library` can be used as a fallback if desired for some reason, though I think it becomes irrelevant for configuration / data if the XDG base dir spec is followed. (As indicated in https://stackoverflow.com/questions/3373948/equivalents-of-xdg-config-home-and-xdg-data-home-on-mac-os-x caching is a little more subtle).

dmateusp:
Resolves #889

### DIFF
--- a/src/sqlfluff/core/config.py
+++ b/src/sqlfluff/core/config.py
@@ -285,7 +285,7 @@ class ConfigLoader:
         else:
             p = os.path.dirname(path)
 
-        d = os.listdir(p)
+        d = os.listdir(os.path.expanduser(p))
         # iterate this way round to make sure things overwrite is the right direction
         for fname in filename_options:
             if fname in d:
@@ -299,14 +299,24 @@ class ConfigLoader:
         self._config_cache[str(path)] = configs
         return configs
 
-    def load_user_appdir_config(self) -> dict:
+    def load_user_appdir_config(self) -> Tuple[Optional[str], dict]:
         """Load the config from the user's OS specific appdir config directory."""
         appname = "sqlfluff"
         appauthor = "sqlfluff"
-        user_config_dir_path = appdirs.user_config_dir(appname, appauthor)
+
+        # On Mac OSX follow Linux XDG base dirs
+        # https://github.com/sqlfluff/sqlfluff/issues/889
+        user_config_dir_path = os.path.expanduser("~/.config/sqlfluff")
+        if appdirs.system == "darwin":
+            appdirs.system = "linux2"
+            user_config_dir_path = appdirs.user_config_dir(appname, appauthor)
+            appdirs.system = "darwin"
+        if not os.path.exists(user_config_dir_path):
+            user_config_dir_path = appdirs.user_config_dir(appname, appauthor)
+
         if os.path.exists(user_config_dir_path):
-            return self.load_config_at_path(user_config_dir_path)
-        return {}
+            return user_config_dir_path, self.load_config_at_path(user_config_dir_path)
+        return None, {}
 
     def load_user_config(self) -> dict:
         """Load the config from the user's home directory."""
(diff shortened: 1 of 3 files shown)


src/sqlfluff/core/config.py
```diff
@@ -260,14 +260,24 @@ def load_config_at_path(self, path: str) -> dict:
...
+    def load_user_appdir_config(self) -> Tuple[Optional[str], dict]:
         """Load the config from the user's OS specific appdir config directory."""
         appname = "sqlfluff"
         appauthor = "sqlfluff"
-        user_config_dir_path = appdirs.user_config_dir(appname, appauthor)
+
+        # On Mac OSX follow Linux XDG base dirs
+        # https://github.com/sqlfluff/sqlfluff/issues/889
+        user_config_dir_path = os.path.expanduser("~/.config/sqlfluff")
+        if appdirs.system == "darwin":
```
barrywhart, line 271: This looks a little tricky, switching the OS. Is there a known issue with `appdirs` requiring this technique?

src/sqlfluff/core/config.py
```diff
@@ -260,14 +260,24 @@ def load_config_at_path(self, path: str) -> dict:
...
+            appdirs.system = "linux2"
+            user_config_dir_path = appdirs.user_config_dir(appname, appauthor)
+            appdirs.system = "darwin"
+        if not os.path.exists(user_config_dir_path):
+            user_config_dir_path = appdirs.user_config_dir(appname, appauthor)
+
         if os.path.exists(user_config_dir_path):
-            return self.load_config_at_path(user_config_dir_path)
-        return {}
+            return user_config_dir_path, self.load_config_at_path(user_config_dir_path)
```
barrywhart, line 279: It looks like the first return value exists solely so the automated test can check it. I wonder if we could refactor (split) this into two functions:
* One that returns the directory
* Another that calls the first and returns the config path

I think that may be cleaner, so we're not returning unneeded data to the "real" code. 🤔

dmateusp: 100%, lazy on my end ^^

src/sqlfluff/core/config.py
dmateusp: the context is here: https://github.com/ActiveState/appdirs/issues/78#issuecomment-366947045

barrywhart (approved):
Looks good! I see the automated test. This seems like something we'd want to test manually as well. Have you done that?

dmateusp:
Thanks for the suggestion @barrywhart! So I tested it manually like this:

``` bash
❯ git status
On branch dmateusp/#889/use_linux_style_xdg_on_mac
❯ export XDG_CONFIG_HOME=~/.config/linters
❯ echo "select 1 FROM table_a" > test.sql
❯ echo "[sqlfluff:rules:L010]\ncapitalisation_policy = upper" > ~/Library/Application\ Support/sqlfluff/.sqlfluff
❯ echo "[sqlfluff:rules:L010]\ncapitalisation_policy = lower" > ${XDG_CONFIG_HOME}/sqlfluff/.sqlfluff
❯ sqlfluff fix test.sql
❯ cat test.sql
select 1 from table_a
❯ git checkout master
Switched to branch 'master'
❯ sqlfluff fix test.sql
❯ cat test.sql
SELECT 1 FROM table_a
```

So I think it's working as expected

dmateusp:
@barrywhart this made me realize that the error message in L010 isn't very helpful so I opened https://github.com/sqlfluff/sqlfluff/issues/1128

PR #960, opened 2021-04-13
add TSQL dialect

swanderz:
support for MSFT TSQL


src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -0,0 +1,37 @@
...
+)
+from sqlfluff.core.dialects import load_raw_dialect
+from sqlfluff.dialects.tsql_keywords import (
+    # BARE_FUNCTIONS,
+    RESERVED_KEYWORDS,
+    # UNRESERVED_KEYWORDS,
+)
+
+ansi_dialect = load_raw_dialect("ansi")
+tsql_dialect = ansi_dialect.copy_as("tsql")
```
swanderz, line 33: @barrywhart rookie mistake here I'm sure... but I thought is all that is needed to create a new dialect? When I try and parse the simple SQL statement below I'm getting the below exception that I don't understand ([Gist](https://gist.github.com/swanderz/fee18cc6caa43767e44640d68ead38a1) w/ .sqlfluff and full stacktrace). I thought that `.copy_as()` copies over a dialect and all associated grammars?

```
  File "/Users/anders.swanson/repos/sqlfluff/src/sqlfluff/core/dialects/base.py", line 233, in ref
    name, self.name
RuntimeError: Grammar refers to 'IntervalKeywordSegment' which was not found in the tsql dialect
```

```sql
SELECT * FROM database.dbo.table_a
```

barrywhart: Interesting! I haven't created a new dialect with this technique, so nothing occurs to me immediately. Just to gain understanding, could you try taking an existing dialect such as BigQuery and do some "trial and error", removing code selectively until it stops working?

I'll try take a look when I have time, but it might be a day or two.

@alanmcruickshank, do you have any thoughts?

barrywhart: I commented out these two lines, and now it works.

```
#tsql_dialect.sets("reserved_keywords").clear()
#tsql_dialect.sets("reserved_keywords").update(RESERVED_KEYWORDS)
```

I don't know offhand **why** this fixed it, but I hope this helps!

tunetheweb: I wonder if it's this "word" - cause it's actually two words?:
```
"WITHIN GROUP",
```

tunetheweb:
Superseded by #1313

PR #970, opened 2021-04-15
Trial pypy test suite

alanmcruickshank:
This is a testing PR to attempt testing and building of sqlfluff on pypy, with the intent of more officially supporting pypy as an option in performance work. Once testing is set up and tests pass, this PR should eventually also include sufficient documentation on how to run sqlfluff on pypy.


barrywhart:
Looks like CircleCI was down, causing the build to fail.

tox.ini
```diff
@@ -9,7 +9,14 @@ allowlist_externals =
...
     -Urrequirements.txt
-    -Urrequirements_dev.txt
+    # All test cases require pytest
+    pytest
+    pytest-cov
+    pytest-sugar
+    hypothesis
+    coverage
+    # If we're in pypy don't install all of these dependencies, otherwise grab the full set.
+    !pypy37: -Urrequirements_dev.txt
```
barrywhart, line 19: Do you know why it's necessary to exclude PyPy here?

alanmcruickshank: `mypy` doesn't support `pypy`. It's because the dependency `typed_ast` doesn't support `pypy`.

alanmcruickshank:
Based on automated testing taking ages, and pypy not performing that well I'm going to abandon this.

PR #979, opened 2021-04-17
Issue 975: Add rule code to diff_cover output

### LINKED ISSUE #975: Add rule identifier to the diff_cover output
We just started using sqlfluff in our organization, and our first streamlined usage is through the `diff_cover` plugin, which is used to write a comment on the PR:

image

Since we are just getting started, there will be a lot of tweaking of rules as we use it "in the wild". Something that would really help would be to have the rule identifier on the output. Is there any particular reason why this is not included?

For context, this is the PR on `sqlfluff` adding itself as a `diff_cover` plugin: https://github.com/sqlfluff/sqlfluff/pull/90.

(amazing to see that you folks were the ones adding a plugin system to `diff_cover` 🙌 )

barrywhart:
Resolves #975

### DIFF
--- a/src/sqlfluff/diff_quality_plugin.py
+++ b/src/sqlfluff/diff_quality_plugin.py
@@ -30,7 +30,7 @@ class SQLFluffViolationReporter(BaseViolationReporter):
         for violation in linted_path.get_violations():
             try:
                 # Normal SQLFluff warnings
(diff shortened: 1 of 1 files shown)


alanmcruickshank (approved):
👍

Issue #630, opened 2020-12-09
UnboundLocalError: local variable 'sub_section' referenced before assignment


NiallRees:
From slack:

```
Traceback (most recent call last):
  File "/usr/local/bin/sqlfluff", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/sqlfluff/cli/commands.py", line 323, in lint
    ignore_files=not disregard_sqlfluffignores,
  File "/usr/local/lib/python3.7/site-packages/sqlfluff/core/linter.py", line 1278, in lint_paths
    ignore_files=ignore_files,
  File "/usr/local/lib/python3.7/site-packages/sqlfluff/core/linter.py", line 1252, in lint_path
    target_file.read(), fname=fname, fix=fix, config=config
  File "/usr/local/lib/python3.7/site-packages/sqlfluff/core/linter.py", line 1043, in lint_string
    parsed = self.parse_string(in_str=in_str, fname=fname, config=config)
  File "/usr/local/lib/python3.7/site-packages/sqlfluff/core/linter.py", line 797, in parse_string
    in_str=in_str, fname=fname, config=config
  File "/usr/local/lib/python3.7/site-packages/sqlfluff/core/templaters/dbt.py", line 181, in process
    return self._unsafe_process(fname, in_str, config)
  File "/usr/local/lib/python3.7/site-packages/sqlfluff/core/templaters/dbt.py", line 227, in _unsafe_process
    raw_sliced, sliced_file = self.slice_file(node.raw_sql, node.injected_sql)
  File "/usr/local/lib/python3.7/site-packages/sqlfluff/core/templaters/python.py", line 284, in slice_file
    split_sliced, raw_occurrences, templated_occurances, templated_str
[...]

PR #982, opened 2021-04-17
Issue 968: Jinja templater throwing UnboundLocalError with "for" loop

### LINKED ISSUE #968: [BigQuery] Jinja for loop failing to parse
A simple Jinja loop with BigQuery fails to parse using sqlfluff 0.5.2, last working for this is version 0.3.6

## Expected Behaviour
No errors

## Observed Behaviour

```
[stijnzanders@stijn]$ sqlfluff parse test.sql
  File "/home/stijnzanders/.local/bin/sqlfluff", line 8, in <module>
    sys.exit(cli())
  File "/home/stijnzanders/.local/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/stijnzanders/.local/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/stijnzanders/.local/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/stijnzanders/.local/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/stijnzanders/.local/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/stijnzanders/.local/lib/python3.8/site-packages/sqlfluff/cli/commands.py", line 582, in parse
    for parsed_string in result:
  File "/home/stijnzanders/.local/lib/python3.8/site-packages/sqlfluff/core/linter.py", line 1469, in parse_path
    yield self.parse_string(
  File "/home/stijnzanders/.local/lib/python3.8/site-packages/sqlfluff/core/linter.py", line 874, in parse_string
    templated_file, templater_violations = self.templater.process(
  File "/home/stijnzanders/.local/lib/python3.8/site-packages/sqlfluff/core/templaters/jinja.py", line 299, in process
    raw_sliced, sliced_file, out_str = self.slice_file(
  File "/home/stijnzanders/.local/lib/python3.8/site-packages/sqlfluff/core/templaters/python.py", line 287, in slice_file
    sliced_file = list(
  File "/home/stijnzanders/.local/lib/python3.8/site-packages/sqlfluff/core/templaters/python.py", line 950, in _split_uniques_coalesce_rest
    "        Attempting Subsplit [post]: %s", sub_section
UnboundLocalError: local variable 'sub_section' referenced before assignment
```

## Steps to Reproduce

SQL File:
```
{% set event_columns = ['campaign', 'click_item'] %}

SELECT
    event_id
    {% for event_column in event_columns %}
    , {{ event_column }}
    {% endfor %}
FROM events
```

## Version
```
==== sqlfluff ====
sqlfluff:           0.5.2 python:             3.8.7
dialect:         bigquery verbosity:              2
templater:          jinja
```
## Configuration
```
[sqlfluff]
dialect = bigquery

[sqlfluff:indentation]
indented_joins = False

[sqlfluff:rules:L019]
comma_style = leading
```

barrywhart:
Resolves #968, #630


src/sqlfluff/core/templaters/python.py
```diff
@@ -947,7 +949,11 @@ def _split_uniques_coalesce_rest(
                     template_idx + raw_len,
                 )
 
-            if starts[1] < stops[1] and last_owu_idx is not None:
+            if (
+                starts[1] < stops[1]
+                and last_owu_idx is not None
+                and sub_section is not None
```
barrywhart, line 955: I am not familiar with the `_split_uniques_coalesce_rest()` function, but just as a simple experiment, I made a simple change to the code to avoid the crash, and it works correctly, at least for the test case provided with #968. I did not try to reproduce the issue with the test case in #630. I am optimistically assuming this fix addresses that issue as well. It should definitely avoid the crash, at least.

src/sqlfluff/core/templaters/python.py
```diff
@@ -837,6 +839,7 @@ def _split_uniques_coalesce_rest(
                     # If we succeeded in one of the above, we can also recurse
                     # and be more intelligent with the other sections.
                     if source_slice:
+                        assert sub_section is not None  # Makes MyPy happy
```
alanmcruickshank, line 842: should probably add

```
... and in this case we never set source_slice without also setting subsection.
```

alternatively we could change it to `if source_slice and subsection:`

src/sqlfluff/core/templaters/python.py
alanmcruickshank: Reviewing this code now, it seems risky to be referring to `sub_section` here _at all_, which means I wonder whether this is a patch fix for something which could be more effectively fixed. I'm going to try running this code locally now the test case is in to see what results we get locally and will report back...

barrywhart: Agreed.

src/sqlfluff/core/templaters/python.py
```diff
@@ -739,17 +739,17 @@ def _split_uniques_coalesce_rest(
...
                             IntermediateFileSlice(
                                 "compound",
                                 # Slicing is easy here, we have no choice
                                 slice(starts[0], int_file_slice.source_slice.stop),
                                 slice(starts[1], int_file_slice.templated_slice.stop),
-                                sub_section,
+                                # Calculate the subsection to deal with.
+                                int_file_slice.slice_buffer[
+                                    bookmark_idx : len(int_file_slice.slice_buffer)
+                                ],
```
alanmcruickshank, line 752: I've totally removed the `sub_section` variable here, it having the same name as the other patch is confusing.

src/sqlfluff/core/templaters/python.py
```diff
@@ -818,15 +818,15 @@ def _split_uniques_coalesce_rest(
...
+                    sub_section: Optional[List[RawFileSlice]] = None
                     # If it's the start, the slicing is easy
                     if starts[1] == int_file_slice.templated_slice.stop:
-                        source_slice = slice(starts[0], raw_idx)
                         sub_section = int_file_slice.slice_buffer[:this_owu_idx]
                     # If we are AFTER the previous in the template, then it's
                     # also easy. [assuming it's not the same owu]
                     elif raw_idx > starts[0] and last_owu_idx != this_owu_idx:
-                        source_slice = slice(starts[0], raw_idx)
                         if last_owu_idx:
```
alanmcruickshank, line 830: @barrywhart I think that this version of the logic is much cleaner. thoughts?

src/sqlfluff/core/templaters/python.py
```diff
@@ -949,9 +952,7 @@ def _split_uniques_coalesce_rest(
 
             if starts[1] < stops[1] and last_owu_idx is not None:
                 # Yield the end bit
-                templater_logger.debug(
-                    "        Attempting Subsplit [post]: %s", sub_section
-                )
+                templater_logger.debug("        Attempting Subsplit [post].")
```
alanmcruickshank, line 955: I think this was just lazy code on my part - `sub_section` wasn't used in anything other than logging, so I've just removed it. 🤷

src/sqlfluff/core/templaters/python.py
barrywhart: I agree -- if one variable is totally dependent on another, eliminating it or narrowing the scope within which it exists is 💯.

src/sqlfluff/core/templaters/python.py
barrywhart: Done -- I added your text to the comment. I think I like the assertion because makes the expected behavior (i.e. the relationship between the two variables) clear. `if source_slice and subsection` would give us the correct behavior but wouldn't document the assumption.

barrywhart:
@alanmcruickshank: The changes are great! I think this is ready to merge. I can't approve it since it's my own PR, but feel free to do so yourself. Thanks for the help!

PR #987, opened 2021-04-21
Add alter table x rename to y to ansi dialect

### LINKED ISSUE #986: Snowflake ALTER Table RENAME unsupported

## Expected Behaviour
sql lint should not break on .sql file containing valid Snowflake statements like:
```
ALTER TABLE old_table_name RENAME TO new_table_name;
```
## Observed Behaviour
Following error is thrown:
```
L:   1 | P:  35 | L014 | Inconsistent capitalisation of unquoted identifiers.
L:   1 | P:  37 |  PRS | Found unparsable section: ' new_table_name'
```
## Steps to Reproduce
* run sqlfluff lint <path/to/file>

## Version
```
sqlfluff, version 0.5.2
```
## Configuration
```
[sqlfluff]
verbose = 0
nocolor = False
dialect = snowflake
templater = jinja
#templater = dbt
rules = None
exclude_rules = L009
recurse = 0
output_line_length = 80
runaway_limit = 10
ignore_templated_areas = True
ignore = PRS
```

NiallRees:
Resolves #986


test/fixtures/parser/ansi/like_operators.yml
```diff
@@ -1,38 +1,38 @@
-file:
```
barrywhart, line 1: I keep seeing these large diffs. I assume it's because of platform-specific line endings. If so, think we should change the `test/generate_parse_fixture_yml.py` script to use consistent (probably Unix) line endings regardless of where it runs?

test/fixtures/parser/ansi/transactions.yml
```diff
@@ -1,31 +1,31 @@
-file:
```
barrywhart, line 1: These diffs are weird, too. Maybe different versions of the YAML package default to different indentation?

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2010,6 +2010,12 @@ class AlterTableStatementSegment(BaseSegment):
                         optional=True,
                     ),
                 ),
+                # Rename
```
barrywhart, line 2013: I think some existing dialects already have similar support. Should we look at removing/consolidating those here?

I didn't look closely, but I did a code search and saw this:
```
src/sqlfluff/dialects/dialect_teradata.py
187:    RENAME TABLE OLD_TABLENAME (TO|AS) NEW_TABLENAME
192:        "RENAME",

src/sqlfluff/dialects/dialect_exasol.py
1033:    """ALTER TABLE RENAME.."""
1037:        "RENAME",
1112:                "RENAME",
1182:# RENAME
1186:    """`RENAME` statement.
1197:    match_grammar = StartsWith("RENAME")
1199:        "RENAME",

src/sqlfluff/dialects/dialect_snowflake.py
614:            Sequence("RENAME", "TO", Ref("ObjectReferenceSegment")),
```

barrywhart (approved):
Maybe we want to consolidate other dialects' (e.g. Teradata) existing support for this syntax? Not a blocker, though -- we could create a separate issue for this. Also a couple of unrelated thoughts about the strange parser `.yml` diffs we keep seeing.

test/fixtures/parser/ansi/like_operators.yml
barrywhart: I went ahead and had a go at this. See #988.

src/sqlfluff/dialects/dialect_ansi.py
NiallRees: It looks like exasol doesn't support the `alert table x rename to y` syntax: https://docs.exasol.com/sql/alter_table(column).htm?tocpath=SQL%20Reference%7CSQL%20Statements%7CDefinition%20of%20Database%20(DDL)%7C_____7

I'm going to keep this separate for now. Potentially to be revisited though if the need arises 😁

PR #997, opened 2021-04-25
Issue 991: Fix bug with BigQuery comparison operator parsing

### LINKED ISSUE #991: Failing to parse combination of comparison operators (BigQuery)
## Expected Behaviour

No parsing errors for `SELECT 1 < 2, 2 > 1` with BigQuery dialect. Note that `SELECT 1 > 2, 2 < 1` works fine, so this seems to be an issue with the bracket/comparison operator distinction.

## Observed Behaviour

`Found unparsable section: ', 2 > 1'`

## Steps to Reproduce

http://sqlfluff-online.herokuapp.com/fluffed?sql=H4sIAG59gmAC_wt29XF1DlEwVLBRMNJRMFKwUzAEAPAOThMTAAAA&dialect=bigquery

## Version
0.5.2

## Configuration
```
dialect = bigquery
```

barrywhart:
Resolves #991 

This was the latest in a series of bugs where `<` and `>` were being misinterpreted as angle brackets. I made a design change which should reduce the likelihood of future bugs like this. Now BigQuery only looks for angle brackets in areas of the code where they are supported, basically in function definitions.

I wouldn't be surprised if we need to revisit this someday if we see situations where comparison operators and angle brackets can be commingled. If that happens, one possible idea is to update `_bracket_sensitive_look_ahead_match()` and the parts of the grammar that use it so that they support multiple passes, i.e. `_bracket_sensitive_look_ahead_match()` can return **all** the valid interpretations, and the caller (e.g. `Sequence`, `Delimited`) tries them until it finds one that works.

The coverage report is failing for an interesting reason, and I'm unsure what we ought to do. Now that the angle brackets have been separated from the other brackets, the "back up and retry" logic in `_bracket_sensitive_look_ahead_match()` is no longer being used. I'm hesitant to remove it, though, because it may prove useful again someday, e.g. in the scenario described above. I'm inclined to leave it in place, dormant for now. I could be persuaded to add a unit test for this behavior if y'all think it worthwhile.


src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -438,7 +438,13 @@ def _trim_elem(seg):
...
     @classmethod
     def _bracket_sensitive_look_ahead_match(
-        cls, segments, matchers, parse_context, start_bracket=None, end_bracket=None
+        cls,
+        segments,
+        matchers,
+        parse_context,
+        start_bracket=None,
+        end_bracket=None,
+        bracket_pairs_set="bracket_pairs",
```
barrywhart, line 447: Added this parameter, which allows the available bracket pairs to vary per segment

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -369,6 +369,14 @@
         "LIMIT", "GROUP", "ORDER", "HAVING", "QUALIFY", "WINDOW"
     ),
     PrimaryKeyGrammar=Sequence("PRIMARY", "KEY"),
+    # Odd syntax, but prevents eager parameters being confused for data types
+    FunctionParameterGrammar=OneOf(
```
barrywhart, line 373: Extracted this from the existing `FunctionParameter` grammar to minimize duplication in BigQuery dialect

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2407,20 +2415,13 @@ class CreateFunctionStatementSegment(BaseSegment):
 
 
 @ansi_dialect.segment()
-class FunctionParameterGrammar(BaseSegment):
+class FunctionParameterListGrammar(BaseSegment):
```
barrywhart, line 2418: Renamed this to distinguish function parameter **list** from a single function **parameter**

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -290,20 +294,47 @@ class DatatypeSegment(BaseSegment):
...
+class FunctionParameterListGrammar(BaseSegment):
+    """The parameters for a function ie. `(string, number)`."""
+
+    # Function parameter list. Note that the only difference from the ANSI
+    # grammar is that BigQuery provides overrides bracket_pairs_set.
+    match_grammar = Bracketed(
+        Delimited(
+            Ref("FunctionParameterGrammar"),
+            delimiter=Ref("CommaSegment"),
+            bracket_pairs_set="angle_bracket_pairs",
```
NiallRees, line 333: This is really neat.

NiallRees (approved):

NiallRees:
> is no longer being used

I'd suggest removing the code, but I'm pretty ruthless when it comes to deleting stuff 😄 we have a record of discussing its deletion and it can be retrieved if helpful.

barrywhart:
@NiallRees: I went ahead and looked at removing it. It's going to be a little tricky to remove, like removing one side if an `if/else` statement. This raises questions like should we remove the `if` entirely? Should we add an `assert` to verify the thing that was previously being checked with `if` is now **always** true. So I'm going to keep it for now at least, because I'm worried about the risk of either breaking it or making it extra hard to read or understand. 😬

PR #998, opened 2021-04-25
Issues 989, 999: Add support for 1) empty ARRAY literals 2) ARRAY literals with explicit type

### LINKED ISSUE #989: Failing to parse BigQuery array literals
## Expected Behaviour

No parsing errors for `SELECT [] AS foo` with BigQuery dialect. Should also support an explicit type like `SELECT ARRAY<string>[] AS foo`.

## Observed Behaviour

`Found unparsable section: '[] AS foo'`

## Steps to Reproduce

http://sqlfluff-online.herokuapp.com/fluffed?sql=H4sIAH95gmAC_wt29XF1DlGIjlVwDFZIy88HANgBUWYQAAAA&dialect=bigquery

## Version
0.5.2

## Configuration
```
dialect = bigquery
```

barrywhart:
Resolves #989, #999

**NOTE**: An early version of this PR addressed only the parse **errors** occurring with empty array literals, but the parse **result** for explicitly typed array literals was incorrect. I created a separate issue for that (#999), thinking it was lower priority and more complex, but it wasn't that complex, so I added it to this PR.


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -428,9 +428,16 @@ class ArrayLiteralSegment(BaseSegment):
...
     type = "array_literal_type"
-    match_grammar = Bracketed(
-        Delimited(Ref("ExpressionSegment")),
-        bracket_type="square",
+    match_grammar = OneOf(
+        Bracketed(
+            Delimited(Ref("ExpressionSegment")),
+            bracket_type="square",
+        ),
+        Sequence(
```
barrywhart, line 436: I was surprised to notice that `Bracketed` would not parse an empty list (`[]`). I added this to make it work, but I wonder if `Bracketed` **should** support this, either by default or as a new keyword option.

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -428,9 +428,16 @@ class ArrayLiteralSegment(BaseSegment):
     """An array literal segment."""
 
     type = "array_literal_type"
```
NiallRees, line 430: I realise this is separate to your PR, but shouldn't this just be `array_literal`?

src/sqlfluff/dialects/dialect_ansi.py
NiallRees: Does making `Delimited(Ref("ExpressionSegment")),` `Delimited(Ref("ExpressionSegment"), optional=True),` solve this?

src/sqlfluff/dialects/dialect_ansi.py
barrywhart: Yes, I think so. I'll change it unless something blocks me (unlikely).

src/sqlfluff/dialects/dialect_ansi.py
barrywhart: Ha, of course! That's a much better way to do it. I created a mental "trap" for myself, thinking this should be a feature of `Bracketed` instead. 🙄

barrywhart:
@NiallRees: I made the changes you suggested, if you want to review it again.

test/fixtures/parser/bigquery/select_mixture_of_array_literals.yml
```diff
@@ -0,0 +1,82 @@
...
+              end_square_bracket: ']'
+          alias_expression:
+            keyword: AS
+            identifier: strcol1
+      - comma: ','
+      - select_clause_element:
+          expression:
+          - keyword: ARRAY
+          - comparison_operator: <
+          - data_type_identifier: string
```
NiallRees, line 47: Looks to me as though the angle brackets are parsed as less-than and greater-than operators still?

barrywhart: IIRC, we chose not to add a new token type, thinking that was tidier. Do you think we should?

barrywhart: (In other words, this was a decision several months back when we first added support for angle brackets, not a recent decision.)

Let me know what you think -- I could see this current behavior causing some linter rules to get confused.

barrywhart: @NiallRees: I'll go ahead and change this, i.e. change it so the angle brackets parse output uses a new token. Definitely easier to understand the parse output this way.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -56,6 +57,12 @@
         "double_quote", name="quoted_literal", type="literal", trim_chars=('"',)
     ),
     StructKeywordSegment=KeywordSegment.make("struct", name="struct"),
+    StartAngleBracketSegment=SymbolSegment.make(
+        "<", name="start_angle_bracket", type="start_angle_bracket"
```
NiallRees, line 61: Referring to these as `open_angle_bracket` and `close_angle_bracket` is more common in what I've read but it's not a deal breaker!

NiallRees (approved):

PR #1003, opened 2021-04-26
Now all brackets are "definite"

barrywhart:
Followup to PR #997. @NiallRees had suggested removing the unused code for backtracking and retrying on "not definite" brackets. This PR removes that code, along with the declaration for whether brackets are definite or not.


barrywhart:
The failed coverage check is due to some "missing close bracket" error checking code not being hit. I'm pretty sure this code was already "not covered", but cleaning up the code is flagging this now.

NiallRees (approved):
I'm going to approve but let's wait to hear from @alanmcruickshank!

alanmcruickshank (approved):
Nice - 💯 for proactive clearup 👍

PR #1011, opened 2021-04-29
Issue 1004: Failing to parse nested structs (BigQuery)

### LINKED ISSUE #1004: Failing to parse nested structs (BigQuery)
## Expected Behaviour

No parsing errors for `SELECT STRUCT(STRUCT(1 AS b) AS a) AS foo` with BigQuery dialect. Using `STRUCT` inside of a function call also fails (e.g. `SELECT ARRAY_AGG(STRUCT(a AS a, b AS b)) FROM foo`).

## Observed Behaviour

`Found unparsable section: ' AS b'`

## Steps to Reproduce

http://sqlfluff-online.herokuapp.com/fluffed?sql=H4sIANTthmAC_wt29XF1DlEIDgkKdQ7RgFKGCo7BCkmaIDIRTKbl5_NyAQBx15-eKwAAAA%3D%3D&dialect=bigquery

## Version
0.5.2

## Configuration
```
dialect = bigquery
```

barrywhart:
Resolves #1004


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -379,6 +379,14 @@
     ),
     # This is a placeholder for other dialects.
     SimpleArrayTypeGrammar=Nothing(),
+    BaseExpressionElementGrammar=OneOf(
```
barrywhart, line 382: Extracted this from `SelectClauseElementSegment` because the BigQuery dialect wants to extend it.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -152,30 +160,6 @@ class SelectClauseSegment(ansi_dialect.get_segment("SelectClauseSegment")):  # t
     )
 
 
-@bigquery_dialect.segment(replace=True)
```
barrywhart, line 155: This is no longer needed because we now get the same effect by copying and appending to the new `BaseExpressionElementGrammar`.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -360,13 +344,7 @@ class TypelessStructSegment(BaseSegment):
             Delimited(
                 AnyNumberOf(
                     Sequence(
-                        OneOf(
```
barrywhart, line 363: This was about 80% the same but wasn't allowing `TypelessStructSegment`. It seemed better to reuse the existing grammar than maintain this separately when it seems to be converging.

test/fixtures/parser/bigquery/select_struct.yml
```diff
@@ -56,15 +56,15 @@ file:
           typeless_struct:
           - keyword: struct
           - start_bracket: (
-          - object_reference:
+          - column_reference:
```
barrywhart, line 59: This parsing change _seems_ like an improvement to me.

src/sqlfluff/dialects/dialect_ansi.py
alanmcruickshank: Sounds very reasonable 👍

alanmcruickshank (approved):
This looks sensible to me :shipit:

PR #1013, opened 2021-05-01
Issue 642: Add --parallel option

### LINKED ISSUE #642: --threads argument for sqlfluff lint
Hey sqfluff devs! :sheep: 

Are all files linted in isolation? I suspect they have no dependencies on each other. Seems ripe for concurrency?

The `dbt` bin supports a `--threads` argument for this. Could be a huge optimization tweak?

For context, we are using sqlfluff with dbt and we have hundreds of models.
I don't have an accurate runtime estimate for our entire project because of #641, but I suspect it's several hours right now.

barrywhart:
Resolves #642

Initial performance numbers:
```
sqlfluff lint --parallel 1 sql/  91.35s user 0.82s system 99% cpu 1:32.32 total
sqlfluff lint --parallel 4 sql/  95.34s user 1.05s system 268% cpu 35.891 total
```

2.5x faster when running with `--parallel 4` on a small-ish project. 🎉💯

Performance numbers after the big refactoring:
```
sqlfluff lint --parallel 1 sql/  106.60s user 0.85s system 99% cpu 1:47.65 total
sqlfluff lint --parallel 4 sql/  112.39s user 1.12s system 234% cpu 48.358 total
```

**Performance got noticeably slower in both the serial and parallel cases.** It would probably be good to do some performance tuning to see where the added time is being spent.


pwildenhain:
Cool stuff Barry! I think this is going the be a game changer

barrywhart:
I think the failing tests are caused by the changes to `raw.py` and/or `meta.py`, possibly stale state hanging around across tests.

barrywhart:
I need to add tests for the new parallel support. Otherwise ready for review.

barrywhart:
Still seeing some problems using `--parallel` on a real-life project. Feel free to review, but putting back in "Draft" status until that's addressed.

alanmcruickshank (commented):
I like this approach, I think the "parallel-ness" is happening in exactly the right place.

Things I like:
- Having new instances of the config and the linter for the different threads makes lots of sense.

Things I'm not sure about:
- Lots of pickling. In theory the only objects we need to pass around are violations (which should serialise simply), fixes (which should serialise simply, providing segments do) and segments. 

To solve the latter I wonder if we can do something simpler. It looks like some of the acrobatics here are to make sure that when we define e.g. `CommaSegment`, that when we rehydrate it, we all refer to the same `CommaSegment`, and I don't know if we *need* to. The only place we really utilise the segment inheritance (apart from syntactic sugar) is in `BaseSegment.is_type()`, and all that does is look at the names of the parent classes. I wonder whether we can serialise the whole parse tree to some kind of pure-python object (or even json?) which could be passed much more easily between threads?

If we could achieve something like that I think it removes a load of the complexity around the pickling. For bonus points, this would also open the door to non-python implementations of the parser 🤷  - which would be the eventual endpoint of lots of the performance work if we want to go to the next level.

barrywhart:
@alanmcruickshank: One question about the pickling -- the `sqlfluff fix` command is a bit more complex, because it runs in two phases:

1. **Generate** a lint report and the proposed fixes.
2. Ask the user to approve making the fixes.
3. Apply the fixes.

Unlike the "lint-only" case, here there's a need for the linting phase to return something other than just the text output -- i.e. the `LintFix` objects. Do you know whether this includes returning segment objects from the "lint" phase?

One clarification about the acrobatics -- it's not so much that the class object has to be the same, it's more that pickling will fail if the class doesn't already exist on the other side. So there's a general need to precreate things. It _may_ be okay to create them again (not sure). Part of the reason I started tracking creation was to make it easier to debug (so we'd know if something already existed or not.)

Let me know about the `LintFix` question above. Depending on that, I'll revisit this.

barrywhart:
@alanmcruickshank: Based on your suggestions, I wonder if we could rework the keyword and symbol segments to avoid creating these classes dynamically (in other words, instead of creating a new class for each one, we create instances of existing classes like `KeywordSegment` or `SymbolSegment`). If we can do that, then segment classes are only created by dialects, not by lint rules. I think that would eliminate around 50% of the PR's complexity.

alanmcruickshank:
> I wonder if we could rework the keyword and symbol segments

No need to do the work in two places. All we would need to do to achieve this is to rework the `RawSegment` class, and specifically the `RawSegment.make()` method. If we can rework that so that it doesn't return a new derived class, but instead it returns some object that generates instances of the parent class.

Currently:
```
NewKindOfSegmentClass = RawSegment.make(...)
segment = NewKindOfSegmentClass()
```

Proposal:
```
ThingWhichReturnsKindsOfSegmentsWhenCalled = RawSegment.make(...)
segment = ThingWhichReturnsKindsOfSegmentsWhenCalled()
```

We would probably need to rethink what are class attributes and what are instance attributes (notably a few things like `_template` and suchlike), but that could also be achieved with some `@property` elements.

Shouldn't be too painful.

Want to have a go?

barrywhart:
Sounds good! Sure, I'll give it a try. I'll start it in a separate branch off `master`, then integrate it with this code once it's working. Seems simpler that way.

barrywhart:
@alanmcruickshank: I marked this as "ready for review". I've tested it on one of our company projects, and I think it is stable enough to release. Tests are passing, coverage is pretty good. Happy to wait if you prefer, but I think it could be cool to release this and then incorporate your parser/lexer cleanup later. 🤷‍♂️

src/sqlfluff/core/config.py
```diff
@@ -547,3 +551,18 @@ def process_inline_config(self, config_line: str):
         config_path = [elem.strip() for elem in config_line.split(":")]
         # Set the value
         self.set_value(config_path[:-1], config_path[-1])
+
```
barrywhart, line 554: Can we remove this?

src/sqlfluff/core/dialects/__init__.py
```diff
@@ -29,10 +29,16 @@
 }
 
 
+def load_raw_dialect_and_module(label, base_module="sqlfluff.dialects"):
```
barrywhart, line 32: Can we remove this?

src/sqlfluff/core/dialects/base.py
```diff
@@ -43,6 +43,7 @@ def __init__(
         self._sets = sets or {}
         self.inherits_from = inherits_from
         self.root_segment_name = root_segment_name
+        self.module = None
```
barrywhart, line 46: And this?

barrywhart:
This is ready for review again. All the `inspect` and `module` ugliness is gone now. 🎉 @alanmcruickshank

src/sqlfluff/cli/commands.py
```diff
@@ -285,8 +285,17 @@ def dialects(**kwargs):
     is_flag=True,
     help=("Perform the operation regardless of .sqlfluffignore configurations"),
 )
+@click.option(
+    "--parallel",
+    type=int,
+    default=1,
+    help="If set to a value higher than 1, runs SQLFluff in parallel, "
```
NiallRees, line 292:
```suggestion
    help="If set to a value higher than 1, run SQLFluff in parallel, "
```

Just for consistency

src/sqlfluff/cli/commands.py
```diff
@@ -382,8 +392,15 @@ def do_fixes(lnt, result, formatter=None, **kwargs):
 @click.option(
     "--fixed-suffix", default=None, help="An optional suffix to add to fixed files."
 )
+@click.option(
+    "--parallel",
+    type=int,
+    default=1,
+    help="If set to a value higher than 1, runs SQLFluff in parallel, "
```
NiallRees, line 399:
```suggestion
    help="If set to a value higher than 1, run SQLFluff in parallel, "
```

src/sqlfluff/core/linter.py
```diff
@@ -46,6 +51,8 @@
 # Instantiate the linter logger
 linter_logger: logging.Logger = logging.getLogger("sqlfluff.linter")
 
+tblib.pickling_support.install()
+
```
NiallRees, line 55: Could you add a comment to explain this and why it's here?

src/sqlfluff/core/linter.py
```diff
@@ -789,6 +796,24 @@ def tree(self) -> Optional[BaseSegment]:
...
 
 
+class DelayedException(Exception):
+    """Multiprocessing process pool uses this to propagate exceptions."""
+
+    def __init__(self, ee):
+        self.ee = ee
+        __, __, self.tb = sys.exc_info()
+        self.fname = None
+        super(DelayedException, self).__init__(str(ee))
```
NiallRees, line 806:
```suggestion
        super().__init__(str(ee))
```

src/sqlfluff/core/linter.py
barrywhart: Will do!

barrywhart:
The code coverage check is complaining about existing uncovered code that I moved around, not new code. 😝

src/sqlfluff/core/linter.py
NiallRees: Does it make sense for this to go in __init__.py? I feel it does

barrywhart: I'll move it. I don't have a strong feeling about where, but I do think code like this feels a little weird anywhere. Gotta have it, though! Argh.

barrywhart:
@NiallRees: I moved the `tblib` stuff to the `__init__.py` file, as you suggested. Care to review again? Thanks!

NiallRees (approved):
Let's do this!

NiallRees:
> @NiallRees: I moved the `tblib` stuff to the `__init__.py` file, as you suggested. Care to review again? Thanks!

Sorry this took me a few goes. Looking really nice.

barrywhart:
@NiallRees: No worries about the feedback -- quality is important!! Thanks again.

Issue #897, opened 2021-03-26
Refactor Position Markers


alanmcruickshank:
Putting together #892 has reminded my what state the position marker code is in.

Position markers are the primitives which allow us to keep track of where in a file we are, and in the "enriched" version, also where in the source file we are.

This is one of the oldest parts of the codebase, and was also significantly disrupted when adding in the source mapping code in the last major release.

Currently a `FilePositionMarker` is *required* on any segment. It is created by default and there is a reasonable amount of diligence around making sure they are consistent. On a templated file we then enrich this with additional metadata to make an `EnrichedFilePositionMarker`, this is patchy however and the treatment and enrichment of particular segments has been patchy and changing. This means some bugs occur when we think we have one kind but actually have another.

At the very least this code needs some TLC. I also propose that:
- Any parsed tree has a reliable kind of position marker which is templating ready from the start (i.e. it assumes an untemplated file but presents a consistent interface).
- No code outside of the `markers` module should need to care what kind of marker we have or to need to validate what it's looking for. Methods should be consistently callable regardless of source.
- Position markers should contain references to positions in both raw and templated files, but *also* the raw code which generated it for the purposes of linting or referring to that code.
- Any logic for "realigning" positions after edits or when elements are inserted should be centralised into one method which is then reused everywhere else.

PR #1014, opened 2021-05-03
Refactor KeywordSegment and SymbolSegment to eliminate dynamic class creation

barrywhart:
Refactoring to support #1013

### DIFF
--- a/src/sqlfluff/core/parser/grammar/base.py
+++ b/src/sqlfluff/core/parser/grammar/base.py
@@ -21,6 +21,12 @@ from sqlfluff.core.parser.context import ParseContext
 MatchableType = Union[Matchable, Type[BaseSegment]]
 
 
+def fluff_issubclass(cls, class_or_tuple):
+    if hasattr(cls, "cls"):
+        cls = getattr(cls, "cls")
+    return issubclass(cls, class_or_tuple)
+
+
 class BracketInfo(NamedTuple):
     """BracketInfo tuple for keeping track of brackets during matching.
 
@@ -84,7 +90,7 @@ class BaseGrammar(Matchable):
 
         for instance, init_type, init_func in initialisers:
             if (instance and isinstance(elem, init_type)) or (
-                not instance and issubclass(elem, init_type)
+                not instance and fluff_issubclass(elem, init_type)
             ):
                 return init_func(elem)
         raise TypeError(
--- a/src/sqlfluff/core/parser/grammar/base.py
+++ b/src/sqlfluff/core/parser/grammar/base.py
@@ -21,6 +21,12 @@ from sqlfluff.core.parser.context import ParseContext
 MatchableType = Union[Matchable, Type[BaseSegment]]
 
 
+def fluff_type(self):
+    if hasattr(self, "_cls"):
+        return self._cls
+    return type(self)
+
+
 def fluff_issubclass(cls, class_or_tuple):
     if hasattr(cls, "cls"):
         cls = getattr(cls, "cls")
@@ -539,7 +545,7 @@ class BaseGrammar(Matchable):
                             # the innermost start bracket? E.g. ")" matches "(",
                             # "]" matches "[".
                             start_index = start_brackets.index(
-                                type(bracket_stack[-1].bracket)
+                                fluff_type(bracket_stack[-1].bracket)
                             )
                             end_index = end_brackets.index(matcher)
                             bracket_types_match = start_index == end_index
@@ -658,7 +664,7 @@ class BaseGrammar(Matchable):
 
         e.g. `OneOf(foo) == OneOf(foo, optional=True)`
         """
-        return type(self) is type(other) and self._elements == other._elements
+        return fluff_type(self) is fluff_type(other) and self._elements == other._elements
 
     def copy(
         self,
--- a/src/sqlfluff/core/parser/segments/keyword.py
+++ b/src/sqlfluff/core/parser/segments/keyword.py
@@ -52,13 +52,13 @@ class _ProtoKeywordSegment(RawSegment):
 
     @classmethod
     @match_wrapper(v_level=4)
-    def match(cls, segments, parse_context, factory=None):
+    def match(cls, segments, parse_context):
         """Compare input segments for a match, return a `MatchResult`.
 
         Note: For Keyword matching, we only consider the *first* element,
         because we assume that a keyword can only span one raw segment.
         """
-        cls_ = factory if factory else cls
+        cls_ = parse_context._factory if parse_context._factory else cls
         # If we've been passed the singular, make it a list
         if isinstance(segments, BaseSegment):
             segments = [segments]
--- a/src/sqlfluff/core/parser/segments/keyword.py
+++ b/src/sqlfluff/core/parser/segments/keyword.py
@@ -48,7 +48,8 @@ class _ProtoKeywordSegment(RawSegment):
         we return a tuple in case there is more than one option.
         """
         # NB: We go UPPER on make, so no need to convert here
-        return [cls._template]
+        cls_ = parse_context._factory if parse_context._factory else cls
+        return [cls_._template]
 
     @classmethod
     @match_wrapper(v_level=4)
--- a/src/sqlfluff/core/parser/segments/keyword.py
+++ b/src/sqlfluff/core/parser/segments/keyword.py
@@ -127,6 +127,7 @@ class ReSegment(_ProtoKeywordSegment):
         we assume that ._template is a r"" string, and is formatted
         for use directly as a regex. This only matches on a single segment.
         """
+        cls_ = parse_context._factory if parse_context._factory else cls
         # If we've been passed the singular, make it a list
         if isinstance(segments, BaseSegment):
             segments = [segments]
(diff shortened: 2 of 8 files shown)


barrywhart:
@alanmcruickshank: This is still very preliminary, but I have all the tests in `test/core/parser/` passing. My sense thus far is that if we go this way rather than the design in #1013, we're trading one kind of complexity for another:
* The design in #1013 makes some areas more static, i.e. materializing segment types early (e.g. at dialect or other module load time) rather than late (at runtime).
* This design involves redesigning some areas to support both actual segment classes or with "class-like" segment objects. There may be a way to encapsulate this better, but I want to get things mostly working before trying to clean it up.

It's a bit unfortunate that both of these make the code messier. I'm still undecided which is better. Either could potentially be made cleaner than their present state, making it difficult to decide without further exploration.

Any thoughts on whether this matches what you have in mind, and whether this is preferable?

test/core/parser/segments_common_test.py
```diff
@@ -18,7 +18,7 @@ def test__parser__core_keyword(raw_seg_list):
     # First make a keyword
     FooKeyword = KeywordSegment.make("foo")
     # Check it looks as expected
-    assert issubclass(FooKeyword, KeywordSegment)
+    assert issubclass(FooKeyword.cls, KeywordSegment)
```
alanmcruickshank, line 21: In this test case, let's rename `FooKeyword` to `foo_keyword_factory`. That makes it more explicit in what's happening in these tests (and below).

Should probably `assert isinstance(foo_keyword_factory, RawSegmentFactory)` too.

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -148,3 +151,51 @@ def get_end_pos_marker(self):
...
+
+class RawSegmentFactory:
+    optional = False
+    _anti_template = None
+
+    def __init__(self, cls, classname, _template, _name, **kwargs):
+        # if classname.startswith("semicolon"):
+        #     import pdb; pdb.set_trace()
+        self.cls = cls
+        self.__name__ = classname
```
alanmcruickshank, line 164: Do we need to play with `__name__`? This feels like overcomplication. If it's to replicate some of the features of `RawSegment` then I wonder if we should get rid of those features.

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -148,3 +151,51 @@ def get_end_pos_marker(self):
     def get_start_pos_marker(self):
         """Return the pos marker at the start of this segment."""
         return self.pos_marker
+
+
+class RawSegmentFactory:
+    optional = False
+    _anti_template = None
```
alanmcruickshank, line 158: Why are these two left as class attributes? I'm surprised that `optional` is here but `_template` is an instance attribute. In my head they're the same kind of property.

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -148,3 +151,51 @@ def get_end_pos_marker(self):
...
+        # if classname.startswith("semicolon"):
+        #     import pdb; pdb.set_trace()
+        self.cls = cls
+        self.__name__ = classname
+        self._template = _template
+        self._name = _name
+        self._kwargs = kwargs
+
+    def __repr__(self):
+        return f"<{self.cls.__name__}: {self.__name__} {self._template} {self._name} {self._kwargs}"
```
alanmcruickshank, line 170: I think the `repr` should be more explicit that this is a `RawSegmentFactory` rather than trying to disguise it as the other class. Unless I'm missing something?

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -148,3 +151,51 @@ def get_end_pos_marker(self):
...
+        result = self.cls(
+            raw,
+            pos_marker,
+            _classname=self.__name__,
+            _template=self._template,
+            _name=self._name,
+            _cls=self,
+            **self._kwargs
+        )
+        return result
```
alanmcruickshank, line 182: 👍  - this is exactly what I had in mind.

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -148,3 +151,51 @@ def get_end_pos_marker(self):
...
+
+    def match(self, segments, parse_context):
+        parse_context._factory = self
+        result = self.cls.match(segments, parse_context)
+        parse_context._factory = None
```
alanmcruickshank, line 187: Eventually, setting `_factory` might be better as a context manager. Then you can do something like:

```
with parse_context.set_factory(self):
    result = self.cls.match(segments, parse_context)
```

(ditto in `simple` below)

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -75,14 +79,7 @@ def make(cls, template, case_sensitive=False, name=None, **kwargs):
...
         classname = "{0}_{1}".format(name, cls.__name__)
-        # This is the magic, we generate a new class! SORCERY
-        newclass = type(
-            classname,
-            (cls,),
-            dict(_template=_template, _name=name, **kwargs),
-        )
-        # Now we return that class in the abstract. NOT INSTANTIATED
-        return newclass
+        return RawSegmentFactory(cls, classname, _template, name, **kwargs)
```
alanmcruickshank, line 82: Exactly!

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -75,14 +79,7 @@ def make(cls, template, case_sensitive=False, name=None, **kwargs):
         name = name or _template
         # Now lets make the classname (it indicates the mother class for clarity)
         classname = "{0}_{1}".format(name, cls.__name__)
```
alanmcruickshank, line 81: I'm wondering whether we *need* to make this classname. I'm not sure what the implications of ditching it would be, but what if all of our `KeywordSegment` classes just still had the name `KeywordSegment`? That would keep things simpler.

It could also cause a big mess, so it's a pain then feel free to forget it.

src/sqlfluff/core/parser/segments/keyword.py
```diff
@@ -125,6 +127,7 @@ def match(cls, segments, parse_context):
         we assume that ._template is a r"" string, and is formatted
         for use directly as a regex. This only matches on a single segment.
         """
+        cls_ = parse_context._factory if parse_context._factory else cls
```
alanmcruickshank, line 130: Why do we need to do this? I'm a little confused.

alanmcruickshank: If we do need this little idiom (i.e. `parse_context._factory if parse_context._factory else cls`), then I wonder if we should wrap it as a method because we use it quite a few times.

src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -21,6 +21,18 @@
...
+    if hasattr(self, "_cls"):
+        return self._cls
+    return type(self)
+
+
+def fluff_issubclass(cls, class_or_tuple):
+    if hasattr(cls, "cls"):
+        cls = getattr(cls, "cls")
+    return issubclass(cls, class_or_tuple)
+
```
alanmcruickshank, line 34: We should tie these into `BaseSegment.is_type()` so that the functionality is consistent. Either these become `staticmethod` on `BaseSegment`, or these methods move to a `helpers.py` file and then `.is_type()` calls them.

alanmcruickshank (commented):
I've got a few questions, and I thin we might be able to simplify this - but the direction looks good.

src/sqlfluff/core/parser/segments/raw.py
barrywhart: Eventually, perhaps not. When I work on things like this, I tend to take an exploratory approach and try to get various tests working, then go back and clean up the ugly bits. This is definitely ugly.

I was thinking last night, I think we're implicitly moving in a direction where we don't use `@classmethod` anymore, nor do we use `type()`, `isinstance()`, `issubclass()`, etc.

From my previous experience with parsers in C++, it's common to have parallel class hierarchies, basically descriptors and then parser objects which are "instances" (but not in the C++ sense) of those descriptors. Essentially, we build our own type system in the parser.

I wonder if we have to do 100% of this work now (or ever). I've had some success thus far by adding the `factory` property to `ParseContext`. We can see how far that takes us.

src/sqlfluff/core/parser/segments/raw.py
barrywhart: This is a shortcut sometimes used in Python -- attributes on a class can be accessed using `self.whatever`, so they act a bit like default-valued fields. Specific instances of the class can "override" them by storing them as fields instead. This pattern is not supported by statically-typed languages like C++ or Java, so it's not well known. These could be moved to `__init__()` and treated like normal instance fields.

src/sqlfluff/core/parser/segments/raw.py
barrywhart: Probably so. This was to help me with debugging the parser -- I'm flipping back and forth between this code and the `master` branch, so it's been helpful thus far if these objects resemble the traditional ones. Can definitely revisit this once things are working better.

src/sqlfluff/core/parser/segments/keyword.py
barrywhart: In `master`, we frequently create new classes "on the fly", and we rely on built-in Python functionality to work with them:
* `classmethod`
* `type()` 
* `issubclass()`
* etc.

`classmethod` in particular is problematic now with the introduction of `RawSegmentFactory`. It **acts** like a class, but it's not a class, so it doesn't work with the techniques above. Thus, in `classmethod`s of `RawSegment`, we need a way to pass in the `RawSegmentFactory` as well, so the method can act correctly (because the **type** alone, e.g. `KeywordSegment`, no longer contains all the necessary info).

This bit of code extracts the factory and substitutes it. It may be that in `RawSegment`, we don't need this trickery and should just **always** use `parse_context._factory`.

As I mentioned earlier, we could eventually redesign things more thoroughly so we are not using Python class-related features for parser objects.

alanmcruickshank:
@barrywhart - hold up! Brainwave. I was sitting down to have a go at #897 and I think these two issues are actually really inextricably linked in how they impact the lexer. Mind if I have a crack at trying to address them both together?

barrywhart:
Closing this -- most likely it will be supplanted by #1019 and subequent PRs.

PR #1015, opened 2021-05-03
Move rules library out of sqlfluff.core

alanmcruickshank:
Since we've already done this with the dialects, and we got most of the way with plugins, I think it makes sense to move the main rules library out of `sqlfluff.core`. This means reduces the size of the core library and makes it easier to test and debug.

Part of the intent of this is also so that when testing `pypy` we can easily test a smaller subset of the project.


pwildenhain (approved):
😍 Love the de-coupling, great move. I notice you left the `doc_decorators` in the `core` path. Is that intentional?

alanmcruickshank:
> I notice you left the `doc_decorators` in the `core` path. Is that intentional?

Yeah - that feels like meaningful "central" code which plugins could also refer to. My intent was that anything which another role could meaningfully refer to should stay in core, but anything specific to a rule should move out.

barrywhart (approved):
Looks great!

One possible future change: Based on the `exclude_rules` configuration, we could entirely skip importing of those modules. No big deal for now, but as the number of rules increases, it might be more impactful on startup time, etc.

Issue #1019, opened 2021-05-03
The Big Segment Refactor ™


alanmcruickshank:
This combines #897 & #1014.

Following from discussion betweem @barrywhart and @alanmcruickshank - this supports #642.

This tracks a big refactor which simplifies a chunk of the parser codebase. The following steps need to happen:
1. Refactor Lexer Configs so that Segments are only generated at the last minute, next to where position markers are enriched.
2. Refactor Position markers (and resolve #897) so that markers always have knowledge of source and template. Templated files are always generated, even if a string is passed to `.lex()`.
3. Fully implement segment factories and remove the `.make()` method used to dynamically create segments. Review `BaseSegment.is_type()` to make sure it still works with that in mind.

pwildenhain:
Just have to say, I love the name of this issue 🤣

barrywhart:
@alanmcruickshank: Can this one be closed now? Great work, BTW!

alanmcruickshank:
Yep. I'm going to declare this done 👍 . You should be all good for the parallel work now.

NiallRees:
💰💰💰

PR #1021, opened 2021-05-04
Phase 1 of big refactor (#1019)

alanmcruickshank:
This is going to look like an a) random and b) large PR.

This is phase 1 of #1019. It's a large refactor of the Lexer to:
1. Simplify it significantly.
2. Ensure we never create a base `FilePositionMarker`. This change means we always create `EnrichedFilePositionMarker` objects from the start in the lexer, which is important to achieve before the refactor of the markers themselves in phase 2.

Notably I've also:
- Created a more diverse set of raw segments (`CodeSegment`, `CommentSegment`, `WhitespaceSegment`, `NewlineSegment`) which will eventually support the SegmentFactory work in #1014 which eventually enables #1013 & #642.
- Lexers are now configured using classes directly rather than the slightly obscure tuples we had before. I think this is much clearer. Dialects now invoke `StringMatcher` & `RegexMatcher` directly. Accordingly the methods on the dialect now refer to `lexer_matchers` rather than `lexer_struct`.
- `SingletonMatcher` is now `StringMatcher` because it makes it much more useful and only needed a slight change.
- Removed the `is_segment()` methods and properties, because they weren't being used.

Phase 2 will involve a big simplification of position markers, but this feels like a sensible checkpoint along the way.


alanmcruickshank:
I'm going to continue with the position marker refactor from this point, but I'll do that on a separate branch so we can merge this before that's complete so that these changes don't all land at once.

alanmcruickshank:
There are notes in the code to this effect, but I intent to revisit:
- `TemplateElement.to_segment()`
- `StringMatcher._make_segment_class()`
- `TemplatedFile.make_position_marker()`

..in later phases of the refactor. They currently contain some either odd code, or a lot of warnings and reminders. That's because they will eventually be replaced or reworked either as part of phase 2 (position markers) or 3 (segment factories).

src/sqlfluff/core/parser/lexer.py
```diff
@@ -21,16 +21,50 @@
 lexer_logger = logging.getLogger("sqlfluff.lexer")
 
 
-class LexMatch(namedtuple("LexMatch", ["new_string", "new_pos", "segments"])):
+@dataclass
```
barrywhart, line 24: Are these objects mutable? If they are immutable, did you think about using `typing.NamedTuple`?

src/sqlfluff/core/parser/lexer.py
```diff
@@ -42,125 +76,121 @@ def __init__(
...
+        """The private match function. Just look for a literal string."""
+        if forward_string.startswith(self.template):
+            return LexedElement(self.template, self)
         else:
             return None
 
-    def _trim(self, matched, start_pos):
+    def search(self, forward_string: str) -> Optional[Tuple[int, int]]:
+        """Use string methods to find a substring."""
+        if self.template in forward_string:
```
barrywhart, line 103: Performance nitpick: Here we are searching for the string twice, first with `in`, then with `find`. If `find()` doesn't find the string, it returns `-1`, so we could potentially streamline this.

src/sqlfluff/core/parser/lexer.py
```diff
@@ -310,200 +288,226 @@ def lex(
...
-
+            # The calculated source slice will include any source only slices.
+            # We should consider all of them in turn to see whether we can
+            # insert them.
             for source_only_slice in source_only_slices:
+                # If it's later in the source, stop looking. Any later
+                # ones *also* won't match.
                 if source_only_slice.source_idx > source_slice.start:
                     break
+                # Is there a templated section within this soure slice?
```
barrywhart, line 361:
```suggestion
                # Is there a templated section within this source slice?
```

barrywhart (commented):
I'm about halfway through. Will continue reviewing later today.

src/sqlfluff/core/parser/lexer.py
```diff
@@ -310,200 +288,226 @@ def lex(
...
+                res = matcher.match(forward_string)
+                if res.elements:
+                    # If we have new segments then whoop!
+                    elem_buff += res.elements
+                    forward_string = res.forward_string
+                    # Cycle back around again and start with the top
+                    # matcher again.
+                    break
+                else:
+                    continue
```
barrywhart, line 487: Does this line do anything? It seems like execution would go to the top of the `for` loop next regardless.

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -11,8 +11,9 @@ class RawSegment(BaseSegment):
     """This is a segment without any subsegments."""
 
     type = "raw"
-    _is_code = False
+    _is_code = True
```
barrywhart, line 14: Why this change?

barrywhart (approved):
Looks good overall! It's difficult to grok design details from reading a PR, but. I did a quick read through of all the code.

src/sqlfluff/core/parser/segments/raw.py
alanmcruickshank: Just ease. It's more likely to be true than false for any given segment.

tbh, this whole section might get trashed when we do the `ProtoSegment`.

src/sqlfluff/core/parser/lexer.py
alanmcruickshank: Good idea :+1:

PR #1027, opened 2021-05-07
Phase 2 of big refactor (#1019)

### LINKED ISSUE #897: Refactor Position Markers

alanmcruickshank:
Another biggie. Resolves #897 . Supports #1019 .

This is phase 2 of #1019. It's a large refactor of position markers.
- Deprecation of `FilePositionMarker` and `EnrichedFilePositionMarker`. We now just create `PositionMarker`. This total rewrite of that code means that position markers support positioning in the source, templated and fixed file more explicitly as seperate references.
- `PositionMarker` also holds a reference to the `TemplatedFile` (which based on phase 1 is now always present). This allows the marker to introspect several things about it's position itself.
- `PositionMarker._position_segments()` does all of the heavy lifting to fill missing markers and readjust those positions as fixes are made. This removes all the position logic from rules, simplifying the logic there and making segment creation much simpler.
- The new `PositionMarker` has no concept of `statement_index` and so accordingly I've adjusted the formatting of how positions are displayed to display the character in the file, the line number and the line position instead.
- Notably, the core code changes here are restricted mostly to `markers.py`, however the implications of those changes are **very wide**, so most of this PR is actually dealing with the implications of those changes.

I've made this PR initially as a draft for review, and also while I merge in upstream PRs including #1021


NiallRees:
😍
image

src/sqlfluff/core/errors.py
```diff
@@ -12,6 +12,12 @@ class SQLBaseError(ValueError):
 
     def __init__(self, *args, **kwargs):
         self.ignore = kwargs.pop("ignore", False)
+        pos = kwargs.pop("pos", None)
```
barrywhart, line 15: Nitpick: I think the `pop()` calls could be replaced by adding:

```
pos=None, line_no=0, line_pos=0
```

to the function parameter list.

src/sqlfluff/core/parser/markers.py
```diff
@@ -1,190 +1,205 @@
...
+        - This combines the previous functionality of FilePositionMarker
+          and EnrichedFilePositionMarker. Additionally it contains a reference
+          to the original templated file.
+        - It no longer explicitly stores a line number or line position in the
+          source or template. This is extrapolated from the templated file as required.
+        - Positions in the source and template are with slices and therefore identify
+          ranges.
+        - Positions within the fixed file are identified with a line number and line
+          position, which identify a point.
+          the fixed file, this is as a combination of line number and line position.
```
barrywhart, line 27: This line looks like it doesn't belong, maybe a leftover from editing?

src/sqlfluff/core/parser/markers.py
```diff
@@ -1,190 +1,205 @@
...
+        - Arithmetic comparisons are on the location in the fixed file.
+    """
+
+    source_slice: slice
```
barrywhart, line 31: I assume `slice` is a type? I was briefly confused by a lower-case type, which is typically just built-in types such as `str` or `int`.

src/sqlfluff/core/parser/markers.py
```diff
@@ -1,190 +1,205 @@
...
+    templated_slice: slice
+    templated_file: "TemplatedFile"
+    # If not set, these will be initialised in the post init.
+    working_line_no: int = -1
+    working_line_pos: int = -1
+
+    def __post_init__(self):
```
barrywhart, line 38: Today I learned about this function!

src/sqlfluff/core/parser/markers.py
```diff
@@ -1,190 +1,205 @@
...
+            working_line_no=line_no,
+            working_line_pos=line_pos,
+        )
+
+    def is_literal(self) -> bool:
+        """Infer literalness from context."""
+        # NOTE: We should define what the implication of is_literal is here.
+        # TODO: I think we were inconsistent in whether a zero length source slice
+        # is (or is not) literal. The lexer initially defined it as False, this
+        # function will say True. Not sure what the implications of that are yet.
```
barrywhart, line 195: This is one of those moments when it feels somewhat comforting to have solid test coverage. 💪🏽

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -131,13 +117,21 @@ def __init__(self, segments, pos_marker=None, validate=True):
...
-        # Equal if type, content and pos are the same
         # NB: this should also work for RawSegment
         return (
             # Same class NAME. (could be constructed elsewhere)
             self.__class__.__name__ == other.__class__.__name__
             and (self.raw == other.raw)
-            and (self.pos_marker == other.pos_marker)
+            # Both must have a non-null position marker to compare.
+            and self.pos_marker
+            and other.pos_marker
```
barrywhart, line 127: Is there potential for confusing behavior now or someday if one has a `pos_marker` and the other doesn't? (Wondering if, for example, an exception should be raised rather than simply returning `False`?)

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -281,52 +275,79 @@ def expand(segments, parse_context):
...
-            running_pos = running_pos.advance_by(raw=seg.raw, idx=idx)
-            # Add the buffer to my new segment
-            seg_buffer.append(seg)
-        return tuple(seg_buffer)
+                linter_logger.warning("SEG: %r, POS: %r", segments, parent_pos)
+                raise ValueError("Unable to find working position.")
+
+        # Use the index so that we can look forward
+        # and backward.
+        for idx in range(len(segments)):
```
barrywhart, line 313: Nitpick: Since the loop body frequently refers to `segments[idx]`, should we consider changing the loop to:

```
for idx, segment in enumerate(segments):
```

Then, all references to `segments[idx]` could be replaced with `segment`. This is. not a huge deal, but it seems more Pythonic and slightly more readable.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -281,52 +275,79 @@ def expand(segments, parse_context):
...
+                # Search forward for a following one, if we have to?
+                else:
+                    for fwd_seg in segments[idx + 1 :]:
+                        if fwd_seg.pos_marker:
+                            segments[
+                                idx
+                            ].pos_marker = fwd_seg.pos_marker.start_point_marker()
+                            break
+                    else:
+                        raise ValueError("Unable to positon new segment")
```
barrywhart, line 333:
```suggestion
                        raise ValueError("Unable to position new segment")
```

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -281,52 +275,79 @@ def expand(segments, parse_context):
...
-        if not todo_buffer:
-            return ()
-        # If starting pos not provided, take it from the first of the buffer.
-        running_pos = starting_pos
-        if not running_pos:
-            for seg in todo_buffer:
-                if not seg.is_meta:
-                    running_pos = seg.pos_marker
+    @classmethod
+    def _position_segments(cls, segments, parent_pos=None):
```
barrywhart, line 279: This is nice code! I expected it to be really intricate and hard to follow by its nature, but the naming, comments, etc. are all 💯.

src/sqlfluff/core/rules/base.py
```diff
@@ -129,9 +131,17 @@ def __init__(self, edit_type, anchor, edit=None):
...
+            # Check that any edits don't have a position marker set.
+            # We should rely on realignment to make position markers.
             # Strip position markers of anything enriched, otherwise things can get blurry
             for seg in self.edit:
-                seg.pos_marker = seg.pos_marker.strip()
+                if seg.pos_marker:
+                    # Developer warning.
+                    rules_logger.debug(
+                        "Developer Note: Edit segment found with preset position marker. "
+                        "These should be unset and calculated later. %r"
```
barrywhart, line 142: The log message contains `%r`, but you're not passing a parameter. Should you be passing `seg` as a parameter to `debug()`?

src/sqlfluff/core/templaters/base.py
```diff
@@ -159,7 +160,7 @@ def get_line_pos_of_char_pos(
         while nl_idx + 1 < len(ref_str) and ref_str[nl_idx + 1] < char_pos:
             nl_idx += 1
 
-        # NB: +1 because character position is 0-indexed, but the character
+        # NB: +1 because character position is 0-indexed, but the line
```
barrywhart, line 163: I'm having trouble interpreting this comment. Is it referring _just_ to the `else` block, and if so, is it referring to the first value (`1`) or the second one (`char_pos + 1`)?

src/sqlfluff/rules/L018.py
```diff
@@ -99,9 +99,7 @@ def indent_size_up_to(segs):
                                     LintFix(
                                         "create",
                                         seg,
-                                        self.make_whitespace(
-                                            " " * (-indent_diff), seg.pos_marker
-                                        ),
+                                        self.make_whitespace(" " * (-indent_diff)),
```
barrywhart, line 102: Just noticing this... does this mean lint rules don't need to provide a position for fixes anymore? If so, that's a really nice simplification for rule developers! 💯 🎉  I (and possibly other people) are probably guilty of just passing **some** value from some segment without much thought or understanding of the significance.

src/sqlfluff/rules/L046.py
```diff
@@ -45,25 +45,19 @@ def _get_whitespace_ends(s):
         pos = main.find(inner)
         return main[:pos], inner, main[pos + len(inner) :]
 
-    def _eval(self, segment, templated_file, memory, **kwargs):
-        # Extract some data from the segment. Importantly, all
-        # of these have defaults in case we don't have an
-        # enriched position marker.
-        source_slice = getattr(segment.pos_marker, "source_slice", None)
-        is_literal = getattr(segment.pos_marker, "is_literal", None)
```
barrywhart, line 53: Was this variable (`is_literal `) actually used anywhere? I see its removal but no corresponding uses of it being removed.

test/cli/formatters_test.py
```diff
@@ -26,8 +27,15 @@ def test__cli__formatters__violation():
...
+            slice(10, 19),
+            slice(10, 19),
+            TemplatedFile.from_string("      \n\n  foobarbar"),
+        ),
+    )
     r = RuleGhost("A", "DESC")
     v = SQLLintError(segment=s, rule=r)
     f = format_violation(v)
-    assert escape_ansi(f) == "L:  20 | P:  11 |    A | DESC"
+    assert escape_ansi(f) == "L:   3 | P:   3 |    A | DESC"
```
barrywhart, line 41: Could you perhaps add a comment explaining why the expected position is 3, 3?

test/core/parser/markers_test.py
```diff
@@ -1,43 +1,42 @@
...
+import pytest
+
+from sqlfluff.core.templaters import TemplatedFile
+from sqlfluff.core.parser.markers import PositionMarker
+
+
+@pytest.mark.parametrize(
+    "raw,start_pos,end_pos",
+    [
+        ("fsaljk", (0, 0), (0, 6)),
```
barrywhart, line 12: Great test! I love parametrized tests because they emphasize the values rather than the code -- very helpful for future maintainers.

test/core/parser/markers_test.py
```diff
@@ -1,43 +1,42 @@
...
+def test_markers__infer_next_position(raw, start_pos, end_pos):
+    """Test that we can correctly infer positions from strings."""
+    assert end_pos == PositionMarker.infer_next_position(raw, *start_pos)
+
+
+def test_markers__setting_position_raw():
+    """Test that we can correctly infer positions from strings & locations."""
+    templ = TemplatedFile.from_string("foobar")
+    # Check inference in the template
+    assert templ.get_line_pos_of_char_pos(2, True) == (1, 3)
```
barrywhart, line 29: Nitpick: Could you pass the `True` / `False` values as keyword parameters? That'd make it a bit easier to understand what it means without having to refer to the called function.

test/core/parser/segments_base_test.py
```diff
@@ -83,25 +91,27 @@ def test__parser__base_segments_base(raw_seg_list, fresh_ansi_dialect):
         (raw_seg_list[0].to_tuple(), raw_seg_list[1].to_tuple()),
     )
     # Check Formatting and Stringification
-    assert str(base_seg) == repr(base_seg) == "<DummySegment: ([3](1, 1, 4))>"
+    assert str(base_seg) == repr(base_seg) == "<DummySegment: ([C:   0, L:  1, P:  1])>"
     assert base_seg.stringify(ident=1, tabsize=2) == (
-        "[3](1, 1, 4)        |  dummy:\n"
-        "[3](1, 1, 4)        |    raw:                                                      'foobar'\n"
-        "[9](1, 1, 10)       |    raw:                                                      '.barfoo'\n"
+        "[C:   0, L:  1, P:  1]|  dummy:\n"
```
barrywhart, line 96: What does the `C:` value mean?

barrywhart (changes requested):
Looks pretty good overall. A few suggestions, one possible logging bug.

src/sqlfluff/core/parser/markers.py
alanmcruickshank: `slice` is a built in type indeed. https://docs.python.org/3/glossary.html#term-slice

test/core/parser/segments_base_test.py
alanmcruickshank: character position in the file. Do you think it adds anything? I could always remove it.

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: Potentially but unlikely. `null` position markers are assigned positions almost straight after addition in the fix routines. I've been thinking of the missing position marker more as _delegating_ the positioning to the parent segment, rather than trying to work it out itself.

src/sqlfluff/core/parser/markers.py
alanmcruickshank: I think I've flushed out all the the cases where this matters. I'll update this comment.

src/sqlfluff/core/errors.py
alanmcruickshank: I think this is a hangover from python2 or maybe python3.5. Back in those days you could have `*args`, `**kwargs` **and** specified keyword arguments that weren't also in `args`. That means we broke some of the Exception code. I think you're right that it should work now though 👍

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: Excellent idea!

src/sqlfluff/core/rules/base.py
alanmcruickshank: Great question. I'm not sure. I'll remove the reference.

src/sqlfluff/core/templaters/base.py
alanmcruickshank: Yes - and I'll move the comment to make this clearer 👍

src/sqlfluff/rules/L018.py
alanmcruickshank: yes yes yes 🎉 .

The logging actually warns people now when the *do* provide a position - because while you *can*, you really shouldn't. `_position_segments` handles all of that.

src/sqlfluff/rules/L046.py
alanmcruickshank: Yeah it was used in line 56 below, but I'm now calling `.is_literal()` directly.

test/core/parser/segments_base_test.py
alanmcruickshank: It used to be the number in the square brackets of the original position readout. i.e. `[9](1, 1, 10)` would now be `[C:   9, L:  1, P: 10]`, but I'm in two minds about whether anyone cares. I find it a little useful for debugging, but not actually while using it as SQLFluff as a tool. The other option would be to just make it `[L:  1, P: 10]` for simplicity. @barrywhart @NiallRees what do you guys think?

barrywhart: I suggest removing it, at least for user-facing output. I can remember very few cases where I was thinking about byte offsets into a text file.

src/sqlfluff/core/parser/markers.py
```diff
@@ -188,16 +187,19 @@ def with_working_position(self, line_no: int, line_pos: int):
...
-        # function will say True. Not sure what the implications of that are yet.
-        # NOTE: So far I can see it only controls:
-        # 1. The patch_type of a FixPatch, which appears to then never be referred to.
-        # 2. The ignoring of templated sections in `remove_templated_errors`.
-        # 3. Whether we can return simply in `iter_patches`.
-        # 4. The application of L046 which is looking for tags.
+        """Infer literalness from context.
+
+        is_literal should return True if a fix can be applied across this area
+        without concern. This obviously applies to any slices which are the same
```
barrywhart, line 193: Can you clarify what you mean by "without concern"? Concern that ______? If you could, please consider providing an example or two, also a counterexample.

barrywhart (approved):
Looks good! A couple small remaining things if you like:
* Removing the "C" (character position in the file) output
* Clarifying what "without concern" means (new comment).

src/sqlfluff/core/parser/markers.py
alanmcruickshank: Yeah - not very clear is it. I've made it clearer.

PR #1030, opened 2021-05-09
Add filter grammar

### LINKED ISSUE #346: [postgres] Support FILTER clause
Please support the `FILTER (WHERE ...)` clause on aggregate expressions:
* https://www.postgresql.org/docs/12/sql-expressions.html#SYNTAX-AGGREGATES

Simple failing example:
```sql
SELECT
    COUNT(*) FILTER (WHERE c_expires > CURRENT_TIMESTAMP) AS c_active
FROM t_test;
```
...results into:
```
== [db/test.sql] FAIL
L:   2 | P:  14 | L012 | Implicit aliasing of column not allowed. Use explicit
                       | `AS` clause.
L:   2 | P:  14 | L029 | Keywords should not be used as identifiers.
L:   2 | P:  20 |  PRS | Found unparsable section: ' (WHERE c_expires >
                       | CURRENT_TIMESTAMP) A...'
L:   3 | P:   6 | L014 | Inconsistent capitalisation of unquoted identifiers.
```

NiallRees:
Resolves #346

### DIFF
--- a/src/sqlfluff/dialects/dialect_exasol.py
+++ b/src/sqlfluff/dialects/dialect_exasol.py
@@ -193,6 +193,10 @@ exasol_dialect.replace(
     BooleanLiteralGrammar=OneOf(
         Ref("TrueSegment"), Ref("FalseSegment"), Ref("UnknownSegment")
     ),
+    PostFunctionGrammar=Sequence(
+        Sequence(OneOf("IGNORE", "RESPECT"), "NULLS", optional=True),
+        Ref("OverClauseSegment"),
+    ),
 )
 
 ############################
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -387,6 +387,9 @@ ansi_dialect.add(
         Ref("ColumnReferenceSegment"),
         Ref("ExpressionSegment"),
     ),
+    FilterClauseGrammar=Sequence(
+        "FILTER", Bracketed(Sequence("WHERE", Ref("ExpressionSegment")))
+    ),
 )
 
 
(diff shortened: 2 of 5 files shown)


NiallRees:
@barrywhart I'm a bit stumped by

```
RuntimeError: Grammar refers to 'FilterKeywordSegment' which was not found in the exasol dialect
```

Any ideas?

src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -193,6 +193,10 @@
     BooleanLiteralGrammar=OneOf(
         Ref("TrueSegment"), Ref("FalseSegment"), Ref("UnknownSegment")
     ),
+    PostFunctionGrammar=Sequence(
```
NiallRees, line 196: I was tempted to just add `FILTER` as an unreserved keyword to the exasol dialect to overcome this, but exasol doesn't support the filter syntax so I decided to replace the `PostFunctionGrammar` even though this feels verbose.

barrywhart: This seems reasonable to me -- not that many lines of additional code. If you wanted to avoid it, one possibility would be to refactor the ANSI `PostFunctionGrammar` into something like:

```
    PostFunctionGrammar=OneOf(
        # Optional OVER suffix for window functions.
        # This is supported in biquery & postgres (and its derivatives)
        # and so is included here for now.
        Sequence(
            Sequence(OneOf("IGNORE", "RESPECT"), "NULLS", optional=True),
            Ref("OverClauseSegment"),
        ),
        Ref("BaseFilterClauseGrammar"),
    ),
```

Then define `BaseFilterClauseGrammar` in ANSI and override it to be blank in EXASOL. I did something slightly similar in #1011.

I'm curious, what was the error you were getting in EXASOL? I encountered an interesting one a week or so ago with EXASOL. I was enhancing the lexer for floating-point values, and I was initially allowing trailing decimals (e.g. `5.`). That caused an error with EXASOL, which supports some sort of range syntax, e.g. (`5..10`). EXASOL has 🐉 🐉 ! 😛

barrywhart (approved):
:shipit:

NiallRees:
I hadn't considered that @barrywhart, neat idea.

The error was

```
Grammar refers to 'FilterKeywordSegment' which was not found in the exasol dialect
```

Once I worked out how it was generate it made sense.

PR #1036, opened 2021-05-10
Remove Indent.when (#1019)

alanmcruickshank:
This follows on from #1034, and should be reviewed after that is merged. It relates to #1019,

This addresses Indent and Dedent segments so that they don't involve dynamic class generation. Specifically the `Indent.when()` method has been replaced by a `Conditional` grammar, which emulates the same functionality.

This is less compact than the previous syntax, but it avoids any type creation, and probably more accurately explains what's going on. It also provides a basis for more iteration.

Before:
```
Indent.when(indented_joins=False)
```

After:
```
Conditional(Indent, config_type="indent", indented_joins=False)
```

The effect of the two above is equivalent.

Notably, this simplifies the creation of indents around templated blocks during lexing, but that's so far down in the codebase that I'm not sure the effect of that is very widespread 🤷 .


src/sqlfluff/core/parser/grammar/conditional.py
```diff
@@ -0,0 +1,67 @@
...
+                "Conditional is only designed to work with Indent segments."
+            )
+        if len(args) != 1:
+            raise ValueError(
+                "Conditional is only designed to work with a single element."
+            )
+        if not config_type:
+            raise ValueError("Conditional config_type must be set.")
+        if not kwargs:
+            raise ValueError("Conditional requires rules to be set.")
```
barrywhart, line 35: What's the connection between `kwargs` and `rules`? Are we assuming that `kwargs` will contain `rules` as a key if it's not empty?

src/sqlfluff/core/parser/grammar/conditional.py
```diff
@@ -0,0 +1,67 @@
...
+        self._config_type = config_type
+        self._config_rules = kwargs
+        super().__init__(*args)
+
+    def is_enabled(self, parse_context):
+        """Evaluate conditionals and return whether enabled."""
+        # If no config rules are set then it's always enabled.
+        try:
+            config = {"indent": parse_context.indentation_config}[self._config_type]
+        except KeyError:
```
barrywhart, line 45: This code reads a bit strangely to me. IIUC, it's doing two things:
* Verifying that `self._config_type` is `"indent"`
* Assigning `config = parse_context.indentation_config`

Could we do this separately? E.g.
```
if self._config_type != "indent":
    raise ValueError(...)
config = parse_context.indentation_config
```

I think this version is easier to read.

barrywhart (changes requested):
A small suggested change for readability... Overall, looks good!

src/sqlfluff/core/parser/grammar/conditional.py
alanmcruickshank: you read it correctly. The idea here is that we might have other kinds of config other than "indent" might be possible in future. The code is kind of hard to read though, I'll split it up a bit.

src/sqlfluff/core/parser/grammar/conditional.py
alanmcruickshank: correct. I might just rename `kwargs` to `rules` for clarity

src/sqlfluff/core/parser/grammar/conditional.py
alanmcruickshank: Ok, I've simplified it a load, and taken out some of the logic around other configuration paths for now. I've extended the documentation a lot to describe where we those points might be. Would love your views on the changes.

barrywhart: Sure! I'll try and take a look today.

src/sqlfluff/core/parser/grammar/conditional.py
```diff
@@ -12,15 +12,36 @@
...
+    | NOTE: The Conditional grammar is assumed to be operating
+    | within a Sequence grammar, and some of the functionality
+    | may not function within a different context.
+
     Args:
-        enforce_whitespace_preceeding (:obj:`bool`): Should the GreedyUntil
-            match only match the content if it's preceded by whitespace?
-            (defaults to False). This is useful for some keywords which may
-            have false alarms on some array accessors.
+        *args: A meta segment which is intantiated
```
barrywhart, line 20:
```suggestion
        *args: A meta segment which is instantiated

```

src/sqlfluff/core/parser/grammar/conditional.py
```diff
@@ -12,15 +12,36 @@
...
+        rules: A set of `rule=boolean` pairs, which are
+            evaluated when understanding whether conditions
+            are met for this grammar to be enabled.
+
+    Example:
+        .. code-block::
+
+            Conditional(Dedent, config_type="indent", indented_joins=False)
+
+        This effectively says that if `indented_joins` in the "indent" section
```
barrywhart, line 33: Great explanation! 💯

src/sqlfluff/core/parser/grammar/conditional.py
```diff
@@ -31,26 +52,30 @@ def __init__(self, *args, config_type: str = "", **kwargs):
...
         super().__init__(*args)
 
     def is_enabled(self, parse_context):
         """Evaluate conditionals and return whether enabled."""
-        # If no config rules are set then it's always enabled.
-        try:
-            config = {"indent": parse_context.indentation_config}[self._config_type]
-        except KeyError:
+        # NOTE: Because only "indentation" is the only current config_type
+        # supported, this code is much simpler that would be requried in
```
barrywhart, line 68:
```suggestion
        # supported, this code is much simpler that would be required in

```

barrywhart (approved):
The latest changes looks great! I noticed a couple of typos in comments, otherwise 💯 .

PR #1039, opened 2021-05-11
Read config from pyproject.toml

### LINKED ISSUE #890: Look for project configuration in pyproject.toml.
A growing number of Python tools look for project-specific configuration in the `[tool.{app}]` section of the `pyproject.toml` file. It would be convenient if `sqlfluff` also inspected this section, especially for Python projects that use `poetry` as a build backend and thus do not rely on `setup.cfg` or potentially any of the other files currently inspected.

edgarrmondragon:
This PR adds support for `pyproject.toml` as a configuration source.

It would help keeping all configuration for a Python project in a single file, since this file is used by other major Python projects like [black] and [pytest].

Some implementation details:

- Takes precedence over `.sqlfluff`.

- `core` section lives in `[tool.sqlfluff.core]`.

    ```toml
    [tool.sqlfluff.core]
    templater = "jinja"
    sql_file_exts = [
        ".sql",
        ".sql.j2",
        ".dml",
        ".ddl",
    ]

    [tool.sqlfluff.indentation]
    indented_joins = false
    template_blocks_indent = false

    [tool.sqlfluff.templater]
    unwrap_wrapped_queries = true

    [tool.sqlfluff.templater.jinja]
    apply_dbt_builtins = true
    ```

Closes #890

[black]: https://black.readthedocs.io/en/stable/pyproject_toml.html
[pytest]: https://docs.pytest.org/en/6.2.x/customize.html#pyproject-toml


docs/source/configuration.rst
```diff
@@ -19,13 +19,44 @@ any vales read from earlier files.
...
-are delimited by a semicolon. For example the *jinjacontext* section will
-be indicated in the section started with *[sqlfluff:jinjacontext]*.
+Within these files, the first four will be read like an `cfg file`_, and
+*SQLFluff* will look for sections which start with *SQLFluff*, and where
+subsections are delimited by a semicolon. For example the *jinjacontext*
+section will be indicated in the section started with
+*[sqlfluff:jinjacontext]*.
+
+For the `pyproject.toml file`_, all valid sections start with `tool.sqlfluff`
+and subsections are delimited by a dot and suffixed with *__options*.
```
barrywhart, line 31: You mentioned the `__options` suffix in the PR description. Can you give an example of the kind of conflict you're avoiding by doing this?

If I understand correctly, could the toml reader skip sections that don't start with `tools.sqlfluff` or `sqlfluff`?

src/sqlfluff/core/config.py
```diff
@@ -144,6 +144,34 @@ def get_global(cls) -> "ConfigLoader":
...
+        return buff
+
+    @staticmethod
+    def _get_config_elems_from_toml(fpath: str) -> List[Tuple[tuple, Any]]:
+        """Load a config from a TOML file and return a list of tuples.
+
+        The return value is a list of tuples, were each tuple has two elements,
+        the first is a tuple of paths, the second is the value at that path.
+        """
+        import toml
```
barrywhart, line 168: Could this import be moved to the top of the file? That's generally the preferred location for imports.

barrywhart (changes requested):
The PR looks pretty good. I would like to understand the reason for the "__options" suffix in the .toml file. I wonder if there may be other options.

Also, I think we need to update the configuration docs in `ocs/source/configuration.rst`, which documents how configuration works, what files it looks at, etc.

src/sqlfluff/core/config.py
edgarrmondragon: Sure. I'll do that.

docs/source/configuration.rst
edgarrmondragon: So a TOML is basically a nested  object/dictionary so the keys would collide. The following code raises an error

```python
import toml

content = """
[tool.sqlfluff]
templater = "jinja"

[tool.sqlfluff.templater]
unwrap_wrapped_queries = true
"""

toml.loads(content)
```

The error is:

```
What? templater already exists?{'templater': 'jinja'} (line 5 column 1 char 38)
```

> If I understand correctly, could the toml reader skip sections that don't start with tools.sqlfluff or sqlfluff?

Yes, the other sections are ignored. The convention I believe is to put every set of configurations for `my_tool` under `[tool.my_tool*]` sections.

edgarrmondragon:
@barrywhart 

> The PR looks pretty good. I would like to understand the reason for the "__options" suffix in the .toml file. I wonder if there may be other options.

I think one option, though it would existing sections names, would be to avoid the same keys for settings and sections, e.g. `templater=jinja` and `[sqlfluff:templater]`.

> Also, I think we need to update the configuration docs in `ocs/source/configuration.rst`, which documents how configuration works, what files it looks at, etc.

I added a few lines and an example TOML block to the first part of the configuration page. Also added `pyproject.toml` to the list of files. Do you think I'm missing something else?

src/sqlfluff/core/config.py
```diff
@@ -144,6 +146,31 @@ def get_global(cls) -> "ConfigLoader":
...
+
+    @classmethod
+    def _get_config_elems_from_toml(cls, fpath: str) -> List[Tuple[tuple, Any]]:
+        """Load a config from a TOML file and return a list of tuples.
+
+        The return value is a list of tuples, were each tuple has two elements,
+        the first is a tuple of paths, the second is the value at that path.
+        """
+        config = toml.load(fpath)
+        tool = config.get("tool", {}).get("sqlfluff", {})
```
barrywhart, line 170: Now that you replaced `__options` with the `tool.sqlfluff.core` technique, do you need to move or place the `core` values to match the structure of the other config file types?

In other words, have you verified that SQLFluff correctly uses the values from the `core` section?
```
[tool.sqlfluff.core]
    templater = "jinja"
    sql_file_exts = [
        ".sql",
        ".sql.j2",
        ".dml",
        ".ddl",
    ]
```

edgarrmondragon: @barrywhart I added a test case with

```toml
[tool.sqlfluff.core]
testing_int = 5
testing_bar = 7.698
testing_bool = false
testing_arr = [ "a", "b", "c" ]
testing_inline_table = { x = 1 }
 
[tool.sqlfluff.bar]
foo = "foobar"

[tool.sqlfluff.fnarr.fnarr]
foo = "foobar"
```

which gets collected into the following dict:

```python
{
    "core": {
        "testing_int": 5,
        "testing_bar": 7.698,
        "testing_bool": False,
        "testing_arr": ["a", "b", "c"],
        "testing_inline_table": {"x": 1},
    },
    "bar": {"foo": "foobar"},
    "fnarr": {"fnarr": {"foo": "foobar"}},
}
```

does it look right?``

barrywhart: TBH, I'd feel much better if you could do some manual tests to confirm the settings in the TOML file are being read correctly and actually affecting the runtime behavior of SQLFluff. There are sometimes limits to the effectiveness of automated tests and manual review, and I think this is one of those times.

edgarrmondragon: @barrywhart  That makes a lot of sense. I created [this repo](https://github.com/edgarrmondragon/sqlfluff-pyproject) to test some configurations. Let me know if you think there are edge cases I could add.

barrywhart: I tested with your example repo. Works great! I also added a simple test with templating. This is based on the Variable Templating [example](https://docs.sqlfluff.com/en/stable/configuration.html#variable-templating) from the SQLFluff docs.

`test.sql`:
```
SELECT {{ num_things }} FROM {{ tbl_name }} WHERE id > 10 LIMIT 5
```

Additions to `pyproject.toml`:
```
[tool.sqlfluff.core]
templater = "jinja"

[tool.sqlfluff.templater.jinja.context]
num_things=456
tbl_name="my_table"
```

barrywhart (approved):
Great work! Thanks for adding the test project. 👍

src/sqlfluff/core/config.py
edgarrmondragon: I added that templated example to the example repo :smile:.

PR #1044, opened 2021-05-12
Fail fast on fatal errors

alanmcruickshank:
We're currently having an issue where SQLFluff takes ages to fail, but results in the same failure on every file. Specifically it's where the dbt templater fails to compile and then keeps attempting (This clearly takes a LONG time).

This adds a `fatal` flag to some errors which enables them to terminate linting entirely, to return an error to the user faster.


NiallRees (approved):
👍

PR #1052, opened 2021-05-15
Clean up some uses of "# type: ignore"

barrywhart:
No functional changes here, just small changes to reduce the use of `# type: ignore`


src/sqlfluff/core/linter.py
```diff
@@ -285,7 +286,8 @@ def fix_string(self) -> Tuple[Any, bool]:
         bencher("fix_string: start")
 
         linter_logger.debug("Original Tree: %r", self.templated_file.templated_str)
-        linter_logger.debug("Fixed Tree: %r", self.tree.raw)  # type: ignore
+        assert self.tree
```
NiallRees, line 289: I'm not a typing wiz - what exactly is going on here?

barrywhart: Previously, mypy would complain because `self.tree` is an `Optional` value (i.e. could be `None`). Mypy pays attention to assertions, so when we assert, this guarantees the value is not `None`, therefore the types match.

I can't find any docs explaining this exact case, but see the "Note" on this page: it's pretty similar, although it uses `isinstance()`.

https://mypy.readthedocs.io/en/stable/casts.html#casts-and-type-assertions

barrywhart: Ah, I found an official mention of this in the [MyPy docs](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type):
>Sometimes mypy doesn’t realize that a value is never `None`. This notably happens when a class instance can exist in a partially defined state, where some attribute is initialized to `None` during object construction, but a method assumes that the attribute is no longer `None`. Mypy will complain about the possible None value. You can use `assert x is not None` to work around this in the method

NiallRees (approved):

src/sqlfluff/core/linter.py
NiallRees: Sweet thanks for the link 👍

Issue #1037, opened 2021-05-10
When running "sqlfluff fix" from script, "WARNING" is output to stdout (please output to "stderr")


yaegassy:
## Description

I started to make a plugin that works with `sqlfluff` in an Editor (Vim + coc.nvim). 😇

![coc-sqlfluff-demo](https://user-images.githubusercontent.com/188642/117739387-3959d280-b239-11eb-87cf-e31f12a2f129.gif)

I noticed during the formatter that when "WARNING" is present, it is output to `stdout`.

If there is unnecessary output in stdout, there will be problems with the formatter feature (e.g., unnecessary strings must be removed).

I think it needs to be modified to output to `stderr`.

## Expected Behaviour

WARNING message should be output to `stderr`

## Observed Behaviour

WARNING message has been output to `stdout`.

## Steps to Reproduce

You can see that "WARNING" is output to `stdout` with this code, You can also check that `stderr` is empty.

### Prepare

```
mkdir -p /tmp/sqlfluff-check
python3 -m venv /tmp/sqlfluff-check/venv
/tmp/sqlfluff-check/venv/bin/pip3 install sqlfluff

cd /tmp/sqlfluff-check/
source venv/bin/activate
```

### Sample code (check.py)

```python
import subprocess

#
# stdin data
#

input_text = """WITH all_upstream_matches AS (
                            SELECT
                                business_type,
                                ROW_NUMBER() OVER (
                                     PARTITION BY
                 low_business_type,
                 low_size_label,
                 low_gender_label,
                low_age_label
            ORDER BY
                business_type DESC,
                size_label DESC,
                gender_label DESC,
                age_label DESC
        ) AS rownum
    FROM
        acceptable_buckets
    JOIN
        small_buckets
        ON
            (business_type = low_business_type
             AND size_label = low_size_label
             AND gender_label = low_gender_label
             AND age_label = low_age_label)
)

SELECT
    business_type,
    user_counts
FROM
    acceptable_buckets
UNION ALL
SELECT
    business_type,
    user_counts
FROM
    substituted_buckets
""".strip()

p = subprocess.run(["sqlfluff", "fix", "--force", "-"], input=input_text, text=True, capture_output=True)

#
# debug print
#

print("======== STDOUT ========¥n")
print(p.stdout)

print("======== STDERR ========¥n")
print(p.stderr)
```

### RESULT

```

(venv) sqlfluff-check$ python3 check.py
======== STDOUT ========

[...]

PR #1055, opened 2021-05-15
Send logging to stderr when stdin is used in CLI fix

nolanbconaway:
A proposal to fix https://github.com/sqlfluff/sqlfluff/issues/1037: send all logging to stderr.

The options as I saw it were:

1. (annoying) whenever logging is being done, check if stdin is being used for input. if so, do not log.
2. (more annoying) whenever logging is being done, check if stdin is being used for input. if so, configure the logger stream to stderr instead of stdout.
3. (least annoying) send everything to stderr.

I can see the places where the issue in #1037 arises and can easily implement solution 1 there, but this doesn't fix the general issue that we are using stdout for output AND logging at times.

I don't know about the use cases of loggers enough to know if everything _oughtn't_ to go to stderr, would love some more expert perspectives!


barrywhart:
I think this is a good idea. We may want to use one of the techniques here to ensure that stuff written to stdout is flushed before any subsequent writes to stderr. Otherwise, I think we could see some weird intermingling of the two.

Reason: By default, stdout is buffered, but stderr is unbuffered.

barrywhart:
Forgot the link: https://blog.finxter.com/what-is-python-output-buffering-and-how-to-disable-it/

nolanbconaway:
I thought that click.echo() was supposed to handle some of this complexity? I don't have a full view into all the reasons that we use logging, but I will say that it is extremely annoying to get tests  (notably, `test__cli__command_lint_warning_explicit_file_ignored`) to pass under this stderr stream handler, given that pytest and click are both trying to intercept output.

Does #1037 go deeper than I thought? Does it cut down to our philosophy about whether fluff's output is _data_ or _logging_? 

Another perspective: we could implement my solution 1 via a logging method on the linter class (`linter.log(...)`) which includes conditional logic given the config or filename.

alanmcruickshank:
I think in principle, all logging going to `stderr` is a good idea. The only things that *need* to come out of `stdout` are when we return parsed files or structures.

Logging in `sqlfluff` is all a little convoluted however. If you check out `sqlfluff.cli.commands.py::set_logging_level` you'll see some of the machinery is already there. It's the function that handles all of the `cli` output, and I think that's where we should change the logic. If someone imports `sqlfluff` as a library, then they should have the flexibility to handle logging however they please.

This will probably need some tweeking to make sure all the tests still pass properly, but I think it's a good idea.

nolanbconaway:
Ah yes, thanks for the pointer @alanmcruickshank ! So this seems like its as easy as setting logging to stderr for the fix CLI whenever stdin is being used. I'll set that up and get the tests passing.

nolanbconaway:
Looks like tests are passing so this is ready for :eyes: ! I tested locally via the `check.py` script that was pasted in #1037. I didn't add that particular setup as a unit test because it seemed hyper specific, but would be happy to do so if this group thinks we should.

src/sqlfluff/cli/commands.py
```diff
@@ -421,7 +421,7 @@ def fix(force, paths, parallel, bench=False, fixed_suffix="", logger=None, **kwa
     formatter.dispatch_config(lnt)
 
     # Set up logging.
-    set_logging_level(verbosity=verbose, logger=logger)
+    set_logging_level(verbosity=verbose, logger=logger, stderr_output=fixing_stdin)
```
barrywhart, line 424: For consistency, should we consider making `lint` and `parse` behave the same as `fix`?

barrywhart (commented):
The changes look good. One question about possibly changing "lint" and "parse" as well.

src/sqlfluff/cli/commands.py
nolanbconaway: They're already set up for this! They both send logs to stderr if the output is meant to be machine readable (json, yaml), so that's already covered.

nolanbconaway:
thx all!

Issue #1062, opened 2021-05-16
PostgreSQL support for zero argument functions


bolajiwahab:
PostgreSQL supports zero-argument functions. An example:

````
CREATE OR REPLACE FUNCTION a() RETURNS integer
AS
$$
    SELECT 1;
$$
LANGUAGE SQL;
CREATE FUNCTION

````
From sqlfluff, we have the following:

````
echo 'CREATE OR REPLACE FUNCTION a() RETURNS integer AS $$ SELECT 1; $$ LANGUAGE SQL;' | sqlfluff lint --dialect postgres -    

== [stdin] FAIL
L:   1 | P:   1 |  PRS | Found unparsable section: 'CREATE OR REPLACE FUNCTION
                       | a() RETURNS i...'
````
While this is parseable

````
echo 'CREATE OR REPLACE FUNCTION a(text) RETURNS integer AS $$ SELECT 1; $$ LANGUAGE SQL;' | sqlfluff lint --dialect postgres -


````

bolajiwahab:
PR ~ https://github.com/sqlfluff/sqlfluff/pull/1063

PR #1063, opened 2021-05-16
pg/support_for_zero_argument_functions

bolajiwahab:


bolajiwahab:
#1063

bolajiwahab:
@barrywhart could you help review this?

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -102,6 +102,22 @@ class FunctionDefinitionGrammar(BaseSegment):
     )
 
 
+@postgres_dialect.segment(replace=True)
```
barrywhart, line 105: Is it possible we want to modify `FunctionParameterListGrammar` in the base (ANSI) dialect rather than making this Postgres-specific?

bolajiwahab: Yes sure, I will move it. I decided to make it postgres centric because I haven't worked with other dialect and not sure it they also support zero argument.

barrywhart: I'm not certain about other databases, but it seems very likely. Also, we tend to have a flexible, permissive approach to the ANSI dialect if it helps us keep the overall code simpler. I hope that makes sense! We may formalize and document this philosophy someday.

bolajiwahab: Yeah, it makes lot of sense. Since it's going to be flexible, I don't expect it to break the other databases.

bolajiwahab:
> * Rerun the `test/generate_parse_fixture_yml.py` script and see if it modifies that file. That would probably fix the test for you.
Indeed this fixed the tests. Thanks man @barrywhart

bolajiwahab:
PR to close issue https://github.com/sqlfluff/sqlfluff/issues/1062

bolajiwahab:
@alanmcruickshank I think we can merge this now?

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2572,10 +2572,13 @@ class FunctionParameterListGrammar(BaseSegment):
...
-            Ref("FunctionParameterGrammar"),
-            delimiter=Ref("CommaSegment"),
-        )
+        Sequence(
+            Delimited(
+                Ref("FunctionParameterGrammar"),
+                delimiter=Ref("CommaSegment"),
+            ),
+            optional=True,
+        ),
```
NiallRees, line 2581:
```suggestion
        Delimited(
            Ref("FunctionParameterGrammar"),
            delimiter=Ref("CommaSegment"),
            optional=True,
        ),
```

NiallRees (changes requested):
Just a small thing then we're good to merge 👍

src/sqlfluff/dialects/dialect_ansi.py
bolajiwahab: Now, I feel like I need to study the code base more, didn't know `delimited` support `optional`.

NiallRees (approved):

NiallRees:
Thank you @bolajiwahab!

Issue #962, opened 2021-04-13
Improve the performance of sqlfluff


dclong:
I know that there has been effort to improve the performance of sqlfluff. However, sqlfluff still runs extremely slow. For example, I have a SQL file (containing about 3000 lines of code after formatted using sqlfluff) and it takes about 1 hour to lint it and close to 3 hours to run `sqlfluff fix` on it.

barrywhart:
This is more of a question or suggestion than a solution, but have we tried running a big SQLFluff job using PyPy rather than CPython? For CPU-bound applications, PyPy is often 5-7x faster. I've used it on a handful of projects over the past several years. Sometimes there are compatibility issues, but that's gotten much better recently.

https://www.pypy.org/

dmateusp:
@dclong would it be possible for you to remove any sensitive information from that 3000 line SQL file and submit it to that issue? (I know it's a big ask, it's a lot of SQL!)

Then we could do some profiling using your query directly

alanmcruickshank:
Thanks @dclong . As options for what we do here:
- As suggested by @barrywhart , I've just checked and sqlfluff does run out of the box on pypy (using `pypy -m sqlfluff ...`). I haven't gone deep into functionality but everything looks good so far. On your large project I would definitely be interested on the performance gains you see there.
- In the last release (`0.5.2`) there were some logging improvements to see explicitly where the timing was going, especially for the dbt templater. Can you run the command you normally run but with `-vv` on the end to see the how long the compilation step takes?
- If neither of these are the issue then I have a feeling that this is going to be the `rule` evaluation process. I've noticed this on our codebase too which has a log of sql, and while it doesn't take as long as yours, it certainly takes a long time. I think it might be worth us re-evaluating the way that rules works to enable them to work faster. In particular at the moment, the application of rules is really expensive at the moment in that they're called on *every* segment in the parsed structure, and the rule then skips the segment if it's not relevant. There are definitely options to improve this.
- To potentially confirm this, when you run `fix` can you add `--bench` to the command which should output a section at the end with a readout of where sqlfluff was spending it's time. Unfortunately most of the benchmarking code is focussed on the _parsing_ elements of the process and so we might not get much granularity on the rules element if that is the culprit.
- For the very large SQL file, if you're not comfortable posting it on the issue here, I'm happy for you to DM me on the sqlfluff slack or email it to me for me to have a look at and unpick where the performance issues are.

dclong:
[...]

PR #1064, opened 2021-05-17
Bracket Matching Refactor

### LINKED ISSUE #1046: Performance: Bracket Matching
sqlfluff 0.5.3

When parsing and linting very long files (10k chars+), in particular one that I'm working with that has many many brackets, an unreasonable amount of time is spent in the bracket matching routines. In particular, within the `_bracket_sensitive_look_ahead_match` of `Delimited` grammars. I think there could be a significant speedup on large files by addressing this.

alanmcruickshank:
Another biggie. Resolves #1046, relates to #962.

This took the linting time of out large dbt project from 16:37 to 7:30. That's over 50% so I'm pretty pleased.

The bottom line of what I've done is make bracketed elements their own segment and then made sure that in most cases, even where matching isn't successful, that this parsing is persisted. This has the following effects:
- The long parsing steps at the start of any file where we're mapping out the overall structure are *much* faster. This is because we only do the bracket matching once, and then later searches through the file (using the bracket sensitive matching) can go much faster because they reuse the previous markings around bracketed regions. This is the core idea of this PR.
- It changes a lot of the dialect steps around brackets. Unfortunately this changes *a lot* of the test cases, grammar routines and rules. These all follow from the first point above, but it makes this a big PR.
- I've added a few small things at the same time while I was in those places in the codesbase (adding python implementation to the config output, updating parts of the Exasol dialect to use `OptionallyBracketed`, basic `pyproject.toml` file, moving some dialect tests into their own `.sql` files rather than being in the python test cases).


barrywhart:
So much chonk, ha ha! 🐈 😹  I will take a look.

alanmcruickshank:
> So much chonk, ha ha! 🐈 😹 I will take a look.

Yeah sorry.

With this one I wanted to make sure it would actually deliver the performance gains I thought it would before putting a PR together. Unfortunately, because bracket matching is so far down in the stack, it involved a lot of refactoring to get to that point.

src/sqlfluff/core/parser/grammar/sequence.py
```diff
@@ -245,48 +251,57 @@ def match(
...
         end_bracket = self.end_bracket or end_bracket
 
-        # Look for the first bracket
-        with parse_context.deeper_match() as ctx:
-            start_match = start_bracket.match(seg_buff, parse_context=ctx)
-        if start_match:
-            seg_buff = start_match.unmatched_segments
+        # Are we dealing with a pre-existing BracketSegment?
+        if seg_buff[0].is_type("bracketed"):
+            seg: BracketedSegment = seg_buff[0]  # type: ignore
```
barrywhart, line 264:
```suggestion
            seg: BracketedSegment = cast(BracketedSegment, seg_buff[0])
```

src/sqlfluff/core/parser/grammar/sequence.py
```diff
@@ -295,43 +310,59 @@ def match(
             pre_nc = ()
             post_nc = ()
 
-        # If we don't have anything left after trimming, act accordingly.
+        # If we've got a case of empty brackets check whether that is allowed.
         if not content_segs:
             if not self._elements or (
-                all(e.is_optional() for e in self._elements) and self.allow_gaps
+                all(e.is_optional() for e in self._elements)
+                and (self.allow_gaps or (not pre_nc and not post_nc))
```
barrywhart, line 317: I should probably know this, but what's `nc`?

src/sqlfluff/core/parser/grammar/sequence.py
```diff
@@ -295,43 +310,59 @@ def match(
...
             content_match = super().match(content_segs, parse_context=ctx)
 
         # We require a complete match for the content (hopefully for obvious reasons)
         if content_match.is_complete():
+            # Reconstruct the bracket segment post match.
+            # We need to realign the meta segments so the pos markers are correct.
+            # Have we already got indents?
+            meta_idx = None
+            for idx, seg in enumerate(bracket_segment.segments):
+                if seg.is_meta and seg.indent_val > 0:  # type: ignore
```
barrywhart, line 339:
```suggestion
                if seg.is_meta and cast(MetaSegment, seg).indent_val > 0:
```

src/sqlfluff/core/parser/grammar/sequence.py
```diff
@@ -295,43 +310,59 @@ def match(
...
+                    meta_idx = idx
+                    break
+            # If we've already go indents, don't add more.
```
barrywhart, line 342:
```suggestion
            # If we've already got indents, don't add more.
```

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1033,6 +1037,53 @@ def iter_patches(self, templated_str: str) -> Iterator[FixPatch]:
...
+class BracketedSegment(BaseSegment):
+    """A segment containing a bracketed expression."""
+
+    type = "bracketed"
+    additional_kwargs = ["start_bracket", "end_bracket"]
+
+    def __init__(
+        self,
+        *args,
+        start_bracket: Tuple[BaseSegment] = None,
```
barrywhart, line 1049: Why is it a tuple rather than just a `BaseSegment`?

src/sqlfluff/core/parser/segments/ephemeral.py
```diff
@@ -67,17 +69,22 @@ def wrapped_match_method(self, segments: tuple, parse_context):
...
-            return MatchResult.from_matched(
-                (
-                    EphemeralSegment(
-                        segments=segments,
-                        pos_marker=None,
-                        # Ephemeral segments get a copy of the parent grammar.
-                        parse_grammar=new_grammar,
-                        name=self.ephemeral_name,
-                    ),
+            # We shouldn't allow nested ephemerals. If they're present. Don't create another.
```
barrywhart, line 72:
```suggestion
            # We shouldn't allow nested ephemerals. If they're present, don't create another.
```

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -154,12 +155,16 @@
...
+# NOTE: The `persists` value controls whether this type
+# of bracket is persisted during matching to speed up other
+# parts of the matching process. Round brackets are the most
+# common and match the largest areas and so are sufficient.
 ansi_dialect.sets("bracket_pairs").update(
     [
-        ("round", "StartBracketSegment", "EndBracketSegment"),
-        ("square", "StartSquareBracketSegment", "EndSquareBracketSegment"),
-        ("curly", "StartCurlyBracketSegment", "EndCurlyBracketSegment"),
+        ("round", "StartBracketSegment", "EndBracketSegment", True),
```
barrywhart, line 165: Nitpick: The comment above is helpful. Maybe someday use `typing.NamedTuple` as well, to make the code more self documenting?

src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -335,9 +335,7 @@ class GroupByClauseSegment(BaseSegment):
                 Ref("ExpressionSegment"),
                 Ref("CubeRollupClauseSegment"),
                 Ref("GroupingSetsClauseSegment"),
-                Sequence(
-                    Ref("StartBracketSegment"), Ref("EndBracketSegment")
-                ),  # () possible
+                Bracketed(),  # () possible
```
barrywhart, line 338: The `() possible` comment is a bit cryptic. Does this mean "Empty parentheses are possible"?

src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -396,9 +394,7 @@ class GroupingSetsClauseSegment(BaseSegment):
             Delimited(
                 Ref("CubeRollupClauseSegment"),
                 Ref("GroupingExpressionList"),
-                Sequence(
-                    Ref("StartBracketSegment"), Ref("EndBracketSegment")
-                ),  # () possible
+                Bracketed(),  # () possible
```
barrywhart, line 397: Ditto

src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -1000,9 +994,7 @@ class AlterTableAddColumnSegment(BaseSegment):
         "ADD",
         Ref.keyword("COLUMN", optional=True),
         Ref("IfNotExistsGrammar", optional=True),
-        Ref("StartBracketSegment", optional=True),
```
barrywhart, line 1003: Heh, this old code was a little scary -- allowing unmatched brackets!!

src/sqlfluff/rules/L018.py
```diff
@@ -81,13 +81,18 @@ def indent_size_up_to(segs):
 
             balance = 0
             with_indent, with_indent_str = indent_size_up_to(raw_stack_buff)
-            for seg in segment.iter_segments(expanding=["common_table_expression"]):
+            print("SEG", segment, with_indent, repr(with_indent_str))
```
barrywhart, line 84: Looks like debugging code -- should this `print()` call be removed?

src/sqlfluff/rules/L018.py
```diff
@@ -81,13 +81,18 @@ def indent_size_up_to(segs):
...
+            for seg in segment.iter_segments(
+                expanding=["common_table_expression", "bracketed"], pass_through=True
+            ):
+                print(seg)
```
barrywhart, line 88: Ditto

src/sqlfluff/rules/L018.py
```diff
@@ -81,13 +81,18 @@ def indent_size_up_to(segs):
...
                 if seg.name == "start_bracket":
                     balance += 1
                 elif seg.name == "end_bracket":
                     balance -= 1
                     if balance == 0:
                         closing_bracket_indent, _ = indent_size_up_to(raw_stack_buff)
+                        print("INDENT", closing_bracket_indent)
```
barrywhart, line 95: Ditto

test/dialects/snowflake_test.py
```diff
@@ -11,86 +11,10 @@
 @pytest.mark.parametrize(
     "segment_cls,raw",
     [
-        ("UseStatementSegment", 'USE ROLE "MY_ROLE";'),
```
barrywhart, line 14: Why remove these test cases?

barrywhart (changes requested):

barrywhart:
@alanmcruickshank: Re "chonk", I am just joking. Great work on all the things recently! And farther back also, of course. 🙏

test/dialects/snowflake_test.py
alanmcruickshank: I moved them to a file based test rather than being here. As per the note a few lines above, the `.sql`/`.yml` tests are more robust than these ones, unless they're testing parts of segments rather than a full file.

src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -537,11 +552,32 @@ def _bracket_sensitive_look_ahead_match(
                             bracket_types_match = start_index == end_index
                             if bracket_types_match:
                                 # Yes, the types match. So we've found a
-                                # matching end bracket. Pop the stack and carry
+                                # matching end bracket. Pop the stack, construct
+                                # a bracketed segment and carry
                                 # on.
+
+                                # Complete the bracketed info
+                                bracket_stack[-1].segments += (
```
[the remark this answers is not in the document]
alanmcruickshank: I've been through the ones in this function and they're _similar_ but not the same. I think in this case I'll leave it as is.

src/sqlfluff/core/parser/grammar/sequence.py
alanmcruickshank: ah, _non code_. I'll add some comments

alanmcruickshank: Actually.... I'll just rename them to `pre_segs` and `post_segs` 👍

src/sqlfluff/dialects/dialect_ansi.py
alanmcruickshank: Agreed. I'm going to leave it for now, but if it gets any more complicated we probably should.

src/sqlfluff/dialects/dialect_exasol.py
alanmcruickshank: It does. I'm going to change it. 👍

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: Good question. I did this initially so that `tuple` arithmetic was easy, but these should always be single segments 🤔. For someone reading the code raw, do you think `BaseSegment` would make more sense? If so I make change it to do that instead.

src/sqlfluff/core/parser/grammar/delimited.py
```diff
@@ -184,13 +179,13 @@ def match(
                 # First check we're had enough delimiters, because if we haven't then
                 # there's no sense to try matching
                 if self.min_delimiters and len(delimiters) < self.min_delimiters:
-                    return MatchResult.from_unmatched(segments)
+                    return MatchResult.from_unmatched(mutated_segments)
```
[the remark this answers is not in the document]
alanmcruickshank: Good call - I'll see if I can make a test.

src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -604,16 +641,26 @@ def _bracket_sensitive_look_ahead_match(
                                 v_level=3,
                                 got=matcher,
                             )
-                            return ((), MatchResult.from_unmatched(segments), None)
+                            # We return with the mutated segments so we can reuse any bracket matching.
+                            return (
+                                (),
```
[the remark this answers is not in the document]
alanmcruickshank: Good call - I'll see if I can make a test.

src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -604,16 +641,26 @@ def _bracket_sensitive_look_ahead_match(
...
                             raise NotImplementedError(
                                 "This shouldn't happen. Panic in _bracket_sensitive_look_ahead_match."
                             )
                     else:
                         # Not in a bracket stack, but no match. This is a happy
-                        # unmatched exit.
-                        return ((), MatchResult.from_unmatched(segments), None)
+                        # unmatched exit. We return with the mutated segments so we can
+                        # reuse any bracket matching.
+                        return (
```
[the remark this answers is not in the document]
alanmcruickshank: Good shout - I'll see what I can do...

src/sqlfluff/core/parser/segments/base.py
barrywhart: I'm torn, because the syntax for creating a tuple "on the fly" is really obtuse and easy to overlook. Let's stick with the current code, but maybe add a clarifying comment that there'll be only one?

src/sqlfluff/core/parser/grammar/base.py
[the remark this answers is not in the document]
alanmcruickshank: Yep - I've managed to combine the _three_ instances of this code into a single code path. Great idea! 🎉

alanmcruickshank:
@barrywhart - I think I've addressed all of your feedback points, could I get another review?

Great idea on getting some better test coverage for some of the paths. I've added a few extra test cases, updated comments and rationalized the exit points from `_bracket_sensitive_look_ahead_match`. I've also added some type hints to `_bracket_sensitive_look_ahead_match` because I realise it was otherwise pretty opaque.

I'm going to be AFK for a few days, if you like the changes, would you mind merging?

barrywhart:
Sure! I'll do a final review tomorrow.

barrywhart (approved):

PR #1071, opened 2021-05-18
Upgrade syntax to Python 3.6+ using pyupgrade

cclauss:
`pyupgrade --py36-plus **/*.py` was used to generate all of these changes.
https://github.com/asottile/pyupgrade

This PR touches many files which may make it too difficult to review so it could be broken into multiple PRs to ease review.


cclauss:
The use of f-strings instead of format should speed up the _bench_ tests.  Does it?

pwildenhain (approved):
> The use of f-strings instead of format should speed up the bench tests. Does it?

Based off recent bench tests, it doesn't seem like it made an impact. Nice to get new syntax though. Even nicer to get it for free.

Looks like we've got:
- More f-strings ✔️ 
- Trailing commas ✔️ 
- Better use of `super.__init__()` ✔️ 
- Better use of `open()` default mode ✔️ 
- Better `set` creation ✔️ 
- Better use of newer standard lib ✔️ 

🤔 Interesting how it decided whether or not to use `str.format()` vs. f-strings but I'm sure there's a good heuristic I'm unaware of.

Can you fix the conflicts? Then you're good to merge 👍🏻 
Thanks so much!

NiallRees:
This is💯

NiallRees:
@cclauss is there any chance you could run this again? There have been a lot of internal changes and the merge conflicts are best resolved by re-running the command I expected.

PR #1072, opened 2021-05-18
flake8 ignore=E501 and --max-line-length are mutually exclusive

cclauss:
These two settings are about the same thing: The length of long lines.

Setting `max-line-length = 196` avoids all flake8 E501 errors and allows us to:
1. Flag attempts to add of even longer lines
2. Wrap long lines and lower max-line-length toward the recommended 88 chars per line.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -65,7 +65,7 @@ commands = mypy src/sqlfluff
 # D105: Don't require docstrings on magic methods
 ignore = W503, D107, D105, D418
 exclude = .git,__pycache__,env,.tox,build,.venv,venv
-max-line-length = 88
+max-line-length = 154
 extend-ignore =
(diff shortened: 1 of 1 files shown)


alanmcruickshank (approved):
Great idea. I'd quite like to make it a little shorter than that, but I realize that involves actually changing some of the offending bits of the codebase (so I'm happy to merge this PR first). Would you be interested in helping tackle some of our longer lines and maybe even bring us down to [PEP8 compliant](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) eventually?

- 196 covers everything already.
- 156 would require changing 5 lines.
- 128 would require changing 17 lines.
- 99 would require changing 88 lines.

99 feels like an eventually achievable goal, even if we take a few steps to get there 🤷 
If 196 currently covers everything

PR #1073, opened 2021-05-18
Update Getting Started Documentation

linchun3:
The existing documentation's example is outdated due to the addition of new rules. This change uses the same SQL example and adds in the following rules detected:
1. L034
2. L036
3. L039

EDIT:
The change has also been updated to reflect the optional use of --rules in `sqlfluff fix `.


docs/source/gettingstarted.rst
```diff
@@ -125,18 +135,23 @@ it's a good place to start. Another thing to note is that when fixing,
 you must always be specific about which rules you wish to fix. This
 is to minimise any unintended consequences from making large scale
 changes to your code. In this case we want to try and fix rules
-*L003*, *L009* and *L010*.
+*L003*, *L009*, *L010*, *L034*, *L036* and *L039*.
 
 .. code-block:: bash
 
-    $ sqlfluff fix test.sql --rules L003,L009,L010
+    $ sqlfluff fix test.sql --rules L003,L009,L010,L034,L036,L039
```
alanmcruickshank, line 142: The `--rules` argument is no longer compulsory for `fix`. I wonder if we should either:
- remove the reference to `--rules` here, and instead only demo it in the later section (line 219).
- adjust the narrative here so we don't fix *all* of the rules but intentionally only choose a subset to fix. In that case I think the later example _shouldn't_ use `--rules` just to show that it's not necessary anymore.

@linchun3 - which do you think is more helpful for a new user reading this?

alanmcruickshank (changes requested):
Great edits, thanks for taking the time to put this together.

I've got one question about how we demo the `--rules` argument to `fix`. It used to be _required_ and we changed that more recently, I think we should update this getting started guide accordingly, but I'm not sure what's most helpful. There are two places this guide references it and I feel like one should specify `--rules` and the other should not, to show that it's optional. I'm open to suggestions on which would be clearest however. 

What do you think?

docs/source/gettingstarted.rst
linchun3: hey @alanmcruickshank thanks for raising this up! I think the 2nd approach of adjusting the narrative (fixing subset followed by all) would be clearer to the reader that `--rules` is optional. I'll update this PR with that.

linchun3: Updated the narrative for the documentation to indicate optional use of `--rules`

docs/source/gettingstarted.rst
```diff
@@ -105,12 +110,17 @@ looks like this:
     c AS bar from my_table
 
 Rerun the same command as before, and you'll see that the original
-problems now no longer show up.
+error (violation of *L006*) no longer show up.
```
NiallRees, line 113:
```suggestion
error (violation of *L006*) no longer shows up.
```

NiallRees (approved):

PR #1075, opened 2021-05-18
Better simple API parsing exceptions

### LINKED ISSUE #569: Simple API raises the first parsing violation instead of "all of the violations"
This came up in https://github.com/sqlfluff/sqlfluff/pull/541/files#r526472027

Where the "simple api" `parse(` function only raises the first violation thrown by the function.

To avoid throwing information away, we could define an Exception which takes a list of violations so the API consumer can use the validation errors.

As a reference, in Scala cats there is this structure called "Validated" which accumulates errors https://typelevel.org/cats/datatypes/validated.html

alanmcruickshank:
This resolves #569 (an oldie), it's also a good one to fit into a major release if we've got 0.6.0 coming up, because it could break code which depends on the simple api.

Exactly as suggested by @dmateusp - this groups together any parsing exceptions and prints them nicely. Here's an example traceback if running in the console:

```
>>> import sqlfluff
>>> sqlfluff.parse("Select (1 + 2 +++) FROM mytable as blah blah")
  File "<stdin>", line 1, in <module>
  File "c:\...\sqlfluff\src\sqlfluff\api\simple.py", line 89, in parse
    raise APIParsingError(parsed.violations)
sqlfluff.api.simple.APIParsingError: Found 2 issues while parsing string.
Line 1, Position 14: Found unparsable section: ' +++'
Line 1, Position 41: Found unparsable section: 'blah'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Contributors:
 - [@GitHub-Username](Link to GitHub profile) ([#PR-Number](Link to PR))
 
+### Changed
+- Better exception handling for the simple parsing API (`sqlfluff.parse`)
+  which now raises an exception which holds all potential parsing issues
+  and prints nicely with more than one issue.
+
 ## [0.6.0a1] - 2021-05-15
 ### Added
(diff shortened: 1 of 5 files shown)


dmateusp (approved):
Nice @alanmcruickshank :+1:

PR #1089, opened 2021-05-21
Fix L016 false positive on templated string

### LINKED ISSUE #845: L016 reports "Line is too long" AFTER template expansion
## Expected Behaviour

No lint warnings.

## Observed Behaviour

Note the position (122). This line is only 53 characters long, pre-templating.
```
sqlfluff lint test.sql
== [test.sql] FAIL
L:   3 | P: 122 | L016 | Line is too long
```

## Version
SQLFluff version 0.4.1
Python version 3.7.3

## Configuration
```
[sqlfluff]
verbose = 0
nocolor = False
dialect = bigquery
templater = jinja
rules = None
recurse = 0

table_at_job_start=FOR SYSTEM_TIME AS OF CAST('2021-03-02T01:22:59+00:00' AS TIMESTAMP)
bi_ecommerce_orders=mc-business-intelligence.user.ecommerce_orders
```

Query:
```
SELECT user_id
FROM
    `{{bi_ecommerce_orders}}` {{table_at_job_start}}
```

pwildenhain:
(Eventually) closes #845 

🎗️ Remember to update `CHANGELOG.md` before merging

### DIFF
--- a/test/core/linter_test.py
+++ b/test/core/linter_test.py
@@ -303,6 +303,7 @@ def test__linter__empty_file():
                 ("L006", 3, 16),
                 ("L006", 3, 39),
                 ("L006", 3, 39),
+                ("L016", 4, 32),
             ],
         ),
     ],
@@ -312,7 +313,7 @@ def test__linter__mask_templated_violations(ignore_templated_areas, check_tuples
     lntr = Linter(
         config=FluffConfig(
             overrides={
-                "rules": "L006",
+                "rules": "L006,L016",
                 "ignore_templated_areas": ignore_templated_areas,
             }
         )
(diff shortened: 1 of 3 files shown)


pwildenhain:
📢 @barrywhart @alanmcruickshank 
I've narrowed down the issue to this algorithm:
https://github.com/sqlfluff/sqlfluff/blob/850758d93a1027d6639fb173e003cb85ff50b0d8/src/sqlfluff/core/templaters/base.py#L327-L348

With the change I added to the test, the `source_slice.start` of the `very_long_table_name` templated slice is 107, the `source_slice.stop` is 108, and the `seg_idx` is 117.

So in the main `for` loop, even though this slice has the correct `slice_type='templated`, we never get to the `else` clause to evaluate it, because it breaks out of the loop when it hits the `elif` clause of `seg_idx >= source_slice.stop`.

Where should we go from here? The obvious answer seems to be "just rely on the `slice_type`", but I'm guessing if it were that simple, we would've done it already 😝

tunetheweb:
Can this be closed now as resolved by #1411

Issue #1088, opened 2021-05-21
Dbt templater re-runs compile for each file while running parallel


alanmcruickshank:
While debugging slow running using the new `--parallel` feature, I've found that the dbt templater is running the compile step again for each file.

I *think* this is because each file uses it's own `Linter` object and so doesn't utilise the cached methods on the dbt templater.

The solution to this I think is to run the templating step in the main thread before dispatching out for parsing and linting. This should be possible by modifying the runner classes and the method calls available on `Linter`. The tricky part will be splitting out the templating step from `Linter.lint_string()` and `Linter.parse_string()`.

@barrywhart , having thought about this today, I was going to suggest this:
- We *don't* create a new `Linter` for each file.
- Instead, we abstract the parsing and linting methods out into `staticmethods` which can be safely called by the threads. Behind the existing methods of `parse_string()` and `lint_string()` there will be private methods of `_render_string()`, `_parse()` and `_lint()` which are static methods and handle all of the meaningful logic. These can be called via the existing methods _or directly from the runner classes_.
- In the parallel runners, the templating step (`_render_string()`) is called in the main thread, and then the resulting `TemplatedFile` in handed off to a child process to parse and lint (`_parse()` and `_lint()`). Those child processes never need their own `Linter()` object because they're just using the static methods, but might need a copy of the `FluffConfig` or similar objects.

Reckon that could work? I'm happy to have a go at this, I should probably get more familiar with the parallel code as is.

barrywhart:
That sounds like a good plan. I like the design of using static methods in the worker processes because it limits the number of object instances being passed to the workers. With `multiprocessing.Pool`, I've seen things get pretty confusing if you call normal instance methods from the workers, because Python serializes those objects automatically, but of course in the child it's not the same object, etc. etc. Far better to pass entire objects that you deliberately want to pass.

@alanmcruickshank: Go for it! I appreciate the help. Suddenly, work has gotten pretty busy for me with new hires, etc. etc.

PR #1092, opened 2021-05-23
Refactor Linter (pre parallel work).

alanmcruickshank:
This PR introduces no new functionality, it's just a reorganization. The original `sqlfluff.linter` module was almost 1.5k lines of code so it felt like that needed to happen. It supports refactoring as to help with #1088. I'm about to do some deeper reworking to the `Linter` class but this cleanup would have made that hard to review.

This PR:
- breaks up the existing `sqlfluff.core.linter` class into separate files.
- breaks up the `Linter.parse_string()` method into smaller pieces.


src/sqlfluff/core/config.py
```diff
@@ -547,3 +547,11 @@ def process_inline_config(self, config_line: str):
         config_path = [elem.strip() for elem in config_line.split(":")]
         # Set the value
         self.set_value(config_path[:-1], config_path[-1])
+
+    def process_raw_file_for_config(self, raw_str: str):
+        """Process a full raw file for inline config and update self."""
+        # Scan the raw file for config commands.
+        for raw_line in raw_str.splitlines():
+            if raw_line.startswith("-- sqlfluff"):
```
barrywhart, line 555: I didn't know about this!

Complete tangent here...

On Friday, I was talking to someone on Slack who wanted MySQL to support multiple kinds of delimiters between MySQL statements (apparently `;` is common, but people can change it). I suggested he use `OneOf()` to set up a fixed set of delimiters, but this makes me wonder about whether it'd be easy to change delimiters in the middle of a file.

src/sqlfluff/core/linter/common.py
```diff
@@ -0,0 +1,67 @@
...
+
+class NoQaDirective(NamedTuple):
+    """Parsed version of a 'noqa' comment."""
+
+    line_no: int  # Source line number
+    rules: Optional[Tuple[str, ...]]  # Affected rule names
+    action: Optional[str]  # "enable", "disable", or "None"
+
+
+class ProtoFile(NamedTuple):
```
barrywhart, line 32: What does "proto" mean here? It's almost a file but not a file just yet? 🤔

src/sqlfluff/core/linter/common.py
```diff
@@ -0,0 +1,67 @@
...
+class EnrichedFixPatch(NamedTuple):
+    """An edit patch for a source file."""
+
+    source_slice: slice
+    templated_slice: slice
+    fixed_raw: str
+    # The patch type, functions mostly for debugging and explanation
+    # than for function. It allows traceability of *why* this patch was
+    # generated.
+    patch_type: str
```
barrywhart, line 61: Nitpick: The term "type" suggests something fairly formal, like a programming language type. Could another word be more accurate? E.g. description?

src/sqlfluff/core/linter/linted_file.py
```diff
@@ -0,0 +1,491 @@
...
+        """Get a list of violations, respecting filters and ignore options.
+
+        Optionally now with filters.
+        """
+        violations = self.violations
+        # Filter types
+        if types:
+            try:
+                types = tuple(types)
+            except TypeError:
```
barrywhart, line 82: What input are we expecting that would cause a `TypeError` as we are handling here?

src/sqlfluff/core/linter/linted_file.py
```diff
@@ -0,0 +1,491 @@
...
+        dedupe_buffer = []
+        # We use enumerate so that we get an index for each patch. This is entirely
+        # so when debugging logs we can find a given patch again!
+        patch: Union[EnrichedFixPatch, FixPatch]
+        for idx, patch in enumerate(
+            self.tree.iter_patches(templated_str=self.templated_file.templated_str)
+        ):
+            linter_logger.debug("  %s Yielded patch: %s", idx, patch)
+
+            # This next bit is ALL FOR LOGGING AND DEBUGGING
```
barrywhart, line 259: Nitpick: The need for this comment makes me wonder -- would it potentially be more readable to move all this code to a separate function?

src/sqlfluff/core/linter/linted_file.py
```diff
@@ -0,0 +1,491 @@
...
+                source_slice=source_slice,
+                templated_slice=patch.templated_slice,
+                patch_type=patch.patch_type,
+                fixed_raw=patch.fixed_raw,
+                templated_str=self.templated_file.templated_str[patch.templated_slice],
+                source_str=self.templated_file.source_str[source_slice],
+            )
+
+            # Deal with the easy case of only literals
+            if set(local_type_list) == {"literal"}:
```
barrywhart, line 325: Minor performance question (feel free to ignore -- I'm probably being obsessive, ha ha): How big is `local_type_list` likely to be? If it could be more than a few elements in size, we could end up materializing a large-ish temporary collection for not much reason.

I found a couple of web pages with some clever ways to check the same thing more efficiently.
https://stackoverflow.com/questions/3844801/check-if-all-elements-in-a-list-are-identical
https://thispointer.com/python-check-if-all-elements-in-a-list-are-same-or-matches-a-condition/

src/sqlfluff/core/linter/linted_path.py
```diff
@@ -0,0 +1,118 @@
+"""Defines the LintedPath class.
+
+This stors the idea of a collection of linted files at a single start path
```
barrywhart, line 3:
```suggestion
This stores the idea of a collection of linted files at a single start path

```

src/sqlfluff/core/linter/linted_path.py
```diff
@@ -0,0 +1,118 @@
...
+from sqlfluff.core.errors import (
+    CheckTuple,
+)
+from sqlfluff.core.parser.segments.base import BaseSegment
+
+from sqlfluff.core.linter.linted_file import LintedFile
+
+
+class LintedPath:
+    """A class to store the idea of a collection of linted files at a single start path."""
```
barrywhart, line 26: Nitpick: Would "directory" be a better term than "path" here?

https://unix.stackexchange.com/questions/131561/what-is-the-difference-between-path-and-directory/131585

src/sqlfluff/core/linter/linted_path.py
```diff
@@ -0,0 +1,118 @@
...
+        self.path: str = path
+
+    def add(self, file: LintedFile) -> None:
+        """Add a file to this path."""
+        self.files.append(file)
+
+    @overload
+    def check_tuples(self, by_path: Literal[False]) -> List[CheckTuple]:
+        """Return a List of CheckTuples when by_path is False."""
+        ...
```
barrywhart, line 39: I haven't seen this before -- what does the "..." do here?

src/sqlfluff/core/linter/linter.py
```diff
@@ -0,0 +1,760 @@
...
+                linter_logger.info("Found unparsable segment...")
+                linter_logger.info(unparsable.stringify())
+        return parsed, violations
+
+    @staticmethod
+    def _generate_short_fname(fname: Optional[str] = None):
+        # Handle nulls gracefully
+        if not fname:
+            return None
+        return fname.replace("\\", "/").split("/")[-1]
```
barrywhart, line 195: Could this be replaced by something from a standard Python module, e.g. `os.path.basename()` or `pathlib.Path.name`?

src/sqlfluff/core/linter/linter.py
```diff
@@ -0,0 +1,760 @@
...
+        self,
+        in_str: str,
+        fname: Optional[str] = None,
+        recurse: bool = True,
+        config: Optional[FluffConfig] = None,
+    ) -> ParsedString:
+        """Parse a string.
+
+        Returns:
+            `ParsedString` of (`parsed`, `violations`, `time_dict`, `templated_file`).
```
barrywhart, line 225: Now that `ParsedString` is defined as a class, could these comments be removed (or moved to `ParsedString` if not already documented there)?

src/sqlfluff/core/linter/linter.py
```diff
@@ -0,0 +1,760 @@
...
+        """
+        if not os.path.exists(path):
+            if ignore_non_existent_files:
+                return []
+            else:
+                raise IOError("Specified path does not exist")
+
+        # Files referred to exactly are also ignored if
+        # matched, but we warn the users when that happens
+        is_exact_file = not os.path.isdir(path)
```
barrywhart, line 577: Nitpick: Would it be correct (and clearer) to say:
```
is_exact_file = os.path.isfile(path)
```

src/sqlfluff/core/linter/linter.py
```diff
@@ -0,0 +1,760 @@
...
+                (
+                    os.path.dirname(ignore_file_path),
+                    None,
+                    # Only one possible file, since we only
+                    # have one "ignore file name"
+                    [os.path.basename(ignore_file_path)],
+                )
+                for ignore_file_path in ignore_file_paths
+            ]
+            path_walk: Union[
```
barrywhart, line 601: This type hint is fairly complex. Could we use aliases here to make it more readable?

https://docs.python.org/3/library/typing.html#type-aliases

>Type aliases are useful for simplifying complex type signatures. For example:...

src/sqlfluff/core/linter/linter.py
```diff
@@ -0,0 +1,760 @@
...
+        if self.formatter:
+            self.formatter.dispatch_path(path)
+        fnames = list(
+            self.paths_from_path(
+                path,
+                ignore_non_existent_files=ignore_non_existent_files,
+                ignore_files=ignore_files,
+            )
+        )
+        runner: runner_module.BaseRunner
```
barrywhart, line 691: After I recently wrote this section of code that selects the runner type, I wondered if I should move it to the file that implements the runner classes. I think that'd be a little tidier. Thoughts?

barrywhart (approved):
I noticed one typo in a comment, also several other low-priority questions or suggested changes. Overall, looks good.

src/sqlfluff/core/config.py
alanmcruickshank: Oh I didn't know about that - do you have links to any docs about it? I use mysql quite a lot and hadn't come across it.

src/sqlfluff/core/linter/common.py
alanmcruickshank: Great question - I _think_ so, but I might try and come up with a better name for this. I'm going to need to rework some of this anyway. 🤔

src/sqlfluff/core/linter/common.py
alanmcruickshank: It's a bit more formal than a description here, but I'll rename this to something like `_category`.

src/sqlfluff/core/linter/linted_file.py
alanmcruickshank: Great question, I think it's singular violation types. That code path is covered by tests, so it's being used, but I'll see if I can either refactor this - or at least add an explanatory comment

src/sqlfluff/core/linter/linted_file.py
alanmcruickshank: Yes 👍 . I just lifted and shifted this so far, but good idea.

src/sqlfluff/core/linter/linted_file.py
alanmcruickshank: It's almost always a small sample here. I'm going to leave it as is - but it's a good suggestion.

src/sqlfluff/core/linter/linter.py
alanmcruickshank: Yes - probably `os.path.basename()`. I'll do some testing first, but I think that's the best option.

src/sqlfluff/core/linter/linter.py
alanmcruickshank: Yes. Will do.

src/sqlfluff/core/linter/linter.py
alanmcruickshank: I _think_ so. Will try it at least 🤔

src/sqlfluff/core/linter/linter.py
alanmcruickshank: great idea.

src/sqlfluff/core/linter/linter.py
alanmcruickshank: Yes definitely. I was planning that as part of the second round of this PR. I might come back to this next rather than addressing it in this PR.

src/sqlfluff/core/linter/linted_path.py
alanmcruickshank: yes - good point.

src/sqlfluff/core/linter/linted_path.py
alanmcruickshank: This is for overloading. https://python.readthedocs.io/en/stable/library/typing.html#typing.overload

src/sqlfluff/core/config.py
barrywhart: Here's one: https://www.mysqltutorial.org/mysql-stored-procedure/mysql-delimiter/

src/sqlfluff/core/linter/linted_file.py
barrywhart: Heh, yeah, moving code around makes reviews tricky -- pre-existing code gets a fresh look. Don't stress too much about this. FWIW, some IDEs (such as PyCharm) have a "Refactor"/"Extract Method" menu choice that streamlines things like this. Very handy -- it's pretty smart about figuring out the parameters required by the function, etc.

src/sqlfluff/core/linter/linter.py
barrywhart: No problem. I'm really pushing the limits, asking you to clean up my messy code. 😁

src/sqlfluff/core/linter/linted_path.py
barrywhart: Got it -- thanks!

src/sqlfluff/core/config.py
alanmcruickshank: Oh wow. That is serious dark magic. I'm not sure how best to support those custom delimiters. We might have to have it as something that we knowingly don't support for a while.

src/sqlfluff/core/linter/common.py
alanmcruickshank: Turns out that this class is never referred to! Must be a chunk of Legacy code. I'm removing it.

src/sqlfluff/core/config.py
barrywhart: No worries -- he's working on a fixed set of typical delimiters, and may have a PR soon.

PR #1093, opened 2021-05-23
Update changelog!

nolanbconaway:
I've merged a couple PRs recently and forgot to update the changelog both times. This PR adds entries for both.

Maybe we ought to set up a check on PR to determine if changelog was updated? This is very easy to forget about.


barrywhart (approved):
A check could be helpful -- I forget all the time!

nolanbconaway:
Sounds good! I'll open an issue to track progress and will assign myself.

Issue #658, opened 2021-01-06
Document the intent of the ANSI dialect.


alanmcruickshank:
I think we should have it somewhere in the docs, maybe for contributors or inside Architecture, specified that the ANSI dialect is our base dialect and it hosts logic that is common across multiple languages, so it does not strictly represent the ANSI specification

_Originally posted by @dmateusp in https://github.com/sqlfluff/sqlfluff/pull/325#r550441316_

PR #1096, opened 2021-05-23
Architecture docs rewrite

### LINKED ISSUE #621: Improve architecture docs
In particular I find the parser section really hard to understand. It is the most complicated part by nature, and I think providing some examples of the difference between match grammar and parse grammar would be helpful. An example parse tree too I think could help. As a contributor this was (and still is) a really scary part of the codebase, and I think it could help us improve the frequency of contributions.

https://docs.sqlfluff.com/en/stable/architecture.html

NiallRees:
Resolves #621 and #658

### DIFF
--- a/docs/source/architecture.rst
+++ b/docs/source/architecture.rst
@@ -100,7 +100,7 @@ to keep in mind.
 
 - Grammars are contained in *dialects*, the root dialect being
   the *ansi* dialect. The ansi dialect is used to host logic common
-  to all dialects, and so does not necessary adhere to the formal
+  to all dialects, and so does not necessarily adhere to the formal
   ansi specification. Other SQL dialects inherit from the ansi dialect,
   replacing or patching any segments they need to. One reason for the *Ref*
   grammar is that it allows name resolution of grammar elements at runtime
--- a/docs/source/architecture.rst
+++ b/docs/source/architecture.rst
@@ -1,23 +1,29 @@
 Architecture
 ============
 
+It is recommended that the following is read in conjunction with exploring
+the codebase. `dialect_ansi.py` in particular is helpful to understand the
+recursive structure of segments and grammars.
+
 Stage 1, the templater
 ----------------------
 
-The templater takes raw files and optionally fills in any configurable
-sections before passing onto the linter. In particular most templating
-systems use curly brackets *{}* to indicate templatable sections and
-these are not currently set up in any of the lexers, and so will fail
-at the next step if not dealt with here.
+This stage only applies to templated SQL, most commonly Jinja and dbt. Vanilla
+SQL is sent straight to stage 2, the lexer.
 
-*SQLFluff* supports 2 templating engines:
+In order to lint templated SQL, SQLFluff must first convert the 'raw' or
+pre-templated code into valid SQL, which can then be parsed. The templater
+returns both the raw and post-templated SQL so that any rule violations which
+occur in templated sections can be ignored and the rest mapped to their
+original line location for user feedback.
 
 .. _Jinja: https://jinja.palletsprojects.com/
 .. _dbt: https://docs.getdbt.com/
 
-Jinja_ is the default templater used by *SQLFluff*. Under the covers, dbt_ also
-uses Jinja_ but in *SQLFluff* it is a separate templater which relies directly
-on the dbt_ compiler.
+*SQLFluff* supports two templating engines: Jinja_ and dbt_.
+
+Under the hood dbt also uses Jinja, but in *SQLFluff* uses a separate
+mechanism which interfaces directly with the dbt python package.
 
 For more details on how to configure the templater see :ref:`templateconfig`.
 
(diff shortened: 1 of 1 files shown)


docs/source/architecture.rst
```diff
@@ -1,69 +1,80 @@
...
 
 
 Stage 2, the lexer
 ------------------
 
-The lexer takes raw text and splits it into tokens. Nothing is *removed* but
-whitespace and code are separated. In principle all identifiers should be
-separated at this stage, but they should not be imparted any meaning at this
-stage. Any files which cannot be lexed, should raise a *SQLLexError*.
+The lexer takes valid SQL and separates it into segments of whitespace and
```
barrywhart, line 34: Nitpick: Saying "valid SQL" seems potentially misleading, since sometimes SQL is invalid. Maybe say "SQL" or "presumably valid SQL"?

docs/source/architecture.rst
```diff
@@ -1,69 +1,80 @@
...
+code. No meaning is imparted, that is the job of the parser.
```
barrywhart, line 35:
```suggestion
code. No meaning is imparted; that is the job of the parser.
```

docs/source/architecture.rst
```diff
@@ -1,69 +1,80 @@
...
+lifting.
+
+#. The lexed segments are parsed using the specified dialect's grammars. In
+   SQLFluff, grammars describe the shape of SQL statements (or their
+   components). The parser attempts to apply each potential grammar to the
+   lexed segments until all the segments have been matched.
+
+#. In SQLFluff, segments form a tree-like structure. The top-level segment is
+   a :code:`FileSegment`, which contains zero or more
+   :code:`StatementSegment`\ s, and so on. Before the segments have been parsed
```
barrywhart, line 52: Is the `\ s` deliberate? (I'm not a markdown expert, so I wasn't sure if this was correct or not.)

docs/source/architecture.rst
```diff
@@ -1,69 +1,80 @@
...
+   and named according to their type, they are are 'raw', meaning they have no
```
barrywhart, line 53:
```suggestion
   and named according to their type, they are 'raw', meaning they have no
```

docs/source/architecture.rst
```diff
@@ -1,69 +1,80 @@
...
+   classification other than their literal value.
+
+#. The three key components to the parser are segments, :code:`match_grammar`s
+   and :code:`parse_grammar`s. A segment can be a leaf in the parse tree, such
+   as a :code:`NumericLiteralSegment`, which is simply a number, or can contain
+   many other segments, such as a :code:`SelectStatementSegment`. Each segment
+   can specify a :code:`parse_grammar`, and a :code:`match_grammar`. The
+   :code:`match_grammar` is used to determine whether or not a given sequence
+   of segments is an instance of the segment in question. If defined, the
+   :code:`parse_grammar` is then used to match those segments further.
```
barrywhart, line 63: Think we could add more detail on the distinction between match and parse grammars, with perhaps a illustrative example or two? I have done a fair amount of work on the dialects and am still unsure how I would know that I need to use both instead of just one.

docs/source/architecture.rst
```diff
@@ -73,60 +84,40 @@ We recursively parse each of the elements, using their in built grammars.
       segments, and the *expand* step will let each of those segments refine
       the content within them.
 
-   #. Eventually in that recursive operation we reach segments which have no
-      children (raw elements containing a single token), and so the recursion
+   #. During the recursion, segments which have no children are eventually
```
barrywhart, line 87: Possible alternate wording:

During the recursion, the parser eventually reaches segments which have no children

docs/source/architecture.rst
```diff
@@ -73,60 +84,40 @@ We recursively parse each of the elements, using their in built grammars.
...
 
+
 Stage 4, the linter
 -------------------
 
-Given the complete parse tree, we now walk that tree to assess the tree
-for linting errors. A linter is an object which is able to traverse the
-tree itself, allowing it to choose which objects it examines and which
-it alerts as a problem.
+Given the complete parse tree, rule classes assess for linting errors by
```
barrywhart, line 114: Maybe replace "assess" with "check" or "look"?

docs/source/architecture.rst
```diff
@@ -73,60 +84,40 @@ We recursively parse each of the elements, using their in built grammars.
...
+traversing the tree, looking for segments and patterns of concern. If
+the rule discovers a violation, it returns a :code:`LintResult` pointing
+to the segment which has caused the violation.
```
barrywhart, line 117:
```suggestion
to the segment which caused the violation.
```

docs/source/architecture.rst
```diff
@@ -73,60 +84,40 @@ We recursively parse each of the elements, using their in built grammars.
...
 
-Some linters, may optionally be able to *fix* the problems they find. If
-this is the case, they will optionally return a mutated tree as one of
-their return values, which can be passed to the next linter. In normal
-operation this will not be what is returned, because it becomes confusing
-with line references for a user fixing issues manually.
+Some rules are able to *fix* the problems they find. If
+this is the case, they will return a mutated segment, containing
```
barrywhart, line 120: This doesn't seem quite accurate. Rules can return a list of `fixes`, which describe mutations (i.e. changes) to be made to the tree. This can include edits, inserts, or deletions to the tree.

barrywhart (changes requested):

docs/source/architecture.rst
NiallRees: Agreed, SQL it is!

docs/source/architecture.rst
NiallRees: Yep, from reading it seems that the escaped whitespace tells the doc compiler that the code segment has ended. I'm also not a markdown expert, but it works like this!

docs/source/architecture.rst
NiallRees: Quite right - thanks!

docs/source/architecture.rst
NiallRees: @alanmcruickshank can I rope you in here? My understanding is that match_grammars are used as a very quick way of matching a segment to tokens, vs the parse_grammar which takes longer. I'm sure there's more to it though.

alanmcruickshank: Think of the dialect like a really complicated tree diagram. Match grammars prune branches and choose which branch to follow. Parse grammars lay out which options will be available in the next round (kind of).

NiallRees: Cool ok that's what I had in mind, while I'm sure it's more nuanced I think that will suffice for the docs. Really the aim of this doc is that people can read it a few times with the source code in hand and contribute to dialects and rules.

NiallRees: And in the instances where there is only a match_grammar, it performs both jobs?

docs/source/architecture.rst
```diff
@@ -1,69 +1,84 @@
...
+   leaf in the parse tree, such as a :code:`NumericLiteralSegment`, which is
+   simply a number, or can contain many other segments, such as a
+   :code:`SelectStatementSegment`. Each segment can specify a
+   :code:`parse_grammar`, and a :code:`match_grammar`. If both a
+   :code:`match_grammar` and :code:`parse_grammar` are defined in a segment,
+   :code:`match_grammar` is used to quickly prune the tree for branches which
+   do not match segments being parsed, and the :code:`parse_grammar` is then
+   used to refine the branch identified as correct. If only a
+   :code:`match_grammar` is defined, then it serves the purpose of both pruning
+   and refining.
```
barrywhart, line 67: 💯

docs/source/architecture.rst
```diff
@@ -73,60 +88,39 @@ We recursively parse each of the elements, using their in built grammars.
...
+When working on the parser there are a couple of design principles
 to keep in mind.
 
-- The core of the grammar is stored in a *dialect*, the root dialect being
-  the *ansi* dialect. The intent here is that for other SQL dialects that
-  they will inherit most of the logic from a parent dialect and then just
-  reimplement the sections that they need to. One reason for the *Ref*
+- Grammars are contained in *dialects*, the root dialect being
+  the *ansi* dialect. The ansi dialect is used to host logic common
+  to all dialects, and so does not necessary adhere to the formal
```
barrywhart, line 103:
```suggestion
  to all dialects, and so does not necessarily adhere to the formal
```

docs/source/architecture.rst
```diff
@@ -73,60 +88,39 @@ We recursively parse each of the elements, using their in built grammars.
...
-  which the grammar will use for matching. These may in turn be a mixture
-  of segments and grammars. A segment can be used as either a class or an
-  instance of a class. While matching, the class itself is used, but if
-  it does match, rather than just returning the original segments it was
-  passed unchanged, it will optionally return mutated versions of those
-  segments which may have been given a more specific meaning by the
-  matcher. For example a *RawSegment* containing *123.4* may actually
-  be mutated to an instance of the *NumericLiteralSegment* class when
-  matched against the *NumericLiteralSegment* class.
+  lower level elements which haven't been redeclared within the dialect
```
barrywhart, line 108:
```suggestion
  lower-level elements which haven't been redeclared within the dialect
```

barrywhart (approved):
I noticed a couple of typos, otherwise good to go. Great work!!

NiallRees:
> I noticed a couple of typos, otherwise good to go. Great work!!

⭐

PR #1098, opened 2021-05-25
First attempt, non-working

silverbullettruck2001:
@barrywhart here's what I have so far.  I am not sure why it detected a change in the `bnech_002_pearson_fix` but it was not intentional, so I can revert that if its not correct.

### DIFF
--- a/src/sqlfluff/dialects/dialect_mysql.py
+++ b/src/sqlfluff/dialects/dialect_mysql.py
@@ -5,6 +5,8 @@ https://dev.mysql.com/doc/refman/8.0/en/differences-from-ansi.html
 """
 
 from sqlfluff.core.parser import (
+    BaseSegment,
+    Anything,
     Ref,
     AnyNumberOf,
     Sequence,
@@ -14,8 +16,13 @@ from sqlfluff.core.parser import (
     CommentSegment,
     NamedParser,
     CodeSegment,
+    StringParser,
+    SymbolSegment,
+    KeywordSegment,
+    Delimited,
 )
 from sqlfluff.core.dialects import load_raw_dialect
+from sqlfluff.core.parser.grammar.greedy import StartsWith
 
 ansi_dialect = load_raw_dialect("ansi")
 mysql_dialect = ansi_dialect.copy_as("mysql")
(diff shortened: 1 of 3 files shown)


barrywhart:
About `bench_002_pearson_fix`, I'm guessing somehow `sqlfluff fix` got run against that file. You'll want to undo that before this PR could be merged, but don't worry about it for now. I think your branch is behind the current `master`, as I mentioned before. (That can also wait.)

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -90,3 +97,49 @@ class CreateTableStatementSegment(
...
+
+mysql_dialect.replace(
+    DelimiterSegment=OneOf(Ref("SemicolonSegment"), Ref("TildeSegment")),
+    TildeSegment=StringParser(
+        "~", SymbolSegment, name="tilde", type="statement_terminator"
+    ), 
+)
+
+@mysql_dialect.segment()
+class FunctionDelimiterGrammar(BaseSegment):
```
barrywhart, line 118: I suggest omitting the `parse_grammar` for now -- a `parse_grammar` is sufficient in most cases. I notice you nested a `Sequence` inside a `Sequence`, which seems unnecessary, and the inner one is marked `optional`. I'm not sure about that. This is my best guess for now:
```
@mysql_dialect.segment()
class FunctionDelimiterGrammar(BaseSegment):
    parse_grammar = Sequence(
        "DELIMITER",
        OneOf(Ref("DelimiterSegment")),
        optional=True,
    )
```

I think the reason this isn't working is that it's not hooked into the rest of the grammar. When you add a new segment type, it doesn't really have any effect until it is mentioned in some other place as a potential option to be used in that place.

barrywhart: I think you'll need to modify the MySQL grammar to inherit from / add to `class StatementSegment` from the ANSI grammar. `StatementSegment` lists all the possible kinds of statements.

silverbullettruck2001: @barrywhart I updated the PR with your suggestions.  I had to add the `match_grammar` because when I took it out it was giving me an error of :
`AttributeError: type object 'DelimiterGrammar' has no attribute 'match_grammar'`
If I add the `match_grammar` back, then it doesn't error, but it also doesn't match and I still get the original error:
`L:   1 | P:   1 |  PRS | Found unparsable section: 'delimiter'`

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -90,3 +97,45 @@ class CreateTableStatementSegment(
...
+    ), 
+)
+
+@mysql_dialect.segment()
+class DelimiterGrammar(BaseSegment):
+    match_grammar = StartsWith("DELIMITER")
+    parse_grammar = Sequence(
+        "DELIMITER",
+        OneOf(Ref("DelimiterSegment")),
+        optional=True,
```
barrywhart, line 123: Is it necessary to specify `optional=True` here?

silverbullettruck2001: I did that because delimiter isn't always specified.  If you don't specify the `DELIMTER` keyword then it just assumes `;`

PR #1099, opened 2021-05-25
Add support for multiple delimiters and the `Delimiter` keyword

silverbullettruck2001:
Modified the ansi dialect to allow for multiple delimiters for languages that want to override this to add multiple delimiters.  Also, updated the mysql dialect to allow parsing of the `Delimiter` keyword.


silverbullettruck2001:
@barrywhart here's the PR with the updates that don't seem to be identifying the trailing `DELIMITER ;` and the ending `~` as expected.  Thoughts?

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -181,6 +178,7 @@
 
 ansi_dialect.add(
     # Real segments
+    DelimiterSegment=OneOf(Ref("SemicolonSegment")),
```
NiallRees, line 181:
```suggestion
    DelimiterSegment=Ref("SemicolonSegment"),
```

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1,11 +1,8 @@
 """The core ANSI dialect.
-
```
NiallRees, line 2: Please could you add all the line breaks back that seem to have gone?

NiallRees (changes requested):
Thanks for this @silverbullettruck2001 !

silverbullettruck2001 (commented):
This is done

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -90,3 +93,39 @@ class CreateTableStatementSegment(
...
+
+
+@mysql_dialect.segment()
+class DelimiterStatement(BaseSegment):
+    """DELIMITER statement."""
+
+    type = "delimiter_statement"
+    match_grammar = Sequence(
+        "DELIMITER",
+    )
```
NiallRees, line 122:
```suggestion
    match_grammar = Ref.keyword("DELIMITER")
```

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -62,7 +65,7 @@
         CodeSegment,
         name="quoted_literal",
         type="literal",
-        trim_chars=('"',),
+        trim_chars=('"'),
```
barrywhart, line 68: It looks like here, you changed a tuple that contains a string to just be a string. Was this change intentional? If not, can you change it back?

barrywhart (changes requested):
Looks good overall. I noticed one change that looks like it may have been accidental. Can you check and maybe undo that change?

src/sqlfluff/dialects/dialect_mysql.py
silverbullettruck2001: This has been fixed now

barrywhart (approved):

NiallRees (approved):

PR #1101, opened 2021-05-25
Rip out BenchIt and replace with TimingSummary

alanmcruickshank:
The procedural nature of `BenchIt` was causing issues with the changes I'd like to make to the `Linter` class to support #1088 . I asked on the slack thread and nobody voted to keep the exiting `BenchIt` implementation, but I think there's still value in profiling and benchmarking tools.

This PR removes the `BenchIt` library and replaces the functionality with a `TimingSummary` object which displays more useful descriptive stats about where time was spent during the operation.

### DIFF
--- a/src/sqlfluff/core/timing.py
+++ b/src/sqlfluff/core/timing.py
@@ -1,0 +1,40 @@
+"""Timing summary class."""
+
+from typing import Optional, List, Dict
+from collections import defaultdict
+
+
+class TimingSummary:
+    """An object for tracking the timing of similar steps across many files."""
+
+    def __init__(self, steps: Optional[List[str]] = None):
+        self.steps = steps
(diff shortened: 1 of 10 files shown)


src/sqlfluff/core/timing.py
```diff
@@ -0,0 +1,40 @@
...
+
+        for timing_dict in self._timings:
+            for step in self.steps:
+                if step in timing_dict:
+                    vals[step].append(timing_dict[step])
+        summary = {}
+        for step in self.steps:
+            if vals[step]:
+                summary[step] = {
+                    "cnt": len(vals[step]),
```
NiallRees, line 34: 👀

NiallRees (approved):

barrywhart (approved):

PR #1103, opened 2021-05-26
Extract runner selection logic

alanmcruickshank:
As suggested in #1101 and to support #1088, this extracts runner selection logic into the `runner` module.

### DIFF
--- a/src/sqlfluff/core/linter/runner.py
+++ b/src/sqlfluff/core/linter/runner.py
@@ -13,7 +13,7 @@ import multiprocessing.dummy
 import signal
 import sys
 import traceback
-from typing import Callable, List, Type
+from typing import Callable, List
 
 from sqlfluff.core.dialects import dialect_selector
 from sqlfluff.core.parser import Lexer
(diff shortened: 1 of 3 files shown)


src/sqlfluff/core/linter/runner.py
```diff
@@ -211,3 +204,28 @@ def __init__(self, ee):
...
+
+
+def get_runner(
+    linter,
+    config,
+    parallel: int,
+    allow_process_parallelism: bool = True,
+) -> BaseRunner:
+    """Generate a runner instance based on parallel and sytem configuration."""
+    if parallel > 1 and sys.version_info > (3, 7):
```
barrywhart, line 216: I noted this on the other PR as well -- I think the version check should be `sys.version_info > (3, 7)`. (Minor bug.)

barrywhart (approved):
One minor pre-existing bug I noticed in the code that's being moved. Would be cool to fix but not vital.

src/sqlfluff/core/linter/runner.py
alanmcruickshank: @barrywhart I'm confused. The check is _already_ `(3, 7)`, did you mean it should be something else? I'd also love to add a comment saying _why_ the earlier versions aren't supported - do you know what the issue is with them?

alanmcruickshank: Scratch that.... saw your other comment. Will fix it there.

PR #1106, opened 2021-05-26
Draft: Resolve parallel dbt templating issue

### LINKED ISSUE #1088: Dbt templater re-runs compile for each file while running parallel

alanmcruickshank:
This PR has a couple of things to still shake out. Namely that one test is failing for a reason I don't understand, and it looks like the CLI output is treading on toes when running in parallel. Those issues aside, I think it's ready for a review, at least from an architecture standpoint.

This resolves #1088 .

I've tested it on our large dbt project with the dbt templater and it's FAST.

It also builds on #1103 so will be easier to review once that's merged.

In short:
- We render the sql files in the main thread and then pass them to the child threads for parsing and linting.
- The rest of the changes are making enough methods in the linter _static_ or _class_ methods to do that in a way that we don't have to create a new `Linter` object each time.

### DIFF
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -366,6 +366,7 @@ def lint(
 
     if bench:
         click.echo("==== overall timings ====")
+        click.echo(cli_table([("Clock time", result.total_time)]))
         timing_summary = result.timing_summary()
         for step in timing_summary:
             click.echo(f"=== {step} ===")
@@ -516,6 +517,7 @@ def fix(force, paths, parallel, bench=False, fixed_suffix="", logger=None, **kwa
 
     if bench:
         click.echo("==== overall timings ====")
+        click.echo(cli_table([("Clock time", result.total_time)]))
         timing_summary = result.timing_summary()
         for step in timing_summary:
             click.echo(f"=== {step} ===")
@@ -633,6 +635,7 @@ def parse(path, code_only, format, profiler, bench, nofail, logger=None, **kwarg
                     click.echo(cli_table(parsed_string.time_dict.items()))
             if verbose >= 2 or bench:
                 click.echo("==== overall timings ====")
+                click.echo(cli_table([("Clock time", result.total_time)]))
                 timing_summary = timing.summary()
                 for step in timing_summary:
                     click.echo(f"=== {step} ===")
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -3,6 +3,7 @@
 import sys
 import json
 import logging
+import time
 
 import oyaml as yaml
 
@@ -598,6 +599,7 @@ def parse(path, code_only, format, profiler, bench, nofail, logger=None, **kwarg
         pr.enable()
 
     try:
+        t0 = time.monotonic()
         # handle stdin if specified via lone '-'
         if "-" == path:
             # put the parser result in a list to iterate later
@@ -609,6 +611,7 @@ def parse(path, code_only, format, profiler, bench, nofail, logger=None, **kwarg
         else:
             # A single path must be specified for this command
             result = lnt.parse_path(path, recurse=recurse)
+        total_time = time.monotonic() - t0
 
         # iterative print for human readout
         if format == "human":
@@ -635,7 +638,7 @@ def parse(path, code_only, format, profiler, bench, nofail, logger=None, **kwarg
                     click.echo(cli_table(parsed_string.time_dict.items()))
             if verbose >= 2 or bench:
                 click.echo("==== overall timings ====")
-                click.echo(cli_table([("Clock time", result.total_time)]))
+                click.echo(cli_table([("Clock time", total_time)]))
                 timing_summary = timing.summary()
                 for step in timing_summary:
                     click.echo(f"=== {step} ===")
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -648,7 +648,7 @@ def parse(path, code_only, format, profiler, bench, nofail, logger=None, **kwarg
                     if parsed
                     else None,
                 )
-                for filepath, (parsed, _, _, _, _) in zip(filepaths, result)
+                for filepath, (parsed, _, _, _, _, _) in zip(filepaths, result)
             ]
 
             if format == "yaml":
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -644,17 +644,16 @@ def parse(path, code_only, format, profiler, bench, nofail, logger=None, **kwarg
                     click.echo(f"=== {step} ===")
                     click.echo(cli_table(timing_summary[step].items()))
         else:
-            # collect result and print as single payload
-            # will need to zip in the file paths
-            filepaths = ["stdin"] if "-" == path else lnt.paths_from_path(path)
             result = [
                 dict(
-                    filepath=filepath,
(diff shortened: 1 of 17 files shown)


alanmcruickshank:
The outstanding test is `test__linter__skip_dbt_model_disabled`, which I'm wondering whether this is actually a better solution for. The existing test checks that a `LintedFile` is still returned but that it has a null `templated_file`. In this reworking, no `LintedFile` is returned at all. It's still effectively _skipped_, but skipped in a slightly different way. @NiallRees, @dmateusp  - do you have views on how this might play out for users?

src/sqlfluff/core/linter/runner.py
```diff
@@ -211,3 +207,28 @@ def __init__(self, ee):
...
+
+
+def get_runner(
+    linter,
+    config,
+    parallel: int,
+    allow_process_parallelism: bool = True,
+) -> BaseRunner:
+    """Generate a runner instance based on parallel and sytem configuration."""
+    if parallel > 1 and sys.version_info > (3, 7):
```
barrywhart, line 219: I'm thinking now that the version check should use `>=`, i.e.: `sys.version_info >= (3, 7)` because it's 3.6.x that is not supporting parallel.

barrywhart:
About the CLI output, could we potentially capture the output in the workers and return it to the main process and print it there? That's what I think I did with the linter output in the first parallel implementation.

src/sqlfluff/core/linter/runner.py
alanmcruickshank: I've updated the comparison. Do you have something I can link to and put in a comment about *why* 3.6 isn't supported. I had a quick google just now and couldn't work it out.

alanmcruickshank:
> About the CLI output, could we potentially capture the output in the workers and return it to the main process and print it there? That's what I think I did with the linter output in the first parallel implementation.

This is what I _intended_ to do here, but evidently didn't manage. I'll have another look.

src/sqlfluff/core/linter/runner.py
barrywhart: TBH, I don't remember. I know it wasn't working. I recall thinking we could possibly get it working, but 3.6 is so old, was it worth the effort.

I'll see if I can reproduce the error on 3.6 again.

barrywhart:
@alanmcruickshank: I replied to your question about the issue with 3.6. Mentioning it here in case that comment thread gets hidden by GitHub!!

dmateusp:
@alanmcruickshank I think not returning a LintedFile makes sense!

alanmcruickshank:
I think this is ready to review properly.

The additional changes are:
- Solved the issue with double logging of paths and threaded logging (@barrywhart - I'd just duplicated the code 🤦 - one was inside the thread).
- In debugging this I came across an odd failure with the `fatal` templating errors I added a while back. I've added a counter in the `dbt` templater so that they only actually trigger a fatal error if they happen 3 times in a row. This isn't really related to this PR, but it's a tiny change that was easy to make while I was in the code.
- File names are now explicit fields on `ParsedString` and `RenderedFile`. Previously we introspected these from the contained `TemplatedFile` but in the case of a templating error, this was `None` and so triggered some issues and odd failures.
- There's an additional logging line in the `dbt` templater for the _completion_ of templating so that if you're not running in high verbosity modes it doesn't just look like it's stuck on `templating...` for ages.

alanmcruickshank:
I realise the `--bench` command isn't that useful in a parallel world because the timings overlap. I've added a total `Clock time` output to the timing outputs so that we can evaluate performance better.

On 858 files using the dbt templater:
- `-p 1`: Clock time:   418.02
- `-p 4`: Clock time:   261.33

Not quite a 50% reduction, but I'm still pretty happy with that. 🎉

src/sqlfluff/core/linter/linter.py
```diff
@@ -455,13 +461,16 @@ def lint_parsed(
             # We're only going to return the *initial* errors, rather
             # than any generated during the fixing cycle.
             violations += initial_linting_errors
+        else:
+            # If no parsed tree, set to Nonw
```
barrywhart, line 465:
```suggestion
            # If no parsed tree, set to None

```

src/sqlfluff/core/templaters/dbt.py
```diff
@@ -199,12 +206,18 @@ def process(self, *, fname, in_str=None, config=None, formatter=None):
...
 
         try:
-            return self._unsafe_process(fname, in_str, config)
+            processed_result = self._unsafe_process(fname, in_str, config)
+            # Reset the fail counter
+            self._sequential_fails = 0
+            return processed_result
         except DbtCompilationException as e:
+            # Increment the counter
+            self._sequential_fails += 1
```
barrywhart, line 215: Why would compilation fail the first time but subsequently work?

barrywhart (approved):
A couple of questions, nothing huge. Looks good!

src/sqlfluff/core/templaters/dbt.py
alanmcruickshank: some compilation fails are fatal (the kind that initially inspired this), but I've realised some aren't. They are where the compilation fails in a file-specific way.

src/sqlfluff/cli/commands.py
```diff
@@ -648,7 +648,7 @@ def parse(path, code_only, format, profiler, bench, nofail, logger=None, **kwarg
                     if parsed
                     else None,
                 )
-                for filepath, (parsed, _, _, _, _) in zip(filepaths, result)
+                for filepath, (parsed, _, _, _, _, _) in zip(filepaths, result)
```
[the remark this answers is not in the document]
alanmcruickshank: Yeah great idea, I think this can be much simpler.

src/sqlfluff/cli/commands.py
```diff
@@ -644,17 +644,16 @@ def parse(path, code_only, format, profiler, bench, nofail, logger=None, **kwarg
...
-                    if parsed
+                    filepath=linted_result.fname,
+                    segments=linted_result.tree.as_record(
+                        code_only=code_only, show_raw=True
+                    )
+                    if linted_result.tree
                     else None,
                 )
-                for filepath, (parsed, _, _, _, _, _) in zip(filepaths, result)
+                for linted_result in result
```
barrywhart, line 656: 🎉

barrywhart (approved):
Very nice!!

PR #1112, opened 2021-05-28
Issue 1082: Add support for BigQuery "select as struct <<column list>>" syntax

### LINKED ISSUE #1082: BigQuery's "SELECT AS STRUCT" mangled by fix
With this query as a simple case (saved as `models/test.sql`):

```sql
select array(
  select as struct
    b as bb,
    a as aa
  from example
)
```

The issue is present even without the `select array()` bit, but this above is a real use case.

## Expected Behaviour

Pass without errors and not get mangled as it's valid BigQuery SQL, see: https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_as_struct

## Observed Behaviour

I get:

```
== [models/test.sql] FAIL
L:   1 | P:   1 | L036 | Select targets should be on a new line unless there is
                       | only one select target.
L:   1 | P:   8 | L034 | Use wildcards then simple select targets before
                       | calculations and aggregates.
```

And the "fixed" SQL is:

```sql
select array(
  select
     a as aa,
    as struct
    b as bb
  from example
)
```

## Steps to Reproduce

```sh
sqlfluff fix --force models/test.sql
```

## Version

```sh
app@27fc52961e2b:/app$ sqlfluff --version
sqlfluff, version 0.5.6
app@27fc52961e2b:/app$ python --version
Python 3.7.10
```

## Configuration

```
[sqlfluff]
templater = dbt
dialect = bigquery
exclude_rules = L031

[sqlfluff:rules]
tab_space_size = 2
max_line_length = 100

[sqlfluff:rules:L030]
capitalisation_policy = lower

```

barrywhart:
Resolves #1082 

Fairly simple PR adding new syntax to BigQuery dialect, with a bit of refactoring the pre-existing `select as struct` grammar.


src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -165,8 +165,10 @@ class SelectClauseSegment(ansi_dialect.get_segment("SelectClauseSegment")):  # t
             Sequence(
                 "AS",
                 "STRUCT",
-                Ref("StarSegment"),
-                Ref("StarModifierSegment", optional=True),
```
barrywhart, line 169: Oddly, this segment is not defined anywhere, and apparently never was. 🤯 I wonder why this isn't failing. Maybe because it's optional?

barrywhart:
I tested `sqlfluff fix` with the example query provided in the issue. It no longer introduces a syntax error.  The indentation does have a slight issue (the `a as aa` line is over-indented by 1 space), but that appears to be unrelated, and should be addressed in another issue (maybe there's already an issue for this?).

```
select array(
        select as struct
                b as bb,
                 a as aa
        from example
)
```

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -80,6 +80,11 @@
     RightArrowSegment=StringParser(
         "=>", SymbolSegment, name="right_arrow", type="right_arrow"
     ),
+    SelectClauseElementListGrammar=Delimited(
```
barrywhart, line 83: Moved this to a grammar because now it is used twice in the BigQuery dialect.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -140,6 +145,22 @@
 )
 
 
+@bigquery_dialect.segment(replace=True)
+class SelectClauseModifierSegment(BaseSegment):
```
barrywhart, line 149: Initially, the BigQuery dialect was supporting `SELECT AS STRUCT` by replacing `SelectClauseSegment` (see below). In working on this issue, I noticed that `AS STRUCT` is a general modifier that can be used on any `SELECT` statement. Implementing it this way is simpler and more generic -- doing it this way will likely avoid some future issues involving combinations of "all the things".

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -153,30 +174,6 @@ class IntervalExpressionSegment(BaseSegment):
     )
 
 
-@bigquery_dialect.segment(replace=True)
```
barrywhart, line 156: Removing this in favor of overriding `SelectClauseModifierSegment`, as explained above.

src/sqlfluff/dialects/dialect_bigquery.py
dmateusp: sounds good :+1:

src/sqlfluff/dialects/dialect_bigquery.py
dmateusp: yes, more code re-use from the base dialect :)

dmateusp (approved):

PR #1115, opened 2021-05-30
Issue 1079: L026 and BigQuery record fields

### LINKED ISSUE #1079: L026 and BigQuery record fields
## Expected Behaviour

The linter only recognises tables from the before the first`.` in the column name when using the bigquery dialect.

## Observed Behaviour

The linter sees names in nested fields which are separated by dots as table names which are not referenced in the query. Maybe it is assuming that the first name in the column string is a bigquery project?

```
== [/example.sql] FAIL
L:   5 | P:   9 | L026 | Reference 'et2.transaction.amount' refers to table/view
                       | 'transaction' not found in the FROM clause or found in
                       | parent subquery.

```

## Steps to Reproduce

SQL:
```
CREATE TABLE `example_dataset1.example_table1`
AS
SELECT
    et2._partitiontime AS partitiontime,
    et2.transaction.amount AS amount
FROM `example_dataset2.example_table2` AS et2
```

example_table2 has the following structure:

```
{
  name: 'transaction',
  type: 'RECORD',
  mode: 'REQUIRED',
  fields: [
    {
      name: 'amount',
      type: 'INTEGER',
      mode: 'REQUIRED'
    }
  ]
}
```
(date partitioned)

## Version
```
% python --version 
Python 3.8.8

% sqlfluff --version
sqlfluff, version 0.5.3
```

## Configuration
```
[sqlfluff]
verbose = 0
nocolor = False
dialect = bigquery
templater = jinja
rules = None

exclude_rules = L031
# L031 - Aliases in joins 

recurse = 0
output_line_length = 80
runaway_limit = 10
ignore_templated_areas = True

[sqlfluff:indentation]
indented_joins = False
template_blocks_indent = True

[sqlfluff:templater]
unwrap_wrapped_queries = True

ds=2020-01-01

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 120
indent_unit = space
comma_style = trailing
allow_scalar = True

[sqlfluff:rules:L003]
lint_templated_tokens = True

[sqlfluff:rules:L016]
ignore_comment_lines = False

[sqlfluff:rules:L038]

[sqlfluff:rules:L042]
forbid_subquery_in = join

prefer_count_1 = False
```

barrywhart:
Resolves #1079


test/fixtures/rules/std_rule_cases/L025.yml
```diff
@@ -42,3 +42,12 @@ test_6:
   configs:
     core:
       dialect: bigquery
+
+test_7:
+  # Test ambiguous column reference b/c of BQ STRUCT: https://github.com/sqlfluff/sqlfluff/issues/1079
```
NiallRees, line 47: b/c?

barrywhart: Are you asking why it's ambiguous?

barrywhart: I updated the comment to clarify why it's ambiguous.

NiallRees: Sorry - I'm asking what b/c means!

barrywhart: Oh, it means "because". Lazy typing... I don't know if this is a USA thing or a "me" thing!

CHANGELOG.md
```diff
@@ -19,12 +19,14 @@ Contributors:
...
   logging had been sent to stdout when reading data from stdin.
 - Add a little bit of fun on CLI exit 🎉!
-- Disabled models in the dbt templater are now skipped enitrely rather than
+- Disabled models in the dbt templater are now skipped entirely rather than
   returning an untemplated file.
 - Add a changelog check to SQLFluff continuous integration.
 - Fix bug [#1083](https://github.com/sqlfluff/sqlfluff/issues/1083), adding
   support for BigQuery named function arguments, used with functions such as
   [ST_GEOGFROMGEOJSON()](https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions#st_geogfromgeojson)
+- Fix bug [#1079](https://github.com/sqlfluff/sqlfluff/issues/1079), addressing
```
dmateusp, line 28: The CHANGELOG is new to me but shouldn't this be added after ## Unreleased?

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -417,8 +417,33 @@ class LiteralCoercionSegment(BaseSegment):
     )
 
 
+@bigquery_dialect.segment(replace=True)
+class ObjectReferenceSegment(ansi_dialect.get_segment("ObjectReferenceSegment")):  # type: ignore
```
dmateusp, line 421: not sure I understand why you need to use the ansi_dialect segment as a base and then not overwrite anything?

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -417,8 +417,33 @@ class LiteralCoercionSegment(BaseSegment):
...
+@bigquery_dialect.segment(replace=True)
+class ColumnReferenceSegment(ObjectReferenceSegment):
+    """A reference to column, field or alias."""
+
+    type = "column_reference"
+
+    def extract_possible_references(self, level: int):
+        """Extract possible references of a given level."""
+        refs = list(self.iter_raw_references())
+        if level == 3 and len(refs) >= 3:
```
dmateusp, line 436: I think I need a little help (or comment) to understand which level corresponds to which reference, column / field / alias?

src/sqlfluff/dialects/dialect_bigquery.py
dmateusp: Can't you just import the ObjectReferenceSegment from the ansi_dialect package as a class?

src/sqlfluff/dialects/dialect_bigquery.py
dmateusp: If each level has a meaning, could it be turned into a named enumerator?

src/sqlfluff/rules/L026.py
```diff
@@ -28,6 +28,17 @@ class Rule_L026(Rule_L025):
 
     """
 
+    @staticmethod
+    def _is_bad_tbl_ref(table_aliases, parent_select, tbl_ref):
+        if tbl_ref and tbl_ref[0] not in [a.ref_str for a in table_aliases]:
```
dmateusp, line 33: those if statements are a little complex to me, could you explain/comment ?

src/sqlfluff/rules/L026.py
```diff
@@ -28,6 +28,17 @@ class Rule_L026(Rule_L025):
...
+            # Last check, this *might* be a correlated subquery reference.
+            if parent_select:
+                parent_aliases, _ = get_aliases_from_select(parent_select)
+                if parent_aliases and tbl_ref[0] in [a[0] for a in parent_aliases]:
```
dmateusp, line 37: could do
```suggestion
                return not (parent_aliases and tbl_ref[0] in [a[0] for a in parent_aliases])
```

dmateusp (commented):
I have some questions but I haven't looked into the project's code in a few weeks so I'm a little lost :)

CHANGELOG.md
barrywhart: Yes. I'll move it. I think a PR yesterday incorrectly added a 0.6.1 section, which is not an actual release. I like the CHANGELOG, but I'm starting to wonder if there's a technical solution that would automate it a bit more. Manual editing is a bit error prone and creates lots of merge "conflicts".

src/sqlfluff/dialects/dialect_bigquery.py
barrywhart: I'll take a look at that. I'm in unfamiliar territory here -- replacing a subclass of `ObjectReferenceSegment` in a child dialect (BigQuery). 🤯

src/sqlfluff/dialects/dialect_bigquery.py
barrywhart: I had the same thought -- I'll look into it for this PR. Until I do that, this [section of`dialect_ansi.py`](https://github.com/sqlfluff/sqlfluff/blob/master/src/sqlfluff/dialects/dialect_ansi.py#L612L614) documents the meaning of the values.

src/sqlfluff/dialects/dialect_bigquery.py
barrywhart: Done! IIUC, Alan did some refactoring a few months back, and dialects shouldn't directly `import` from other dialects, so I did this:
```
ObjectReferenceSegment = ansi_dialect.get_segment("ObjectReferenceSegment")
```

This seems to work fine -- I can use `ObjectReferenceSegment` as a base class, which was why I had created this unnecessary new class. 👍

src/sqlfluff/rules/L026.py
barrywhart: Makes sense, but I think I'd like to keep the existing structure -- now that I have added more comments as you suggested, I find that the various `return` statements are good places to provide context on the significance and context of reaching various lines of code. I'll push up this changes in a few minutes.

barrywhart:
@dmateusp: Thanks for the thoughtful comments! I added comments and made other changes to hopefully make things simpler and more maintainable. I believe it's ready for another review.

CHANGELOG.md
```diff
@@ -22,6 +24,10 @@ Contributors:
   ANSI dialect which defaults to the semicolon, but allows it to be more intuitive 
   when overriden in a specific child dialect (mysql) [#901](https://github.com/sqlfluff/sqlfluff/issues/901))
 - Added support for the DELIMITER statement in the mysql dialect [#901](https://github.com/sqlfluff/sqlfluff/issues/901))
+
+Contributors:
```
dmateusp, line 28: I think you might have added this by mistake, (I'm seeing many issues managing CHANGELOGs on PRs :/ )

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -606,18 +607,37 @@ def qualification(self):
         """Return the qualification type of this reference."""
         return "qualified" if self.is_qualified() else "unqualified"
 
-    def extract_reference(self, level: int) -> Optional[ObjectReferencePart]:
-        """Extract a reference of a given level.
+    class ReferencePart(Enum):
+        """Labels for the "levels" of a reference.
+
+        Note: Since SQLFluff does not have access to database catalog
```
dmateusp, line 613: that's very clear, thanks!

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -606,18 +607,37 @@ def qualification(self):
...
+        and intended mainly to clarify the intent of the code -- no guarantees!
+        Additionally, the terminology may vary by dialect, e.g. in BigQuery,
+        "project" would be a more accurate term than "schema".
+        """
+
+        OBJECT = 1
+        TABLE = 2
+        SCHEMA = 3
+
+    def extract_possible_references(self, level: int) -> List[ObjectReferencePart]:
```
dmateusp, line 627: can't that be `level: ReferencePart` now?

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -409,8 +409,32 @@ class LiteralCoercionSegment(BaseSegment):
...
+            return [refs[0]]
+        if level == 2 and len(refs) >= 3:
```
dmateusp, line 429: can you replace the values with the enum now?

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -409,8 +409,32 @@ class LiteralCoercionSegment(BaseSegment):
...
+            return [refs[0], refs[1]]
+        if level == 1 and len(refs) >= 3:
+            return [refs[1], refs[2]]
+        return super().extract_possible_references(level)
+
+
 @bigquery_dialect.segment()
-class HyphenatedObjectReferenceSegment(ansi_dialect.get_segment("ObjectReferenceSegment")):  # type: ignore
+class HyphenatedObjectReferenceSegment(ObjectReferenceSegment):  # type: ignore
```
dmateusp, line 437: :)

src/sqlfluff/rules/L026.py
```diff
@@ -28,6 +28,25 @@ class Rule_L026(Rule_L025):
...
+            return False
+
+        # Not a table alias. It it referring to a correlated subquery?
+        if parent_select:
+            parent_aliases, _ = get_aliases_from_select(parent_select)
+            if parent_aliases and tbl_ref[0] in [a[0] for a in parent_aliases]:
+                # Yes. Therefore okay.
+                return False
+
+        # It's not referring to an alias or a correlated subquery. Looks like a
```
dmateusp, line 46: very clear now with the comments :)

dmateusp (commented):
some additional questions ^^

CHANGELOG.md
barrywhart: Probably so. This file has conflicts all the time, because people are changing same/adjacent lines. I wonder if there's a way to avoid this somehow.

barrywhart: Will remove.

src/sqlfluff/dialects/dialect_ansi.py
barrywhart: I want to support both integers or enums here, because of the ambiguity I commented on in the enum. You make a good point, though -- I have updated the type hint to specify that either is allowed and to handle both correctly. 👍

In BigQuery, for example, a reference could have literally dozens of parts (for example, a highly nested structure). The enum is useful for the 3 typical or "canonical" parts. Currently, the only code that calls this function happens to use the enum, but I could easily see other future users of this class & function wanting to just specify a number, e.g. if they were looping over all the parts or something. I could see revisiting this code in the future. It's challenging to find the right balance of clarity and flexibility here, because of the limited number of "data points" (i.e. code that calls this function).

src/sqlfluff/dialects/dialect_bigquery.py
barrywhart: Done! I also made some other changes, hinted at in my comment above.

barrywhart:
@dmateusp: Ready for another review!

dmateusp (approved):
:+1: You might just need to add a test with a Schema reference and another one with an Object reference for the code coverage to stop complaining

all the rest lgtm :)

barrywhart:
I'm going to go ahead and merge this in spite of a bit of missing coverage. I'm technically on PTO, and starting to get some "looks" from my family. 😄

PR #1118, opened 2021-06-01
Ignore select statements with macros in L034

NiallRees:
Several users have now reported that L034 will apply broken fixes to select statements which contain macros (https://getdbt.slack.com/archives/CV2KM39KR/p1622420620013500), most commonly dbt's `{{ dbt_utils.surrogate_key() }}`. This change anchors L034 on the select statement segment so that statements using macros are completed ignored by the rule.

@barrywhart I realise this changes the behaviour that you added in https://github.com/sqlfluff/sqlfluff/pull/918, so I'm interested for your thoughts. For us to keep the behaviour you added and be able to ignore the statement if it contains a templated section we'll need to add additional functionality to the ignore_templated_areas logic.

I've had to remove the dbt017-py38 tests due to an incompability with dbt utils 0.6.3 which was throwing the following error as a result of https://getdbt.slack.com/archives/C2JRRQDTL/p1610568190132200: 
```
sqlfluff.core.errors.SQLTemplaterError: dbt compilation error on file 'models/my_new_project/L034_test.sql', 'dbt.context.providers.ParseDatabaseWrapper object' has no attribute 'dispatch'
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,13 +133,6 @@ jobs:
       - image: circleci/python:3.9
         environment:
           TOXENV: mypy
-  dbt017-py38:
-    <<: *common
-    docker:
-      - image: circleci/python:3.8
-        environment:
-          TOXENV: dbt017-py38
-          TOX_ARGS: -m dbt
   dbt018-py38:
     <<: *common
     docker:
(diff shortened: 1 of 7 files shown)


barrywhart (approved):
Looks good!

As far as my earlier PR, no concerns.

PR #1119, opened 2021-06-01
L034 should ignore union statements

### LINKED ISSUE #1104: fix for L034 reorder of columns can break query
The fix for L034 reorders columns. However, when two select queries are being unioned the fix may change the result of the query and/or cause the query to break. 

The following is a case where there is no re-ordering of the columns that will work without altering the result of the query. 

## Expected Behaviour
Don't fix this. I'm not sure if L034 should apply to this situtation, but I don't think there is an automated fix. 

## Observed Behaviour
sqlfluff fixes as follows which significantly changes the query:
```
select
    4 as first_col,
    case
        when 4 = 5
            then 'x'
        else 'y'
    end as second_col
union all
select
    'l' as second_col,
    case
        when 4 = 5
            then 9
        else 10
    end as first_col
```
## Steps to Reproduce
Run sqlfluff fix on the following SQL:
```
select
   4 as first_col,
   case
       when 4 = 5
           then 'x'
       else 'y'
   end as second_col
union all
select
   case
       when 4 = 5
           then 9
       else 10
   end as first_col,
   'l' as second_col
```
## Version
sqlfluff 0.5.6

NiallRees:
Resolves #1104

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - L034 now ignores select statements which contain macros.
+- L034 now ignores select statements part of a set expression, most commonly a union.
 
 Contributors:
 - [@GitHub-Username](Link to GitHub profile) ([#PR-Number](Link to PR))
--- a/src/sqlfluff/rules/L034.py
+++ b/src/sqlfluff/rules/L034.py
@@ -45,7 +45,7 @@ class Rule_L034(BaseRule):
         self.current_element_band = i
         self.seen_band_elements[i].append(segment)
 
-    def _eval(self, segment, **kwargs):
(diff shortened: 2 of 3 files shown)


barrywhart:
Ah, I had wondered if `SelectCrawler` might be useful here, but its main value comes when recursing over a tree of `SELECT` statements. In this case, since you're looking for a `set_expression` parent, I wonder if it'd be more appropriate just check the `parent_stack`, as some existing rules do.

NiallRees:
> Ah, I had wondered if `SelectCrawler` might be useful here, but its main value comes when recursing over a tree of `SELECT` statements. In this case, since you're looking for a `set_expression` parent, I wonder if it'd be more appropriate just check the `parent_stack`, as some existing rules do.

That's exactly what I needed!

barrywhart (approved):
Looks good!

PR #1132, opened 2021-06-04
Placeholder positioning and tests

### LINKED ISSUE #1124: weird "ValueError: Zero length string passed to ReSegment!?" error
I have the following query with Jinja templating (actually my initial query was more complex, I just simplified it as much as possible for the bug report):

```
{% set some_condition %}TRUE{% endset %}

WITH cust AS
    (SELECT SNAPSHOT_DATE
    FROM DATAHUB.SNAPSHOT_DAILY
    WHERE {{some_condition}}
    )

SELECT DISTINCT cust.SNAPSHOT_DATE
FROM cust
```

and when running "sqlfluff lint ./addons.template.sql" it throws the following exception:
ValueError: Zero length string passed to ReSegment!?

The full stacktrace is attached:
[stacktrace.txt](https://github.com/sqlfluff/sqlfluff/files/6583612/stacktrace.txt)

## Observed Behaviour
What makes it very strange is that if I remove almost any of the parts of this query, it doesn't throw an exception anymore. 
For example, the following alternatives work just fine:

```
{% set some_condition %}TRUE{% endset %}

WITH cust AS
    (SELECT SNAPSHOT_DATE
    FROM DATAHUB.SNAPSHOT_DAILY
    )

SELECT DISTINCT cust.SNAPSHOT_DATE
FROM cust
```

```
{% set some_condition %}TRUE{% endset %}

WITH cust AS
    (SELECT SNAPSHOT_DATE
    FROM DATAHUB.SNAPSHOT_DAILY
    WHERE {{some_condition}}
    )
```

```
WITH cust AS
    (SELECT SNAPSHOT_DATE
    FROM DATAHUB.SNAPSHOT_DAILY
    WHERE {{some_condition}}
    )

SELECT DISTINCT cust.SNAPSHOT_DATE
FROM cust
```

```
{% set some_condition %}TRUE{% endset %}

WITH cust AS
    (SELECT SNAPSHOT_DATE
    FROM SNAPSHOT_DAILY
    WHERE {{some_condition}}
    )

SELECT DISTINCT cust.SNAPSHOT_DATE
FROM cust
```

## Version
sqlfluff, version 0.5.6
Python 3.8.8

## Configuration
I use the default configuration

alanmcruickshank:
This resolves #1124 .

More importantly in fixing that I realised that we don't really have any test coverage on the placement of meta segments - and it was not good. This adds two test cases in the `jinja` test suite to test their handling and refactors a few places which control their generation.

This also adds significantly more logging in the lexing and templating stages to support future work in this area.

It's not perfect but it's a lot better than it was.


barrywhart (approved):
I skimmed over it -- looks good!

PR #1133, opened 2021-06-05
Rename --parallel to --processes

### LINKED ISSUE #1057: Better name for --parallel?
I feel that --threads instead of --parallel is clearer in terms of what the input value does. We can also then change the help string to specify that the number passed is the number of threads used.

NiallRees:
Resolves #1057

### DIFF
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -77,7 +77,7 @@ def set_logging_level(verbosity, logger=None, stderr_output=False):
     # Set up a handler to colour warnings red.
     handler.addFilter(RedWarningsFilter())
     if logger:
-        focus_logger = logging.getLogger("sqlfluff.{0}".format(logger))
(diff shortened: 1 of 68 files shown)


src/sqlfluff/cli/commands.py
```diff
@@ -288,7 +288,7 @@ def dialects(**kwargs):
 )
 @click.option(
     "-p",
-    "--parallel",
+    "--processes",
     type=int,
     default=1,
     help="If set to a value higher than 1, run SQLFluff in parallel, "
```
barrywhart, line 294:
```suggestion
    help="If set to a value higher than 1, starts multiple SQLFluff processes, "
```

barrywhart (changes requested):
I think we should change the wording of the CLI help message, otherwise good to go. Thanks for doing this!

barrywhart (approved):

PR #1139, opened 2021-06-06
Release 0.6.0

NiallRees:


CHANGELOG.md
```diff
@@ -37,13 +44,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   SET SCHEMA and DROP SCHEMA support to ANSI dialect.
-- Added support for additional delimiters by creating a new DelimiterSegment in the 
-  ANSI dialect which defaults to the semicolon, but allows it to be more intuitive 
+- Added support for additional delimiters by creating a new DelimiterSegment in the
+  ANSI dialect which defaults to the semicolon, but allows it to be more intuitive
   when overriden in a specific child dialect (mysql) [#901](https://github.com/sqlfluff/sqlfluff/issues/901))
 
-Contributors:
-- [@bolajiwahab](https://github.com/bolajiwahab) ([#1063])(https://github.com/sqlfluff/sqlfluff/pull/1063)
-
```
alanmcruickshank, line 46: I think we should either keep this bit - or reference the user in the but that they changed. In this case that would be _Allow `CREATE FUNCTION` syntax without arguments_.

I realise we haven't been very consistent about using the contributors section, but we should either use it or just remove it.

alanmcruickshank (changes requested):

CHANGELOG.md
NiallRees: Yeah it felt a bit silly only referencing one contributor. I'll just mention them in the feature added

NiallRees: Can you approve for when it's sorted? Then I can get this merged

alanmcruickshank (approved):
:shipit:

NiallRees:
Waiting on https://github.com/sqlfluff/sqlfluff/pull/898 mostly as I'm concerned that there are some strange logging side effects going on that would be good to straighten out.

NiallRees:
> Waiting on #898 mostly as I'm concerned that there are some strange logging side effects going on that would be good to straighten out.

We worked out the problem and will address separately.

Issue #1131, opened 2021-06-04
Add support for `DEFINER`


silverbullettruck2001:
Within the mysql dialect there is a keyword for `DEFINER` that exists for `CREATE PROCEDURE`/`CREATE FUNCTION` but is not currently supported.  This would be good to add so that the mysql dialect is more completely supported.

PR #1141, opened 2021-06-07
Add mysql dialect support for DECLARE and IF

silverbullettruck2001:
Also, added support for `DEFINER` to resolve [#1131](https://github.com/sqlfluff/sqlfluff/issues/1131)


src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -129,21 +136,129 @@ class CreateTableStatementSegment(
...
+    ),
+)
+
+mysql_dialect.patch_lexer_matchers(
+    [
+        RegexLexer(
+            "ampersand",
+            r"[@][a-zA-Z0-9_]*",
+            CodeSegment,
+        ),
```
alanmcruickshank, line 169: Why `ampersand`? An ampersand is an `&` not a `@`. I believe the latter is called an `at sign`: https://en.wikipedia.org/wiki/List_of_typographical_symbols_and_punctuation_marks

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -129,21 +136,129 @@ class CreateTableStatementSegment(
...
+
+@mysql_dialect.segment(replace=True)
+class StatementSegment(ansi_dialect.get_segment("StatementSegment")):  # type: ignore
+    """Overriding StatementSegment to allow for additional segment parsing."""
+
+    parse_grammar = ansi_dialect.get_segment("StatementSegment").parse_grammar.copy(
+        insert=[
+            Ref("DelimiterStatement"),
+            Ref("CreateProcedureStatementSegment"),
+            Ref("TransactionStatementSegment"),
```
alanmcruickshank, line 248: Because you're replacing an existing segment (e.g. `TransactionStatementSegment`) you don't need to reinsert it here, it's already part of `StatementSegment`. You should only need to add segments which don't exist at all in the parent dialect.

test/fixtures/parser/mysql/begin.sql
```diff
@@ -0,0 +1,4 @@
+blocks:BEGIN
+#set @abc = 1;
```
alanmcruickshank, line 2: did you mean to comment out line 2 here? I'm not sure what purpose it serves either way.

alanmcruickshank (dismissed):
A few small things, but broadly this looks good. I love the detailed test coverage and links to documentation 👍 .

Thanks for putting this together.

test/fixtures/parser/mysql/begin.sql
silverbullettruck2001: yeah...sorry, I cleaned that up now.

src/sqlfluff/dialects/dialect_mysql.py
silverbullettruck2001: @alanmcruickshank  I got that cleaned up now

src/sqlfluff/dialects/dialect_mysql.py
silverbullettruck2001: @alanmcruickshank  this was a good one, I am not sure why my brain thought that it was an ampersand...lol. I renamed this, and because `atsign` is not an existing lexer_matcher I had to change the method to be `insert_lexer_matchers` instead.  All good now.

bolajiwahab:
Just a small comment. Is it possible to move the if logic to ansi?

silverbullettruck2001:
> Just a small comment. Is it possible to move the if logic to ansi?

I would rather wait and treat that in a separate ticket.

bolajiwahab:
> > Just a small comment. Is it possible to move the if logic to ansi?
> 
> I would rather wait and treat that in a separate ticket.

What I mean is that other dialects can easily benefit from this just by putting it in ANSI and IIRC we always prefer putting things in ANSI to reduce code repetitions.

silverbullettruck2001:
> > > Just a small comment. Is it possible to move the if logic to ansi?
> > 
> > 
> > I would rather wait and treat that in a separate ticket.
> 
> What I mean is that other dialects can easily benefit from this just by putting it in ANSI and IIRC we always prefer putting things in ANSI to reduce code repetitions.

I understand, but I am new to this project and feel that a Chang elite that would be more than I am comfortable doing.  So I would rather deal with that as a separate issue and address it later and then when it's refactored, that PR could update the MySQL dialect and remove the dialect specific syntax when it is added to ANSI.

NiallRees (approved):
Thank you @silverbullettruck2001 !!

PR #1150, opened 2021-06-10
Add select stmt segments

silverbullettruck2001:
I am doing a PR for this knowing it is not ready.  I need some assistance with two areas...

1. figuring out why my refactor of the IndexHintClauseSegment is not working correctly with the `ORDER BY` and `GROUP BY` segments that can exist within an index hint.  It keeps picking it up as a normal ORDER BY or GROUP BY statement, which it is not, but I don't know how to adjust this.

Here are the tests that aren't parsing because of the `ORDER BY` and `GROUP BY` segments in the clause:
[GROUP BY example](https://github.com/silverbullettruck/sqlfluff/blob/AddSelectStmtSegments/test/fixtures/parser/mysql/force_index_for_group_by.yml)

[ORDER BY example](https://github.com/silverbullettruck/sqlfluff/blob/AddSelectStmtSegments/test/fixtures/parser/mysql/force_index_for_order_by.sql)

This is what I would expect them to parse as:
(expected GROUP BY)
```yaml
file:
  statement:
    select_statement:
      select_clause:
        keyword: SELECT
        select_clause_element:
          wildcard_expression:
            wildcard_identifier:
              star: '*'
      from_clause:
        keyword: FROM
        from_expression:
          from_expression_element:
            table_expression:
              table_reference:
                identifier: onetable
            index_hint_clause:
            - keyword: FORCE
            - keyword: INDEX
            - keyword: FOR
            - keyword: GROUP
            - keyword: BY
            - bracketed:
                start_bracket: (
                object_reference:
                  identifier: i2
                end_bracket: )
  statement_terminator: ;
```

expected (ORDER BY)
```yaml
```yaml
file:
  statement:
    select_statement:
      select_clause:
        keyword: SELECT
        select_clause_element:
          wildcard_expression:
            wildcard_identifier:
              star: '*'
      from_clause:
        keyword: FROM
        from_expression:
          from_expression_element:
            table_expression:
              table_reference:
                identifier: onetable
            index_hint_clause:
            - keyword: FORCE
            - keyword: INDEX
            - keyword: FOR
            - keyword: ORDER
            - keyword: BY
            - bracketed:
                start_bracket: (
                object_reference:
                  identifier: i2
                end_bracket: )
  statement_terminator: ;
```
2. I have added comments on lines 605-607, 649-650, 678-679 and 712-713 for some code that I believe is not correct, but I am not sure how to avoid it.  I am looking for suggestions on how to do this better or be written differently to remove the unnecessary code.


src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -270,11 +410,374 @@ class DropStatementSegment(BaseSegment):
...
+        Ref("IntoClauseSegment", optional=True),
+        Ref("FromClauseSegment", optional=True),
+        Ref("WhereClauseSegment", optional=True),
+        Ref("GroupByClauseSegment", optional=True),
+        Ref("HavingClauseSegment", optional=True),
+        Ref("ForClauseSegment", optional=True),
+    )
+
+
+# I am unclear why I have to override this segement, but if I don't then new segments won't parse
```
NiallRees, line 649: The issue is that when a segment is overridden, the parser looks for all Ref() segments within the present dialect. To fix this, you need to explicitly refer to the segments in the source dialect from which you are overriding. See https://github.com/sqlfluff/sqlfluff/blob/master/src/sqlfluff/dialects/dialect_postgres.py#L195 for an example.

silverbullettruck2001: I am not clear on your suggestion.  I would like to completely remove the overrides for `SelectClauseElementSegment`, `SelectClauseSegment`, and `SelectStatementSegment` since I am not altering them at all.  Could you provide a specific example of what you are asking me to change?

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -270,11 +410,374 @@ class DropStatementSegment(BaseSegment):
...
+                    ),
+                    optional=True,
+                ),
+            ),
+        ),
+    )
+
+
+# I wanted to do an override and just insert the new Refs but I need
+# to have the IntoClause be before FromClauseSegment, but it doesn't work to add the `before` clause
```
NiallRees, line 606: Reading into this more - can you change the PR to be the way you're expecting this to work so that we can debug more easily? As in:

```
    parse_grammar = ansi_dialect.get_segment(
        "UnorderedSelectStatementSegment"
    ).parse_grammar.copy(
        insert=[Ref("IntoClauseSegment", optional=True)],
        before=Ref("FromClauseSegment", optional=True),
    )
```

silverbullettruck2001: I added the statements as I think it should work, but then it stops parsing correctly.  Here's an example of something that did parse, but now doesn't after this change.
[select into outfile](https://github.com/silverbullettruck/sqlfluff/blob/AddSelectStmtSegments/test/fixtures/parser/mysql/select_into_outfile_fields_enclosed.yml)

src/sqlfluff/dialects/dialect_mysql.py
barrywhart: IIUC, Niall is saying (for example), that at line 678, you should replace this code:
```
            Ref("SelectClauseElementSegment"),
```

with something like this:
```
            ansi_dialect.get_segment("SelectClauseElementSegment"),
```

I haven't encountered this situation before, so I'm not sure this is exactly right. Can you give it a try and see if it works? If it does, you should be able to remove lines 624-650 (the copy pasted `SelectClauseElementSegment` from the ANSI dialect).

And once we get this working, a similar technique will allow removing the other segments mentioned above.

PR #1151, opened 2021-06-10
postgres dialect: add support for array types and array literals

### LINKED ISSUE #1142: PostgreSQL: Array notation

## Expected Behaviour
I would expect these  statements to be parsed by sqlfluff, as they are supported by PostgreSQL.

L:   1 | P:  16 | L008 | Commas should be followed by a single whitespace unless
                       | followed by a comment.
L:   1 | P:  29 |  PRS | Found unparsable section: ',4'
L:   1 | P:  30 | L008 | Commas should be followed by a single whitespace unless
                       | followed by a comment.
## Steps to Reproduce
>SELECT ARRAY[1,2] || ARRAY[3,4];

Lint the file with the following command(s) :
>sqlfluff lint test.sql --dialect postgres or sqlfluff lint test.sql
## Version
I am using sqlfluff version 0.5.6 and python version 3.8.5.
## Configuration
This example comes from the official PostgreSQL [documentation](https://www.postgresql.org/docs/8.4/arrays.html). I found similar results when executing queries from Github Project [vertx-jooq](https://github.com/jklingsporn/vertx-jooq).

boblannon:
This PR adds the following to the postgres dialect

- support for array literals
- support for array type declarations 

I'm new to this code, so I'm happy to take any feedback.  I didn't intend for this change to alter the parses of existing fixtures, but I can't find another way, yet.

## Note: Test fixtures changed!

Fixtures were updated in order to make tests pass, because (as noted in comments) this addition changes the structure of the `data_type_identifier` parse.  Maintainers should definitely make sure this change is okay before merging.


boblannon:
This is the main difference that my change introduces, and it affects several of the postgres fixture tests:

```diff
diff --git a/test/fixtures/parser/postgres/postgres_create_table.yml b/test/fixtures/parser/postgres/postgres_create_table.yml
index 3ab8135b..aa7c1a5a 100644
--- a/test/fixtures/parser/postgres/postgres_create_table.yml
+++ b/test/fixtures/parser/postgres/postgres_create_table.yml
@@ -12,5 +12,6 @@ file:
           data_type:
             identifier: public
             dot: .
-            data_type_identifier: MY_TYPE
+            data_type_identifier:
+              raw: MY_TYPE
         end_bracket: )
```

The `RegexParser` by itself produces a differently-structured parse than when its used as part of a `Sequence`.    I've tried to avoid it but can't figure out how to do so while keeping the array additions to the `DatatypeIdentifierSegment`.  I guess a different approach would have been to just complicate the original `RegexParser` used in ansi dialect?

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -86,6 +87,69 @@
...
+    match_grammar = Sequence(
+        "ARRAY",
+        Bracketed(
+            Delimited(Ref("LiteralGrammar"), optional=False),
+            bracket_type="square",
+        ),
+    )
+
+
+postgres_dialect.replace(
```
barrywhart, line 104: Rather than copying and modifying the source code from the ANSI `BaseExpressionGrammar`, could you look at extending it as seen in the BigQuery dialect [here](https://github.com/sqlfluff/sqlfluff/blob/master/src/sqlfluff/dialects/dialect_bigquery.py#L112L114)? This technique is more maintainable, in that changes to the ANSI base will be automatically reflected in Postgres.

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -86,6 +87,69 @@
...
+        Ref("FunctionSegment"),
+        Ref("IntervalExpressionSegment"),
+        Ref("ColumnReferenceSegment"),
+        Ref("ExpressionSegment"),
+    )
+)
+
+
+@postgres_dialect.segment(replace=True)
+class DatatypeIdentifierSegment(BaseSegment):
```
barrywhart, line 118: Did you consider replacing `DataTypeSegment` rather than `DataTypeIdentifierSegment`? Typically, in SQLFluff, `Identifier` is closer to a pure name. Structurally, this segment looks a bit more like `DataTypeIdentifierSegment`.

src/sqlfluff/dialects/dialect_postgres.py
boblannon: ah, sure thing, makes sense.

src/sqlfluff/dialects/dialect_postgres.py
boblannon: I did consider this. That was my first pass, but i second-guessed it.  I can switch back to that.

barrywhart: I'm not 100% sure it'll work in terms of handling array literals, but it seems cleaner in terms of fitting within the existing ANSI and Postgres dialects. Let us know how it goes!

boblannon:
flipping this PR back to draft to make sure it isn't merged before I get time to respond to comments

tunetheweb:
I'm going to close this. The code has moved on a lot so think we should reapply this change to latest code base in a new PR if we want to finish it out.

boblannon:
That's fine, thanks for the heads-up. I haven't had bandwidth to follow up with requested changes, but I'll  open a new PR when I do.

PR #1176, opened 2021-07-01
Add syntax for mysql specific SELECT statements

silverbullettruck2001:
Shoutout to @barrywhart, @dmateusp and @NiallRees for their help and responsiveness to my questions regarding this enhancement!  Thanks again!


src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -53,11 +77,29 @@
             Ref("DoubleQuotedLiteralSegment"),
         ]
     ),
-    PostTableExpressionGrammar=Sequence(
```
barrywhart, line 56: Why was this removed?

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -53,11 +77,29 @@
...
-        Bracketed(Ref("ObjectReferenceSegment")),
+    FromClauseTerminatorGrammar=OneOf(
+        "WHERE",
+        "LIMIT",
+        "GROUP",
+        "ORDER",
+        "HAVING",
+        "QUALIFY",
+        "WINDOW",
+        Ref("IndexHintClauseSegment"),
```
barrywhart, line 88: It looks like this element is the same as the ANSI dialect except for lines 88-90. Should we look at using `.copy()` to base this segment on the ANSI one, to reduce copy/paste? (NOTE: I _think_ this is supported for grammars as well as segments.)

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -53,11 +77,29 @@
...
+        Ref("WithNoSchemaBindingClauseSegment"),
+    ),
+    BaseExpressionElementGrammar=OneOf(
+        Ref("LiteralGrammar"),
+        Ref("BareFunctionSegment"),
+        Ref("FunctionSegment"),
+        Ref("IntervalExpressionSegment"),
+        Ref("ColumnReferenceSegment"),
+        Ref("ExpressionSegment"),
+        Ref("SessionVariableNameSegment"),
```
barrywhart, line 101: Similar to the above, it looks like this is the same as ANSI except for these two added items. Consider using `.copy()` here as well?

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -475,6 +517,214 @@ class DefinerSegment(BaseSegment):
...
+            OneOf("NOWAIT", Sequence("SKIP", "LOCKED"), optional=True),
+        ),
+        Sequence("LOCK", "IN", "SHARE", "MODE"),
+        optional=True,
+    )
+
+
+@mysql_dialect.segment()
+class IndexHintClauseSegment(BaseSegment):
+    """This is the body of a index hint clause."""
```
barrywhart, line 710:
```suggestion
    """This is the body of an index hint clause."""
```

test/fixtures/parser/mysql/call_statement.sql
```diff
@@ -1 +1 @@
-CALL zproc_validate_send_adminuser_otp('a');
\ No newline at end of file
+CALL somefunction('a');
```
barrywhart, line 1: Is there a reason for changing the function name in this and other test `.sql` files?

test/fixtures/parser/mysql/force_index.sql
```diff
@@ -1,8 +1 @@
-SELECT
-       onetable.f1,
-       twotable.f1
-FROM onetable
-         LEFT JOIN
-     twotable
-         FORCE INDEX FOR JOIN (idx_index)
-     ON onetable.f1 = twotable.f1
\ No newline at end of file
+SELECT * FROM onetable FORCE INDEX (idx_index);
```
barrywhart, line 1: Why so many changes to this file? Could this (should this) be a new test file instead?

barrywhart (changes requested):
This looks pretty close to ready! I have a few questions and suggestions.

src/sqlfluff/dialects/dialect_mysql.py
silverbullettruck2001: > Why was this removed?

It wasn't needed after I expanded the scope for the INDEX functionality by added in the IndexHintClauseSegment at the bottom of the dialect definition so I removed this.

test/fixtures/parser/mysql/call_statement.sql
silverbullettruck2001: I used a name of a proc that is specific to me and wanted to make it generic instead.

test/fixtures/parser/mysql/force_index.sql
silverbullettruck2001: I simplified this one since the original statement for the test was more complex than necessary.  I will add a new test that is specific to using the `FORCE INDEX` on a join so that the original test will still be validated.

src/sqlfluff/dialects/dialect_mysql.py
silverbullettruck2001: > 

I will get this correction made

test/fixtures/parser/mysql/force_index.sql
silverbullettruck2001: @barrywhart good catch, after adding the test back, my changes no longer covered it.  But I have it updated now to cover both this test and my existing tests.  Thanks!

src/sqlfluff/dialects/dialect_mysql.py
silverbullettruck2001: Will do, good suggestion.

src/sqlfluff/dialects/dialect_mysql.py
silverbullettruck2001: Will do, good suggestion.

silverbullettruck2001:
> This looks pretty close to ready! I have a few questions and suggestions.

@barrywhart I made the suggested changes, let me know if there's anything else that needs to be adjusted.

barrywhart (approved):

NiallRees:
Your commitment to this has been great to see @silverbullettruck2001!

barrywhart:
PR looks good! Will merge once the build passes.

silverbullettruck2001:
> Your commitment to this has been great to see @silverbullettruck2001!

Thanks @NiallRees!

PR #1177, opened 2021-07-03
 Issue 654 - add file encoding detection for read/write of files  

barrywhart:
Replaces #899, which had been unofficially approved but was not merged and became out of date. This is a pretty mechanical update of that PR, no additional features added.

As with the original PR, it does not fully address #654, because it does not add CLI support, but the decision on the earlier PR was to add that in a subsequent PR.


NiallRees (approved):
Thanks @barrywhart!

Issue #1191, opened 2021-07-08
Directory in .sqlfluffignore not ignored when running SQLFluff on it


scrambldchannel:
As discussed in this slack thread - https://sqlfluff.slack.com/archives/C01GX4L213J/p1625669293157000

If a directory is listed in `.sqlfluffignore` eg:

```
queries
```

Running sqlfluff on said directory will lint / fix the files in `queries` even though it should be ignored. For example:

```
$ sqlfluff lint queries
== [queries/test.sql] FAIL
L:   1 | P:   1 | L044 | Query produces an unknown number of result columns.
L:   1 | P:   7 | L039 | Unnecessary whitespace found.
L:   2 | P:   1 | L010 | Inconsistent capitalisation of keywords.
L:   2 | P:   5 | L039 | Unnecessary whitespace found.
L:   3 | P:   1 | L010 | Inconsistent capitalisation of keywords.
L:   3 | P:  17 | L010 | Inconsistent capitalisation of keywords.
L:   3 | P:  20 | L009 | Files must end with a trailing newline.
L:   3 | P:  20 | L010 | Inconsistent capitalisation of keywords.
All Finished 📜 🎉!
```

However running on the parent directory or a file in `queries` does observe the contents of `.sqlfluffignore`:

```
$ sqlfluff lint queries/test.sql 
WARNING    Exact file path queries/test.sql was given but it was ignored by a .sqlfluffignore pattern, re-run with `--disregard-sqlfluffignores` to skip .sqlfluffignore 
All Finished 📜 🎉!
```

```
$ sqlfluff lint .
All Finished 📜 🎉!
```

## Expected Behaviour

I would expect the files in the ignored directory to be ignored

## Observed Behaviour

Invoking sqlfluff on an ignored folder does not ignore it

## Steps to Reproduce

* create a directory
* add directory to the sqlfluffignore file in its parent directory
* run `sqlfluff lint directory`

## Version

sqlfluff, version 0.6.0

Python 3.8.10

## Configuration
```
Manifests itself with default config (i.e. no configuration file in place)
```

barrywhart:
This looks like a possible duplicate of #556. Can you let us know, @scrambldchannel?

scrambldchannel:
> This looks like a possible duplicate of #556. Can you let us know, @scrambldchannel?

Ah, it does look suspiciously like a duplicate, sorry for not noticing this in the first place :shrug:

barrywhart:
[...]

PR #1192, opened 2021-07-08
fix: sqlfluffignore not being observed when linting/fixing ignored path explicitly

scrambldchannel:
https://github.com/sqlfluff/sqlfluff/issues/1191

WIP - adds a failing test to illustrate issue


tunetheweb:
I'm going to close this. The code has moved on a lot so think we should reapply this change to latest code base in a new PR if we want to finish it out.

PR #1195, opened 2021-07-09
Fix runtime error in diff-cover plugin caused by new diff-cover release 6.1.0

barrywhart:


src/sqlfluff/diff_quality_plugin.py
```diff
@@ -37,6 +37,9 @@ def violations(src_path):
             result.append(Violation(violation.line_no, message))
         return result
 
+    def measured_lines(self, src_path):
```
barrywhart, line 40: The latest release of `diff-cover` inexplicably changed this method to be "abstract", so now plugin implementers have to implement it. The PR that made this change was apparently created by "fixing" pylint issues, so I think it was probably a mistake on their part, but it's an easy enough fix to make on our side, so here we go...

alanmcruickshank (approved):

PR #1200, opened 2021-07-12
Fix quote escapes in snowflake

alanmcruickshank:
This fixes an issue in the snowflake dialect where backslash escapes within single-quoted strings would fail. This was because the regex matcher in the snowflake dialect overrode the inherited regex with a simpler one. This resolves that.


test/fixtures/parser/snowflake/snowflake_escape.sql
```diff
@@ -0,0 +1,2 @@
+-- Backslash escapes work in snowflake
+select 'c\' ' as escaped, 'c\' '' ' as escaped_double
```
barrywhart, line 2: Is there a discrepancy between this test and the CHANGELOG? If I read this test case correctly, Snowflake dialect also needs to allow escaping single quotes by having two in a row `''`, but the CHANGELOG entry talks about fixing an issue with "backslash escaped single quoted strings", which would be `\'`.

Based on the change to the regex, do we need to test backlash followed by some other character?

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -37,7 +37,7 @@
     [
         # In snowflake, a double single quote resolves as a single quote in the string.
         # https://docs.snowflake.com/en/sql-reference/data-types-text.html#single-quoted-string-constants
-        RegexLexer("single_quote", r"'([^']|'')*'", CodeSegment),
+        RegexLexer("single_quote", r"'([^'\\]|\\.|'')*'", CodeSegment),
```
barrywhart, line 40: Reading regexes is hard. 😛  Is this changing single-quoted strings to allow a backlash followed by any other character?

barrywhart (commented):
Some questions, trying to make sure I understand the CHANGELOG, dialect change, and the code. Are they consistent with one another?

src/sqlfluff/dialects/dialect_snowflake.py
alanmcruickshank: In effect, yes. What the middle bit says is actually:
- Anything which isn't a `'` or a `\`.
- or a `\` followed by any single character.
- or a `''`.

The reason I've made the second entry quite broad, is to cover the case that someone enters an escape sequence but not a standard one (e.g. `\q`). In this case the current regex would match that and assume it was part of the string (the desired behaviour). If however we were more specific there, then the second case wouldn't catch it and the string literal would fail to parse entirely.

What we're trying to do here is parse the string literal, not accurately capture escape sequences - if that makes sense?

test/fixtures/parser/snowflake/snowflake_escape.sql
alanmcruickshank: I don't *think* so - but see if my explanation below on why makes sense.

barrywhart (approved):

PR #1213, opened 2021-07-18
Add dbt 0.20.0 to test suite

alanmcruickshank:
dbt 0.20.0 adds some big changes (some which I suspect break the dbt templater). This adds it to the test suite, and if anything breaks then I'll also add the changes to fix it.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = linting, doclinting, cov-init, py{36,37,38,39}, dbt01{7,8,9}-py{37,38}, cov-report, bench, mypy
+envlist = linting, doclinting, cov-init, py{36,37,38,39}, dbt0{17,18,19,20}-py{37,38}, cov-report, bench, mypy
 
 [testenv]
 passenv = CI CIRCLECI CIRCLE_* HOME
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -18,13 +18,14 @@ deps =
     dbt017: dbt==0.17.2
     dbt018: dbt==0.18.2
     dbt019: dbt==0.19.1
+    dbt020: dbt==0.20.0
```
barrywhart, line 21: Is there a way to also test with "latest version of dbt"? (Or would that even be a good idea?)

alanmcruickshank: I think we *could* by using something like `dbt>=0.20.0,<0.21.0` but I'm not sure we *should* 🤔 

What we trade in less maintenance, we potentially lose in unexpected failures if they break something. Ideally dbt would be stable enough that we could rely on things not breaking, but I don't know if that's a reasonable expectation at the moment.

barrywhart: This is something we are trying to solve with other dependencies at my job. I think it's useful to _know_ if a new version of a dependency breaks something, but to make it a "private" failure, i.e. don't allow (or at least, don't *encourage* end users) to use new versions until they're known to work.

For us, it's Great Expectations, which is a fast-moving project that frequently breaks things.

alanmcruickshank: > For us, it's Great Expectations, which is a fast-moving project that frequently breaks things.

If it makes you feel any better, I've been bitten by that one too 🤣

PR #1214, opened 2021-07-18
Issue 1207: Fix looping interaction between L008 and L030

### LINKED ISSUE #1207: COUNT dup in sqlfluff fix in 0.6.0 (from 0.5.3)
## Expected Behaviour

```
SELECT FLOOR(dt) ,count(*) FROM test
```
## Observed Behaviour

```
~ # sqlfluff fix --rules L008,L030 -f test.sql
WARNING    One fix for L008 not applied, it would re-cause the same error.
WARNING    One fix for L008 not applied, it would re-cause the same error.
WARNING    One fix for L008 not applied, it would re-cause the same error.
WARNING    One fix for L008 not applied, it would re-cause the same error.
WARNING    One fix for L008 not applied, it would re-cause the same error.
WARNING    One fix for L008 not applied, it would re-cause the same error.
WARNING    One fix for L008 not applied, it would re-cause the same error.
WARNING    One fix for L008 not applied, it would re-cause the same error.
WARNING    One fix for L008 not applied, it would re-cause the same error.
== [test.sql] FAIL
L:   1 | P:  19 | L008 | Commas should be followed by a single whitespace unless
                       | followed by a comment.
L:   1 | P:  19 | L008 | Commas should be followed by a single whitespace unless
                       | followed by a comment.
L:   1 | P:  19 | L030 | Inconsistent capitalisation of function names.
==== fixing violations ====
FORCE MODE: Attempting fixes...
Persisting Changes...
== [test.sql] PASS
```
fixed(?) sql.
```
SELECT FLOOR(dt) , COUNTCOUNTCOUNTCOUNTCOUNTCOUNTCOUNTCOUNTCOUNTCOUNTcount(*) FROM test
```

## Steps to Reproduce

on minimal environment.

```
docker run --rm -it python:3.9-alpine
pip install sqlfluff
```

## Version

```
~ # sqlfluff --version
sqlfluff, version 0.6.0
```

version 0.5.2 (SQLFluff Online version) run correctly, and 0.5.3 generate COUNT dup.

```
~ # python --version
Python 3.9.6
```
## Configuration

w/o .sqlfluff

barrywhart:
Resolves #1207


barrywhart:
This PR (which makes a couple of fixes to linter rules) causes a seemingly unrelated test to fail:
```
py.test -v 'test/rules/std_fix_auto_test.py::test__std_fix_auto[bigquery-002_templating]'
INFO     sqlfluff.linter:linted_file.py:289       - Skipping. Source space Value Error. i.e. attempted insertion within templated section.
```

Any ideas, @alanmcruickshank or others?

src/sqlfluff/rules/L008.py
```diff
@@ -38,16 +38,24 @@ def _eval(self, segment, raw_stack, **kwargs):
...
+            return None
+
+        cm2 = raw_stack[-1]
+        if cm2.name == "comma":
+            # comma followed by something that isn't whitespace?
+            if segment.name not in ["whitespace", "newline"]:
+                ins = WhitespaceSegment(raw=" ")
+                return LintResult(
+                    anchor=cm2, fixes=[LintFix("edit", segment, [ins, segment])]
+                )
```
alanmcruickshank, line 51: Structure looks good. Nitpick: `cm2` refers to _something-minus-2_ and in this case we're `raw_stack[-1]`. We should either call it `cm1` or maybe even just refer to the element in the raw stack directly. The `cm2` here I think is confusing.

Arguably the `cm1` and `cm2`  in the part below aren't much clearer, but I don't think they _need_ to be refactored as part of this PR.

alanmcruickshank (commented):
Small nitpick, but the fix looks good. I'll take a look at the fail later this week and see if I can work out why that error is happening 🤔

src/sqlfluff/rules/L008.py
barrywhart: Oh yes, I meant to do that. Will clean this up. It may have been possible to fix this without splitting the two cases, but it seemed good to avoid going any farther up `raw_stack` than necessary

barrywhart:
Thanks, @alanmcruickshank!

alanmcruickshank:
Found the issue with the test! The second clause of rule L008 was now improperly flagging on _non-whitespace_ segments, this meant that some other fixes were being ignored because they would _recause the previous error_ (which shouldn't actually be an error). I've updated the code (and added a little more logging).

barrywhart:
@alanmcruickshank or @NiallRees: Care to review this PR? Thanks!

barrywhart:
@alanmcruickshank: Thanks for looking into the bug!!

alanmcruickshank (approved):

PR #1221, opened 2021-07-21
skip annotation test on windows

alanmcruickshank:
This test fails on windows because of path issues. Given our automated tests are all on linux, and so is the environment that this is most likely to be called, I think this is fine to skip.

### DIFF
--- a/test/cli/commands_test.py
+++ b/test/cli/commands_test.py
@@ -8,6 +8,7 @@ import json
 import oyaml as yaml
 import subprocess
 import chardet
+import sys
 
 # Testing libraries
 import pytest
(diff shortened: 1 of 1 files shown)


test/cli/commands_test.py
```diff
@@ -522,6 +523,7 @@ def test__cli__command_lint_serialize_multiple_files(serialize):
         raise Exception
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="Paths won't match test on windows")
```
barrywhart, line 526: Looks fine. Would it be possible (and worthwhile) to modify the test so it works on Windows as well?

barrywhart (approved):

test/cli/commands_test.py
barrywhart: Heh, I think I wrote this test. I'll take a quick look to see about Windows compatibility.

barrywhart:
@alanmcruickshank: See PR #1223. With this change, the test should pass on any platform. Sorry about the broken test!

test/cli/commands_test.py
alanmcruickshank: Don't stress - I think it's a great test for unix style platforms - which are also the main targets for it, especially given the github action. I'm fine just skipping it for windows unless we fancy working out a way of properly normalising both sides of the comparison.

PR #1223, opened 2021-07-21
Do not check the path in the GitHub annotation test so it'll work on Windows

barrywhart:
Proposed alternative to #1221

### DIFF
--- a/test/cli/commands_test.py
+++ b/test/cli/commands_test.py
@@ -533,11 +533,11 @@ def test__cli__command_lint_serialize_github_annotation():
         ret_code=65,
     )
     result = json.loads(result.output)
-    for r in result:
-        r.pop("file")
     assert result == [
         {
             "annotation_level": "warning",
+            # Normalise paths to control for OS variance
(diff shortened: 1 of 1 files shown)


alanmcruickshank (changes requested):
I think it's still worth testing the file output so that we can make sure it still works for most users. In my mind, either we skip the test (as per #1221) or we call `normpath` on all the filenames on both side of the equation. Happy to go with the latter - but I'm also happy to go with the former. @barrywhart What do you reckon?

barrywhart:
I'll add `normpath()`. I knew there was a function which would do this, but I was too lazy yesterday to go and look it up. 😿

barrywhart:
@alanmcruickshank: This is ready for another review. The tests are taking a while to run, unfortunately. 🤔

alanmcruickshank (approved):
I like this approach. I've closed my other PR. This is failing linting checks, but otherwise looks good!

barrywhart:
I fixed the lint check (just needed to run Black). Merging...

PR #1239, opened 2021-07-31
Issue 939: Comma is removed when fixing query with multiple CTEs

### LINKED ISSUE #939: Comma is removed when fixing query with multiple CTEs
When fixing a query with multiple CTEs I find that the comma separating them is occasionally removed, resulting in invalid SQL. I'm using `dialect = snowflake` and a simple example query is:

```sql
WITH first_cte AS
(SELECT
    id
    , one
    FROM first),

second_cte AS
(SELECT
    id
    , two
    FROM {{ source('schema', 'table') }} )

SELECT
    id
    , one
    , two
FROM first_cte
LEFT JOIN second_cte
    ON first_cte.id = second_cte.id;
```

I've done some detective work and the comma is removed only when I set `comma_style=leading` for rule L019. 

Interestingly, if I manually replace the jinja macro `{{ source('schema', 'table') }}` with `schema_table` then the leading comma is correctly inserted and the fixed query is valid!

## Expected Behaviour
I'm expect that `sqlfluff fix ...` correctly removes the trailing command after the first CTE in the query above and places a leading comma before the second CTE.

## Observed Behaviour
The trailing comma is removed and no leading comma is inserted.

## Steps to Reproduce
1. Copy/paste the query into "bug.sql"
2. Create `.sqlfluff` file (see below) in the same directory 
3. Run `sqlfluff fix bug.sql` -> trailing comma is removed, no leading comma is inserted

**And to reproduce some of the odd behaviour I saw (in case it's helpful to locate the issue):**
- Replace `{{ source('schema', 'table') }}` with `schema_table` in the query
  - Trailing comma is correctly replaced with a leading comma

-  Reset `{{ source('schema', 'table') }}`, change `comma_style = leading` with `comma_style = trailing` in `.sqlfluff`
   - Trailing comma is correctly replaced with a leading comma

## Version
```bash
$ sqlfluff --version
sqlfluff, version 0.5.0

$ python --version
Python 3.8.5
```

## Configuration
```
[sqlfluff]
dialect = snowflake
rules = L019

[sqlfluff:rules]
comma_style = leading
```

## Extra info
Output from running `sqlfluff fix`:
```bash
$ sqlfluff fix bug.sql
== [bug.sql] FAIL
L:   5 | P:  13 | L019 | Found trailing comma. Expected only leading.
==== fixing violations ====
Attempting fixes...
Persisting Changes...
== [bug.sql] PASS
```

Resulting `.sql` file (notice the comma missing between the first and second CTE):
```sql
WITH first_cte AS
(SELECT
        id
        , one
        FROM first)

second_cte AS
  (SELECT
    id
    , two
    FROM {{ source('schema', 'table') }} )

SELECT id
        , one
        , two
FROM first_cte
LEFT JOIN second_cte
        ON first_cte.id = second_cte.id;
```

---
## Finally

I'd be willing to try to fix the code and open a PR. However, I'm very new to `sqlfluff` and I don't know if this is an easy fix or a hard one. Also, I would need some guidance about where to look and what to investigate.

barrywhart:
Resolves #939

### DIFF
--- a/src/sqlfluff/rules/L019.py
+++ b/src/sqlfluff/rules/L019.py
@@ -191,9 +191,9 @@ class Rule_L019(BaseRule):
                         fixes=[
                             LintFix("delete", anchor=last_comma_seg),
                             LintFix(
-                                "edit",
+                                "create",
                                 anchor=segment,
-                                edit=[last_comma_seg, new_whitespace_seg, segment],
+                                edit=[last_comma_seg, new_whitespace_seg],
                             ),
                         ],
                     )
--- a/test/rules/std_test.py
+++ b/test/rules/std_test.py
@@ -31,7 +31,7 @@ from test.fixtures.dbt.templater import (  # noqa
         # Check we get comma (with leading space/newline) whitespace errors
         # NB The newline before the comma, should report on the comma, not the newline for clarity.
         ("L005", "whitespace_errors.sql", [(2, 9)]),
-        ("L019", "whitespace_errors.sql", [(4, 1)]),
+        ("L019", "whitespace_errors.sql", {(2, 12), (4, 1)}),
         # Check we get comma (with incorrect trailing space) whitespace errors,
         # but also no false positives on line 4 or 5.
         ("L008", "whitespace_errors.sql", [(3, 12)]),
--- a/test/rules/std_test.py
+++ b/test/rules/std_test.py
@@ -31,7 +31,7 @@ from test.fixtures.dbt.templater import (  # noqa
         # Check we get comma (with leading space/newline) whitespace errors
         # NB The newline before the comma, should report on the comma, not the newline for clarity.
         ("L005", "whitespace_errors.sql", [(2, 9)]),
-        ("L019", "whitespace_errors.sql", {(2, 12), (4, 1)}),
+        ("L019", "whitespace_errors.sql", {(4, 1)}),
         # Check we get comma (with incorrect trailing space) whitespace errors,
         # but also no false positives on line 4 or 5.
         ("L008", "whitespace_errors.sql", [(3, 12)]),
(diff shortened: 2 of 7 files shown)


src/sqlfluff/rules/L019.py
```diff
@@ -191,9 +191,9 @@ def _eval(self, segment, raw_stack, memory, **kwargs):
                         fixes=[
                             LintFix("delete", anchor=last_comma_seg),
                             LintFix(
-                                "edit",
+                                "create",
                                 anchor=segment,
-                                edit=[last_comma_seg, new_whitespace_seg, segment],
```
barrywhart, line 196: Editing `segment` where the replacement **includes** segment` causes unpredictable behavior. I've seen this in other rules previously.

test/rules/std_test.py
```diff
@@ -31,7 +31,7 @@
         # Check we get comma (with leading space/newline) whitespace errors
         # NB The newline before the comma, should report on the comma, not the newline for clarity.
         ("L005", "whitespace_errors.sql", [(2, 9)]),
-        ("L019", "whitespace_errors.sql", [(4, 1)]),
+        ("L019", "whitespace_errors.sql", {(2, 12), (4, 1)}),
```
barrywhart, line 34: I'm not sure why the behavior changed here, but if you look at `whitespace_errors.sql`, the new behavior seems correct (unless I'm missing something).

pwildenhain (approved):
Makes sense! When I was first creating the fix for this rule I had no idea how fixes worked. This is great, thanks Barry!

barrywhart:
Thanks @pwildenhain! People at work are starting to complain about a few configuration issues and bugs with our SQLFluff tooling. This gives me a good ~excuse~ reason to work on SQLFluff now. 😛

PR #1240, opened 2021-08-01
L019 should specify "_works_on_unparsable = False" to avoid corrupting queries

barrywhart:


pwildenhain (approved):
👏🏻 Wonderful, great test as well ✅

PR #1249, opened 2021-08-06
Issue 1246: the "--nocolor" command-line option should suppress emoji output

### LINKED ISSUE #1246: Add the option to remove/change the emojis of the output
It would be nice if we can have a option to change/modify/remove the emojis of the output.

image

barrywhart:
Resolves #1246


src/sqlfluff/cli/commands.py
```diff
@@ -405,7 +405,7 @@ def lint(
             click.echo(cli_table(timing_summary[step].items()))
 
     if not nofail:
-        if not non_human_output:
+        if not non_human_output and not config.get("nocolor"):
```
reidwil-jb, line 408: Shouldn't this have some type of place holder even if `nocolor` is selected?

Thinking something like:
```python
if not non_numan_output:
  if not config.get("nocolor"):
    click.echo("All Finished 📜 🎉!!")
  else:
    click.echo("All finished!")
```

@barrywhart I love that someone got to this feature!

barrywhart: Great point! Making that change now...

CHANGELOG.md
```diff
@@ -7,14 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+
+- Support for primary index name, collect stats improvement, COMMENT statement for teradata dialect [#] (https://github.com/sqlfluff/sqlfluff/issues/)
```
barrywhart, line 11: Moved several items that had mistakenly been placed in 0.6.2. Oops!!

alanmcruickshank (approved):

PR #1252, opened 2021-08-08
Add DETERMINISTIC support for BigQuery

### LINKED ISSUE #1251: Support DETERMINISTIC in BgQuery User Defined Functions
-- Please add any additional labels that are related to the issue before submitting! --

BigQuery supports an optional [DETERMINISTIC](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_function_statement) keyword for User Defined Functions (UDFs):

```sql
CREATE FUNCTION
qs(
    y STRING
)
DETERMINISTIC
LANGUAGE js
AS " return y; "
```

The value is `DETERMINISTIC` or `NOT DETERMINISTIC`.

This can be a good performance enhancement to cache the function result so it's not re-executed when the function is called for another row with the same params.

Currently this fails linting with an unparsable error.

It looks like this is supported for MySQL but not for BigQuery so it looks easy enough to add so will submit a PR for this.

tunetheweb:
Fixes #1251 

Note technically this is only supported for `LANGUAGE js` but so is `OPTIONS` and it wasn't restricted when it was implemented, so I kept this simple and did the same for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 ### Changed
 
+- Support for DETERMINISTIC user defined functions [#1251](https://github.com/sqlfluff/sqlfluff/issues/1251)
+
 ## [0.6.2] - 2021-07-22
 ### Added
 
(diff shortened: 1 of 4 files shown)


barrywhart:
The code looks good. Can you update the "Changed" section at the top of CHANGELOG.md?

Thanks for the contribution!

tunetheweb:
> The code looks good. Can you update the "Changed" section at the top of CHANGELOG.md?
> 

Done!

> Thanks for the contribution!

No probs. FYI, working on a couple more too.

Thanks for the project!

barrywhart (approved):

PR #1254, opened 2021-08-08
Handle more parameter names in BigQuery

### LINKED ISSUE #1253: BigQuery Parameter Names allow more characters than currently permitted
We have an issue whereby the following are not accepted in BigQuery linting:

```sql
CREATE TEMP FUNCTION
qs(
    y STRING
)
RETURNS STRUCT<_product_id ARRAY<INT64>>
LANGUAGE js AS """
    CODE GOES HERE
"""
```

The underscore in `_product_id` is unparsable.

The [BigQuery documentation on identifiers](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical) has this to say:

> Unquoted identifiers must begin with a letter or an underscore character. Subsequent characters can be letters, numbers, or underscores.

Similarly, this doesn't work:

```sql
CREATE TEMP FUNCTION
qs(
    y STRING
)
RETURNS STRUCT<`$=` ARRAY<INT64>>
LANGUAGE js AS """
    CODE GOES HERE
"""
```

even though the documentation has the following to say:

> Quoted identifiers must be enclosed by backtick (`) characters.
> - Quoted identifiers can contain any character, such as spaces or symbols.
> - Quoted identifiers cannot be empty.
> - Quoted identifiers have the same escape sequences as string literals.

tunetheweb:
Fixes #1253


barrywhart (approved):
Nice work -- thanks!

Two Barry's working on the BigQuery dialect. 😎

PR #1263, opened 2021-08-09
Support COUNT(0) preference over COUNT(*) or COUNT(1)

### LINKED ISSUE #1260: Support `COUNT(0)` as well as `COUNT(1)`
L047 allows enforcing consistency of use of `COUNT(*)` or `COUNT(1)`.

However some teams use `COUNT(0)` instead of either. Personally `COUNT(1)` has always made more sense to me, but that's by the by as sometimes you have to swallow your own preferences to follow team style.

So I would like to add the option to be able to enforce `COUNT(0)`. I could see this being done in two ways:
- The easiest would be to add a `prefer_count_0` flag similar to `prefer_count_1`, and then decide a precedence if both are set (I'd suggest `prefer_count_1` for backwards compatibility).
- Perhaps a cleaner way (in terms of config anyway) would be to to change this to a `prefer_count="1"` option, that could also be set to `"0"`, though that would be a breaking change (unless you supported both - in which case you might as well just go with the first option).

Any thoughts on this?

Happy to tackle this one myself btw.

tunetheweb:
Fixes #1260 

I've gone with the simpler option of two separate configs to avoid breaking backwards compatibility. So you can have:

```yml
prefer_count_1: True
```

or:

```yml
prefer_count_0: True
```

And if both are specified:
```yml
prefer_count_1: True
prefer_count_0: True
```

Then `prefer_count_1` takes precedence (no matter what order they are specified in).

As I say, ideally you'd not be able to have contradictory params like this, so let me know if you think we should introduce a breaking change here instead.

Barry

P.S. I swear I'm not trying to annoy you @barrywhart - I just seem to get each next PR finished just at the exact same time as you merge the last one! 😁

### DIFF
--- a/src/sqlfluff/rules/L047.py
+++ b/src/sqlfluff/rules/L047.py
@@ -12,8 +12,12 @@ from sqlfluff.core.rules.doc_decorators import (
 class Rule_L047(BaseRule):
     """Use consistent syntax to express "count number of rows".
 
-    COUNT(*) and COUNT(1) are equivalent syntaxes in many SQL engines
-    due to optimizers interpreting these instructions as
+    Note:
+        If both `prefer_count_1` and `prefer_count_0` are set to true
+        then `prefer_count_1` has preference.
+
+    COUNT(*), COUNT(1), and even COUNT(0) are equivalent syntaxes in many SQL
+    engines due to optimizers interpreting these instructions as
     "count number of rows in result".
 
     The ANSI-92_ spec mentions the COUNT(*) syntax specifically as
@@ -24,8 +28,9 @@ class Rule_L047(BaseRule):
 
     So by default, SQLFluff enforces the consistent use of COUNT(*).
 
-    If the SQL engine you work with, or your team, prefers COUNT(1)
-    over COUNT(*) you can configure this rule to consistently enforce COUNT(1).
+    If the SQL engine you work with, or your team, prefers COUNT(1) or COUNT(0)
+    over COUNT(*) you can configure this rule to consistently enforce your
+    preference.
 
     .. _ANSI-92: http://msdn.microsoft.com/en-us/library/ms175997.aspx
 
@@ -38,7 +43,8 @@ class Rule_L047(BaseRule):
         from table_a
 
     | **Best practice**
-    |   Use count(*) unless specified otherwise by config ``prefer_count_1``
+    | Use ``count(*) unless specified otherwise by config ``prefer_count_1``,
+    | or ``prefer_count_0`` as preferred.
 
     .. code-block:: sql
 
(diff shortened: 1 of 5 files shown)


src/sqlfluff/rules/L047.py
```diff
@@ -38,7 +43,8 @@ class Rule_L047(BaseRule):
         from table_a
 
     | **Best practice**
-    |   Use count(*) unless specified otherwise by config ``prefer_count_1``
+    | Use ``count(*) unless specified otherwise by config ``prefer_count_1``,
```
barrywhart, line 46:
```suggestion
    | Use ``count(*)`` unless specified otherwise by config ``prefer_count_1``,
```

src/sqlfluff/rules/L047.py
```diff
@@ -87,7 +93,64 @@ def _eval(self, segment, **kwargs):
...
+                expression_content = [
+                    seg for seg in f_content[0].segments if not seg.is_meta
+                ]
+                if (
+                    len(expression_content) == 1
+                    and expression_content[0].is_type("literal")
+                    and expression_content[0].raw == "0"
+                ):
+                    return LintResult(
+                        anchor=segment,
```
barrywhart, line 106: I notice it says this line is not covered by tests?

barrywhart:
No worries -- it's a great "problem" to have! I posted a couple comments. Will review more thoroughly later!

src/sqlfluff/rules/L047.py
tunetheweb: Just submitted some more tests in https://github.com/sqlfluff/sqlfluff/pull/1263/commits/7ed3c89982b08b990499425ce15245724fb80559. will monitor to make sure coverage is back to normal.

tunetheweb: Looks better now.

src/sqlfluff/rules/L047.py
```diff
@@ -87,7 +93,64 @@ def _eval(self, segment, **kwargs):
                         ),
                     ],
                 )
-            if not self.prefer_count_1 and f_content[0].is_type("expression"):
+            elif self.prefer_count_1 and f_content[0].is_type("expression"):
```
barrywhart, line 96: Looking at this code, it seems pretty repetitive now -- i.e. the code for `count(0)` and `count(1)` is often identical. Do you think we could DRY this up (i.e. https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)?

tunetheweb: You were right. This could be simplified - a lot!

barrywhart (approved):
Nice work! I ❤️ the cleanup/dedupe work you did. 🚀

tunetheweb:
> Nice work! I ❤️ the cleanup/dedupe work you did. 🚀

Just used what was there already basically. Took me a bit to figure out how it worked to be honest.

barrywhart:
Yep. I find all the rule code a little dense. It's not super hard to read, but I have to stare at it or run it in the debugger so I can see the element objects

Issue #1233, opened 2021-07-27
Allow plugins to dynamically register themselves


matthieucan:
The current plugin registration process uses pluggy's `load_setuptools_entrypoints`, which enforces plugins to be packaged and installed in a Python environment to be able to be seen by sqlfluff.

It would be great if any piece of code could dynamically register plugins, e.g. by calling `plugin_manager.register(some_module)`.

The `get_plugin_manager()` function in `src/sqlfluff/core/plugin/host.py` returns a new instance of a plugin manager at every call. Ideally, it should return a singleton or be called only once during sqlfluff execution. I realized this while exploring monkey-patching options to dynamically register a plugin.

My use-case for this is as follows:
- I'm running sqlfluff through its Python API, i.e. I call `sqlfluff.lint()`, because I want to apply transformations to the SQL content before linting it.
- It's not convenient to have to create a separate Python package for the plugin, due to the name constraints (`sqlfluff-plugin-foo`).

Thank you for your consideration!

barrywhart:
@matthieucan: It sounds like you have a pretty good idea how to implement this. Would you be interested in submitting a PR to make the change?

matthieucan:
@barrywhart Yes, I'd be interested :)
What do you think of the singleton approach? Something like saving the plugin manager at the module level, so always the same instance is used, while exposing it to the outside so plugins can call any registering function on it?

barrywhart:
It sounds sensible to me. I wonder if there's a reason it wasn't written that way initially. 🤔

barrywhart:
I found an interesting discussion from some of the core maintainers of `pluggy`, the plugin framework used by SQLFluff. This [comment](https://github.com/pytest-dev/pluggy/issues/130#issuecomment-381459376) seems the most relevant:
>the PluginManager is not supposed to be a singleton. Most often the way you see it's usefulness is in projects that use pluggy whose dependencies also use pluggy resulting in > 1 pluggy.PluginManager instances being used simultaneously. I don't think I've yet seen a project using pluggy that has more than one PluginManager internally although conceivably you could...

This is a bit "deep" for me to try and understand, but it sounds like in practice, making it a singleton is fine (even though not the "intended" usage) as long as the entire package using `pluggy`) (i.e. SQLFluff) isn't trying to support multiple, distinct sets of plugins.

[...]

PR #1264, opened 2021-08-10
Move the dbt templater into a plugin

### LINKED ISSUE #1236: Make templaters installable as plugins

alanmcruickshank:
This resolves #1236.

As discussed before - this extracts the dbt templater into a plugin. It's still within the main project, but from here could be extracted to it's own repo (which I think it eventually should).

This also moves the handling on the plugin manager to the `FluffConfig` as discussed in #1233 . Would be interested in feedback on this approach.


plugins/sqlfluff-templater-dbt/src/dbt_templater/templater.py
```diff
@@ -1,379 +1,378 @@
...
+        super().__init__(**kwargs)
+
+    def config_pairs(self):
+        """Returns info about the given templater for output by the cli."""
+        return [("templater", self.name), ("dbt", self.dbt_version)]
+
+    @cached_property
+    def dbt_version(self):
+        """Gets the dbt version."""
+        from dbt.version import get_installed_version
```
barrywhart, line 51: Now that it's a plugin, would it be safe to assume that dbt is installed and (thus) move the dbt imports to the top of the file?

plugins/sqlfluff-templater-dbt/src/dbt_templater/templater.py
```diff
@@ -1,379 +1,378 @@
...
+        )
+
+    @staticmethod
+    def _check_dbt_installed():
+        try:
+            import dbt  # noqa: F401
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError(
+                "Module dbt was not found while trying to use dbt templating, "
+                "please install dbt dependencies through `pip install sqlfluff[dbt]`"
```
barrywhart, line 225: Is this error message still accurate? (The part about `sqlfluff[dbt]`?)

src/sqlfluff/core/config.py
```diff
@@ -429,15 +438,32 @@ def __init__(
...
             self._configs["core"]["templater"]
         )
 
+    def __getstate__(self):
+        # Copy the object's state from self.__dict__ which contains
+        # all our instance attributes. Always use the dict.copy()
+        # method to avoid modifying the original state.
+        state = self.__dict__.copy()
+        # Remove the unpicklable entries.
+        del state["_plugin_manager"]
```
barrywhart, line 455: IIUC, this would prevent using "not installed" linter rules in combination with `--processes`. See #1233. I'm not sure we need to solve this, but maybe need to document it when #1233 is implemented.

src/sqlfluff/core/config.py
```diff
@@ -429,15 +438,32 @@ def __init__(
...
+        return state
+
+    def __setstate__(self, state):
+        # Restore instance attributes
+        self.__dict__.update(state)
+        # NB: We don't reinstate the plugin manager, but this should only
+        # be happening between threads where the plugin manager should
```
barrywhart, line 462:
```suggestion
        # be happening between threads or processes where the plugin manager should
```

src/sqlfluff/core/config.py
```diff
@@ -429,15 +438,32 @@ def __init__(
...
+        # probably be fresh in any case.
+        # NOTE: This means that registering user plugins directly is not
+        # thread safe.
```
barrywhart, line 465: To me, "not thread safe" means it's subject to race conditions. In this case, it's more that they'll be lost in other processes or threads. Maybe clarify this?

src/sqlfluff/core/config.py
```diff
@@ -484,9 +515,35 @@ def from_kwargs(
             overrides["rules"] = ",".join(rules)
         return cls(overrides=overrides)
 
+    def get_templater(self, templater_name="jinja", **kwargs):
+        """Fetch a templater by name."""
+        templater_lookup = {
+            templater.name: templater
+            for templater in chain.from_iterable(
+                self._plugin_manager.hook.get_templaters()
+            )
```
barrywhart, line 524: Teeny nitpick: It's tempting to put `if templater.name == templater_name` here, so the dictionary would at most have the one entry we want. But I think that's my OCD talking. Why bother?!?! Sorry... 🤪

src/sqlfluff/core/rules/__init__.py
```diff
@@ -25,8 +21,25 @@
...
+        for rule in plugin_rules:
+            std_rule_set.register(rule)
+
+    return std_rule_set
+
+
 def get_ruleset(name: str = "standard") -> RuleSet:
     """Get a ruleset by name."""
-    lookup = {std_rule_set.name: std_rule_set}
+    std_rules = _load_standard_rules()
```
barrywhart, line 42: Will rule plug-ins be discovered and loaded in each child process or thread? (Context: I want to be know if custom rules and `--processes` work correctly together.)

src/sqlfluff/core/config.py
barrywhart: Random thought: I wonder if there may be a way to convert the plugin manager instance to a pickleable format, e.g. for each plugin, get its module and class/function name, then on the other side, create an empty plug-in manager and build it up again. It wouldn't be guaranteed to work in every case, but my intuition is that it could handle 99% of cases. Worth creating an issue for this? (Or adding these notes to #1233?)

barrywhart (commented):
Looks good so far. A few small suggestions and observations.

plugins/sqlfluff-templater-dbt/src/dbt_templater/templater.py
alanmcruickshank: yes good point

plugins/sqlfluff-templater-dbt/src/dbt_templater/templater.py
alanmcruickshank: I've just removed this now - if we're assuming that dbt is now installed because it's a requirement - then this will never be called.

src/sqlfluff/core/config.py
alanmcruickshank: Agreed. I actually wonder whether it might be more practical to pickle any use defined plugins and pass them to the processes rather than passing around the plugin manager? It would need to be implemented with #1233 but might be more practical. For now I'll add in a warning in the code.

src/sqlfluff/core/rules/__init__.py
alanmcruickshank: right now they won't - see comments above. I think that would probably need to be addressed as part of a seperate PR.

CHANGELOG.md
```diff
@@ -15,12 +15,14 @@ Note: Changes are now automatically tracked in [GitHub](https://github.com/sqlfl
 
 ### Changed
 - Fix typo in the in the wild page [#1285](https://github.com/sqlfluff/sqlfluff/pull/1285)
+- Moved the dbt templater into a plugin rather than part of the core. This should now be installed
```
barrywhart, line 18: With the addition of the GitHub action, CHANGELOG no longer needs updating prior to a release.

CHANGELOG.md
```diff
@@ -15,12 +15,14 @@ Note: Changes are now automatically tracked in [GitHub](https://github.com/sqlfl
...
+  using `pip install sqlfluff-templater-dbt` rather than `pip install sqlfluff[dbt]`.
 
 ## [0.6.3] - 2021-08-16
 ### Added
 
-- Moved the dbt templater into a plugin rather than part of the core. This should now be installed
-  using `pip install sqlfluff-templater-dbt` rather than `pip install sqlfluff[dbt]`.
+- Support for primary index name, collect stats improvement, COMMENT statement for teradata dialect [#] (https://github.com/sqlfluff/sqlfluff/issues/)
```
barrywhart: Are these two new items related to this PR?

barrywhart (approved):
Looks good!

plugins/sqlfluff-templater-dbt/src/dbt_templater/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""Defines the hook endpoints for the dbt templater plugin."""
+
+from dbt_templater.templater import DbtTemplater
+from sqlfluff.core.plugin import hookimpl
+
+
+@hookimpl
+def get_templaters():
+    """Get templaters."""
+    return [DbtTemplater]
```
dmateusp, line 10: if I understand correctly @alanmcruickshank, this PR makes templaters pluggable in general?
If that's the case, maybe it's worth reflecting in the PR title/description so it appears in the CHANGELOG, and also updating the plugin Documentation?

I've seen a few issues in the past with people having some very custom ways of templating SQL, this PR might address those too

dmateusp (approved):
LGTM in general, however I'm not sure about moving the code to a separate repository in the Org, what are the upsides? I'm just thinking it's more CI to maintain

CHANGELOG.md
tunetheweb: Removed during merging to `main`. Resolving.

alanmcruickshank:
> LGTM in general, however I'm not sure about moving the code to a separate repository in the Org, what are the upsides? I'm just thinking it's more CI to maintain

The intent would be to reduce the CI load on each one individually and allow different release cycles. However, this PR is only the first step in that direction. Once this is merged, I'd like to segregate the tests, and then finally we can decide on whether to extract it to a separate repo.

barrywhart:
About separating the repos (or not), let's consider the current state and maturity of the project. Recently, we've been fixing some tricky bugs involving templating. Dbt is often the environment where those issues were discovered and reported, even if they were core linter or Jinja template bugs. Let's be sure we avoid creating a situation where important tests for core functionality are located elsewhere and get overlooked, leading to regressions in future releases.

PR #1266, opened 2021-08-10
Add optional end of line operators config for L007

### LINKED ISSUE #1261: Change L007 to specify end of line as well as beginning of line consistency
L007 enforces "Operators near newlines should be after, not before the newline."

That is this:

```sql
SELECT
    a
    + b
FROM foo
```

is preferred over this:

```sql
SELECT
    a +
    b
FROM foo
```

I'd like to suggest an `operator-before-end-of-line` config param (defaulted to `False` for backwards compatibility) which can be used the consistently enforce the latter behaviour for teams that prefer that.

If accepted, we should reword this rule to "Operators near newlines should be consistent in regards to placement before or after the newline (default is to prefer before)."

Happy to look at this one myself.

tunetheweb:
Fixes #1261


alanmcruickshank (approved):
Nice! Thanks for the great test coverage too.

Issue #1255, opened 2021-08-08
Support BigQuery function member fields


tunetheweb:
BigQuery supports accessing function member fields for functions returning a `STRUCT` (usually UDFs):

```sql
SELECT testFunction(a).id
FROM table1
```

But that `.id` causes the parse to bomb out.

PR #1269, opened 2021-08-12
Add wildcard support for function fields access in BigQuery

tunetheweb:
A further enhancement to #1255 to allow wildcard support as well:

```sql
SELECT testFunction(a).*
FROM table1
```

### DIFF
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -243,9 +243,16 @@ class FunctionSegment(BaseSegment):
                     ephemeral_name="FunctionContentsGrammar",
                 )
             ),
+            # Functions returning STRUCTs in BigQuery can have the fields
+            # elements referenced (e.g. ".a"), including wildcards (e.g. ".*")
+            # Note: we currently don't support field of fields (e.g. ".a.b", or ".a.b.c")
+            # But bad practice to use them directly, as can't guarantee they exist.
             Sequence(
                 Ref("DotSegment"),
-                Ref("ParameterNameSegment"),
+                OneOf(
+                    Ref("ParameterNameSegment"),
+                    Ref("StarSegment"),
+                ),
                 optional=True,
             ),
         ),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support alternative indentation for USING and ON clauses [#1250](https://github.com/sqlfluff/sqlfluff/issues/1250)
 - Support COUNT(0) preference over COUNT(*) or COUNT(1) [#1260](https://github.com/sqlfluff/sqlfluff/issues/1260)
 - Support for BigQuery "CREATE table OPTIONS ( description = 'desc' )" [#1205](https://github.com/sqlfluff/sqlfluff/issues/1205)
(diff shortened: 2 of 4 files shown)


barrywhart (approved):
Looks good!

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -243,9 +243,16 @@ class FunctionSegment(BaseSegment):
...
+            # Note: we currently don't support field of fields (e.g. ".a.b", or ".a.b.c")
+            # But bad practice to use them directly, as can't guarantee they exist.
             Sequence(
                 Ref("DotSegment"),
-                Ref("ParameterNameSegment"),
+                OneOf(
+                    Ref("ParameterNameSegment"),
+                    Ref("StarSegment"),
+                ),
                 optional=True,
```
tunetheweb, line 256: Note that I tried to add this too, just for completeness sake, by adding an `AnyNumberOf` section:

```py
            Sequence(
                Ref("DotSegment"),
                AnyNumberOf(
                    Sequence(
                        Ref("ParameterNameSegment"),
                        Ref("DotSegment"),
                    ),
                    optional=True,
                 ),
                OneOf(
                    Ref("ParameterNameSegment"),
                    Ref("StarSegment"),
                ),
                optional=True,
            ),
```

However the parser didn't like this, as when I added a test case with:

```sql
SELECT testFunction(a).b.c
FROM table1
```

and:

```yml
file:
  statement:
    select_statement:
      select_clause:
        keyword: SELECT
        select_clause_element:
          function:
            function_name:
              function_name_identifier: testFunction
            bracketed:
              start_bracket: (
              expression:
                column_reference:
                  identifier: a
              end_bracket: )
            dot: .
            parameter: b
            dot: .
            parameter: c
      from_clause:
        keyword: FROM
        from_expression:
          from_expression_element:
            table_expression:
              table_reference:
                identifier: table1
```

The parser seemed to completely ignore `b.` and so complained of a mismatch, and I'm not sure why:

```
E                           ('dot',
E                            '.'),
E                           ('parameter',
E             +              'b'),
E             +             ('dot',
E             +              '.'),
E             +             ('parameter',
E                            'c'))),)))),
E                     ('from_clause',
```

But to be honest I think use of nested field names directly from the function return is pretty bad practice (what if `b` is null, and you try to access `b.c`?) and so is likely not used much (I certainly don't use this), even if BigQuery technically supports it.

So I thought I'd keep it simple and just add the wildcard support (which my project does use on occasion - even if it's not great practice in itself).

barrywhart: Interesting! The parser does have some quirks, which I don't understand and thus tend to surprise me. I'm curious if something like the following might work instead (replacing the `AnyNumberOf` bit above):
```
                Delimited(
                    Ref("ParameterNameSegment"),
                    delimiter=Ref("DotSegment")
               ),
```

barrywhart:
@tunetheweb: Yep, I've got some insomnia on this side of the Atlantic, and I was already working on something. I'm about to go try and sleep a couple more hours. 🥱

src/sqlfluff/dialects/dialect_bigquery.py
tunetheweb: Tried it and it caused a few other issues. But tweaking, basically ended up back with the same error.

Anyway, not going to worry about it as, as I said, I'd imagine usage of this is rare and so don't want to over complicated the code for something that might never even be used.

If someone does use it then we can look into it further then and we have the Note there to explain it.

Issue #1277, opened 2021-08-14
Better ways to handle YML test files


tunetheweb:
When adding a SQL fixture test file you need a corresponding YAML file. These can be automatically generated with the following command:

```
python test/generate_parse_fixture_yml.py
```

New contributors may not realise this (I didn’t!) and may try to craft these manually by hand. This can lead to slight differences and even confusion and errors (see #1273 ).

Can we handle this better? Two thoughts immediately spring to mind:

If the YML files can be automatically generated then why do we need them in source control? Can’t we generate as part of `tox` and gitignore them so they are not committed? I don’t know `tox` myself but presume this would be possible?

However, if for some reason they can’t always be automatically generated, then an alternative solution would be to add a CI check on pull request to confirm any submitted YAML file matches the auto generated one. It would be a simple enough GitHub action to checkout the code, build the YAML files with above command, and then check if the branch is clean and fail the test if it is not. This should avoid any bad YAML files getting into the source in future. I’d be happy enough to create this if this is the route we wanna go, though the first option is the better one in my mind.

Thoughts?

barrywhart:
They're in source control in order to catch potential (mainly accidental) changes in parsing behavior.

I think when generating these files, we should add a checksum that makes it impractical to manually generate or edit the files. Then, as you say, the tests that read the YAMLs would check not only the data but also the checksum.

tunetheweb:
> They're in source control in order to catch potential (mainly accidental) changes in parsing behavior.

Ah OK that makes sense.

> I think when generating these files, we should add a checksum that makes it impractical to manually generate or edit the files. Then, as you say, the tests that read the YAMLs would check not only the data but also the checksum.

What do you think of my second suggestion? Probably a lot easier than the checksum but gives same protection. And you’d basically need that check anyway to double check someone’s not checking in a copied checksum. It also allows people to still manually handcraft them if they are complete sadists 😁

barrywhart:
Your idea sounds good, too. I would suggest one change: could we do this check in the test itself (`test/dialects/dialects_test.py::test__dialect__base_parse_struct`) rather than a GitHub action? I think that'd be easier for people to work with. For example, people often run the tests locally (before committing or pushing their changes), so they could detect any issues sooner.

tunetheweb:
[...]

Issue #1273, opened 2021-08-12
Test suite doesn't handle some multiples in YAML files


tunetheweb:
I noticed this in https://github.com/sqlfluff/sqlfluff/pull/1269#discussion_r687560810 but also stumbled across it again in a more recent PR.

To repeat edit the `test/fixtures/parser/bigquery/select_mixture_of_array_literals.sql` file and change this line:

```sql
    ['a'] AS strcol1,
```

to this:

```sql
    ['a', 'b'] AS strcol1,
```

Still valid SQL right?

We also change the equivalent YAML file from this:

```yml
            array_literal:
              start_square_bracket: '['
              expression:
                literal: "'a'"
              end_square_bracket: ']'
```

to this:

```yml
            array_literal:
              start_square_bracket: '['
              expression:
                literal: "'a'"
              comma: ','
              expression:
                literal: "'b'"
              end_square_bracket: ']'
```

We then run the test suite:

```
tox -e cov-init,py38,cov-report,linting
```

The SQL test completes fine (it's valid SQL after all):

```
test/dialects/dialects_test.py::test__dialect__base_file_parse[bigquery-select_mixture_of_array_literals.sql] ✓                                                                            35% ███▌      
 ```

However the YAML check later fails:

```py
――――――――――――――――――――――――――――――――――――――― test__dialect__base_parse_struct[bigquery-select_mixture_of_array_literals.sql-True-select_mixture_of_array_literals.yml] ――――――――――――――――――――――――――――――――――――――――

dialect = 'bigquery', sqlfile = 'select_mixture_of_array_literals.sql', code_only = True, yamlfile = 'select_mixture_of_array_literals.yml', yaml_loader = <function load_yaml at 0x7feb35a62940>

    @pytest.mark.parametrize("dialect,sqlfile,code_only,yamlfile", parse_structure_examples)
    def test__dialect__base_parse_struct(
        dialect, sqlfile, code_only, yamlfile, yaml_loader
    ):
        """For given test examples, check parsed structure against yaml."""
        # Load the right dialect
        config = FluffConfig(overrides=dict(dialect=dialect))
        # Load the SQL
        raw = load_file(dialect, sqlfile)
        # Lex and parse the file
        tokens, _ = Lexer(config=config).lex(raw)
        parsed = Parser(config=config).parse(tokens)
        # Load the YAML
        res = yaml_loader(make_dialect_path(dialect, yamlfile))
        if parsed:
[...]

PR #1241, opened 2021-08-01
Issue 1205: BigQuery support for CREATE table OPTIONS ( description = 'desc' )

barrywhart:
Resolves #1205

PR #1279, opened 2021-08-14
Issue 1277: Enforce that YML test files are computer generated and not edited

barrywhart:


test/generate_parse_fixture_yml.py
```diff
@@ -3,32 +3,40 @@
...
-from sqlfluff.core import FluffConfig
 from sqlfluff.cli.commands import quoted_presenter
 
-from dialects.parse_fixtures import get_parse_fixtures, load_file
+from conftest import compute_parse_tree_hash, parse_example_file
+from dialects.parse_fixtures import get_parse_fixtures
 
 yaml.add_representer(str, quoted_presenter)
 
-parse_success_examples, _ = get_parse_fixtures()
```
barrywhart, line 14: I refactored this script, splitting it into functions and moving some reusable code to `conftest.py`, where it is also used by `test/dialects/dialects_test.py::test__dialect__base_parse_struct`.

test/conftest.py
```diff
@@ -37,6 +43,36 @@ def process_struct(obj):
         raise TypeError(f"Not sure how to deal with type {type(obj)}: {obj!r}")
 
 
+def parse_example_file(dialect, sqlfile):
```
barrywhart, line 46: Extracted these functions from `test/generate_parse_fixture_yml.py`, which are now used both there and by `test/dialects/dialects_test.py::test__dialect__base_parse_struct`.

test/conftest.py
```diff
@@ -14,6 +19,10 @@
     CodeSegment,
 )
 from sqlfluff.core.templaters import TemplatedFile
+from test.dialects.parse_fixtures import load_file
+
+# When writing YAML files, double quotes string values needing escapes.
+oyaml.add_representer(str, quoted_presenter)
```
barrywhart, line 25: Moved this code here from `test/generate_parse_fixture_yml.py` to ensure consistent YAML quoting format between that script and the test (`test/dialects/dialects_test.py::test__dialect__base_parse_struct`) that verifies those files.

test/generate_parse_fixture_yml.py
```diff
@@ -1,34 +1,40 @@
...
+            )
             yaml.dump(r, f, default_flow_style=False)
         else:
             f.write("")
+
+
+def main():
+    """Find all example SQL files, parse and create YAML files."""
+    parse_success_examples, _ = get_parse_fixtures()
+    with multiprocessing.Pool(multiprocessing.cpu_count()) as pool:
```
barrywhart, line 34: Introduced multiprocessing so this script can process the `.sql` files more quickly. On my Mac, this change results in a ~3.5x speedup.

tunetheweb: Given this, are you 100% sure this PR is still the right approach?

With this speed up we could add this command to the tox.ini file with less of an impact and add the GitHub action as a fall back check just in case any contributor didn’t run it first.

This PR seems a lot of effort and change. If you still think it’s a good change then will bow to your better knowledge here but thought I’d just ask the question in case yours down the rabbit hole here.

barrywhart: I'm not concerned about the amount of change -- it's mostly just moving code around with slight changes in ways that improve the maintainability. (TBH, I think the `test/generate_parse_fixture_yml.py` script began with copy-pasted code from other places, and this PR cleans that up a bit.) 

Can you remind me what the `tox` change would look like and how it would impact contributors' workflow? My thinking (possibly wrong) is that it would change the workflow for people who are already familiar with it. I actually don't use `tox` at all for SQLFluff development, but I don't know if that's common. I have manually created a virtualenv using `pyenv`. I use that for all my local development and testing, and I rely on the GitHub builds to do all the other testing (different Python versions, etc.)

barrywhart: BTW, this PR is**tiny** compared to some of mine or Alan's. 😅 One reason I contribute to this project is because I often don't write much code in my real job. (I also contribute because it's useful to my real job. Kind of a win-win for me and my employer.)

tunetheweb: So my workflow is the following:

- Generate a virtualenv (only needs to be done once when first setting up the repo)
- Activate that virtual env (needs to be done each time you open a new shell):
```
source .venv/bin/activate
```
-  Create branch
- Make code change
- Generate YAML files (this is the step I missed in past as did them by hand - it also takes a minute or two, but maybe not with above change?):
```
```
- Run the quicker dialect tests only (I only discovered this command recently as hidden in a deeper README):
```
pytest test/dialects/dialects_test.py
```
- When fully finished, run the full tests for one python env only (takes a good few mins, even for one python version, so only really wanna run this when pretty much done):
```
tox -e cov-init,py38,cov-report,linting
```
- Check file changed, and config you're happy with the changes, and then commit them.
- Push to GitHub.
- Open PR.
- GitHub will automatically run all tests for ALL supported python versions in parallel.

I would propose the one line config change to `tox` as shown in https://github.com/sqlfluff/sqlfluff/issues/1277#issuecomment-898891659 to generate the YAML files when doing the full `tox` run. If people use `tox` before opening PR (like I do in above workflow) then they will automatically get the updated YAML files when they run it as so they will be included in the PR. If they don't, then that's fine too. And people can continue to run the generate command themselves if they want.

Either way, when they submit the PR a new GitHub Action will check the YAML files are up to date so if a contributor forgot to run it (or didn't know they had to), and it would produce any changes if it was run, then they will be informed and can run it manually and add to PR. If they always run it as part of their workflow (either manually, or via `tox`) then this check will never fail so is not additional effort. This would be a simple GitHub Action workflow to checkout the code, setup the python env for one python version (e.g. 3.8), run the YAML generation command, then use `git status` to see if the branch has any uncommitted changes and, if so, raise an alert. I'm happy to create this workflow as quite a lot of experience with GitHub Actions.

Does that make sense? Happy to answer any questions of create a sample PR to show you.

tunetheweb: > One reason I contribute to this project is because I often don't write much code in my real job. (I also contribute because it's useful to my real job. Kind of a win-win for me and my employer.)

Oh I hear you there! Same reasons I do open source in my spare time 😀

tunetheweb: Btw what would your workflow be with your change?

Is it that running `pytest` (either directly or via `tox`) fails if the YAML generation would have produced changes (i.e. you didn’t run it)? And so that would be caught by the existing checks on PRs and also if anyone runs `pytest` offline? If so, then that’s a good workflow too (maybe even a better one!) and works for me too. Just saying if that involves a lot of change and is much more complicated to implement, then there is the easier option of my proposal.

But there may be additional benefits to your change and would work better for those not using `tox` but running `pytest` directly.

barrywhart: Yes, I'm pretty sure (but not 100% sure) that running pytest without running the script catches me if I forget to run the script.

IIUC, adding this step to the tox config would make it mandatory to also add the GitHub Action check as well, to ensure the files are up to date. If I'm correct that pytest already detects this, I don't see how this additional work will add any benefit.

Would you be willing to test the pytest approach to try and confirm or deny that it detects problems? The most common issues are:
* SQL file with no YAML file
* SQL and YAML files out of sync

Less common, but possible:
* YAML file with no SQL file
* Am I overlooking any others?

We may still decide to add the tox and GitHub steps, but having this background info will help us decide. I'd like to hear from @NiallRees  before we merge this change, since he was the original champion of adding YAML checks for every SQL file.

tunetheweb: > Yes, I'm pretty sure (but not 100% sure) that running pytest without running the script catches me if I forget to run the script.

If it does that already, then I'm confused what the point of this PR is?

> IIUC, adding this step to the tox config would make it mandatory to also add the GitHub Action check as well, to ensure the files are up to date.

That is correct. The `tox` additions just make it less likely to be a problem for those that run that.

> If I'm correct that pytest already detects this, I don't see how this additional work will add any benefit.
> Would you be willing to test the pytest approach to try and confirm or deny that it detects problems?

It doesn't catch minor errors and differences (e.g. using `'a'` over `"a"`) or spacing differences. Otherwise the extra changes in https://github.com/sqlfluff/sqlfluff/pull/1241 wouldn't have been caught?

It does capture when you handcraft YAML files and get it wrong. So thats good. But I didn't know to run that command so thought it was a bug. Hence why I raised #1273 

So the additional GitHub Action would capture those and have the following benefits:
1) Make contributors more aware of the auto generation.
2) Check PRs are including the correct YAML (as would be auto generated).
3) Prevent these randoms changes when people don't run this script before submitting a PR.
And, as I say, there's no real downside if people are running it already.

barrywhart: I'm good with the tox and GitHub changes, too. Would you like to create a PR for my PR to add those changes?

tunetheweb: See https://github.com/sqlfluff/sqlfluff/pull/1280

tunetheweb: So, given that, do you still want to go the Hash route here? Not sure if that's overkill now or there's still other good changes and clean up in here?

We should at least do the parallelisation of the generation though as that's definitely a good change.

Oh and just realised you meant to add the PR to this PR - sorry I did a separate one!

barrywhart: Definitely the parallel part. The hash stuff and the related code cleanup seem "nice to have", definitely not "must have". Let's see what Niall or other maintainers think. (I can't merge my own changes without someone else's review, regardless.)

barrywhart: That's fine about your PR. I thought it may somehow depend on or touch the same code as this one, but they were completely independent. It worked out well -- I was happy to go ahead and get your changes in place! Good ideas there.

barrywhart:
@tunetheweb: All the checks except one are passing. Ironically, it's the "Check generated YAML files" step. Does the failure make sense to you? I wonder if either:
* My PR moves some code around so there's an import failure trying to find something that is gone or moved.
* The check is not setting up the Python environment using the same procedure as the other steps.

barrywhart:
BTW, I tried running tox to see I could reproduce this issue locally, but I'm on an old Mac and can't install xcode or `brew install python@3.8`. Maybe when I get the new machine working, I can do it. That could be a few days or longer, depending on what else is happening at work.

tunetheweb:
It's basically executing this:

```
pip install -Ur requirements.txt -Ur requirements_dev.txt
python setup.py develop
```

I get the exact same error when I run those commands locally.

But changing this line from test/conftest.py:

```py
from .dialects.parse_fixtures import load_file
```

to this:

```py
from dialects.parse_fixtures import load_file
```

Resolves the error.

test/conftest.py
```diff
@@ -14,6 +19,10 @@
     CodeSegment,
 )
 from sqlfluff.core.templaters import TemplatedFile
+from .dialects.parse_fixtures import load_file
```
tunetheweb, line 22: Not sure why this needs to be relative to current path (`.`), rather than relative to this file location, which seems to be causing you your issue:

```suggestion
from dialects.parse_fixtures import load_file
```

tunetheweb: Actually ignore this. This causes other errors.

tunetheweb:
Actually this causes other errors.

The only reliable way to get it to work seems to be to move `load_file` and `make_dialect_path` from `test/dialects/parse_fixtures.py` to `test/conftest.py` and then update the references in `test/dialects/dialects_test.py` to load them from the parent:

```py
from ..conftest import compute_parse_tree_hash, parse_example_file, load_file, make_dialect_path
from .parse_fixtures import get_parse_fixtures
```

barrywhart:
Thanks, @tunetheweb. Interestingly, making the changes you suggested broke it for me. I feel like we've wandered into some "Twilight Zone" of Python import behavior. I moved _all_ the code from `parse_fixtures.py` into `test/conftest.py` to see if that works. 🤞🏽 This is all test code, so it'll have no impact on users.

test/generate_parse_fixture_yml.py
```diff
@@ -1,34 +1,39 @@
...
     path = os.path.join("test", "fixtures", "parser", dialect, root + ".yml")
     with open(path, "w", newline="\n") as f:
-        if r:
+        r = None
+        if tree:
+            r = dict(
+                [("_hash", _hash)]
+                + list(tree.as_record(code_only=True, show_raw=True).items())
+            )
             yaml.dump(r, f, default_flow_style=False)
```
tunetheweb, line 25: Could we generate a comment line at the top of each YML file as well to really hammer home the point for those that won't run the tests and see the message until after?

For example something like this:
```
# YML test files are auto-generated from SQL files and should not be edited by hand as the hash needs to be calculated.
# Please run the `python test/generate_parse_fixture_yml.py` command to generate them after adding/altering SQL files.
```

tunetheweb (changes requested):

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ Note: Changes are now automatically tracked in [GitHub](https://github.com/sqlfl
 - Update L019 (leading/trailng comma rule) so it doesn't run on unparsable code.
 - The "--nocolor" command-line option should suppress emoji output (https://github.com/sqlfluff/sqlfluff/issues/1246)
 - Added HTTP Archive to the [In The Wild page](https://docs.sqlfluff.com/en/stable/inthewild.html)
+- Enforce that YML test files are computer generated and not edited (https://github.com/sqlfluff/sqlfluff/issues/1277)
```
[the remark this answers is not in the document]
barrywhart: Sure!

test/generate_parse_fixture_yml.py
barrywhart: Will do.

barrywhart:
@tunetheweb: This is ready for another review.

test/generate_parse_fixture_yml.py
tunetheweb: That's great. Definitely think that makes this clearer and further avoids people hand editing. Plus points them in the right direction to generate!

test/conftest.py
```diff
@@ -37,6 +94,31 @@ def process_struct(obj):
...
+def compute_parse_tree_hash(tree):
+    """Given a parse tree, compute a consistent hash value for it."""
+    if tree:
+        r = tree.as_record(code_only=True, show_raw=True)
+        if r:
+            r_io = io.StringIO()
+            oyaml.dump(r, r_io)
+            result = hashlib.blake2s(r_io.getvalue().encode("utf-8")).hexdigest()
+            # print(result)
+            # print(r_io.getvalue())
```
tunetheweb, line 117: Can we remove these debug lines?

tunetheweb (approved):
LGTM but maybe some with better python skills then me should review too?

One minor nit.

PR #1280, opened 2021-08-15
Add fixture YAML generation checks

### LINKED ISSUE #1277: Better ways to handle YML test files

tunetheweb:
Fixes #1277 

This PR adds the following:
- Adds the generate command to `tox.ini` config so YAML files will be generated when contributors run `tox` locally
- Adds a new GitHub Action to check that contributors have included any necessary YAML changes by running the script and checking for diffs. The tests already pick up diffs, but ignores whitespace changes for manually edited files - which can then pollute future PRs when the next person runs the generate scripts.
 
YAML files can still be generated via the python script, via the new tox config above, or via manual editing - as long as they match what would auto generate.

### DIFF
--- a/.github/workflows/check-yaml-files-generation.yml
+++ b/.github/workflows/check-yaml-files-generation.yml
@@ -11,8 +11,7 @@ jobs:
         python-version: '3.8'
     - name: Install dependencies
       run: |
-        pip install -Ur requirements.txt -Ur requirements_dev.txt
-        python setup.py develop
+        pip install -e -Ur requirements.txt -Ur requirements_dev.txt
     - name: Generate the YAML files
       run: |
         python test/generate_parse_fixture_yml.py
@@ -20,7 +19,7 @@ jobs:
       run: |
         if [ -n "$(git status --porcelain)" ]; then
           git diff
-          echo "Generated YAML files do no match branch."
+          echo "Generated YAML files do not match branch."
           echo "Please run the following command to generate these:"
           echo "  python test/generate_parse_fixture_yml.py"
           exit 1
--- a/.github/workflows/check-yaml-files-generation.yml
+++ b/.github/workflows/check-yaml-files-generation.yml
@@ -12,7 +12,7 @@ jobs:
     - name: Install dependencies
       run: |
         pip install -Ur requirements.txt -Ur requirements_dev.txt
-        pip install -e
+        python setup.py develop
     - name: Generate the YAML files
       run: |
         python test/generate_parse_fixture_yml.py
--- a/.github/workflows/check-yaml-files-generation.yml
+++ b/.github/workflows/check-yaml-files-generation.yml
@@ -1,0 +1,27 @@
+name: "Check generated YAML files"
+on: pull_request
+jobs:
+  changelog:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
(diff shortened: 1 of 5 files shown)


test/fixtures/parser/bigquery/select_mixture_of_array_literals.yml
```diff
@@ -34,7 +34,7 @@ file:
             array_literal:
               start_square_bracket: '['
               expression:
-                literal: "'a'"
+                literal: '"a"'
```
tunetheweb, line 37: This manual edit is only included to show the check failing.

Once happy with this PR then let me know and I will revert this change and the test will pass.

.github/workflows/check-yaml-files-generation.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
+    - name: Install dependencies
+      run: |
+        pip install -Ur requirements.txt -Ur requirements_dev.txt
+        python setup.py develop
```
barrywhart, line 15: Nitpick: Nowadays, I think it's more common to use `pip install -e .` rather than `python setup.py develop`. I'm not sure there's any real difference in behavior, though.

.github/workflows/check-yaml-files-generation.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Generate the YAML files
+      run: |
+        python test/generate_parse_fixture_yml.py
+    - name: Test the generated YAML files
+      run: |
+        if [ -n "$(git status --porcelain)" ]; then
+          git diff
+          echo "Generated YAML files do no match branch."
```
barrywhart, line 23:
```suggestion
          echo "Generated YAML files do not match branch."
```

barrywhart (commented):
Looks good! If you want to revert that test change, we can merge this.

tunetheweb:
> Looks good! If you want to revert that test change, we can merge this.

Actually the tests DID fail for this dummy change. So think it's only whitespace changes that can sneak through rather than actual errors.

.github/workflows/check-yaml-files-generation.yml
tunetheweb: Can this me merged with the line above? Or does it still need to be two commands?

tunetheweb: Couldn't get it to work so reverted back to `python setup.py develop`.

barrywhart: Interesting. No worries!

barrywhart (approved):

PR #1193, opened 2021-07-08
Add QUALIFY statement to teradata dialect

silverbullettruck2001:
Resolves #1184

PR #1281, opened 2021-08-15
Add QUALIFY clause to BigQuery

### LINKED ISSUE #1242: Add Qualify statement to Bigquery

## Expected Behaviour
Qualify should be parsed for BigQuery dialect

## Observed Behaviour
```
found unparsable section: '\n qualify row_num = 1'
```

## Version
```
sqlfluff, version 0.6.2
```
## Configuration
```
```

tunetheweb:
Fixes #1242 

Mostly based on #1193

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support ARRAYS of STRUCTs in BigQuery dialect [#1271](https://github.com/sqlfluff/sqlfluff/issues/1271)
 - Support fields of field references in BigQuery dialect [#1276](https://github.com/sqlfluff/sqlfluff/issues/1276)
 - Support OFFSET and ORDINAL clauses of Array Functions in BigQuery dialect [#1171](https://github.com/sqlfluff/sqlfluff/issues/1171)
+- Support QUALIFY to BigQuery dialect [#1242](https://github.com/sqlfluff/sqlfluff/issues/1242)
 
 
 ### Changed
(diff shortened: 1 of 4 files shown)


barrywhart (approved):
Looks good! Can you resolve the `CHANGELOG.md` conflict? Then we can merge.

tunetheweb:
> Looks good! Can you resolve the `CHANGELOG.md` conflict? Then we can merge.

Done. I wonder if there is a better way here? But can’t think of one - except maintainers resolving these simple CHANGES.md ones and merging of that’s the one issue (by default GitHub opens PRs with “allow maintainers to edit this branch” mode).

barrywhart:
I've wondered about that, too. Also wondered if maintainers can make changes even if the PR is based on someone's personal clone of the repo.

I'm curious what other projects do. One other _possible_ idea would be to generate the change log automatically, either from the PRs or closed issues, something like that. We're definitely figuring this out as we go. 😝

tunetheweb:
> I've wondered about that, too. Also wondered if maintainers can make changes even if the PR is based on someone's personal clone of the repo.

Yes they can as have done that on other projects. Only on that branch the PR is for and only if they've not unticked the "allow edits" box.

> I'm curious what other projects do. One other _possible_ idea would be to generate the change log automatically, either from the PRs or closed issues, something like that. We're definitely figuring this out as we go. 😝

I've used the [release-drafter GitHub Action](https://github.com/marketplace/actions/release-drafter) on other projects. It basically creates a draft Release in GitHub and updates it every time a PR is merged. This draft PR is only visible to maintainers.

Here's the one for my example:

image

You coudl then copy that into the CHANGES.md file at release time? That would avoid the continual merge conflicts. The downside though is non-maintainers can't see what's upcoming, which is useful feature of the current process.

Or we auto add to CHANGES.md on merge? That might be a better way...

Probably should open a separate issue to discuss.

barrywhart:
@tunetheweb: That sounds cool! Definitely worth creating another issue to discuss. I'd like to understand a bit more about how it works, but yes, the CHANGELOG has been useful but also the inherent merge conflicts add "drag" to every PR now.

PR #1294, opened 2021-08-17
Support empty OVER() clause in Window Specification

tunetheweb:
Fixes #1293

PR #1295, opened 2021-08-17
Add release drafter

### LINKED ISSUE #1287: Better ways to handle CHANGELOG.md
The [CHANGELOG.md](https://github.com/sqlfluff/sqlfluff/blob/main/CHANGELOG.md) file lists all the changes, and there is even a [GitHub Action](https://github.com/sqlfluff/sqlfluff/blob/main/.github/workflows/check-changelog-edited.yml) to ensure it's updated for each PR.

However this does create a merge conflict with nearly every PR (unless it's raised, and closed without any other changes happening in the meantime)

It also means (as noted in a [Slack thread](https://sqlfluff.slack.com/archives/C01JS7H598E/p1625104720071100)) you often have to raise a change with a blank entry (or let it fail), then add an entry - meaning extra commits.

As this project grows, this minor annoyance could grow more and more annoying, so is there a better way of doing this?

I've used the [release-drafter GitHub Action](https://github.com/marketplace/actions/release-drafter) on other projects. It basically creates a draft Release in GitHub and updates it every time a PR is merged. This draft PR is only visible to maintainers.

Here's an example for another repo I work on:

image

This also helps with release notes as can just convert that Draft to the next release. The Action will then automatically open a new draft on the next merge.

You could then copy that into the CHANGELOG.md file at release time? That would avoid the continual merge conflicts. The downside though is non-maintainers can't see what's upcoming, which is useful feature of the current process and maintaining these changes in CHANGELOG.md, which is visible to all.

The other alternative would be auto adding to CHANGELOG.md on merge to `main`? That might be a better way but would mean either updating `main` directly in a GitHub workflow (not really recommended!) or opening a Pull Request for each successfully merged PR - meaning we have two PRs for each one! So neither option is great to be honest.

Anyone any other thoughts on better ways to handle this?

tunetheweb:
Fixes #1287 

Also removes the GitHub Action checking whether CHANGELOG.md is updated as it will now be maintained in GitHub Release and then copied over as part of release process.

Can one of the maintainers create a `skip-changelog` label? You can then use that for issues you don't want to be added to the change log (e.g. release PRs). Or alternatively just allow them to be added and then edit them out. 😉


barrywhart:
I went ahead and created a `skip-changelog` label.

barrywhart (approved):
Looks good to me!

barrywhart:
@tunetheweb: Can you fix the `CHANGELOG.md` conflict? Hopefully the last one of these ever...

tunetheweb:
Can you confirm this PR was added to https://github.com/sqlfluff/sqlfluff/releases and add a screenshot so I can see? Non-maintainers can't see draft releases.

barrywhart:
I see it.

image

tunetheweb:
That's interesting. I did not expect to see #1285 and #1294 in that as they were merged before this was added. Guess it must be clever enough to look at all merges since the last release?

barrywhart:
I was surprised, too. Each release has an associated tag, so maybe it's comparing the latest `main` against that tag. That does make me wonder if it's possible to edit the draft release notes after each PR is merged, or only at release time. It would be technically tricky to allow the former, since it'd involve mixing human and computer-generated text (a bit like what SQLFluff is doing with respect to the Jinja templating behavior we were discussing in the other issue just now!)

PR #1303, opened 2021-08-20
Fix spacing issue for BigQuery UNNEST statement for rules L003 and L025

### LINKED ISSUE #1302: L025 rule firing in fix (but not lint) and breaking SQL for BigQuery UNNEST statement.
Consider the following SQL in BigQuery dialect:

```sql
SELECT
    category,
    value
FROM
    table1,
UNNEST(1, 2, 3) AS category
```

If I `lint` it, it correct picks up that the `UNNEST` statement has bad indentation (should be the same indentation as `table1`):

```
% sqlfluff lint --dialect bigquery test.sql
== [test.sql] FAIL
L:   6 | P:   1 | L003 | Indentation not consistent with line #5
All Finished 📜 🎉!
```

However if I run `fix` on this it identifies **two** errors:

```
% sqlfluff fix --dialect bigquery test.sql
== [test.sql] FAIL
L:   6 | P:   1 | L003 | Indentation not consistent with line #5
L:   6 | P:  20 | L025 | Alias 'category' is never used in SELECT statement.
==== fixing violations ====
2 fixable linting violations found
```

The alias `category` **IS** being used - so why does `fix` think it's not being used? And why is it different than what `lint` found? I think it's because  it sees `UNNEST` as a table and thinks the `category` is a table alias rather than a special column alias just for `UNNEST`?

And if I then `fix` this, it (correctly!) fixes the indentation but (incorrectly!) removes the `category` alias:

```sql
SELECT
    category,
    value
FROM
    table1,
    UNNEST(1, 2, 3)
```

This breaks the query as the `category` in the `SELECT` clause now refers to nothing!

So it's almost like `fix` for rule L025 is not aware of this special syntax for BigQuery but `lint` is?

What's even stranger (and might point us in the direction of the fix) is that if I lint it without passing a dialect (so using the standard ANSI dialect) then even `lint` picks up the "error":

```
% sqlfluff lint test.sql
== [test.sql] FAIL
L:   2 | P:   5 | L027 | Unqualified reference 'category' found in select with
                       | more than one referenced table/view.
L:   3 | P:   5 | L027 | Unqualified reference 'value' found in select with more
                       | than one referenced table/view.
L:   6 | P:   1 | L003 | Indentation not consistent with line #5
L:   6 | P:  20 | L025 | Alias 'category' is never used in SELECT statement.
All Finished 📜 🎉!
```

## Expected Behaviour

Fix should not attempt to fix non-linting errors and shouldn't break the SQL.

## Observed Behaviour

Fix broke the SQL.

`fix` identifying more issues than `lint` was also identified in #1149 but have opened this as a separate issue as it's a more specific example and that issue didn't say whether fix would break something or not.

## Steps to Reproduce

As per above

## Dialect

BigQuery

## Version

sqlfluff, version 0.6.3

## Configuration

No config (other than `--dialect bigquery` in command line), so default.

Will try to figure out what's going on here, and submit a fix if I do, but any pointers greatly appreciated as bit confused how `lint` and `fix` can be different...

tunetheweb:

Fixes #1302 so when SQL fails L003 (and adds indentation spacing) it doesn't break rule L025 (which has a special case for BigQuery that doesn't work when spacing is added).

Nope

### Pull Request checklist

  - Other.

I don't know how to add a test case for this to be honest as it only happens for a combination of rules. If someone can suggest how to add test cases for multiple rules then let me know as would be good to prevent this regressing in future.

I tried adding a rules fixture test case with the following:

```
  configs:
    core:
      dialect: bigquery
      rules: L003,L025
```

But still only rule L025 is run.

However it's a simple change, and coverage won't change since the line is already covered so if this is not possible then think it's still good to merge.

### DIFF
--- a/src/sqlfluff/core/rules/analysis/select.py
+++ b/src/sqlfluff/core/rules/analysis/select.py
@@ -132,6 +132,7 @@ def _has_value_table_function(table_expr, dialect):
         return False
 
     for function_name in table_expr.recursive_crawl("function_name"):
-        if function_name.raw.lower() in dialect.sets("value_table_functions"):
+        # Other rules can increase whitespace in the function name so use strip
+        if function_name.raw.lower().strip() in dialect.sets("value_table_functions"):
             return True
     return False
--- a/src/sqlfluff/core/rules/analysis/select.py
+++ b/src/sqlfluff/core/rules/analysis/select.py
@@ -132,7 +132,8 @@ def _has_value_table_function(table_expr, dialect):
         return False
 
     for function_name in table_expr.recursive_crawl("function_name"):
-        # Other rules can increase whitespace in the function name so use strip
+        # Other rules can increase whitespace in the function name, so use strip to remove
+        # See: https://github.com/sqlfluff/sqlfluff/issues/1304
         if function_name.raw.lower().strip() in dialect.sets("value_table_functions"):
             return True
     return False
--- a/test/fixtures/linter/autofix/bigquery/005_unnest_spacing/after.sql
+++ b/test/fixtures/linter/autofix/bigquery/005_unnest_spacing/after.sql
@@ -1,0 +1,6 @@
+SELECT
+    category,
+    value
+FROM
+    table1,
+    UNNEST(1, 2, 3) AS category
--- a/test/fixtures/linter/autofix/bigquery/005_unnest_spacing/before.sql
+++ b/test/fixtures/linter/autofix/bigquery/005_unnest_spacing/before.sql
@@ -1,0 +1,6 @@
+SELECT
+    category,
+    value
+FROM
+    table1,
+UNNEST(1, 2, 3) AS category
--- a/test/fixtures/linter/autofix/bigquery/005_unnest_spacing/test-config.yml
+++ b/test/fixtures/linter/autofix/bigquery/005_unnest_spacing/test-config.yml
@@ -1,0 +1,4 @@
+test-config:
+  rules:
+    - L003
+    - L025


barrywhart:
>how to add test cases for multiple rules

This is supported, although maybe not documented. See the subdirectories under `test/fixtures/linter/autofix/bigquery`. Each subdirectory is one test. `test-config.yml` lists the rules to run. (TBH, I don't know why these tests don't just use `.sqlfluff` configuration.)

src/sqlfluff/core/rules/analysis/select.py
```diff
@@ -132,6 +132,7 @@ def _has_value_table_function(table_expr, dialect):
         return False
 
     for function_name in table_expr.recursive_crawl("function_name"):
-        if function_name.raw.lower() in dialect.sets("value_table_functions"):
+        # Other rules can increase whitespace in the function name so use strip
+        if function_name.raw.lower().strip() in dialect.sets("value_table_functions"):
```
barrywhart, line 136: I saw your explanation of this fix on the related issue. I think that behavior (rule adding whitespace to function names) is a deeper bug that we should investigate and fix as well. The correct behavior would be to add the whitespace as a new segment (or possibly edit/extend existing whitespace to the left).

We seem to discover these kinds of rule bugs every month or two: a rule that is doing something wrong, but it doesn't cause an obvious problem most of the time.

I have this idea that SQLFluff should have various runtime checks on the `LintFix` objects returned by rules to prevent them from misbehaving. I should probably create a ticket for that. A "linter for the linter", kind of.

barrywhart: In fact, I think if we fix that deeper issue, this change will no longer be needed, if that makes sense. Even during the "fix" phase, segments should always correspond to something the parser and the dialect would've created, but in this case, that's not true (whitespace and code in the same segment).

tunetheweb:
> > how to add test cases for multiple rules
> 
> This is supported, although maybe not documented. See the subdirectories under `test/fixtures/linter/autofix/bigquery`. Each subdirectory is one test. `test-config.yml` lists the rules to run. (TBH, I don't know why these tests don't just use `.sqlfluff` configuration.)

Nice! Added test case and confirmed that worked (broke without the fix, passed with the fix).

src/sqlfluff/core/rules/analysis/select.py
tunetheweb: I don't disagree.

However I still think we should accept this fix in the meantime because:
1. This is a nasty little bug that breaks SQL. I think it's pretty common to have no indentation on `UNNEST` statements - it certainly is in our code base and don't think we're unusually in that regard - and running `sqlfluff fix` on any such code will break it.
2. The fix is tiny! And there's no real downsides to it being in there (the overhead of a `strip()` of a text which is just space and a function name is negligible!) so even though we could removed it if the larger fix is done, there's no downside to leaving it in place.
3. I have now added the test case, which will be useful for when we do the get round to the bigger fix.
4. The bigger fix might be quite large! Presumably the space is there to allow rule 10 to correct indentation. So that whole logic may have to be rewritten. This could also impact all the other rules that do indentation (and there's lots of them).
 
I can't commit to being able to do this rewrite any time soon and presume no one else can either. So my big preference is to merge this and open an issue for the bigger fix if you agree.

And btw this is such a nasty little bug (that broke lots of our queries!), that I was going to push for a 0.6.4 (even though we only released 0.6.3 fairly recently) once merged.

barrywhart: That sounds reasonable. Could you go ahead and open that other issue, then add a comment to the line above to explain why the `.strip()` is there? Once someone fixes the other issue, it will be cool to see if we can remove the above line without breaking the test.

barrywhart (commented):
Can you add the "related to issue XXXX" comment I suggested? Otherwise good to merge.

barrywhart:
I'm happy to work on a release once this is merged. (Will need another maintainer to approve the release PR, but I can do the legwork.)

src/sqlfluff/core/rules/analysis/select.py
tunetheweb: Done!

tunetheweb:
> I'm happy to work on a release once this is merged. (Will need another maintainer to approve the release PR, but I can do the legwork.)

That would be fantastic thanks! Don't forget to copy the changes to CHANGELOG.md this time.

barrywhart (approved):

PR #1307, opened 2021-08-23
Show latest version number on unsupported Python error message

zhongjiajie:

Just some simple change for error python version hint

### Pull Request checklist

Just some simple change for error python version hint


tunetheweb:
LGTM. Just wanna check it does what it looks to do when back at a computer.

tunetheweb (approved):

zhongjiajie:
Thanks @tunetheweb  for changing the PR title

PR #1334, opened 2021-08-28
Bugfix PR, fixes issue #1333, wherein test___main___help() defaults to your default Python installation

### LINKED ISSUE #1333: test___main___help() fails if sqlfluff isn't installed in your default python distribution
Subprocess creates a subshell which doesn't instantiate your virtual environment. If you've created a venv, and haven't installed sqlfluff in your default instance of python, it will fail the test. This is because the src directory is not naturally in sys.path. It is added to the sys.path either with a manually set PYTHONPATH, or by the distutils on setup (Pycharm adds it to the venv on creation due to the presence of setup.py). PYTHONPATH and sys.path are not set at any point, so cannot be expected to include src at runtime.

```
def test___main___help():
    """Test that the CLI can be access via __main__."""
    # nonzero exit is good enough
    subprocess.check_output(["python", "-m", "sqlfluff", "--help"])
```

## Expected Behaviour

The subprocess should run using your venv executable, and the test should pass

## Observed Behaviour

The test fails

## Steps to Reproduce

Uninstall sqlfluff on your default python distribution
Create a virtual environment
Activate virtual environment
Install tox
Run tox

## Version
Python 3.8.2

WittierDinosaur:
Fixes #1333. Changes the call to python in test___main___help() to sys.executable, which is the path of the current instance of Python that you are running. This ensures that sys.path is the same in the subshell. Also passes in os.environ into the subshell, in case the path is defined for any developers in the PYTHONPATH.
...

No

### Pull Request checklist

  - Other.

As it is not new functionality, just fixing a test case, there are no additional tests added. Tested the change locally, and the test is passing.

### DIFF
--- a/test/cli/commands_test.py
+++ b/test/cli/commands_test.py
@@ -628,7 +628,9 @@ def test__cli__command_lint_serialize_github_annotation():
 def test___main___help():
     """Test that the CLI can be access via __main__."""
     # nonzero exit is good enough
-    subprocess.check_output([sys.executable, "-m", "sqlfluff", "--help"], env=os.environ)
+    subprocess.check_output(
+        [sys.executable, "-m", "sqlfluff", "--help"], env=os.environ
+    )
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


barrywhart (approved):

barrywhart:
Thanks!

PR #1335, opened 2021-08-30
Add unittest for rules docstring

zhongjiajie:

Add unittest for sqlfluff rules docstring for more readable doc and avoid regression.
before this patch, our rules ref have wrong syntax of code
image

Some of the rule without example
image

And this path should fix this.

### DIFF
--- a/plugins/sqlfluff-plugin-example/src/example/rules.py
+++ b/plugins/sqlfluff-plugin-example/src/example/rules.py
@@ -47,7 +47,7 @@ class Rule_Example_L001(BaseRule):
     | **Anti-pattern**
     | Using ORDER BY one some forbidden columns.
 
-    .. code-block::
+    .. code-block:: sql
 
         SELECT *
         FROM foo
@@ -58,7 +58,7 @@ class Rule_Example_L001(BaseRule):
     | **Best practice**
     | Do not order by these columns.
 
-    .. code-block::
+    .. code-block:: sql
 
         SELECT *
         FROM foo
--- a/src/sqlfluff/rules/L009.py
+++ b/src/sqlfluff/rules/L009.py
@@ -8,7 +8,7 @@ from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 
 @document_fix_compatible
 class Rule_L009(BaseRule):
-    r"""Files must end with a trailing newline.
+    """Files must end with a trailing newline.
 
     | **Anti-pattern**
     | The content in file without ends without a trailing newline, the $ represents end of file.
--- a/src/sqlfluff/rules/L009.py
+++ b/src/sqlfluff/rules/L009.py
@@ -9,6 +9,7 @@ from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 @document_fix_compatible
 class Rule_L009(BaseRule):
     """Files must end with a trailing newline."""
+
     _check_docstring = False
 
     def _eval(self, segment, siblings_post, parent_stack, **kwargs):
--- a/src/sqlfluff/rules/L009.py
+++ b/src/sqlfluff/rules/L009.py
@@ -8,7 +8,27 @@ from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 
 @document_fix_compatible
 class Rule_L009(BaseRule):
-    """Files must end with a trailing newline."""
+    """Files must end with a trailing newline.
+
+    | **Anti-pattern**
+    | The content in file without end with a trailing newline.
+
+    .. code-block:: sql
+
+        SELECT
+            a
+        FROM foo
+
+    | **Best practice**
+    | Add trailing newline to the end, the \n character represents a newline symbol.
+
+    .. code-block:: sql
+
+        SELECT
+            a
+        FROM foo\n
+
+    """
 
     _check_docstring = False
 
--- a/src/sqlfluff/rules/L009.py
+++ b/src/sqlfluff/rules/L009.py
@@ -18,7 +18,7 @@ class Rule_L009(BaseRule):
         SELECT
             a
         FROM foo$
-        
+
         -- Ending on an indented line means there is no newline at the end of the file, the • represents space.
 
         SELECT
@@ -43,7 +43,7 @@ class Rule_L009(BaseRule):
             a
         FROM foo
         $
-        
+
         -- Ensuring the last line is not indented so is just a newline.
 
         SELECT
--- a/src/sqlfluff/rules/L009.py
+++ b/src/sqlfluff/rules/L009.py
@@ -46,7 +46,7 @@ class Rule_L009(BaseRule):
         
         -- Ensuring the last line is not indented so is just a newline.
 
-       SELECT
+        SELECT
         ••••a
         FROM
         ••••foo
--- a/src/sqlfluff/rules/L009.py
+++ b/src/sqlfluff/rules/L009.py
@@ -8,7 +8,7 @@ from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 
 @document_fix_compatible
 class Rule_L009(BaseRule):
-    """Files must end with a trailing newline.
+    r"""Files must end with a trailing newline.
 
     | **Anti-pattern**
     | The content in file without end with a trailing newline.
@@ -29,8 +29,6 @@ class Rule_L009(BaseRule):
         FROM foo\n
 
     """
-
-    _check_docstring = False
 
     def _eval(self, segment, siblings_post, parent_stack, **kwargs):
         """Files must end with a trailing newline.
--- a/src/sqlfluff/rules/L009.py
+++ b/src/sqlfluff/rules/L009.py
@@ -9,6 +9,7 @@ from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 @document_fix_compatible
 class Rule_L009(BaseRule):
     """Files must end with a trailing newline."""
+    _check_docstring = False
 
     def _eval(self, segment, siblings_post, parent_stack, **kwargs):
         """Files must end with a trailing newline.
--- a/src/sqlfluff/rules/L009.py
+++ b/src/sqlfluff/rules/L009.py
@@ -11,22 +11,54 @@ class Rule_L009(BaseRule):
     r"""Files must end with a trailing newline.
 
     | **Anti-pattern**
-    | The content in file without end with a trailing newline.
+    | The content in file without ends without a trailing newline, the $ represents end of file.
+
+    .. code-block:: sql
+
+        SELECT
+            a
+        FROM foo$
+        
+        -- Ending on an indented line means there is no newline at the end of the file, the • represents space.
+
+        SELECT
+        ••••a
+        FROM
+        ••••foo
+        ••••$
+
+        -- Ending on a semi-colon means the last line is not a newline.
+
+        SELECT
+            a
+        FROM foo
+        ;$
+
+    | **Best practice**
+    | Add trailing newline to the end, the $ character represents end of file.
 
     .. code-block:: sql
 
         SELECT
             a
         FROM foo
+        $
+        
+        -- Ensuring the last line is not indented so is just a newline.
 
-    | **Best practice**
-    | Add trailing newline to the end, the \n character represents a newline symbol.
+       SELECT
+        ••••a
+        FROM
+        ••••foo
+        $
 
-    .. code-block:: sql
+        -- Even when ending on a semi-colon, ensure there is a newline after
 
         SELECT
             a
-        FROM foo\n
+        FROM foo
+        ;
+        $
 
     """
 
--- a/src/sqlfluff/rules/L012.py
+++ b/src/sqlfluff/rules/L012.py
@@ -10,7 +10,7 @@ class Rule_L012(Rule_L011):
     separate so that they can be enabled and disabled separately.
 
     | **Anti-pattern**
-    | In this example, the alias 'alias_col' is implicit.
+    | In this example, the alias for column 'a' is implicit.
 
     .. code-block:: sql
 
--- a/src/sqlfluff/rules/L012.py
+++ b/src/sqlfluff/rules/L012.py
@@ -9,6 +9,24 @@ class Rule_L012(Rule_L011):
     NB: This rule inherits its functionality from obj:`Rule_L011` but is
     separate so that they can be enabled and disabled separately.
 
+    | **Anti-pattern**
+    | In this example, the alias 'alias_col' is implicit.
+
+    .. code-block:: sql
+
+        SELECT
+            a as alias_col
+        FROM foo
+
+    | **Best practice**
+    | Add `AS` to make it explicit.
+
+    .. code-block:: sql
+
+        SELECT
+            a AS alias_col
+        FROM foo
+
     """
 
     _target_elems = ("select_clause_element",)
--- a/src/sqlfluff/rules/L012.py
+++ b/src/sqlfluff/rules/L012.py
@@ -15,7 +15,7 @@ class Rule_L012(Rule_L011):
     .. code-block:: sql
 
         SELECT
-            a as alias_col
+            a
         FROM foo
 
     | **Best practice**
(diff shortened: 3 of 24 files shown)


src/sqlfluff/rules/L020.py
```diff
@@ -7,7 +7,29 @@
...
+    | Make all table with the some aliases.
+
+    .. code-block:: sql
+
+        SELECT
+            f.a,
+            f.a
+        FROM foo f, foo f
+
+    """
```
tunetheweb, line 32: I think you've misunderstood the intent of this rule.

Would suggest this instead:

```suggestion
    """Table aliases should be unique within each clause.

    | **Anti-pattern**
    | In this example, the alias 't' is reused for two different ables:

    .. code-block:: sql

        SELECT
            t.a,
            t.b
        FROM foo AS t, bar AS t
        
        -- this can also happen when using schemas where the implicit alias is the table name:

        SELECT
            a,
            b
        FROM
            2020.foo,
            2021.foo

    | **Best practice**
    | Make all tables have a unique alias

    .. code-block:: sql

        SELECT
            f.a,
            b.b
        FROM foo AS f, bar AS b
        
        -- Also use explicit alias's when referencing two tables with same name from two different schemas

        SELECT
            f1.a,
            f2.b
        FROM
            2020.foo AS f1,
            2021.foo AS f2

    """
```

src/sqlfluff/rules/L040.py
```diff
@@ -15,6 +15,40 @@ class Rule_L040(Rule_L010):
     """Inconsistent capitalisation of boolean/null literal.
 
     The functionality for this rule is inherited from :obj:`Rule_L010`.
+
+    | **Anti-pattern**
+    | In this example, 'null' and 'false' is in lower-case whereas 'TURE' is in upper-case.
```
tunetheweb, line 20:
```suggestion
    | In this example, 'null' and 'false' is in lower-case whereas 'TRUE' is in upper-case.
```

src/sqlfluff/rules/L040.py
```diff
@@ -15,6 +15,40 @@ class Rule_L040(Rule_L010):
...
+
+    .. code-block:: sql
+
+        select
+            a,
+            null,
+            TURE,
```
tunetheweb, line 27:
```suggestion
            TRUE,
```

src/sqlfluff/rules/L040.py
```diff
@@ -15,6 +15,40 @@ class Rule_L040(Rule_L010):
...
+        from foo
+
+        -- Also good
+
+        select
+            a,
+            NULL,
+            TRUE,
+            FALSE
+        from foo
```
tunetheweb, line 50: Change order to match text:

```suggestion
        select
            a,
            NULL,
            TRUE,
            FALSE
        from foo

        -- Also good

        select
            a,
            null,
            true,
            false
        from foo
```

test/core/rules/docstring_test.py
```diff
@@ -0,0 +1,37 @@
+"""Test rules docstring"""
+import pytest
+
+from sqlfluff.core.plugin.host import get_plugin_manager
+
+KEYWORD_ANTI = "\n    | **Anti-pattern**"
+KEYWORD_BEST = "\n    | **Best practice**"
+KEYWORD_CODE_BLOCK = "\n    .. code-block:: sql\n"
```
tunetheweb, line 8: Do we need the white space here? Seems fragile.

```suggestion
KEYWORD_ANTI = "**Anti-pattern**"
KEYWORD_BEST = "**Best practice**"
KEYWORD_CODE_BLOCK = ".. code-block:: sql\n"
```

test/core/rules/docstring_test.py
```diff
@@ -0,0 +1,37 @@
...
+        (KEYWORD_CODE_BLOCK, 2),
+    ]
+)
+def test_content_count(content, min_count):
+    """Test docstring have specific content"""
+    for plugin_rules in get_plugin_manager().hook.get_rules():
+        for rule in plugin_rules:
+            if rule._check_docstring is True:
+                assert rule.__doc__.count(content) >= min_count, \
+                    f"{rule.__name__} content {content} do not occurrences at less {min_count} times"
```
tunetheweb, line 25:
```suggestion
                    f"{rule.__name__} content {content} does not occur at less {min_count} times"
```

test/core/rules/docstring_test.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def test_keyword_anti_before_best():
+    """Test docstring anti pattern before best pattern"""
+    for plugin_rules in get_plugin_manager().hook.get_rules():
+        for rule in plugin_rules:
+            if rule._check_docstring is True:
+                assert rule.__doc__.index(KEYWORD_ANTI) < rule.__doc__.index(KEYWORD_BEST), \
+                    f"{rule.__name__} keyword {KEYWORD_BEST} appear before {KEYWORD_ANTI}"
```
tunetheweb, line 34:
```suggestion
                    f"{rule.__name__} keyword {KEYWORD_BEST} appears before {KEYWORD_ANTI}"
```

src/sqlfluff/rules/L009.py
```diff
@@ -9,6 +9,7 @@
 @document_fix_compatible
 class Rule_L009(BaseRule):
     """Files must end with a trailing newline."""
+    _check_docstring = False
```
tunetheweb, line 12: Could we not add an example for this? With a symbol for newline (like we use a symbol for "space" in L001)?

src/sqlfluff/rules/L012.py
```diff
@@ -9,6 +9,24 @@ class Rule_L012(Rule_L011):
...
     separate so that they can be enabled and disabled separately.
 
+    | **Anti-pattern**
+    | In this example, the alias 'alias_col' is implicit.
+
+    .. code-block:: sql
+
+        SELECT
+            a as alias_col
+        FROM foo
```
tunetheweb, line 19:
```suggestion
        SELECT
            a
        FROM foo
```

tunetheweb (changes requested):
Love the idea of this!

Spotted a few things as noted, and you will also need to apply the changes to:
```
plugins/sqlfluff-plugin-example/src/example/rules.py
```

And also fix the black errors:

```
./src/sqlfluff/rules/L009.py:12:1: BLK100 Black would make changes.
./test/core/rules/docstring_test.py:1:1: D415 First line should end with a period, question mark, or exclamation point
./test/core/rules/docstring_test.py:17:6: BLK100 Black would make changes.
./test/core/rules/docstring_test.py:20:1: D415 First line should end with a period, question mark, or exclamation point
./test/core/rules/docstring_test.py:29:1: D415 First line should end with a period, question mark, or exclamation point
./test/core/rules/docstring_test.py:37:1: W391 blank line at end of file
```

src/sqlfluff/rules/L040.py
zhongjiajie: Oh…… a bad copy and paste for typo

test/core/rules/docstring_test.py
zhongjiajie: @tunetheweb I add space to avoid the indentation issue for docstring, and I do not have strong opinion here

src/sqlfluff/rules/L020.py
zhongjiajie: Oh, I totally go to the way here, Thanks

src/sqlfluff/rules/L009.py
zhongjiajie: But we here talking about file, not sql syntax. So I am not sure use sql text would be a good idea.

zhongjiajie: And maybe we should use bash here?
```bash
> cat filename_anti.sql
select
        id
from foo
>
> cat filename_best.sql
select
        id
from foo

>
```

test/core/rules/docstring_test.py
zhongjiajie: I change my mind, I think white space have to add here, to ensure rule reference have same indent.
```diff
    | **Anti-pattern**
    | The • character represents a space.

    .. code-block:: sql

        SELECT
            a
        FROM foo••

    | **Best practice**
    | Remove trailing spaces.

-.. code-block:: sql
-
-    SELECT
-         a
-    FROM foo
+    .. code-block:: sql
+
+        SELECT
+            a
+        FROM foo
```
without white space, the red part would pass test but what we need is the green part

src/sqlfluff/rules/L009.py
zhongjiajie: I add describe and make sense for users

plugins/sqlfluff-plugin-example/src/example/rules.py
```diff
@@ -47,7 +47,7 @@ class Rule_Example_L001(BaseRule):
     | **Anti-pattern**
     | Using ORDER BY one some forbidden columns.
 
-    .. code-block::
+    .. code-block:: sql
```
zhongjiajie, line 50: I am not sure could we add unittest for plugin, So I just add syntax hint for it

zhongjiajie:
CI pass, PTAL @tunetheweb , and we still have discuss in [discuss-1](https://github.com/sqlfluff/sqlfluff/pull/1335#discussion_r698397699) and [discuss-2](https://github.com/sqlfluff/sqlfluff/pull/1335#discussion_r698965863)

plugins/sqlfluff-plugin-example/src/example/rules.py
tunetheweb: Yes that's all I meant. Resolving.

test/core/rules/docstring_test.py
tunetheweb: OK. Was just concerned if we ever changed the indentation in future. But guess it's a good check until then. Resolving.

src/sqlfluff/rules/L009.py
```diff
@@ -8,7 +8,27 @@
...
+        FROM foo
+
+    | **Best practice**
+    | Add trailing newline to the end, the \n character represents a newline symbol.
+
+    .. code-block:: sql
+
+        SELECT
+            a
+        FROM foo\n
```
tunetheweb, line 29: Perhaps include some of the more common mistakes I often see?

I also suggest changing from using \n (to represent newline) to $ (to represent end of file) as \n could be confusing as to which line it appears on.

```suggestion
    | **Anti-pattern**
    | The content in file without ends without a trailing newline, the $ represents end of file.

    .. code-block:: sql

        SELECT
            a
        FROM foo$
        
        -- Ending on an indented line means there is no newline at the end of the file, the • represents space.

        SELECT
        ••••a
        FROM
        ••••foo
        ••••$

        -- Ending on a semi-colon means the last line is not a newline.

        SELECT
            a
        FROM foo
        ;$

    | **Best practice**
    | Add trailing newline to the end, the $ character represents end of file.

    .. code-block:: sql

        SELECT
            a
        FROM foo
        $
        
        -- Ensuring the last line is not indented so is just a newline.

       SELECT
        ••••a
        FROM
        ••••foo
        $

        -- Even when ending on a semi-colon, ensure there is a newline after

        SELECT
            a
        FROM foo
        ;
        $
```

src/sqlfluff/rules/L012.py
```diff
@@ -9,6 +9,24 @@ class Rule_L012(Rule_L011):
     NB: This rule inherits its functionality from obj:`Rule_L011` but is
     separate so that they can be enabled and disabled separately.
 
+    | **Anti-pattern**
+    | In this example, the alias 'alias_col' is implicit.
```
tunetheweb, line 13: As 'alias_col' is not mentioned (do you mean the code is called that), let's just remove that from the anti-pattern example:

```suggestion
    | In this example, the alias for column 'a' is implicit.
```

src/sqlfluff/rules/L014.py
```diff
@@ -31,6 +31,35 @@ class Rule_L014(Rule_L010):
     """Inconsistent capitalisation of unquoted identifiers.
 
     The functionality for this rule is inherited from :obj:`Rule_L010`.
+
+    | **Anti-pattern**
+    | In this example, unquoted identifiers 'a' is in lower-case and
```
tunetheweb, line 36:
```suggestion
    | In this example, unquoted identifier 'a' is in lower-case but
```

src/sqlfluff/rules/L014.py
```diff
@@ -31,6 +31,35 @@ class Rule_L014(Rule_L010):
...
+
+    .. code-block:: sql
+
+        select
+            a,
+            B
+        from foo
+
+    | **Best practice**
+    | Make all unquoted identifiers either in upper-case or in lower-case
```
tunetheweb, line 47:
```suggestion
    | Ensure all unquoted identifiers are either in upper-case or in lower-case
```

src/sqlfluff/rules/L040.py
```diff
@@ -15,6 +15,40 @@ class Rule_L040(Rule_L010):
     """Inconsistent capitalisation of boolean/null literal.
 
     The functionality for this rule is inherited from :obj:`Rule_L010`.
+
+    | **Anti-pattern**
+    | In this example, 'null' and 'false' is in lower-case whereas 'TRUE' is in upper-case.
```
tunetheweb, line 20:
```suggestion
    | In this example, 'null' and 'false' are in lower-case whereas 'TRUE' is in upper-case.
```

src/sqlfluff/rules/L040.py
```diff
@@ -15,6 +15,40 @@ class Rule_L040(Rule_L010):
...
+
+        select
+            a,
+            null,
+            TRUE,
+            false
+        from foo
+
+    | **Best practice**
+    | Make all literal null/true/false either in upper-case or in lower-case
```
tunetheweb, line 32: It's not just lower and upper as there is "capitalise" (First letter only capitalised)

```suggestion
    | Ensure all literal null/true/false literals cases are used consistently
```

tunetheweb (commented):
Think we are almost there, but I have a few more comments.

src/sqlfluff/rules/L009.py
```diff
@@ -8,7 +8,27 @@
 
 @document_fix_compatible
 class Rule_L009(BaseRule):
-    """Files must end with a trailing newline."""
+    r"""Files must end with a trailing newline.
```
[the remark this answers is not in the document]
zhongjiajie: Cause docstring have `\n` which flake8-docstring got error, add `r` to pass flake8

src/sqlfluff/rules/L012.py
zhongjiajie: Good catch

src/sqlfluff/rules/L009.py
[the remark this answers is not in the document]
tunetheweb: Ah of course. Well if more the $ as suggested for other reasons, then don't need that anymore 😀

src/sqlfluff/rules/L009.py
zhongjiajie: What a good example for common mistake! 👍

src/sqlfluff/rules/L009.py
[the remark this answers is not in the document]
zhongjiajie: Yeah, you're right

src/sqlfluff/rules/L009.py
```diff
@@ -8,7 +8,59 @@
...
+    .. code-block:: sql
+
+        SELECT
+            a
+        FROM foo
+        $
+        
+        -- Ensuring the last line is not indented so is just a newline.
+
+       SELECT
```
zhongjiajie, line 49: Missing a white space
```suggestion
        SELECT
```

zhongjiajie: Fix it

tunetheweb (approved):
This is a great addition @zhongjiajie !

And thanks for working through all the feedback with me.

zhongjiajie:
@tunetheweb  Thanks for review and useful suggest.

PR #1336, opened 2021-08-30
Improve function support in Postgres dialect

### LINKED ISSUE #1288: Add support for function endings in dialect `postgres`
Defining functions in Postgres and ending the with some typical indications:

```
[...]
LANGUAGE plpgsql IMMUTABLE PARALLEL SAFE
```
or variations, such as:
```
[...]
LANGUAGE plpgsql STABLE PARALLEL SAFE
```

Executed with

```
sqlfluff lint --dialect postgres
```

causes following error:

```
L:  71 | P:  17 |  PRS | Line 71, Position 17: Found unparsable section: '
                       | IMMUTABLE PARALLEL SAFE'
```

It would be nice to get this added. Many thanks for the nice linter!

WittierDinosaur:
This is an enhancement of the FunctionDefinitionGrammar class for Postgres. It is comprehensive to allow all of the ending keywords, regardless of ordering.

Fixes #1288.

Includes all function endings as specified at https://www.postgresql.org/docs/9.1/sql-createfunction.html 
...

No

### Pull Request checklist
Tests have been enhanced in postgres_create_function.sql to cover the changes.


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -88,15 +87,62 @@
 
 @postgres_dialect.segment(replace=True)
 class FunctionDefinitionGrammar(BaseSegment):
-    """This is the body of a `CREATE FUNCTION AS` statement."""
+    """This is the body of a `CREATE FUNCTION AS` statement.
+
+    Options supported as defined in https://www.postgresql.org/docs/9.1/sql-createfunction.html
```
tunetheweb, line 92: Looks to me like you've included TRANSFORM which was added since 9.1 so should we update this to 9.5?

```suggestion
    Options supported as defined in https://www.postgresql.org/docs/9.5/sql-createfunction.html
```

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -88,15 +87,62 @@
...
+    """
 
     match_grammar = Sequence(
-        "AS",
-        OneOf(Ref("QuotedLiteralSegment"), Ref("DollarQuotedLiteralSegment")),
+        AnyNumberOf(
+            Sequence("TRANSFORM", "FOR", "TYPE", Ref("ParameterNameSegment")),
+            OneOf("WINDOW", "IMMUTABLE", "STABLE", "VOLATILE", "STRICT"),
+            Sequence(OneOf("NOT", optional=True), "LEAKPROOF"),
```
tunetheweb, line 99: I'm not 100% sure how to read tthe spec](https://www.postgresql.org/docs/9.5/sql-createfunction.html) but why is WINDOW on a separate line in that but not here? Even if it doesn't matter it might be good to align with the spec. Similar for `LEAKPROOF`, and leaving `STRICT` until the next line:

```suggestion
            Ref("WINDOW"),
            OneOf("IMMUTABLE", "STABLE", "VOLATILE", Sequence(Ref("NOT", optional=True), "LEAKPROOF"),),
```

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -88,15 +87,62 @@
...
+            Sequence(
+                OneOf("CALLED", Sequence("RETURNS", "NULL")), "ON", "NULL", "INPUT"
```
tunetheweb, line 101: I wonder if it would be less confusing just to have some repetition here?

```suggestion
                OneOf(
                    Sequence("CALLED", "ON", "NULL", "INPUT"),
                    Sequence("RETURNS", "NULL", "ON", "NULL", "INPUT"),
                    Sequence("STRICT"),
```

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -88,15 +87,62 @@
...
+            ),
+            Sequence(
+                OneOf("EXTERNAL", optional=True),
```
tunetheweb, line 104: Use `Ref` instead of `OneOf` if you need to use the optional

```suggestion
                Ref("EXTERNAL", optional=True),
```

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -88,15 +87,62 @@
...
+                "SECURITY",
+                OneOf("INVOKER", "DEFINER"),
```
tunetheweb, line 106: Again should we just use some repetition here to align more with the spec even if it is slightly more code? WDYT?

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -88,15 +87,62 @@
...
+            ),
+            Sequence("LANGUAGE", Ref("ParameterNameSegment")),
```
tunetheweb, line 108: Any reason not to have this at the top as per the spec?

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -88,15 +87,62 @@
...
+                "SET",
+                Ref("ParameterNameSegment"),
+                OneOf(
+                    Sequence(
+                        OneOf("TO", Ref("EqualsSegment")),
+                        Delimited(
+                            OneOf(
+                                Ref("ParameterNameSegment"),
+                                Ref("NumericLiteralSegment"),
+                                Ref("QuotedLiteralSegment"),
```
tunetheweb, line 120: Should this be just LiteralGrammar to include the other types (Bool, Date):

```suggestion
                                Ref("LiteralGrammar"),
```

test/fixtures/parser/postgres/postgres_create_function.sql
```diff
@@ -8,4 +8,15 @@ CREATE OR REPLACE FUNCTION increment(i integer) RETURNS integer AS '
...
+CREATE OR REPLACE FUNCTION increment(i integer) RETURNS integer AS '
+    BEGIN
+        RETURN i + 1;
+    END;
+' LANGUAGE plpgsql WINDOW IMMUTABLE STABLE LEAKPROOF RETURNS NULL ON NULL INPUT EXTERNAL SECURITY DEFINER
+ROWS 5 SET test_param = 3;
+
+CREATE OR REPLACE FUNCTION increment(i integer) RETURNS integer AS 'C:\\my_file.c', 'symlink_c'
+LANGUAGE plpgsql WINDOW IMMUTABLE STABLE NOT LEAKPROOF CALLED ON NULL INPUT EXTERNAL SECURITY DEFINER COST 123
+ROWS 5 SET test_param = 3 WITH (isStrict);
```
tunetheweb, line 22: It might be good to have some more examples of create function here. Two examples for all that syntax isn't a huge amount to be honest.

tunetheweb (changes requested):
This is great!

I do have some comments and suggestions though (mostly to do with making our code closer aligned to the Postgres definition spec).

Let me know what you think, of if you disagree and think it should stay as is.

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: Sorted, I've also brought it forward to 13, and added in support for PARALLEL and SUPPORT

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: Okay, so the reason it's on a separate line in the Postgres spec, is because it's a separate option. The spec reads that you can have IMMUTABLE or VOLATILE or STABLE, but the only other option for WINDOW is nothing. This comment probably works as an answer to a few of them, but the way i've grouped things in my implementation is based on syntax structure, not their logical position in the function definition. So all options that are just one keyword are together, keyword followed by param are together, keyword - numeric are together, etc. I think we need a new logical operator (though there is a chance this may already exist and i cant find it). The operator should be like AsNumberOf, but limit the number of times each of the *args can appear. Then we could use it to list the options by logical structure, and it would be meaningful. A test case being, "WINDOW WINDOW" would pass in this current implementation, but shouldn't, but if we could pop each of the *args when its found, we could engineer that to fail.

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: Fixed

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: Yes! I couldn't find a type like that. Fixed.

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: Fixed

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: I personally think it's neater this way, and not particularly confusing. Happy to defer to your opinion if you feel strongly about it.

src/sqlfluff/dialects/dialect_postgres.py
tunetheweb: Actually you CAN have what you want here! `AnyNumberOf` takes an optional `max_times` argument so just add `max_times=1` to line 140 and you can avoid "WINDOW WINDOW"!!

src/sqlfluff/dialects/dialect_postgres.py
tunetheweb: My concern is not necessarily understanding what you've written, but when the next version of Postgres comes out with one extra line of syntax - will it be obvious where and how to insert that new one? I think being closer to the spec will make that process easier as a future contributor can see exactly what's been implemented already and therefore what's missing. It took me a bit of matching your code to the spec to understand what you'd implemented was correct.

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: Edit, not fixed. If I use Ref, both for this and NOT, it says that 
`
RuntimeError: Grammar refers to 'EXTERNAL' which was not found in the postgres dialect
`
Possibly a bug in Ref?

tunetheweb: Actually no, it seems `Ref` doesn't do what I thought it did when passed a literal value! `Ref.keyword` is what I should have said:

```suggestion
                Ref.keyword("EXTERNAL", optional=True),
```

test/fixtures/parser/postgres/postgres_create_function.sql
WittierDinosaur: Added one more example. I can add more if needed, but i think all bases are covered by the tests now

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: I've just tried it out and I think max_times is 'the number of times a thing matches' rather than 'the number of times each thing matches'

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: Changed

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: Fixed

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: I think I agree with you on the CALLED ON NULL INPUT set, but constructing the SECURITY makes more sense to me. I've implemented the rest of your changes, can you see if it makes more sense, now that the thing as a whole marries up with the Postgres spec?

src/sqlfluff/dialects/dialect_postgres.py
tunetheweb: Yes think you are right. So it's set to `1` for `OneOf` but unset for `AnyNumberOf`. Shame. Looks like we need a `no_dups` optional parameter for this `AnyNumberOf`. Anyway, let's ignore for now...

test/fixtures/parser/postgres/postgres_create_function.sql
tunetheweb: What do you think about adding all the examples from the spec as well? So it's not jsut examples we've come up wiht (using whatever assumptions we've made)?

```sql

-- Exmaples from https://www.postgresql.org/docs/13/sql-createfunction.html

CREATE FUNCTION add(integer, integer) RETURNS integer
    AS 'select $1 + $2;'
    LANGUAGE SQL
    IMMUTABLE
    RETURNS NULL ON NULL INPUT;

CREATE OR REPLACE FUNCTION increment(i integer) RETURNS integer AS $$
        BEGIN
                RETURN i + 1;
        END;
$$ LANGUAGE plpgsql;


CREATE FUNCTION dup(in int, out f1 int, out f2 text)
    AS $$ SELECT $1, CAST($1 AS text) || ' is text' $$
    LANGUAGE SQL;

SELECT * FROM dup(42);


CREATE TYPE dup_result AS (f1 int, f2 text);

CREATE FUNCTION dup(int) RETURNS dup_result
    AS $$ SELECT $1, CAST($1 AS text) || ' is text' $$
    LANGUAGE SQL;

SELECT * FROM dup(42);

CREATE FUNCTION dup(int) RETURNS TABLE(f1 int, f2 text)
    AS $$ SELECT $1, CAST($1 AS text) || ' is text' $$
    LANGUAGE SQL;

SELECT * FROM dup(42);

CREATE FUNCTION check_password(uname TEXT, pass TEXT)
RETURNS BOOLEAN AS $$
DECLARE passed BOOLEAN;
BEGIN
        SELECT  (pwd = $2) INTO passed
        FROM    pwds
        WHERE   username = $1;

        RETURN passed;
END;
$$  LANGUAGE plpgsql
    SECURITY DEFINER
    -- Set a secure search_path: trusted schema(s), then 'pg_temp'.
    SET search_path = admin, pg_temp;


BEGIN;
CREATE FUNCTION check_password(uname TEXT, pass TEXT)
RETURNS BOOLEAN AS $$
DECLARE passed BOOLEAN;
BEGIN
        SELECT  (pwd = $2) INTO passed
        FROM    pwds
        WHERE   username = $1;

        RETURN passed;
END;
$$  LANGUAGE plpgsql
    SECURITY DEFINER;
REVOKE ALL ON FUNCTION check_password(uname TEXT, pass TEXT) FROM PUBLIC;
GRANT EXECUTE ON FUNCTION check_password(uname TEXT, pass TEXT) TO admins;
COMMIT;
```

src/sqlfluff/dialects/dialect_postgres.py
tunetheweb: Yup that's fine as more obvious. Looks great now. Just one more comment on the test cases then think we're good to merge!

test/fixtures/parser/postgres/postgres_create_function.sql
tunetheweb: These all pass with your code btw - I checked!

WittierDinosaur: Yeah, happy to add them. It exposed another problem though - the dollar_quote lexer is greedy. This regex: \$(\w*)\$[^\1]*\$\1\$ would match the entirety of this string "$$ function_stuff $$ stuff after function $$ another_function's stuff $$

tunetheweb: And that's why extra test cases are useful 😁

Agree on your regex change.

Though looks like it still doesn't like `in`/`out` params nor `RETURNS TABLE`  Both should be possible by overriding the appropriate pieces:

```py
     FunctionParameterGrammar=OneOf(
         Sequence(
             OneOf("IN", "OUT", optional=True),
             Ref("ParameterNameSegment", optional=True),
             OneOf(Sequence("ANY", "TYPE"), Ref("DatatypeSegment")),
         ),
```

Weirdly though I can't get it to recognise nameless In params (`in int`) like in the example, but named ones do work (`in a1 int`). Spot anything wrong with this?

```py
         Ref("FunctionParameterListGrammar"),
         Sequence(  # Optional function return type
             "RETURNS",
             OneOf(
                 Ref("DatatypeSegment"),
                 Sequence(
                     Ref.keyword("TABLE"),
                     Ref("FunctionParameterListGrammar"),
                 ),
             ),
             optional=True,
```

WittierDinosaur: Yeah, I've been having a fight with it for almost an hour. I've solved the vars problem, just can't quite get the table syntax right, The issue you're having is that DataTypeSegment isn't well-defined, its regex is basically anything, So it's finding the data type as the optional ParameterName, then looking after it for the required DataType. This code works, wrapping it in OneOf instead of using optional:

`
    FunctionParameterGrammar=Sequence(
        OneOf("IN", "OUT", "INOUT", "VARIADIC", optional=True),
        OneOf(Ref("DatatypeSegment"),
              Sequence(Ref("ParameterNameSegment"),
                       Ref("DatatypeSegment"))),
        Sequence(
            OneOf(
                "DEFAULT",
                Ref("EqualsSegment")
            ),
            Ref("LiteralGrammer"),
            optional=True
        )
    )
`

WittierDinosaur: Right, changes made, all tests passing locally... i think we might finally be there 😄

test/fixtures/parser/postgres/postgres_create_function.sql
```diff
@@ -8,4 +8,87 @@ CREATE OR REPLACE FUNCTION increment(i integer) RETURNS integer AS '
...
+        FROM    pwds
+        WHERE   username = $1;
+
+        RETURN passed;
+END;
+$$  LANGUAGE plpgsql
+    SECURITY DEFINER;
+REVOKE ALL ON FUNCTION check_password(uname TEXT, pass TEXT) FROM PUBLIC;
+GRANT EXECUTE ON FUNCTION check_password(uname TEXT, pass TEXT) TO admins;
+COMMIT;
```
tunetheweb, line 94: Nit:

```suggestion
COMMIT;

```

tunetheweb (approved):
Cheers for all your work on this one @WittierDinosaur ! And for dealing with the seemingly endless feedback. Huge amount of functionality added here!!

ErunamoJAZZ:
Excellent!, thank you very much!

Issue #1314, opened 2021-08-24
Increase pytest coverage to 100%


tunetheweb:
It's annoying when the coverage checker fails when you haven't made any code changes (see #1311 as an example), or even when you have they are code changes that shouldn't result in coverage changes.

I believe this is often this is due to rounding errors, or black reformatting increasing the number of lines, while not increasing the number of tests - leading to a percentage difference.

We're actually pretty close to 100% test coverage (see report below below, including the lines missing test coverage), so think with a little effort we could close this gap and avoid this issue AND prevent any regressions in code coverage.

Not every line of code can be tested and to cover this, I think we should make liberal use of `# pragma: no cover` to exclude untestable lines. I think that will close a lot of the gap, though for others there undoubtedly are missing tests, so we should only add this as a last resort.

Thoughts? Is this an unrealistic goal (I don't think so btw or wouldn't be proposing! 😀)

```
coverage report --show-missing
Name                                                 Stmts   Miss  Cover   Missing
----------------------------------------------------------------------------------
src/sqlfluff/__init__.py                                 9      0   100%
src/sqlfluff/api/__init__.py                             2      0   100%
src/sqlfluff/api/info.py                                 9      0   100%
src/sqlfluff/api/simple.py                              34      2    94%   22-23
src/sqlfluff/cli/__init__.py                             0      0   100%
src/sqlfluff/cli/commands.py                           299     35    88%   210-212, 400-405, 423-425, 515, 532, 538-539, 543, 551-556, 651-653, 681, 684, 686, 691, 722-729, 741, 750
src/sqlfluff/cli/formatters.py                         145      9    94%   58, 62-64, 67, 183, 287-288, 304, 308
src/sqlfluff/cli/helpers.py                             76      1    99%   92
src/sqlfluff/core/__init__.py                            8      0   100%
src/sqlfluff/core/config.py                            278      8    97%   72, 209, 244, 380, 469, 547, 596-599
src/sqlfluff/core/dialects/__init__.py                  17      0   100%
src/sqlfluff/core/dialects/base.py                     116     18    84%   48, 67, 114, 145, 148, 170, 180, 190, 192, 204, 206, 219, 226-232, 253, 262, 284, 296
src/sqlfluff/core/dialects/common.py                     9      0   100%
src/sqlfluff/core/errors.py                             72      3    96%   66, 70, 88
src/sqlfluff/core/file_helpers.py                        9      0   100%
[...]

PR #1311, opened 2021-08-23
Update documentation to make the project more discoverable

tunetheweb:
### Brief summary of the change made

Various updates to the README and other documentation pages to reflect the latest state of the projects and make the project more discoverable.

Here's the changes and reasoning:

#### CHANGELOG.md
- Remove the empty future section as just confusing and no longer required since we get these from GitHub draft release since #1295 
- Made the comment an HTML comment so it doesn't show for anyone viewing the changes outside of source.

#### CONTRIBUTING.md
- Updated example to 0.6.0 as 0.4.0 was a bit dated now

#### README.md
- Correct heading level and capitalisation for our headline ("The SQL Linter for Humans")
- Update GitHub badges broken since we moved from `master` to `main`.
- Highlighted SQLFLuff in bold consistently.
- Added section on Dialects supported and Templates supported. This will require some manual maintenance admittedly but think it's better for searchability and to let potential users easily see what we support (and how many different dialects we support so adding new ones is entirely possible!)
- Add hyperlinks to CLI and Rules for ease of use.
- Added note that the docs website is maintained from this repo so changes can be suggested here.
- Hide long Slack URL and made it a link instead.
- Added Twitter link.
- Reformatted Contributing section line endings and added link to GitHub contributors.

#### docs.sqlfluff.com
- Consistent capitalisation for our tagline ("The SQL Linter for Humans")
- Added Twitter link

#### setup.py (for PyPi)
- Consistent description ("The SQL Linter for Humans")
- Removed embedded CHANGELOG and added link instead
- Added more keywords for dialects we support + dbt. I didn't bother adding Jinja as that's quite generic.
- Added more links (based on [Flask](https://pypi.org/project/Flask/) which has these link types).

#### GitHub Issue Templates
- Adds a Documentation issue template for future suggested changes.

### Are there any other side effects of this change that we should be aware of?

Shouldn't be!

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/parser` (note YML files can be auto generated with `python test/generate_parse_fixture_yml.py` or by running `tox` locally).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #1337, opened 2021-08-31
Using github star instead of watch in docs

zhongjiajie:

We should using github star instead of github watch in our docs,
and I think we should remove repo code coverage in our docs too.

**Before**
image

**After**
image

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -86,7 +86,7 @@ html_theme_options = {
     "github_repo": "sqlfluff",
     # Github Fork button
     "github_banner": True,
-    # Github link button
+    # Github star button
     "github_type": "star",
     "github_count": "true",
 }
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -87,9 +87,8 @@ html_theme_options = {
     # Github Fork button
     "github_banner": True,
     # Github link button
-    "github_button": True,
-    # Codecov button
-    "codecov_button": True,
+    "github_type": "star",
+    "github_count": "true",
 }
 
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -88,6 +88,7 @@ html_theme_options = {
     "github_banner": True,
     # Github star button
     "github_type": "star",
+    # Use `"true"` instead of `True` for counting github star, see https://ghbtns.com for details
     "github_count": "true",
     # Codecov button
     "codecov_button": True,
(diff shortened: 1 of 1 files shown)


zhongjiajie:
Change parameters according to [alabaster#service-links](https://alabaster.readthedocs.io/en/latest/customization.html#service-links-and-badges)

docs/source/conf.py
```diff
@@ -86,10 +86,9 @@
     "github_repo": "sqlfluff",
     # Github Fork button
     "github_banner": True,
-    # Github link button
-    "github_button": True,
-    # Codecov button
-    "codecov_button": True,
+    # Github star button
+    "github_type": "star",
+    "github_count": "true",
```
tunetheweb, line 91: I like the move to Star instead of Watch.

Bit confused however as to why `github_button` didn't need a count option, but `github_action` does? Does Watch automatically add a count but star does not? If so that makes sense.

Should we use `True` though rather than `true` - abide by our own rules! 😁

I'm not sure about removing the code coverage button. Yes it's not really needed, and yes it's a bit geeky, but this is a project for geeky people and we're very proud of our high code coverage - though I personally want it even higher! See #1314 

```suggestion
    # Github star button
    "github_type": "star",
    "github_count": "True",
    # Codecov button
    "codecov_button": True,
```

zhongjiajie: Cause `github_button` is convenient function of `github_type="watch" and  github_count="true"`

And it's odd that `github_count` only work when we set to `"true"` instead of `True`, maybe it just pass value to https://ghbtns.com and it need `"true"`

Ok, I reset the code coverage button, for geeky 👍

tunetheweb: OK let's add a comment explaining it explicitly needs to be `"true"` in case anyone thinks this is a mistake.

tunetheweb: Updated suggested code change.

zhongjiajie: Ok, I'd like to add it

tunetheweb (approved):
LGTM!

Ignoring code  coverage error as false fail.

zhongjiajie:
[Question] why the code coverage will decrease?

zhongjiajie:
Thanks for merging

tunetheweb:
> [Question] why the code coverage will decrease?

It won't. See #1314  which starts with:

> It's annoying when the coverage checker fails when you haven't made any code changes (see #1311 as an example), or even when you have they are code changes that shouldn't result in coverage changes.
> 
> I believe this is often this is due to rounding errors, or black reformatting increasing the number of lines, while not increasing the number of tests - leading to a percentage difference.

Issue #350, opened 2020-05-16
[postgres] Support more GRANT privileges


ancoron:
Please support at least the following additional `GRANT` privileges for the PostgreSQL dialect:
* `EXECUTE` (e.g. `ON FUNCTION ...`)
* `USAGE` (e.g. `ON SCHEMA ...`)
* `<role_name> TO <role_specification>`

For more details: https://www.postgresql.org/docs/current/sql-grant.html

Simple examples:
```sql
GRANT USAGE ON SCHEMA api TO web_anon;
GRANT EXECUTE ON FUNCTION api.test TO web_anon;
GRANT web_anon TO my_user;
```

...results into:
```
== [db/test.sql] FAIL
L:   1 | P:   1 |  PRS | Found unparsable section: 'GRANT USAGE ON SCHEMA api TO
                       | web_anon'
L:   2 | P:   1 |  PRS | Found unparsable section: 'GRANT EXECUTE ON FUNCTION
                       | api.test TO we...'
L:   3 | P:   1 |  PRS | Found unparsable section: 'GRANT web_anon TO my_user'
```

Examples actually derived from [PostgREST docs](http://postgrest.org/en/v7.0.0/tutorials/tut0.html#step-4-create-database-for-api).

alanmcruickshank:
Agreed. Thanks for the examples too.

The additional base permissions should be very straightforward for this one. We'll need to add some additional segments to handle the roles and role specification but should be fairly straightforward.

creste:
@alanmcruickshank  - I see #1032 was merged and released but this issue hasn't been closed.  Are `GRANT` privileges supposed to work correctly now?  If not, then you can probably ignore the rest of this comment.

If they _are_ supposed to work correctly, then sqlfluff 0.6.4 is still failing to parse some `GRANT`s correctly.

Given this SQL file:
```sql
GRANT CONNECT ON DATABASE mydb TO app;
```
I see this output:
```
> sqlfluff lint --verbose .\test_grant.sql
==== sqlfluff ====
sqlfluff:                0.6.4 python:                  3.9.6
implementation:        cpython dialect:              postgres
verbosity:                   1 templater:               jinja

==== readout ====

=== [ path: .\test_grant.sql ] ===

== [test_grant.sql] FAIL
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'GRANT
                       | CONNECT ON DATABASE mydb TO app'
==== summary ====
[...]

PR #1032, opened 2021-05-09
Parse alter default privileges and more Postgres access privileges

NiallRees:
Resolves #348 and #350

PR #1339, opened 2021-08-31
Support additional GRANTs in Postgres

creste:
This PR adds support for all variants of the GRANT command documented at https://www.postgresql.org/docs/13/sql-grant.html.  This is based on prior work done by @NiallRees in https://github.com/sqlfluff/sqlfluff/pull/1032.

Fixes https://github.com/sqlfluff/sqlfluff/issues/350.

Some of the postgres-specific syntaxes, like `WITH ADMIN OPTION` and `GRANTED BY`, are now also considered valid in other dialects.  I didn't see a way to make them postgres-specific because `AccessStatementSegment` mixes all dialects in a single segment.

I'm not sure I understand when to put unreserved keywords in `ansi_keywords.py` vs `dialect_postgres.py`.  My strategy was to first try putting the missing keywords in `dialect_postgres.py` and if that didn't work then I promoted them to `ansi_keywords.py`.

### Pull Request checklist

I have included at least one test case per variant of the GRANT command, but I didn't do an exhaustive check of all GRANT command permutations.

All tests pass locally on python 3.9.


barrywhart (approved):

barrywhart:
Looks great, thanks!

Issue #1345, opened 2021-09-02
Snowflake Is Inheriting From Postgres


WittierDinosaur:
Snowflake dialect inherits from Postgres. It shouldn't, it isn't based on Postgres. It adds a dependency when developing in Postgres, to ensure that a) your code is Snowflake compliant, b) you add a hack into your code to keep it Snowflake compliant, or c) you fix Snowflake's issues. 

An example:

ANSI Dialect defines DESCRIBE parse grammar as 

```
    parse_grammar = Sequence(
        "DESCRIBE",
        Ref("NakedIdentifierSegment"),
        Ref("ObjectReferenceSegment"),
    )
```

where NakedIdentifierSegment is:

```
        lambda dialect: RegexParser(
            r"[A-Z0-9_]*[A-Z][A-Z0-9_]*",
            CodeSegment,
            name="naked_identifier",
            type="identifier",
            anti_template=r"^(" + r"|".join(dialect.sets("reserved_keywords")) + r")$",
        )
```

In Postgres, USER is a reserved keyword.

In Snowflake, this is a valid command:

```
DESCRIBE USER my_user;
```

PR #1347, opened 2021-09-02
Postgres keyword support

### LINKED ISSUE #1156: PostgreSQL: timestamptz not parsed

## Expected Behaviour

I would expect statements with keyword 'timestamptz' to be parsed by sqlfluff, as they are supported by PostgreSQL.

## Observed Behaviour

> ]$ sqlfluff lint test.sql --dialect postgres
== [test.sql] FAIL
L:   1 | P:  20 | L012 | Implicit aliasing of column not allowed. Use explicit
                       | `AS` clause.
L:   1 | P:  41 |  PRS | Line 1, Position 41: Found unparsable section: " -
                       | timestamptz '2013-03-01 12:00:00'"

> $ sqlfluff parse test.sql --dialect postgres
[L:  1, P:  1]      |file:
[L:  1, P:  8]      |                    column_reference:
[L:  1, P:  8]      |                        identifier:                           'timestamptz'
[L:  1, P: 19]      |                    whitespace:                               ' '
[L:  1, P: 20]      |                    alias_expression:
[L:  1, P: 20]      |                        literal:                              "'2013-07-01 12:00:00'"
[L:  1, P: 41]      |                    unparsable:                               !! Expected: 'Nothing...'
[L:  1, P: 41]      |                        whitespace:                           ' '
[L:  1, P: 42]      |                        raw:                                  '-'
[L:  1, P: 43]      |                        whitespace:                           ' '
[L:  1, P: 44]      |                        raw:                                  'timestamptz'
[L:  1, P: 55]      |                        whitespace:                           ' '
[L:  1, P: 56]      |                        raw:                                  "'2013-03-01 12:00:00'"
[L:  1, P: 77]      |            [META] dedent:
[L:  1, P: 77]      |    statement_terminator:                                     ';'
[L:  1, P: 78]      |    newline:                                                  '\n'
==== parsing violations ====
L:   1 | P:  41 |  PRS | Line 1, Position 41: Found unparsable section: " - timestamptz
                       | '2013-03-01 12:00:00'"

## Steps to Reproduce

> SELECT timestamptz '2013-07-01 12:00:00' - timestamptz '2013-03-01 12:00:00';

## Version

sqlfluff version 0.6.0a2
Python 3.8.1

This example comes from the official PostgreSQL [documentation](https://www.postgresql.org/docs/10/functions-datetime.html).

WittierDinosaur:
Keyword dump into Postgres. Uses the keywords as listed in the docs, and some I've found to not be there. Also created some functions to get keywords, and merge keyword list (so that I could have a docs matching list, and a custom list). 

Fixes #1156 

It created a bug in Snowflake, as Postgres is a dependency (I've submitted a Bug report on this), but I fixed the specific error that was thrown.

### Pull Request checklist
Added test cases to /test/dialects/postgres_test.py and some .yml test cases to Postgres


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -92,6 +93,7 @@
     ]
 )
 
+snowflake_dialect.sets("reserved_keywords").discard("USER")
```
tunetheweb, line 96: I'm a bit worried about the potential for breakage here given what you raised in #1345 

Until we can pull out Postgres from Snowflake (assuming we even want to do this - not familair enough with them to know), should we instead reset Snowflake to what it was before by deleting all the postgres keywords it will inherit (including the extra ones you've now added in this PR), and have it reimport them from ANSI - plus adding any of the keywords it used to get on top of that from the old postgres code?

A bit of work I admit, and not really fair to lump you, a postgres developer with presumably no interest in Snowflake, with this, but do worry about this being a breaking change if we don't do that.

Alternatively we could just set up the Snowflake keyword list as part of this PR if we can get a definitive list for them?

src/sqlfluff/dialects/postgres_keywords.py
```diff
@@ -0,0 +1,899 @@
...
+    keywords = [x[0] for x in keyword_list if x[1].startswith(keyword_type)]
+
+    return keywords
+
+
+postgres_docs_keywords = [
+    ("A", "not-keyword"),
+    ("ABORT", "non-reserved"),
+    ("ABS", "not-keyword"),
+    ("ABSENT", "not-keyword"),
```
tunetheweb, line 53: Any reason you didn't got the `ansi_keywords.py` and `exasol_keywords.py` route?

WittierDinosaur: Yeah, a few. Firstly, there are types in the list that are like "reserved-(can-be-function-or-type)", this allows us to implement this extra functionality later. Second, I wanted to track documented keywords, and non-documented separately, and when I do the PostGIS code, I want those keywords separately tracked as well. Thirdly, this includes words to deallocate as "not-keyword". And finally, the docs keywords conforms to the table in https://www.postgresql.org/docs/13/sql-keywords-appendix.html, it's harder to track and more work to pull them out into separate reserved and non-reserved lists.

src/sqlfluff/dialects/dialect_snowflake.py
WittierDinosaur: I see what you mean, but think it may be out of scope of this PR. From my perspective, Snowflake is a certain amount of broken, and that doesn't change with this PR. Every time we have done Postgres PRs, this issue has been present. and Snowflake's dialect definition has mutated. Snowflake currently passes all of the defined tests, so there's nothing other than USER that is an obvious symptom.

src/sqlfluff/dialects/postgres_keywords.py
WittierDinosaur: * i didn’t mean this to sound sassy or dismissive... but reason it back, it might do 😬

src/sqlfluff/dialects/dialect_snowflake.py
tunetheweb: I've had a look at Snowflake before you and another contributor started all the great Postgres work recently and it was fairly basic. As such I think it might be best to pull out Snowflake from Postgres anyway before next release in case your other changes break that anyway. Working on a PR now, which we might merge first before this, to resolve this issue.

src/sqlfluff/dialects/postgres_keywords.py
tunetheweb: No, no, it's all good. Good to hear the reasons. I'm still not 100% convinced though. But just noticed that exasol did a different implementation so maybe it doesn't need to be standard?

On the last point I do like matching the spec as much as possible (as you've noticed when I've ask you to align grammer more closely to the spec) but not as convinced on the keywords. They could change that table at any time.

Anyway not a complete deal breaker so will give it some more thought and come back to it once we resolve my first (bigger) concern above.

But good to know there was a good reason for it (several in fact!).

WittierDinosaur: They could, and I might implement an explicit test set on these keywords, to make sure we don’t lose any. I look forward to your snowflake pull... not against working on snowflake myself, I’m only working on Postgres because, I’ve worked in it enough to know the syntax well, and I noticed that you guys were looking for people to work on it

src/sqlfluff/dialects/postgres_keywords.py
```diff
@@ -0,0 +1,899 @@
...
+    ("COLLATION_CATALOG", "not-keyword"),
+    ("COLLATION_NAME", "not-keyword"),
+    ("COLLATION_SCHEMA", "not-keyword"),
+    ("COLLECT", "not-keyword"),
+    ("COLUMN", "reserved"),
+    ("COLUMNS", "non-reserved"),
+    ("COLUMN_NAME", "not-keyword"),
+    ("COMMAND_FUNCTION", "not-keyword"),
+    ("COMMAND_FUNCTION_CODE", "not-keyword"),
+    ("COMMENT", "non-reserved"),
```
miketheman, line 156: I was testing this branch today, since it has a LOT of the keywords that are currently unparseable in my postgresql files - thanks! I found one example that triggered a parsing failure, and wondered how/best to add to this changeset.

This SQL

```sql
COMMENT ON FUNCTION dummy() IS NULL;
```

raises:

```
sqlfluff.api.simple.APIParsingError: Found 1 issues while parsing string.
Line 2, Position 1: Found unparsable section: 'COMMENT ON FUNCTION dummy() IS NULL'
```

src/sqlfluff/dialects/dialect_snowflake.py
tunetheweb: Raised this: https://github.com/sqlfluff/sqlfluff/pull/1353

Got a little obsessed with closing out the issues while I was in there!

tunetheweb (approved):
OK I'm happy to merge this now we've a way forward for Snowflake with #1353 

But wouldn't mind the other @sqlfluff/sqlfluff-maintainers 's thoughts on the keywords approach here.

src/sqlfluff/dialects/postgres_keywords.py
```diff
@@ -0,0 +1,899 @@
+"""Keywords in the Postgres Dialect.
+
+Most of the keywords come from https://www.postgresql.org/docs/13/sql-keywords-appendix.html
+Here, "not-keyword" refers to a word not being a keyword, and will be removed from any default keyword definition,
+these keywords are, or have been, an ANSI keyword.
+
+There are also some keywords that are(n't) supported as types and function, but htere isn't support for that
```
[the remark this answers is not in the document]
WittierDinosaur: Fixed

src/sqlfluff/dialects/postgres_keywords.py
WittierDinosaur: That won't be a keyword problem, it's probably undeveloped grammar, can you add this as an issue?

miketheman: Done!

WittierDinosaur:
Is this good to go?

NiallRees:
> Is this good to go?

Looks like there are a couple of merge conflicts to address - then sure!

WittierDinosaur:
Should be sorted now

PR #1348, opened 2021-09-02
Show all LintResult in Rule_L020

zhongjiajie:

Rule_L020 only show first LintResult before,
after this patch merge Rule_L020 would show
all duplicate table aliases, without duplicate
position

I think this PR no need add `std_rule_cases` here, am I wrong?

### Pull Request checklist

  - ~`.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.`~  Already exists and no need

### DIFF
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -132,7 +132,9 @@ class BaseSegment:
         )
 
     def __hash__(self):
-        return hash((self.__class__.__name__, self.raw, self.pos_marker.source_position()))
+        return hash(
+            (self.__class__.__name__, self.raw, self.pos_marker.source_position())
+        )
 
     def __repr__(self):
         return f"<{self.__class__.__name__}: ({self.pos_marker})>"
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -131,6 +131,9 @@ class BaseSegment:
             )
         )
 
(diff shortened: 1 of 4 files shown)


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -131,6 +131,9 @@ def __eq__(self, other):
             )
         )
 
+    def __hash__(self):
+        return hash((self.__class__.__name__, self.raw, self.pos_marker.source_position()))
+
```
zhongjiajie, line 136: Add `__hash__` for `set.add` function which need hashable for remove duplicate element.

tunetheweb:
> I think this PR no need add `std_rule_cases` here, am I wrong?

I think a test with two missing aliases are both should would be good to ensure this doesn’t regress in future.

tunetheweb:
For more complex tests where you want to look at the actual output, you don't add a fixture, but instead add python code test/rules/. Have a look at this one for example:

https://github.com/sqlfluff/sqlfluff/blob/e21d200c35b75d771f2d7635460f9c0e7d9d6f06/test/rules/std_L007_test.py#L46-L63

zhongjiajie:
Thanks for review, I add more test for rule L020

zhongjiajie:
CI pass, PTAL thanks @tunetheweb

tunetheweb (approved):
LGTM! Nice test cases.

zhongjiajie:
Thanks for merging.

Issue #348, opened 2020-05-16
[postgres] Support ALTER DEFAULT PRIVILEGES


ancoron:
Please support `ALTER DEFAULT PRIVILEGES` in the PostgreSQL dialect:
* https://www.postgresql.org/docs/current/sql-alterdefaultprivileges.html

Simple failing example (actually from the [PostgREST docs](http://postgrest.org/en/v7.0.0/schema_structure.html)):
```sql
ALTER DEFAULT PRIVILEGES IN SCHEMA api REVOKE EXECUTE ON FUNCTIONS FROM public;
```
...results into:
```
== [db/test.sql] FAIL
L:   1 | P:   1 |  PRS | Found unparsable section: 'ALTER DEFAULT PRIVILEGES IN
                       | SCHEMA api R...'
```

alanmcruickshank:
Yep. Needs adding. Makes sense to do as part of one PR with #350 I think too.

PR #1350, opened 2021-09-02
Added full support for postgres's ALTER DEFAULT PRIVILEGES.

creste:

This is a follow-up to https://github.com/sqlfluff/sqlfluff/pull/1032 and fixes https://github.com/sqlfluff/sqlfluff/issues/348.  It adds support for all known variants of `ALTER DEFAULT PRIVILEGES...` in PostgreSQL 13.  See https://www.postgresql.org/docs/13/sql-alterdefaultprivileges.html.

The changes made in https://github.com/sqlfluff/sqlfluff/pull/1032 attempted to [re-use the `GRANT` and `REVOKE` statements](https://github.com/sqlfluff/sqlfluff/pull/1032/files#diff-928e7640b7dcb8820cec8f60a1e37aafea5721724dcebf7c51116476cc083748R2154) defined in `AccessStatementSegment`.  Unfortunately, those `GRANT` and `REVOKE` statements [allow many database objects](https://github.com/sqlfluff/sqlfluff/blob/ab11418d1c82ad8c7609cafe2b7650eba51faa3e/src/sqlfluff/dialects/dialect_ansi.py#L2464-L2498) that are not allowed in `ALTER DEFAULT PRIVILEGES` and don't allow the object types required by `ALTER DEFAULT PRIVILEGES`.  The only object types allowed in `ALTER DEFAULT PRIVILEGES` are:
- TABLES
- SEQUENCES
- FUNCTIONS
- ROUTINES
- TYPES
- SCHEMAS

Note that unlike `GRANT`/`REVOKE`, the `ALTER DEFAULT PRIVILEGES` command requires all object types to be pluralized.

Given those significant differences and some other minor differences, I decided to remove `AlterDefaultPrivilegesStatementSegment`'s dependency on `AccessStatementSegment` and separate it into it's own top-level statement with several child segments that are independent of `AccessStatementSegment`.

Finally, I did some research and it looks like PostgreSQL is the only database that supports `ALTER DEFAULT PRIVILEGES`.  Therefore, I decided to move `AlterDefaultPrivilegesSegment` from `dialect_ansi.py` to `dialect_postgres.py`.

There may be an opportunity in the future to reduce some of the duplication between `AlterDefaultPrivilegesStatementSegment` and `AccessStatementSegment`.  For example, the logic in `AlterDefaultPrivilegesToFromRolesSegment` is almost the same between the two.  I decided against doing that in this pull request because I'm not familiar enough with non-postgres dialects that depend on the current behavior of `AccessStatementSegment`.

If a user was relying on `ALTER DEFAULT PRIVILEGES` being available in the `ansi` dialect then this change will break them because it was moved to the `postgres` dialect.  As I mentioned above, I don't think anyone has been relying on this behavior.

### Pull Request checklist

Included test cases for all variants of `ALTER DEFAULT PRIVILEGES`.


WittierDinosaur (commented):
On my phone, so may be looking at this wrong... but doesn’t seem right. I’ll review in the morning, but please don’t approve beforehand

tunetheweb:
> On my phone, so may be looking at this wrong... but doesn’t seem right. I’ll review in the morning, but please don’t approve beforehand

Looks OK to me @WittierDinosaur - what have you spotted that I missed?

BTW looks like Amazon Redshift support this, but 1) we don't support this yet and 2) apparently they are based on Postgres 8 so would likely be based on Postgres if they ever did get supported added. So I agree with moving it out of ANSI.

WittierDinosaur:
Okay, whatever I thought was up wasn't. Much easier to read on pc, all good here!

tunetheweb (approved):

PR #1360, opened 2021-09-06
Rule L030 bugfix

### LINKED ISSUE #1323: Postgres dialect seems to be missing some keywords
I believe there are some keywords missing in the Postgres dialect. For example, based on this [list](https://www.postgresql.org/docs/current/sql-keywords-appendix.html), `CURRENT_DATE` should be considered a reserved keyword. However, the [ansi keywords](https://github.com/sqlfluff/sqlfluff/blob/b195e2585483fa220d88ced6c8ec271d08919170/src/sqlfluff/dialects/ansi_keywords.py) and the [postgres dialect](https://github.com/sqlfluff/sqlfluff/blob/b195e2585483fa220d88ced6c8ec271d08919170/src/sqlfluff/dialects/dialect_postgres.py#L39-L55) code don't include it. There may be other reserved words in that list that are missing (`CURRENT_TIME`, etc)

## Steps to Reproduce
```
DATEADD(day, -2, current_date)
```

## Expected Behaviour
```
DATEADD(day, -2, CURRENT_DATE)
```

## Observed Behaviour
```
DATEADD(day, -2, current_date)
```

## Dialect
postgres

## Version
sqlfluff, version 0.6.4
Python 3.7.11

## Configuration
```
[sqlfluff]
nocolor = True
dialect = postgres
templater = dbt
output_line_length = 120
ignore = parsing, templating

[sqlfluff:rules]
comma_style = leading
max_line_length = 120
tab_space_size = 2

capitalisation_policy = upper

[sqlfluff:rules:L014]  # Unquoted Identifiers

capitalisation_policy = upper

capitalisation_policy = upper
```

WittierDinosaur:
Rule L030 including bare functions, such as CURRENT_TIMESTAMP. This PR includes them, with a test case added to ensure they're caught

Fixes #1323

No

### Pull Request checklist
Test case added to the /test/rules set.


test/fixtures/rules/std_rule_cases/L030.yml
```diff
@@ -16,6 +16,16 @@ test_2:
       L030:
         capitalisation_policy: lower
 
+test_bare_functions:
+  fail_str: SELECT current_timestamp from table
+
+  fix_str: SELECT CURRENT_TIMESTAMP from table
```
tunetheweb, line 22: Might be good to mix some test cases, to ensure your early return does still cover both (I tested and it does btw):

```suggestion
test_bare_functions:
  fail_str: SELECT current_timestamp, min(a) from table

  fix_str: SELECT CURRENT_TIMESTAMP, MIN(a) from table
```

tunetheweb (commented):
Change looks good (one small comment), but not sure why CircleCI is not running?

It says this:

> No .circleci/config.yml was found in your project. Add a configuration file to your repository or refer to documentation.

But the file does exist!??!

Any ideas?

test/fixtures/rules/std_rule_cases/L030.yml
WittierDinosaur: Yeah, sure. I think the _get_fix is per segment anyway, but no harm in the extra tests. I've added it as a second test case

WittierDinosaur:
I have absolutely no idea... no files have been removed on this PR

tunetheweb:
> I have absolutely no idea... no files have been removed on this PR

Hmmm see this:

image

@alanmcruickshank / @sqlfluff/sqlfluff-maintainers any ideas?

src/sqlfluff/rules/L030.py
```diff
@@ -48,13 +51,18 @@ def _get_fix(self, segment, fixed_raw):
...
-            child
-            for child in segment.segments
-            if child.get_type() == "function_name_identifier"
-        ][0]
-        return LintFix(
-            "edit",
-            child_segment,
-            child_segment.edit(fixed_raw.strip()),
-        )
+
```
tunetheweb, line 54: Black is complaining about this blank line (though can't see it as CircleCI not running)

```suggestion
```

WittierDinosaur: Done

barrywhart:
I haven't seen this CircleCI issue. Possible it's a short term breakage on their side? Or a change in behavior? E.g. the file needs to be in a different place now?

tunetheweb:
Did you see my last update @barrywhart ? Looks like we've been suspended from CircleCI 😱

tunetheweb:
CircleCI is back!

tunetheweb (approved):

Issue #1342, opened 2021-09-01
BigQuery Parameterized Queries


rileyrunnoe:
[BigQuery](https://cloud.google.com/bigquery/docs/parameterized-queries) allows for adding parameters to queries using the `@VAR_NAME` syntax.

I am hoping to add this to the bigquery language so that these queries can be parsed.

tunetheweb:
Can you give an example query?

rileyrunnoe:
```sql
SELECT 1
from x
where y = @z
```

Z is specified as a specific value with whatever SDK you are using to interact with BQ.

PR #1363, opened 2021-09-07
Add BigQuery Parameter Lexing and Parsing

rileyrunnoe:
This feature will close https://github.com/sqlfluff/sqlfluff/issues/1342

Bigquery allows for named and ordered parameters to help prevent sql injection.

https://cloud.google.com/bigquery/docs/parameterized-queries

Just a dialect enhancement. There should be no side effects other than enhancing bigquery parsing coverage.

### Pull Request checklist
- [x]  `.sql`/`.yml` parser test cases in `test/fixtures/parser` (note YML files can be auto generated with `python


rileyrunnoe:
Big thanks to @tunetheweb as he basically wrote all of this code. I just added the tests.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -41,6 +41,8 @@
     # JSON Operators: https://www.postgresql.org/docs/9.5/functions-json.html
     [
         StringLexer("right_arrow", "=>", CodeSegment),
+        StringLexer("question_mark", "?", CodeSegment),
+        RegexLexer("atsign_literal", r"@[a-zA-Z_][\w]*", CodeSegment),
```
tunetheweb, line 45: I moved this from this:
```py
RegexLexer("atsign_literal", r"[@][\w]*", CodeSegment),
```

For the following reasons:
- It's not just the `atsign` but `atsign` followed by literal identifier. So this is a better name to make that obvious
- In BigQuery literals should start with a character or an underscore so `@1abc` is not valid but would be accepted by the original regex.
- It prevents just `@` being accepted.

I've committed that change and added test cases so this is just an FI an no further action required.

tunetheweb (approved):

PR #1364, opened 2021-09-07
Issues 854, 1321: Handle Jinja leading whitespace-only lines

### LINKED ISSUE #854: Fix command mangles file when jinja comment is present
## Expected Behaviour
Correctly fix trailing commas from input file.

Input file `stg_payments.sql`:
```sql
with source as (

    {#-
    Normally we would select from the table here, but we are using seeds to load
    our data in this project
    #}
    select * from {{ ref('raw_payments') }}

),

renamed as (

    select
        id as payment_id
        , order_id
        , payment_method
        --`amount` is currently stored in cents, so we convert it to dollars
        , amount / 100 as amount

    from source

)

select * from renamed
```

`sqlfluff fix` command line output:
```bash
== [models/staging/stg_payments.sql] FAIL
L:  10 | P:   9 | L019 | Found leading comma. Expected only trailing.
L:  11 | P:   9 | L019 | Found leading comma. Expected only trailing.
L:  13 | P:   9 | L019 | Found leading comma. Expected only trailing.
==== fixing violations ====
Attempting fixes...
Persisting Changes...
== [models/staging/stg_payments.sql] PASS
```

Expected `stg_payments.sql` output:
```sql
with source as (

    {#-
    Normally we would select from the table here, but we are using seeds to load
    our data in this project
    #}
    select * from {{ ref('raw_payments') }}

),

renamed as (

    select
        id as payment_id,
        order_id,
        payment_method,
        --`amount` is currently stored in cents, so we convert it to dollars
        amount / 100 as amount

    from source

)

select * from renamed
```

## Observed Behaviour
Actual output:

```sql
with source as (

    {#-
    Normally we would select from the table here, but we are using seeds to load
    our data in this project
    #}
    select * from {{ ref('raw_payments'renamed as (

    select
        id as payment_id,
        order_id,
        payment_method,
        --`amount` is currently stored in cents, so we convert it to dollars
        amount / 100 as amount

    from source

) source

)

select * from renamed
```

## Steps to Reproduce
`sqlfluff fix stg_payments.sql`

## Version
qlffluff: 0.4.1
dbt: 0.19.0
python: 3.7.10

## Configuration
Tested with both
```
[sqlfluff]
templater = jinja
exclude_rules = L009
```
and
```
[sqlfluff]
templater = dbt
exclude_rules = L009
```

## Notes

When the comments from lines 3-6 are remove, `sqlfluff` works as expected.

### LINKED ISSUE #1321: SQLFluff produces corrupt code with Jinja whitespace trimming
## Input

```sql
-- Beginning of file, two blank lines follow


{%- set x = 42 %}
SELECT 1, 2

```

(the blank lines are important; the comment is just to prevent the GitHub Markdown renderer from trimming the leading blank lines)

## Command

```console
$ sqlfluff fix --force test.sql
```

## Expected output

Probably something like:

```sql
-- Beginning of file, two blank lines follow


{%- set x = 42 %}
SELECT
    1,
    2
```

## Actual output

```sql
-- Beginning of file, two blank lines follow


{%- set x = 42 SELECT
    1,
    2, 2
```

## Version
Reproducible with SQLFluff 0.6.4 and master on Python 3.9.6

## Configuration

None.

barrywhart:
Fixes #854, #1321 
Preserves leading whitespace-only lines in SQL files when the Jinja templater is used.

### Pull Request checklist

  - Other.


barrywhart:
The CircleCI steps won't run until next week or we get the usage limits raised/removed, whichever comes first.

tunetheweb (approved):

tunetheweb (changes requested):
Actually going to un-approve as using non-collapsing whitespace now breaks, whereas previously it worked.

Example:

```sql



{% set x = 42 %}
SELECT 1,2
```

Now returns:

```sql



{% set x = 42 %}
SELSELECT
1,
2
```

Because we're now adding the whitespace twice with this change.

I think it's best to just say we don't support `{%-` and `-%}` for now,

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -326,8 +327,16 @@ def _slice_template(cls, in_str: str) -> Iterator[RawFileSlice]:
...
+        def lex(env, in_str):
+            # This is ugly. Jinja discards leading whitespace-only lines. We
+            # have to detect and add them back in order for SQLFluff's source
+            # position markers to align with the actual file.
+            m = re.search(r'^\s+\n', in_str)
+            if m:
+                yield None, "data", m.group(0)
+            # https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.Environment.lex
+            yield from env.lex(in_str)
+        for _, elem_type, raw in lex(env, in_str):
```
tunetheweb, line 339: Actually I found a real cheeky fix that seems to work for both #1321 and #854 now you've done the hard work and led the way:

```suggestion

        # Using Jinja whitespace stripping (e.g. `{#-` or `-%}`) breaks the
        # position markers between unlexed and lexed file. So let's ignore any
        # request to do that before lexing, by replacing '-' with '+'
        #
        # Note: '+' is the default, so shouldn't really be needed but we
        # explicitly state that to preserve the space for the missing '-' character
        # so it looks the same.
        in_str = in_str.replace('{%-', '{%+')
        in_str = in_str.replace('-%}', '+%}')
        in_str = in_str.replace('{#-', '{#+')
        in_str = in_str.replace('-#}', '+#}')

        for _, elem_type, raw in env.lex(env, in_str):
```

If so we should remove the `re` import above as no longer needed and also enhance this with other test cases (the example above I suggested and the one from #854).

src/sqlfluff/core/templaters/python.py
```diff
@@ -934,7 +934,7 @@ def _split_uniques_coalesce_rest(
                         if last_owu_idx is None or last_owu_idx + 1 >= len(
                             int_file_slice.slice_buffer
                         ):
-                            cur_idx = 0
+                            cur_idx = 0  # pragma: noqa
```
tunetheweb, line 937:
```suggestion
                            cur_idx = 0  # pragma: no cover
```

tunetheweb (approved):
One small fix (without which CI won't pass anyway I don't think) but other than that this LGTM!

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -326,6 +326,18 @@ def _slice_template(cls, in_str: str) -> Iterator[RawFileSlice]:
...
+        # position markers between unlexed and lexed file. So let's ignore any
+        # request to do that before lexing, by replacing '-' with '+'
+        #
+        # Note: '+' is the default, so shouldn't really be needed but we
+        # explicitly state that to preserve the space for the missing '-' character
+        # so it looks the same.
+        in_str = in_str.replace("{%-", "{%+")
+        in_str = in_str.replace("-%}", "+%}")
+        in_str = in_str.replace("{#-", "{#+")
+        in_str = in_str.replace("-#}", "+#}")
```
tunetheweb, line 340: Looks like dbt doesn't need this workaround and inherits this, so need to stick an `if` check in here:

```suggestion
        #
        # Note dbt inherits this class and does not need this workaround
        if cls.name == "jinja":
            in_str = in_str.replace("{%-", "{%+")
            in_str = in_str.replace("-%}", "+%}")
            in_str = in_str.replace("{#-", "{#+")
            in_str = in_str.replace("-#}", "+#}")
```

tunetheweb (changes requested):

barrywhart:
I dug through the voluminous log for that failed test. I think this is the error:
```
Line is too long
```

So... I wonder if dbt would also benefit from this fix, but a side effect of our fix is that the post-rendered code has a line that is too long. Note there is a [known issue](https://github.com/sqlfluff/sqlfluff/issues/845) where SQLFluff incorrectly checks line length after template expansion, not before.

Would you be okay if I tweaked that test to ignore "line too long" errors? Since most of our users use dbt, I hate to deliberately not fix a nasty bug that is likely to affect them. Happy to loop in another maintainer who is more familiar with dbt if you think that's wise. I am just a Jinja person myself. 🙂

tunetheweb:
SGTM

barrywhart:
@NiallRees: With the changes in this PR, one of the dbt tests is failing for reasons I don't understand -- L003 is reporting an issue. I am not familiar with the dbt templater. I think this fix is important both for jinja and for dbt, so I'd like to have both templaters use this fix if we are confident it's working okay.

src/sqlfluff/core/templaters/dbt.py
```diff
@@ -259,13 +259,16 @@ def process(self, *, fname, in_str=None, config=None, formatter=None):
...
             # Increment the counter
             self._sequential_fails += 1
-            return None, [
-                SQLTemplaterError(
-                    f"dbt compilation error on file '{e.node.original_file_path}', {e.msg}",
-                    # It's fatal if we're over the limit
-                    fatal=self._sequential_fails > self.sequential_fail_limit,
-                )
-            ]
+            if e.node:
```
barrywhart, line 262: This change fixes an unrelated bug I discovered: For some errors, a `DbtCompilationException` object has a `node` value of `None`, in which case referencing `e.node.original_file_path` fails. Worse, this prevents the user from seeing the original error message.

tunetheweb (approved):

CyberShadow:
Thank you!

I tried this on our codebase and this seems to be the next problem: https://github.com/sqlfluff/sqlfluff/issues/1371

Issue #1366, opened 2021-09-08
Non-terminated Delimited() functions affect the behaviour of later Delimited() functions in the match grammar


WittierDinosaur:
In the following match grammar, the first Delimited() block causes the second to not work, because the first doesn't have the terminator kwarg.

```
        Sequence(
            "GRANT",
            OneOf(
                Sequence(
                    Delimited(
                        OneOf(_global_permissions, _permissions),
                        delimiter=Ref("CommaSegment"),
                    ),
                    "ON",
                    _objects,
                ),
                Sequence("ROLE", Ref("ObjectReferenceSegment")),
                Sequence("OWNERSHIP", "ON", "USER", Ref("ObjectReferenceSegment")),
                # In the case where a role is granted non-explicitly,
                # e.g. GRANT ROLE_NAME TO OTHER_ROLE_NAME
                # See https://www.postgresql.org/docs/current/sql-grant.html
                Ref("ObjectReferenceSegment"),
            ),
            "TO",
            OneOf("GROUP", "USER", "ROLE", "SHARE", optional=True),
            Delimited(
                Ref("ObjectReferenceSegment"),
                Ref("FunctionSegment"),
                "PUBLIC",
            ),
            OneOf(
                Sequence("WITH", "GRANT", "OPTION"),
                Sequence("WITH", "ADMIN", "OPTION"),
                Sequence("COPY", "CURRENT", "GRANTS"),
                optional=True,
            ),
            Sequence(
                "GRANTED",
                "BY",
                OneOf(
                    "CURRENT_USER",
                    "SESSION_USER",
                    Ref("ObjectReferenceSegment"),
                ),
                optional=True,
            ),
        )
```

Separately, as Delimited() inherits from OneOf() and not Sequence(), it doesn't create a Segment, like something like Bracketed() does. This issue does not occur if the second Delimited() block exists in its own segment. If it were wrapped in Bracketed() for example, there isn't a scope problem, and the code works as expected. Perhaps in this fix we should make Delimited() inherit from Sequence()?

## Expected Behaviour
The first Delimited() block should naturally terminate when it hits a segment that doesn't fit its criteria.

## Observed Behaviour
The first Delimited() block stops the second Delimited() block from working for any list size above 1. 

## Steps to Reproduce
The match block above is for a bugfix in the ANSI GRANT grammar. Without a terminator specified for the first Delimited() block, this example would fail:

```
[...]

PR #1367, opened 2021-09-08
Added ability to Grant and Revoke Grant to multiple users

### LINKED ISSUE #1365: Granting to multiple roles in postgres
## Expected Behaviour

Parsing and linting correctly handle a `GRANT` statement to multiple roles/grantees in postgres dialect.

From [postgres docs](https://www.postgresql.org/docs/current/sql-grant.html):
```
GRANT { { SELECT | INSERT | UPDATE | DELETE | TRUNCATE | REFERENCES | TRIGGER }
    [, ...] | ALL [ PRIVILEGES ] }
    ON { [ TABLE ] table_name [, ...]
         | ALL TABLES IN SCHEMA schema_name [, ...] }
    TO role_specification [, ...] [ WITH GRANT OPTION ]
```

## Observed Behaviour

Parse error:

```
❯ echo 'GRANT SELECT ON abc TO xyz, mno;' | sqlfluff parse --dialect postgres -
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |        unparsable:                                           !! Expected: 'StatementSegment'
[L:  1, P:  1]      |            raw:                                              'GRANT'
[L:  1, P:  6]      |            whitespace:                                       ' '
[L:  1, P:  7]      |            raw:                                              'SELECT'
[L:  1, P: 13]      |            whitespace:                                       ' '
[L:  1, P: 14]      |            raw:                                              'ON'
[L:  1, P: 16]      |            whitespace:                                       ' '
[L:  1, P: 17]      |            raw:                                              'abc'
[L:  1, P: 20]      |            whitespace:                                       ' '
[L:  1, P: 21]      |            raw:                                              'TO'
[L:  1, P: 23]      |            whitespace:                                       ' '
[L:  1, P: 24]      |            raw:                                              'xyz'
[L:  1, P: 27]      |            comma:                                            ','
[L:  1, P: 28]      |            whitespace:                                       ' '
[L:  1, P: 29]      |            raw:                                              'mno'
[L:  1, P: 32]      |    statement_terminator:                                     ';'
[L:  1, P: 33]      |    newline:                                                  '\n'

==== parsing violations ====
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'GRANT SELECT ON abc
                       | TO xyz, mno'


❯ echo 'GRANT SELECT ON abc TO xyz, mno;' | sqlfluff lint --dialect postgres -
== [stdin] FAIL
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'GRANT
                       | SELECT ON abc TO xyz, mno'
All Finished 📜 🎉!
```

## Dialect

Postgres

## Version
```
❯ sqlfluff --version
sqlfluff, version 0.6.4

❯ python3 --version
Python 3.9.6
```

## Configuration

No custom configuration

WittierDinosaur:
Allows GRANT and REVOKE GRANT for multiple users. A simple change that changes a OneOf() to Delimited()

No. But it did bring to light a bug with Delimited(), as explained in #1366 

Fixes #1365


tunetheweb (approved):
The only downside (and probably worth noting on the "Are there any other side effects..." section) is this will enable this multiuser syntax in all dialects.

Now I don't think that's necessarily a bad thing (BigQuery supports it for example and others may too). Plus the access control is already not 100% accurate for all dialects. Also it won't raise positive positives (it just want block false negatives is a dialect that doesn't support this uses it.

Anyway, I thought worth pointing out...

WittierDinosaur:
Yeah, I think that's a decision to be made on this block.

From the comments at the top of the class:

```
    In order to help reduce code duplication we decided to implement other dialect specific grants (like Snowflake)
    here too which will help with maintainability. We also note that this causes the grammar to be less "correct",
    but the benefits outweigh the con in our opinion.
```

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2489,10 +2490,11 @@ class AccessStatementSegment(BaseSegment):
             ),
             "TO",
             OneOf("GROUP", "USER", "ROLE", "SHARE", optional=True),
-            OneOf(
+            Delimited(
                 Ref("ObjectReferenceSegment"),
                 Ref("FunctionSegment"),
                 "PUBLIC",
+                delimiter=Ref("CommaSegment"),
```
tunetheweb, line 2497: I'm not sure it's entirely obvious that `Delimited` is the same as `OneOf` (even though it is as you note in #1366). should we wrap this in an actual `OneOf` to make this more obvious (also less to fix if we ever do change `OneOf` to inherit from `Sequence` as you've suggested).

```suggestion
            Delimited(
                OneOf(
                    Ref("ObjectReferenceSegment"),
                    Ref("FunctionSegment"),
                    "PUBLIC",
                ),
                delimiter=Ref("CommaSegment"),
```

There are that many instances of `Delimited` with more than one child, but seems a mixed bag of whether to use a nested OneOf or not (lines 722, 1626, 2088, 2226 in this file do use `OneOf` for example, lines 1397, 1472, 1738 do not).

Personally I think it's clearer if it is in there, even if it's not strictly necessary. WDYT?

WittierDinosaur: No strong opinion, so I'll change to that.

tunetheweb (approved):
LGTM - thanks @WittierDinosaur !

PR #1369, opened 2021-09-09
Enhanced the Postgres grammar for create table

### LINKED ISSUE #1095: PostgreSQL support for "PARTITION BY"

## Expected Behaviour
I would expect statements with "PARTITION BY"  to be parsed by sqlfluff, as they are supported by PostgreSQL.

L:   6 | P:   2 |  PRS | Found unparsable section: ' PARTITION BY RANGE
                       | (logdate)'

## Steps to Reproduce
>CREATE TABLE measurement (
    city_id int NOT NULL,
    logdate date NOT NULL,
    peaktemp int,
    unitsales int
) PARTITION BY RANGE (logdate);

Lint the file with the following command(s) :
>sqlfluff lint test.sql --dialect postgres or sqlfluff lint test.sql
## Version
I am using sqlfluff version 0.5.6 and python version 3.8.5.

This example comes from the official PostgreSQL [documentation](https://www.postgresql.org/docs/10/ddl-partitioning.html). I found similar results when executing queries from Github Project [vertx-jooq](https://github.com/jklingsporn/vertx-jooq).

### LINKED ISSUE #1172: Postgres: CREATE UNLOGGED unparsable
Parsing issue: when parsing code:
```
CREATE UNLOGGED TABLE staging (
    event_type INTEGER
    , event_time TIMESTAMP
    , user_email VARCHAR
    , phone_number VARCHAR
    , processing_date DATE
    , PRIMARY KEY (event_type, event_time, user_email, phone_number, processing_date)
);
```
linter reports unparsable section. The same with parser: unparsable:  !! Expected: 'StatementSegment'

## Expected Behaviour
This is correct code, working as expected - creating an UNLOGGED table. When UNLOGGED keyword is removed, linter works and no errors are detected.

## Observed Behaviour
==== sqlfluff ====
sqlfluff:                0.6.0       python:                  3.8.2
implementation:  cpython  dialect:              postgres
verbosity:            1             templater:               jinja

== [main\models\templates\event_staging_table.sql] FAIL
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'CREATE UNLOGGED TABLE staging (\n even...'

## Steps to Reproduce
run linter on the cited code

## Version
sqlfluff, version 0.6.0
Python 3.8.2

## Configuration
```
[sqlfluff]
dialect = postgres
templater = jinja
output_line_length = 120
runaway_limit = 10
#disabled rule L031 [avoid table aliases] - aliasing reduces clutter in code
exclude_rules = L031

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 120
indent_unit = space
comma_style = leading

[sqlfluff:rules:L010]
capitalisation_policy = upper

[sqlfluff:rules:L014]

[sqlfluff:rules:L030]
capitalisation_policy = upper
```

WittierDinosaur:
Enhanced the Create Table grammar for Postgres, as per the docs https://www.postgresql.org/docs/13/sql-createtable.html.

There is some crossover between create and alter table, so this PR helped point out a couple of errors in that code, which I've fixed on the way.

Fixes #1172, Fixes #1095

No

### Pull Request checklist
Added all examples from the Postgres create table docs as tests. 5 of them have been commented out, as they won't parse due to unrelated missing features. 

4 removed due to us not being able to parse the Interval data type grammar, 1 because we can't parse GIST expression grammar. Eg. gist(c WITH &&), we don't have the operators added yet.

Added examples from the issues as test cases.


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -429,7 +584,7 @@ class AlterTableActionSegment(BaseSegment):
             Ref("ColumnReferenceSegment"),
             Ref("DatatypeSegment"),
             Sequence("COLLATE", Ref("QuotedLiteralSegment"), optional=True),
-            Ref("ColumnOptionSegment", optional=True),
+            AnyNumberOf(Ref("ColumnConstraintSegment", optional=True)),
```
tunetheweb, line 587: Do we need `optional=True` for `AnyNumberOf`?

tunetheweb (approved):
Looks good! One small question.

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: Nope, removed!

PR #1376, opened 2021-09-10
Simplify rule 030

### LINKED ISSUE #1374: Function dataset prefix keeps getting prepended
_If this is a parsing or linting issue, please include a minimal SQL example which reproduces the issue, along with the `sqlfluff parse` output, `sqlfluff lint` output and `sqlfluff fix` output when relevant._

## Expected Behaviour

The prefix should continue to occur once in the output.

## Observed Behaviour

The prefix seems to be prepended until the fix loop limit is reached.

## Steps to Reproduce

```console
$ printf 'SELECT project1.FOO(value) AS value\n' | sqlfluff fix -  
SELECT
    project1.PROJECT1.PROJECT1.PROJECT1.PROJECT1.PROJECT1.PROJECT1.PROJECT1.PROJECT1.PROJECT1.PROJECT1.FOO(
        value
    ) AS value
```

## Dialect

None specified

## Version
SQLFluff c3a1f703e6d34fd2b293ef17387fe603cf910b82, Python 3.9.6

## Configuration

None

tunetheweb:

Fixes #1374 

Not sure why rule R030 needs to be so complicated - am I missing something?

Don't _think_ so.... unless I'm missing something?

### Pull Request checklist

  - Other.


alanmcruickshank:
I think the complexity here is legacy from when `function_name_identifier` didn't exist, and we only had `function_name`. Assuming tests pass, I don't see why we can't merge this.

tunetheweb:
But the rule depends on `function_name_identifer` existing 🤷‍♂️🤷‍♂️🤷‍♂️

tunetheweb:
Maybe previous it got the identifier as a subset of `function_name` and when that was updated to `function_name_identifier` the rule was simply updated without realising it could be simplified at same time?

tunetheweb:
> Maybe previous it got the identifier as a subset of `function_name` and when that was updated to `function_name_identifier` the rule was simply updated without realising it could be simplified at same time?

Yeah think that was the case: https://github.com/sqlfluff/sqlfluff/commit/93c4702804a63da6b39a7b77ecdcf4f515dbe74a#

Anyway this seems good to merge now. Ignore the Circle CI error - needs to be turned off.

barrywhart (approved):

src/sqlfluff/rules/L030.py
```diff
@@ -40,28 +39,10 @@ class Rule_L030(Rule_L010):
...
     ]
     _description_elem = "Function names"
 
     def _get_fix(self, segment, fixed_raw):
-        """Overrides the base class.
-
-        We need to do this because function_name nodes have a child
-        function_name_identifier that holds the actual name.
-        """
-        if segment.type == "function_name":
```
barrywhart, line 54: I'm trying to remember -- I have a vague thought that maybe this was for BigQuery support, where there can be a project name, dataset name, AND function name. Does `git blame` show who did this?

If the above is correct, seems like we're missing a test case.

barrywhart: It's possible the parse structure has changed since this code was added; maybe necessary at some point. I wouldn't stress too much about it. We keep refining, fixing, etc.

tunetheweb: I wouldn't like to name and shame anyone @barrywhart but have a look at above commit https://github.com/sqlfluff/sqlfluff/commit/93c4702804a63da6b39a7b77ecdcf4f515dbe74a :-)

barrywhart: I think I named myself earlier. As much as I could without checking for myself. Another reason it is good to have a second Barry on the project! 🎉🤣

tunetheweb: Can never have enough Barrys on a project! 😁

PR #1377, opened 2021-09-10
Postgres analyze

### LINKED ISSUE #1041: ANALYZE/ANALYSE unparseable
## Expected Behaviour

`ANALYZE` should be accepted. Postgres suggests `ANALYSE` as equally valid (https://www.postgresql.org/docs/current/sql-keywords-appendix.html). 

I'm not clear if `ANALYZE` is part of ANSI (I suspect not), but it's supported by Postgres, MySQL, and Oracle (but not SQL Server AFAICT). I don't have an opinion on whether SQLFluff should allow it under ANSI dialect or only under specific dialects, I'd just like it to have a way to lint SQL that contains `ANALYZE`.

## Observed Behaviour

Rejected as unparsable

## Steps to Reproduce

```
$ echo "ANALYZE foo;" | sqlfluff lint -
== [stdin] FAIL
L:   1 | P:   1 |  PRS | Found unparsable section: 'ANALYZE foo'
WARNING: Parsing errors found and dialect is set to 'ansi'. Have you configured your dialect?
```

## Version
```
$ sqlfluff --version
sqlfluff, version 0.5.2

$ python -V
Python 3.8.8
```

## Configuration
I'm using the standard 0.5.2 config. I get the same outcome when using Postgres dialect.

WittierDinosaur:
Full support added for ANALYZE, as per https://www.postgresql.org/docs/13/sql-analyze.html

Fixes #1041

No

### Pull Request checklist
No examples in Postgres docs, so added enough examples for every permutation.


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1396,6 +1396,31 @@ class CreateIndexStatementSegment(BaseSegment):
...
+    )
+
+    _tables_and_columns = Sequence(
+        Ref("TableReferenceSegment"),
+        Bracketed(Delimited(Ref("ColumnReferenceSegment")), optional=True),
+    )
+
+    match_grammar = Sequence(
+        OneOf("ANALYZE", "ANALYSE"),
+        OneOf(Bracketed(Delimited(_option)), "VERBOSE", optional=True),
```
tunetheweb, line 1419: Did you see this bit:
> When the option list is surrounded by parentheses, the options can be written in any order. The parenthesized syntax was added in PostgreSQL 11; the unparenthesized syntax is deprecated.

Should we support the unparenthesized in case any on else is still using it? Or is it deprecated enough that we don't need to support it?

```suggestion
        OneOf(Bracketed(Delimited(_option)), Delimited(_option), "VERBOSE", optional=True),
```

WittierDinosaur: The unparenthesized syntax it's referring to is just VERBOSE. Parenthesized was introduced in pg11, and SKIP_LOCKED as an option was introduced in pg12. So there is no postgres version where unparenthesized SKIP_LOCKED or VERBOSE boolean would run

tunetheweb: Ah gotcha. Resolving.

tunetheweb (approved):

PR #1378, opened 2021-09-10
Improve datatypes: CHARACTER VARYING for ANSI, and Postgres and also TIMESTAMP AT TIME ZONE for Postgres

### LINKED ISSUE #733: [Postgres] unparseable error on `character varying` type

## Expected Behaviour

```sql
CREATE TABLE main.test_table (
    "col1" character varying(40) NOT NULL
);
```

Linting the DDL statement would be success.

## Observed Behaviour

The linting would be failed.

```bash
$ sqlfluff lint --dialect postgres postgres/table/test.sql
== [postgres/table/test.sql] FAIL
L:   1 | P:   1 |  PRS | Found unparsable section: 'CREATE TABLE main.test_table
                       | (\n "col1...'
```

If we update the DDL to use `varchar` instead linting will success.

```diff
CREATE TABLE main.test_table (
-    "col1" character varying(40) NOT NULL
+    "col1" varchar(40) NOT NULL
);
```

## Version

```bash
$ sqlfluff --version
sqlfluff, version 0.3.6

$ python -V
Python 3.9.1
```

## Configuration

No configurations, using default.

### LINKED ISSUE #1331: Parsing support for postgres `at time zone`
Hello folks, Postgres's `at time zone` conversion syntax seems to throw a parser error. Seems like `with time zone`/`without time zone` also. Postgres doesn't seem to have a function call version so I believe this is the only way to do it.

`example.sql`

```sql
SELECT ('2015-10-24 16:38:46'::timestamp AT TIME ZONE 'UTC');
```

## Observed Behaviour

```
==== sqlfluff ====
sqlfluff:                0.6.4 python:                 3.7.11
implementation:        cpython dialect:              postgres

==== readout ====

=== [ path: example.sql ] ===

== [example.sql] FAIL
L:   1 | P:  41 |  PRS | Line 1, Position 41: Found unparsable section: " AT TIME
                       | ZONE 'UTC'"
==== summary ====
violations:        1 status:         FAIL
All Finished 📜 🎉!
```

## Steps to Reproduce

```bash
sqlfluff lint --dialect postgres -v example.sql 
```

WittierDinosaur:
Added CHARACTER VARYING as parsable data types to ANSI
Added CHARACTER VARYING and TIMESTAMP AT TIME ZONE "MY_TIMEZONE" as parsable data types to Postgres

Fixes #1331 
Fixes #733

No

### Pull Request checklist
Tests added for both of the bugs' examples.


tunetheweb:
> Added to ANSI, as I don't think having these available is breaking to other languages.

I'm not so sure about that decision to be honest. BigQuery, TeraData, and T-SQL already have their own definition of `DataTypeSegment`. Should we not just add one for Postgres unless there's a good reason to add this to ANSI.

WittierDinosaur:
Can we make a deal? Character Varying is an ANSI data type. TIMESTAMP AT TIME ZONE is dialect-specific. But, ANSI is also missing Data Types. For example, CHARACTER LARGE OBJECT. This PR fixes live issues. I agree that TIMESTAMP AT TIME ZONE should be Postgres-only. But I think the types logic needs an overhaul anyway, it's too generic right now. So can we merge this, then look at setting the types right across dialects?

tunetheweb:
> Can we make a deal? Character Varying is an ANSI data type. TIMESTAMP AT TIME ZONE is dialect-specific. But, ANSI is also missing Data Types. For example, CHARACTER LARGE OBJECT. This PR fixes live issues. I agree that TIMESTAMP AT TIME ZONE should be Postgres-only. But I think the types logic needs an overhaul anyway, it's too generic right now. So can we merge this, then look at setting the types right across dialects?

What's your thinking here for not splitting ANSI and Postgres now? While I agree ANSI is not complete, that's not a reason to knowingly add an incorrect one to that, when we're going to have to split our Postgres at some point anyway. Especially when lots of other dialects already have their own implementations of this section. If it required a whole lot of reengineering I could maybe see your point but since it simply requires a copy and paste of the ANSI dialog to the Postgres dialect and then removing the "TIMESTAMP AT TIME ZONE" that your added from ANSI, then why not just do it in this PR?

WittierDinosaur:
Okay, fair point. Changed.

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -527,19 +527,28 @@ class DatatypeSegment(BaseSegment):
         Sequence(
             OneOf("time", "timestamp"),
             Bracketed(Ref("NumericLiteralSegment"), optional=True),
-            Sequence(OneOf("WITH", "WITHOUT"), "TIME", "ZONE", optional=True),
+            OneOf(
+                Sequence(OneOf("WITH", "WITHOUT"), "TIME", "ZONE"),
+                optional=True,
+            ),
```
tunetheweb, line 533: Might as well revert this change now?

```suggestion
            Sequence(OneOf("WITH", "WITHOUT"), "TIME", "ZONE"), optional=True),
```

tunetheweb (approved):
Thanks for humouring me!

Just one small thing I spotted and then think we're good to merge.

PR #1380, opened 2021-09-11
CREATE, ALTER, DROP SEQUENCE support, with Postgres extensions

WittierDinosaur:
This PR adds CREATE, ALTER, and DROP SEQUENCE into the dialects. It specifies the statement blocks in ANSI, and Postgres. ANSI written to match Oracle Docs, Postgres to match Postgres 13 docs. Also adds the sequence options functionality to Postgres' Alter Table statement. 

Adds these blocks to all Dialects where previously unsupported - might need dialect-specific additions.

### Pull Request checklist
Tests written in ANSI and Postgres trying to cover all individual options, and some random combinations. 
Docstrings added, linking to online docs.
Tests added to Alter Table for new functionality.


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -786,7 +790,7 @@ class ColumnConstraintSegment(BaseSegment):
     This matches the definition in https://www.postgresql.org/docs/13/sql-altertable.html
     """
 
-    type = "column_constraint"
+    type = "column_constraint_segment"
```
tunetheweb, line 793: This is called `column_coinstraint` in ANSI. Should we leave as is, or change both?

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -887,7 +893,7 @@ class TableConstraintSegment(BaseSegment):
     As specified in https://www.postgresql.org/docs/13/sql-altertable.html
     """
 
-    type = "table_constraint_definition"
+    type = "table_constraint_segment"
```
tunetheweb, line 896: Same comment here as it's `table_constraint_definition` in ANSI.

tunetheweb (changes requested):
Looks good. Two small comments that I think we should resolve one way or another before merging.

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: In my opinion we should go for column_constraint and table_constraint respectively. We aren't necessarily consistent with type naming, but those would follow closest to the general theme

tunetheweb: I agree yours are better names. I don’t agree with being inconsistent between dialects for the same thing.

So, as I say, happy either reverting to the slightly worse names. Or updating the ANSI dialect to the new names as part of this PR.

barrywhart: I agree with @tunetheweb: Names (and parse trees) should be consistent across dialects whenever practical, because the lint rules examine that structure when looking for issues. We don't want the rules to have to worry about unnecessary differences between dialects.

WittierDinosaur: Yeah, just pushed a commit aligning them for TableConstraintSegment and ColumnConstraintSegment across the dialects

tunetheweb (approved):

PR #1381, opened 2021-09-11
Bugfix, frame clauses in window functions were not working

### LINKED ISSUE #1379: Support redshift 'ignore nulls' clause
We've been using postgres as a standin for redshift, and this is the first time i've seen some redshift-specific syntax get in the way (credit to @WittierDinosaur).

I adapted this [query](https://online.sqlfluff.com/fluffed?sql=H4sIAENlpWAC_2WOSwqDQBBE9wHv0EuF3EnmU8rA2C09PUpuH40kDqSWrx5UFWQE6x50ZANXFHOGZwvgrLSA3fIVpqTFxs3liv7XUZpZFMQ15zJcomzQfnVqyZIw-Vczdhn_EY3QWz1vUEQJt6-yF_KwHWCq7KVyRKRVERATz-Q4NnySnGU_-ECuUJA8nne7x6SykDmf8QFvWEDYMRMBAAA%3D&dialect=postgres) from the [redshift docs](https://docs.aws.amazon.com/redshift/latest/dg/r_WF_first_value.html):

```sql
select
    venuestate,
    venueseats,
    venuename,
    first_value(venuename ignore nulls)
    over(partition by venuestate
                      order by venueseats desc
        rows between unbounded preceding and unbounded following) as col_name
from table_name
```

I suppose this belongs on the laundry list of reasons to one day support a dedicated redshift dialect.

WittierDinosaur:
IGNORE/ RESPECT NULLS were incorrectly in the PostFunctionGrammar, making window functions including them unparsable. This has been fixed. Also made the FrameClauseSegment more explicit, as it's specified in the Redshift docs.

Fixes #1379

No. Some test cases now have more explicit Frame Clauses.

### Pull Request checklist
Tested, no previous tests have been broken.
Added example in #1379 to Postgres tests, and parses correctly.
Docs link to ANSI and Postgres docs specifying Frame Clause


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -947,16 +945,40 @@ class PartitionClauseSegment(BaseSegment):
...
     type = "frame_clause"
-    match_grammar = StartsWith("ROWS")
-    # TODO: Expand a parse statement here properly to actually
-    # parse rather than assuming that it's good.
-    # parse_grammar = Sequence(
-    #    'ROWS',
-    #    ...
-    # )
+    match_grammar = Sequence(
+        "ROWS",
```
tunetheweb, line 955: Should we support `RANGE` and `GROUPS` too?

https://www.postgresql.org/docs/13/sql-expressions.html

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1603,7 +1625,7 @@ class OrderByClauseSegment(BaseSegment):
                 # for now.
                 Sequence("NULLS", OneOf("FIRST", "LAST"), optional=True),
             ),
-            terminator=Ref.keyword("LIMIT"),
+            terminator=OneOf(Ref.keyword("LIMIT"), Ref.keyword("ROWS")),
```
tunetheweb, line 1628: This would need to change too if we did add `RANGE` and `GROUPS`.

tunetheweb (commented):
Looks good. One suggestion.

src/sqlfluff/dialects/dialect_ansi.py
WittierDinosaur: Ahh okay, this is bigger than I thought. It varies by dialect. I'll rewrite for ANSI and Postgres

WittierDinosaur: It's much beefier now

tunetheweb (approved):
LGTM if you can fix the merge conflict.

Nice use of reducing duplicated code through those `_frame*` definitions.

PR #1382, opened 2021-09-11
Issue 1344: L016: "sqlfluff fix" adds too many newlines

### LINKED ISSUE #1344: sqlfluff fix adding too many newlines to bench_002_pearson.sql
This happens when running tox -e bench, output can be observed in /benchmarks/bench/bench_002_pearson_fix.sql

## Version
Current main, I jumped back about 20 commits, and it still happens

## Configuration
No deviation from standard configs in main branch

barrywhart:
Fixes #1344 

L016 "Line is too long" has a bug that causes it to behave pathologically if it finds a line with an inline comment that is:
* Is on a line with other code
* Exceeds the maximum line length even after moving it to a line by itself

E.g.
```
    ,--  the "y variable" and uses_small_subject_line to be the "x variable" in terms of the regression line.
```

In this case, L016 keeps adding blank lines endlessly until it hits the `sqlfluff fix` iteration limit (typically 10).
### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L016.py
+++ b/src/sqlfluff/rules/L016.py
@@ -391,7 +391,6 @@ class Rule_L016(Rule_L003):
 
             # We'll need the indent, so let's get it for fixing.
             line_indent = []
-            idx = 0
             for s in this_line:
                 if s.name == "whitespace":
                     line_indent.append(s)
@@ -429,14 +428,25 @@ class Rule_L016(Rule_L003):
                         idx -= 1
                     else:
                         break  # pragma: no cover
+                create_elements = line_indent + [this_line[-1], segment]
+                if sum(len(s.raw) for s in create_elements) > self.max_line_length:
+                    # The inline comment is NOT on a line by itself, but even if
+                    # we move it onto a line by itself, it's still too long. In
+                    # this case, the rule should do nothing, otherwise it
+                    # triggers an endless cycle of "fixes" that simply keeps
+                    # adding blank lines.
+                    self.logger.info(
+                        "Unfixable inline comment, too long even on a line by itself: %s",
+                        this_line[-1],
+                    )
+                    if self.ignore_comment_lines:
+                        return LintResult()
+                    else:
+                        return LintResult(anchor=segment)
                 # Create a newline before this one with the existing comment, an
                 # identical indent AND a terminating newline, copied from the current
                 # target segment.
-                create_buffer = [
-                    LintFix(
-                        "create", this_line[0], line_indent + [this_line[-1], segment]
-                    )
-                ]
+                create_buffer = [LintFix("create", this_line[0], create_elements)]
                 return LintResult(anchor=segment, fixes=delete_buffer + create_buffer)
 
             fixes = self._eval_line_for_breaks(this_line)
--- a/test/fixtures/linter/autofix/ansi/008_looping_rules_l003_l016_l019/before.sql
+++ b/test/fixtures/linter/autofix/ansi/008_looping_rules_l003_l016_l019/before.sql
@@ -1,0 +1,13 @@
+SELECT
+COUNT(1) AS campaign_count,
+state_user_v_peer_open
+,business_type
+
+-- The following is the slope of the regression line. Note that CORR (which is the Pearson's correlation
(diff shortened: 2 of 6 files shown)


src/sqlfluff/rules/L016.py
```diff
@@ -391,7 +391,6 @@ def _eval(self, segment, raw_stack, **kwargs):
 
             # We'll need the indent, so let's get it for fixing.
             line_indent = []
-            idx = 0
```
barrywhart, line 394: My IDE identified this as an unused variable, so I removed it. This is not related to the fix.

src/sqlfluff/rules/L016.py
```diff
@@ -429,14 +428,25 @@ def _eval(self, segment, raw_stack, **kwargs):
...
                     else:
                         break  # pragma: no cover
+                create_elements = line_indent + [this_line[-1], segment]
+                if sum(len(s.raw) for s in create_elements) > self.max_line_length:
+                    # The inline comment is NOT on a line by itself, but even if
+                    # we move it onto a line by itself, it's still too long. In
+                    # this case, the rule should do nothing, otherwise it
+                    # triggers an endless cycle of "fixes" that simply keeps
+                    # adding blank lines.
+                    self.logger.info(
```
barrywhart, line 438: Reviewers: The lines inside the `if` are very similar to existing lines 407-413, only the condition is different.

test/fixtures/linter/autofix/ansi/008_looping_rules_l003_l016_l019/before.sql
```diff
@@ -0,0 +1,13 @@
+SELECT
```
barrywhart, line 1: This is sort of a real-world test for the fix, adapted from the existing file mentioned in the issue, `benchmarks/bench/bench_002_pearson_fix.sql`. 

During the analysis of the issue, three rules were primarily involved in cleaning up the file, and this seemed like a good test case to add to the suite.

tunetheweb (approved):
LGTM - nice find!

One question: Why didn't the timing query bomb out when the LOOP LIMIT was reached? Should that be changed from a WARNING to an ERROR to help us spot these issues in the future?

barrywhart:
The loop limit did stop it, but it was adding one additional blank line each time, so it added a bunch of them before it stopped.

I think I'd be okay changing it to an error if we also raised the default limit a bit, maybe to 20 or 25.

Related: Did you see my note about how it was moving the comma up one line at a time? IMHO, that's at best a poorly written rule, _probably_ a bug. Imagine if there's a really long block of comments between the initial location of a leading comma and its final, desired, trailing location above. It's poor design for the linter to require 'n' iterations to move a comma by 'n' lines. Think I should create an issue for that?

tunetheweb:
> The loop limit did stop it, but it was adding one additional blank line each time, so it added a bunch of them before it stopped.
> I think I'd be okay changing it to an error if we also raised the default limit a bit, maybe to 20 or 25.


My point is reaching the loop limit implies there’s a problem surely? But that wasn’t surfaced. Should reaching the loop limit throw an exception rather than silently swallowing it?

> Related: Did you see my note about how it was moving the comma up one line at a time? IMHO, that's at best a poorly written rule, _probably_ a bug. Imagine if there's a really long block of comments between the initial location of a leading comma and its final, desired, trailing location above. It's poor design for the linter to require 'n' iterations to move a comma by 'n' lines. Think I should create an issue for that?

Possibly. It might as well skip that it if can. But it’s not the only rule that can result in a rerun surfacing more fixes.

barrywhart:
Yes, reaching the loop limit most likely means there was a problem. I'm a little torn about having it throw an exception:
* Pros: Issues get surfaced and reported.
* Cons: Perhaps we improved the user's SQL. Are we talking about stopping the run, or reporting an exception within that file and still processing other files? Do we still rewrite their SQL file?

I'm generally +1 on your suggestion, but want to consider the details. I'll create an issue for this so we can improve on it.

barrywhart:
See #1386 re: loop limit.

barrywhart:
See #1387 re: moving the comma one line at a time.

PR #1383, opened 2021-09-11
Issue 1310: L003 fix mixes hanging and clean indents

### LINKED ISSUE #1310: `sqlfluff fix` mixes hanging and regular indents
## Steps to Reproduce

Run `sqlfluff fix` on this code:
``` sql
select *
from foo
where a like 'a%'
  and b like 'b%'
  and c = false
```

## Expected Behaviour

Either one of these options:
``` sql
select *
from foo
where a like 'a%'
      and b like 'b%'
      and c = false
```
or
``` sql
select *
from foo
where a like 'a%'
    and b like 'b%'
    and c = false
```

## Observed Behaviour

`sqlfluff fix` produces this output, which looks worse than any of the alternatives:

``` sql
select *
from foo
where a like 'a%'
      and b like 'b%'
    and c = false
```

## Version
SQLFluff 0.6.3
Python 3.9.4

barrywhart:
Fixes #1310

Rule L003 was only allowing hanging indent for a single line. This results in inconsistent indentation. As an experiment, I disabled the single-line limitation. No tests broke, so I guess this is okay?

I added a new test based on the user-reported issue, so if future concerns arise about this rule, we'll have a test case to make sure we can converge towards a better solution and not "flap" back and forth.
...

As mentioned above, hanging indent is now allowed for any number of consecutive lines.  Is this ever a bad thing? Unsure.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L003.py
+++ b/src/sqlfluff/rules/L003.py
@@ -94,7 +94,7 @@ class Rule_L003(BaseRule):
                     # Indent balance is the indent at the start of the first content
                     "indent_balance": this_indent_balance,
                     "hanging_indent": hanger_pos if line_indent_stack else None,
-                    # Clean indent is true if the line *ends* win an indent
+                    # Clean indent is true if the line *ends* with an indent
                     # or has an indent in the initial whitespace.
                     "clean_indent": clean_indent,
                 }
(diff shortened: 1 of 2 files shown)


tunetheweb (approved):
Nice and simple!

Issue #1340, opened 2021-09-01
Return non zero exit code on fix when there are lint violations


mvgijssel:
We use [lefthook](https://github.com/evilmartians/lefthook) in combination with a pre-commit git hook to fix and commit updated dbt models when committing changes to the repository. This works great with the only downside that `poetry run sqlfluff --fix` will return a 0 exit code when it's done fixing even though there are still unfixable lint violations. 

As an enhancement if would be great if sqlfluff could return a non zero exit code on fix when there are still unfixable linting violations.

nolanbconaway:
I got this one unless someone faster wants it!

tunetheweb:
> I got this one unless someone faster wants it!

Actually I started looking at this already and think I have a fix. Would appreciate a PR review though!

PR #1385, opened 2021-09-11
Handle template and unfixable errors when fixing stdin

### LINKED ISSUE #1196: AttributeError: 'NoneType' object has no attribute 'templated_str'
A bug surfaced in slack: I have no idea what's going on here! The SQL is

```sql
--Airflow: ml_cxe_service_metrics -
CREATE TABLE IF NOT EXISTS {{ params.ml_schema }}.daily_service_metrics (
    source_type varchar(10) NOT NULL,
    visit_date date NOT NULL,
    event_name text NOT NULL,
    event_count bigint NOT NULL
)
diststyle even
compound sortkey(visit_date, source_type)
;
```

## Expected Behaviour

No errors!

## Observed Behaviour

linting:

```
➜ echo '--Airflow: ml_cxe_service_metrics -
CREATE TABLE IF NOT EXISTS {{ params.ml_schema }}.daily_service_metrics (
    source_type varchar(10) NOT NULL,
    visit_date date NOT NULL,
    event_name text NOT NULL,
    event_count bigint NOT NULL
)
diststyle even
compound sortkey(visit_date, source_type)
;' | sqlfluff lint - --verbose
==== sqlfluff ====
sqlfluff:                0.6.0 python:                  3.8.7

== [stdin] FAIL
L:   0 | P:   0 |  TMP | Unrecoverable failure in Jinja templating: 'params' is
                       | undefined. Have you configured your variables?
L:   2 | P:  31 |  TMP | Undefined jinja template variable: 'params'
```

fixing:
```
➜ echo '--Airflow: ml_cxe_service_metrics -
CREATE TABLE IF NOT EXISTS {{ params.ml_schema }}.daily_service_metrics (
    source_type varchar(10) NOT NULL,
    visit_date date NOT NULL,
    event_name text NOT NULL,
    event_count bigint NOT NULL
)
diststyle even
compound sortkey(visit_date, source_type)
;' | sqlfluff fix - --dialect postgres
...
AttributeError: 'NoneType' object has no attribute 'templated_str
```

`params` is not defined in my (empty) project, but the fixer should not except out!

## Steps to Reproduce

see above!

## Version

0.6.0

## Configuration

zilch

nolanbconaway:

Handles templater errors and otherwise unfixable errors better when fixing stdin.

fixes #1196 

Instead of excepting like:

```
➜ echo '--Airflow: ml_cxe_service_metrics -
CREATE TABLE IF NOT EXISTS {{ params.ml_schema }}.daily_service_metrics (
    source_type varchar(10) NOT NULL,
    visit_date date NOT NULL,
    event_name text NOT NULL,
    event_count bigint NOT NULL
)
diststyle even
compound sortkey(visit_date, source_type)
;' | sqlfluff fix - --dialect postgres
...
AttributeError: 'NoneType' object has no attribute 'templated_str
```

This PR allows fluff to exit like

```
➜ echo '--Airflow: ml_cxe_service_metrics -
CREATE TABLE IF NOT EXISTS {{ params.ml_schema }}.daily_service_metrics (
    source_type varchar(10) NOT NULL,
    visit_date date NOT NULL,
    event_name text NOT NULL,
    event_count bigint NOT NULL
)
diststyle even
compound sortkey(visit_date, source_type)
;' | sqlfluff fix - --dialect postgres

The stdin input contains template variables that could not be parsed. Aborting fix.
```

I am not feeling 100% good about this approach. It is better than current state but the _best_ outcome would be to return fixed sql if at all possible. I don't know much about the templater so will leave it to others to decide if it is at all possible to fix via stdin with templated variables..

...

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -4,7 +4,6 @@ import sys
 import json
 import logging
 import time
-
 import oyaml as yaml
 
 import click
@@ -477,19 +476,18 @@ def fix(force, paths, processes, bench=False, fixed_suffix="", logger=None, **kw
     if fixing_stdin:
         stdin = sys.stdin.read()
         result = lnt.lint_string_wrapped(stdin, fname="stdin", fix=True)
-        # exit early if there is a templater error.
-        if any((isinstance(i, SQLTemplaterError) for i in result.get_violations())):
-            click.echo(
-                colorize(
-                    "The stdin input contains template variables that could not be "
-                    + "parsed. Aborting fix.",
-                    "red",
+
+        templater_error = result.num_violations(types=SQLTemplaterError) > 0
+        if result.num_violations(types=SQLLintError, fixable=True) > 0:
+            stdout = result.paths[0].files[0].fix_string()[0]
+        else:
+            if verbose and templater_error:
+                click.echo(
+                    "Fix aborted due to template variables that could not be parsed."
                 )
-            )
-            sys.exit(1)
-        stdout = result.paths[0].files[0].fix_string()[0]
+            stdout = stdin
         click.echo(stdout, nl=False)
-        sys.exit()
+        sys.exit(1 if templater_error else 0)
 
     # Lint the paths (not with the fix argument at this stage), outputting as we go.
     click.echo("==== finding fixable violations ====")
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -33,6 +33,7 @@ from sqlfluff.core import (
     Linter,
     FluffConfig,
     SQLLintError,
+    SQLTemplaterError,
     dialect_selector,
     dialect_readout,
     TimingSummary,
@@ -476,6 +477,16 @@ def fix(force, paths, processes, bench=False, fixed_suffix="", logger=None, **kw
     if fixing_stdin:
         stdin = sys.stdin.read()
         result = lnt.lint_string_wrapped(stdin, fname="stdin", fix=True)
+        # exit early if there is a templater error.
+        if any((isinstance(i, SQLTemplaterError) for i in result.get_violations())):
+            click.echo(
+                colorize(
+                    "The stdin input contains template variables that could not be "
+                    + "parsed. Aborting fix.",
+                    "red",
+                )
+            )
+            sys.exit(1)
         stdout = result.paths[0].files[0].fix_string()[0]
         click.echo(stdout, nl=False)
         sys.exit()
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -475,19 +475,23 @@ def fix(force, paths, processes, bench=False, fixed_suffix="", logger=None, **kw
     # handle stdin case. should output formatted sql to stdout and nothing else.
     if fixing_stdin:
         stdin = sys.stdin.read()
-        result = lnt.lint_string_wrapped(stdin, fname="stdin", fix=True)
 
+        result = lnt.lint_string_wrapped(stdin, fname="stdin", fix=True)
         templater_error = result.num_violations(types=SQLTemplaterError) > 0
+        unfixable_error = result.num_violations(types=SQLLintError, fixable=False) > 0
+
         if result.num_violations(types=SQLLintError, fixable=True) > 0:
             stdout = result.paths[0].files[0].fix_string()[0]
         else:
-            if verbose and templater_error:
-                click.echo(
-                    "Fix aborted due to template variables that could not be parsed."
-                )
+            if verbose:
+                if templater_error:
+                    click.echo("Fix aborted due to unparseable template variables.")
+                if unfixable_error:
+                    click.echo("Unfixable violations detected.")
             stdout = stdin
+
         click.echo(stdout, nl=False)
-        sys.exit(1 if templater_error else 0)
+        sys.exit(1 if templater_error or unfixable_error else 0)
 
     # Lint the paths (not with the fix argument at this stage), outputting as we go.
     click.echo("==== finding fixable violations ====")
--- a/test/cli/commands_test.py
+++ b/test/cli/commands_test.py
@@ -409,12 +409,13 @@ def test__cli__command_fix_stdin_safety():
 @pytest.mark.parametrize(
     "sql,exit_code",
     [
-        ("create TABLE {{ params.dsfsdfds }}.t (a int)", 1),
-        ("create TABLE a.t (a int)", 0),
-        ("create table a.t (a int)", 0),
+        ("create TABLE {{ params.dsfsdfds }}.t (a int)", 1),  # template error
+        ("create TABLE a.t (a int)", 0),  # fixable error
+        ("create table a.t (a int)", 0),  # perfection
+        ("select col from a join b using (c)", 1),  # unfixable error (using)
     ],
 )
-def test__cli__command_fix_stdin_template_error_exit_code(sql, exit_code):
+def test__cli__command_fix_stdin_error_exit_code(sql, exit_code):
     """Check that the CLI fails nicely if fixing a templated stdin."""
     if exit_code == 0:
         invoke_assert_code(args=[fix, ("-",)], cli_input=sql)
--- a/test/cli/commands_test.py
+++ b/test/cli/commands_test.py
@@ -406,6 +406,16 @@ def test__cli__command_fix_stdin_safety():
     assert result.output.strip() == perfect_sql
 
 
+def test__cli__command_fix_stdin_template_error():
+    """Check that the CLI fails nicely if fixing a templated stdin."""
+    sql = "create table {{ params.something }}.t (a int)"
+
+    with pytest.raises(SystemExit) as exc_info:
+        invoke_assert_code(args=[fix, ("-",)], cli_input=sql)
+
+    assert exc_info.value.args[0] == 1
+
+
 @pytest.mark.parametrize(
     "rule,fname,prompt,exit_code",
     [
--- a/test/cli/commands_test.py
+++ b/test/cli/commands_test.py
@@ -406,14 +406,23 @@ def test__cli__command_fix_stdin_safety():
     assert result.output.strip() == perfect_sql
 
 
-def test__cli__command_fix_stdin_template_error():
+@pytest.mark.parametrize(
+    "sql,exit_code",
+    [
+        ("create TABLE {{ params.dsfsdfds }}.t (a int)", 1),
+        ("create TABLE a.t (a int)", 0),
+        ("create table a.t (a int)", 0),
+    ],
+)
+def test__cli__command_fix_stdin_template_error_exit_code(sql, exit_code):
     """Check that the CLI fails nicely if fixing a templated stdin."""
-    sql = "create table {{ params.something }}.t (a int)"
-
-    with pytest.raises(SystemExit) as exc_info:
+    if exit_code == 0:
         invoke_assert_code(args=[fix, ("-",)], cli_input=sql)
+    else:
+        with pytest.raises(SystemExit) as exc_info:
+            invoke_assert_code(args=[fix, ("-",)], cli_input=sql)
 
-    assert exc_info.value.args[0] == 1
+        assert exc_info.value.args[0] == exit_code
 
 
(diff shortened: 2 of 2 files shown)


barrywhart:
Does this change prevent using templated stdin by providing a .sqlfluff file with those variables?

nolanbconaway:
Nope! All it does is make the error message more informative!

barrywhart:
>decide if it is at all possible to fix via stdin with templated variables..

I'll try and take a closer look at this soon.

nolanbconaway:
> > decide if it is at all possible to fix via stdin with templated variables..
> 
> I'll try and take a closer look at this soon.

For context; another approach I tried was to write the stdin to a tempfile and fix _that_, so there is a file. But the error did not vanish and I found myself too far out my zone

barrywhart:
This fix seems good to me. It may be possible to do more, but linting from stdin is using a different Linter function, so the code path is completely different.

One thing I thought about, if we want, is to write the contents of stdin to a temp file, fix that, then print the result to stdout. That way, both flavors of "fix" are using the same code path. Think that's worthwhile, or is this good enough? I can go either way.

barrywhart (approved):

nolanbconaway:
> One thing I thought about, if we want, is to write the contents of stdin to a temp file, fix that, then print the result to stdout. That way, both flavors of "fix" are using the same code path. Think that's worthwhile, or is this good enough? I can go either way.

LOL Barry! https://github.com/sqlfluff/sqlfluff/pull/1385#issuecomment-917483389

> For context; another approach I tried was to write the stdin to a tempfile and fix that, so there is a file. But the error did not vanish and I found myself too far out my zone

I'll give this another run but i was getting the same error after fixing upon a tempfile

src/sqlfluff/cli/commands.py
```diff
@@ -476,6 +477,16 @@ def fix(force, paths, processes, bench=False, fixed_suffix="", logger=None, **kw
...
+        if any((isinstance(i, SQLTemplaterError) for i in result.get_violations())):
+            click.echo(
+                colorize(
+                    "The stdin input contains template variables that could not be "
+                    + "parsed. Aborting fix.",
+                    "red",
+                )
+            )
+            sys.exit(1)
         stdout = result.paths[0].files[0].fix_string()[0]
```
tunetheweb, line 490: When you test it using files it doesn't "fix" the file either. It just doesn't bomb out like it does with the CLI. So switching to linting a temp file won't help any more than the current fix.

Digging into why this is, it is because it does this for the file option:

```py
    # NB: We filter to linting violations here, because they're
    # the only ones which can be potentially fixed.
    if result.num_violations(types=SQLLintError, fixable=True) > 0:
        click.echo("==== fixing violations ====")
        ...
```

So it only attempts to fix violations if there are any fixable violations! Whereas the stdin method attempts to fix regardless.

Therefore I think you should implement this same thing instead of checking for just `SQLTemplaterError` errors:

```suggestion
    # handle stdin case. should output formatted sql to stdout and nothing else.
    if fixing_stdin:
        stdin = sys.stdin.read()
        result = lnt.lint_string_wrapped(stdin, fname="stdin", fix=True)
        if result.num_violations(types=SQLLintError, fixable=True) > 0:
            stdout = result.paths[0].files[0].fix_string()[0]
        else:
            if verbose:
                click.echo("No fixable violations found.")
            stdout = stdin
```

Note that I'm suggesting returning `stdin` as `stdout`, which is what happens in non-templated SQL that doesn't need fixing, but that only happens because `result.paths[0].files[0].fix_string()[0]` is set to the same as input, whereas that doesn't happen on templating failures, but net effect is the same. So might as well be explicit about using stdin.

tunetheweb (changes requested):
Dug into this more and think that there is a better fix.

I also think we should add a test case to this PR to prevent this regressing in future.

src/sqlfluff/cli/commands.py
barrywhart: That sounds great to me!

nolanbconaway: Waiiiiit, does this error only raise when there are no fixable errors? I ask because:

```
➜ echo 'create TABLE {{ params.dsfsdfds }}.t (a int)' | sqlfluff fix - --verbose
The stdin input contains template variables that could not be parsed. Aborting fix.

➜ echo 'create TABLE a.t (a int)' | sqlfluff fix - --verbose
create table a.t (a int)
```

Both cases have a fixable error and in one we ALSO have a template error which could (?) be handled and a fix returned. So is this now a two-case issue?

1. if no fixable errors, return stdin even if there was an exception on the way (duh)
2. Use a tempfile approach to support fixing what can be fixed using a single code path

am i understanding the suggestion correctly?

nolanbconaway: FYI- with this implementation we still raised a templated file error

```py
    if fixing_stdin:
        stdin = sys.stdin.read()
        with tempfile.NamedTemporaryFile(mode="w+", suffix=".sql") as f:
            f.write(stdin)
            f.flush()
            result = lnt.lint_path(f.name, fix=True)
            stdout = result.files[0].fix_string()[0]
        click.echo(stdout, nl=False)
        sys.exit()
```


```
squidward 🐙 in sqlfluff on  nolan/fix-stdout-error [!] via sqlfluff via 🐍 sqlfluff 
➜ echo 'create TABLE a.t (a int)' | sqlfluff fix - --verbose                    
create table a.t (a int)

squidward 🐙 in sqlfluff on  nolan/fix-stdout-error [!] via sqlfluff via 🐍 sqlfluff 
➜ echo 'create TABLE {{ params.dsfsdfds }}.t (a int)' | sqlfluff fix - --verbose
  File "/home/squidward/.pyenv/versions/sqlfluff/bin/sqlfluff", line 33, in <module>
    sys.exit(load_entry_point('sqlfluff', 'console_scripts', 'sqlfluff')())
  File "/home/squidward/.pyenv/versions/sqlfluff/lib/python3.8/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/home/squidward/.pyenv/versions/sqlfluff/lib/python3.8/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/home/squidward/.pyenv/versions/sqlfluff/lib/python3.8/site-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/squidward/.pyenv/versions/sqlfluff/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/squidward/.pyenv/versions/sqlfluff/lib/python3.8/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/home/squidward/Documents/repos/sqlfluff/src/sqlfluff/cli/commands.py", line 483, in fix
    stdout = result.files[0].fix_string()[0]
  File "/home/squidward/Documents/repos/sqlfluff/src/sqlfluff/core/linter/linted_file.py", line 242, in fix_string
    linter_logger.debug("Original Tree: %r", self.templated_file.templated_str)
AttributeError: 'NoneType' object has no attribute 'templated_str'

tunetheweb: > Waiiiiit, does this error only raise when there are no fixable errors? I ask because:

No. It only raises when there is a templating error. As in that case the `templated_file` is not set and that is needed by `fix_string()`.

For non-templates, no fixable errors that variable is set - but to the original SQL anyway. BUT for case the fixed is not outputted in file mode (try it with a file and the SUFFIX option and you’ll see it’s not written).

> Both cases have a fixable error and in one we ALSO have a template error

That is not possible. Template errors result in unparsable SQL and hence no fixable errors. Which may seem a little harsh but is actually right IMHO. If you don’t understand the templates pieces then you don’t understand the SQL. For simple cases you could argue to just ignore the templates but think that’s a risky option. To be honest it’s the same with unparseable sections but I think we do allow them to be fixed (but not sure we should).

tunetheweb: > FYI- with this implementation we still raised a templated file error

Are you sure? Not at computer now but can look later but was definitely working earlier. Maybe I copied and pasted the fix in, incorrectly.

nolanbconaway: Ohhhh i see- the parser will be unable to find any fixable violations whenever the templater fails. That makes sense and nooooowwww i understand the suggestion. In that case we have two choices:

1. Spit the stdin back out or not. It feels good to do so from a unix/posix sort of perspective
2. exit zero or nonzero, optionally logging an informative message

I lean towards exiting nonzero AND spitting the stdin out. users should be _somehow_ notified that something is not right

tunetheweb: Yup works for me. But think that should be done for ANY non-fixable errors not just template errors?

But that might cause other breakages?

BTW if we ever did start supporting fixing SQL with template errors (which as I say I’m not sure we should, but anyway) then my suggestion would seem less let support that (as fixable errors would now be > 0).

nolanbconaway: Yeah aboslutely; i think exit nonzero if any templater error but not if the sql is perfect. will add tests to make sure all is good.

barrywhart: I don't know the details, but some of the lint rules _will_ run on unparseable code. Search the rule code for `_works_on_unparsable`. Some rules specify `False`, but not all.

nolanbconaway:
Ok so in the latest:

- if no fixable violations, spit the stdin back out unchanged
- if any templater errors, return nonzero exit and print message if verbose.

thx for the help everyone!

src/sqlfluff/cli/commands.py
tunetheweb: Linting is fine. Fixing is more questionable.

But maybe OK for simple fixes (e.g. capitalisation, some spacing).

Anyway going off track.

BTW I do wonder if `fix_string()` should be made more robust rather than failing so ungracefully when `templated_file` doesn’t exist? Yes we probably shouldn’t waste our time fixing SQL with no fixable errors but still…

tunetheweb:
What if there are non-template unfixable errors? Should we exit non-zero there rather than just on template errors?

Also can we add a test case or two?

src/sqlfluff/cli/commands.py
nolanbconaway: My take is that the use of fix_string here is something of a holdover that we never re-wrote to be more robust. It goes back to like my first fluff pr i think. @barrywhart is right and we should seek to unify the code path between stdin and file fixing; probably a tempfile solution would work.

nolanbconaway (commented):
Am i crazy or did i write a test case or two? am i crazy? https://github.com/sqlfluff/sqlfluff/pull/1385/files#diff-999a57d85b8049e8718a87b7789ce83e7a575f2d702facc42990cb374e5ec130R409-R427

tunetheweb:
> Am i crazy or did i write a test case or two? am i crazy? https://github.com/sqlfluff/sqlfluff/pull/1385/files#diff-999a57d85b8049e8718a87b7789ce83e7a575f2d702facc42990cb374e5ec130R409-R427

Sorry missed them! Used to looking for SQL files and so expected an extra file since your original PR.

src/sqlfluff/cli/commands.py
tunetheweb: Well I think they are unified now. We only call fix if there are fixable errors. That’s what the difference was and that is now fixed.

nolanbconaway:
I could go either way on the unfixable lint errors! i suppose i lean towards exiting zero because that's been fluffs behavior this whole time. but could be easily convinced otherwise.

WittierDinosaur:
#1340 references exiting non-zero for unfixable lint errors. Seems there's appetite for it (for fix)

nolanbconaway:
lol aight u convinced me

barrywhart:
Whee, we are aligned!

nolanbconaway:
cool! ready for :eyes: !

tunetheweb (approved):
LGTM!

The only thing I'd note is this this introduce a difference between the `stdin` and the file version. Should the file version similarly exit non-zero on unfixable and template errors similarly? Or is that out of scope of this PR and should be a separate PR?

nolanbconaway:
My take is a separate PR- this one has already creeped a couple levels deeper in scope!

PR #1391, opened 2021-09-13
Improve rules L003 and L019 by processing multi-line fixes in one pass.

### LINKED ISSUE #1387: L019: When moving a comma from leading to trailing, skip over blank and/or "comment only" lines
In working on #1344, we discovered a quirk (bug?) in L019: when it has to move a comma across several physical lines (i.e. if the lines in between are blank or "comment only") to get it to the correct location, _it moves the comma just one line at a time_. So the linter has to run multiple iterations, one per line the comma is to be moved. This is slow, and it also means _the linter may hit the loop limit_ before finishing its work. This is not good, and arguably a bug.

Let's make the rule smarter, i.e. ignore those blank/comment-only lines and **move the comma to the right location in one step**.

For context, see the original explanation of the issue (with an example) [here](https://github.com/sqlfluff/sqlfluff/issues/1344#issuecomment-917409372).

To reproduce, run `sqlfluff fix` on this query:
```
SELECT
COUNT(1) AS campaign_count,
state_user_v_peer_open
,business_type

-- The following is the slope of the regression line. Note that CORR (which is the Pearson's correlation
--  coefficient is symmetric in its arguments, but since STDDEV_POP(open_rate_su) appears in the
--  numerator this is the slope of the regression line considering STDDEV_POP(open_rate_su) to be
--  the "y variable" and uses_small_subject_line to be the "x variable" in terms of the regression line.
,SAFE_DIVIDE(SAFE_MULTIPLY(CORR(open_rate_su, uses_small_subject_line), STDDEV_POP(open_rate_su)), STDDEV_POP(uses_small_subject_line))

FROM
global_actions_states
```

Expected outcome:
```
SELECT
    state_user_v_peer_open,
    business_type,
        COUNT(1) AS campaign_count,

    -- The following is the slope of the regression line. Note that CORR (which is the Pearson's correlation
    --  coefficient is symmetric in its arguments, but since STDDEV_POP(open_rate_su) appears in the
    --  numerator this is the slope of the regression line considering STDDEV_POP(open_rate_su) to be
    --  the "y variable" and uses_small_subject_line to be the "x variable" in terms of the regression line.
    SAFE_DIVIDE(
        SAFE_MULTIPLY(
            CORR(open_rate_su, uses_small_subject_line), STDDEV_POP(open_rate_su)
        ),
        STDDEV_POP(uses_small_subject_line)
    )

FROM
    global_actions_states
```

Actual outcome:
```
SELECT
    state_user_v_peer_open,
    business_type,
        COUNT(1) AS campaign_count

-- The following is the slope of the regression line. Note that CORR (which is the Pearson's correlation
,--  coefficient is symmetric in its arguments, but since STDDEV_POP(open_rate_su) appears in the
    --  numerator this is the slope of the regression line considering STDDEV_POP(open_rate_su) to be
    --  the "y variable" and uses_small_subject_line to be the "x variable" in terms of the regression line.
    SAFE_DIVIDE(
        SAFE_MULTIPLY(
            CORR(open_rate_su, uses_small_subject_line), STDDEV_POP(open_rate_su)
        ),
        STDDEV_POP(uses_small_subject_line)
    )

FROM
    global_actions_states
```

`.sqlfluff`:
```
[sqlfluff]
runaway_limit = 3
```

(I artificially lowered the `runaway_limit` in order to reproduce the issue with a smaller example file.)

barrywhart:
Fixes #1387 

This PR fixes a couple of issues:
- Primary fix: Previously, when run in "trailing comma" mode, L019 was only moving commas one line at a time, even if there were multiple comment or blank lines between the initial location and the desired final location. This caused the linter to require additional loops (bad) and could result in hitting the linter loop limit while the comma was in mid-journey.
- Secondary fix: I found that when I fixed the above issue, it surfaced a bug in L003 that had previously been hidden in some cases. Amusingly, when a comma would move through a comment block, it would sometimes trigger useful indentation changes along the way. Now that the comma jumps immediately to the final destination, this was no longer occurring. So I updated L003 to explicitly align a comment **block** preceding a code line with the code line, rather than (old behavior) just aligning the **immediately preceding** comment.

These changes "broke" a few existing tests, but in every case, the new behavior seemed correct/better to me.

The update to L003  changes existing formatting behavior. I believe it's an improvement, but if we have concerns, we can revert to the previous behavior. Note that the previous behavior was inconsistent -- the indentation of comments depended on whether a comma had taken a joyride through the comment block. 🤣 So I recommend we take these changes, or at least refine them farther, _not_ go back.
...

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/linter/runner.py
+++ b/src/sqlfluff/core/linter/runner.py
@@ -7,6 +7,7 @@ Implements various runner types for SQLFluff:
   - Multithread (used only by automated tests)
 """
 from abc import ABC
+import bdb
 import functools
 import logging
 import multiprocessing.dummy
@@ -92,7 +93,7 @@ class SequentialRunner(BaseRunner):
         for fname, partial in self.iter_partials(fnames, fix=fix):
             try:
                 yield partial()
-            except KeyboardInterrupt:
+            except (bdb.BdbQuit, KeyboardInterrupt):
                 raise
             except Exception as e:
                 self._handle_lint_path_exception(fname, e)
--- a/src/sqlfluff/core/linter/runner.py
+++ b/src/sqlfluff/core/linter/runner.py
@@ -92,6 +92,8 @@ class SequentialRunner(BaseRunner):
         for fname, partial in self.iter_partials(fnames, fix=fix):
             try:
                 yield partial()
+            except KeyboardInterrupt:
+                raise
             except Exception as e:
                 self._handle_lint_path_exception(fname, e)
 
--- a/src/sqlfluff/core/linter/runner.py
+++ b/src/sqlfluff/core/linter/runner.py
@@ -93,7 +93,7 @@ class SequentialRunner(BaseRunner):
         for fname, partial in self.iter_partials(fnames, fix=fix):
             try:
                 yield partial()
-            except (bdb.BdbQuit, KeyboardInterrupt):
+            except (bdb.BdbQuit, KeyboardInterrupt):  #pragma: no cover
                 raise
             except Exception as e:
                 self._handle_lint_path_exception(fname, e)
--- a/src/sqlfluff/core/linter/runner.py
+++ b/src/sqlfluff/core/linter/runner.py
@@ -93,7 +93,7 @@ class SequentialRunner(BaseRunner):
         for fname, partial in self.iter_partials(fnames, fix=fix):
             try:
                 yield partial()
-            except (bdb.BdbQuit, KeyboardInterrupt):  #pragma: no cover
+            except (bdb.BdbQuit, KeyboardInterrupt):  # pragma: no cover
                 raise
             except Exception as e:
                 self._handle_lint_path_exception(fname, e)
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -14,6 +14,7 @@ should flag on the segment FOLLOWING, the place that the desired element is
 missing.
 """
 
+import bdb
 import copy
 import logging
 import pathlib
@@ -309,7 +310,7 @@ class BaseRule:
                 path=pathlib.Path(fname) if fname else None,
                 templated_file=templated_file,
             )
-        except KeyboardInterrupt:
+        except (bdb.BdbQuit, KeyboardInterrupt):
             raise
         # Any exception at this point would halt the linter and
         # cause the user to get no results
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -310,7 +310,7 @@ class BaseRule:
                 path=pathlib.Path(fname) if fname else None,
                 templated_file=templated_file,
             )
-        except (bdb.BdbQuit, KeyboardInterrupt):  #pragma: no cover
+        except (bdb.BdbQuit, KeyboardInterrupt):  # pragma: no cover
             raise
         # Any exception at this point would halt the linter and
         # cause the user to get no results
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -310,7 +310,7 @@ class BaseRule:
                 path=pathlib.Path(fname) if fname else None,
                 templated_file=templated_file,
             )
-        except (bdb.BdbQuit, KeyboardInterrupt):
+        except (bdb.BdbQuit, KeyboardInterrupt):  #pragma: no cover
             raise
         # Any exception at this point would halt the linter and
         # cause the user to get no results
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -309,6 +309,8 @@ class BaseRule:
                 path=pathlib.Path(fname) if fname else None,
                 templated_file=templated_file,
             )
+        except KeyboardInterrupt:
+            raise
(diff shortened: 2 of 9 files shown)


src/sqlfluff/core/linter/runner.py
```diff
@@ -92,6 +93,8 @@ def run(self, fnames, fix):
         for fname, partial in self.iter_partials(fnames, fix=fix):
             try:
                 yield partial()
+            except (bdb.BdbQuit, KeyboardInterrupt):
```
barrywhart, line 96: Unrelated to the overall issue, but this makes SQLFluff better behaved: If a user or developer hits ^C at runtime or quits out of the debugger, don't catch the exception.

src/sqlfluff/core/rules/base.py
```diff
@@ -309,6 +310,8 @@ def crawl(
                 path=pathlib.Path(fname) if fname else None,
                 templated_file=templated_file,
             )
+        except (bdb.BdbQuit, KeyboardInterrupt):
```
barrywhart, line 313: Unrelated to the overall issue, but this makes SQLFluff better behaved: If a user or developer hits ^C at runtime or quits out of the debugger, don't catch the exception.

src/sqlfluff/rules/L003.py
```diff
@@ -578,35 +578,41 @@ def _eval(self, segment, raw_stack, memory, **kwargs):
...
-                            break
-                    # Make fixes.
-                    fixes = self._coerce_indent_to(
-                        desired_indent="".join(
-                            elem.raw for elem in this_line["indent_buffer"]
-                        ),
-                        current_indent_buffer=prev_line["indent_buffer"],
-                        current_anchor=anchor,
-                    )
+            # Given that this line is ok, consider if the preceding lines are
```
barrywhart, line 581: This change is smaller than it looks due to indentation changes. It just changes a check for "previous comment line" to "all contiguous preceding comment lines"

test/fixtures/linter/autofix/ansi/008_looping_rules_l003_l016_l019/after.sql
```diff
@@ -2,7 +2,7 @@ SELECT
     COUNT(1) AS campaign_count,
     state_user_v_peer_open,
     business_type,
-    
+
```
barrywhart, line 5: Incidental change to output. The new behavior seems better (completely blank line rather than line with 4 spaces).

test/fixtures/rules/std_rule_cases/L019.yml
```diff
@@ -24,7 +24,7 @@ leading_comma_violation_with_inline_comment:
   fix_str: |
     SELECT
         a,
-        b ,-- inline comment
+        b, -- inline comment
```
barrywhart, line 27: This test "broke" after changing L019. The new behavior seems better.

test/fixtures/linter/autofix/bigquery/004_templating/after.sql
```diff
@@ -1,89 +1,89 @@
-
```
barrywhart, line 1: I actually only changed lines 27 and 28. The whole file looks different here. Maybe it was in Windows format previously?

test/fixtures/rules/std_rule_cases/L019.yml
```diff
@@ -198,3 +198,79 @@ leading_comma_fixing_flows_around_comments:
...
+    global_actions_states
+  configs:
+    core:
+      # Set runaway_limit=1 to verify the fix only requires one pass. In an
+      # earlier version, the comma before "SAFE_DIVIDE()" was being moved one
+      # line per pass. Too lazy!
+      runaway_limit: 1
+
+
+leading_comma_move_past_several_comment_lines:
```
barrywhart, line 239: I added this new test for the "leading comma" case to complement the new "trailing comma" case above. For this test, no code changes were required -- I guess L019 was already handling this case efficiently. 👍🏽

src/sqlfluff/rules/L003.py
tunetheweb: FYI, GitHub has an option to remove whitespace when reviewing files, which can make changes like this easier to review:

image

src/sqlfluff/core/linter/runner.py
tunetheweb:
```suggestion
            except (bdb.BdbQuit, KeyboardInterrupt):  #pragma: no cover
```

src/sqlfluff/core/rules/base.py
tunetheweb:
```suggestion
        except (bdb.BdbQuit, KeyboardInterrupt):  #pragma: no cover
```

tunetheweb (approved):
LGTM!

Just had to fix some coverage issues.

tunetheweb:
Going to suggest a better title for the release notes. What do you think about:

"Improve rules L003 and L019 by processing multi-line fixes in one pass."

Or soemthing like that?

src/sqlfluff/rules/L003.py
barrywhart: 👍

barrywhart:
I'll change the PR title. You're right -- when I'm creating a title for the PR, I'm often unconsciously thinking of the reviewers as the audience, rather than the eventual users. 👍

Issue #1392, opened 2021-09-13
[Doc] Rule ref without highlighting


zhongjiajie:
## URL or Command Line Option with the issue being reported

Our [rules ref](https://docs.sqlfluff.com/en/stable/rules.html) code seem without highlighting, It seem we could fix by adding `:force:` option to our docstrings according to https://github.com/sphinx-doc/sphinx/pull/6416.

![image](https://user-images.githubusercontent.com/15820530/133038811-36fab3ea-c211-4c0c-b099-e4229fe8850d.png)

And when we try to build our doc, we would get somme error as below

```log
sqlfluff/rules/L001.py:docstring of sqlfluff.rules.L001.Rule_L001:8: WARNING: Could not lex literal_block as "sql". Highlighting skipped.
sqlfluff/rules/L002.py:docstring of sqlfluff.rules.L002.Rule_L002:16: WARNING: Could not lex literal_block as "sql". Highlighting skipped.
sqlfluff/rules/L002.py:docstring of sqlfluff.rules.L002.Rule_L002:25: WARNING: Could not lex literal_block as "sql". Highlighting skipped.
sqlfluff/rules/L003.py:docstring of sqlfluff.rules.L003.Rule_L003:21: WARNING: Could not lex literal_block as "sql". Highlighting skipped.
sqlfluff/rules/L003.py:docstring of sqlfluff.rules.L003.Rule_L003:32: WARNING: Could not lex literal_block as "sql". Highlighting skipped.
sqlfluff/rules/L004.py:docstring of sqlfluff.rules.L004.Rule_L004:19: WARNING: Could not lex literal_block as "sql". Highlighting skipped.
sqlfluff/rules/L004.py:docstring of sqlfluff.rules.L004.Rule_L004:29: WARNING: Could not lex literal_block as "sql". Highlighting skipped.
sqlfluff/rules/L005.py:docstring of sqlfluff.rules.L005.Rule_L005:12: WARNING: Could not lex literal_block as "sql". Highlighting skipped.
sqlfluff/rules/L008.py:docstring of sqlfluff.rules.L008.Rule_L008:19: WARNING: Could not lex literal_block as "sql". Highlighting skipped.
sqlfluff/rules/L009.py:docstring of sqlfluff.rules.L009.Rule_L009:8: WARNING: Could not lex literal_block as "sql". Highlighting skipped.
[...]

PR #1393, opened 2021-09-13
Fix documentation SQL highlight error

zhongjiajie:
closes: https://github.com/sqlfluff/sqlfluff/issues/1392

Correct sql code highlighting by add `force` option to docstring,  when add force option,  sql code highlight work, but error character show with red rectangle as below

### before

image

### after

image


tunetheweb:
Who knew that Sphinx had a SQL Parser built int!

I’m wondering should we add force to all the SQL blocks and not just those ones that need it? My worry is someone copies another rule, adjusts it to contain invalid SQL for the new rule they are basing it on, and doesn’t realise the need to add force.

I’d there any harm in adding force to them all?

zhongjiajie:
I would try to find out if that a way to set `:force:` options globally.

but if it could not be set, maybe we should add test for our doc build when run `make html`, when doc build with error or warning make this test fail and github checklist would block maintainers merge to `main` branch

docs/source/conf.py
```diff
@@ -108,7 +108,15 @@ def ultimate_replace(app, docname, source):
...
-ultimate_replacements = {"|release|": release}
+SQLFLUFF_RELEASE_FLAG = "|release|"
+SQLFLUFF_CODE_SPACE_NUM = 4
+SQLFLUFF_SQL_CODE_FLAG = f"{' ' * SQLFLUFF_CODE_SPACE_NUM}.. code-block:: sql"
+SQLFLUFF_SQL_CODE_FORCE_FLAG = f"{SQLFLUFF_SQL_CODE_FLAG}\n{' ' * (SQLFLUFF_CODE_SPACE_NUM + 3)}:force:"
+
+ultimate_replacements = {
+    SQLFLUFF_RELEASE_FLAG: release,
+    SQLFLUFF_SQL_CODE_FLAG: SQLFLUFF_SQL_CODE_FORCE_FLAG
+}
```
zhongjiajie, line 119: Cann't found any config for global `:force:` option, but I found out we could use function `ultimate_replace` to make it

zhongjiajie: cc: @tunetheweb

tunetheweb: I can't get this to work.

I'm doing the following:
```
cd docs
pip install -r requirements.txt
make html
python -m http.server --directory build/html
```

But none of the ultimate replacements are replace and I see alerts in the `make html`.

Am I doing something wrong?

BTW a README for this docs directory would be very useful to tell people how to build and test these sorts of changes.

zhongjiajie: Oh…… my bad, I run `python setup.py install` with commit https://github.com/sqlfluff/sqlfluff/pull/1393/commits/e94c179bc9923e28511229ec386aa6489b1a41c3, so that it would render with `:force:` option

tunetheweb: I'm confused. Then I should run the following?:

```
python setup.py install
cd docs
pip install -r requirements.txt
make html
python -m http.server --directory build/html
```

That still doesn't work for me?

zhongjiajie: Yeah, It's should not work event in my laptop . It's work in my env by accident, cause my docstring with `:force:` option, and just commit https://github.com/sqlfluff/sqlfluff/commit/e94c179bc9923e28511229ec386aa6489b1a41c3 look like.

BTW, I test the doc by

```sh
pip install sqlfluff
python setup.py install
cd docs
pip install -r requirements.txt
make html
```

and then just double click the `index.html` file

zhongjiajie:
Now I try the second solution in https://github.com/sqlfluff/sqlfluff/pull/1393#issuecomment-918726221, I add new ci job `docbuild` which test if our doc could build without any warning.
* If it with warning, ci job would failed, we could see example in [error build][1]
* On the other hand, CI job pass until our doc build without error,

[1]: https://github.com/sqlfluff/sqlfluff/pull/1393/checks?check_run_id=3600218555

tunetheweb:
LGTM. Maybe should remove the `force` now from the rules that don't need it now you've a test for it? Or should we just leave it in? WDYT?

It would be nice to try to figure out why the `ultimate_replacements` don't work. Seems like useful functionality. But that's separate to this PR.

tunetheweb:
> LGTM. Maybe should remove the `force` now from the rules that don't need it now you've a test for it? Or should we just leave it in? WDYT?

Ah wait In see you already did that. OK looks good.

docs/Makefile
```diff
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
-SPHINXOPTS    ?=
+SPHINXOPTS    ?= -W --keep-going
```
tunetheweb, line 6: Think we should add a comment this:

```suggestion
# We want SPHINX to warn on error (-W) for example if we don't force SQL
# formatting, but we also want to continue "--keep-going" and check the
# other docs before exiting non-zero
SPHINXOPTS    ?= -W --keep-going
```

tox.ini
```diff
@@ -10,6 +10,7 @@ deps =
     # -U keeps all requirements up-to-date
     -Urrequirements.txt
     -Urrequirements_dev.txt
+    -Urdocs/requirements.txt
```
tunetheweb, line 13: Let's move this further down as only needed for some tests:

```suggestion
```

tox.ini
```diff
@@ -83,6 +84,9 @@ commands = flake8
 [testenv:doclinting]
 commands = doc8 docs/source --file-encoding utf8
 
+[testenv:docbuild]
+commands = make -C docs html
```
tunetheweb, line 88:
```suggestion
[testenv:docbuild]
deps =
    # -U keeps all requirements up-to-date
    -Urrequirements.txt
    -Urrequirements_dev.txt
    -Urdocs/requirements.txt
commands = make -C docs html

```

tunetheweb (changes requested):
Some small suggestions.

tunetheweb (approved):
Committing minor changes

zhongjiajie:
> It would be nice to try to figure out why the `ultimate_replacements` don't work. Seems like useful functionality. But that's separate to this PR.

I inspect and found out it only work for code, but we try to make it work with code syntax `.. code-block:: sql`, so it did not work in our case

zhongjiajie:
Thanks for reviewing and merging.

PR #1396, opened 2021-09-13
Fix extra underscores in case rules (L010 and L014)

### LINKED ISSUE #1394: L014 Inserting underscores after numbers in unquoted identifiers
## Expected Behaviour

Only casing of unquoted identifiers is affected by L014, with no other modifications, eg:

Running `sqlfluff fix --rules L014 test_file.sql` on sql script:

```
select a from b;

insert into A1B2C3
select A from B;
```

should produce sql:

```
select a from b;

insert into a1b2c3
select a from b;
```

## Observed Behaviour

Underscores are inserted after numbers in unquoted identifiers when a casing change is applied. EG.

`sqlfluff fix --rules L014 test_file.sql` 

on

```
select a from b;

insert into A1B2C3
select A from B;
```

produces

```
select a from b;

insert into a1_b2_c3
select a from b;
```

## Steps to Reproduce

1. Create a `test_file.sql` containing the sql in the example above.
2. Run `sqlfluff fix --rules L014 test_file.sql`

## Version

```
sqlfluff, version 0.6.4
Python 3.9.6
```

## Configuration
```
[tool.sqlfluff.core]
templater = "jinja"
sql_file_exts = ".sql,.sql.j2,.dml,.ddl"

[tool.sqlfluff.indentation]
indented_joins = false
indented_using_on = true

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 88
indent_unit = space
comma_style = leading

```

tunetheweb:
Fixes #1394 

Currently the rules were setting Pascal capitalisation policy to be an alternative to underscores, and so adding and removing underscores when fixing case issues.

`ThisIsAnExample` (pascal case) -> `this_is_an_example` (lowercase)
`this_is_an_example` (lowercase) -> `ThisIsAnExample` (pascal case)

This is not my understanding of Pascal Case should be converted, and is more like Snake Case.

With this PR, we now leave underscores alone.

The rules were also lowercasing Capital letters in the middle or words which is, again, not my understanding of PascalCase (as it's name suggests!)

### Pull Request checklist

  - Other.


tunetheweb:
See also this comment where some of these rules were decided: https://github.com/sqlfluff/sqlfluff/issues/820#issuecomment-918226281

tunetheweb:
FYI: @panasenco

WittierDinosaur:
I agree with most of your comments, barring one. https://github.com/sqlfluff/sqlfluff/issues/820#issuecomment-918238258

tunetheweb:
> I agree with most of your comments, barring one. [#820 (comment)](https://github.com/sqlfluff/sqlfluff/issues/820#issuecomment-918238258)

We've discussed on that thread and explained my reasoning, and think we've agreed now on what's been implemented here. So all good to review.

But thanks for looking and challenging! Always good to hear opinions and thrash things out to make sure what's being proposed is correct.

barrywhart (approved):
Looks reasonable to me, but I'm not the most qualified person for the casing stuff. Think we should tag the original author of this work? IIRC, it was contributed about 4 months ago, so they may still be available.

tunetheweb:
> Looks reasonable to me, but I'm not the most qualified person for the casing stuff. Think we should tag the original author of this work? IIRC, it was contributed about 4 months ago, so they may still be available.

See the comment two above this 😁

There was a robust conversation with @WittierDinosaur on this in that other thread and I eventually managed to persuade him round to what we have here. Will give it a day or two to see if any more comments and, if not, merge.

barrywhart:
@tunetheweb: Ah yes, I see some of the earlier discussion now. I'm 👍 for not adding/removing underscores due to the danger of breaking previously working code. I'll throw out there, one option that I don't think y'all considered would be to _warn_ about non-complying code but not provide automatic **fixes** if it would involve adding/removing underscores or something dangerous like that. No strong feelings about whether to do this. Possibly a future option depending on other reviewers' thoughts and/or the eventual user feedback.

Issue #1155, opened 2021-06-13
Postgres: TIMESTAMP WITH TIME ZONE not parsed 


dflss:
-- If this is a parsing or linting issue, please include a minimal SQL example which reproduces the issue, along with the `sqlfluff parse` output, `sqlfluff lint` output and `sqlfluff fix` output when relevant. Finally, please add any labels that are related to the issue before submitting so that it can be addressed easily! --

## Expected Behaviour

I would expect statements with keyword 'timestamp with time zone' to be parsed by sqlfluff, as they are supported by ANSI SQL:2011 standard.

## Observed Behaviour

> $ sqlfluff lint test.sql --dialect postgres 
== [test.sql] FAIL
L:   1 | P:  18 | L012 | Implicit aliasing of column not allowed. Use explicit
                       | `AS` clause.
L:   1 | P:  18 | L029 | Keywords should not be used as identifiers.
L:   1 | P:  22 |  PRS | Line 1, Position 22: Found unparsable section: " time
                       | zone '2005-04-02 12:00:00-07' + in..."

> $ sqlfluff parse test.sql --dialect postgres 
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    statement:
[L:  1, P:  1]      |        select_statement:
[L:  1, P:  1]      |            select_clause:
[L:  1, P:  1]      |                keyword:                                      'SELECT'
[L:  1, P:  7]      |                [META] indent:
[L:  1, P:  7]      |                whitespace:                                   ' '
[L:  1, P:  8]      |                select_clause_element:
[L:  1, P:  8]      |                    column_reference:
[L:  1, P:  8]      |                        identifier:                           'timestamp'
[L:  1, P: 17]      |                    whitespace:                               ' '
[L:  1, P: 18]      |                    alias_expression:
[L:  1, P: 18]      |                        identifier:                           'with'
[L:  1, P: 22]      |                    unparsable:                               !! Expected: 'Nothing...'
[L:  1, P: 22]      |                        whitespace:                           ' '
[L:  1, P: 23]      |                        raw:                                  'time'
[L:  1, P: 27]      |                        whitespace:                           ' '
[L:  1, P: 28]      |                        raw:                                  'zone'
[L:  1, P: 32]      |                        whitespace:                           ' '
[L:  1, P: 33]      |                        raw:                                  "'2005-04-02 12:00:00-07'"
[L:  1, P: 57]      |                        whitespace:                           ' '
[...]

PR #1398, opened 2021-09-13
Support TIMESTAMPTZ in TIME ZONE queries for Postgres

tunetheweb:
Further fix to #1155 

Nope, very simple fix.

### Pull Request checklist

  - Other.


WittierDinosaur:
@tunetheweb It's actually a bit more quirky than that:

Valid:
```
TIMESTAMPTZ AT TIME ZONE 'UTC'
TIMESTAMP WITH TIME ZONE AT TIME ZONE 'UTC'
```

Invalid:
```
TIMESTAMPTZ WITH TIME ZONE
```

tunetheweb:
At shouldn’t try to step in in places I obviously know little about! 😁

Will fix!

tunetheweb:
Could you have another look when you get a chance @WittierDinosaur ?

WittierDinosaur:
I might be reading it wrong, but I think this would fail on TIMESTAMP WITH TIME ZONE AT TIME ZONE 'UTC'?

tunetheweb:
> I might be reading it wrong, but I think this would fail on TIMESTAMP WITH TIME ZONE AT TIME ZONE 'UTC'?

Yes that would still fail. It wasn't one of the things requested in the that other issue but now see you did mention that in your comment above.

Fixing that was painful as there appears to be two types of similar things going on here:
- Casting values to `TIMESTAMP` type with the datatype after including optional `TIME ZONE` info (e.g. `SELECT '2015-10-24 16:38:46'::TIMESTAMP WITH TIME ZONE AT TIME ZONE 'UTC';`)
- Setting the Datatype upfront followed by a Date literal, which also might have optional `TIME ZONE` info (e.g. `SELECT timestamp with time zone '2025-11-20 00:00:00+00' AT TIME ZONE 'Africa/Cairo';`).

So we end up with both `DatatypeSegment` and `DateTimeLiteralGrammar` including `TIME ZONE` grammar for similar, but slightly different things. Which confused the heck out of me for a long time!

Anyway, if you could have another look that would be greatly appreciated.

And next time I'm definitely leaving even simple Postgres follow ups directed to you, for you! 😁

WittierDinosaur:
So looking at it again, there are a few things (this is less simple than I first thought):

I don't like the DateTimeLiteralGrammar, as written in ANSI, and when you replaced it, you freed up this to parse better:

```
        - interval_expression:
            keyword: interval
            literal: "'1 second'"
```

I think it might be worth replacing it with Nothing(), in Postgres at least, and just adding an optional LiteralGrammar to the sequence.

Separately, i think it might be worth extracting this to it's own class, along the lines of DatetimeCastingGrammar

```
Sequence("AT", "TIME", "ZONE", Ref("QuotedLiteralSegment"), optional=True)
```

That way we can just add it on the end, after optional LiteralGrammar.

And finally, just to be extra annoying, this is valid SQL. I just tested it on a Postgres instance:

```
SELECT '2018-09-02 07:09:19-04'::timestamptz AT TIME ZONE 'America/Chicago' AT TIME ZONE 'America/Chicago' AT TIME ZONE 'America/Chicago' AT TIME ZONE 'America/Chicago' AT TIME ZONE 'America/Chicago';
```

So maybe we should make it AnyNumberOf?

WittierDinosaur:
Oh, one more thing:

AT TIME ZONE works after both date and time. Just occurred to me that it might, and it does.

tunetheweb:
> So looking at it again, there are a few things (this is less simple than I first thought):
> 
> I don't like the DateTimeLiteralGrammar, as written in ANSI, and when you replaced it, you freed up this to parse better:
> 
> ```
>         - interval_expression:
>             keyword: interval
>             literal: "'1 second'"
> ```
>
> I think it might be worth replacing it with Nothing(), in Postgres at least, and just adding an optional LiteralGrammar to the sequence.
> 

I **did** like this. Isn't that what it should be?

> Separately, i think it might be worth extracting this to it's own class, along the lines of DateTimeCastingGrammar
> 
> ```
> Sequence("AT", "TIME", "ZONE", Ref("QuotedLiteralSegment"), optional=True)
> ```
>
> That way we can just add it on the end, after optional LiteralGrammar.

Is this just to avoid the repetition cause we use this in two places? If so I'm cool with that. Though I think might be better to leave the `optional=True` to the calling function, rather than embed in here. Have added in https://github.com/sqlfluff/sqlfluff/pull/1398/commits/ec19bd65c004f761a56526527be78de8b68b15b6

> And finally, just to be extra annoying, this is valid SQL. I just tested it on a Postgres instance:
> 
> ```
> SELECT '2018-09-02 07:09:19-04'::timestamptz AT TIME ZONE 'America/Chicago' AT TIME ZONE 'America/Chicago' AT TIME ZONE 'America/Chicago' AT TIME ZONE 'America/Chicago' AT TIME ZONE 'America/Chicago';
> ```
> 
> So maybe we should make it AnyNumberOf?

Good spot. Included in https://github.com/sqlfluff/sqlfluff/pull/1398/commits/ec19bd65c004f761a56526527be78de8b68b15b6

> Oh, one more thing:
> 
> AT TIME ZONE works after both date and time. Just occurred to me that it might, and it does.

Sorry not sure what you mean by this?

WittierDinosaur:
> I did like this. Isn't that what it should be?

Yes. What I mean is I don't like this in the ANSI dialect:
```
    DateTimeLiteralGrammar=Sequence(
        OneOf("DATE", "TIME", "TIMESTAMP", "INTERVAL"), Ref("QuotedLiteralSegment")
    ),
```
Replacing that definition allowed it to parse as interval_expression.

> Is this just to avoid the repetition cause we use this in two places? If so I'm cool with that. Though I think might be better to leave the optional=True to the calling function, rather than embed in here. Have added in ec19bd6

Agreed.

> Sorry not sure what you mean by this?

What I mean is DATE AT TIME ZONE 'UTC' is valid but DATE WITH TIME ZONE isn't, so can you include DATE in the OneOf before TimezoneGrammar?

Also, can you try replacing DateTimeLiteralGrammar with Nothing() in Postgres, I think it should be removed if it doesn't break anything.

WittierDinosaur:
Or maybe replace it in postgres with:

```
    DateTimeLiteralGrammar=Sequence(
        OneOf("DATE", "TIME", "TIMESTAMP", "INTERVAL"), Ref("QuotedLiteralSegment"), Ref("TimeZoneGrammar", optional=True),
```

WittierDinosaur:
At this point I'm on the verge of spamming this conversation, but i think the solution might be:

Spin out a new class for DateTimeTypeIdentifier grammar - which is DATE, TIME(STAMP(TZ)) (NumericLiteral) (WITH(OUT) TIME ZONE), INTERVAL

DataType Segment can then be: DateTimeTypeIndentifier, AnyNumberOf(TimezoneGrammar). 

And DatetimeLiteralGrammar can then be, DateTimeTypeIdentifier, QuotedLiteralSegment, TimezoneGrammar.

I think that would cover all bases

tunetheweb:
> SELECT DATE '2021-01-01';

That still parses for me:

```
% echo "SELECT DATE '2021-01-01';" | sqlfluff parse --dialect postgres -                                                           (non-zero-exit)
[L:  1, P:  1]      |file:
[L:  1, P:  8]      |                    expression:
[L:  1, P:  8]      |                        data_type:
[L:  1, P:  8]      |                            data_type_identifier:             'DATE'
[L:  1, P: 12]      |                        whitespace:                           ' '
[L:  1, P: 13]      |                        literal:                              "'2021-01-01'"
[L:  1, P: 25]      |            [META] dedent:
[L:  1, P: 25]      |    statement_terminator:                                     ';'
[L:  1, P: 26]      |    newline:                                                  '\n'
```

I'm not sure it's needed in ANSI to be honest. It says it's needed for EXASOL so maybe it shouldn't even be in ANSI.

It is needed in Postgres, but not for `DATE`

> What I mean is DATE AT TIME ZONE 'UTC' is valid but DATE WITH TIME ZONE isn't, so can you include DATE in the OneOf before TimezoneGrammar?

Again that parses for me without any changes needed:

```
% echo "SELECT DATE '2025-11-20' AT TIME ZONE 'Africa/Cairo';" | sqlfluff parse --dialect postgres -                               (non-zero-exit)
[L:  1, P:  1]      |file:
[L:  1, P:  8]      |                    expression:
[L:  1, P:  8]      |                        data_type:
[L:  1, P:  8]      |                            data_type_identifier:             'DATE'
[L:  1, P: 12]      |                        whitespace:                           ' '
[L:  1, P: 13]      |                        date_type_literal:
[L:  1, P: 13]      |                            literal:                          "'2025-11-20'"
[L:  1, P: 25]      |                            whitespace:                       ' '
[L:  1, P: 26]      |                            time_zone_grammar:
[L:  1, P: 26]      |                                keyword:                      'AT'
[L:  1, P: 28]      |                                whitespace:                   ' '
[L:  1, P: 29]      |                                keyword:                      'TIME'
[L:  1, P: 33]      |                                whitespace:                   ' '
[L:  1, P: 34]      |                                keyword:                      'ZONE'
[L:  1, P: 38]      |                                whitespace:                   ' '
[L:  1, P: 39]      |                                literal:                      "'Africa/Cairo'"
[L:  1, P: 53]      |            [META] dedent:
[L:  1, P: 53]      |    statement_terminator:                                     ';'
[L:  1, P: 54]      |    newline:                                                  '\n'
```

So not sure we need any further changes?

WittierDinosaur:
The reason most of this stuff parses though is the broadness of expression.

Most of the things I've commented are pedantic. But for consistency, the `DATE '2020-01-01'` in `SELECT DATE '2020-01-01'` should all parse as a datetime_literal. Whereas right now it parses as expression, datatype, literal

tunetheweb:
OK Gotcha. And thanks for the detailed instructions. Have committed those changes now and seems to work.

Phew this is definitely a lot more complicated than I originally thought it would be!!!

WittierDinosaur:
Yeah, all looks good now!

WittierDinosaur (approved):

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -91,6 +91,45 @@
 )
 
 
+@postgres_dialect.segment()
+class TimeZoneGrammar(BaseSegment):
+    """Literal Date Time with optional casting to Time Zone."""
+
+    type = "time_zone_grammar"
+    match_grammar = AnyNumberOf(
```
barrywhart, line 99: I'm not a frequent Postgres user, but I'm surprised this is `AnyNumberOf()`!

barrywhart (approved):
Looks good!

src/sqlfluff/dialects/dialect_postgres.py
tunetheweb: Agree but it is - see comment at the bottom of this https://github.com/sqlfluff/sqlfluff/pull/1398#issuecomment-918991891

SQL dialects are weird 🤷‍♂️

PR #1406, opened 2021-09-15
Remove unused deps in tox test docbuild

zhongjiajie:

ATT, remove unused deps in tox job

### DIFF
--- a/src/sqlfluff/rules/L008.py
+++ b/src/sqlfluff/rules/L008.py
@@ -25,6 +25,7 @@ class Rule_L008(BaseRule):
     | Keep a single space after the comma.
 
     .. code-block:: sql
+       :force:
 
         SELECT
             *
--- a/src/sqlfluff/rules/L008.py
+++ b/src/sqlfluff/rules/L008.py
@@ -25,7 +25,6 @@ class Rule_L008(BaseRule):
     | Keep a single space after the comma.
 
     .. code-block:: sql
-       :force:
 
         SELECT
             *
--- a/tox.ini
+++ b/tox.ini
@@ -86,11 +86,8 @@ commands = doc8 docs/source --file-encoding utf8
 [testenv:docbuild]
 deps =
     # -U keeps all requirements up-to-date
-    -Urrequirements.txt
-    -Urrequirements_dev.txt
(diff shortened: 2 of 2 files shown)


zhongjiajie:
As the result show on [CI result][1], we don't need reinstall dependent `requirements.txt` and `requirements_dev.txt` in job `docbuild` could in already install in `[testenv]`

So I think we could remove code in https://github.com/sqlfluff/sqlfluff/pull/1406/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449L89-L90

[1]: https://github.com/sqlfluff/sqlfluff/pull/1406/checks?check_run_id=3605563486

zhongjiajie:
Ref PR https://github.com/sqlfluff/sqlfluff/pull/1393

tunetheweb:
The results show that requirements.txt and requirements_dev.txt are not run. Thats why I added them. It would have been nice if it was possible for the job to inherit the original dependencies and add to it.

However, I do agree it looks like they are not actually needed though (which surprises me as the docs requirements.txt is so light!). I wonder if that's intentionally independent or by luck and the assumption will be that developers will have the parents requirements.txt and requirements_dev.txt installed?

Then again we have the check so if this ever changes, we can easily either fix the dependencies or add these lines back in. So will merge.

tunetheweb (approved):

zhongjiajie:
> The results show that requirements.txt and requirements_dev.txt are not run. 

I found out after you said, my initial think job `docbuild` had parent dependent 
https://github.com/sqlfluff/sqlfluff/blob/845c8eb3bb8d5c31a90c0d9c626c59834b92040e/tox.ini#L9-L21
So it only need to add `docs/requirements.txt` for addition
https://github.com/sqlfluff/sqlfluff/blob/845c8eb3bb8d5c31a90c0d9c626c59834b92040e/tox.ini#L87-L89

but after you said and I checkout the CI logs, and doc build just install dependent `docs/requirements.txt` and work fine
https://github.com/sqlfluff/sqlfluff/pull/1406/checks?check_run_id=3605563486#step:5:8
I make me surprised too.

PR #1407, opened 2021-09-15
Better error messages on error

tunetheweb:
Follow on from #1385 

Following changes:
- The current error is not highlighted in red so is a bit lost.
- Also we only show the error when in `verbose` mode which I think in hindsight is not great.
- Should use stderr so anyone using output from stdout can continue to
- Explain how to override

Nope.

### Pull Request checklist

  - Other.


nolanbconaway (approved):
Ah yeah good call

PR #1409, opened 2021-09-15
Added support for escape character strings to Postgres

### LINKED ISSUE #1069: Postgres dialect fails to parse some string literals with embedded backslashes, resulting in message "Zero length Lex item returned from '<unlexable>'"
Consider the following test case:

```
$ cat test.sql
SELECT '\';
```
When executing `sqlfluff fix test.sql` (or `sqlfluff fix --dialect postgres test.sql`) the following output is printed:

```
WARNING    Zero length Lex item returned from '<unlexable>'. Report this as a bug. 
```
I think the warning is unexpected, since, for example, the statement is a valid Postgres or SQLite query.

I'm using sqlfluff version 0.5.6 and found this bug by trying it on some statements generated by [SQLancer](https://github.com/sqlancer/sqlancer). I assume it reproduces on the latest version, since executing the query using the online demo version results in a `Internal Server Error`.

WittierDinosaur:
Added support for the different escape characters and string types in Postgres. 
Fixes #1069

Creating the correct regex may have cost me my sanity.

### Pull Request checklist
Test cases added
Added a comment block to explain the regex


test/fixtures/parser/postgres/postgres_single_quote.sql
```diff
@@ -0,0 +1,15 @@
...
+
+';
+
+SELECT '''aaa''';
+
+SELECT '
+''
+';
+
+SELECT '\';
```
tunetheweb, line 15: Can you fix the newlines:

image

```suggestion
SELECT '\';

```

Not least of all cause it would make my editorconfig change easier 😁

test/fixtures/parser/postgres/postgres_unicode_single_quote.sql
```diff
@@ -0,0 +1,24 @@
...
+
+SELECT U&'asdf' UESCAPE 'P';
+
+SELECT U&'
+somestuff
+'
+UESCAPE
+
+
+'?';
```
tunetheweb, line 24:
```suggestion
'?';

```

tunetheweb (approved):
Some questions and minor nits. But in general, assuming that horrendous regex is accurate the change LGTM.

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -27,15 +27,64 @@
...
+        #                                '(?!')    Followed by a single, which is not followed by a single quote
+        # - (\s*UESCAPE\s*'[^0-9A-Fa-f'+\-\s)]')?
+        #    \s*UESCAPE\s*                         Whitespace, followed by UESCAPE, followed by whitespace
+        #                 '[^0-9A-Fa-f'+\-\s)]'    Any character that isn't A-F, a-f, 0-9, +-, or whitespace, in quotes
+        #                                       ?  This last block is optional
+        RegexLexer(
+            "unicode_single_quote",
+            r"(?s)U&(('')+?(?!')|('.*?(?<!')(?:'')*'(?!')))(\s*UESCAPE\s*'[^0-9A-Fa-f'+\-\s)]')?",
+            CodeSegment,
+        ),
```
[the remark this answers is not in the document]
barrywhart: I ran all these regexes through a tool called `dlint`, which checks for the possibility of catastrophic backtracking, where a regex runs basically forever.

Reference:
* https://github.com/duo-labs/dlint/blob/master/docs/linters/DUO138.md
* https://github.com/duo-labs/dlint

The example script below includes an example regex known to have this issue, along with the 4 regexes from the PR.
```
import re

subject = 'x' * 64
re.search(r'(x+x+)+y', subject)  # Boom

# "unicode_single_quote",
re.search(r"(?s)U&(('')+?(?!')|('.*?(?<!')(?:'')*'(?!')))(\s*UESCAPE\s*'[^0-9A-Fa-f'+\-\s)]')?", subject)
# "escaped_single_quote",
re.search(r"(?s)E(('')+?(?!')|'.*?((?<!\\)(?:\\\\)*(?<!')(?:'')*|(?<!\\)(?:\\\\)*\\(?<!')(?:'')*')'(?!'))", subject)
# "unicode_double_quote",
re.search(r'(?s)U&".+?"(\s*UESCAPE\s*\'[^0-9A-Fa-f\'+\-\s)]\')?', subject)
# "single_quote"
re.search(r"(?s)('')+?(?!')|('.*?(?<!')(?:'')*'(?!'))", subject)
```

Here's the command I ran plus the output. Note that only the test regex is flagged:
```
(sqlfluff-3.7.9) ➜  sqlfluff git:(bhart-issue_845_l016_compute_line_length_before_template_expansion) ✗ python -m flake8 --select=DUO t.py
t.py:4:1: DUO138 catastrophic "re" usage - denial-of-service possible
```

PR #1410, opened 2021-09-15
Add .editorconfig config and enforce style rules

### LINKED ISSUE #1356: Add a .editorconfig file
As discussed in https://github.com/sqlfluff/sqlfluff/issues/1354#issuecomment-913054119 we should add a `.editorconfig` file.

[Editorconfig](https://editorconfig.org/) is a file which many IDEs support (sometimes via plugins) which enforce consistent styling which can be applied automatically (e.g. tabs versus spaces, newlines at the bottom).

It is added by placing an `.editorconfig` file in the root directory with a format like the below:

```
# editorconfig.org
root = true

[*]
indent_style = space
end_of_line = lf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true

[*.{html,md,js,css,sql}]
indent_size = 2

[*.py]
indent_size = 4

[*.rst]
indent_size = 3
```

Any IDE that a contributor uses, that supports this, will automatically enforce these standards.

We already have `black` to enforce this consistency for our python files, but enforcing (some of?) this for `.md`, `.rst`, and `.sql` files would help bring consistency and avoid "nit"-ty feedback on PRs.

Some items to consider:
- What rules to enforce? For Python this should be obvious as can follow `black`'s convention but for others might not be as obvious.
- Tabs versus spaces - I presume spaces since `black` does this for Python files. Anyone fancy opening the argument or can we just go with that?
- Indentation spacing for no-Python files? Again `black` has set the standard to 4 spaces for Python, but often 2 spaces are used for JS and other files. Luckily `.editorconfig` supports setting this by file type (see above) if we want to use this. Also our .rst files seem to use 3 spaces so is that the convention for there?
- Spot checking a few files, to ensure that adding a small tweak to any of those file types doesn't result in a load of white space noise in PRs forevermore.
- Maybe cleaning up all the files that don't match that to avoid item above?
- Add as [a pre-commit hook](https://github.com/editorconfig-checker/editorconfig-checker.python#usage-with-the-pre-commit-git-hooks-framework)?
- Add documentation for contributors to tell them how to install any necessary plugins for popular IDEs?

tunetheweb:
Fixes #1356 

See https://editorconfig.org/

This adds the `.editorconfig` config file which IDEs (e.g. Visual Studio Code) can use to automatically enforce certain rules on file save:
- space over tabs
- LF over CRLF (prevents the weird `^M` character on linux/Mac machines)
- UTF-8 charset (prevents encoding errors for non-ASCII characters)
- No trailing whitespace (waste of space)
- Ensuring files end in a newline (see [why this is better](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file)).
- Indent spaces for Python (4), .rst (3) and everything else (2).

Black (and the flake8 run during our automatic tests) already enforces many of these rules for our Python files, but this adds it for the other files.

Editorconfig is optional and depends on developers IDE enforcing the rules. But many tools (and projects) do use these, so good to have standards.

We also add a `.gitattributes` which will enforce the LF over CRLF rule when files are checked into git.

I have had to exclude certain files (`.sql` and `.yml`) as they need trailing or leading whitespace (e.g. to test whitespace rules!).

Additionally I have converted a lot of the files, so this PR looks bigger than it is but choose to ignore whitespaces to make it easier to review:

image

However I have not converted ALL the files. I have the left the test `.sql` and `.yml` files unconverted because there are a lot of them (201!) and I didn't want to convert them for two reasons: 1) the PR would be massive to review (though ignoring whitespace changes will help) and 2) the potential for merge conflicts is too great. So I've only converted the main code (so we don't see big diffs in future PR), but not all the test files. We are we therefore see changes with test cases as future developers check in code with suddenly enforced CRLF to LF changes, for example.

### Pull Request checklist

  - Other.


WittierDinosaur:
Won't this have the side effect that if contributors not using VSCode create or edit non-Python files, the next VSCode user's PR will include linting changes in (apparently) random files?

Also, do we know why this is affecting python files? Shouldn't flake8 have been catching them?

NiallRees:
> Won't this have the side effect that if contributors not using VSCode create or edit non-Python files, the next VSCode user's PR will include linting changes in (apparently) random files?
> 
> Also, do we know why this is affecting python files? Shouldn't flake8 have been catching them?

There's also a plugin for editorconfig in [atom](https://atom.io/packages/editorconfig) and [vim](https://github.com/editorconfig/editorconfig-vim).

tunetheweb:
For CRLF they will now be enforced on git checkin. That’s the main thing flake8 doesn’t currently catch but it’ll automatically be taken care of so won’t cause needless issues.

This might still cause a few other issues for missing new lines or missing whitespaces but only if those files are being edited in a future PR.

.gitatttributes
```diff
@@ -0,0 +1,10 @@
+# We'll let Git's auto-detection algorithm infer if a file is text. If it is,
+# enforce LF line endings regardless of OS or git configurations.
+* text=auto eol=lf
+
+# Allow Batch files ot be CRLF:
```
barrywhart, line 5:
```suggestion
# Allow Batch files to be CRLF:
```

barrywhart (approved):

tunetheweb:
Also going to look at a linter for our non-python files. Probably through the GitHub Super Linter (it’s how I found out about this project!).

WittierDinosaur:
> For CRLF they will now be enforced on git checkin. That’s the main thing flake8 doesn’t currently catch but it’ll automatically be taken care of so won’t cause needless issues.
> 
> This might still cause a few other issues for missing new lines or missing whitespaces but only if those files are being edited in a future PR.

Ah, makes sense!

barrywhart:
Should we consider adding a GitHub action to detect bad whitespace during the PR build/test process?

tunetheweb:
> Should we consider adding a GitHub action to detect bad whitespace during the PR build/test process?

Yup - see comment above:

> Also going to look at a linter for our non-python files. Probably through the GitHub Super Linter (it’s how I found out about this project!).

Though this will primarily be for the `.md` and `.rst` files since the `.sql` and `.yml` files are ignored for whitespace (though admittedly not for the end of file newlines).

PR #1411, opened 2021-09-15
Issue 845: L016 should compute line length prior to template expansion

### LINKED ISSUE #845: L016 reports "Line is too long" AFTER template expansion

barrywhart:
Fixes #845 
...

...

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L016.py
+++ b/src/sqlfluff/rules/L016.py
@@ -373,7 +373,9 @@ class Rule_L016(Rule_L003):
     def _compute_segment_length(cls, segment):
         if segment.is_type("newline"):
             return 0
-        slice_length = segment.pos_marker.source_slice.stop - segment.pos_marker.source_slice.start
+        slice_length = (
+            segment.pos_marker.source_slice.stop - segment.pos_marker.source_slice.start
+        )
         if slice_length:
             return slice_length
         else:
@@ -386,7 +388,10 @@ class Rule_L016(Rule_L003):
         # Use a set to avoid counting segments twice.
         seen_slices = set()
         for segment in segments:
-            slice = (segment.pos_marker.source_slice.start, segment.pos_marker.source_slice.stop)
+            slice = (
+                segment.pos_marker.source_slice.start,
+                segment.pos_marker.source_slice.stop,
+            )
             if slice not in seen_slices:
                 seen_slices.add(slice)
                 line_len += cls._compute_segment_length(segment)
--- a/src/sqlfluff/rules/L016.py
+++ b/src/sqlfluff/rules/L016.py
@@ -372,26 +372,36 @@ class Rule_L016(Rule_L003):
     @classmethod
     def _compute_segment_length(cls, segment):
         if segment.is_type("newline"):
+            # Generally, we won't see newlines, but if we do, simply ignore
+            # them. Rationale: The intent of this rule is to enforce maximum
+            # line length, and newlines don't make lines longer.
             return 0
+        # Compute the length of this segments in SOURCE space (before template
+        # expansion).
         slice_length = (
             segment.pos_marker.source_slice.stop - segment.pos_marker.source_slice.start
         )
         if slice_length:
             return slice_length
         else:
+            # If a segment did not originate from the original source, its slice
+            # length slice length will be zero. This occurs, for example, when
+            # other lint rules add indentation or other whitespace. In that
+            # case, compute the length of its contents.
             return len(segment.raw)
 
     @classmethod
     def _compute_source_length(cls, segments):
         line_len = 0
-        # Often, a single area of a template will supply many SQL tokens.
-        # Use a set to avoid counting segments twice.
         seen_slices = set()
         for segment in segments:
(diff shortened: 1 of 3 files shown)


barrywhart:
One test is failing -- need to look into that:
```
test/rules/std_fix_auto_test.py::test__std_fix_auto[bigquery-002_templating]
```

test/fixtures/linter/autofix/bigquery/002_templating/after.sql
```diff
@@ -1,14 +1,17 @@
--- A subset of the hairy test.
-SELECT
-    COUNT(1) AS campaign_count,
-    {{corr_states}}
-    {% for action in considered_actions %}
-        , aaa(
-            bbb(ccc({{metric}}_r, {{action}}), ddd({{metric}}_r)),
```
barrywhart, line 7: This test specifies `max_line_length` = 50, but this line was 67 characters long. The new result looks correct to me.

tunetheweb: A reverse case of what you are trying to fix here!

Post templating this is 47-48 characters long as the template values where shorter than the template place holders.

So I agree it's correct that it now splits it with your god change as it's now looking at the placeholder lengths.

src/sqlfluff/rules/L016.py
```diff
@@ -369,6 +369,34 @@ def _gen_line_so_far(raw_stack, initial_buff=None):
                 break  # pragma: no cover
```
tunetheweb, line 369: Not related to your change, but since you're in the code here anyway, is there a test case we should add for this?

test/fixtures/linter/autofix/bigquery/002_templating/after.sql
barrywhart: Exactly. The previous behavior was incorrect, and often very confusing from a user perspective.

src/sqlfluff/rules/L016.py
```diff
@@ -369,6 +369,44 @@ def _gen_line_so_far(raw_stack, initial_buff=None):
...
+        line_len = 0
+        seen_slices = set()
+        for segment in segments:
+            slice = (
+                segment.pos_marker.source_slice.start,
+                segment.pos_marker.source_slice.stop,
+            )
+            # Often, a single templated area of a source file will expand to
+            # multiple SQL tokens.Here, we use a set to avoid double counting
+            # the length of that text.
```
tunetheweb, line 404: Is this due to FOR loops for example? Is it worth mentioning that to explain it?

```suggestion
            # Often, a single templated area of a source file will expand to
            # multiple SQL tokens (for example when a For loop expands multiple
            # times). We use a set to avoid double counting the length of that
            # pre-templated text.
```

Or, if not, then you at least need a space between the sentences.

src/sqlfluff/rules/L016.py
```diff
@@ -429,7 +467,7 @@ def _eval(self, segment, raw_stack, **kwargs):
                     else:
                         break  # pragma: no cover
```
tunetheweb, line 468: Same comment here about a test case if you know how to add one through your understanding of this code.

tunetheweb (approved):
LGTM. A few comments for you to consider.

src/sqlfluff/rules/L016.py
barrywhart: I'll add an example here to the comment. 👍

src/sqlfluff/rules/L016.py
barrywhart: I think I'll pass on adding the test cases. Some of these rules are so large and complex, you can end up fixing an issue while gaining little or no insight into other areas. 😢

tunetheweb: Can't blame me for trying! 😉 Resolving.

src/sqlfluff/rules/L016.py
tunetheweb: That's much clearer now. Wasn't getting that before. Thanks.

src/sqlfluff/rules/L016.py
tunetheweb: Resolving as per above.

PR #1415, opened 2021-09-16
Remaining line endings

tunetheweb:

A follow on to #1410 these are the remaining files. Thought it over and though best to knock off in one go so we can see what's properly change with these files in future PRs.

Also corrects the name of `.gitattributes` which had one too many t's in it before!

Nope.

### Pull Request checklist

  - Other.


tunetheweb:
Make sure you use this setting when reviewing and you'll see there are no actual changes in this PR except one file rename:

image

tunetheweb:
This is the one real change I had to make to get the DBT tests to pass:

image

To be honest I don't really understand why!

It's complaining the line was too long without this, and it is just on the verge of what's acceptable (80 characters) but changing the file from CRLF to LF makes it think the line is too long now (I tested in `main` to confirm nothing else weird I did in this branch).

It's like rule L016 allows an extra character or two for CRLF files which is enough to allow it to pass. You'd think if anything it would be the other way around and it would allow one less character - weird. 🤷‍♂️ Another reason to shoot for consistency!

Anyway fixing the test case is easy and tests now pass.

barrywhart (approved):

PR #1416, opened 2021-09-16
TSQL: Remove match possibilities for segments with no TSQL equivalent

jpers36:
Remapping segments to Nothing() that have no TSQL equivalent.  This will keep the linter from pursing paths that include TSQL non-keywords and failing with the runtime error "Grammar refers to '<Keyword>KeywordSegment' which was not found in the tsql dialect".

### DIFF
--- a/src/sqlfluff/dialects/dialect_tsql.py
+++ b/src/sqlfluff/dialects/dialect_tsql.py
@@ -299,17 +299,6 @@ class DropModelStatementSegment(BaseSegment):
 
 
 @tsql_dialect.segment(replace=True)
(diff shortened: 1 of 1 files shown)


jpers36:
Something unusual is going on with UnorderedSelectStatementSegment.  It's failing to match on the create_view_with_cte.sql tsql test case when LimitClauseSegment and NamedWindowSegment are redefined to Nothing(), even though neither of those segments actually exists in the test case.  Removing those redefines for now, since this is meant to be a simple first PR.

tunetheweb (approved):

Issue #1274, opened 2021-08-13
Add a parameter to provide a .sqlfluff configuration file when using the .fix or .lint API methods


mayconrpgm:
Currently, when we use the sqlfluff API on a Python script as described in https://docs.sqlfluff.com/en/stable/api.html, it doesn't allow you to specify a configuration file built using the instructions at https://docs.sqlfluff.com/en/stable/configuration.html, it will only do the linting and fixing of SQL statements using the default configurations, allowing only to specify which dialect will be used:

```
import sqlfluff

sql_string = "SeLEct  *, 1, blah as  fOO  from myTable"

lint_result = sqlfluff.lint(sql_string, dialect="postgres")
print(lint_result)

fix_result = sqlfluff.fix(sql_string, dialect="postgres")
print(fix_result)
```

I was able to implement this funcionality by using the internal API as follows:

```
sql_string = "SeLEct  *, 1, blah as  fOO  from myTable"

parsed_sql = sqlfluff.parse(sql_string)

fluff_config = FluffConfig.from_path(path='.sqlfluff')
fluff_linter = Linter(config=fluff_config)

fixed_sql, linter_result = fluff_linter.fix(tree=parsed_sql.tree, config=fluff_config)

print(fixed_sql.raw)
```

But it's a little more cumbersome to use and I had to take some time digging into the code to better understand how to get the information needed from the returned objects.

It would be nice if the `sqlfluff.lint` and `sqlfluff.fix` functions had an optional parameter to pass a path to a configuration file or even a FluffConfig object by using the `FluffConfig.from_path` method from above:

```
import sqlfluff

sql_string = "SeLEct  *, 1, blah as  fOO  from myTable"

lint_result = sqlfluff.lint(sql_string, config="<path_to_sqlfluff_config_file>")
print(lint_result)

fix_result = sqlfluff.fix(sql_string, config="<path_to_sqlfluff_config_file>")
print(fix_result)
```

This issue is similar to this one, only that it applies to the API instead of the CLI: #1244

andres-lowrie:
don't want to hijack this thread but I think this is the correct place for these comments given the linked issues.

It would be helpful to be able to pass in an argument to the cli to have it override the various directories it looks for config files for something like 

```
sqlfluff --path-to-config-file some/path/here
```

[...]

PR #1419, opened 2021-09-17
Allow configuration to be passed when using simple API

matthieucan:

Add an optional `config` parameter to the "simple" API (`lint`, `fix` and `parse`). This allows to use sqlfluff programmatically with specific configuration.
My use case for this feature is to configure specific rules, but I believe this is beneficial to any API user who wants to change specific configuration options.

I'm not aware of any side-effect this feature brings, the configuration parameter is totally optional and is simply forwarded to `Linter`.

### Pull Request checklist

### DIFF
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -1,6 +1,8 @@
 """The simple public API methods."""
 
-from sqlfluff.core import Linter
+from typing import Optional
+
+from sqlfluff.core import Linter, FluffConfig
 
 
 class APIParsingError(ValueError):
@@ -24,7 +26,7 @@ def _unify_str_or_file(sql):
     return sql
 
 
-def lint(sql, dialect="ansi", rules=None):
+def lint(sql, dialect="ansi", rules=None, config: Optional[FluffConfig] = None):
     """Lint a sql string or file.
 
     Args:
@@ -34,12 +36,13 @@ def lint(sql, dialect="ansi", rules=None):
             to be linted. Defaults to `ansi`.
         rules (:obj:`str` or iterable of :obj:`str`, optional): A subset of rule
             reference to lint for.
+        config (:obj:`FluffConfig`): Optional configuration object.
 
     Returns:
         :obj:`list` of :obj:`dict` for each violation found.
     """
     sql = _unify_str_or_file(sql)
-    linter = Linter(dialect=dialect, rules=rules)
+    linter = Linter(dialect=dialect, rules=rules, config=config)
 
     result = linter.lint_string_wrapped(sql)
     result_records = result.as_records()
@@ -47,7 +50,7 @@ def lint(sql, dialect="ansi", rules=None):
     return [] if not result_records else result_records[0]["violations"]
 
 
-def fix(sql, dialect="ansi", rules=None):
+def fix(sql, dialect="ansi", rules=None, config: Optional[FluffConfig] = None):
     """Fix a sql string or file.
 
     Args:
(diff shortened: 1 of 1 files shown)


barrywhart (changes requested):
The code changes look good -- thanks! Do you think you could also update or add tests in `./test/api/simple_test.py` to use this new parameter? I'm thinking maybe one new test for each function you changed, using `FluffConfig` rather than the `dialect` and `rules` parameters to configure SQLFluff's behavior.

tunetheweb:
I’m not convinced about this change so please hold off merging and other work on this. Will respond with a fuller explanation in a bit.

tunetheweb:
I'm not convinced about this change to be honest.

While I appreciate the simplicity of the code change, and definitely do see the need for something like this (without which the Simple API is pretty limited in functionality to be honest!), I'm concerned about the fact this exposes internal workings of SQLFluff, and specifically the `FluffConfig` object, to the Simple API.

From [the API documentation](https://docs.sqlfluff.com/en/stable/api.html):

> The simple API presents only a fraction of the functionality present within the core SQLFluff library. For more advanced use cases, users can import the `Linter()` and `FluffConfig()` classes from `sqlfluff.core`. As of version 0.4.0 this is considered as _experimental only_ as **the internals may change without warning in any future release**. If you come to rely on the internals of SQLFluff, please post an issue on GitHub to share what you’re up to. This will help shape a more reliable, tidy and well documented public API for use.

That seems clear to me on the risk such users are agreeing to, by using the full API and not the simple API. I think this change is blurring those lines, but also not making this clear.

My understanding of that is the Simple API is supposed to be simple by definition, and retain API compatibility across versions and therefore pulling that `FluffConfig` object breaks that assumption.

I think a better change would be as suggested in #1274 to be able to pass a `.sqlfluff` file location or, alternatively, to be able to pass a config string of a file's contents rather than a path. To me either of those would avoid the potential future API compatibility issue I've highlighted. Though I admit they would be more effort than the change proposed here.

@barrywhart and the other @sqlfluff/sqlfluff-maintainers what do you think?

If we did go ahead with this change then we should at the very least add a few tests cases for it to [`simple-test.py`](https://github.com/sqlfluff/sqlfluff/blob/main/test/api/simple_test.py) as @barrywhart suggests, but I would suggest also updating the [example](https://github.com/sqlfluff/sqlfluff/blob/main/examples/01_basic_api_usage.py) which is pulled into [the API documentation page](https://docs.sqlfluff.com/en/stable/api.html) and also adding a warning to the config documentation changes added here.

barrywhart:
That's a great point. TBH, I'm not that familiar with the two APIs. Definitely don't want them to blend into one another.

Two questions:
* Are there specific user-level reasons for wanting this change? It shouldn't be "I want to pass a config object." It should be "I need to do X."
* Given we do have a specific use case for this change, how difficult is it to use the full API for that? If it's too difficult, could we address it some other way, such as improved documentation, examples, etc.?

tunetheweb:
I think there's definitely use cases where you would want to be able to pass config for a Simple API call.

Lots of rules have config options that are useful to set based on the condign style you wan. Currently we allow you to specify `upper` or `lower` capitalisation for example. To **only** be able to use the default for the Simple API is very limiting.

And using the full API is overkill for that IMHO, and also adds risk if we do change this.

So I can clearly see the need for this change.

However, as I say, I think the answer is to be able to specify a config file location (or string representation of that file's contents), rather than a `FluffConfig` object is the more robust solution that maintains the internal independence.

pwildenhain:
I completely agree with Barry 😉 


For real, I'm aligned with @tunetheweb -- we should allow a `.sqlfluff` to be passed, but we shouldn't expose the `FluffConfig` object since that isn't guaranteed to stay the same between minor versions.

matthieucan:
Hi all,
Thank you so much for the very quick feedback. I totally understand the concerns.
I think I could implement the option to forward the content of a `.sqlfluff`. Is that better than giving the configuration through Python datastructures? (as in, the content of an .ini as opposed to a dictionary?)

matthieucan:
> Lots of rules have config options that are useful to set based on the coding style you want. Currently we allow you to specify `upper` or `lower` capitalisation for example. To **only** be able to use the default for the Simple API is very limiting.

This was exactly my use-case, actually. But it will also happen for other rules :)

tunetheweb:
> I think I could implement the option to forward the content of a `.sqlfluff`. Is that better than giving the configuration through Python datastructures? (as in, the content of an .ini as opposed to a dictionary?)

I think it is to be honest. Obviously it's not as nice programmatically compared to a structure, but as I say it keeps the internal code independent. And the .sqlfluff config file format is our API "contract" of how to configure SQLFluff.

tunetheweb:
Closing for now. Let's open a new PR with the alternative approach when ready (would be good to reference this PR in it).

PR #1420, opened 2021-09-17
TSQL: Remove Limit and NamedWindow segments as not supported in T-SQL

jpers36:

Removed two segments not supported by T-SQL to prevent keyword errors if it tries to go down these paths.


src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -896,3 +896,10 @@ def match(self, segments, parse_context):
         dialects.
         """
         return MatchResult.from_unmatched(segments)
+
+    def simple(self, parse_context: ParseContext) -> Optional[List[str]]:
+        """Does this matcher support a lowercase hash matching route?
+
+        Yes, trivially so.
+        """
+        return ["1"]
```
tunetheweb, line 905: I'm a bit concerned with the unintended consequences of this to be honest. But don't understand it enough.

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -307,3 +307,25 @@ class OverlapsClauseSegment(BaseSegment):
...
+
+@tsql_dialect.segment(replace=True)
+class NamedWindowSegment(BaseSegment):
+    """A WINDOW clause.
+
+    Not present in T-SQL.
+    """
+
+    type = "named_window"
+    match_grammar = Nothing()
```
tunetheweb, line 331: I'm not sure exactly what you are trying to achieve here but not sure litering the dialect with all these `Nothing()`s is the way to go about it.

Surely you would be better instead to update the segments referencing them?:

```suggestion


@tsql_dialect.segment(replace=True)
class UnorderedSelectStatementSegment(BaseSegment):
    """A `SELECT` statement without any ORDER clauses or later.

    We need to change ANSI slightly
    """

    type = "select_statement"
    # Remove the LimitClauseSegment and NamedWindowSegment from ANSI
    match_grammar = StartsWith(
        Ref("SelectClauseSegment"),
        terminator=OneOf(
            Ref("SetOperatorSegment"),
            Ref("WithNoSchemaBindingClauseSegment"),
            Ref("OrderByClauseSegment"),
        ),
        enforce_whitespace_preceeding_terminator=True,
    )

    parse_grammar = ansi_dialect.get_segment(
        "UnorderedSelectStatementSegment"
    ).match_grammar.copy()


@tsql_dialect.segment(replace=True)
class SelectStatementSegment(BaseSegment):
    """A `SELECT` statement.

    We need to change ANSI slightly
    """

    type = "select_statement"
    match_grammar = ansi_dialect.get_segment(
        "SelectStatementSegment"
    ).match_grammar.copy()

    # Remove the Limit and Window statements from ANSI
    parse_grammar = UnorderedSelectStatementSegment.parse_grammar.copy(
        insert=[
            Ref("OrderByClauseSegment", optional=True),
        ]
    )
```
 
Note above code also requires importing `StartsWith`

src/sqlfluff/core/parser/grammar/base.py
jpers36: I went to bed Friday evening confident in my code and woke up Saturday morning with the same question.  I was hoping someone with experience in the match logic could chime in and allay or confirm my concerns.

barrywhart: Maybe @alanmcruickshank or @dmateusp? I've worked on the parser, but I haven't remembered much; I have to go and review the code if I am getting into anything deep.

src/sqlfluff/dialects/dialect_tsql.py
jpers36: That approach is an option.  My consideration is that overwriting more central segments like UnorderedSelectStatementSegment is a bigger code change.  If two or three months from now UnorderedSelectStatementSegment receives a rewrite for some reason -- performance, breakfix, functionality -- the TSQL dialect wouldn't receive the benefits of that rewrite.  On the other hand, of course, TSQL wouldn't receive any negative impacts of that rewrite.  Adding the Nothing() stubs feels more elegant to me, but I'm the neophyte here regarding both python and SQLFluff.  Let me know if any of that makes sense.

tunetheweb: I think for complex segments you're right. But in that case, it could also be refactored to make it easier to overwrite only certain functions in different dialects.

For simple change like here - where the change is almost the same size as the other proposal, I think this is better. And avoids the need for the change to the `Nothing()` segment above.

The `Nothing()` approach feels like a patch on top of a patch, so should be used sparingly IMHO (we have very few instances of these, and mostly for simple Grammars).

tunetheweb (approved):
Made some quick changes to get tests to pass.

Happy to merge now.

PR #1423, opened 2021-09-19
Added support for psql meta commands

### LINKED ISSUE #1327: Getting unlexable error when parsing psql meta-commands
Here is the minimal SQL example which reproduces the issue

The linter bugs out anytime the code includes psql meta-commands

```sql
begin;

\echo 'Creating the api schema...'  -- <<< the offending code

CREATE schema authorization api;

comment on schema api is
    'Schema that defines an API suitable to be exposed through PostgREST';

set role api;
```

## Expected Behaviour
It should ignore where it does not have a point of view.

## Error returned
```fish
FAIL - Zero length Lex item returned from '<unlexable>'. Report this as a bug.
```

## Steps to Reproduce
sqlfluff lint file_with_meta_commands.sql

## Dialect
postgres

## Version
fluff: 0.6.4
pip: 3.9.0

## Configuration
no configuration

WittierDinosaur:
This PR adds support for psql meta commands. Psql commands here start with \, followed by at least one alphabet character, and are then terminated by newline or \\. The File and Statement Segments had to be edited to accommodate for this new functionality.
 
Fixes #1327 

No, Postgres-specific, and all dialect tests are passing

### Pull Request checklist
Test cases added for psql meta commands, and sql commands in the same file (or same line for \\)
Documented regex in file

### DIFF
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -15,6 +15,7 @@ from sqlfluff.core.parser import (
     NamedParser,
     SymbolSegment,
     StartsWith,
+    GreedyUntil
 )
 
 from sqlfluff.core.dialects import load_raw_dialect
@@ -74,6 +75,29 @@ postgres_dialect.insert_lexer_matchers(
     before="not_equal",
 )
 
+postgres_dialect.insert_lexer_matchers(
+    [
+        # Explanation for the regex
+        # \\([^(\\\r\n)])+((\\\\)|(?=\n)|(?=\r\n))?
+        # \\                                        Starts with backslash
+        #   ([^(\\\r\n)])+                          Anything that is not a newline or a backslash
+        #                 (
+        #                  (\\\\)                   Double backslash
+        #                        |                  OR
+        #                         (?=\n)            The next character is a newline
+        #                               |           OR
+        #                                (?=\r\n)   The next 2 characters are a carriage return and a newline
+        #                                        )
+        #                                         ? The previous clause is optional
+        RegexLexer(
+            "meta_command",
+            r"\\([^(\\\r\n)])+((\\\\)|(?=\n)|(?=\r\n))?",
+            CodeSegment
+        )
+    ],
+    before="code",  # Final thing to search for - as psql specific
+)
+
 postgres_dialect.patch_lexer_matchers(
     [
         # In Postgres, the only escape character is ' for single quote strings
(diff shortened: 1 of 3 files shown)


tunetheweb:
I was curious how comments were handled as seems like a similar use case.

From a quick look it looks like it basically trims the details as part of the lexing, so those symbols aren’t sent to the parser and are effectively ignored.

The downside would be that you can’t check the grammar of meta commands (as they will effectively not exist to the parser), but doesn’t look like you are doing that here anyway so for now that wouldn’t matter. But your solution at least opens that possibility later.

Any thoughts on whether following the comments process would be the better solution than what’s proposed here?

WittierDinosaur:
Yeah, that might be easier. The only thing I think could cause an issue, is:
`\echo "thing" \\ SELECT 1;`
Comments are easy to ignore because they can't be terminated mid-line, psql meta commands can. We would just need to make sure the whitespace is preserved during linting, if the command gets ignored in parsing. I could add an optional \s after the \\\\?

tunetheweb:
> Comments are easy to ignore because they can't be terminated mid-line, psql meta commands can.

Sure they can, with `/*...*/` syntax:

```sql
SELECT *
/* this is our main table */ FROM table 1
```

WittierDinosaur:
Good point. I've updated meta commands to function like comments.

tunetheweb:
> Good point. I've updated meta commands to function like comments.

Wow that seems almost too simple now!!! I took ages trying to hunt for the code to figure out where the actual magic was happening - but guess it's all handled just by calling it a `CommentSegment`?

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -74,6 +75,27 @@
...
+        #                 (
+        #                  (\\\\)                   Double backslash
+        #                        |                  OR
+        #                         (?=\n)            The next character is a newline
+        #                               |           OR
+        #                                (?=\r\n)   The next 2 characters are a carriage return and a newline
+        #                                        )
+        #                                         ? The previous clause is optional
+        RegexLexer(
+            "meta_command", r"\\([^(\\\r\n)])+((\\\\)|(?=\n)|(?=\r\n))?", CommentSegment
```
tunetheweb, line 93: What do you think about a comment here?

```suggestion
            # For now we'll just treat meta syntax like comments and so just ignore them.
            # In future we may want to enhance this to actually parse them to ensure they are
            # valid meta commands.
            "meta_command", r"\\([^(\\\r\n)])+((\\\\)|(?=\n)|(?=\r\n))?", CommentSegment
```

tunetheweb (approved):
LGTM! One small suggestion.

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: Committed.

tunetheweb (approved):

PR #1426, opened 2021-09-20
Add default parameter to dbt "var" macro stub

CyberShadow:

Add the default parameter for the stub for the dbt `var` macro.

https://docs.getdbt.com/reference/dbt-jinja-functions/var#variable-default-values

Not that I'm aware

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -31,7 +31,7 @@ apply_dbt_builtins = True
 dbt_ref = {% macro ref(model_ref) %}{{model_ref}}{% endmacro %}
 dbt_source = {% macro source(source_name, table) %}{{source_name}}_{{table}}{% endmacro %}
 dbt_config = {% macro config() %}{% for k in kwargs %}{% endfor %}{% endmacro %}
-dbt_var = {% macro var(variable) %}item{% endmacro %}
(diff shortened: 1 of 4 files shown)


tunetheweb:
Can someone with more dbt knowledge than me review this? Seems simple enough but no idea if it'll have any unintended side effects.

barrywhart (approved):
Looks good to me, but I have almost no knowledge of dbt.

Can you take a look, @pwildenhain?

dmateusp (approved):
LGTM :+1:

dmateusp:
To explain a little bit that macro, dbt allows variables to be passed as a dictionary through the CLI, and defaults to be set in `dbt_project.yml`. The API for the macro works a little bit like `.get` on a Python dictionary, where the second argument is a default value in case no default was already provided.
I guess because of our current stub, @CyberShadow must have hit an issue where he was using the syntax to provide a default but the SQLFluff stub didn't expect a second argument (but I'm guessing)

CyberShadow:
>  I guess because of our current stub, @CyberShadow must have hit an issue where he was using the syntax to provide a default but the SQLFluff stub didn't expect a second argument (but I'm guessing)

Yes, that's right. (Well, that's one of the many issues that prevents SQLFluff from parsing our code base, but I guess we can take things one step at a time.)

PR #1430, opened 2021-09-21
Fix runtime error (IndexError ) when linting file with jinja "if"

### LINKED ISSUE #1414: Runtime error (IndexError ) linting file with jinja "if"
## Expected Behaviour

Get list of rule violations by line  for the file:

```sql
-- This file combines product data from individual brands into a staging table
{% set products =  [
  'tabe1',
  'table2'] %}

{% for product in products %}
SELECT
  brand,
  country_code,
  category,
  name,
  id
FROM
  {{ ref(product) }}
{% if not loop.last -%} UNION ALL {%- endif %}
{% endfor %}
```

## Observed Behaviour

Cannot make sqlfluff work with jinja for loops.

Error:

```
WARNING    Unable to lint /app/models/3_INFORMATION_MART/orders.sql due to an internal error. Please report this as an issue with your query's contents and stacktrace below!
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/linter/runner.py", line 94, in run
    yield partial()
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/linter/linter.py", line 512, in lint_rendered
    parsed = cls.parse_rendered(rendered)
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/linter/linter.py", line 294, in parse_rendered
    tokens, lvs, config = cls._lex_templated_file(
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/linter/linter.py", line 127, in _lex_templated_file
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/parser/lexer.py", line 319, in lex
    segments: Tuple[RawSegment, ...] = self.elements_to_segments(
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/parser/lexer.py", line 346, in elements_to_segments
    source_slice = templated_file.templated_slice_to_source_slice(
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/templaters/base.py", line 287, in templated_slice_to_source_slice
    start_slices = [self.sliced_file[ts_start_sf_start]]
IndexError: list index out of range

```

## Steps to Reproduce

run lint against the target file

## Dialect
dbt

## Version
0.6.4

## Configuration
```
[sqlfluff]
templater = dbt
exclude_rules = L031,L029

[sqlfluff:rules]
tab_space_size = 2

[sqlfluff:templater:dbt]
profiles_dir = .
profile = dev

capitalisation_policy = upper

```

barrywhart:

Fixes #1414

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/templaters/base.py
+++ b/src/sqlfluff/core/templaters/base.py
@@ -285,8 +285,12 @@ class TemplatedFile:
                 ts_start_sf_stop, ts_stop_sf_stop
             )
         ]
-        if ts_start_sf_start == ts_start_sf_stop:  # pragma: no cover TODO?
-            start_slices = [self.sliced_file[ts_start_sf_start]]
+        if ts_start_sf_start == ts_start_sf_stop:
+            if ts_start_sf_start < len(self.sliced_file):
+                start_slices = [self.sliced_file[ts_start_sf_start]]
+            else:
+                assert ts_start_sf_start == len(self.sliced_file)
+                return self.sliced_file[-1].source_slice
         else:
             start_slices = self.sliced_file[ts_start_sf_start:ts_start_sf_stop]
         if ts_stop_sf_start == ts_stop_sf_stop:  # pragma: no cover TODO?
--- a/src/sqlfluff/core/templaters/base.py
+++ b/src/sqlfluff/core/templaters/base.py
@@ -286,10 +286,12 @@ class TemplatedFile:
             )
         ]
         if ts_start_sf_start == ts_start_sf_stop:
+            if ts_start_sf_start > len(self.sliced_file):  # pragma: no cover TODO?
+                # We should never get here
+                raise ValueError("Starting position higher than sliced file position")
             if ts_start_sf_start < len(self.sliced_file):
-                start_slices = [self.sliced_file[ts_start_sf_start]]  # pragma: no cover TODO?
+                return self.sliced_file[1].source_slice
             else:
-                assert ts_start_sf_start == len(self.sliced_file)
                 return self.sliced_file[-1].source_slice
         else:
             start_slices = self.sliced_file[ts_start_sf_start:ts_start_sf_stop]
--- a/src/sqlfluff/core/templaters/base.py
+++ b/src/sqlfluff/core/templaters/base.py
@@ -286,7 +286,7 @@ class TemplatedFile:
             )
         ]
         if ts_start_sf_start == ts_start_sf_stop:
-            if ts_start_sf_start > len(self.sliced_file):  # pragma: no cover TODO?
+            if ts_start_sf_start > len(self.sliced_file):  # pragma: no cover
                 # We should never get here
                 raise ValueError("Starting position higher than sliced file position")
             if ts_start_sf_start < len(self.sliced_file):
--- a/src/sqlfluff/core/templaters/base.py
+++ b/src/sqlfluff/core/templaters/base.py
@@ -287,7 +287,7 @@ class TemplatedFile:
         ]
         if ts_start_sf_start == ts_start_sf_stop:
             if ts_start_sf_start < len(self.sliced_file):
-                start_slices = [self.sliced_file[ts_start_sf_start]]
+                start_slices = [self.sliced_file[ts_start_sf_start]]  # pragma: no cover TODO?
             else:
                 assert ts_start_sf_start == len(self.sliced_file)
                 return self.sliced_file[-1].source_slice
--- a/test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/after.sql
+++ b/test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/after.sql
@@ -1,0 +1,16 @@
+-- This file combines product data from individual brands into a staging table
+{% set products =  [
+  'table1',
+  'table2'] %}
+
+{% for product in products %}
+SELECT
+    brand,
+    country_code,
+    category,
+    name,
+    id
+FROM
+    product
+{% if not loop.last -%} UNION ALL {%- endif %}
+{% endfor %}
--- a/test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/before.sql
+++ b/test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/before.sql
@@ -1,0 +1,16 @@
+-- This file combines product data from individual brands into a staging table
+{% set products =  [
+  'table1',
+  'table2'] %}
+
+{% for product in products %}
+SELECT
+  brand,
+  country_code,
+  category,
+  name,
+  id
+FROM
+  product
+{% if not loop.last -%} UNION ALL {%- endif %}
+{% endfor %}
--- a/test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/before.sql
+++ b/test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/before.sql
@@ -11,6 +11,6 @@ SELECT
   name,
   id
 FROM
-  product
+  {{ product }}
 {% if not loop.last -%} UNION ALL {%- endif %}
 {% endfor %}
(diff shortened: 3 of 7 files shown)


test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/after.sql
```diff
@@ -0,0 +1,16 @@
...
+
+{% for product in products %}
+SELECT
+    brand,
+    country_code,
+    category,
+    name,
+    id
+FROM
+    product
```
tunetheweb, line 14: This to me is wrong as `product` is a variable, so this should be:

```suggestion
    {{ product }}
```

However testing this leads to another problem (discussed below)...

test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/after.sql
```diff
@@ -0,0 +1,16 @@
...
+{% if not loop.last -%} UNION ALL {%- endif %}
```
tunetheweb, line 15: Can we also add a test case as well without the `-%}`? I.e. with:

```jinja
{% if not loop.last %} UNION ALL {% endif %}
```

And also another test case with something afterwards (a `;` or even an `ORDER BY`)? When you add this, this causes the middle path (`ts_start_sf_start < len(self.sliced_file)`) to be covered (hence no need for `pragma: no cover` in there!).

test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/before.sql
```diff
@@ -0,0 +1,16 @@
...
+
+{% for product in products %}
+SELECT
+  brand,
+  country_code,
+  category,
+  name,
+  id
+FROM
+  product
```
tunetheweb, line 14: This really should be this as per above:

```suggestion
  {{ product }}
```

However when I test this with `{% if not loop.last -%} UNION ALL {%- endif %}` syntax it _dedents_ this as part of the fix (so fixes it to `{{ product}}` instead of `••••{{ product }}`(where `•` represents a space) but it correctly indents it with `{% if not loop.last %} UNION ALL {% endif %}` so a bit confused as to why that happens. Any ideas?

tunetheweb (changes requested):
Suggesting some changes below.

However think there is more going on as discovered some other oddness when testing multiple `for` loops in the same file.

For example this fails with an L016 exception:

```sql
-- This file combines product data from individual brands into a staging table
{% set products =  [
  'table1',
  'table2'] %}

{% for product in products %}
SELECT
  brand,
  country_code,
  category,
  name,
  id
FROM
  {{ product }};
{% endfor %}

{% for product in products %}
SELECT
  brand,
  country_code,
  category,
  name,
  id
FROM
  {{ product }};
{% endfor %}
```

Seems to be unrelated to your change (it's broken with `main` code, wiht your suggested changes, and also with my suggested further changes to your change) so maybe raise a separate issue for this one?

src/sqlfluff/core/templaters/base.py
```diff
@@ -285,8 +285,12 @@ def templated_slice_to_source_slice(
...
             )
         ]
-        if ts_start_sf_start == ts_start_sf_stop:  # pragma: no cover TODO?
-            start_slices = [self.sliced_file[ts_start_sf_start]]
+        if ts_start_sf_start == ts_start_sf_stop:
+            if ts_start_sf_start < len(self.sliced_file):
+                start_slices = [self.sliced_file[ts_start_sf_start]]  # pragma: no cover TODO?
+            else:
+                assert ts_start_sf_start == len(self.sliced_file)
+                return self.sliced_file[-1].source_slice
```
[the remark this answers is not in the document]
barrywhart: Note the assert is to capture my (limited) understanding of the code. I could remove the assert, but I think it'll be helpful if we encounter future bugs in this area. Note this is an area of the code with lots of "no cover". Maybe we can add tests as users find issues or our understanding of this code improves.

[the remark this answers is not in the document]
barrywhart: Can you add the suggested test case here?
>Also, while this fixes the test case you have, it doesn't work if there is anything after the test case (even a semi-colon!)

test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/after.sql
barrywhart: Oops, I just intended to use a placeholder table name. I'll change it to something else like `foo`.

src/sqlfluff/core/templaters/base.py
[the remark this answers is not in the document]
tunetheweb: Happy to commit my changes to this branch now if you want?

test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/after.sql
tunetheweb: Actually I think we should use the real templated name (`{{ product }}`) since it has shown another error. Even if we don;'t fix that here. Again happy to commit my change if you want.

src/sqlfluff/core/templaters/base.py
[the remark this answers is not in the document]
tunetheweb: I went ahead and did this in https://github.com/sqlfluff/sqlfluff/pull/1430/commits/62687a79887205a0c877ccdd8bf18be44e39738e

test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/after.sql
tunetheweb: Added this in https://github.com/sqlfluff/sqlfluff/pull/1430/commits/62687a79887205a0c877ccdd8bf18be44e39738e

test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/after.sql
```diff
@@ -0,0 +1,16 @@
...
+
+{% for product in products %}
+SELECT
+    brand,
+    country_code,
+    category,
+    name,
+    id
+FROM
+{{ product }}
```
tunetheweb, line 14: This is wrong and shouldn't dedent. But not sure why it does. Any ideas @barrywhart ?

src/sqlfluff/core/templaters/base.py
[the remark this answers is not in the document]
tunetheweb: Resolving this comment now assuming you're happy with that extra code I committed (revert it if not).

test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/after.sql
barrywhart: I don't know exactly, but I do know that SQLFluff is conservative about applying fixes in or near templated areas (i.e. curly braces). Probably related to that. Want to create another issue for that? I'm happy to keep working on these, but trying to avoid letting any single issue become too much of a 🐇 hole.

barrywhart:
I'll create an issue for the other thing you found. I thought _possibly_ my other PR (#1431) may help with that, but it didn't.

barrywhart:
Created issue #1433.

I think this is ready for another review. Want me to also create an issue for the indentation thing with `{{ product }}`? It seems like a little lower priority than the L016 exceptions, but happy to do it if you like. 🙏🏽

test/fixtures/linter/autofix/ansi/016_index_error_with_jinja_if/after.sql
barrywhart: I confirmed just now that it indents correctly when run on a post-expansion version of the file, i.e.:
```
-- This file combines product data from individual brands into a staging table
SELECT
    brand,
    country_code,
    category,
    name,
    id
FROM
    table1
UNION ALL
SELECT
    brand,
    country_code,
    category,
    name,
    id
FROM
    table2
```

tunetheweb (approved):
I'm happy with this if you are @barrywhart . Gave it a thorough review earlier and suggested changes which have been applied.

As discussed, investigations have thrown up another few related issues, but they didn't work previously either and this seems a good improvement.

Issue #1425, opened 2021-09-19
`sqlfluff fix` corrupts Jinja for loop


CyberShadow:
## Expected Behaviour

Output should be semantically equivalent to input, and have valid syntax.

## Observed Behaviour

Output seems to be corrupted.

## Steps to Reproduce

```console
$ printf 'SELECT\n    1,\n{%% for _ in [1, 2, 3] %%} 2,{%%endfor %%}\n' |  sqlfluff fix -
Unfixable violations detected.
SELECT
    1,
    {% for _ in [1, 2, 3] %}
 {%endfor %}
```

Note that the contents of the `for` loop was completely deleted.

Slightly varying the input causes the output to be corrupted in other ways (e.g. `2` is present but the comma is not, or the `2` is duplicated multiple times).

## Dialect

None specified

## Version
SQLFluff ce4e5a344526f7ee61a0950adc079e4d3b5af438, Python 3.9.7

## Configuration

None

tunetheweb:
Is the double percentage valid Jinja syntax? Never seen that before to be honest.

tunetheweb:
> Is the double percentage valid Jinja syntax? Never seen that before to be honest.

Oh is this needed for `printf`? So can be ignored?

CyberShadow:
It's to escape the special meaning of % in `printf`. Sorry for the confusion, I'll avoid it in the future if you can recommend a simpler method to unambiguously include test cases.

tunetheweb:
No that’s fine. Just confused me for a second.

Seems similar to #1162 and in particular this comment: https://github.com/sqlfluff/sqlfluff/issues/1162#issuecomment-914611474

barrywhart:
Reproduced this issue using a file (no printf):
```
SELECT
    1,
{% for _ in [1, 2, 3] %} 2,{% endfor %}
```

The result after `sqlfluff fix` was:
```
SELECT
    1,
    {% for _ in [1, 2, 3] %}
 {% endfor %}
```

barrywhart:
Slightly different example -- planning to use this one for deeper debugging as it only triggers a single rule violation, and that one is inside the `{% for %}` loop.

```
SELECT
{% for _ in [1, 2, 3] %} 2,
{% endfor %}
    10
```

barrywhart:
@CyberShadow: Do you recall how you produced the behavior
>the 2 is duplicated multiple times).

[...]

PR #1431, opened 2021-09-22
Issues 1162, 1425: Fix some issues where the SQL file is corrupted by lint "fixes" in or near Jinja loops

### LINKED ISSUE #1162: SQL JINJA IS EXPLODING with sqlfluff fix

SIMPLE SQL QUERY
```
SELECT
    apple
    , {% for num in ('1', '2', '3') %}
            CASE WHEN {{ num }} = 1 THEN cool
            ELSE moo END as {{ num }}_col
      {% endfor -%}
    , pear
from fruit f
```

## Expected Behaviour

```
SELECT
    apple,
    {% for num in ('1', '2', '3') %}
            CASE WHEN {{ num }} = 1 THEN cool
            ELSE moo END as {{ num }}_col
      {% endfor -%}
    pear
FROM fruit
```

## Observed Behaviour
```
SELECT
    apple,
    {% for num in ('1', '2', '3') %}
            CASE WHEN {{ num }} = 1 THEN cool
            ELSE moo END as {{ num }}_col
      {% endfor -%}
    CASE WHEN 1 = 1 THEN cool
            ELSE moo END as 1_col

            CASE WHEN 2 = 1 THEN cool
            ELSE moo END as 2_col

            CASE WHEN 3 = 1 THEN cool
            ELSE moo END as 3_col
FROM fruit
```

## Steps to Reproduce
create a file called test.sql with
```
SELECT
    apple,
    {% for num in ('1', '2', '3') %}
            CASE WHEN {{ num }} = 1 THEN cool
            ELSE moo END as {{ num }}_col
      {% endfor -%}
    pear
FROM fruit

```

## Version

sqlfluff, version 0.6.0
Python 3.7.9
## Configuration
```
```
```
[sqlfluff]
dialect = snowflake
load_macros_from_path = macros

[sqlfluff:rules:L010]
capitalisation_policy = upper

#[sqlfluff:rules:L016]
#max_line_length = 999

[sqlfluff:rules:L019]
comma_style = trailing
```

barrywhart:
Fixes #1162, #1425 

This PR adds a couple of new internal features to SQLFluff:
* `TemplatedFile` adds a new property, `raw_slice_block_info`, which identifies logical sections of the templated file: e.g. blocks and loops.
* All `LintResult`s now get their fixes automatically analyzed in terms of template structure. Unsafe fixes are automatically discarded, e.g. moving code into or out of a loop.

### Pull Request checklist

  - Other.


src/sqlfluff/core/linter/linted_file.py
```diff
@@ -205,15 +205,16 @@ def _log_hints(
     ):
         """Log hints for debugging during patch generation."""
         # This next bit is ALL FOR LOGGING AND DEBUGGING
-        if patch.templated_slice.start >= 10:
+        max_log_length = 10
```
barrywhart, line 208: Unrelated to the fix... this change makes it easier to tweak the max size of log hints for patches. Larger size can be helpful for debugging some issues.

src/sqlfluff/core/linter/linted_file.py
```diff
@@ -325,6 +333,19 @@ def fix_string(self) -> Tuple[Any, bool]:
 
             # Deal with the easy case of only literals
             if set(local_type_list) == {"literal"}:
+                # Is there leftover source following the patch?
```
barrywhart, line 336: The changes in this section fixed the initial issue (code replaced by whitespace), but not the overall issue (e.g. duplicate code inside a loop). This change may still be useful in some circumstances, as it is not specific to templates or loops, so I recommend we keep it.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -971,6 +971,11 @@ def iter_patches(self, templated_str: str) -> Iterator[FixPatch]:
             return
 
         # If we're here, the segment doesn't match the original.
+        linter_logger.debug("%s at %s: Original: [%r] Fixed: [%r]",
```
barrywhart, line 974: This new message helps see how generating patches descends the parse tree to zero in on which segment(s) were changed.

src/sqlfluff/core/linter/linted_dir.py
```diff
@@ -57,19 +57,26 @@ def check_tuples(self, by_path: bool = False):  # pragma: no cover
         """Default overload method."""
         ...
 
-    def check_tuples(self, by_path=False):
+    def check_tuples(self, by_path=False, raise_on_non_linting_violations=True):
```
barrywhart, line 60: Changed this test-related function so linter tests can include SQL files with parse errors.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -971,6 +971,13 @@ def iter_patches(self, templated_str: str) -> Iterator[FixPatch]:
             return
 
         # If we're here, the segment doesn't match the original.
+        linter_logger.debug(
```
barrywhart, line 974: I added this logger message to better understand how the linter descends the parse tree when generating patches, a process which involves determining those segments of the parse tree that were affected by fixes.

src/sqlfluff/core/templaters/base.py
```diff
@@ -59,6 +59,7 @@ class RawFileSlice(NamedTuple):
     raw: str
     slice_type: str
     source_idx: int
+    slice_subtype: Optional[str] = None
```
barrywhart, line 62: Either `"loop"` or `None`. We use this in the heuristic to detect bad patches to a templated loop body.

src/sqlfluff/core/linter/linted_file.py
```diff
@@ -263,6 +267,9 @@ def fix_string(self) -> Tuple[Any, bool]:
         # Make sure no patches overlap and divide up the source file into slices.
         # Any Template tags in the source file are off limits.
         source_only_slices = self.templated_file.source_only_slices()
+        raw_slices_in_template_loop = set(
```
tunetheweb, line 270: Could we add a comment here explaining this line and why it's here?

src/sqlfluff/core/linter/linted_file.py
```diff
@@ -314,6 +321,21 @@ def fix_string(self) -> Tuple[Any, bool]:
...
+                if len(patch_split_list) - len(patch_split_set) >= 2:
+                    # This is a heuristic. If the patch contains 2 or more
+                    # duplicate strings, assume it's a poorly computed patch
+                    # where the fixes to expanded code couldn't be applied
+                    # cleanly to the source code.
+                    linter_logger.info(
+                        "      - Skipping apparent bad patch inside template loop: %s",
+                        patch,
+                    )
+                    continue
```
tunetheweb, line 337: Seems this could be prone to false positives. Wonder what we can do to reduce that? Could we add lots of test cases?

test/fixtures/linter/autofix/ansi/016_no_fix_in_template_loops/after.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+    {% for _ in [1, 2, 3] %}      {% endfor %}
```
tunetheweb, line 2: Why have we lost the newline from before?

test/fixtures/linter/autofix/ansi/016_no_fix_in_template_loops/after.sql
```diff
@@ -0,0 +1,7 @@
...
+    10;
+
+SELECT
+    1,
+      {% for _ in [1, 2, 3] %}    2{%endfor %}
```
tunetheweb, line 7: Why has indentation changed?

tunetheweb (commented):
Some comments. Also should look at https://github.com/sqlfluff/sqlfluff/issues/1425#issuecomment-925518526

src/sqlfluff/core/linter/linted_file.py
barrywhart: Sure!

src/sqlfluff/core/linter/linted_file.py
barrywhart: Difficult to say. I think this will only trigger for loops that don't do any templating inside the loop, which, now that I think about it, is a **pretty strange thing to do**. I asked the author of the original issue what he's actually trying to do, because none of his examples generate useful looking code, e.g.:
```
SELECT 2, 2, 2, 2
```

Umm, what?

Ultimately, I think it's hard to solve this 100% robustly without getting deeper into Jinja. As Alan mentioned on the podcast, Jinja doesn't seem to support this. I have a _suspicion_ that if we look really closely at Jinja, we may be able to do some things to make this work, although they may be hacky at first. But if we find something promising, maybe we can submit an upstream PR.

Basically, if we could determine which sections of code came from different iterations of a loop, we could avoid these issues much more reliably and straightforwardly.

barrywhart: Some other random thoughts:
* Currently, SQLFluff sometimes corrupts the source code. That's a very severe bug. A code fixer should never _break_ the code.
* In the short run, I'm willing to have SQLFluff make some mistakes of omission (not changing code) or even poor formatting in order to avoid corrupting the code.
* In the long run, I expect we'll see user reports where this isn't quite behaving correctly, and we'll address those.

In a perfect world, I'd take a few engineers and a QA person and we'd go off and work on this for a few months in isolation. In practice, we have limited engineering time, and our users provide a lot of the QA. It's not perfect, but SQLFluff _does_ say it's in beta. I was wondering if we may want our next release to be a [pre-release](https://www.python.org/dev/peps/pep-0440/#pre-releases). That will prevent people from accidentally picking up unstable code, while still giving us a chance to get preliminary feedback from the community.

tunetheweb: Yeah I'm still veering towards the "if Jinja doesn't support this easily, then maybe we should gracefully bail out as say we can't fix for loop" argument I suggested on the other issue, rather than risk breaking things. I didn't 100% follow this code, but is that what this PR is basically attempting to do by checking this? In which cse maybe bailing out (even if it's more often than it used to) is the "right" thing to do.

I just never like the word "heuristics!" - to me it just suggests magic AI 😁

test/fixtures/linter/autofix/ansi/016_no_fix_in_template_loops/after.sql
barrywhart: I honestly don't know. Think we should address it in this PR or create a new one? 

I think we've gotten into an area of SQLFluff with tons of squirrelly behavior. 😛 I can tell this because of the number of "no cover" comments (I'm seeing now how these can be helpful!). I worked on a somewhat similar project in the early 2000s, where there was a 3-6 month period that, whenever you fixed a bug, you'd find two more. It felt pretty crazy at the time, but we kept adding new test cases as we found them. Eventually, it became really stable and Oracle bought us, but I was gone by then. 

https://www.oracle.com/industries/retail/products/science-platform/

src/sqlfluff/core/linter/linted_file.py
tunetheweb: > I was wondering if we may want our next release to be a [pre-release](https://www.python.org/dev/peps/pep-0440/#pre-releases). That will prevent people from accidentally picking up unstable code, while still giving us a chance to get preliminary feedback from the community.

Do we get much feedback for pre-releases? If not, then question the value of doing this. I think it makes sense for certain things (e.g. Alan is proposing a pre-release for moving dbt template to a plugin and can have some the maintainers test that), but for others it seems to just put off the release with the hope of gaining more confidence without any real knowledge that it will give us that confidence...

test/fixtures/linter/autofix/ansi/016_no_fix_in_template_loops/after.sql
tunetheweb: > Think we should address it in this PR or create a new one?

Not sure. But makes me nervous that we're proposing a heuristic-based fix, that I don't fully understand if I'm honest (no offence to your code - it's just complicated area as you say), and even the test cases are showing "funnies"...

src/sqlfluff/core/linter/linted_file.py
barrywhart: I had the same thought yesterday afternoon (don't do it!), but we have existing tests that rely on this, and if you look at those tests, they are pretty useful things to support. We have code like this at my job -- picture a "for" loop where each iteration expands to a 50-line SELECT that uses just a little templating -- a column name here, a FROM table there. So I see both sides of this -- the user side (I want!) and the engineer side (it's flaky!).

Basically, this code is looking for what appear to be duplicate strings in the expanded code, and if it sees them, discard the patches in that area.

I used the word heuristic somewhat deliberately. To me, it's a fancy word for "guess'. I leave the magic AI to the data scientists on the team I work with. 🤣

test/fixtures/linter/autofix/ansi/016_no_fix_in_template_loops/after.sql
barrywhart: >Why have we lost the newline from before?

This is no longer happening.

tunetheweb: Looks better in latest code. Resolving.

test/fixtures/linter/autofix/ansi/016_no_fix_in_template_loops/after.sql
tunetheweb: Looks better in latest code. Resolving.

barrywhart: This is no longer happening.

src/sqlfluff/core/rules/base.py
```diff
@@ -444,6 +447,40 @@ def matches_target_tuples(seg: BaseSegment, target_tuples: List[Tuple[str, str]]
...
+    def discard_results_that_span_block_boundaries(
+        lint_result: LintResult, templated_file: TemplatedFile
+    ) -> bool:
+        """Given a LintResult, remove its fixes if they span block boundaries.
+
+        Reason: Applying changes that span block boundaries may corrupt the
+        file, e.g. by moving code in or out of a template loop.
+        """
+        # Get the set of slices touched by any fixes.
+        fix_slices = set()
```
barrywhart, line 460: This is the fix for #1162.

src/sqlfluff/core/rules/base.py
```diff
@@ -444,6 +444,44 @@ def matches_target_tuples(seg: BaseSegment, target_tuples: List[Tuple[str, str]]
             return True
         return False
 
+    @staticmethod
+    def discard_fixes_that_span_block_boundaries(
```
barrywhart, line 448: This function fixes #1162.

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -339,18 +343,27 @@ def _slice_template(cls, in_str: str) -> Iterator[RawFileSlice]:
             # parts of the tag at a time.
             if elem_type.endswith("_end") or elem_type == "raw_begin":
                 block_type = block_types[elem_type]
+                block_subtype = None
                 # Handle starts and ends of blocks
                 if block_type == "block":
                     # Trim off the brackets and then the whitespace
-                    trimmed_content = str_buff[2:-2].strip()
```
barrywhart, line 345: The old code assumed Jinja block markers were always two characters at beginning (`{%`) and end (`%}`). Updating it to use regex that knows about the optional `+` or `-` for whitespace control.

src/sqlfluff/core/linter/linted_file.py
```diff
@@ -47,7 +47,7 @@ class LintedFile(NamedTuple):
     templated_file: TemplatedFile
     encoding: str
 
-    def check_tuples(self) -> List[CheckTuple]:
+    def check_tuples(self, raise_on_non_linting_violations=True) -> List[CheckTuple]:
```
barrywhart, line 50: More changes to optionally ignore parse errors in tests.

src/sqlfluff/core/linter/linted_file.py
tunetheweb: Resolving as you went a different approach in the end.

tunetheweb (approved):
Had another look at the latest changes and more comfortable with them now.

Still would be good if someone else, more familiar with this code, could review them but happy enough to approve in the meantime and can merge if that's not possible.

Good job on a complicated issue - this could be a big step forward.

PR #1440, opened 2021-09-24
Stop exception in L016 for long Jinja comments

### LINKED ISSUE #1439: Unexpected exception on "too long" jinja-comment
## Expected Behaviour

Successful run of `sqlfluff lint`

## Observed Behaviour

Exception raised on Rule L016 when having a jinja comment with more than 80 characters.

```python
  File "/usr/local/lib/python3.9/site-packages/sqlfluff/core/rules/base.py", line 302, in crawl
    res = self._eval(
  File "/usr/local/lib/python3.9/site-packages/sqlfluff/rules/L016.py", line 503, in _eval
    fixes = self._eval_line_for_breaks(this_line)
  File "/usr/local/lib/python3.9/site-packages/sqlfluff/rules/L016.py", line 266, in _eval_line_for_breaks
    raise ValueError("Is this possible?")  # pragma: no cover
ValueError: Is this possible?
== [a.sql] FAIL
L:   3 | P:  82 | L016 | Unexpected exception: Is this possible?; Could you open
                       | an issue at https://github.com/sqlfluff/sqlfluff/issues
                       | ? You can ignore this exception for now, by adding
                       | '--noqa: L016' at the end of line 3
L:   4 | P:   9 | L016 | Line is too long
```

## Steps to Reproduce

Lint this file:
```sql
SELECT *
{# comment #}
{# ........................................................................... #}
FROM table
```

## Dialect

ANSI

## Version

`sqlfluff, version 0.6.6`

## Configuration

Default config

tunetheweb:
Fixes #1439 

We don't see to have access to `place_holder` content in rules so can't fix this. So for now it just fails the rule if too long without an auto fix.

Nope

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L016.py
+++ b/src/sqlfluff/rules/L016.py
@@ -448,9 +448,9 @@ class Rule_L016(Rule_L003):
                 else:
                     break
 
-            # Does the line end in an inline comment that we can move back?
-            if this_line[-1].name == "inline_comment":
-                # Is this line JUST COMMENT (with optional preceding whitespace) if
+            # Does the line end in an inline comment or jinja placeholder that we can move back?
+            if this_line[-1].name == "inline_comment" or this_line[-1].type == "placeholder":
+                # Is this line JUST comment/placeholder (with optional preceding whitespace) if
                 # so, user will have to fix themselves.
                 if len(this_line) == 1 or all(
                     elem.name == "whitespace" or elem.is_meta for elem in this_line[:-1]
--- a/src/sqlfluff/rules/L016.py
+++ b/src/sqlfluff/rules/L016.py
@@ -454,6 +454,7 @@ class Rule_L016(Rule_L003):
                 self.logger.info("Unfixable template segment: %s", this_line[-1])
                 return LintResult(anchor=segment)
 
+            # Does the line end in an inline comment that we can move back?
             if this_line[-1].name == "inline_comment":
                 # Is this line JUST COMMENT (with optional preceding whitespace) if
                 # so, user will have to fix themselves.
--- a/src/sqlfluff/rules/L016.py
+++ b/src/sqlfluff/rules/L016.py
@@ -448,7 +448,8 @@ class Rule_L016(Rule_L003):
                 else:
                     break
 
-            # Don't even attempt to handle template placeholders as don't have content
+            # Don't even attempt to handle template placeholders as we don't have the content
+            # to move it (appears as empty space to us). It will remain as unfixable.
             if this_line[-1].type == "placeholder":
                 self.logger.info("Unfixable template segment: %s", this_line[-1])
                 return LintResult(anchor=segment)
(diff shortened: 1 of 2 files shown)


src/sqlfluff/rules/L016.py
```diff
@@ -448,6 +448,12 @@ def _eval(self, segment, raw_stack, **kwargs):
                 else:
                     break
 
+            # Don't even attempt to handle template placeholders as we don't have the content
+            # to move it (appears as empty space to us). It will remain as unfixable.
+            if this_line[-1].type == "placeholder":
```
barrywhart, line 453: What kind of lines or segments are marked as `placeholder`? I wonder why we don't have access? Maybe because it's not considered part of the text after template expansion?

test/fixtures/rules/std_rule_cases/L016.yml
```diff
@@ -129,3 +129,30 @@ test_compute_line_length_before_template_expansion_2:
...
+    {# ........................................................................... #}
+    FROM table
+  configs:
+    rules:
+      max_line_length: 80
+      L016:
+        ignore_comment_lines: False
+
+
+test_long_jina_comment_ignore:
```
barrywhart, line 147: IIUC, are we essentially testing for the presence of a bug here (i.e. still failing on Jinja comments)? Should we consider omitting this test?

barrywhart (approved):
Looks good! A question about the second test.

Is there a corresponding issue for this? What was the exception -- is this the one that was something like, "Is this possible?"

src/sqlfluff/rules/L016.py
tunetheweb: OK that comment isn't entirely accurate so will update it.

`placeholder` are templated areas that don't end up in the final SQL, including Jinja comments (`{#...#}`), assignments (`{% set barry=xxx %}`) or even logic (`{% for barry in pollard %}...{% endfor %}`). If they end up as the cause for long lines then we've limited options anyway.

Yes, in theory, we should be able to fix this:

```sql
SELECT *
FROM {# really long comment #}
```

to this:

```sql
SELECT *
 {# really long comment #}
FROM
```

And similar for `{% set... %}` commands, but we can't fix this:

```sql
{% for... %}
SELECT *
FROM really_long_table_name {% endfor %}
```

And yes the reason we don't have access to the segment, is it's because it's not considered part of the test after template expansion and this rule (mostly) works on the post templated file. We do have access to the source string (in the `.source_str` variable), which is used in the one area we look at pre-templated, but we don't (currently) have access to the original segment. We could add it to handle the above case, but honestly don't think it's worth the complexity.

test/fixtures/rules/std_rule_cases/L016.yml
tunetheweb: Kind of. However at the moment we are assuming `placeholders` are not treated as comments. And in fact there can be problems if it is assumed to be (the `{% endfor %}` example above. So I still think best to have this test for now.

Plus if we ever add the logic to treat Jinja comments the same as SQL comments, then we'd need a test case and here we have a perfect one already created - just need to update `fail_str` to `pass_str`.

tunetheweb:
Replied to your comments @barrywhart . Know you've approved this but will hold off merging in case you disagree with my replies or have any further comments. Let me know either way!

barrywhart (approved):
Looks good -- thanks for the explanations!

PR #1442, opened 2021-09-24
Add initial support for TSQL Azure Synapse Analytics dialect (tsql_asa)

jpers36:

Adding support for the TSQL dialect specific to Azure Synapse Analytics.  New dialect file, new tests, new keyword list.

Additionally, added the SWITCH and COLUMNSTORE keywords as unreserved keywords to the main TSQL dialect, and updated the main TSQL dialect to pull in its unreserved keywords.

The main TSQL dialect will now recognize SWITCH and COLUMNSTORE as unreserved keywords.


tunetheweb:
Can't say I know enough about the two versions but would it be so terrible to have tsql dialect support both, rather than adding another dialect? Even if that means tsql technically supports syntax that is invalid in standard tsql?

Or are they different enough you think it should be a separate dialog?

jpers36:
@tunetheweb The two are definitely different. When describing it to people familiar with vanilla T-SQL I say it's only a 5% difference but the differences are critical.  T-SQL allows things that Azure Synapse Analytics doesn't and vice versa.  The most important question is whether there's any place where the two dialect definitions would be irreconcilable rather than just an extension on one side or the other, and I don't have a quick answer for that. I'll see if I can think of a concrete example.

jpers36:
I asked a question on Slack that could bring some clarity.

barrywhart:
Since the two dialects overlap a lot, but each has unique bits, should we create a new "core TSQL" dialect, then inherit both TSQL and TSQL Azure Synapse Analytics from that? That seems like it may be cleaner and easier to maintain than having TSQL Azure Synapse Analytics inherit from TSQL. This mimics the design used for the ANSI and other dialects. In general, I wouldn't want to see a subdialect _removing_ features of its parent. 🤯

Thoughts?

tunetheweb:
That's creating three T-SQL dialects - I was trying to reduce the number we need to manage not increase!

We already support some syntax in the ANSI version that is inherited by the other dialects, even when it is not part of all those dialects.

As I say I guess it depends how far apart they are, but I'd still be temped to manage them both under `tsql` dialect for now. Nothing to stop us pulling them apart again at a later date if that proves unwieldy...

barrywhart:
More dialects is not good, but child dialect removing lots of stuff from the parent is also not good. We don't have to do anything now, but I suggest we keep an eye on how much deletion there is. I'm picturing a scenario where every time you add something to TSQL, you have to go and add corresponding code so it doesn't leak into the other. I would be hesitant to contribute to either of those dialects for risk of accidentally breaking something.

jpers36:
Created #1478 as an alternate PR to this one.

tunetheweb:
Closing this one for now in preference to #1478

PR #1443, opened 2021-09-25
Issue 1438: sqlfluff fix keeps adding then removing whitespace

### LINKED ISSUE #1438: `sqlfluff fix` keeps adding then removing whitespace
## Expected Behaviour

After a finite number of iterations (ideally 1 :slightly_smiling_face:), `sqlfluff fix` should eventually converge on an output that it is happy with.

## Observed Behaviour

`sqlfluff fix` keeps adding whitespace before some `for` / `endfor` Jinja tags, and then removing it again upon a successive invocation (which makes it challenging to use as, e.g., a pre-commit hook).

## Steps to Reproduce

```console
$ echo '{% set elements = "foo" %}

SELECT
    {% for elem in elements %}
    {{ elem }},
    {% endfor %}

    {% for elem in elements %}
    {{ elem }}+
    {% endfor %}
    0
' > test_0.sql

$ for n in $(seq 5) ; do cp test_$((n-1)).sql test_$n.sql ; sqlfluff fix --force test_$n.sql ; done
(sqlfluff output omitted)
$ diff -u test_4.sql test_5.sql
--- test_4.sql	2021-09-23 16:36:30.696291287 +0000
+++ test_5.sql	2021-09-23 16:36:31.232985009 +0000
@@ -3,10 +3,10 @@
 SELECT
     {% for elem in elements %}
     {{ elem }},
-    {% endfor %}
+{% endfor %}
 
-        {% for elem in elements %}
+    {% for elem in elements %}
     {{ elem }} +
-    {% endfor %}
+{% endfor %}
     0

```

## Dialect

None specified

## Version
SQLFluff a50454c963ee8d942d2fa60ab2b48eca31ce5175, Python 3.9.7

## Configuration

None

barrywhart:
Fixes #1438

Modifies L001 to check the raw source code following whitespace at the end of a line. There may be template code following it, in which case we want to keep the whitespace because it's not really "trailing whitespace". This avoids an endless cycle of "fixes" between L001 and L039.

n/a

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -589,7 +589,7 @@ class Linter:
         config.process_raw_file_for_config(in_str)
         rendered = self.render_string(in_str, fname, config, encoding)
         if templated_file is not None:
-            templated_file[:] = rendered
+            templated_file[:] = [rendered.templated_file]
         violations += rendered.templater_violations
 
         # Dispatch the output for the parse header
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -573,6 +573,7 @@ class Linter:
         recurse: bool = True,
         config: Optional[FluffConfig] = None,
         encoding: str = "utf-8",
+        templated_file: Optional[List[TemplatedFile]] = None
     ) -> ParsedString:
         """Parse a string."""
         violations: List[SQLBaseError] = []
@@ -587,6 +588,8 @@ class Linter:
         # Scan the raw file for config commands.
         config.process_raw_file_for_config(in_str)
         rendered = self.render_string(in_str, fname, config, encoding)
+        if templated_file is not None:
+            templated_file[:] = rendered
         violations += rendered.templater_violations
 
         # Dispatch the output for the parse header
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -573,7 +573,6 @@ class Linter:
         recurse: bool = True,
         config: Optional[FluffConfig] = None,
         encoding: str = "utf-8",
-        templated_file: Optional[List[TemplatedFile]] = None,
     ) -> ParsedString:
         """Parse a string."""
         violations: List[SQLBaseError] = []
@@ -588,8 +587,6 @@ class Linter:
         # Scan the raw file for config commands.
         config.process_raw_file_for_config(in_str)
         rendered = self.render_string(in_str, fname, config, encoding)
-        if templated_file is not None:
-            templated_file[:] = [rendered.templated_file]
         violations += rendered.templater_violations
 
         # Dispatch the output for the parse header
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -573,7 +573,7 @@ class Linter:
         recurse: bool = True,
         config: Optional[FluffConfig] = None,
         encoding: str = "utf-8",
-        templated_file: Optional[List[TemplatedFile]] = None
+        templated_file: Optional[List[TemplatedFile]] = None,
     ) -> ParsedString:
         """Parse a string."""
         violations: List[SQLBaseError] = []
--- a/src/sqlfluff/core/templaters/jinja.py
+++ b/src/sqlfluff/core/templaters/jinja.py
@@ -354,15 +354,17 @@ class JinjaTemplater(PythonTemplater):
                         trimmed_content = str_buff[
                             len(m_open.group(0)) : -len(m_close.group(0))
                         ]
-                    if trimmed_content.startswith("end"):
+                    first_keyword = trimmed_content.split()[0]
+                    if first_keyword.startswith("end"):
                         block_type = "block_end"
-                    elif trimmed_content.startswith("el"):
-                        # else, elif
+                    elif first_keyword in ("else", "elif"):
                         block_type = "block_mid"
-                    else:
+                    elif first_keyword in ("for", "if"):
                         block_type = "block_start"
                         if trimmed_content.split()[0] == "for":
                             block_subtype = "loop"
+                    else:
+                        block_type = "directive"
                 yield RawFileSlice(str_buff, block_type, idx, block_subtype)
                 idx += len(str_buff)
                 str_buff = ""
--- a/src/sqlfluff/core/templaters/jinja.py
+++ b/src/sqlfluff/core/templaters/jinja.py
@@ -354,17 +354,15 @@ class JinjaTemplater(PythonTemplater):
                         trimmed_content = str_buff[
                             len(m_open.group(0)) : -len(m_close.group(0))
                         ]
-                    first_keyword = trimmed_content.split()[0]
-                    if first_keyword.startswith("end"):
+                    if trimmed_content.startswith("end"):
                         block_type = "block_end"
-                    elif first_keyword in ("else", "elif"):
+                    elif trimmed_content.startswith("el"):
+                        # else, elif
                         block_type = "block_mid"
-                    elif first_keyword in ("for", "if"):
+                    else:
                         block_type = "block_start"
                         if trimmed_content.split()[0] == "for":
                             block_subtype = "loop"
-                    else:
-                        block_type = "directive"
                 yield RawFileSlice(str_buff, block_type, idx, block_subtype)
                 idx += len(str_buff)
                 str_buff = ""
--- a/src/sqlfluff/rules/L001.py
+++ b/src/sqlfluff/rules/L001.py
@@ -56,6 +56,11 @@ class Rule_L001(BaseRule):
             next_raw_slice = templated_file.raw_slices_spanning_source_slice(
                 slice(last_deletion_slice.stop, last_deletion_slice.stop)
             )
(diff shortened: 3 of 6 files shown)


src/sqlfluff/core/linter/linter.py
```diff
@@ -573,6 +573,7 @@ def parse_string(
         recurse: bool = True,
         config: Optional[FluffConfig] = None,
         encoding: str = "utf-8",
+        templated_file: Optional[List[TemplatedFile]] = None
```
barrywhart, line 576: Optional output parameter: Receives the `TemplatedFile` returned by rendering the raw SQL

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -354,15 +354,17 @@ def _slice_template(cls, in_str: str) -> Iterator[RawFileSlice]:
                         trimmed_content = str_buff[
                             len(m_open.group(0)) : -len(m_close.group(0))
                         ]
-                    if trimmed_content.startswith("end"):
```
barrywhart, line 357: Not related to the PR, but tidies up block types. "directive" means a Jinja command such as `% set elements = "foo" %}`. It is a standalone command, so not a block start, middle, or end.

src/sqlfluff/testing/rules.py
```diff
@@ -87,11 +87,14 @@ def assert_rule_pass_in_sql(code, sql, configs=None):
     # Configs allows overrides if we want to use them.
     cfg = FluffConfig(configs=configs)
     r = get_rule_from_set(code, config=cfg)
-    parsed = Linter(config=cfg).parse_string(sql)
+    templated_file = []
+    parsed = Linter(config=cfg).parse_string(sql, templated_file=templated_file)
```
barrywhart, line 91: Previously, rule tests using `pass_str` didn't receive a `templated_file` parameter, which is now needed by L001.

test/fixtures/linter/autofix/bigquery/004_templating/after.sql
```diff
@@ -66,7 +66,7 @@ imputed_effect_sizes AS (
         , COALESCE(IF(IS_NAN(o.{{metric}}_{{action}}), 0, o.{{metric}}_{{action}}), 0) AS {{metric}}_{{action}}
         , COALESCE(IF(IS_NAN(n.{{metric}}_{{action}}), 0, n.{{metric}}_{{action}}), 0) AS new_{{metric}}_{{action}}
         , n.campaign_count_{{action}}
-    {% endfor %}
+        {% endfor %}
```
barrywhart, line 69: This test uses L001, and the output changed -- now L001 leaves this line alone rather than messing up the indentation.

src/sqlfluff/rules/L001.py
```diff
@@ -45,7 +45,20 @@ def _eval(self, segment, raw_stack, **kwargs):
...
+            # Check the raw source (before template expansion) immediately
+            # following the whitespace we want to delete. Often, what looks
+            # like trailing whitespace in rendered SQL is actually a line like:
+            # "    {% for elem in elements %}\n", in which case the code is
+            # fine -- it's not trailing whitespace from a source code
+            # perspective.
+            next_raw_slice = templated_file.raw_slices_spanning_source_slice(
+                slice(last_deletion_slice.stop, last_deletion_slice.stop)
             )
+            if next_raw_slice[0].slice_type == "literal":
```
tunetheweb, line 59: This appears to work but took a long while for me to understand why!

So are you basically getting all the raw slices which are either templated (`block_start`, `block_end`...etc.) or, if not templated, `literal`? So a "slice" is basically a "templated slice" of either templated language or raw SQL?

If so, it might be nice to explain this:

```suggestion
            # If there are no templated slices, and only raw SQL ("literal"), then safe to delete
            if next_raw_slice[0].slice_type == "literal":
```

test/fixtures/linter/autofix/bigquery/004_templating/after.sql
tunetheweb: Good improvement!

tunetheweb (changes requested):

src/sqlfluff/core/linter/linter.py
```diff
@@ -587,6 +588,8 @@ def parse_string(
         # Scan the raw file for config commands.
         config.process_raw_file_for_config(in_str)
         rendered = self.render_string(in_str, fname, config, encoding)
+        if templated_file is not None:
+            templated_file[:] = [rendered.templated_file]
```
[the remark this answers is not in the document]
barrywhart: Python parameters are always "pass by value". So to implement "output parameters", you have to pass in a collection and update the collection instead. If we did this, it would assign `templated_file` to point to a new, different list, _not_ updating the one passed in by the caller:
```
templated_file = [rendered.templated_file]
```

https://www.learnbyexample.org/python-list-slicing/#modify-multiple-list-values

I considered updating the function to return two values rather than use this trickery, but it's a public `Linter` function, so I thought it inappropriate to change the API to return something very few people care about.

test/fixtures/rules/std_rule_cases/L001.yml
```diff
@@ -8,3 +8,15 @@ test_fail_trailing_whitespace:
...
+test_pass_trailing_whitespace_before_template_code:
+  pass_str: |
+    SELECT
+        {% for elem in ["a", "b"] %}
+        {{ elem }},
+        {% endfor %}
+        0
+  configs:
+    templater:
+      jinja: {}
```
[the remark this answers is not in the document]
barrywhart: Oh cool -- I'll change that! I thought at one point it was not expanding the template.

src/sqlfluff/rules/L001.py
barrywhart: Close! The way I'm constructing the slice I pass to the function, I know it'll only return one slice, and that is the next slice of raw source (pre-template) immediately following the white space it's considering deleting.

barrywhart: I'll add a comment similar to the above. Note there's an existing comment a few lines above which helps explain this as well.

barrywhart:
This is ready for another review.

src/sqlfluff/core/linter/linter.py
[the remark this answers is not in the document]
tunetheweb: Oh I see now. And that is why we're passing in `Optional[List[TemplatedFile]]` rather than just `Optional[TemplatedFile]` I presume?

But not sure I like it to be honest. Seems a bit dishonest to pass input parameters and then alter them with this trickery. At the very least should add a comment to explain this.

However I can also understand hesitancy to alter `parse_string` (even though technically you have already in this PR, with the optional input param - but at least that's optional). What do you think about creating a `parse_string_and_template` function (perhaps with a better name!) that does this properly by returning two values?

Then `parse_string` could either be left along (there's not much code to duplicate to be honest!) or, perhaps better, it could call the `parse_string_and_template` function with blank template and strip out the returned template, so the API of `parse_string` remains the same?

src/sqlfluff/rules/L001.py
tunetheweb: Thanks. It's the `literal` that through me for a long time. I was wondering why you weren't checking for `"\n"` rather than any literal. But get it now - `literal` here means non-templated code, rather than any particular literal value.

src/sqlfluff/core/linter/linter.py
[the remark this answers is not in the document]
barrywhart: I'm a little hesitant about adding another public function to this user-facing class. I'll see if there's a way to add it as a standalone function. It'll still be "public", but not "staring users in the face", if that makes sense.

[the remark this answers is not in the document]
barrywhart: Actually, the `Linter` class already has a large number of "public" functions that are intended mainly for internal use, so I'm okay adding another function there.

src/sqlfluff/rules/L001.py
barrywhart: Yes, it's literal in the template sense, not in the SQL sense. If you think about what Jinja or other templating engines do, they execute the control structures, and other stuff just passes through. If you mentally translate the template into the code you'd write by hand if you _didn't_ have Jinja, "other stuff" would be a `print()` statement, e.g. the string "0" in a template becomes:
```
print('0')
```

when rendering the template.

src/sqlfluff/testing/rules.py
```diff
@@ -87,11 +87,15 @@ def assert_rule_pass_in_sql(code, sql, configs=None):
     # Configs allows overrides if we want to use them.
     cfg = FluffConfig(configs=configs)
     r = get_rule_from_set(code, config=cfg)
-    parsed = Linter(config=cfg).parse_string(sql)
+    linter = Linter(config=cfg)
```
barrywhart, line 90: Changed this code so we can get and pass along the `TemplatedFile` instance to the `crawl()` function.

It turns out we don't need to change the public interface to `Linter` at all. The existing API works, although some of these secondary functions like `render_string()` have a poorly designed API (seems like I should be able to pass it just the `sql` string, but it requires those 3 other parameters).

tunetheweb (approved):
LGTM!

As you say could clean up `render_string` API by making arguments optional but maybe not one for this PR.

Issue #1446, opened 2021-09-26
EXASOL_FS: Unable to lex characters


sti0:
_If this is a parsing or linting issue, please include a minimal SQL example which reproduces the issue, along with the `sqlfluff parse` output, `sqlfluff lint` output and `sqlfluff fix` output when relevant._

## Expected Behaviour
Ignore the text in the comment section. Note that the ``@lang`` and `$` which are passed with parameter 2 and 3 are no problem. Only the code within the parameter `json_str`.

## Observed Behaviour
I received the following message
```
L:  14 | P:  17 |  LXR | Unable to lex characters: '@lang'
L:  14 | P:  33 |  LXR | Unable to lex characters: '$'
L:  14 | P:  49 |  LXR | Unable to lex characters: '@lang'
L:  14 | P:  65 |  LXR | Unable to lex characters: '$'
```

## Steps to Reproduce
Lint this SQL script by command line (`sqlfluff lint --dialect=exasol_fs`)

```
CREATE
OR REPLACE PYTHON3 SCALAR SCRIPT MYSCHEMA.MYPYTHONSCRIPT(
    JSON_STR VARCHAR(2000000),
    LANGUAGE_KEY VARCHAR(50),
    TXT_KEY VARCHAR(50)
) EMITS (
    X VARCHAR(2000000)
)
AS
"""
/*====================================================================
    e.g.:
    SELECT MYSCHEMA.MYPYTHONSCRIPT(
            '[{"@lang":"de-DE","$":"Krztxt"}, {"@lang":"en-GB","$":"Shrttxt"}]',
            '@lang',
            '$'
        );
 ====================================================================*/
"""
def run (ctx):
    pass
/
```

## Dialect
exasol_fs

## Version
Include the output of `sqlfluff --version` along with your Python version
sqlfluff, version 0.6.6
Python 3.8.12

## Configuration

default config

tunetheweb:
Question for you, should the function not end in a "/" sign as per https://docs.exasol.com/sql/create_script.htm?

That would allows us to be liberal in what we accept in script values rather than adding each symbol just in case it's used.

But it is easy enough to add these if that's not the case.

tunetheweb:
> Note that the @lang and $ which are passed with parameter 2 and 3 are no problem. Only the code within the parameter json_str.

Actually that's not true - it just is incorrectly reporting everything as part of line 14 even though it shouldn't really:

```
[...]

PR #1447, opened 2021-09-26
Add @ and $ symbols to Exasol to avoid lexing errors

tunetheweb:
Fixes: #1446

A better solution is probably the one followed by BigQuery where we swallow the function up in the Parser, rather than current process of parsing each component of the function as a Raw (which then leads to these types of errors when an unrecognised symbol is used).

But don't know enough about exasol to make that change, the example function doesn't seem to follow the [Exasol spec to make this easy](https://docs.exasol.com/sql/create_script.htm) (question raised with reporter), and this change is relatively simple.

### Pull Request checklist

  - Other.


barrywhart (approved):
Looks good!

PR #1453, opened 2021-09-26
Add SQLite dialect

### LINKED ISSUE #1452: support SQLite dialect
SQLite is the the [most used](https://sqlite.org/mostdeployed.html) database engine in the world. I believe it would be a great addition to this project to support its dialect.

Following the discussion on #1450, it should be similar to `Postgres` with one difference being the `REGEXP` keyword.

tunetheweb:
Fixes #1452 

I sorted dialects by alphabetical order on the documentation page.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_sqlite.py
```diff
@@ -0,0 +1,24 @@
...
+"""
+
+from sqlfluff.core.parser import (
+    OneOf,
+    Ref,
+)
+
+from sqlfluff.core.dialects import load_raw_dialect
+
+postgres_dialect = load_raw_dialect("postgres")
```
WittierDinosaur, line 13: are we sure about adding this dependency? SQLite, as far as I'm aware, isn't based on Postgres.

tunetheweb: I choose to clone Postgres based on the original comment in #1452 and also for this reason:

https://en.wikipedia.org/wiki/SQLite
> SQLite generally follows PostgreSQL syntax.

But do note, that there's no mention of Postgres here: https://sqlite.com/lang.html

I'm easy either way. @tlgs what's your thoughts?

WittierDinosaur: It makes sense to a degree - I just want to avoid a repeat of the Snowflake issue - where developing on Postgres broke tests in Snowflake. Risk is probably lower here though, as SQLite isn't anything like as heavyweight as the other languages

tlgs: @tunetheweb I'm really not familiar enough with the world of SQL dialects to be comfortable making suggestions here. I'm just a fan of sqlfluff that happens to use SQLite. I'm sure are more like me that would benefit from adding support for it as a dialect.

That said, SQLite is not a traditional database engine. For one, it [doesn't use static, rigid typing](https://sqlite.com/datatype3.html).
I imagine that this and other quirks of the engine could warrant different linting rules.

docs/source/dialects.rst
```diff
@@ -52,54 +52,92 @@ to be a representation of vanilla SQL before any other project adds their
...
+Hive
+------
+
+The dialect for `Hive`_.
+
+.. _`Hive`: https://hive.apache.org/
+
+.. _mysql_dialect_ref:
+
+MySql
```
barrywhart, line 85:
```suggestion
MySQL
```

src/sqlfluff/dialects/dialect_sqlite.py
barrywhart: At this time, SQLFluff is very unaware of types. It parses them, but the linting rules are more around formatting, aliasing, joins, fairly high-level things. That may change as the project evolves.

I think whatever decision we make now may need to be revisited in a year or two, so I wouldn't stress _too_ much about this.

In a way, I think the tests are more important than the dialect code itself. In other words, with a solid set of tests for each dialect, we can refactor as needed, changing the "inheritance" structure, etc. To be clear, I don't know if our tests are good enough yet, just pointing out that effort in developing good tests is well worth it.

barrywhart (approved):
Looks good. One small suggestion about capitalization of MySQL (this was pre-existing text).

docs/source/dialects.rst
tunetheweb: Fixed!

src/sqlfluff/dialects/dialect_sqlite.py
tunetheweb: The example query still parses when using ANSI so happy to switch to that instead of Postgres. Resolving.

PR #1449, opened 2021-09-26
Move Exasol test statements fixtures from Python to SQL files

tunetheweb:
### Brief summary of the change made
Fixes #1117

Tests which aren't complete statements (e.g. tests against individual segments) have been left in Python.

This makes it easier to add new Exasol SQL statements and also spot changes affecting old statements (as we'll see a YML change).

### Are there any other side effects of this change that we should be aware of?
Nope.

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/parser` (note YML files can be auto generated with `python test/generate_parse_fixture_yml.py` or by running `tox` locally).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #1457, opened 2021-09-27
Preceding spelling

jpers36:
Fix preceeding in one place introduced in a separate PR concurrently with the original spelling fix.


tunetheweb:
Thanks - but taking care of this in #1449

Issue #1472, opened 2021-09-28
Add file_path to FileSegment


sti0:
It would be nice to add a property `file_path` to the FileSegment to indicate where the statement comes from.

This might be useful for some custom rules: E.g. we store all our `view` scripts in one folder, `tables` in another. In such case I could write a rule that make sure that views statements are saved in the right location.

sti0:
solved by #1473

PR #1473, opened 2021-09-28
Adding a new BaseFileSegment class for FileSegments to inherit from

sti0:
This adds a new property `file_path` to the BaseSegment which is filled in the RootSegment e.g. FileSegment. This solves the request in  #1472.

This PR is made to discuss the implementation, maybe someone with more knowledge about the architecture has a suggestion to improve this. Tests will be added.
...

...

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -181,7 +181,7 @@ class Linter:
         tokens: Sequence[BaseSegment],
         config: FluffConfig,
         recurse: bool = True,
-        fname: str = None,
+        fname: Optional[str] = None,
     ) -> Tuple[Optional[BaseSegment], List[SQLParseError]]:
         parser = Parser(config=config)
         violations = []
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -178,13 +178,18 @@ class Linter:
 
     @staticmethod
     def _parse_tokens(
-        tokens: Sequence[BaseSegment], config: FluffConfig, recurse: bool = True
+        tokens: Sequence[BaseSegment],
+        config: FluffConfig,
+        recurse: bool = True,
+        fname: str = None,
     ) -> Tuple[Optional[BaseSegment], List[SQLParseError]]:
         parser = Parser(config=config)
         violations = []
         # Parse the file and log any problems
         try:
-            parsed: Optional[BaseSegment] = parser.parse(tokens, recurse=recurse)
+            parsed: Optional[BaseSegment] = parser.parse(
+                tokens, recurse=recurse, fname=fname
+            )
         except SQLParseError as err:
             linter_logger.info("PARSING FAILED! : %s", err)
             violations.append(err)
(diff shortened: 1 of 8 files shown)


src/sqlfluff/core/linter/linter.py
```diff
@@ -178,13 +178,18 @@ def _lex_templated_file(
 
     @staticmethod
     def _parse_tokens(
-        tokens: Sequence[BaseSegment], config: FluffConfig, recurse: bool = True
+        tokens: Sequence[BaseSegment],
+        config: FluffConfig,
+        recurse: bool = True,
+        fname: str = None,
```
tunetheweb, line 184: Should this be:

```suggestion
        fname: Optional[str] = None,
```

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -79,11 +79,20 @@ class (which still inherits from this one).
     # What other kwargs need to be copied when applying fixes.
     additional_kwargs: List[str] = []
 
-    def __init__(self, segments, pos_marker=None, name: Optional[str] = None):
+    def __init__(
+        self,
+        segments,
+        pos_marker=None,
+        name: Optional[str] = None,
+        fname: str = None,
```
tunetheweb, line 87: Should this be:

```suggestion
        fname: Optional[str] = None,
```

src/sqlfluff/dialects/dialect_exasol_fs.py
```diff
@@ -122,6 +122,11 @@ class FileSegment(BaseSegment):
         allow_trailing=True,
     )
 
+    @property
+    def file_path(self):
+        """File path of a parsed SQL file."""
+        return self._file_path
+
```
tunetheweb, line 129: Won't this automatically be inherited from `ansi`?

```suggestion
```

tunetheweb (commented):
The approach looks like it'll work but seems like we';d be adding a new attribute to every segment, even though we'd only be using it on one of them (`FileSegment`).

Or would it be set for all Segments? And, if it's not, would some expect this to be set if the attribute exists?

I wonder would it be better to define a `FileSegment` type in `src/sqlfluff/core/parser/segments` with just this attribute? Or would you have to basically make all the same code changes anyway to ensure it was passed through to set it?

Also have some other minor comments below:

src/sqlfluff/dialects/dialect_exasol_fs.py
tunetheweb: Ah no it won't because you explicitly overwrite it. Can we set it to copy from ANSI version so not repeating the code?

sti0: I think it needs to be overwritten because scripts not delimited by `;` Instead they are delimited by `/`. But maybe there is another way?

Currently I added a new class `FileSegment` as you suggested.

sti0:
I think adding this to the FileSegment is the right place, because ...it's a file 😃 Maybe this could even helpful for someone in other context.

tunetheweb (approved):
LGTM!

Issue #1208, opened 2021-07-15
does not scale with large SQL files well


splickz:
-- If this is a parsing or linting issue, please include a minimal SQL example which reproduces the issue, along with the `sqlfluff parse` output, `sqlfluff lint` output and `sqlfluff fix` output when relevant. Finally, please add any labels that are related to the issue before submitting so that it can be addressed easily! --

## Expected Behaviour
be able to parse SQL file with thousands of lines

## Observed Behaviour
seems to take over half an hour + for large sql files (I'd imagine a reasonable small change to algorithm could make this tool scale much better hopefully?)

## Steps to Reproduce
try lint a  8000 line file with just `select * from a;` on each line
![image](https://user-images.githubusercontent.com/10183003/125725135-877e850d-c89c-4970-97bd-8c2638f71e46.png)

## Version
Include the output of `sqlfluff --version` along with your Python version
0.6.0, 3.9.5
## Configuration
```
Include your SQLFluff configuration here
```
default

barrywhart:
@splickz: I'm a bit worried that your actual file may be different enough from simply repeating 8000 lines of `select * from a` that fixing this issue won't help with the performance issue you're seeing. We can look into this, just want to note that this is a risk.

barrywhart:
#962 has another large query with slow parse times. IIIUC, it is a more complex query, rather than a sequence of simple ones.

barrywhart:
Changed from "bug" to "enhancement".

barrywhart:
@WittierDinosaur, @alanmcruickshank, @tunetheweb: Think we can close this issue with the work in PR #1744?

tunetheweb:
Yup!

PR #1475, opened 2021-09-29
Modest parser performance improvement

NathanHowell:

When attempting to parse a large legacy file, thousands of lines, I noticed sqlfluff was taking quite a while to complete (#1208). While most of the time is spent dealing with matchers, I found a couple improvements that help a bit (but not a lot):

* `get_line_pos_of_char_pos` was iterating when a binary search would work
* `iter_indices` searched for matches twice: once via `in` and a second with `index`

The parser should (always?) be faster. Here's one example using a more typical query:

#### Before

```
==== overall timings ====
Clock time:     0.00
=== templating ===
avg:            0.01
=== lexing ===
cnt:               1 sum:            0.06
min:            0.06 max:            0.06
avg:            0.06
=== parsing ===
cnt:               1 sum:            0.48
min:            0.48 max:            0.48
avg:            0.48
```

#### After

```
==== overall timings ====
Clock time:     0.00
=== templating ===
avg:            0.01
=== lexing ===
cnt:               1 sum:            0.04
min:            0.04 max:            0.04
avg:            0.04
=== parsing ===
cnt:               1 sum:            0.38
min:            0.38 max:            0.38
avg:            0.38
```


src/sqlfluff/core/templaters/base.py
```diff
@@ -168,18 +169,15 @@ def get_line_pos_of_char_pos(
...
             ref_str = self._templated_newlines
 
-        while nl_idx + 1 < len(ref_str) and ref_str[nl_idx + 1] < char_pos:
-            nl_idx += 1
+        nl_idx = bisect_right(ref_str, char_pos)
 
-        if nl_idx >= 0:
-            return nl_idx + 2, char_pos - ref_str[nl_idx]
+        if nl_idx > 0:
+            return nl_idx + 1, char_pos - ref_str[nl_idx - 1]
```
tunetheweb, line 180: The first change seems fine, but this change is causing some of the tests to fail so they musn't be equivalent.

tunetheweb (changes requested):

src/sqlfluff/core/templaters/base.py
NathanHowell: interesting, the lexer/parser tests passed (at least locally).. looks like this is failing in the linter. I'll take a peek Friday or Monday.

tunetheweb: Here's what I run locally:

```
tox -e generate-fixture-yml,cov-init,py38,cov-report,linting
```

Quite slow, but covers everything the CI does. Or just update branch and let GitHub do the hard work 😁 - which basically does the above, plus more versions of Python, plus dbt tests and also some doc linting.

NathanHowell: @tunetheweb can you kick the CI workflow? tests are passing locally now

tunetheweb: Running now!

tunetheweb: All tests passing. Late here now so will give it a once over over the weekend unless someone else gets in there first.

Not seeing any real noticeable improvement on the CI run time. Was hoping you were being modest about the improvements! 😁 Still every little helps!

NathanHowell: it's really query dependent, some of the larger ones we have saw a 10-20% improvement. it shouldn't make anything worse though. most queries spend so much time in the matcher lookahead logic that other minor perf bumps barely register.

barrywhart: Sounds great!

Running the automated test suite is often a poor indicator of performance because tests involve lots of tiny examples, lots of setup and teardown, etc.

tunetheweb (approved):

PR #1482, opened 2021-10-01
Disable rule L033 for dialects that do not support it (e.g. Exasol, Postgres)

### LINKED ISSUE #1481: EXASOL: L033 false fixing
In the [Exasol](https://docs.exasol.com/sql/table_operators.htm) dialect `UNION DISTINCT` doesn't exists. Is it possible to exclude this rule for the whole dialect or does everybody need to exclude this explicit?

## Expected Behaviour
Rule should be ignored for Exasol

## Observed Behaviour
A simple `UNION` was fixed into `UNION DISTINCT`

## Steps to Reproduce
```
SELECT c FROM x 
UNION
SELECT c FROM y
```
## Dialect
exasol

## Version
sqlfluff, version 0.6.6
Python 3.8.12

## Configuration
default

tunetheweb:
Fixes #1481 

Gave tests proper name and added uppercase/lowercase support while at it (based on case of `UNION` keyword already set.)

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L033.py
+++ b/src/sqlfluff/rules/L033.py
@@ -26,13 +26,19 @@ class Rule_L033(BaseRule):
 
     """
 
-    def _eval(self, segment, raw_stack, **kwargs):
+    def _eval(self, segment, raw_stack, dialect, **kwargs):
         """Look for UNION keyword not immediately followed by DISTINCT or ALL.
 
         Note that UNION DISTINCT is valid, rule only applies to bare UNION.
         The function does this by looking for a segment of type set_operator
         which has a UNION but no DISTINCT or ALL.
+
+        Note some dialects (e.g. Exasol) do not have concept of UNION DISTINCT,
+        so rule is ignored for them.
         """
+        if dialect.name == "exasol":
+            return LintResult()
+
         if segment.is_type("set_operator"):
             if "UNION" in segment.raw.upper() and not (
                 "ALL" in segment.raw.upper() or "DISTINCT" in segment.raw.upper()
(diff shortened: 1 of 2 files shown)


WittierDinosaur:
Postgres also doesn't support this syntax. This rule seems very dialect specific - is it worth switching it round so it's off unless it's a supported syntax?

tunetheweb:
> Postgres also doesn't support this syntax. This rule seems very dialect specific - is it worth switching it round so it's off unless it's a supported syntax?
> 
> ```
> if dialect in [my_dialects]:
>     do rule
> ```

A bit on the fence. I originally though that if in ANSI (difficult to tell as you have to pay 😞) then default should be to support, but a quick check shows the majority of our dialects do NOT support this. So went with your suggestion.

Also fixed a capitalisation issue while I was in there.

sti0 (approved):
Looks got for me.

src/sqlfluff/rules/L033.py
```diff
@@ -26,15 +26,38 @@ class Rule_L033(BaseRule):
...
+
+        Note only some dialects have concept of UNION DISTINCT, so rule is only
+        applied to dialects that are known to support this syntax.
         """
+        if dialect.name not in ["ansi", "bigquery", "hive", "mysql"]:
+            return LintResult()
+
         if segment.is_type("set_operator"):
-            if "UNION" in segment.raw.upper() and not (
+            if "union" in segment.raw and not (
```
barrywhart, line 43: Is the `.raw` property guaranteed to be lower case? If you're removing `.upper()` here, could it also be removed from line 44 below?

barrywhart (approved):
One question about upper/lower. Looks good!

src/sqlfluff/rules/L033.py
tunetheweb: It's the raw casing - so whatever is in the source.

So I'm using this to say "it's it's lowercase in source, then enter the `distinct` as lowercase." For Line 44 I don't really care about the case, so keep it as case-less matching.

This will match in the unlikely event of:

```sql
SELECT 1
union ALL
SELECT 2
```

Issue #426, opened 2020-08-27
Found unparsable section: ' notnull'


nolanbconaway:
This query is definitely valid postgres:

```sql
select * from t where c notnull
```

Alas, fluff disagrees!

```sh
$ echo 'select * from t where c notnull' | sqlfluff lint - --dialect postgres
== [stdin] FAIL
L:   1 | P:  24 |  PRS | Found unparsable section: ' notnull'
```

For the record:

```sh
$ sqlfluff version    
0.3.5
```

alanmcruickshank:
Interesting! I wasn't familiar with that syntax.

Yes it is valid in postgres (https://www.postgresql.org/docs/9.2/sql-syntax-lexical.html#SQL-SYNTAX-OPERATORS), but not in mysql (https://dev.mysql.com/doc/refman/5.7/en/non-typed-operators.html) or snowflake (https://docs.snowflake.com/en/sql-reference/operators-logical.html). Interestingly isn't also not supported in Redshift (https://docs.aws.amazon.com/redshift/latest/dg/r_null_condition.html) either which means this might be the first feature which justifies a new redshift dialect!

Postgres classifies it as an operator, although I think sqlfluff should probably treat `NOTNULL` and `ISNULL` as keywords for our purposes. The logic for this should go in `Expression_A_Grammar`, but that expression is already quite big a probably needs a little surgery to make it a bit more configurable.

@nolanbconway, want to have a go at this? I don't immediately have a way of doing this tidily, but I'm open to suggestions!

PR #1483, opened 2021-10-01
Fix handling of `ISNULL` and `NOTNULL` keywords

### LINKED ISSUE #1067: Postgres notnull parse error rises again
Looks like https://github.com/sqlfluff/sqlfluff/issues/426 is back!

http://sqlfluff-online.herokuapp.com/fluffed?sql=H4sIALyAomAC_ytOzUlNLlFIVkgrys9VKFEoz0gtSgVy8_JL8kpzcni5AF74kBIhAAAA&dialect=postgres

```
$ sqlfluff version                                             
0.5.6

$ echo 'select c from t where c notnull' | sqlfluff lint - --dialect=postgres
== [stdin] FAIL
L:   1 | P:  24 |  PRS | Found unparsable section: ' notnull'
```

leamingrad:
Start treating `ISNULL` and `NOTNULL` as keywords in an expression rather than aliases

Prior to this change, if we were parsing an expression that contained the `ISNULL` or `NOTNULL`, we would treat them as
aliases. Since they are listed as keywords for Postgres and ANSI, this would cause some files to become unparsable (particularly if they were used in statements with actual aliases).

We fix this by introducing grammars for `ISNULL` and `NOTNULL`, and allowing them to be items in an expression. Since the `ISNULL` and `NOTNULL` keywords are non-standard (but are commonly supported), we remove them for the mysql, snowflake and exasol dialects which do not support them (based on the comment [here](https://github.com/sqlfluff/sqlfluff/issues/426#issuecomment-683465473)).

Technically we should also add a redshift dialogue which removes `ISNULL` and `NOTNULL`, but that is too big of a change to make now.

This change also removes `ISNULL` and `NOTNULL` from the `IsClauseGrammar`, as this would have allowed invalid `IS ISNULL` and `IS NOTNULL` constructs in expressions.

Fixes #1067

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -410,8 +410,6 @@ ansi_dialect.add(
     IsClauseGrammar=OneOf(
         "NULL",
         "NAN",
-        "NOTNULL",
-        "ISNULL",
         Ref("BooleanLiteralGrammar"),
     ),
     SelectClauseSegmentGrammar=Sequence(
@@ -434,6 +432,10 @@ ansi_dialect.add(
         Ref("CommaSegment"),
         Ref("SetOperatorSegment"),
     ),
+    # Define these as grammars to allow child dialects to enable them (since they are non-standard
+    # keywords)
+    IsNullGrammar=Nothing(),
+    NotNullGrammar=Nothing(),
     FromClauseTerminatorGrammar=OneOf(
         "WHERE",
         "LIMIT",
@@ -1434,6 +1436,8 @@ ansi_dialect.add(
                     Ref.keyword("NOT", optional=True),
                     Ref("IsClauseGrammar"),
                 ),
+                Ref("IsNullGrammar"),
+                Ref("NotNullGrammar"),
                 Sequence(
                     # e.g. NOT EXISTS, but other expressions could be met as
                     # well by inverting the condition with the NOT operator
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -187,6 +187,10 @@ postgres_dialect.replace(
         ansi_dialect.get_segment("ColumnReferenceSegment"),
         Ref("TimeZoneGrammar", optional=True),
     ),
+    # Postgres supports the non-standard ISNULL and NONNULL comparison operators. See
+    # https://www.postgresql.org/docs/14/functions-comparison.html
+    IsNullGrammar=Ref.keyword("ISNULL"),
+    NotNullGrammar=Ref.keyword("NOTNULL"),
 )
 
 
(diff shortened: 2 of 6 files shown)


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -409,10 +409,12 @@
...
         "NULL",
         "NAN",
-        "NOTNULL",
-        "ISNULL",
         Ref("BooleanLiteralGrammar"),
     ),
+    # Define these as grammars to allow child dialects to disable them (since they are non-standard
+    # keywords)
+    IsNullGrammar=Ref.keyword("ISNULL"),
+    NotNullGrammar=Ref.keyword("NOTNULL"),
```
tunetheweb, line 417: Should these be `Nothing()` in ANSI if they are non-standard then?

My understanding is that `Postgres` is the only one to support these.

That would also allow you to remove some of the overrides below for Exasol, MySQL, Snowflake.

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_ansi.py
leamingrad: I put these in here because the comments in the code say that `Expression_A_Grammar` was taken from the [Cockroach DB docs](https://www.cockroachlabs.com/docs/v20.2/sql-grammar.html#a_expr) and `ISNULL`/`NOTNULL` appear there.

That said, I wasn't entirely sure about this, so happy to change if preferred.

tunetheweb: Yeah I think better to reverse it if you don't mind. Cockroach DB isn't the same as ANSI and seems most dialects do not support this. But good to know you had a good reason for this!

leamingrad: I've reworked it, so let me know what you think.

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -434,6 +432,10 @@
         Ref("CommaSegment"),
         Ref("SetOperatorSegment"),
     ),
+    # Define these as grammars to allow child dialects to enable them (since they are non-standard
+    # keywords)
+    IsNullGrammar=Nothing()
```
tunetheweb, line 437: Need a comma here?

```suggestion
    IsNullGrammar=Nothing(),
```

tunetheweb (approved):
LGTM! Thanks for making those changes.

nolanbconaway:
FYI! Notnull is supported in redshift!

tunetheweb:
> FYI! Notnull is supported in redshift!

OK but we don't explicitly support redshift (yet), and it didn't work in ANSI before, so is anything broken by introducing this change?

nolanbconaway:
Not at all! I simply saw a note that if a redshift dialect were made then notnull would not be supported in it; my comment was just a note that I am P sure I use notnull in redshift commonly

tunetheweb:
Ah OK missed that original comment.

Well since we shifted this PR from the original comment we now no longer support by default (i.e. in ANSI) so if Redshift were added we'd need to explicitly add this (maybe another reason to add that dialect!).

nolanbconaway:
nope! nothing is broken!

PR #1491, opened 2021-10-01
Exasol: Fix `CREATE TABLE` in-/outline constraint / Adjusted DISTRIBUTE/PARTITION clause

sti0:
> I try to fix the IN- and OUTLINE constraint clause of the `CREATE TABLE` statement.
> 
> It's possible to write a statement with `CONSTRAINT PRIMARY KEY` (without an identifer). This is equal to just `PRIMARY KEY`. It's not a big deal to remove the `CONSTRAINT` keyword but it's technically possible and I want to understand why my implementation doesn't work. I added some new test cases for this behaviour.
> 
> It would be nice if someone can look over this.

- Fixed IN/OUTLINE CONSTRAINT for `CREATE TABLE` statements if there is no identifier
- Furthermore I changed the identifier from `SingleIdentifierGrammar` to `NakedIdentifierSegment` because no quotes are allowed here.
- Adjusted DISTRIBUTE/PARTITION clause
...

...

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -853,8 +853,16 @@ class TableInlineConstraintSegment(BaseSegment):
...
+    match_grammar = StartsWith(
+        OneOf("CONSTRAINT", "NOT", "NULL", "PRIMARY", "FOREIGN"),
+        terminator=OneOf("COMMENT", Ref("CommaSegment"), Ref("EndBracketSegment")),
+    )
+    parse_grammar = Sequence(
+        Sequence(
+            "CONSTRAINT",
+            Ref("NakedIdentifierSegment", optional=True),
+            optional=True,
+        ),
```
tunetheweb, line 865: The problem is it doesn't know the next keyword (PRIMARY) is not the optional naked identifier (since they are UNRESERVED keywords so could be used as identifiers). 

However I recently learned off a clever way around this, by using the `excludes` function (only available in `AnyNumberOf` so have to use that and then limit it to be 0 or 1, which is same as `Ref(..., optional=True):

```suggestion
    parse_grammar = Sequence(
        Sequence(
            "CONSTRAINT",
            AnyNumberOf(
                Ref("NakedIdentifierSegment"),
                max_times=1,
                min_times=0,
                exclude=OneOf("NOT", "NULL", "PRIMARY", "FOREIGN"),
            ),
            optional=True,
        ),
```

src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -872,8 +880,15 @@ class TableOutOfLineConstraintSegment(BaseSegment):
...
-        Sequence("CONSTRAINT", Ref("SingleIdentifierGrammar"), optional=True),
+    match_grammar = StartsWith(
+        OneOf("CONSTRAINT", "PRIMARY", "FOREIGN"),
+    )
+    parse_grammar = Sequence(
+        Sequence(
+            "CONSTRAINT",
+            Ref("NakedIdentifierSegment", optional=True),
+            optional=True,
+        ),
```
tunetheweb, line 891: Same here:

```suggestion
    parse_grammar = Sequence(
        Sequence(
            "CONSTRAINT",
            AnyNumberOf(
                Ref("NakedIdentifierSegment"),
                max_times=1,
                min_times=0,
                exclude=OneOf("PRIMARY", "FOREIGN"),
            ),
            optional=True,
        ),
```

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_exasol.py
sti0: Thank you so much. It needed some extra adjustment but worked quite well 🚀

src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -985,9 +985,10 @@ class CreateTableStatementSegment(BaseSegment):
                         Ref("TableOutOfLineConstraintSegment"),
                         Ref("CreateTableLikeClauseSegment"),
                         Ref("CommaSegment", optional=True),
-                        Ref("TableDistributionPartitonClause", optional=True),
                         min_times=1,
                     ),
+                    Ref("CommaSegment", optional=True),
```
tunetheweb, line 990: Does this make sense without the following `TableDistributionPartitonClause`?

I.e. should they both be in a sequence together that is optional, rather than both be optional individually?

tunetheweb (commented):
One small question. Other than that LGTM!

src/sqlfluff/dialects/dialect_exasol.py
sti0: The `CommaSegment` is not needed at all. It comes from the `AnyNumberOf` sequence above.

tunetheweb: Hmmm.... the only issue is this allows the following:

Trailing comma:
```sql
CREATE TABLE t5 (
    a DOUBLE,
    b DOUBLE,
);
```

and No commas:

```sql
CREATE TABLE t5 (
    a DOUBLE
    b DOUBLE
);
```

and no comma before the `TableDistributionPartitonClause`:

```sql
CREATE TABLE t5 (
    a DOUBLE,
    b DOUBLE
    DISTRIBUTE BY a
);
```

All of which are invalid Exasol SQL I presume?

However, I've played around with this and can't get it to easily prevent this (mostly due to #1366).

So maybe best merging this as it is?

sti0: You're correct, i noticed the same. I couldn't find out how to prevent this, too (at least the DB engine will stop this, as its no vaild statement). But I think this bug happens also at the current version. IMHO we should merge this and keep track on this in a new issue.

tunetheweb (approved):

PR #1502, opened 2021-10-03
Issue 1471: Inline --noqa not always honoured by "sqlfluff fix"

### LINKED ISSUE #1471: Inline noqa not honoured by "sqlfluff fix"
With the below test file:

```sql
--noqa: disable=L034
SELECT DISTINCT
    TO_CHAR(a, 'YYYY-MM-dd HH:MM:ss') as the_date,
    a AS b
FROM
    table1
```

Run the following:

```
% sqlfluff fix -f --fixed-suffix FIXED test.sql
% diff test.sql testFIXED.sql 
3,4c3,4
<     TO_CHAR(a, 'YYYY-MM-dd HH:MM:ss') as the_date,
<     a AS b
---
>     a AS b,
>     TO_CHAR(a, 'YYYY-MM-dd HH:MM:ss') AS the_date
```

## Expected Behaviour

The lowercase `as` should be corrected to `AS` (which does happen)
The column ordering should be left as is, as L034 is ignored (which does not happen)

## Observed Behaviour

The lowercase `as` is corrected to `AS` - good
The column ordering is not left as is - bad

## Steps to Reproduce

As per above.

Interestingly the ignore rule **is honoured** for L034 in the following situations:
- If `sqlfluff lint` is used instead of `sqlfluff fix`
- If defined on command line (`sqlfluff fix -f --fixed-suffix FIXED test.sql --exclude-rules L034`)
- If defined in `.sqlfluff` file (`exclude_rules = L034`)
- If there are no other failures for `sqlfluff fix` to fix (for example if the `as` is uppercase already in above example).

## Dialect
ANSI

## Version
0.6.6 or `main` as of d256450

## Configuration
None

barrywhart:

Fixes #1471

Currently, "noqa" is implemented as a filter against the list of violations on each file. However, the fix command updates the parse tree as it runs. By the time it considers whether to persist the changes or not, all the fixes have already been applied to the tree, regardless of "noqa".

This PR:
* Updates `BaseRule.crawl()` to respect the "noqa" directive
* Adds a pre-commit configuration to the project (unrelated to the PR, but I really hate surprise CI failures involving "Black", which are common)

...

### Pull Request checklist

  - Other.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,13 @@
+repos:
```
barrywhart, line 1: Add a pre-commit configuration to the project

src/sqlfluff/core/linter/linted_file.py
```diff
@@ -170,21 +170,22 @@ def _ignore_masked_violations_line_range(
                 result.append(v)
         return result
 
-    def _ignore_masked_violations(
-        self, violations: List[SQLBaseError]
+    @classmethod
+    def ignore_masked_violations(
```
barrywhart, line 174: Changed to public classmethod so it can be used externally by the linter.

src/sqlfluff/core/linter/linter.py
```diff
@@ -263,22 +263,25 @@ def parse_noqa(comment: str, line_no: int):
 
     @staticmethod
     def remove_templated_errors(
-        linting_errors: List[SQLLintError],
-    ) -> List[SQLLintError]:
+        linting_errors: List[SQLBaseError],
+    ) -> List[SQLBaseError]:
```
barrywhart, line 267: Parsing "noqa" directives can return `SQLParseError`, so that changes the type of this collection, and we have to update the method to handle the non-`SQLLintError` case.

src/sqlfluff/core/linter/linter.py
```diff
@@ -383,6 +388,10 @@ def lint_fix_parsed(
         if formatter:
             formatter.dispatch_lint_header(fname)
 
+        # Look for comment segments which might indicate lines to ignore.
```
barrywhart, line 391: This code moved from line 453, from `lint_parsed()` to `lint_parse_fixed()`.

src/sqlfluff/core/linter/linter.py
```diff
@@ -367,7 +372,7 @@ def lint_fix_parsed(
         fname: Optional[str] = None,
         templated_file: Optional[TemplatedFile] = None,
         formatter: Any = None,
-    ) -> Tuple[BaseSegment, List[SQLLintError]]:
+    ) -> Tuple[BaseSegment, List[SQLBaseError], List[NoQaDirective]]:
```
barrywhart, line 375: Updated return type to include the parsed "noqa" directives.

src/sqlfluff/core/rules/base.py
```diff
@@ -337,16 +339,29 @@ def crawl(
 
         new_lerrs = []
         new_fixes = []
+
+        def _process_lint_result(res):
+            self.discard_unsafe_fixes(res, templated_file)
+            lerr = res.to_linting_error(rule=self)
+            ignored = False
+            if lerr:
+                if ignore_mask:
```
barrywhart, line 348: New logic in this area to ignore fixes based on `ignore_mask` (noqa)

src/sqlfluff/core/rules/base.py
```diff
@@ -337,16 +339,29 @@ def crawl(
...
+            if not ignored:
+                new_fixes.extend(res.fixes)
+
         if res is None:
             # Assume this means no problems (also means no memory)
             pass
         elif isinstance(res, LintResult):
             # Extract any memory
             memory = res.memory
-            self.discard_unsafe_fixes(res, templated_file)
```
barrywhart, line 346: Extracted duplicate code from here and the section below into the new `_process_lint_result()` function above, which now _also_ handles "noqa".

.pre-commit-config.yaml
tunetheweb: More questions - does that update it locally (I presume so?)? Or do they need to do a `git pull`?
Will there ever be a circumstance where you do NOT want what the pre-commit gives you?
What if people do a quick edit on github.com? Does it work there?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - repo: https://github.com/psf/black
+    rev: 21.9b0
```
tunetheweb, line 3: Does this mean we have to keep it up to date? Not that it changes too much but still...

Anyway of NOT specifying a rev so latest is always used? Or is that more dangerous? Or major a major (`21`) only.

Anyway as I say, let's take it to another issue/PR.

tunetheweb (changes requested):
Change looks good but think we should do the following:
- Pull pre-commit to separate issue-change
- Add a test case for inline `noqa` (we only have a test case for disable going forward).
- Maybe add a test case for re-enabling `noqa`?

.pre-commit-config.yaml
barrywhart: I'll remove it. It's a bit slower to commit, but it only runs the checks on the modified files, so it's usually just a second or two.

Users only get it if they have pre-commit installed (it's a Python package and can also be installed using various package manager). Additionally, they must run `pre-commit install` (one time only) in the SQLFluff clone directory.

You can bypass the checks by adding the `--no-verify` option to your `git commit` command.

Pre-commit only runs on the client side, not on github.com. As such, it is a supplement to whatever CI checks are in place, _not_ a replacement.

Yes, it updates the files locally. The initial commit will fail, and you can review the locally changed files, then `git add` and `git commit` again.

.pre-commit-config.yaml
barrywhart: IIRC, you can specify "bleeding edge" if you always want the latest. This has the same downside as with packages, in that now each developer may be running with a different version, with different behavior.

.pre-commit-config.yaml
tunetheweb: Ah OK. Sounds great. But think we should do as separate PR and also add documentation to CONTRIBUTING as part of that explaining how to use and benefits of it!

barrywhart:
I removed the pre-commit file.

About test cases, are you asking for another test case? I had forgotten to check the checkbox in the PR description, but I _did_ add a test case. See the files in `test/fixtures/linter/autofix/ansi/021_fix_respects_noqa/`. Note that there are existing tests for the lower-level "noqa" parsing and filtering (i.e. testing the same code that this PR builds on), so we may not need to add a lot of new tests.

tunetheweb:
Yes I saw you added a test case to disable that rule for the rest of this file.

Does this rule work for a single line disable only or do you have to use disable? If so could we have a test case showing this - and also so it doesn't regress in future?

```
a.a*a.b AS bad_1  -- noqa: L006
```


Similarly does it work to reenable a rule? If so can we show this with a test case, again to avoid regressing this functionality?

```
-- Ignore rule L012 from this line forward
SELECT col_a a FROM foo --noqa: disable=L012

-- Ignore all rules from this line forward
SELECT col_a a FROM foo --noqa: disable=all

-- Enforce all rules from this line forward
SELECT col_a a FROM foo --noqa: enable=all
```

Get me?

barrywhart:
@tunetheweb: There's a pre-existing test [here](https://github.com/sqlfluff/sqlfluff/blob/main/test/core/linter_test.py#L564L595) that I believe addresses your request. The parsing and filtering of "noqa" directives was working correctly prior to this PR, but it had never been wired into the "fix" command, only the "lint" command. So I _think_ in this PR, we need to add a test verifying that "fix" respects "noqa".

tunetheweb:
Yes exactly!

barrywhart:
Will do!

barrywhart:
Ok, I extended `test/fixtures/linter/autofix/ansi/021_fix_respects_noqa/` by adding this more complex query from the other existing test:
```
SELECT
    col_a a,
    col_b b, --noqa: disable=L012
    col_c c,
    col_d d, --noqa: enable=L012
    col_e e,
    col_f f,
    col_g g,  --noqa
    col_h h,
    col_i i, --noqa:L012
    col_j j,
    col_k k, --noqa:L013
    col_l l,
    col_m m,
    col_n n, --noqa: disable=all
    col_o o,
    col_p p --noqa: enable=all
FROM foo
```

tunetheweb:
> Improve the test, fix a bug it discovered

Ha ha! Looks good then. Let's merge!

tunetheweb (approved):

PR #1504, opened 2021-10-04
Disable rules L026 and L028 for BigQuery by default, with option to reenable

### LINKED ISSUE #1503: L026 error while using bigquery nested field
Consider the following sql:

```sql
WITH commit_info AS (
    SELECT
        commit,
        author.name AS author_name
    FROM `bigquery-public-data.github_repos.commits`
    LIMIT 10
)

SELECT * FROM commit_info
```

## Expected Behaviour
Lint should not fail with L026 error

## Observed Behaviour
Lint fails with following error:
```
L:   4 | P:   9 | L026 | Reference 'author.name' refers to table/view not found
                       | in the FROM clause or found in parent subquery.
```

## Dialect
bigquery

## Version
- sqlfluff, version 0.6.6
- Python 3.9.7

tunetheweb:
Fixes #1503 

Rules L026 and L028 will now be disabled by default for BigQuery which technically is a breaking change. However I suspect most BigQuery users will have already disabled these in config. They can manually reenable if desired.

### Pull Request checklist

  - Other.


src/sqlfluff/rules/L026.py
```diff
@@ -2,12 +2,18 @@
...
+from sqlfluff.core.rules.doc_decorators import document_configuration
 from sqlfluff.rules.L025 import Rule_L020
 
 
+@document_configuration
 class Rule_L026(Rule_L020):
     """References cannot reference objects not present in FROM clause.
 
+    NB: This rule is disabled by default for BigQuery due to its use of
+    structs which trigger false positives. It can be enabled with the
```
barrywhart, line 14: Future-looking question (not a "strike" against this PR): Do we believe that someday we could get this rule working better for L026, either working 100% correctly or at least improve it so it can distinguish between:
* Definite issues
* Possible issues

If so, is there (should there be) an issue for that work?

src/sqlfluff/rules/L033.py
```diff
@@ -8,6 +8,9 @@
 class Rule_L033(BaseRule):
     """UNION [DISTINCT|ALL] is preferred over just UNION.
 
+    NB: This rule is only enabled for dialogs that support UNION DISTINCT
```
barrywhart, line 11:
```suggestion
    NB: This rule is only enabled for dialects that support UNION DISTINCT
```

src/sqlfluff/rules/L026.py
```diff
@@ -76,3 +84,14 @@ def _lint_references_and_aliases(
...
                 )
         return violation_buff or None
+
+    def _eval(self, segment, parent_stack, dialect, **kwargs):
+        """Override Rule L020 for dialects that use structs.
+
+        Some dialects use structs (e.g. column.field) which look like
+        table references and so incorrectly trigger this rule.
+        """
+        if dialect.name in ["bigquery"] and not self.force_enable:
```
barrywhart, line 94: Should we consider building in this disabling behavior at the core level to minimize having to add rule-specific code? This code is reasonable, but I'm thinking ahead if we end up doing this for several more rules.

E.g. could rules disable themselves through class-level config or code, e.g.:
```
 _default_disabled = {"dialects": ["bigquery"]}
```

This would be similar in spirit to `_works_on_unparseable`, for example:
https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/rules/L018.py#L39

test/fixtures/rules/std_rule_cases/L026.yml
```diff
@@ -39,4 +45,38 @@ test_pass_object_referenced_4:
...
-      dialect: bigquery
\ No newline at end of file
+      dialect: bigquery
+    rules:
+      L026:
+        force_enable: True
+
+test_pass_object_referenced_5:
+  # Test ambiguous column reference caused by use of BigQuery structure fields.
+  # Here, column,field should not trigger the rule as byu default this is off
```
barrywhart, line 55:
```suggestion
  # Here, column,field should not trigger the rule as by default this rule is disabled.
```

test/fixtures/rules/std_rule_cases/L026.yml
```diff
@@ -39,4 +45,38 @@ test_pass_object_referenced_4:
...
   configs:
     core:
-      dialect: bigquery
\ No newline at end of file
+      dialect: bigquery
+    rules:
+      L026:
+        force_enable: True
+
+test_pass_object_referenced_5:
```
barrywhart, line 53: Since tests 5-7 are closely related, should we indicate that somehow through the names, e.g.
* test_pass_object_referenced_5_a
* test_pass_object_referenced_5_b
* test_pass_object_referenced_5_c

test/fixtures/rules/std_rule_cases/L028.yml
```diff
@@ -44,3 +44,24 @@ test_value_table_functions_do_not_require_qualification:
   configs:
     core:
       dialect: bigquery
+    rules:
+      L028:
+        force_enable: True
+
+test_object_references_1:
```
barrywhart, line 51: Similar to L026 test comment, should we consider renaming these cases as:
* test_object_references_1_a
* test_object_references_1_b
* test_object_references_1_c

barrywhart (changes requested):

src/sqlfluff/rules/L026.py
tunetheweb: No I do not believe this is possible for dialects (like BigQuery) that use structs

It's impossible to distinguish the below as being a non-issue (because `b` is a column reference that is a STRUCT), ot being an issue (because `b` is a missing table alias):

```sql
SELECT 
  a,
  b.c
FROM
  table1
```

src/sqlfluff/rules/L026.py
tunetheweb: Potentially, but then would need to add `force_enable` config to all rules if we wanted to do that.

The other option is to have config at a higher level than rules:

e.g.

```yaml
dialect-exclusions:
   bigquery: L026, L028
   all: L033

dialect-inclusions:
   ansi: L033
   bigquery: L033
   hive: L033
```

Anyway needs some thought, so let's not do that now, but certainly one to think about as we add more dialect-specific rules.

Another thought I have is to avoid the whole "rules are on by default" mentality we have now. I think some should be more advanced/niche and be disabled by default with option to enable.

Anyway, bigger discussions for another time. Will leave this as is for now if you don't mind.

tunetheweb:
Addressed the simple issues and answered the more complex ones but don't think we should address now. Could you have another look @barrywhart ?

barrywhart (approved):

PR #1511, opened 2021-10-05
Issue_1490 TSQL: Support for unicode literals

### LINKED ISSUE #1490: TSQL: Support for unicode literals
TSQL identifies unicode single-quoted literals by prefixing a capital N:

`SELECT N'Weekend'
`
https://docs.microsoft.com/en-us/sql/t-sql/data-types/constants-transact-sql?view=sql-server-ver15#unicode-strings

SQLFluff currently does not parse this.

adam-tokarski:
Fixes #1490 
* It allows to use unicode single-quoted literals in T-SQL

---
### Pull Request checklist

  - Other.

### DIFF
--- a/test/rules/std_L048_test.py
+++ b/test/rules/std_L048_test.py
@@ -1,0 +1,66 @@
+"""Tests the python routines within L048."""
+import pytest
+
+import sqlfluff
+
+
+def test__rules__std_L048() -> None:
+    """L0048 is not raised for quoted literals surrounded by a single whitespace."""
+    sql = "SELECT a + 'b' + 'c' FROM tbl;"
+    result = sqlfluff.lint(sql)
+
+    assert len(result) == 1
+
+    results_l048 = [r for r in result if r["code"] == "L048"]
+    assert not len(results_l048)
+
+
+def test__rules__std_L048_raised() -> None:
+    """L0048 is raised for quoted literals not surrounded by a single whitespace."""
+    sql = "SELECT a +'b'+'c' FROM tbl;"
+    result = sqlfluff.lint(sql)
+
+    assert len(result) == 7
+
+    results_l048 = [r for r in result if r["code"] == "L048"]
+    assert len(results_l048) == 3
+    assert results_l048[0]["description"] == "Missing whitespace before 'b'"
+    assert results_l048[1]["description"] == "Missing whitespace after 'b'"
+    assert results_l048[2]["description"] == "Missing whitespace before 'c'"
+
+
+def test__rules__std_L048_unicode_tsql() -> None:
+    """L0048 is not raised for unicode quoted literals when T-SQL."""
+    sql = "SELECT a + N'b' + N'c' FROM tbl;"
+    result = sqlfluff.lint(sql, dialect="tsql")
+
+    assert len(result) == 1
+
+    results_l048 = [r for r in result if r["code"] == "L048"]
+    assert not len(results_l048)
+
+
+@pytest.mark.parametrize("dialect", ["ansi", "mysql", "postgres", "sqlite", "bigquery"])
+def test__rules__std_L048_unicode_other_dialects(dialect: str) -> None:
+    """L0048 is raised for unicode quoted literals when some other dialects."""
+    sql = "SELECT a + N'b' + N'c' FROM tbl;"
+    result = sqlfluff.lint(sql, dialect=dialect)
+
+    assert len(result) == 3
+
+    results_l048 = [r for r in result if r["code"] == "L048"]
+    assert len(results_l048) == 2
+    assert results_l048[0]["description"] == "Missing whitespace before 'b'"
+    assert results_l048[1]["description"] == "Missing whitespace before 'c'"
+
+
+def test__rules__std_L048_unicode_tsql_small_n() -> None:
+    """L0048 when there is `n` instead of `N` in a T-SQL."""
+    sql = "SELECT a + N'b' + n'c' FROM tbl;"
+    result = sqlfluff.lint(sql, dialect="tsql")
+
+    assert len(result) == 2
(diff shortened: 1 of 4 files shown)


test/rules/std_L048_test.py
```diff
@@ -0,0 +1,66 @@
+"""Tests the python routines within L048."""
```
tunetheweb, line 1: Is there any reason for these not to just be YML tests?

IMHO we should only use the python tests if there's something specific we wanna check that can't be covered by the YML tests  (e.g. two rules running at same time, or a different error message is returned than default when it should be).

Not seeing anything in below tests that couldn't be covered with a YML test case.

Yes you are checking that L048 is passing or failing, but as the YML test cases are only run for the specific rule, it's safe to assume if it fails it's because it's L048.

The text error messages are something we can't really check in YML but, as they are always the same message (though admittedly with a slightly different parameter), I'm not sure if they are worth adding the python test cases just for that and think sufficient to test for pass and fail. Maybe one Python test case to cover that (the `test__rules__std_L048_raised` one seems pretty comprehensive) and move the others to YML?

test/rules/std_L048_test.py
```diff
@@ -0,0 +1,66 @@
...
+
+def test__rules__std_L048() -> None:
+    """L0048 is not raised for quoted literals surrounded by a single whitespace."""
+    sql = "SELECT a + 'b' + 'c' FROM tbl;"
+    result = sqlfluff.lint(sql)
+
+    assert len(result) == 1
+
+    results_l048 = [r for r in result if r["code"] == "L048"]
+    assert not len(results_l048)
```
tunetheweb, line 15: Think this is covered with the previous YML test cases so don't see need for it in Python.

test/rules/std_L048_test.py
```diff
@@ -0,0 +1,66 @@
...
+    sql = "SELECT a +'b'+'c' FROM tbl;"
+    result = sqlfluff.lint(sql)
+
+    assert len(result) == 7
+
+    results_l048 = [r for r in result if r["code"] == "L048"]
+    assert len(results_l048) == 3
+    assert results_l048[0]["description"] == "Missing whitespace before 'b'"
+    assert results_l048[1]["description"] == "Missing whitespace after 'b'"
+    assert results_l048[2]["description"] == "Missing whitespace before 'c'"
```
tunetheweb, line 29: Maybe good to leave this one in Python just to check error messages.

test/rules/std_L048_test.py
```diff
@@ -0,0 +1,66 @@
...
+
+def test__rules__std_L048_unicode_tsql() -> None:
+    """L0048 is not raised for unicode quoted literals when T-SQL."""
+    sql = "SELECT a + N'b' + N'c' FROM tbl;"
+    result = sqlfluff.lint(sql, dialect="tsql")
+
+    assert len(result) == 1
+
+    results_l048 = [r for r in result if r["code"] == "L048"]
+    assert not len(results_l048)
```
tunetheweb, line 40: IMHO covered by `test_pass_tsql_unicode_singe_quote` in YML so unecessary.

test/rules/std_L048_test.py
```diff
@@ -0,0 +1,66 @@
...
+    """L0048 is raised for unicode quoted literals when some other dialects."""
+    sql = "SELECT a + N'b' + N'c' FROM tbl;"
+    result = sqlfluff.lint(sql, dialect=dialect)
+
+    assert len(result) == 3
+
+    results_l048 = [r for r in result if r["code"] == "L048"]
+    assert len(results_l048) == 2
+    assert results_l048[0]["description"] == "Missing whitespace before 'b'"
+    assert results_l048[1]["description"] == "Missing whitespace before 'c'"
```
tunetheweb, line 54: Covered by `test_fails_tsql_unicode_singe_quote` in YML.

Don't really need to test every dialect (and unsustainable as we add more dialects).

Also it's not valid SQL in any of the other AFAIK so even though it throws L048 error it's still not vali so question value of this case on top of the `test_fails_tsql_unicode_singe_quote` you already added in YML.

test/rules/std_L048_test.py
```diff
@@ -0,0 +1,66 @@
...
+def test__rules__std_L048_unicode_tsql_small_n() -> None:
+    """L0048 when there is `n` instead of `N` in a T-SQL."""
+    sql = "SELECT a + N'b' + n'c' FROM tbl;"
+    result = sqlfluff.lint(sql, dialect="tsql")
+
+    assert len(result) == 2
+    results_l048 = [r for r in result if r["code"] == "L048"]
+
+    assert len(results_l048) == 1
+    assert results_l048[0]["description"] == "Missing whitespace before 'c'"
```
tunetheweb, line 66: Think it can be covered in YML test case.

tunetheweb (changes requested):
The change itself looks good, but do have some suggestions for the test cases.

adam-tokarski:
Ok, I moved these tests into `yml` files (except that suggested one). I did it in both places because I created firstly tests in `tests/rules` then realized that there are also these in yml.

tunetheweb:
Good stuff. Thanks for that.

GitHub Actions is having an off day (https://www.githubstatus.com/) so need to wait for that. But LGTM!

test/rules/std_L048_test.py
```diff
@@ -0,0 +1,16 @@
+"""Tests the python routines within L048."""
+import sqlfluff
+
+
+def test__rules__std_L048_raised() -> None:
+    """L0048 is raised for quoted literals not surrounded by a single whitespace."""
```
tunetheweb, line 6: Nit (which is lucky as looking for a small change to kick off CI 😁):

```suggestion
    """L048 is raised for quoted literals not surrounded by a single whitespace."""
```

tunetheweb (approved):

PR #1515, opened 2021-10-05
added type hints to rule files

ttomasz:
makes progress on #1059 

This PR adds type hints to rule files.
It covers only a few files so far but since it will add a type Unioning all Segment types and another person wanted to join the effort I wanted to push this quickly so they can share it.

At first I wanted to use BaseSegment type but since some of the code works on attributes that are available only on some segment types i think the best solution was to create this AnySegmentType that merges all Segment types.

In L003.py in _make_indent method I added else block for unexpected inputs. Function would raise an exception anyway but this way it should have a nicer message.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/parser/segments/__init__.py
+++ b/src/sqlfluff/core/parser/segments/__init__.py
@@ -1,6 +1,7 @@
 """Definitions of the segment classes."""
 
 # flake8: noqa: F401
+from typing import Union
 
 from sqlfluff.core.parser.segments.base import (
     BaseSegment,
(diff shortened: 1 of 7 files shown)


ttomasz:
Ah i didn't know I could run `tox -e mypy` locally 😅 Although I am surprised that it complains about lines that I didn't touch e.g. missing return statements or missing type annotations for variables.

adam-tokarski:
> Ah i didn't know I could run `tox -e mypy` locally

You can call it even without the tox:
```console
mypy src/sqlfluff
```
(make sure that you're using a venv).

ttomasz:
Added some types for rules files and made them compliant with both Pycharm type checker and MyPy. Let me know if you see any issues.

PR #1527, opened 2021-10-06
Add a rule to warn on "= NULL" or "<> NULL" comparisons

barrywhart:

Adds a new rule, L049, that warns on use of `=` or `<>` to compare against `NULL` values. Such code should use `IS` or `IS NOT` instead. Did not add a fix routine -- should we? Are there cases (even if rare) where it's appropriate/necessary to do this?

No.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L049.py
+++ b/src/sqlfluff/rules/L049.py
@@ -1,0 +1,73 @@
+"""Implementation of Rule L006."""
+
+
+from sqlfluff.core.rules.base import BaseRule, LintResult
+from sqlfluff.core.rules.doc_decorators import document_fix_compatible
+
+
+@document_fix_compatible
+class Rule_L049(BaseRule):
+    """Comparisons with NULL should use "IS" or "IS NOT".
+
+    | **Anti-pattern**
+    | In this example, the <> is used to check for NULL values'.
+
+    .. code-block:: sql
+
+        SELECT
+            a
+        FROM foo
+        WHERE a <> NULL
+
+
+    | **Best practice**
+    | Use "IS" or "IS NOT" to check for NULL values.
+
+    .. code-block:: sql
+
+        SELECT
+            a
+        FROM foo
+        WHERE a IS NOT NULL
+    """
+
+    def _eval(self, segment, **kwargs):
+        """Relational operators should not be used to check for NULL values."""
+        # Iterate through children of this segment looking for any of the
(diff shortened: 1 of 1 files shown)


src/sqlfluff/rules/L049.py
```diff
@@ -0,0 +1,65 @@
+"""Implementation of Rule L006."""
```
tunetheweb, line 1:
```suggestion
"""Implementation of Rule L049."""
```

src/sqlfluff/rules/L049.py
```diff
@@ -0,0 +1,65 @@
...
+
+    | **Anti-pattern**
+    | In this example, the <> is used to check for NULL values'.
+
+    .. code-block:: sql
+
+        SELECT
+            a
+        FROM foo
+        WHERE a <> NULL
```
tunetheweb, line 20: Is `=` the more obvious example?

```suggestion
    | In this example, the '=' is used to check for NULL values'.

    .. code-block:: sql

        SELECT
            a
        FROM foo
        WHERE a = NULL
```

src/sqlfluff/rules/L049.py
```diff
@@ -0,0 +1,65 @@
...
+        # equals".
+        if len(segment.segments) <= 2:
+            return LintResult()
+
+        operator = None
+        for idx, sub_seg in enumerate(segment.segments):
+            # Skip anything which is whitespace or non-code.
+            if sub_seg.is_whitespace or not sub_seg.is_code:
+                continue
+            # Look for "=" or "<>".
```
tunetheweb, line 46: What about `>` or `<`? Do they work with `NULL`?

Should you check for `IS` and `NOT NULL` and complain on anything else?

src/sqlfluff/rules/L049.py
barrywhart: Done!

src/sqlfluff/rules/L049.py
barrywhart: SQLFluff can parse `SELECT 1 > NULL`, but BigQuery reports a syntax error: `Operands of > cannot be literal NULL`. Not sure about other databases.

I'm torn about whether to make this change. It's much more likely `=` and `<>` would be used unknowingly, because it seems to make sense. Unless you or someone else feels strongly about changing it, I would prefer (for now at least) to keep it "as is".

Issue #1535, opened 2021-10-07
Zero length Lex item returned from '<unlexable>'


halostatue:
SQLFluff version 0.6.8.
Python 3.9.7 (via pipx venv)

`.sqlfluff` config:

```
[sqlfluff]
dialect = postgres
output_line_length = 100

[sqlfluff:rules]
tab_space_size = 2
max_line_length = 100

[sqlfluff:rules:L010]  # Keywords
capitalisation_policy = upper

[sqlfluff:rules:L014]  # Unquoted identifiers
allow_scalar = true
extended_capitalisation_policy = consistent

[sqlfluff:rules:L030]  # Function names
capitalisation_policy = lower

[sqlfluff:rules:L040]  # Null & Boolean Literals
capitalisation_policy = upper
```

Dialect `postgres`

## Expected Behaviour

Cannot parse

```sql
-- The holding account is created using the config.default_users mix task.
CREATE OR REPLACE FUNCTION hold_account_id()
RETURNS uuid LANGUAGE plpgsql VOLATILE STRICT
AS $$
DECLARE
  hold_account_id users.id%TYPE;

BEGIN
  SELECT id INTO hold_account_id
    FROM users
   WHERE email = 'example@example.com';

  IF hold_account_id IS NULL THEN
    RAISE EXCEPTION 'Cannot find user for example@example.com';
  END IF;

  RETURN hold_account_id;
END;
$$;

\prompt 'Cash Account #: ' account
\prompt 'Region (1 - quebec, 2 - east, 3 - west): ' region_number

SELECT *
  FROM card_matches
 WHERE region::text = CASE :region_number WHEN 1 THEN 'quebec' WHEN 2 THEN 'east' WHEN 3 THEN 'west' ELSE 'unknown' END
   AND customer_number = rpad(lpad((:account)::text, 10, '0'), 20, '0');

-- Information to feed into GraphQL.
SELECT 'reassignCashCard' AS mutation,
       encode_relay_id(hold_account_id(), 'Customer') AS new,
       encode_relay_id(cc.customer_id, 'Customer') AS old,
       encode_relay_id(cc.id, 'CustomerCard') AS card
  FROM customer_cards AS cc
 WHERE cc.region::text = CASE :region_number WHEN 1 THEN 'quebec' WHEN 2 THEN 'east' WHEN 3 THEN 'west' ELSE 'unknown' END
   AND cc.code = rpad(lpad(:'account', 10, '0'), 20, '0')
[...]

PR #1548, opened 2021-10-07
Support brackets in Postgres Meta commands

tunetheweb:
Makes progress on #1535

Don't think so. Think brackets were left there in error. @WittierDinosaur can you confirm?

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -45,16 +45,18 @@ from sqlfluff.core import (
     dialect_readout,
     TimingSummary,
 )
+
+from sqlfluff.core.enums import FormatType, Color
 from sqlfluff.core.linter import ParsedString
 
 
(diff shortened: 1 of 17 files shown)


WittierDinosaur:
I can’t really run any tests right now, so you’re probably right, though I might have meant to use [] instead… if tests are passing I’m happy. Can’t see any .yml changes, so have you definitely ran generate?

tunetheweb:
> Looks good. I'm curious as to why we didn't do this in the first place 🤔 , but it looks like it works, and it's simpler than the previous implementation!

I think it's a lot to do wiht this comment: https://github.com/sqlfluff/sqlfluff/pull/1548#pullrequestreview-775480689 😀

> **Are there any other side effects of this change that we should be aware of?**
> Creating the correct regex may have cost me my sanity.

PR #1553, opened 2021-10-07
TSQL - BEGIN/END blocks

### LINKED ISSUE #1555: TSQL: Incorrect indent in if block
Example code:
```sql
if @debug = 1
  begin
    print '<something>';
    print '<something else>';
    return
  end
```

## Expected Behaviour
The indentation is correct like this.

## Observed Behaviour
The third line (`print '<something else>';`) is de-indented to the same level as line 1.

## Steps to Reproduce
see above

## Dialect
tsql

## Version
sqlfluff version 0.6.8, Python 3.10.0

## Configuration
none

jpers36:
Fixes #1555 

Adding BEGIN/END blocks to TSQL. This will improve our parsing of stored procedures, and lay the foundation for properly parsing IF/ELSE statements and functions.

In order to support BEGIN/END blocks, I'm introducing a new segment called the BatchSegment.  A batch is bounded by GO statement segments within a FileSegment, or wrapped within BEGIN/END within a BeginEndSegment.

I'm overriding FileSegment in order to incorporate this new BatchSegment.  Now, instead of being a delimited set of StatementSegments, a FileSegment is a delimited set of BatchSegments.  A new delimiter segment called BatchDelimiterSegment now exists to perform this delimitation.  The standard delimiter segment has had GO removed from its definition; the segment and its delimitation have been pushed down one level to within BatchSegment.

Because of these changes, the StatementSegment's GreedyUntil match logic no longer works.  Instead, we are moving the parser logic to the match logic.

We are also overriding TransactionSegment now in order to allow the parser to easily distinguish between it and a BEGIN/END block.  In TSQL, a Transaction must start with "BEGIN TRANSACTION" and not just "BEGIN", likely for the same distinguishing purpose.

I have added TSQL test cases begin_end_nested.sql, stored_procedure_begin_end.sql and transaction.sql.

As a result of the new BatchSegment layer under FileSegment, all test case parser YMLs have been updated to include a batch layer. "Use Hide whitespace changes: when reviewing to make ths the changes more obvious:

image

T-SQL now no longer as a quick `GreedyUntil` `match_grammar` option so will be slightly slower. But this is necessary as semi-colons are used within statements.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_tsql.py
+++ b/src/sqlfluff/dialects/dialect_tsql.py
@@ -1050,7 +1050,7 @@ class FileSegment(BaseFileSegment):
 
     We override default as T-SQL allows concept of several
     batches of commands separated by GO as well as usual
-    Semi-colon statement separated lines.
+    semicolon-separated statement lines.
 
     This is also the default "root" segment of the dialect,
     and so is usually instantiated directly. It therefore


jpers36:
@tunetheweb I'll be working on this throughout the day today!

PR #1609, opened 2021-10-12
Progressbar utility

### LINKED ISSUE #1536: Add progress indicator
When you have large SQL file, it would be nice to know how much is done and how much is still to be done. So, please add some kind of progress indicator.

E.g. if you have 100k queries in a file, and have processed already 10k of them, you could show that 10% of that file is processed already.

adam-tokarski:
Fixes #1536 

There is my proposition for the progressbar utility.

- single file
image

- for the whole directory
image

- multiple paths
image

It is enabled by default (can be disabled with `--disable_progress_bar` flag) and rendered for linting and fixing (not for parsing).

The steps are iterated over
* paths (if more than 1)
* files from path (if more than 1)
* rules for each file
* also there is one dummy semi-progress bar rendered during parsing, which is just for showing that something happens by printing a label, not a real progress, but it is because of nature of recurrent parsing.

### Pull Request checklist

  - Other.


adam-tokarski:
I knew that there may be issues with parsing, so I've put that into PR description
> The steps are iterated over
> 
> * paths (if more than 1)
> * files from path (if more than 1)
> * rules for each file
> * also there is one dummy semi-progress bar rendered during parsing, which is just for showing that something happens by printing a label, not a real progress, but it is because of nature of recurrent parsing.

but it might not detailed enough I guess. I was going to record some gif from console to show how it animates, but decided to put some pictures instead. Maybe with that animation it could be more clear from the very beginning.


---
Ok, so if there is no clear way to say how many steps can be done during the parsing, or how long can it take, we can either do some infinity progress bar (going back and forth) or arbitrary set some high number of steps and just move progress during the parsing as long as it is possible.

I'll take another look on that part.

adam-tokarski:
Was the `master` changed to `main` recently?

adam-tokarski:
> Ah looks like GitHub Actions Windows tests are all queuing. Might be that then...

Something wrong happens here https://status.dev.azure.com/
image

tunetheweb:
All fixed from our end.

tunetheweb:
@adam-tokarski could you look at the failing tests?

adam-tokarski:
> @adam-tokarski could you look at the failing tests?

Yes, I'm working on that, but it takes so long because of lack of time. Sorry. But I will fix that all eventually.

tunetheweb:
BTW what is "matching for each file"? Is that lexing? Would we be better calling it that? Not that it stays on screen for very long to even see it to be honest!

adam-tokarski:
@tunetheweb 

> BTW what is "matching for each file"? Is that lexing? Would we be better calling it that? 

This is done in [match](https://github.com/sqlfluff/sqlfluff/pull/1609/files#diff-5d9f753288388b9120f501441bd7e49b6a2d30fe07a9d61fdee41a010006c821R78) method which is called during `parsing`, so it was naturally to put `matching` there. Lexing is done before that AFAIK.
But if simple `matching` is confusing, we can change that, no problem. Any ideas? Would be good to have it short.


> Not that it stays on screen for very long to even see it to be honest!

When I used file with 1 thousand of `select id from tbl9;` lines, that specific phase takes quite a long time. So for such cases it would be good to name it properly.

tunetheweb:
OK happy to leave as matching.

Should we have a `lexing` stage? Though imagine that's even shorter so maybe not worth while?

barrywhart:
"matching" is a SQLFluff-specific term. I would call it "parsing", since that's a standard computer language term.

adam-tokarski:
> Should we have a lexing stage? Though imagine that's even shorter so maybe not worth while?

Right, it will not be very useful. Additionally, this may be another tricky part where we don't know the number of steps a priori.

> "matching" is a SQLFluff-specific term. I would call it "parsing", since that's a standard computer language term.

But there is actually progress bar for [parsing](https://github.com/sqlfluff/sqlfluff/pull/1609/files#diff-a00ff6dd7618133c098d44ce5127dae8f3fc4233a8c239cd65f4d1f9ecde8b95R285). It's called inside `expand` method, but still as part of `parsing`, just after `matching` is done.
I'm open for name changes propositions :)

tunetheweb:
Let's leave as is. As you say it's only in that stage for an unnoticable amount of time most of the time.

This good to merge then?

tunetheweb (approved):

adam-tokarski:
> This good to merge then?

I'd be happy to have it merged. I don't see any further issues, so if no one else finds anything to be adjusted, we can go with that.

PR #1651, opened 2021-10-14
TSQL: Allow for multiple variables DECLAREd in the same statement

jpers36:
The DECLARE segment was not properly parsing multiple variable declarations.  This change moves AnyNumber to the correct location with the correct content.  Added example to an existing test case.


jpers36:
@tunetheweb since we can't separate the statement segments in TSQL using a delimiter, any effort to use Delimited() ends up grabbing all elements up through the next delimiter if it exists in a following statement.  So the last element gets misparsed to include part of the following segment or more.

PR #1657, opened 2021-10-15
Allow :: casting of CASE statements

### LINKED ISSUE #1655: Parser throws error on type case after case/when
Running SQL lint on
```sql
select
    col0,
    case
        when col1
        then col2
        else col3
    end::text as mycol
from table1
```

## Expected Behaviour
Query passes

## Steps to Reproduce
Running linter on code above

## Dialect
postgres

## Version
sqlfluff, version 0.6.9
Python 3.8.11

## Configuration
```
[sqlfluff]
verbose = 0
nocolor = False
dialect = postgres
templater = jinja
rules = None
exclude_rules = L009,L011,L031,L034
recurse = 0
output_line_length = 160
runaway_limit = 10
ignore_templated_areas = True
encoding = autodetect

[sqlfluff:indentation]
indented_joins = False
indented_using_on = True
template_blocks_indent = True

[sqlfluff:templater]
unwrap_wrapped_queries = True

[sqlfluff:templater:jinja]
apply_dbt_builtins = True

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 120
indent_unit = space
comma_style = trailing
allow_scalar = True

[sqlfluff:rules:L003]
lint_templated_tokens = True

operator_new_lines = after

[sqlfluff:rules:L016]
ignore_comment_lines = False

[sqlfluff:rules:L038]

[sqlfluff:rules:L042]
forbid_subquery_in = join

prefer_count_1 = False
prefer_count_0 = False

```

tunetheweb:
Fixes #1655 

It changes the ansi expression C syntax, but this was already added as an extra to Expression D so think that's OK.

### Pull Request checklist

  - Other.


tunetheweb:
> I think in both cases we should also have AnyNumberOf too because I think foo::int::float is still also valid in all of the same places as foo::int

Added a test case for this.

tunetheweb:
> Looks good. I wasn't quite sure the distinction between the two .sql files?

No real difference. The change was in ANSI so added test case there. However the original request was for Postgres, so felt I should also add an explicit test case there too (even though it’s the same!) to avoid it regressing there. Maybe overkill to have in both but adding more test cases is almost never a bad thing!

PR #1684, opened 2021-10-17
Exclude Exasol scripts from rule L003

### LINKED ISSUE #1680: L003 breaks `EXASOL` script body formatting
When fixing a `exasol` script, L003 breaks the script body. The script body could be `Anything()` because it could be LUA, Python....

https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/dialects/dialect_exasol.py#L3480

## Expected Behaviour
L003 should ignore the children of `ScriptContentSegment`.

## Observed Behaviour
L003 removes the idention within the script body
image

## Steps to Reproduce
`sqlfluff fix --dialect=exasol test.sql`

test.sql:
```
CREATE OR REPLACE LUA SCRIPT ASCRIPT (APARAM) RETURNS ROWCOUNT AS
    res = 1
    suc = true
    if not suc then
        error("ERROR")
    end
    return res
/

```

## Dialect
exasol

## Version
sqlfluff, version 0.7.0
Python 3.8.12

## Configuration
```
[sqlfluff]
dialect=exasol

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 120
indent_unit = space
comma_style = trailing
```

tunetheweb:
Fixes #1680

Adds Indent/Dedent clauses to Exasol (as otherwise the Dedent is seen as part of the script and so causes the terminator line to be flagged as incorrect.

I've made the element type list a hardcoded listed. I decided against making it configurable as think this really depends on the internal parsing structure of SQLFluff so not something that really should be exposed internally. But let me know if you think otherwise. We've precendence with this in other rules (e.g. `_target_elems` in L006)

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -20,7 +20,11 @@ from sqlfluff.core.parser import (
 )
 
 from sqlfluff.core.dialects import load_raw_dialect
-from sqlfluff.dialects.dialect_postgres_keywords import postgres_keywords, get_keywords
+from sqlfluff.dialects.dialect_postgres_keywords import (
+    postgres_keywords,
+    get_keywords,
+    postgres_postgis_datatype_keywords,
+)
 
 ansi_dialect = load_raw_dialect("ansi")
 
@@ -137,6 +141,7 @@ postgres_dialect.add(
     DollarQuotedLiteralSegment=NamedParser(
         "dollar_quote", CodeSegment, name="dollar_quoted_literal", type="literal"
     ),
+    SimpleGeometryGrammar=AnyNumberOf(Ref("NumericLiteralSegment")),
 )
 
 postgres_dialect.replace(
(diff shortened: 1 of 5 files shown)


tunetheweb:
> Works as expected. Thank you so much

Good stuff. BTW it doesn't currently try to fix the indentation of the bad separate:

For example:
```sql
CREATE OR REPLACE
LUA SCRIPT ASCRIPT (APARAM) RETURNS ROWCOUNT AS
    res = 1
    suc = true
    if not suc then
        error("ERROR")
    end
    return res
  /
```

Becasue currently you allow spaces in the `function_script_terminator`. 

If you made that more strict then could start to look at that:

```py
         RegexLexer(
             "function_script_terminator",
            r"\n\/\n",
             CodeSegment,
             segment_kwargs={"type": "function_script_terminator"},
            subdivider=RegexLexer(
                 "newline",
                 r"(\n|\r\n)+",
                 NewlineSegment,
             ),
         ),
```

Which would also allow multiple functions per file and also newline characters in functions (providing proceeded by whitespace):

```sql
CREATE OR REPLACE
LUA SCRIPT ASCRIPT (APARAM) RETURNS ROWCOUNT AS
    res = 1
    suc = true
    if not suc then
        error("ERROR")
    end
    /
    return res
/

CREATE OR REPLACE
LUA SCRIPT ASCRIPT (APARAM) RETURNS ROWCOUNT AS
    res = 1
    suc = true
    if not suc then
        error("ERROR")
    end
    return res
/
```

But think that's separate to this PR. But one for you to consider.

PR #1690, opened 2021-10-18
TSQL: Add #, @ to valid identifier characters

jpers36:
Naked identifiers in TSQL are allowed to have hash marks and at signs anywhere after the first character.  These are the things I discover as I test SQLFluff against an existing codebase without prior oversight.


tunetheweb:
Not sure we've overriden `code` before but looks to work fine.

PR #1694, opened 2021-10-18
TSQL: allow for semi-colon after BEGIN in a BEGIN/END block

jpers36:
TSQL: The BEGIN in a BEGIN/END block can be ended by a semi-colon.


jpers36:
@tunetheweb it's why I need a linter so badly!

PR #1744, opened 2021-10-25
Improve Parser Performance By Caching Values

WittierDinosaur:
This PR aims to deliver a performance boost to the parser. On this test file, parser time drops from 711 secs to 20secs:
https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2019/performance/07_04d.sql

Yes, it removes a bunch of generators, and adds a cached property. The cached properties are now reset when __setattr__ is called on segments.

### Pull Request checklist
All current test cases pass


WittierDinosaur:
@alanmcruickshank @barrywhart @tunetheweb Can you guys take another look when you get the chance? I think I've now addressed everything that's been raised

alanmcruickshank (approved):
Nice. Thanks for taking the time to put this together 👍

tunetheweb (approved):
Can't comment too much on the hard code python code (will leave that to the real pythonistas!), but can't spot anything wrong.

And linting that big file goes from 40mins on my lowly powered Macbook to 4 mins! Amazing difference.

WittierDinosaur:
Fantastic! It'd be useful to know the breakdown of times between parser and linter for that file. Could you fire me the output of this?
```
python -m cProfile -s cumtime -m sqlfluff.cli.commands lint mybigfile.sql
```

barrywhart:
Great work!! 🎉💯

I took a look at performance one time and kinda got lost. Tuning recursive code feels very different than the stuff I've tuned in the past.

tunetheweb:
Before your fix: https://gist.github.com/tunetheweb/be324a51669a01731cec49fe16847832

After your fix: https://gist.github.com/tunetheweb/995756d807b733ce5dc190fc1e6b6ccb

PR #1775, opened 2021-10-28
L029: Check quoted identifiers in addition to naked identifiers

### LINKED ISSUE #1747: New Rule Suggestion: Identify and flag object names that are reserved words
We should be able to identify and flag the creation of objects that share names with reserved words.

Anti-pattern:
```
CREATE TABLE DBO.ColumnNames2
(
    [Table] INT,
    [Select] INT,
    [Create] INT,
    [Drop] INT
)
```

jpers36:
Fixes #1747

This PR adjusts L029 to optionally check quoted identifiers in addition to naked identifiers.
In order to do this, we are adding a new configuration keyword, quoted_identifiers_policy, which represents the same thing as unquoted_identifiers_policy but for quoted identifiers.
The options for quoted_identifiers_policy are the same as unquoted_identifiers_policy with the addition of none, representing no checks for quoted identifiers.

We check quoted identifiers by stripping off the first and last character.  If we have a case where this shortcut would fail, we'll need to tweak the code before merging.

Changes:
+quoted_identifiers_policy default of none to default_config.cfg
+quoted_identifiers_policy description in config_info.py
L014.py: Rename unquoted_ids_policy_applicable function to identifiers_policy_applicable in order to reflect change in functionality
L029.py: Reflect renaming of unquoted_ids_policy_applicable function; add quoted_identifiers_policy as a configuration keyword; adjust rule to check quoted identifiers based on the config setting.
L029.yml: Rename test cases to be more descriptive.  Add quoted identifier test cases.

### DIFF
--- a/src/sqlfluff/rules/L014.py
+++ b/src/sqlfluff/rules/L014.py
@@ -11,10 +11,10 @@ from sqlfluff.core.rules.doc_decorators import (
 from sqlfluff.rules.L010 import Rule_L010
 
 
-def unquoted_ids_policy_applicable(
+def identifiers_policy_applicable(
     policy: str, parent_stack: Tuple[BaseSegment, ...]
 ) -> bool:
-    """Does `unquoted_identifiers_policy` apply to this segment?"""
+    """Does `(un)quoted_identifiers_policy` apply to this segment?"""
     if policy == "all":
         return True
     is_alias = parent_stack and parent_stack[-1].is_type(
@@ -70,7 +70,7 @@ class Rule_L014(Rule_L010):
     _description_elem = "Unquoted identifiers"
 
     def _eval(self, context: RuleContext) -> LintResult:
-        if unquoted_ids_policy_applicable(
+        if identifiers_policy_applicable(
             self.unquoted_identifiers_policy, context.parent_stack  # type: ignore
         ):
             return super()._eval(context=context)
--- a/src/sqlfluff/rules/L014.py
+++ b/src/sqlfluff/rules/L014.py
@@ -17,6 +17,8 @@ def identifiers_policy_applicable(
     """Does `(un)quoted_identifiers_policy` apply to this segment?"""
     if policy == "all":
         return True
+    if policy == "none":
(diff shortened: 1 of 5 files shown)


src/sqlfluff/rules/L014.py
```diff
@@ -11,10 +11,10 @@
 from sqlfluff.rules.L010 import Rule_L010
 
 
-def unquoted_ids_policy_applicable(
+def identifiers_policy_applicable(
     policy: str, parent_stack: Tuple[BaseSegment, ...]
 ) -> bool:
-    """Does `unquoted_identifiers_policy` apply to this segment?"""
+    """Does `(un)quoted_identifiers_policy` apply to this segment?"""
```
barrywhart, line 17: While not strictly necessary, I wonder if we may want to add:
```
if policy == "none":
    return False
```

That would make the code easier to understand and potentially a bit faster.

barrywhart (commented):
One small suggestion, but looks pretty good to me.

@tunetheweb: Would you like to review as well? Holding off approving until I hear from you.

tunetheweb:
> @tunetheweb: Would you like to review as well? Holding off approving until I hear from you.

Won’t be near a computer for next few days so happy to go with your approval.

One slight nit from a quick look on phone is all test cases are t-sql, which uses different quotes than most. Might be good to have one ANSI I’ve too. But other than that great set of test cases.

barrywhart:
@jpers36: The other reviewer suggested adding at least one test case that doesn't use TSQL. Could you add an ANSI test case to address that?

This raises another question -- are there any databases where quotes are more than one character long? If so, code like this wouldn't work correctly:
```
context.segment.raw.upper()[1:-1]
```

barrywhart (changes requested):
Request for one new test case. Also a potential question about potentially other dialects using oddball quotes.

tunetheweb (approved):
Think this is good to merge. @barrywhart anything holding it up from your end?

barrywhart (approved):

Issue #1781, opened 2021-10-29
Redshift Create Temp Table Parsing Errors


tdstark:
## Expected Behaviour
While linting Redshift SQL from the command line, SQLFluff ought to be able to parse `CREATE TEMP TABLE ... AS ()` statements using the `redshift` dialect.

## Observed Behaviour
SQLFluff throws unparsable error: `Found unparsable section:`.

## Steps to Reproduce
Using the `redshift` dialect command line arg, run a SQL statement similar to:
```
DROP TABLE IF EXISTS tmp_table;
CREATE TEMP TABLE tmp_table AS (
    SELECT
         campaign_id,
         recipient_type
    FROM (
             SELECT
                 campaign_id,
                 recipient_type,
                 (
                     ROW_NUMBER() OVER (
                        PARTITION BY campaign_id ORDER BY created_at DESC
                    )
                 ) AS sequence
             FROM table2
        )
    WHERE sequence = 1
);
```

## Dialect
Redshift

## Version
sqlfluff version  `0.7.1`, Python version `3.6.9`

## Configuration
Default configuration

PR #1785, opened 2021-10-30
Add CREATE TABLE AS to Postgres and Redshift

tdstark:
### Summary
This PR fixes issue #1781 by adding CREATE TABLE AS into the Postgres dialect. These changes will be inherited by Redshift dialect. In addition, test cases were added to check for compatibility with and without parenthesis surrounding the select statement.

In the future, `TABLE`, `VALUES`, AND `EXECUTE` statements will need to be added as options along with `SELECT`. However, this will likely cover the bulk of use cases for now.

At this time, there are no known side effects.

### DIFF
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -958,10 +958,8 @@ class CreateTableAsStatementSegment(BaseSegment):
         AnyNumberOf(
             Sequence(
                 Bracketed(
-                    Delimited(
-                            Ref("ColumnReferenceSegment")
-                        ),
-                    ),
+                    Delimited(Ref("ColumnReferenceSegment")),
+                ),
                 optional=True,
             ),
             Sequence("USING", Ref("FunctionSegment"), optional=True),
@@ -975,14 +973,14 @@ class CreateTableAsStatementSegment(BaseSegment):
                                 Sequence(
                                     Ref("EqualsSegment"),
                                     Ref("LiteralGrammar"),
-                                    optional=True
+                                    optional=True,
                                 ),
                             )
                         )
                     ),
                 ),
                 Sequence("WITHOUT", "OIDS"),
-                optional=True
+                optional=True,
             ),
             Sequence(
                 "ON",
@@ -990,7 +988,7 @@ class CreateTableAsStatementSegment(BaseSegment):
                 OneOf(Sequence("PRESERVE", "ROWS"), Sequence("DELETE", "ROWS"), "DROP"),
                 optional=True,
             ),
-            Sequence("TABLESPACE", Ref("ParameterNameSegment"), optional=True)
+            Sequence("TABLESPACE", Ref("ParameterNameSegment"), optional=True),
         ),
         "AS",
         OptionallyBracketed(
@@ -998,10 +996,10 @@ class CreateTableAsStatementSegment(BaseSegment):
                 Ref("SelectableGrammar"),
                 Sequence("TABLE", Ref("TableReferenceSegment")),
                 Ref("ValuesClauseSegment"),
-                Sequence("EXECUTE", Ref("FunctionSegment"))
+                Sequence("EXECUTE", Ref("FunctionSegment")),
             )
         ),
-        Sequence("WITH", Ref.keyword("NO", optional=True), "DATA", optional=True)
+        Sequence("WITH", Ref.keyword("NO", optional=True), "DATA", optional=True),
     )
 
 
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -944,6 +944,7 @@ class CreateTableAsStatementSegment(BaseSegment):
     match_grammar = Sequence(
         "CREATE",
         OneOf(
+            Ref("TemporaryGrammar"),
             Sequence(
                 OneOf("GLOBAL", "LOCAL"),
                 Ref("TemporaryGrammar"),
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -991,13 +991,11 @@ class CreateTableAsStatementSegment(BaseSegment):
             Sequence("TABLESPACE", Ref("ParameterNameSegment"), optional=True),
         ),
         "AS",
-        OptionallyBracketed(
-            OneOf(
-                Ref("SelectableGrammar"),
-                Sequence("TABLE", Ref("TableReferenceSegment")),
-                Ref("ValuesClauseSegment"),
-                Sequence("EXECUTE", Ref("FunctionSegment")),
-            )
+        OneOf(
+            OptionallyBracketed(Ref("SelectableGrammar")),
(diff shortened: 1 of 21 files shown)


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -932,6 +932,77 @@ class CreateTableStatementSegment(BaseSegment):
...
+                "COMMIT",
+                OneOf("PRESERVE ROWS", "DELETE ROWS", "DROP"),
+                optional=True,
+            ),
+            Sequence("TABLESPACE", Ref("ParameterNameSegment"), optional=True),
+        ),
+        "AS",
+        OneOf(
+            OptionallyBracketed(Ref("SelectableGrammar"))
+            # TODO Add TABLE, VALUES, AND EXECUTE commands
```
tunetheweb, line 1000: These might be as easy as this to add:

```suggestion
            Sequence("TABLE", Ref("TableReferenceSegment"),
            Ref("ValuesClauseSegment"),
            Sequence("EXECUTE", Ref("FunctionSegment")),
```

Though should add some test examples if willing to add this.

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -932,6 +932,77 @@ class CreateTableStatementSegment(BaseSegment):
...
+    """
+
+    type = "create_table_as_statement"
+
+    match_grammar = Sequence(
+        "CREATE",
+        OneOf(
+            Sequence(
+                OneOf("GLOBAL", "LOCAL", optional=True),
+                Ref("TemporaryGrammar", optional=True),
```
WittierDinosaur, line 949: This optional shouldn't be true. You can't have global/local without temporary grammar.

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -932,6 +932,77 @@ class CreateTableStatementSegment(BaseSegment):
...
+        "TABLE",
+        Ref("IfNotExistsGrammar", optional=True),
+        Ref("TableReferenceSegment"),
+        AnyNumberOf(
+            Sequence(
+                Bracketed(
+                    Delimited(
+                        Sequence(
+                            Ref("ColumnReferenceSegment"),
+                            AnyNumberOf(Ref("ColumnConstraintSegment")),
```
WittierDinosaur, line 963: I don't think CREATE TABLE AS supports column constraints, just renaming of the column name

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -932,6 +932,77 @@ class CreateTableStatementSegment(BaseSegment):
...
+            ),
+            Sequence("USING", Ref("FunctionSegment"), optional=True),
+            OneOf(
+                Sequence(
+                    "WITH",
+                    Bracketed(
+                        AnyNumberOf(
+                            Sequence(
+                                Ref("ParameterNameSegment"),
+                                Sequence(
```
WittierDinosaur, line 977: This second nested sequence isn't needed, just put those terms in the one above

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -932,6 +932,77 @@ class CreateTableStatementSegment(BaseSegment):
...
+                        )
+                    ),
+                    optional=True,
+                ),
+                Sequence("WITHOUT", "OIDS", optional=True),
+            ),
+            Sequence(
+                "ON",
+                "COMMIT",
+                OneOf("PRESERVE ROWS", "DELETE ROWS", "DROP"),
```
WittierDinosaur, line 992: PRESERVE ROWS and DELETE ROWS should be sequences

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: I think this is worth adding in this PR, as it's not a complicated change

WittierDinosaur (changes requested):
Looks good for the most part. See comments for changes. Also, it'd be good to add a bunch more test cases for the various parts of the statement

src/sqlfluff/dialects/dialect_postgres.py
tdstark: @WittierDinosaur This is replicating what exists in `CreateTableStatement`. Would the following work?: 
```
        OneOf(
            Ref("TemporaryGrammar"),
            Sequence(
                OneOf("GLOBAL", "LOCAL"),
                Ref("TemporaryGrammar"),
            ),
            "UNLOGGED",
            optional=True,
        ),
```

src/sqlfluff/dialects/dialect_postgres.py
tdstark: @WittierDinosaur This is also replicating what exists in the `CreateTableSegmentStatement`. I can remove it but the logic is the same according to the documentation.

src/sqlfluff/dialects/dialect_postgres.py
tdstark: This has been added.

src/sqlfluff/dialects/dialect_postgres.py
tdstark: Agree. This has been changed.

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: All you need to do is remove the optional=True from Ref(TemporaryGrammar), leave it in for OneOf(GLOBAL, LOCAL)

src/sqlfluff/dialects/dialect_postgres.py
WittierDinosaur: It's probably unnecessary if CreateTable too. For clarity, I'm just talking about de-nesting the inner sequence, not getting rid of what it contains. Bring the EqualsSegment and LiteralGrammar to the same level as ParameterNameSegment

src/sqlfluff/dialects/dialect_postgres.py
tdstark: Done.

src/sqlfluff/dialects/dialect_postgres.py
tdstark: It should be removed from both now. It didn't seem to have any affect on the tests.

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -911,11 +911,9 @@ class CreateTableStatementSegment(BaseSegment):
...
                             Sequence(
                                 Ref("ParameterNameSegment"),
-                                Sequence(
-                                    Ref("EqualsSegment"),
-                                    Ref("LiteralGrammar"),
-                                    optional=True,
-                                ),
+                                Ref("EqualsSegment"),
+                                Ref("LiteralGrammar"),
+                                optional=True,
```
WittierDinosaur, line 916: I've just reread the docs, and realised that asking you to change this here and in createtableas was bad advice 🤦‍♂️ . Can you change it back? And then I think we're good to go.

tdstark: Done!

WittierDinosaur (approved):

PR #1803, opened 2021-11-01
L019 plus L034 corrupts SQL

### LINKED ISSUE #1734: TSQL linting fix: L019+L034 breaks the code
The following code breaks when fixed:

```
    SELECT
      [ID]
    , CASE
        WHEN [ID] = 1 THEN NULL
        ELSE 1
        END AS [CaseOutput]
    , [DataDate]
    FROM #temp1
```

## Expected Behaviour

Workable code.

## Observed Behaviour

```
SELECT
    [ID],
    [DataDate]
    CASE
        WHEN [ID] = 1 THEN NULL
        ELSE 1
    END AS [CaseOutput],
FROM #temp1
```

## Steps to Reproduce

Run fix on the above

## Dialect

TSQL

## Version

most recent main
Python 3.9.7

## Configuration
```
```

barrywhart:

Fixes #1734

No.

### Pull Request checklist

  - Other.


src/sqlfluff/rules/L019.py
```diff
@@ -176,11 +186,11 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
                                 for d in memory["whitespace_deletions"]
                             ],
                             LintFix(
-                                "edit",
-                                last_code_seg,
-                                # Reuse the previous leading comma violation to
-                                # create a new trailing comma
-                                [last_code_seg, last_leading_comma_seg],
+                                "create",
```
barrywhart, line 189: This is the fix. If we "edit", it corrupts the parse tree by making the newly placed comma segment a _child_ of the code segment. This is wrong. It should be a sibling. We achieve this by using "create" instead. This requires using a different anchor: the segment _following_ the last code segment, not the last code segment itself.

src/sqlfluff/rules/L019.py
```diff
@@ -67,7 +66,7 @@ def _last_comment_seg(raw_stack):
         return None
 
     @staticmethod
-    def _last_code_seg(raw_stack):
+    def _last_code_seg(raw_stack: Tuple[RawSegment, ...]) -> Optional[RawSegment]:
```
barrywhart, line 69: Just adding type hints; no change in functionality.

src/sqlfluff/rules/L019.py
```diff
@@ -77,6 +76,17 @@ def _last_code_seg(raw_stack):
                 return segment
         return None
 
+    @staticmethod
```
barrywhart, line 79: New helper function.

test/fixtures/linter/autofix/ansi/022_l019_corrupts_parse_tree_and_causes_l034_to_corrupt_sql/after.sql
```diff
@@ -0,0 +1,5 @@
+SELECT
```
barrywhart, line 1: Previously, the bug was corrupting the SQL, leaving commas in the wrong place.

src/sqlfluff/rules/L019.py
```diff
@@ -123,7 +133,7 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
             # A comma preceded by a new line == a leading comma
             if context.segment.is_type("comma"):
                 last_seg = self._last_code_seg(context.raw_stack)
-                if last_seg.is_type("newline"):
+                if last_seg and last_seg.is_type("newline"):
```
barrywhart, line 136: Checking that `last_seg` is not `None`. mypy wanted this, and this may have fixed a bug (I didn't study closely).

src/sqlfluff/rules/L019.py
```diff
@@ -77,6 +76,16 @@ def _last_code_seg(raw_stack):
                 return segment
         return None
 
+    @staticmethod
+    def _follows_seg(
```
tunetheweb, line 80: Not loving the name here. I'm not a Python developer but is there a better name? Like `_get_following_seg` or `get_next_seg`?

tunetheweb (approved):
One suggestion. But maybe this is the convention in Python so happy to go with that you have if you think it's right as is.

src/sqlfluff/rules/L019.py
barrywhart: 👍  I'll change it to `_get_following_seg`.

PR #1813, opened 2021-11-03
Spark3: Add CREATE VIEW support

### LINKED ISSUE #1802: CREATE TEMPORARY VIEW not supported in Spark3 dialect
Following syntax is not supported in Spark3 dialect
`CREATE OR REPLACE TEMPORARY VIEW temp AS( SELECT * FROM temp2);`

## Expected Behaviour
Linting should be successful since it is [valid Spark3 syntax](https://spark.apache.org/docs/3.0.0/sql-ref-syntax-ddl-create-view.html#syntax)

## Observed Behaviour
Found unparsable section: 'CREATE OR REPLACE TEMPORARY VIEW temp AS (\n ...'

## Steps to Reproduce
- Try `sqlfluff lint` with above query

## Dialect
Spark3
## Version
sqlfluff, version 0.7.1

## Configuration
```
[tool.sqlfluff.core]
dialect = "spark3"


[tool.sqlfluff.templater.jinja]
apply_dbt_builtins = "False"

[tool.sqlfluff.templater.jinja.context.params]
anchor_date = "2020-01-01"
```

DipeshCS:
fixes #1802

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -113,9 +113,13 @@
         Ref("NotEqualToSegment_b"),
         Ref("LikeOperatorSegment"),
     ),
+    TemporaryGrammar=Sequence(
+        Sequence("GLOBAL", optional=True),
+        OneOf("TEMP", "TEMPORARY", optional=True),
```
tunetheweb, line 118: I don't think this should be optional or you could have `GLOBAL` without `TEMP` or `TEMPORARY`.

```suggestion
        OneOf("TEMP", "TEMPORARY"),
```

Plus the whole block should be wrapped in an optional when called.

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -113,9 +113,13 @@
...
+        optional=True,
```
tunetheweb, line 119: I don't think you need this as it's optional when added:

```suggestion
```

tunetheweb (changes requested):
Some small tweaks but in general looks good.

Nice set of test cases!

src/sqlfluff/dialects/dialect_spark3.py
DipeshCS: Thanks for the review and suggestions!

tunetheweb (approved):

PR #1788, opened 2021-10-31
Update DbtTemplater to use JinjaTracer

barrywhart:
### Brief summary of the change made

Fixes #1783

This PR builds on #1678. That one adds `JinjaTracer` and updates the _Jinja_ templater to use it, and this one updates the dbt templater to use it. **We may want to consider delaying merge of this PR until we've had a release with that one. More people use dbt, and dbt users are often less technical, so this strategy lets us do a sort of "phased" release.**

### Are there any other side effects of this change that we should be aware of?
...

### Pull Request checklist

- [X] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `python test/generate_parse_fixture_yml.py` or by running `tox` locally).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #1826, opened 2021-11-04
Fix: Add missing init file to `sqlfluff.core.templaters.slicers`

CrossNox:
Running `sqlfluff --help` (or any CLI command) fails with `ModuleNotFoundError: No module named 'sqlfluff.core.templaters.slicers'` due to a missing `__init__.py` file on `src/sqlfluff/core/templaters/slicers`.

`None`


CrossNox:
I just noticed that [1788](https://github.com/sqlfluff/sqlfluff/pull/1788) adds this file, but IMHO this quick fix can be isolated into a small PR and then merge those changes, so that `main` has a runnable version.

tunetheweb:
Can you add the comment like #1788 does to prevent linting failures.

@barrywhart you happy for this to go ahead of your PR?

barrywhart:
Sure, go ahead. Not sure when mine will be finished.

CrossNox:
> Can you add the comment like #1788 does to prevent linting failures.
> 
> @barrywhart you happy for this to go ahead of your PR?

Added! @barrywhart I couldn't find a way to checkout the file from your fork so that it appeared as authored by you, sorry.

barrywhart:
No worries!

WittierDinosaur:
Perhaps it's worth adding a test case for this?

CrossNox:
@WittierDinosaur do you mean testing the CLI or testing imports?

tunetheweb:
Presume he means to prevent this issue happening again in the future where a new module is added without a `__init__.py` when it's needed?

Would be good if we could add that. But think out of scope of this PR so let's merge this.

tunetheweb (approved):

PR #1838, opened 2021-11-06
Updated L009 logic to only allow a single trailing newline.

jpy-git:
Currently the logic of L009 will only add a trailing newline if none already exist, however, it will not remove excess trailing newlines. This can lead to files passing linting with any number of trailing new lines.
I have updated the logic to also remove excess trailing newlines so that we only have a single trailing newline (similar to black).

New logic should not impact the current effect of L009 adding a missing trailing newline, but also removes excess trailing newlines.

### Pull Request checklist

  - Other.


tunetheweb (changes requested):
This looks pretty good, though still going through the Python code.

One request: Could we move the tests from `test/rules/std_L009_test.py` to `test/fixtures/rules/std_rule_cases/L009.yml`? I think we should only use the Python when we have to more complex cases than pass and fail, which is all we seem to be doing here.

jpy-git:
@tunetheweb Moved those unit tests over to yaml :)

tunetheweb:
> @tunetheweb Moved those unit tests over to yaml :)

So much easier to read!

src/sqlfluff/rules/L009.py
```diff
@@ -95,11 +100,39 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
-        # no segment after it to match on. Otherwise we would never get a match!
-        return LintResult(
-            anchor=context.segment,
-            fixes=[
-                LintFix("edit", context.segment, [context.segment, NewlineSegment()])
-            ],
-        )
+        # Include current segment for complete stack.
+        complete_stack = list(context.raw_stack)
+        complete_stack.append(context.segment)
```
jpy-git, line 105: @tunetheweb mypy is throwing an error here as the first line implies `List[RawSegment]`
and then I append a `BaseSegment` in the following line. Given that `RawSegment` inherits `BaseSegment` and I'm only touching the inherited properties/methods I'll in-line annotate this `List[BaseSegment]`. Unless there's a raw_to_base (or vice versa) method I've missed.

tunetheweb: Paging @barrywhart and @alanmcruickshank who know much more about this than me!

src/sqlfluff/rules/L009.py
```diff
@@ -36,6 +37,14 @@ class Rule_L009(BaseRule):
         FROM foo
         ;$
 
+        -- Ending on more than a single newline.
```
barrywhart, line 40:
```suggestion
        -- Ending with multiple newlines.
```

test/fixtures/rules/std_rule_cases/L009.yml
```diff
@@ -0,0 +1,12 @@
+rule: L009
```
barrywhart, line 1: Wow, I'm surprised this rule didn't have any `.yml` tests before now!

src/sqlfluff/rules/L009.py
```diff
@@ -95,11 +101,39 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
-        # We're going to make an edit because we're appending to the end and there's
-        # no segment after it to match on. Otherwise we would never get a match!
-        return LintResult(
-            anchor=context.segment,
-            fixes=[
-                LintFix("edit", context.segment, [context.segment, NewlineSegment()])
-            ],
-        )
+        # Include current segment for complete stack.
+        complete_stack: List[BaseSegment] = list(context.raw_stack)
```
barrywhart, line 105: This looks like a good fix for the earlier mypy issue. It's a bit confusing that `raw_stack` is a list of `RawSegment` rather than `BaseSegment`. That is, precisely speaking, the correct type, but it frequently requires some fussy type hints or casting because rule code frequently works with other segment types. I _almost_ wonder if we should change `context.raw_stack` to `List[BaseSegment]`. It's still correct, and would require fewer of these casts. (Out of scope for this PR, just thinking...)

barrywhart (approved):
Looks good. One suggestion for changing the wording in the examples, just a nitpick.

barrywhart:
Note there's a dbt test failing. Will need to fix that before merge.

src/sqlfluff/rules/L009.py
jpy-git: Agreed, plus at the moment `raw_stack` is returned as a `tuple` whereas a `list` might make more sense for a stack. We can stick with this for now though, given that the type mixing is scoped to this function and mypy would highlight if someone then tried to add a `RawSegment` only bit of code in the next couple lines.

jpy-git:
> Note there's a dbt test failing. Will need to fix that before merge.

@barrywhart I believe I've found the failing dbt test and fixed, but having issues installing and running the dbt tests locally due to https://github.com/dbt-labs/dbt-core/issues/4185 so will take a look at it tomorrow :)

test/fixtures/rules/std_rule_cases/L009.yml
jpy-git: Better late than never!

tunetheweb (approved):

tunetheweb:
> > Note there's a dbt test failing. Will need to fix that before merge.
> 
> @barrywhart I believe I've found the failing dbt test and fixed, but having issues installing and running the dbt tests locally due to [dbt-labs/dbt-core#4185](https://github.com/dbt-labs/dbt-core/issues/4185) so will take a look at it tomorrow :)

Looks like the fix worked! Merging this.

PR #1840, opened 2021-11-07
New Rule L050: No leading whitespace

jpy-git:

This is a new rule which lints + fixes SQL beginning with newlines/whitespace. The logic is that the SQL should start with either a comment or the code. See gif below for demonstration.
Screen-Recording-2021-11-07-at-13 46 48

This adds new rule so doesn't impact existing rules.

### Pull Request checklist

  - Other.

I've added the docstrings/unit tests. I believe the documentation is updated from the the docstrings but happy to add any additional required documentation if I've missed anything 😄

### DIFF
--- a/src/sqlfluff/rules/L050.py
+++ b/src/sqlfluff/rules/L050.py
@@ -85,18 +85,26 @@ class Rule_L050(BaseRule):
         # If parent_stack is empty we are currently at FileSegment.
         if len(context.parent_stack) == 0:
             return None
+        
+        # If raw_stack is empty there can be nothing to remove.
+        if len(context.raw_stack) == 0:
+            return None
 
         # If the current segment is either comment or code and all
         # previous segments are forms of whitespace then we can
         # remove these earlier segments.
         # Given the tree stucture, we make sure we are at the
         # first leaf to avoid repeated detection.
+        whitespace_set = {"newline", "whitespace", "Indent", "Dedent"}
         if (
-            (context.segment.is_comment or context.segment.is_code)
-            & (len(context.raw_stack) > 0)
+            # Non-whitespace segment.
+            (context.segment.name not in whitespace_set)
+            # We want first Non-whitespace segment so
+            # all preceding segments must be whitespace.
             & set(segment.name for segment in context.raw_stack).issubset(
-                {"newline", "whitespace", "Dedent"}
+                whitespace_set
             )
+            # Found leaf of parse tree.
             & (not context.segment.is_expandable)
         ):
             # It is possible that a template segment (e.g. {{ config(materialized='view') }})
--- a/src/sqlfluff/rules/L050.py
+++ b/src/sqlfluff/rules/L050.py
@@ -65,7 +65,6 @@ class Rule_L050(BaseRule):
 
         Returns:
             :obj:`bool` indicating a preceding templated raw slice has been detected.
-
         """
         templated_file = context.segment.pos_marker.templated_file
         for segment in context.raw_stack:
--- a/src/sqlfluff/rules/L050.py
+++ b/src/sqlfluff/rules/L050.py
@@ -1,7 +1,7 @@
 """Implementation of Rule L050."""
-from typing import Optional, Tuple
+from typing import Optional
 
-from sqlfluff.core.parser.segments.raw import RawSegment
+from sqlfluff.core.parser.segments.base import BaseSegment
 from sqlfluff.core.rules.base import BaseRule, LintResult, LintFix, RuleContext
 from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 
@@ -61,28 +61,22 @@ class Rule_L050(BaseRule):
     """
 
     @staticmethod
-    def _check_unaccounted_slice_gaps(raw_stack: Tuple[RawSegment, ...]) -> bool:
-        """Check for missing segments by comparing neighbouring templated slices in the raw stack.
+    def _potential_template_collision(segment: BaseSegment) -> bool:
+        """Check for any source only slices that occur at or before the supplied segment.
 
         Returns:
-            :obj:`bool` indicating a missing segment discovered.
+            :obj:`bool` indicating a preceding source only slice detected.
 
         """
-        missing_segment_flag = False
-        for i, segment in enumerate(raw_stack):
-            if (i == 0) and (segment.pos_marker.templated_slice.start != 0):
-                # First segment does not start at position 0.
-                missing_segment_flag = True
-            elif i > 0:
-                prior_segment = raw_stack[i - 1]
-                if (
-                    segment.pos_marker.templated_slice.start
-                    != prior_segment.pos_marker.templated_slice.stop
-                ):
-                    # There is an unaccounted for gap in templated slices.
-                    missing_segment_flag = True
+        source_only_slices = segment.pos_marker.templated_file.source_only_slices()
+        if source_only_slices:
+            if (
+                source_only_slices[0].source_idx
+                <= segment.pos_marker.source_slice.start
+            ):
+                return True
 
-        return missing_segment_flag
+        return False
 
     def _eval(self, context: RuleContext) -> Optional[LintResult]:
         """Files must not begin with newlines or whitespace."""
@@ -103,11 +97,12 @@ class Rule_L050(BaseRule):
             )
             & (not context.segment.is_expandable)
         ):
-            if self._check_unaccounted_slice_gaps(context.raw_stack):
-                # It is possible that a template segment (e.g. {{ config(materialized='view') }})
-                # renders to an empty string and as such is omitted from the parsed tree.
-                # We therefore should flag any unaccounted for gaps in the earlier templated slices
-                # and skip this rule to avoid risking collisions with template objects.
+            # It is possible that a template segment (e.g. {{ config(materialized='view') }})
+            # renders to an empty string and as such is omitted from the parsed tree.
+            # We therefore should flag if a source only slice occurs at or before the
+            # the current segment and skip this rule to avoid risking collisions
+            # with template objects.
+            if self._potential_template_collision(context.segment):
                 return None
 
             return LintResult(
--- a/src/sqlfluff/rules/L050.py
+++ b/src/sqlfluff/rules/L050.py
@@ -1,7 +1,6 @@
 """Implementation of Rule L050."""
 from typing import Optional
 
-from sqlfluff.core.parser.segments.base import BaseSegment
 from sqlfluff.core.rules.base import BaseRule, LintResult, LintFix, RuleContext
 from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 
--- a/src/sqlfluff/rules/L050.py
+++ b/src/sqlfluff/rules/L050.py
@@ -78,7 +78,6 @@ class Rule_L050(BaseRule):
             )
             & (not context.segment.is_expandable)
         ):
-            context.memory = {"first": True}
             return LintResult(
                 anchor=context.parent_stack[0],
                 fixes=[LintFix("delete", d) for d in context.raw_stack],
--- a/src/sqlfluff/rules/L050.py
+++ b/src/sqlfluff/rules/L050.py
@@ -84,7 +84,7 @@ class Rule_L050(BaseRule):
         # If parent_stack is empty we are currently at FileSegment.
         if len(context.parent_stack) == 0:
             return None
-        
+
         # If raw_stack is empty there can be nothing to remove.
         if len(context.raw_stack) == 0:
             return None
--- a/src/sqlfluff/rules/L050.py
+++ b/src/sqlfluff/rules/L050.py
@@ -100,9 +100,7 @@ class Rule_L050(BaseRule):
             (context.segment.name not in whitespace_set)
             # We want first Non-whitespace segment so
             # all preceding segments must be whitespace.
-            and set(segment.name for segment in context.raw_stack).issubset(
-                whitespace_set
-            )
+            and all(segment.name in whitespace_set for segment in context.raw_stack)
             # Found leaf of parse tree.
             and (not context.segment.is_expandable)
         ):
--- a/src/sqlfluff/rules/L050.py
+++ b/src/sqlfluff/rules/L050.py
@@ -49,7 +49,6 @@ class Rule_L050(BaseRule):
             a
         FROM
             foo
-
     """
 
     def _eval(self, context: RuleContext) -> Optional[LintResult]:
--- a/src/sqlfluff/rules/L050.py
+++ b/src/sqlfluff/rules/L050.py
@@ -1,0 +1,79 @@
+"""Implementation of Rule L050."""
+from typing import Optional
+
+from sqlfluff.core.rules.base import BaseRule, LintResult, LintFix, RuleContext
+from sqlfluff.core.rules.doc_decorators import document_fix_compatible
+
+
+@document_fix_compatible
+class Rule_L050(BaseRule):
+    """Files must not begin with newlines or whitespace.
+
+    | **Anti-pattern**
+    | The content in file begins with newlines or whitespace, the ^ represents the beginning of file.
+
+    .. code-block:: sql
+       :force:
+
+        ^
+
+        SELECT
+            a
+        FROM foo
+
+        -- Beginning on an indented line is also forbidden, the • represents space.
+
+        ••••SELECT
+        ••••a
+        FROM
+        ••••foo
+
+    | **Best practice**
+    | Start file on either code or comment, the ^ represents the beginning of file.
+
+    .. code-block:: sql
+       :force:
+
+
+        ^SELECT
+            a
+        FROM foo
+
+        -- Including an initial comment.
+
+        ^/*
+        This is a description of my SQL code.
+        */
+        SELECT
+            a
+        FROM
+            foo
+
+    """
+
+    def _eval(self, context: RuleContext) -> Optional[LintResult]:
+        """Files must not begin with newlines or whitespace."""
+        # If parent_stack is empty we are currently at FileSegment.
+        if len(context.parent_stack) == 0:
+            return None
+
+        # If the current segment is either comment or code and all
+        # previous segments are forms of whitespace then we can
+        # remove these earlier segments.
+        # Given the tree stucture, we make sure we are at the
+        # first leaf to avoid repeated detection.
+        if (
+            (context.segment.is_comment or context.segment.is_code)
+            & (len(context.raw_stack) > 0)
+            & set(segment.name for segment in context.raw_stack).issubset(
+                {"newline", "whitespace", "Dedent"}
+            )
+            & (not context.segment.is_expandable)
+        ):
+            context.memory = {"first": True}
+            return LintResult(
+                anchor=context.parent_stack[0],
+                fixes=[LintFix("delete", d) for d in context.raw_stack],
+            )
+
+        return None
--- a/src/sqlfluff/rules/L050.py
+++ b/src/sqlfluff/rules/L050.py
@@ -62,7 +62,7 @@ class Rule_L050(BaseRule):
 
     @staticmethod
     def _check_unaccounted_slice_gaps(raw_stack: Tuple[RawSegment, ...]) -> bool:
-        """Check for missing segments by comparing neighbouring source slices in the raw stack.
+        """Check for missing segments by comparing neighbouring templated slices in the raw stack.
 
         Returns:
             :obj:`bool` indicating a missing segment discovered.
@@ -70,16 +70,16 @@ class Rule_L050(BaseRule):
         """
         missing_segment_flag = False
         for i, segment in enumerate(raw_stack):
-            if (i == 0) and (segment.pos_marker.source_slice.start != 0):
+            if (i == 0) and (segment.pos_marker.templated_slice.start != 0):
                 # First segment does not start at position 0.
                 missing_segment_flag = True
             elif i > 0:
                 prior_segment = raw_stack[i - 1]
                 if (
-                    segment.pos_marker.source_slice.start
-                    != prior_segment.pos_marker.source_slice.stop
+                    segment.pos_marker.templated_slice.start
+                    != prior_segment.pos_marker.templated_slice.stop
                 ):
-                    # There is an unaccounted for gap in source slices.
+                    # There is an unaccounted for gap in templated slices.
                     missing_segment_flag = True
 
         return missing_segment_flag
@@ -106,7 +106,7 @@ class Rule_L050(BaseRule):
             if self._check_unaccounted_slice_gaps(context.raw_stack):
                 # It is possible that a template segment (e.g. {{ config(materialized='view') }})
                 # renders to an empty string and as such is omitted from the parsed tree.
-                # We therefore should flag any unaccounted for gaps in the earlier source slices
+                # We therefore should flag any unaccounted for gaps in the earlier templated slices
                 # and skip this rule to avoid risking collisions with template objects.
                 return None
 
--- a/src/sqlfluff/rules/L050.py
+++ b/src/sqlfluff/rules/L050.py
@@ -1,6 +1,7 @@
 """Implementation of Rule L050."""
-from typing import Optional
+from typing import Optional, Tuple
 
+from sqlfluff.core.parser.segments.raw import RawSegment
 from sqlfluff.core.rules.base import BaseRule, LintResult, LintFix, RuleContext
 from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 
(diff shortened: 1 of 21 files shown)


test/rules/std_L003_L036_L039_combo_test.py
```diff
@@ -4,12 +4,14 @@
 and L039 (unnecessary white space) can result in incorrect indentation.
 """
 
+import textwrap
+
 import sqlfluff
 
 
 def test__rules__std_L003_L036_L039():
```
jpy-git, line 12: One interesting thing to note here: This test started failing with this change which is to be expected. When using multiline docstrings to hold the test sql as below it is equivalent to having the SQL with a leading newline and indented e.g. 
```sql

    WITH example AS (
        SELECT
            my_id,
            other_thing,
            one_more
        FROM
            my_table
    )

    SELECT *
    FROM example
```
Agree that it's readable in the code, but to make the tests proper we can escape the first newline and use `textwrap.dedent` to remove the indents.

src/sqlfluff/rules/L050.py
```diff
@@ -0,0 +1,79 @@
...
+        -- Including an initial comment.
+
+        ^/*
+        This is a description of my SQL code.
+        */
+        SELECT
+            a
+        FROM
+            foo
+
```
tunetheweb, line 51:
```suggestion
```

Should we also show an example with `--`?

```sql
-- This is a description of my SQL code
SELECT
    a
...
```

test/fixtures/rules/std_rule_cases/L050.yml
```diff
@@ -0,0 +1,15 @@
+rule: L050
+
+test_pass_leading_whitespace_statement:
+  pass_str: "SELECT foo FROM bar\n"
+
+test_pass_leading_whitespace_comment:
+  pass_str: "/*I am a comment*/\nSELECT foo FROM bar\n"
+
```
tunetheweb, line 8:
```suggestion
test_pass_leading_whitespace_comment:
  pass_str: "/*I am a comment*/\nSELECT foo FROM bar\n"

test_pass_leading_whitespace_inline_comment:
  pass_str: "--I am a comment\nSELECT foo FROM bar\n"

test_pass_leading_whitespace_inline_comment_hash:
  pass_str: "# I am a comment\nSELECT foo FROM bar\n"
  configs:
    core:
      dialect: bigquery

```

src/sqlfluff/rules/L050.py
```diff
@@ -0,0 +1,79 @@
...
+    .. code-block:: sql
+       :force:
+
+        ^
+
+        SELECT
+            a
+        FROM foo
+
+        -- Beginning on an indented line is also forbidden, the • represents space.
```
tunetheweb, line 24: Avoid this unnecessarily overflowing to two lines:

```suggestion
        -- Beginning on an indented line is also forbidden,
        -- (the • represents space).
```

tunetheweb (changes requested):
Some nits and suggested extra test cases.

Can you also look at the DBT errors? No need to install DBT as logs in GitHub usually sufficient (or can run that specific `tox` command locally which will look after all the installs for you)

test/rules/std_L003_L036_L039_combo_test.py
tunetheweb: I'm OK with this. As long as we've caught all the examples - is that perhaps why the DBT tests are failing? I've not looked.

tunetheweb:
> I've added the docstrings/unit tests. I believe the documentation is updated from the the docstrings but happy to add any additional required documentation if I've missed anything 😄

That's correct. Look at the `docs/README.md` file to see how to run the docs web server locally to make sure it looks OK (I did and think it does btw, so just tip for the future).

jpy-git:
This rule has an additional useful benefit. Since it effectively anchors the start of the query to the top left of the file it then allows L003 to correctly align the rest of the query with the left of the page! See example below (1st is current release, 2nd is this PR):
Screen-Recording-2021-11-07-at-16 03 11

test/rules/std_L003_L036_L039_combo_test.py
jpy-git: @tunetheweb It's likely why the dbt tests are failing I'm looking at the GHA logs now

src/sqlfluff/rules/L050.py
```diff
@@ -0,0 +1,79 @@
...
+@document_fix_compatible
+class Rule_L050(BaseRule):
+    """Files must not begin with newlines or whitespace.
+
+    | **Anti-pattern**
+    | The content in file begins with newlines or whitespace, the ^ represents the beginning of file.
+
+    .. code-block:: sql
+       :force:
+
```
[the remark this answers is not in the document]
jpy-git: @tunetheweb I'm not too sure, the other rules seem to keep the newline before the SQL so just stuck with that. Happy to take it out if you want though?

src/sqlfluff/rules/L050.py
jpy-git: I'll add in that extra example

src/sqlfluff/rules/L050.py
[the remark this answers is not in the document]
tunetheweb: Oh I checked a few and thought it was the opposite. Guess we’re inconsistent (or I misread them - was in a bit of a rush as was logging off!). In which case this is fine.

test/fixtures/rules/std_rule_cases/L050.yml
jpy-git: All good additions, hadn't considered the bigquery hash :)

jpy-git:
@tunetheweb , @barrywhart: I've found the dbt failure but would like your input as to how to resolve:

- The failing test is `test__dbt_templated_models_fix_does_not_corrupt_file` and is due to the `WITH` being moved up next to the template parameter at the top of the file:
image

- Since this parameter is not included in the parsed tree the rule just thinks the file starts with 2 new lines and adjusts the `WITH` position accordingly:
image.

I'm not too familiar with dbt/the templater side of the code so any help/pointers would be appreciated? 
Thanks.

tunetheweb:
Hmmm that’s a tough one that there’s not even a meta section in there. @barrywhart knows this better than me so might have some ideas.

The other thing you could do, since this rule is fairly simple, and doesn’t actually need the parsed code, is look at the pre-templated code (i.e. the original file before any dbt or Jinja logic is processed). Rule L016 uses some of the original code to test line length so might be worth a look at for inspiration. Though it needs to do a more complicated mapping for each line - you don’t need that at all and just need to read the original file to see if there is something other than whitespace at the beginning. So yours should be easier than that is.

barrywhart:
@tunetheweb's L016 suggestion sounds good to me. IIUC, the `config` block renders to the empty string. I think that's why there's no `meta` section produced (or maybe you only get a `meta` section for control structures like loops or if statements).

tunetheweb:
Rules don't have access to the original file directly do they @barrywhart ? Only via looking at the `source_slice` of the segments?

That's probably sufficient here (you could look at `source-slice.start` of the "first" segment you see and if it's greater than 1 (or 0?) then you know "something" came before it so your rule can end.

However I wonder would it be good longer term for the rules to have access to the original file (like they do for the templated file) for rules like this, or L009 (only one newline at end of file)?

barrywhart:
Yes, I think the slice info is enough here. I'm not sure if they have access to the original file. Is there a TemplatedFile object on the context? That's where it's stored.

tunetheweb:
> Is there a TemplatedFile object on the context? That's where it's stored.

There is a `TemplatedFile` - is that the original file? Or the file applied after template pieces have been applied? I'm always confused by that name...

jpy-smart:
So both the source and template slices start with the new line as position 0 after `{{ config(materialized='view') }}` so I can't detect the object using slices. Guess it's on to the `TemplatedFile`

jpy-git:
@tunetheweb Happy days, I've managed to find a method to locate any potential hidden template segments before the first comment/code block. 
The logic is scan the raw_stack for any source slices which intersect with templated raw slices and return None. 
Essentially the rule isn't applied when a template is detected within the initial whitespace as we don't want to risk overlapping the template with the moved code.

src/sqlfluff/rules/L050.py
```diff
@@ -0,0 +1,115 @@
...
+            foo
+    """
+
+    @staticmethod
+    def _potential_template_collision(context: RuleContext) -> bool:
+        """Check for any templated raw slices that intersect with source slices in the raw_stack.
+
+        Returns:
+            :obj:`bool` indicating a preceding templated raw slice has been detected.
+
```
tunetheweb, line 68: Nit but we tend not to have newlines at end of docstrings:

```suggestion
```

src/sqlfluff/rules/L050.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        templated_file = context.segment.pos_marker.templated_file
+        for segment in context.raw_stack:
+            source_slice = segment.pos_marker.source_slice
+            raw_slices = templated_file.raw_slices_spanning_source_slice(source_slice)
+            if any(
+                raw_slice
+                for raw_slice in raw_slices
+                if raw_slice.slice_type == "templated"
+            ):
```
tunetheweb, line 78: I've a few questions on this code (my Python is pretty poor so apologies if these are all non-issues):
1. Should the first `raw_slice` be `raw_slices`? If not how can you reference `raw_slice` before it's defined? Or is the point, it's false initially but `any()` will pick up one of the other ones? And then it might be true later?
2. There's nothing under the `for` loop? Or is that taken care of by the `any()`?
3. With the use of `any()` doe that mean the last line is not a necessity for this? Does that mean a non-templated piece could be seen as templated?

Sorry just finding it hard to grok this code. If this is correct, and common enough in Python, then happy to leave as is. Though might get @barrywhart to approve this PR then as a bit out of my comfort zone.

src/sqlfluff/rules/L050.py
```diff
@@ -0,0 +1,115 @@
...
+        if len(context.parent_stack) == 0:
+            return None
+
+        # If the current segment is either comment or code and all
+        # previous segments are forms of whitespace then we can
+        # remove these earlier segments.
+        # Given the tree stucture, we make sure we are at the
+        # first leaf to avoid repeated detection.
+        if (
+            (context.segment.is_comment or context.segment.is_code)
```
tunetheweb, line 95: Do we need to explicitly check for comments or code? What if we have some third type not covered by those (meta maybe?)? Or ever introduce one?

Why not just check for NOT whitespace?

src/sqlfluff/rules/L050.py
```diff
@@ -0,0 +1,115 @@
...
+            & (len(context.raw_stack) > 0)
+            & set(segment.name for segment in context.raw_stack).issubset(
+                {"newline", "whitespace", "Dedent"}
+            )
+            & (not context.segment.is_expandable)
```
tunetheweb, line 100: There's a LOT going on in this - four different IF clauses.

Would it be clearer if they were potentially nested into a few separate `if` statements rather than trying to squeeze it all into one?

```py
# Check if 
if condition1 and condition2:
  # Check if blah blah
  if condition3:
    # Check if blah blah blah
    if condition4:
      # Do something

return None
```

Or do you think clearer as one?

Also why use of `&` over `and`?

src/sqlfluff/rules/L050.py
```diff
@@ -0,0 +1,115 @@
...
+        ):
+            # It is possible that a template segment (e.g. {{ config(materialized='view') }})
+            # renders to an empty string and as such is omitted from the parsed tree.
+            # We therefore should flag if a templated raw slice intersects with the
+            # source slices in the raw stack and skip this rule to avoid risking
+            # collisions with template objects.
+            if self._potential_template_collision(context):
+                return None
```
tunetheweb, line 108: Should this be outside the `if` as an early exit? Or is this more computational expensive and/or less likely to be hit so better to check we're in the `if` statement first before hitting this?

test/fixtures/rules/std_rule_cases/L050.yml
```diff
@@ -0,0 +1,24 @@
...
+    core:
+      dialect: bigquery
+
+test_fail_leading_whitespace_statement:
+  fail_str: "\n  SELECT foo FROM bar\n"
+  fix_str: "SELECT foo FROM bar\n"
+
+test_fail_leading_whitespace_comment:
+  fail_str: "\n  /*I am a comment*/\nSELECT foo FROM bar\n"
+  fix_str: "/*I am a comment*/\nSELECT foo FROM bar\n"
```
tunetheweb, line 24: Given the problems with DBT (and seems those test cases cover this), I wonder should be add a `Jinja` test case.

```suggestion
  fix_str: "/*I am a comment*/\nSELECT foo FROM bar\n"
test_fail_leading_whitespace_jinja_comment:
  pass_str: "{# I am a comment #}\nSELECT foo FROM bar\n"
```

test/rules/std_L003_L036_L039_combo_test.py
```diff
@@ -4,12 +4,14 @@
 and L039 (unnecessary white space) can result in incorrect indentation.
 """
 
+import textwrap
+
```
tunetheweb, line 8: Do we need this new line:

```suggestion
```

tunetheweb (commented):
Some comments - my Python is very basic so feel free to push back on these if you disagree!

src/sqlfluff/rules/L050.py
jpy-git: So this is a pretty common use of a generator to find whether any element in an iterable satisfies a condition, it's very similar to list comprehension.

It means: return each `raw_slice` in the list of `raw_slices` only if that `raw_slice` satisfies the condition `raw_slice.slice_type == "templated"`. The `any` means that as soon as one raw_slice is returned from the generator then the condition is satisfied so no need to check the rest of `raw_slices` and enter the `if` block

tunetheweb: Cool. My lack of Python experience here showing. Resolving!

src/sqlfluff/rules/L050.py
jpy-git: the `&` is just personal preference, it's more concise and means the same thing as `and`. I could also replace the `or` with `|` for consistency or vice versa. I'm easy going on that.

src/sqlfluff/rules/L050.py
jpy-git: It relies on the current context being the first code/comment segment, which is caught by the outer if block. It can then do a simple check back over the raw stack to see if there are any overlapping templated raw slices. Putting it outside the block would require more complex checks/more hits, so definitely worth keeping it here.

test/rules/std_L003_L036_L039_combo_test.py
jpy-git: it's pretty standard to separate import groups into standard library, 3rd party , and local packages, each separated by a new line so I'd leave this one in here

barrywhart:
@tunetheweb:
>There is a TemplatedFile - is that the original file? Or the file applied after template pieces have been applied? I'm always confused by that name...

It's a class with both.
* source_str: Original source
* templated_str: Expanded source (after running Jinja or whatever)
* raw_sliced: List with slices of original source (pre-templater)
* sliced_file: List with slices of templated code (after templating). Also maps back to the original source

`TemplatedFile` is the "nerve center" for raw <---> templated code. This _may_ be one of the areas we could work on simplifying and improving in order to make it easier to write rules.

src/sqlfluff/rules/L050.py
jpy-git: looking into this :)

src/sqlfluff/rules/L050.py
tunetheweb: OK happy yo leave as is. With the suggestion above to stop looking a comments and code only, this might be simplified anyway.

However, I do think we should move from `&` to `and` for consistency though, as seems more common through our code base (plus, as you point out it's more consistent with `or`).

barrywhart: They actually _aren't_ quite the same: `&` operator is bitwise "and". Strongly prefer we change this to regular boolean `and`.

barrywhart (changes requested):

src/sqlfluff/rules/L050.py
jpy-git: @tunetheweb changed it to not whitespace

src/sqlfluff/rules/L050.py
jpy-git: @barrywhart oh interesting thanks, didn't realise! I've made the change to `and` :)

Additionally I've added comments by the conditions to make them clearer.

src/sqlfluff/rules/L050.py
```diff
@@ -0,0 +1,122 @@
...
+        # remove these earlier segments.
+        # Given the tree stucture, we make sure we are at the
+        # first leaf to avoid repeated detection.
+        whitespace_set = {"newline", "whitespace", "Indent", "Dedent"}
+        if (
+            # Non-whitespace segment.
+            (context.segment.name not in whitespace_set)
+            # We want first Non-whitespace segment so
+            # all preceding segments must be whitespace.
+            and set(segment.name for segment in context.raw_stack).issubset(
```
barrywhart, line 103: Would it be clearer to avoid constructing a temporary set by doing something like this?
```
all(segment.name in whitespace_set for segment in context.raw_stack)
```

test/rules/std_L003_L036_L039_combo_test.py
```diff
@@ -4,12 +4,14 @@
...
 import sqlfluff
 
 
 def test__rules__std_L003_L036_L039():
     """Verify that double indents don't flag L039."""
-    sql = """
+    sql = """\
```
barrywhart, line 14: What does the backslash do here? I'm not familiar with using them together with triple-quoted strings.

tunetheweb: I OKed this early. The backslash is to escape the newline to avoid flagging this new rule. It's also why we do a dedent before.

Probably should add a comment to explain this.

Open to alternatives.

```
sql="""WITH example AS (
        SELECT my_id,
...
```

Not quote as clean and nice but would avoid the backslash.

Or you any other suggestions?

Actually this test only needs L003, L036 and L039 so maybe better option is to ensure ONLY those rules are running so this isn't needed?

test/fixtures/rules/std_rule_cases/L050.yml
jpy-git: @tunetheweb @barrywhart I've added a jinja pass test.

With regards to a query like this:
```sql

{# I am a comment #}
SELECT foo FROM bar
```
It has leading whitespace and the template block is detected (this one is included in the parse tree) and given fixes to delete the preceding whitespace by rule L050 however these fixes then get discarded and not raised at the top level. 
I believe this happens in `BaseRule.discard_unsafe_fixes` to not take the risk of overlapping templates, so as far as I'm aware the fact that this specific example is unchanged is an intended effect?

tunetheweb: But would `"\n{# I am a comment #}\nSELECT foo FROM bar\n"` not fail? It's just it wouldn't have a `fix_str` as it's "unfixable" automatically.

So the fact the above test passes is still a good test?

I presume `pass_str` actually tests it passes and doesn't just do a "before fix" and "after fix" comparison so "passes" even if it's not passed?

src/sqlfluff/rules/L050.py
jpy-git: agree with you on this, making the change.

test/rules/std_L003_L036_L039_combo_test.py
jpy-git: yeah otherwise writing these queries as multiline strings effectively gives them unintended leading new lines (+extra indentation without `textwrap.dedent`).

Unfortunately that alternative is equivalent to:
```sql
WITH example AS (
••••••••SELECT my_id,
...
```
i.e. double indented and textwrap can't help there.

Personally I think the best approach is just write these queries in sql files (std_L003_L036_L039_input.sql & std_L003_L036_L039_expected.sql) and import them in the test?
```python
with open(<input_sql_path>, "r") as f:
    input_sql = f.read()

with open(<input_sql_path>, "r") as f:
    input_sql = f.read()
```
as this means we don't worry about messing around with backslashes/textwrap and it's clear exactly what indentations we're using in the test.
maybe put them in the folder `sqlfluff/test/fixtures/rules/custom/`

This only impact the python rules tests so there shouldn't be many.

src/sqlfluff/rules/L050.py
jpy-git: done

test/fixtures/rules/std_rule_cases/L050.yml
jpy-git: @tunetheweb Good news, I understand why `"\n{# I am a comment #}\nSELECT foo FROM bar\n"` is passing rule 50 and it's intentional behaviour.
Within `linter.py` there is a function that filters out all fixes to templated code section unless it is a rule that `targets_templates` (just L046):
image
OR if the user specifies that they want to lint templated areas (defaults to ignore them)
image

jpy-git: So now if I change ignore_templated_areas to false in .sqlfluff, `"\n{# I am a comment #}\nSELECT foo FROM bar\n"` can be linted/fixed :)

jpy-git: image

tunetheweb: `ignore_templated_areas` is not configurable. It's set per rule in the code.

This one should probably be set to true.

test/rules/std_L003_L036_L039_combo_test.py
tunetheweb: The API is about to have an `exclude_rules` option! Working on a PR.

This should make this easier.

test/fixtures/rules/std_rule_cases/L050.yml
jpy-git: Oh nevermind these yaml tests aren't testing the final outputed violations (post templated fix removals) but instead are just looking to see whether or not the violation is raised at any point in the lint, which it rightly is.
image

Guess I can also add a custom test to show the impact of config on this rule

jpy-git: so the pass version cares about the log not the output, but the fail version cares about the output not the log
image

test/rules/std_L003_L036_L039_combo_test.py
jpy-git: reset this test to how it originally was but updated result = `sqlfluff.fix(sql, rules=["L003", "L036", "L039"])` to not include L050

tunetheweb: If you want until this is merged: https://github.com/sqlfluff/sqlfluff/pull/1850 you can use:

```
sqlfluff.fix(sql, exclude_rules=["L050"]
```

jpy-git: sound good :)

test/fixtures/rules/std_rule_cases/L050.yml
jpy-git: > `ignore_templated_areas` is not configurable. It's set per rule in the code.
> 
> This one should probably be set to true.

You're thinking of `targets_templated`, I believe it's specifically for template related rules i.e. L046. This rule should work if I were to set that attribute to true but feel thats not the intention of the attribute.

`ignore_templated_areas` 100% comes from the config (see below). I've tested configuring it with .sqlfluff and can confirm that it is what causes L050 to be applied or not in the case of a leading template. I agree with the existing default config to leave it off by default and allow users to turn it on if they wish.
image

tunetheweb: Ah yes you’re right. Got the two confused.

At the moment what can be passed though the simple API is relatively limited (hence why I just added `exclude_rules`). I’m happy to leave this as a positive test and not a negative test for now.

test/rules/std_L003_L036_L039_combo_test.py
tunetheweb: That’s merged to `main` now if you wanna add to this branch and update the test.

test/fixtures/rules/std_rule_cases/L050.yml
jpy-git: Thanks for adding that `exclude_rules argument`, I'll account for that in my branch now

tunetheweb: > @tunetheweb yeah I was thinking we should really be able to pass a config object to the simple api too for easier testing, happy to pick that up as a separate PR :)

Actually I disagree. See reasoning here: https://github.com/sqlfluff/sqlfluff/pull/1419

jpy-git: Makes sense to me, sound reasoning.

I've accounted for the exclude_rules addition in std_L003_L036_L039_combo_test.py 👍

test/rules/std_L003_L036_L039_combo_test.py
jpy-git: I've accounted for the exclude_rules change

jpy-git:
@tunetheweb @barrywhart I believe I've accounted for all test cases/requested changes now?

barrywhart (approved):

tunetheweb (approved):
Thanks for working through all the feedback. Great job!

jpy-git:
Wonderful, thank you both 😄

barrywhart:
Great work!

PR #1842, opened 2021-11-07
Fix: add tqdm to setup.py installation requirements

skykasko:
At v0.8.0, running `pip install sqlfluff` and then invoking `sqlfluff --help` (or any other command) results in the following error: `ModuleNotFoundError: No module named 'tqdm'`.

This PR adds `tqdm` to the installation requirements in  `setup.py` to fix the bug.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -120,7 +120,7 @@ setup(
         # For returning exceptions from multiprocessing.Pool.map()
         "tblib",
         # For handling progress bars
-        "tqdm==4.62.3",
+        "tqdm",
     ],
     entry_points={
         "console_scripts": [
--- a/setup.py
+++ b/setup.py
@@ -119,6 +119,8 @@ setup(
         "toml",
         # For returning exceptions from multiprocessing.Pool.map()
         "tblib",
+        # For handling progress bars
+        "tqdm==4.62.3",
     ],
     entry_points={
         "console_scripts": [
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,4 +2,4 @@
 sphinx>=2.4.1
 sphinx-click
 doc8
-tqdm==4.62.3
+tqdm
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,4 +16,4 @@ dataclasses; python_version < '3.7'
 chardet
 types-chardet
 toml
-tqdm==4.62.3
+tqdm


setup.py
```diff
@@ -119,6 +119,8 @@ def read(*names, **kwargs):
         "toml",
         # For returning exceptions from multiprocessing.Pool.map()
         "tblib",
+        # For handling progress bars
+        "tqdm==4.62.3",
```
tunetheweb, line 123: Do we need to be so precise in the version?

```suggestion
        "tqdm>=4.62.3",
```

tunetheweb (changes requested):
Thanks for spotting and raising the solution to quickly!

One question.

setup.py
barrywhart: Agree. No pin or loose pin, depending on how stable the package's API is.

skykasko: Thanks for the quick reply! The precise specification of version 4.62.3 comes from [requirements.txt](https://github.com/sqlfluff/sqlfluff/blob/main/requirements.txt#L19) (and [docs/requirements.txt](https://github.com/sqlfluff/sqlfluff/blob/main/docs/requirements.txt#L5)). I'm afraid I don't know why the requirement was so specific, but I figured I should match what's in `requirements.txt` for this PR.

tunetheweb: I think that’s an oversight when we reviewed that. Would you mind removing the version from that too as part of this PR? Then we’ll merge this and release 0.8.1.

We generally try to avoid specific versions unless necessary to allow SQLFluff to play nicely with other projects.

tunetheweb: FYI @adam-tokarski

skykasko: The version of tqdm is no longer pinned in any of those files.

tunetheweb (approved):

adam-tokarski:
My bad. Apparently I'm used to installing that using `requirements` files, that I even didn't think about installing that as a package. Sorry for that.

is there any mechanism to catch such things except running such installation manually?

setup.py
adam-tokarski: There was no particular reason to pin that specific version except my habits. I should rather note that most of other packages also do not use packages versions.

Thanks for taking care of it.

tunetheweb:
Working on adding a test: https://github.com/sqlfluff/sqlfluff/pull/1843

tunetheweb:
v0.8.1 has been release with both this fix, and a test to avoid it happening in future.

Thanks again @skykasko for raising and fixing so quickly.

PR #1843, opened 2021-11-07
Add test to ensure pip install works

tunetheweb:
Fixes the recent issue with 0.8.0 whereby a new dependency was added but not added to setup.py so `pip install sqlfluff` failed.

This adds a test to our CI which does the following:
- `pip install git+https://github.com/@pull_request_author/sqlfluff@branch`
- `sqlfluff --version`

That is, it doesn't set up the full dev environment and instead mimics what users would see with a `pip install` to ensure that works, even if requirements.txt and requirements_dev.txt are not installed.

Proved this failed here when this branch was based on the broken v0.8.0: https://github.com/sqlfluff/sqlfluff/pull/1843/commits/5d3a11a2ce4fb60ae0a68b1c4c7504e923ca5c48
And that it passed when #1843 change was included: https://github.com/sqlfluff/sqlfluff/pull/1843/commits/0d9a2bc1165e26fa7619b5e4bab7d3b7c9ba15f2

An extra CI test, but it's very quick, and runs in parallel.

### Pull Request checklist

  - Other.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -189,7 +189,7 @@ jobs:
       uses: codecov/codecov-action@v1
       with:
         files: ./coverage.xml
-  pip-test-pull=request:
+  pip-test-pull-request:
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
     name: pip install tests for forks
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -195,9 +195,9 @@ jobs:
       matrix:
         include:
           - name: 'main'
-            location: https://github.com/sqlfluf/sqlfluff@main
+            location: sqlfluff/sqlfluff.git@main
           - name: 'branch'
-            location: https://github.com/${GITHUB_REPOSITORY}@${GITHUB_REF##*/}
+            location: ${GITHUB_REPOSITORY}.git@${GITHUB_REF##*/}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-tests.yml
```diff
@@ -189,4 +189,22 @@ jobs:
...
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.10'
+    - name: Install dependencies
+      run: |
+        pip install git+https://github.com/${{ github.event.pull_request.head.repo.full_name }}@${{ github.head_ref }}
+    - name: Run the tests
+      run: |
+        sqlfluff --version
```
barrywhart, line 208: Should we run a command that exercises more of SQLFluff's functionality? Python is a dynamic language, so it's not going to check all the `import` statements, only those in modules that get loaded.

barrywhart (approved):
Looks good. One suggestion for making it more thorough.

.github/workflows/ci-tests.yml
tunetheweb: Added a couple of more simple tests:

```yml
    - name: Run the version test
      run: |
        sqlfluff --version
    - name: Run a simple select parse test
      run: |
        echo "select 1" | sqlfluff parse -
    - name: Run a simple select lint test
      run: |
        echo "select 1" | sqlfluff lint -
```

PR #1869, opened 2021-11-09
Add OVERRIDING SYSTEM/USER VALUE to insert statement in postgres dialect

### LINKED ISSUE #1858: OVERRIDING keyword in postgres INSERT INTO statement is not parsable
## Expected Behaviour
SQLFluff should be able to parse postgres `INSERT INTO` statements that are valid according to the [docs](https://www.postgresql.org/docs/current/sql-insert.html).

## Observed Behaviour
SQLFluff fails to parse valid postgres statement `INSERT INTO test (id, col1) OVERRIDING SYSTEM VALUE VALUES (1, 'val');`

## Steps to Reproduce
Run in `bash`:
```bash
echo "INSERT INTO test (id, col1) OVERRIDING SYSTEM VALUE VALUES (1, 'val');" | sqlfluff lint --dialect=postgres -
```

## Dialect
postgres

## Version
sqlfluff, version 0.8.1
Python 3.7.6

## Configuration
default

jpy-git:

This PR fixes #1858. I have introduced `OVERRIDING SYSTEM/USER VALUE` to postgres insert statements as described in the linked issue and the postgres docs (https://www.postgresql.org/docs/13/sql-insert.html) (N.B. there are other additions to made from those docs in the future but that is out of scope of this PR).

None

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -2303,3 +2303,19 @@ class DropTriggerStatementSegment(BaseSegment):
...
+    type = "insert_statement"
+    match_grammar = StartsWith("INSERT")
+    parse_grammar = Sequence(
+        "INSERT",
+        "INTO",
+        Ref("TableReferenceSegment"),
+        Ref("BracketedColumnReferenceListGrammar", optional=True),
+        Sequence("OVERRIDING", OneOf("SYSTEM", "USER"), "VALUE", optional=True),
+        Ref("SelectableGrammar"),
+    )
```
jpy-git, line 2321:
```SQL
@ansi_dialect.segment()
class InsertStatementSegment(BaseSegment):
    """A `INSERT` statement."""

    type = "insert_statement"
    match_grammar = StartsWith("INSERT")
    parse_grammar = Sequence(
        "INSERT",
        Ref.keyword("OVERWRITE", optional=True),  # Maybe this is just snowflake?
        Ref.keyword("INTO", optional=True),
        Ref("TableReferenceSegment"),
        Ref("BracketedColumnReferenceListGrammar", optional=True),
        Ref("SelectableGrammar"),
    )
```
The current ANSI implementation for the reviewers convenience.

jpy-git: A couple questions on the ANSI version:
1. I couldn't find any evidence that the `OVERWRITE` keyword should be in any other dialect except Snowflake (as the comment says). This should be moved to Snowflake dialect. Happy to do this as part of this PR or make a seperate one.
2. I'm not aware of any dialect (certainly ansi/postgres/snowflake/big query) where the `INTO` keyword is optional? Is there an example of this and if so it should only be made optional in the relevant dialect.

jpy-git: I have corrected the 2 cases above in the postgres dialect but feel they should be updated in ANSI.

WittierDinosaur: So, our ANSI isn't strictly ANSI, it's kind of ANSI or stuff that is easier to just have in ANSI. If you can confirm the INTO isn't optional in any dialect (I'd check SQLite because it's weird), and it doesn't break tests, it's worth doing. Probably not worth removing overwrite from ANSI at this juncture

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -2303,3 +2303,19 @@ class DropTriggerStatementSegment(BaseSegment):
         Ref("TableReferenceSegment"),
         OneOf("CASCADE", "RESTRICT", optional=True),
     )
+
+
+@postgres_dialect.segment(replace=True)
+class InsertStatementSegment(BaseSegment):
+    """An `INSERT` statement."""
```
WittierDinosaur, line 2310: Can you add a reference to the Postgres 14 docs in your docstring?

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -2303,3 +2303,19 @@ class DropTriggerStatementSegment(BaseSegment):
...
+
+    type = "insert_statement"
+    match_grammar = StartsWith("INSERT")
+    parse_grammar = Sequence(
+        "INSERT",
```
WittierDinosaur, line 2315: Appreciate this isn't necessarily what you were going for in this PR, but if you fancy it, could you take a stab at replicating the entire Postgres insert grammar? If you don't want to do this, can you either have a TODO comment, or comment in the docstring that it is an incomplete representation of the grammar?

src/sqlfluff/dialects/dialect_postgres.py
jpy-git: Added in the doc reference :)

src/sqlfluff/dialects/dialect_postgres.py
jpy-git: Looking at the docs, the complete INSERT grammar looks like a bit of a beast so let's maybe finish this PR to address the related issue and then I can look at replicating the whole thing as a separate PR 😄  I've added in a note in the docstring that it is not a complete implementation.

src/sqlfluff/dialects/dialect_postgres.py
jpy-git: makes sense, I figured that was the case. Let me just take a quick look at the optional INTO.

jpy-git: Okay so the 2 culprits I can find are:
Big Query
image

Hive
image
(overwrite as well!)

jpy-git: @WittierDinosaur Hive already has a complete implementation of it's insert grammar (including it's optional INTO) so I've removed the optional INTO in ANSI and moved to big query. No tests are failing and I've also added an extra unit test to big query as well to check the optional INTO in big query.

WittierDinosaur: Good effort - LGTM

WittierDinosaur (approved):

jpy-smart:
@WittierDinosaur resolved the lint test so feel free to merge once the tests have completed 🚀

PR #1873, opened 2021-11-10
Fix progressbar artifacts within linter errors

### LINKED ISSUE #1841: Progress bar issue with lint errors
@adam-tokarski noticed this today with the progress bar feature.

When you get lint errors, they interspace the progress bar, leaving an old copy of the bar around:

Screenshot

It also makes the lints a tiny bit more difficult to read.

Is it possible to delete the progress bar lint before printing any lint error to avoid this?

adam-tokarski:
Fixes #1841 

These changes allowed to avoid disabling progressbar when in verbose mode, which is good I guess.

The printing (echoing/logging) was touched, so it may - but shouldn't - affect messages printed to console and/or file.

### Pull Request checklist

  - Other.


tunetheweb:
Tested this briefly and looks good to me - great work as thought we were going to have to live with it as it was!

Can you look at the coverage issues and either `# pragma: no cover` or add test cases? Then think we can merge.

tunetheweb (approved):

tunetheweb:
@barrywhart / @WittierDinosaur you wanna cast your more experienced Python eyes over this before we merge?

Functionally it works a lot better anyway in my opinion.

src/sqlfluff/cli/commands.py
```diff
@@ -62,6 +64,24 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+    Without it, there were left artifacts in DEBUG mode (not sure about another ones,
+    but probably would happen somewhere).
+    """
+
+    def emit(self, record: LogRecord) -> None:
+        """Behaves like original one except uses `tqdm` to write."""
+        try:
+            msg = self.format(record)
+            tqdm.write(msg, file=self.stream)
+            self.flush()
```
barrywhart, line 80: Just curious -- what happens if you omit this line?

barrywhart (approved):

barrywhart:
The code changes look great! Haven't tested it, but @tunetheweb did, so I feel good about this PR.

src/sqlfluff/cli/commands.py
adam-tokarski: Seems to be working as far I can tell, but it is in original `emit`, so I guess this is for some reason. So it safer just to put it there.

PR #1877, opened 2021-11-10
Added SET Statement to Postgres

### LINKED ISSUE #1832: SET statement not parsed in postgres dialect
Here is a minimal SQL example:
```
markpolyak$ echo "SET search_path TO my_schema, public;" | sqlfluff parse --dialect postgres -
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |        unparsable:                                           !! Expected: 'StatementSegment'
[L:  1, P:  1]      |            raw:                                              'SET'
[L:  1, P:  4]      |            whitespace:                                       ' '
[L:  1, P:  5]      |            raw:                                              'search_path'
[L:  1, P: 16]      |            whitespace:                                       ' '
[L:  1, P: 17]      |            raw:                                              'TO'
[L:  1, P: 19]      |            whitespace:                                       ' '
[L:  1, P: 20]      |            raw:                                              'my_schema'
[L:  1, P: 29]      |            comma:                                            ','
[L:  1, P: 30]      |            whitespace:                                       ' '
[L:  1, P: 31]      |            raw:                                              'public'
[L:  1, P: 37]      |    statement_terminator:                                     ';'
[L:  1, P: 38]      |    newline:                                                  '\n'

==== parsing violations ====
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'SET search_path TO
                       | my_schema, public'
```

## Expected Behaviour
`SET` statement should be parsed, the minimal example is from the postgres [docs](https://www.postgresql.org/docs/9.6/sql-set.html).

## Observed Behaviour
Perfectly valid `SET` statement causes PRS (unparsable section) in SQLFluff.

## Steps to Reproduce
Run in bash:
```
echo "SET search_path TO my_schema, public;" | sqlfluff parse --dialect postgres -
```

## Dialect
`postgres`

## Version
```
sqlfluff, version 0.7.1
Python 3.7.6
```

## Configuration
```
Default 
```

WittierDinosaur:
Added SET statement block to SQL - most of this functionality diverges from ANSI - so haven't added anything to ANSI

Fixes #1832 

No

### Pull Request checklist
Test Cases Added


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -2346,3 +2347,33 @@ class DropTypeStatementSegment(BaseSegment):
...
+
+    As specified in https://www.postgresql.org/docs/14/sql-set.html
+    """
+
+    type = "set_statement"
+
+    match_grammar = StartsWith("SET")
+
+    parse_grammar = Sequence(
+        "SET",
```
tunetheweb, line 2364: I've often wondered is there any benefit in this (this is not the only statement that is like this)?

My understanding of the `match_grammar` is that [it is there to do a quick match before a more intensive `parse_grammar` attempt](https://docs.sqlfluff.com/en/stable/architecture.html#stage-3-the-parser). Both of these will match `"SET"` in equal amount of time surely? So any point?

Should this be collapsed into just `match_grammar` with full definition?

Should we do a clean up elsewhere that uses this?

WittierDinosaur: I think the opposite cleanup is needed. Match should be the minimum required to uniquely identify a segment. In a stack, my understanding is it goes: first thing matches, try second thing, matches, third thing, matches etc. So I think if everything took the minimal match approach it'd parse faster

tunetheweb: If the Match grammar allowed us to to more efficiently match the whole statement (e.g. if it had a terminated by clause) the. I’d agree. As it stands the Match Grammar is just the first part of the Parse Grammar so if it doesn’t match that then the rest of it wouldn’t be run is my thinking? So I don’t see value of having the two here? But maybe I’m misunderstanding how efficiently it could recognise that and move on?

WittierDinosaur: The statements are already terminated by the statement segment greedy until bit. So inside the statement it becomes 'which segment is this'. The statements themselves don't need a terminator because they are the entire contents of the statement segment

tunetheweb: Yes but my point is, is it any quicker to test:

> if it begins with "SET" then check if "SET" followed by lots of other things

versus

> check if "SET" followed by lots of other things

?

Both surely exit out pretty quick if the thing we're checking doesn't begin with "SET"?

WittierDinosaur: Yeah, I think you're right for failure cases. But for success cases, it has to parse the entire tree to succeed, which would be slower

tunetheweb: But for success cases, it does have to parse the entire tree anyway doesn't it? The `match_grammar` in this case doesn't help prevent that?

WittierDinosaur: I think it'll parse it twice. It'll parse the whole thing to identify the match - it'll then parse it to parse it

tunetheweb: Ohhhh if that's the case then I see where you're coming from!

@alanmcruickshank any comments?

barrywhart: I'm outside of my expertise here, but two possibilities concern me:
* Is this premature optimization? I.e., are we trying to make something fast even though we haven't identified it as a bottleneck?
* Does this potentially make the dialect harder to maintain or extend? I.e., we're telling the parser that, if it starts with `SET`, it's definitely a `SetStatementSegment`. But what if there's a future requirement (in this dialect or a parent or child dialect) where a different segment type **also** begins with the keyword `SET`? Perhaps this is unlikely, but my usual philosophy is, don't do something tricky until you have a clear need to do so. I think overall, SQLFluff has a bigger need for simplicity than speed (because we have never-ending requests for dialect enhancements, so we want it to be "super easy" for new contributors).

tunetheweb: I don’t think this is premature optimisation if it does work like @WittierDinosaur thinks it does. It’s not like it’s complicated to set up.

But on the other hand it is a bit pointless if it works like I thought it did.

So either way it would be good to find out! Don’t think it’ll make a massive difference one way or the other to be honest. But mostly I’m just curious to be honest. Always good to know how something works if it’s the same or different than you assumed. Plus we might as well do it in the best way since it’s not difficult to do one or the other.

And doesn’t say this is the ONLY statement that begins with SET. So don’t think that’s a concern at all.

barrywhart: Could we try an artificial benchmark? I.e. create a huge file (thousands of lines or larger) with this kind of statement over and over, then test the parsing performance both ways.

+1 for settling discussions like this with numbers. Less discussion, plus we all learn a little. 💪

WittierDinosaur: Okay - I'm going to have to look more into how this works under the hood. Not using StartsWith(), and giving the full definition was 2% faster. I also tried `match_grammar = Sequence("SET", Anything())`, which was somewhere in between the two. I still think there are performance gains to be made in parsing - clearly not here though. I've changed it to use the full definition for this PR

tunetheweb: Oh cool. Thanks for testing. Definitely interested to hear how it actually works rather than me guessing and presuming.

tunetheweb (approved):

Issue #1216, opened 2021-07-19
sqlalchemy support


thomaspoignant:
I am using [sqlalchemy](https://www.sqlalchemy.org/) on my projects and the syntax is not valid for `sqlfluff`.
It would be great to have a templater for sqlalchemy to recognize the syntax.

Currently running it results with **unparsable errors**.

### Example
```sql
-- get_queues.sql
SELECT
    queue_id,
    queue_name,
    icon,
    code,
    type,
    suggested_reasons,
    COALESCE(ad_count, 0) AS ad_count
FROM
    queues
LEFT JOIN (
    SELECT
        UNNEST(queue_ids) AS id,
        COUNT(*) AS ad_count
    FROM
        moderation_inbox_items
    WHERE (:assigned_entity)::text IS NULL OR assigned_entity = (:assigned_entity)::text
    GROUP BY
        UNNEST(queue_ids)
)
ON id = queue_id
ORDER BY rank, queue_id
```
Is failing because of `:assigned_entity`

alanmcruickshank:
This shouldn't be too difficult. I've got one thing I'm not sure about though. Do you know how sqlalchemy knows not to count ``::text` (which ends with `:text`) as a variable?

jacopofar:
I'd like to try and implement it, and it seems now the best way is to create a plugin for it. I read the code of the DBT one but is not totally clear to me how it interfaces to SQLFluff (I don't know DBT).

There's a [`process`](https://github.com/sqlfluff/sqlfluff/blob/87197306a417a99aaccae62afacda3ac428f5931/plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py#L434) method to parse the file and I see it returns an instance of `[TemplatedFile](https://github.com/sqlfluff/sqlfluff/blob/87197306a417a99aaccae62afacda3ac428f5931/src/sqlfluff/core/templaters/base.py#L64)` which if I understand correctly maintains the reference to the original code so that later it can be "un-templated", is it correct?

pwildenhain:
@jacopofar Have you ever used the pluggy package? That is what we use to manage our plugins -- you can see what we make "pluggable" by checking out the src/sqfluff/core/plugins directory. Right now it's just custom rules and templaters. And then I suggest reading up on pluggy to see how you can plug in the sqlalchemy templater. I can guarantee you won't need anywhere near as much complexity as the dbt templater 😉

jacopofar:
[...]

Issue #1871, opened 2021-11-10
Support Placeholder for SQL


tank-pooja:
## Expected Behaviour
It should support generic SQL Queries.

We are using a query like : 

SELECT  * FROM users WHERE id = ?;

## Observed Behaviour

Showing Error:  Found unparsable section

## Dialect
Postgres

Include your SQLFluff configuration here
```
[sqlfluff]
dialect=postgres

**Comment by barrywhart** [comment]
Is the use of `?` a feature of Postgres itself? A client library? Something else?

**Comment by barrywhart** [comment]
Changed from "bug" to "enhancement" because I'm pretty sure this is not standard Postgres behavior. Rather, it is part of a client library or similar. So this sounds like a request to add a new (albeit simple) templater, not an enhancement to the Postgres dialect.

In fact, `?` is a built-in Postgres operator for working with JSON data. Sometimes this usage collides with the use of `?` as described here, e.g.:
* https://stackoverflow.com/questions/38370972/how-do-i-use-postgresql-jsonb-operators-containing-a-question-mark-via-jdb
* https://github.com/metabase/metabase/issues/1964

**Comment by tank-pooja** [comment]
@barrywhart , Thank you for your reply.
Yes, We are using this client library of Go - https://github.com/jmoiron/sqlx. So, is that possible to have support for `$variable` or `:variable` ?

**Comment by barrywhart** [comment]
It's possible. As I mentioned above, I think we'd treat this as a new templater. I think it's similar to #1216.

**Comment by barrywhart** [comment]
I think this is addressed by #1887. Can you confirm, @tank-pooja?

**Comment by tank-pooja** [comment]
@barrywhart , Yes, that's working. But, it's failing for this :

`L:   7 | P:  15 | L048 | Missing whitespace before '5'`

Here is config

```
[sqlfluff]
dialect= postgres
templater = sqlalchemy
[sqlfluff:templater:sqlalchemy:context]
id='5' 
```

And Query
```
SELECT
    *
FROM public.users
WHERE id = :id;

```

tank-pooja:
@jacopofar , thank you for adding support for a different variable placeholder.

@barrywhart , But, I am still facing issues as mentioned above, regarding whitespace.

jacopofar:
I can reproduce it, I thought the templater was adding whitespaces but I don't see any, and the automated tests are checking for an exact match, so I don't know where's the issue originating.

PR #1878, opened 2021-11-10
[many dialects] Add support for SQLAlchemy templating

jacopofar:
This closes in #1216 by adding support for the SQLAlchemy parameter passing. SQLALchemy is a very common library to interface with DBs and it supports multiple engines (e.g. Postgres, MySQL, SQLite, etc.)

Not that I can see, the usage is identical to the other templaters but much simpler

### Pull Request checklist
- [X ] Please confirm you have completed any of the necessary steps below.


WittierDinosaur:
Probably one for @barrywhart or @alanmcruickshank to review

src/sqlfluff/core/templaters/sqlalchemy.py
```diff
@@ -0,0 +1,174 @@
...
+            raw_slices.append(
+                RawFileSlice(
+                    raw=in_str[last_pos_raw : span[0]],
+                    slice_type="literal",
+                    source_idx=last_pos_raw,
+                )
+            )
+            out_str += in_str[last_pos_raw : span[0]]
+            # add the current replaced element
+            start_teplate_pos = last_pos_templated + last_literal_length
```
barrywhart, line 118:
```suggestion
            start_template_pos = last_pos_templated + last_literal_length
```

src/sqlfluff/core/templaters/sqlalchemy.py
```diff
@@ -0,0 +1,174 @@
...
+            template_slices.append(
+                TemplatedFileSlice(
+                    slice_type="block_start",
+                    source_slice=slice(span[0], span[1], None),
+                    templated_slice=slice(
+                        start_teplate_pos, start_teplate_pos + len(replacement), None
```
barrywhart, line 124:
```suggestion
                        start_template_pos, start_template_pos + len(replacement), None
```

src/sqlfluff/core/templaters/sqlalchemy.py
```diff
@@ -0,0 +1,174 @@
...
+                RawFileSlice(
+                    raw=in_str[span[0] : span[1]],
+                    slice_type="block_start",
+                    source_idx=span[0],
+                )
+            )
+            out_str += replacement
+            # update the indexes
+            last_pos_raw = span[1]
+            last_pos_templated = start_teplate_pos + len(replacement)
```
barrywhart, line 138:
```suggestion
            last_pos_templated = start_template_pos + len(replacement)
```

barrywhart (approved):
Other than a typo in a variable name, looks great!

Hypothetically, how much effort would it be to make this templater more generic, e.g. so we could subclass it to handle other templating syntax? For example, see #1871.

barrywhart (changes requested):
Changing review to "request changes" so we can fix the typos before merging.

src/sqlfluff/core/templaters/sqlalchemy.py
jacopofar: fixed

src/sqlfluff/core/templaters/sqlalchemy.py
jacopofar: done

barrywhart (approved):

barrywhart:
@jacopofar: If you're up for making that change, that would be great! Do you think it'd be better to do that in this or in a separate PR?

jacopofar:
we can close this and I'll open a separate request for that, should be easy to change

barrywhart:
Ok, merging now. For reference, issue #1871 would hopefully be easy to implement after making this templater more generic.

jacopofar:
oh sorry by close I meant without merging, since the new one will replace it completely and break the (unlikely) code already using it

barrywhart:
Oh, I'm sorry. Well, it's not in a release yet, so it's fine to change. It should be possible (assuming we want to) to keep it compatible (I.e. same templater name in the config file).

PR #1879, opened 2021-11-10
Add rule L051 to forbid lone JOIN

### LINKED ISSUE #1712: Add rule to not allow plain JOIN
Add rule to not allow "naked" joins, i.e. `JOIN`.

```sql
SELECT
  ...
FROM
  customer
  -- Should be an explicit join type, e.g. INNER JOIN or LEFT JOIN
  JOIN orders ON customer.id = orders.customer_id
```

Potentially, you could also add a setting to assume all naked `JOIN`s are `INNER JOIN`s (as most dialects do) and implement that when running `fix`.

```sql
SELECT
  ...
FROM
  customer
  -- Fixed
  INNER JOIN orders ON customer.id = orders.customer_id
```

jpy-git:

New rule to forbid the use of lone JOIN. It also provides a fix by replacing lone JOIN with INNER JOIN.
Logic if pretty simple and have included test cases. Fixes #1712 

Screen-Recording-2021-11-10-at-18 13 13

No

### Pull Request checklist

  - Other.


test/cli/commands_test.py
```diff
@@ -149,20 +149,49 @@ def test__cli__command_lint_stdin(command):
     "command",
     [
         # Test basic linting
-        (lint, ["-n", "test/fixtures/cli/passing_b.sql"]),
+        (lint, ["-n", "test/fixtures/cli/passing_b.sql", "--exclude-rules", "L051"]),
```
jpy-git, line 152: FYI I have excluded L051 from these tests as `passing_b.sql` contains a lone JOIN. Alternatively I can remove the lone JOIN from that test query? 😄

jpy-git:
@barrywhart could you take a look at this one please? Should be quite quick :)

src/sqlfluff/rules/L051.py
```diff
@@ -0,0 +1,60 @@
...
+        # We identify non-lone JOIN by looking at first child segment.
+        if context.segment.segments[0].name != "join":
+            return None
+
+        # Replace lone JOIN with INNER JOIN.
+        return LintResult(
+            context.segment.segments[0],
+            fixes=[
+                LintFix(
+                    "edit",
```
barrywhart, line 51: I think we want to use `"create"` here rather than `"edit"`.

Reason: `"create"` adds new **sibling** segments, while `"edit"` creates a new level of the parse tree, replacing the specified segment with a new segment of the same type, with the new segments _as children_. This is a subtle bug, because the final SQL **string** looks okay, but the parse **tree** is wrong. This can cause other lint rules to misbehave.

See an excerpt of `sqlfluff parse` output below for two queries, one using `JOIN`, the other using `INNER JOIN`. In the second, you'll see that `INNER` and the whitespace should be at the same level as the `JOIN`.

```
[L:  1, P: 30]      |                    [META] dedent:
[L:  1, P: 30]      |                    join_clause:
[L:  1, P: 30]      |                        keyword:                              'JOIN'
[L:  1, P: 34]      |                        [META] indent:
[L:  1, P: 34]      |                        whitespace:                           ' '
[L:  1, P: 35]      |                        from_expression_element:
[L:  1, P: 35]      |                            table_expression:
[L:  1, P: 35]      |                                table_reference:
[L:  1, P: 35]      |                                    identifier:               'bar'
[L:  1, P: 38]      |                        [META] dedent:
[L:  1, P: 38]      |    newline:                                                  '\n'
```

```
[L:  1, P: 30]      |                    [META] dedent:
[L:  1, P: 30]      |                    join_clause:
[L:  1, P: 30]      |                        keyword:                              'INNER'
[L:  1, P: 35]      |                        whitespace:                           ' '
[L:  1, P: 36]      |                        keyword:                              'JOIN'
[L:  1, P: 40]      |                        [META] indent:
[L:  1, P: 40]      |                        whitespace:                           ' '
[L:  1, P: 41]      |                        from_expression_element:
[L:  1, P: 41]      |                            table_expression:
[L:  1, P: 41]      |                                table_reference:
[L:  1, P: 41]      |                                    identifier:               'bar'
[L:  1, P: 44]      |                        [META] dedent:
[L:  1, P: 44]      |    newline:                                                  '\n'
```

barrywhart (changes requested):

src/sqlfluff/rules/L051.py
barrywhart: This rule is similar to L033. I see it has the same issue. I'll look into creating a PR to fix that one.

jpy-git: @barrywhart Interesting, I wasn't aware of that! Will bear it in mind going forwards.

I've made the switch to using a create LintFix now, unit tests are all passing still but as you said this should prevent future bugs 😄

jpy-git:
@barrywhart all looks green so should be good to go 🚀

src/sqlfluff/rules/L051.py
barrywhart: Thanks! The behavior I described is pretty tricky and not documented. I hope in the future, we can add automated checking inside the linter to flag it if a rule does something to the parse tree that makes it illegal wrt the dialect.

barrywhart (approved):

PR #1893, opened 2021-11-13
L036 refinement - FROM clause interaction

jpy-git:

This PR aims to resolve an issue with the logic in L036, in which the columns in a query such as `SELECT a, b FROM c;` are moved to a new line without considering the impact on the FROM clause, leaving the FROM clause hanging off to the right of the final column rather than under the SELECT keyword as expected.

This is an example of behaviour prior to this PR:
Screen-Recording-2021-11-13-at-02 02 39

This is an example of the updated logic to account for moved FROM clauses:
Screen-Recording-2021-11-13-at-02 05 17

I have included the additional test cases to validate the new correct behaviour.

No

### Pull Request checklist

  - Other.


jpy-git:
@barrywhart Could you take a look at this when you get the chance, thanks 😄

src/sqlfluff/rules/L036.py
```diff
@@ -82,12 +90,21 @@ def _get_indexes(segment):
             ):
                 first_whitespace_idx = fname_idx
 
+        for seg in context.siblings_post:
+            if seg.is_type("from_clause"):
```
barrywhart, line 94: Nitpick: Would it be clearer to change this to `if seg is from_segment:` ?

src/sqlfluff/rules/L036.py
```diff
@@ -333,5 +369,33 @@ def _eval_single_select_target_element(
...
+                and (
+                    select_clause.pos_marker.working_line_no
+                    != select_targets_info.from_segment.pos_marker.working_line_no
+                )
+                and (
+                    wildcard_select_clause_element.pos_marker.working_line_no
+                    == select_targets_info.from_segment.pos_marker.working_line_no
+                )
+            ):
+                fixes = [
```
barrywhart, line 390: Is this the same code as line 147? I wonder if we ought to create a function with this code and call it from both these places.

barrywhart (approved):
A couple of very minor suggestions, but I'm happy to merge "as is" -- let me know what you think.

barrywhart:
BTW, nice work!!

src/sqlfluff/rules/L036.py
jpy-git: I've implemented this change 👍

src/sqlfluff/rules/L036.py
jpy-git: So my thinking on this was:
On lines 142-157, we are appending these lint fixes to an already existing lint fixes, whereas on lines 378-399 this condition only a wildcard and creates a separate list of fixes to the rest of the code. Given that the conditions in the if clause are also different, the only thing we could add to a function would be the
```python
 fixes.extend(
                        [
                            LintFix("delete", ws)
                            for ws in select_targets_info.pre_from_whitespace
                        ]
                    )
                    fixes.append(
                        LintFix(
                            "create", select_targets_info.from_segment, NewlineSegment()
                        )
                    )
```
lines, which IMO are already pretty self explanatory so maybe not much is gained in terms of moving them to a separate method?

Happy to implement the change if you think it's worth it though 😄

jpy-git:
@barrywhart Thanks, appreciate the feedback! 😄 I think the PR is ready to go if you're happy with my reasoning on the splitting fix logic into new method point?

barrywhart:
Looks good. No word worries about the new function -- truly nitpicky feedback on my part. I'll do a final look and merge tonight.

PR #1894, opened 2021-11-13
Fixed Delimited() logic, added TSQL grammar

### LINKED ISSUE #1366: Non-terminated Delimited() functions affect the behaviour of later Delimited() functions in the match grammar

### LINKED ISSUE #1892: Support view attributes for T-SQL
## Expected Behaviour
SQLFluff using T-SQL dialect should be able to parse a view creation statement that uses WITH ENCRYPTION or WITH SCHEMABINDING or WITH VIEW_METADATA attributes.

Refer to T-SQL create view syntax documentation here: https://docs.microsoft.com/en-us/sql/t-sql/statements/create-view-transact-sql?view=sql-server-ver15

## Observed Behaviour
SQLFluff fails to parse the statement

## Steps to Reproduce
`echo create view vw_view with schemabinding as select A.ID from A| sqlfluff parse -`

```
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    unparsable:                                               !! Expected: 'FileSegment'
[L:  1, P:  1]      |        raw:                                                  'create'
[L:  1, P:  7]      |        whitespace:                                           ' '
[L:  1, P:  8]      |        raw:                                                  'view'
[L:  1, P: 12]      |        whitespace:                                           ' '
[L:  1, P: 13]      |        raw:                                                  'vw_view'
[L:  1, P: 20]      |        whitespace:                                           ' '
[L:  1, P: 21]      |        raw:                                                  'with'
[L:  1, P: 25]      |        whitespace:                                           ' '
[L:  1, P: 26]      |        raw:                                                  'schemabinding'
[L:  1, P: 39]      |        whitespace:                                           ' '
[L:  1, P: 40]      |        raw:                                                  'as'
[L:  1, P: 42]      |        whitespace:                                           ' '
[L:  1, P: 43]      |        raw:                                                  'select'
[L:  1, P: 49]      |        whitespace:                                           ' '
[L:  1, P: 50]      |        raw:                                                  'A'
[L:  1, P: 51]      |        raw:                                                  '.'
[L:  1, P: 52]      |        raw:                                                  'ID'
[L:  1, P: 54]      |        whitespace:                                           ' '
[L:  1, P: 55]      |        raw:                                                  'from'
[L:  1, P: 59]      |        whitespace:                                           ' '
[L:  1, P: 60]      |        raw:                                                  'A'
[L:  1, P: 61]      |    newline:                                                  '\n'

==== parsing violations ====
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'create view vw_view
                       | with schemabinding a...'
```

## Dialect
T-SQL

## Version
SQLFluff - 0.8.1
Python - 3.10.0

## Configuration
```
# Refer to https://docs.sqlfluff.com/en/stable/configuration.html#default-configuration for defaults
[sqlfluff]
dialect = tsql
templater = raw
sql_file_exts = .sql

[sqlfluff:rules]
allow_scalar = False

[sqlfluff:rules:L007]  # Operator location
operator_new_lines = before

[sqlfluff:rules:L010]  # Keyword capitalization
capitalisation_policy = upper

[sqlfluff:rules:L014]  # Identifier capitalization
extended_capitalisation_policy = pascal

[sqlfluff:rules:L030]  # Function capitalization
capitalisation_policy = lower

[sqlfluff:rules:L040]  # Null & Boolean capitalization
capitalisation_policy = lower

[sqlfluff:rules:L042]  # Subqueries
forbid_subquery_in = both

[sqlfluff:rules:L047]  # Count syntax
prefer_count_1 = True
prefer_count_0 = False
```

WittierDinosaur:
This PR does two largely unrelated things. Firstly, it adds options to the tsql CREATE VIEW syntax.
Secondly, it fixes a bug in the Delimited(), which causes rogue behaviour when there is another Delimited within the same statement block.

Fixes #1892 
Fixes #1366

No. Though there are probably some code clean-ups wee can do where people have avoided using Delimited() because of errors they couldn't explain

### Pull Request checklist
Tests added for the TSQL stuff - which served as useful test cases for the Delimited() stuff. I haven't made any explicit unit tests for the delimited change

### DIFF
--- a/src/sqlfluff/core/parser/grammar/delimited.py
+++ b/src/sqlfluff/core/parser/grammar/delimited.py
@@ -167,7 +167,19 @@ class Delimited(OneOf):
                         )
                     # No match - Not allowed
                     if not match:
-                        return MatchResult.from_unmatched(mutated_segments)
+                        if self.allow_trailing:
+                            # If we reach this point, the lookahead match has hit a delimiter
+                            # beyond the scope of this Delimited section. Trailing delimiters are allowed,
+                            # so return matched up to this section.
+                            return MatchResult(
+                                matched_segments.matched_segments,
+                                pre_non_code
+                                + match.unmatched_segments
+                                + post_non_code
+                                + delimiter_match.all_segments(),
+                            )
+                        else:
+                            return MatchResult.from_unmatched(mutated_segments)
 
                     if not match.is_complete():
                         # If we reach this point, the lookahead match has hit a delimiter
--- a/src/sqlfluff/core/parser/grammar/delimited.py
+++ b/src/sqlfluff/core/parser/grammar/delimited.py
@@ -165,9 +165,22 @@ class Delimited(OneOf):
                             # We've already trimmed
                             trim_noncode=False,
                         )
-                    # No match, or an incomplete match: Not allowed
-                    if not match or not match.is_complete():
+                    # No match - Not allowed
+                    if not match:
                         return MatchResult.from_unmatched(mutated_segments)
+
+                    if not match.is_complete():
+                        # If we reach this point, the lookahead match has hit a delimiter
+                        # beyond the scope of this Delimited section. We should return a
+                        # partial match, and the delimiter as unmatched.
+                        return MatchResult(
+                            matched_segments.matched_segments
+                            + pre_non_code
+                            + match.matched_segments,
+                            match.unmatched_segments
+                            + post_non_code
+                            + delimiter_match.all_segments(),
+                        )
 
                     # We have a complete match!
 
--- a/src/sqlfluff/dialects/dialect_tsql.py
+++ b/src/sqlfluff/dialects/dialect_tsql.py
@@ -1006,8 +1006,14 @@ class CreateViewStatementSegment(BaseSegment):
         Sequence("OR", "ALTER", optional=True),
         "VIEW",
         Ref("ObjectReferenceSegment"),
+        Sequence(
+            "WITH",
+            Delimited("ENCRYPTION", "SCHEMABINDING", "VIEW_METADATA"),
+            optional=True,
+        ),
         "AS",
         Ref("SelectableGrammar"),
+        Sequence("WITH", "CHECK", "OPTION", optional=True),
         Ref("DelimiterSegment", optional=True),
     )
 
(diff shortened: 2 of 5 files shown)


tunetheweb (approved):
Woo nice find fixing that delimiter issue!

Though probably should really be two separate PRs in future!

WittierDinosaur:
Definitely should - just ended up with a weird interdependency. The fix made the grammar viable, and the grammar tested the fix 😂

PR #1896, opened 2021-11-13
Add INSERT INTO statements to Redshift Dialect

### LINKED ISSUE #1895: Redshift Dialect Cannot Parse INSERT INTO Statement
## Expected Behaviour
SQLFluff ought to be able to parse the following SQL statement using the Redshift dialect:
```
INSERT INTO s1.t1 (
    SELECT
        col1,
        col2,
        col3
    FROM testtable
  );
```

## Observed Behaviour
SQLFluff notes that this section is unparsable.

## Steps to Reproduce
Run the aforementioned query.

## Dialect
Redshift

## Version
`0.8.1`

## Configuration
Default

tdstark:
Fixes #1895 , the Insert Into statement was added to the Redshift dialect. Note that this differs from the ANSI/Postgres dialects in a few ways. Postgres, for example, can upsert but that is not a possibility in Redshift.

None at this time

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -249,6 +250,34 @@ class CreateTableStatementSegment(BaseSegment):
...
+    type = "insert_statement"
+    match_grammar = Sequence(
+        "INSERT",
+        "INTO",
+        Ref("TableReferenceSegment"),
+        OneOf(
+            OptionallyBracketed(Ref("SelectableGrammar")),
+            Sequence("DEFAULT", "VALUES"),
+            Sequence(
+                Ref("BracketedColumnReferenceListGrammar", optional=True),
```
WittierDinosaur, line 271: Did you work out which bit of this is causing the problem?

tdstark: Yeah, It was because I didn't have it as part of the `StatementSegment`. Once I did that, the clouds opened up :).

WittierDinosaur: That doesn't make sense though - as it is a segment you're replacing, it's already part of StatementSegment

tdstark: I removed that part. It's probably the fact that I added OneOf.

src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -249,6 +250,34 @@ class CreateTableStatementSegment(BaseSegment):
...
+        - https://docs.aws.amazon.com/redshift/latest/dg/r_INSERT_30.html
+        - https://docs.aws.amazon.com/redshift/latest/dg/r_INSERT_external_table.html
+    """
+
+    type = "insert_statement"
+    match_grammar = Sequence(
+        "INSERT",
+        "INTO",
+        Ref("TableReferenceSegment"),
+        OneOf(
```
WittierDinosaur, line 267: This isn't the simplest representation of this grammar. It would be nice if we could work out which element is misbehaving, to cause the issue you posted about in slack

tdstark: What could be simplified?

tdstark: I dug into this a bit more on my own and I shot a couple of messages over to @tunetheweb as well. I can't tell you exactly what's happening here. What I know for sure is that there appears to be something about the parenthesis which is throwing off how the `SELECT` statement is parsed. I tried it with `Bracketed()` it doesn't seem to have the expected affect. Using the following query: 
```
INSERT INTO s1.t1 (
    SELECT
        col1,
        col2,
        col3
    FROM testtable
 );
```
...and the following statement:
```
    type = "insert_statement"
    match_grammar = Sequence(
        "INSERT",
        "INTO",
        Ref("TableReferenceSegment"),
        Ref("BracketedColumnReferenceListGrammar", optional=True),
        Bracketed(Ref("SelectableGrammar")),
    )
```
This throws an unparseable error. However, this *can* parse that statement:
```
  type = "insert_statement"
  match_grammar = Sequence(
      "INSERT",
      "INTO",
      Ref("TableReferenceSegment"),
      Bracketed(Ref("SelectableGrammar")),
  )
```
So my assumption is that it is identifying the parenthesis around the `SELECT` statement as `BracketedColumnReferenceListGrammar` or the `option=True` statement isn't working as expected. I'm assuming it's the latter though. If it is, any suggestions?

barrywhart: This sounds like a legitimately tricky, i.e. ambiguous parsing situation.

Speculation below, but hopefully some useful suggestions...

To make this work, I think you'll need to add some form of "hint" to the grammar, so it can tell the difference between a `SELECT` versus a list of columns. See the use of `match_grammar = StartsWith(` several places in `dialect_ansi.py`. You may have to experiment with different "factorings" of the various segments as well, because the ambiguity (`SELECT` vs column list) both occur inside of parentheses.

This code in `dialect_ansi.py` may offer some ideas as well, because here (and in the related `SelectClauseElementTerminatorGrammar`), you can see the sort of "hints" I'm referring to -- helping the parser identify a `SELECT` statement by some of the elements that start or end it.

```
@ansi_dialect.segment()
class SelectClauseElementSegment(BaseSegment):
    """An element in the targets of a select statement."""

    type = "select_clause_element"
    # Important to split elements before parsing, otherwise debugging is really hard.
    match_grammar = GreedyUntil(
        Ref("SelectClauseElementTerminatorGrammar"),
        enforce_whitespace_preceding_terminator=True,
    )

    parse_grammar = OneOf(
        # *, blah.*, blah.blah.*, etc.
        Ref("WildcardExpressionSegment"),
        Sequence(
            Ref("AliasExpressionSegment", optional=True),
        ),
    )
```

tdstark: I put a few hours into testing `GreedyUntil`, `StartsWith`, and `match_grammar` to no avail. I'm not sure if anyone else sees a clear path to make this work. Since this parses fine with the current approach, do we want to push the changes with a TODO? I'm fine if we want to find the longer-term solution as well, but this could take me some time. I'd like to get this and the `GRANT` and `GROUP` statements ready since this will make Redshift usable for most use cases if those are completed. My thoughts are that once those are done, I can pivot back to this. Thoughts?

tdstark: Sounds good. I'll add the TODO tomorrow.

src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -249,6 +250,34 @@ class CreateTableStatementSegment(BaseSegment):
     )
 
 
+@redshift_dialect.segment(replace=True)
+class InsertStatementSegment(BaseSegment):
+    """A `INSERT` statement.
```
barrywhart, line 255:
```suggestion
    """An `INSERT` statement.
```

src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -56,17 +57,6 @@ class DatePartFunctionNameSegment(BaseSegment):
     match_grammar = OneOf("DATEADD", "DATEDIFF")
 
 
-@redshift_dialect.segment(replace=True)
```
barrywhart, line 59: Why was this moved to a different place in the file? It seems like an unnecessary change.

src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -249,6 +250,37 @@ class CreateTableStatementSegment(BaseSegment):
...
+    # See https://github.com/sqlfluff/sqlfluff/pull/1896
+
+    type = "insert_statement"
+    match_grammar = Sequence(
+        "INSERT",
+        "INTO",
+        Ref("TableReferenceSegment"),
+        OneOf(
+            OptionallyBracketed(Ref("SelectableGrammar")),
+            Sequence("DEFAULT", "VALUES"),
```
barrywhart, line 272: I don't see a test that uses `DEFAULT VALUES`. Should there be?

barrywhart (commented):
This looks good -- just a few small suggestions or questions.

src/sqlfluff/dialects/dialect_redshift.py
tdstark: I was originally going to organize the functions thematically and/or alphabetically but I didn't get that far.

src/sqlfluff/dialects/dialect_redshift.py
tdstark: I believe this already exists [here](https://github.com/sqlfluff/sqlfluff/pull/1896/files#diff-4a8ae40a1739111fbb9a0b5f9be034d4d56d1aa05ec975398112a3049c9784e9R25-R26).

barrywhart (commented):
Updates look great! Approving and merging. Thanks!

barrywhart (approved):

PR #1900, opened 2021-11-14
Documentation: Change inheritance dialect example to Redshift

chwiese:

Changes the inheritance examples on the dialects documentation page from Snowflake to Redshift. The Snowflake dialect inherits from ANSI only, while the Redshift dialect does inherit from Postgres. This way the example matches what is actually happening in the code.

No.

### Pull Request checklist

  - Other.


tunetheweb (approved):
Good spot. Snowflake used to inherit from Postgres (for no real good reason), but since then it was moved to be direct from ANSI and Redshift was added.

PR #1901, opened 2021-11-14
Add support for curly brackets in SnowSQL ampersand variables

chwiese:

Moves support for Snowsql variables from the dialect into the placeholder templater and adds support for curly brackets, as documented [here](https://docs.snowflake.com/en/user-guide/snowsql-use.html#substituting-variables-in-a-session). Since Snowsql variables can hold SQL logic such as optional WHERE clauses, the templater is the better home for them as discussed [here](https://sqlfluff.slack.com/archives/C01JS7H598E/p1636892634079800?thread_ts=1636886949.072400&cid=C01JS7H598E) on Slack.

Yes, anybody how currently has Snowsql variables in their scripts will run into parsing errors. They will need to switch to the placeholder templater and define the placeholders in their Sqlfluff config.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -105,7 +107,7 @@
         type="variable",
     ),
     ReferencedVariableNameSegment=RegexParser(
-        r"(\$|\&)[A-Z][A-Z0-9_]*",
+        r'(?:\$[A-Z][A-Z0-9_]*)|(?:[A-Z0-9_\."]*&{[A-Z0-9_]*}[A-Z0-9_]*)|(?:&[A-Z0-9_\."]*)',
```
chwiese, line 110: For some reason `(?:[\w\."]*&{[A-Z0-9_]*}[\w]*)` does not work, while `(?:[A-Z0-9_\."]*&{[A-Z0-9_]*}[A-Z0-9_]*)` does. Does anybody know why?

tunetheweb: Those should be identical. Can you define "does not work"? Have you been able to reproduce it on https://regex101.com/?

chwiese: Sorry, you are right, "does not work" is rarely helpful.
When I use `r'(?:\$[A-Z][A-Z0-9_]*)|(?:[A-Z0-9_\."]*&{[A-Z0-9_]*}[A-Z0-9_]*)|(?:&[A-Z0-9_\."]*)',`, then `USE WAREHOUSE &WAREHOUSE;` parses alright. If I use `r'(?:\$[A-Z][A-Z0-9_]*)|(?:[\w\."]*&{[A-Z0-9_]*}[A-Z0-9_]*)|(?:&[\w\."]*)',`, the same piece of SQL returns a _Found unparsable section_ error.
I cannot reproduce this on regex101.com - over there both patterns produce the desired match `&WAREHOUSE`.

WittierDinosaur: This is a bug - not in your regex - but in the RegexParser that is being used. RegexParser inherits from StringParser, which calls this in its __init__ (which is called in the RegexParser)
```
self.template = template.upper()
```
So, `\w` becomes `\W`, which matches every character that is not a word character. Basically `^[\w]`

barrywhart (approved):
The code changes look good to me. I do have a question: This PR appears to add support for ampersand variables both in the placeholder templater as well as the Snowflake dialect. Do both of them use the same syntax? Is it possible to use both at once? Do we need to add or update documentation so users will understand what's supported and how to use it?

chwiese:
@barrywhart Actually this PR removes ampersand support from the Snowflake dialect and moves it to the placeholder templater only. There is still dollar sign variable support in the dialect though.

Both can be used concurrently. Dollar sign variables are part of the dialect and can be used with all clients, ampersand variables are specific to the official SnowSQL client.

barrywhart:
Got it! Merging now. Great work!

PR #1902, opened 2021-11-14
New Rule L052: Semi colon alignment

### LINKED ISSUE #1748: New TSQL rule suggestion: Flag statements that do not end with semi-colon
The semi-colon statement terminator is optional but best practice.

This could be fixed automatically.

Anti-pattern:

```
SELECT 1
```

Pattern:

```
SELECT 1;
```

This could potentially apply to other dialects as well; I am not familiar enough with other dialects to say for sure.

An exception to the above rule would be the BEGIN statement, where we should flag/remove the optional semi-colon if we see it.

jpy-git:

This PR introduces a new rule L052. It serves two purposes, one to snap semi-colons to the end of the preceding statement, and the other to add the final trailing semi-colon if it is missing:

Screen-Recording-2021-11-14-at-15 59 30

I have added a configuration option called `allow_final_semi_colon` (which defaults to False) so users can allow missing final semi-colons if they desire (I'm happy to change the default to True on this if you would prefer, but my preference is to add the final semi-colons by default).

This change impacts quite a few unit tests, solely because we weren't previously including final semi-colon, so I have gone through and added the semi-colon/excluded the rule where I think appropriate.

Fixes #1748 

No

### Pull Request checklist

  - Other.


jpy-git:
@barrywhart took a couple attempts to track down all the places this impacted in the dbt tests, but this should be good for review now 😄

WittierDinosaur:
I don't think this is an anti-pattern. It's fairly standard in large queries
```
        -- Newlines between the end of the statement
        -- and the semi-colon are forbidden.
        SELECT
            a
        FROM foo
        ;
```

tunetheweb:
Personally I like this rule!

What about an optional config to also allow them to be on a newline (with no extra newlines in between) to keep the @WittierDinosaur 's of the world happy too? In which case should this be enforced or only allowed? Which should `fix` use if it's missing and this optional param is set?

jpy-git:
@WittierDinosaur @tunetheweb sure thing! I don't think it should be too hard to add that option.

The way it will work is user has an option of either semi-colon on same line OR new line, but not both. Consistentcy should be encouraged.

The default will be to use same line.

alanmcruickshank:
So - potential spanner in the works for this rule. Trailing semi colons are explicitly a problem for dbt users: 
image

Reason being that often queries are "wrapped" by other sql and so the trailing semicolon becomes an issue.

Potential solutions:
- This rule is disabled by default for users when using the dbt templater (although I'm not sure exactly what the mechanism should be to do that).
- That semi-colons _between_ statements are linted by default, but that the default is that _trailing_ semicolons aren't added.

jpy-git:
@alanmcruickshank ah was not aware of that, agree that your approach to disable the trailing semi-colon by default is pragmatic though, will tweak the default config 👍

alanmcruickshank:
> @alanmcruickshank ah was not aware of that, agree that your approach to disable the trailing semi-colon by default is pragmatic though, will tweak the default config 👍

If we're taking that option - would you mind removing the semi-colons from the tests in the dbt tests. I'm happy for them to stay on the core tests - but I think that sets the right expectation for performance with dbt. Bonus points if you make a note in the default config to that effect.

jpy-git:
Will do! 😄

jpy-git:
Addressed the "allow_final_semi_colon" being set to true by default point. Will move on to looking at the option for semi-colons on a new line now

jpy-git:
@WittierDinosaur @tunetheweb @alanmcruickshank I've added the options to configure the behaviour of rule L052 as we described above. I've also added extra unit tests to for the config options.

Demo:
semi-colon-config-options

jpy-git:

barrywhart:
Tagging other commenters for review. I'll take a look as well.

src/sqlfluff/core/default_config.cfg
```diff
@@ -43,6 +43,8 @@ comma_style = trailing
 allow_scalar = True
 single_table_references = consistent
 unquoted_identifiers_policy = all
+semi_colon_new_line = False
```
barrywhart, line 46: Isn't "semicolon" a single (compound) word? I'm thinking we don't want that underscore there.
```suggestion
semicolon_newline = False
```

src/sqlfluff/core/default_config.cfg
```diff
@@ -43,6 +43,8 @@ comma_style = trailing
...
+require_final_semi_colon = False
```
barrywhart, line 47:
```suggestion
require_final_semicolon = False
```

src/sqlfluff/core/default_config.cfg
jpy-git: I think you can also spell it semi-colon, hence the underscore but happy to do one word 👍 
Should I maybe move that config to a rule specific config as well?

barrywhart: I would move it, yes. It's unlikely another rule would use something so specific.

src/sqlfluff/core/default_config.cfg
jpy-git: Flowed these changes through to code and unit tests

src/sqlfluff/core/default_config.cfg
jpy-git: Flowed these changes through to code and unit tests

tunetheweb (approved):
LGTM

jpy-git:
@tunetheweb @barrywhart think we can merge this, L053, and L054 down once the final checks are done, I've addressed all the suggestions and test cases look good

pwildenhain (approved):

Issue #1808, opened 2021-11-02
Rule language to make it easier to add rules


tunetheweb:
Grammar is written in Python but, for the most part you don't need to know Python to write it. It's kind of it's own syntax and non-Python people can easily add a dialect change without knowing much (if any!) Python, other than setting up their environment and knowing how to run `black`. As a result we get a lot of new people able to add dialect changes.

Rules, however are pure Python. You need to know Python data structures and loops and the lot. This makes it quite difficult (and intimidating!) to add or enhance rules, not to mention easier to make a mistake. As a result it's mostly the maintainers that write rules and there are relatively few of them.

At the time of writing we have 49 rules, but 27 open issues for new rules! I don't know how much syntax we support, and how mcuh we know we don't support but I'd be willing to bet the ratio isn't anywhere near 49:27!

Could we have some kind of "rule language", similar to the dialect syntax, that would make it easier to add rules?

For example #1748 is "New TSQL rule suggestion: Flag statements that do not end with semi-colon". It feels like we shouldn't have to write a load of Python to implement that and should be able to say something like:

```
if StatementSegment.nextSegment != ";" then fail
```

Or:

```
if StatementSegment.nextSegment != ";":
    return LintFix("edit", context.segment, [context.segment, SemicolonSegment()])
```

Or maybe we just need some more generic rules, like the block rule proposed in https://github.com/sqlfluff/sqlfluff/issues/361#issuecomment-955778171 ?

Or maybe some more generic functions (like `.nextSegment`), as currently adding rules feels like going back to the start each time, unless there's a rule that's very close to this one.

barrywhart:
I like the idea! I've wondered about something similar.

I agree we can make a lot of progress simply by providing some higher-level reusable helpers. I think until now, we've almost deliberately avoided that, not because it's too hard, but perhaps because the rewards didn't seem worthwhile when adding individual rules, and once there were a lot of rules, the lack of reusable primitives started to seem like an institution.

I would start by doing an inventory of the existing rules to look for the most common patterns, which often involve iterating segments with various filters.

Related: I have probably fixed about 5 past rule bugs where a rule "fix" did one or more of:
* Produce a parse tree where the string rendition was correct, but the parse tree was wrong (e.g. #1803). High-level idea: After applying a fix, can we rerun the parser on that area of the tree? Related idea: #1012
[...]

PR #1904, opened 2021-11-14
Fix bug where L041 was confused by L016's placement of newlines in the parse tree

### LINKED ISSUE #1881: Order of application of rules affects computability
When `SELECT` is uppercase and following `distinct` is lowercase AND the line is long enough to trigger carrying over, SQLFluff puts capitalised `DISTINCT` before carried-over `SELECT`. This breaks SQL syntaxis. 

I believe this is due to 3 Rules being applied simultaneously on the source text, rather than sequentially one after another:
1. Rule_L010 (Inconsistent capitalisation of keywords)
2. Rule_L016 (Line is too long) - the threshold to trigger it is between 49 and 87, so, probably, 60?
3. Rule_L041 (SELECT clause modifiers such as DISTINCT must be on the same line as SELECT.)

Thus the 3rd rule puts capitalised `DISTINCT` on the same line where `SELECT` was (but it is not there anymore).

Request:
```
SELECT
        *
    FROM
        `project_name.schema_name.table1`
    WHERE
        long_varname_to_trigger_Rule_L016_id in (SELECT distinct id FROM `project_name.schema_name.table2` WHERE deletedat IS NULL)
```

## Expected Behaviour
```
SELECT
        *
    FROM
        `project_name.schema_name.table1`
    WHERE
         long_varname_to_trigger_Rule_L016_id IN (SELECT DISTINCT
              id 
            FROM `project_name.schema_name.table2` WHERE deletedat IS NULL
         )
```

## Observed Behaviour
```
SELECT
    *
FROM
    `project_name.schema_name.table1`
WHERE
    long_varname_to_trigger_Rule_L016_id IN ( DISTINCT
        SELECT
            id
        FROM `project_name.schema_name.table2` WHERE deletedat IS NULL
    )
```

## Steps to Reproduce
Use the request above. Substitute `project_name.schema_name.table1/2` with any tables, available in your environment (and use corresponding variable names, but the 6th line of the request should stay long (>60 symbols)

## Dialect
- Jinja SQL in VSCode
- Standard SQL in BigQuery
- SQLFluff is not yet [configured for dbt specifics](https://docs.sqlfluff.com/en/stable/configuration.html?highlight=dbt#dbt-project-configuration)

## Version
- sqlfluff, version 0.7.1 and version 0.8.1 checked
- Python 3.8.9

## Configuration
Standard "vanilla" configuration, no tweaks of rules, no configuration files

barrywhart:

Fixes #1881 

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L041.py
+++ b/src/sqlfluff/rules/L041.py
@@ -35,21 +35,36 @@ class Rule_L041(BaseRule):
     def _eval(self, context: RuleContext) -> Optional[LintResult]:
         """Select clause modifiers must appear on same line as SELECT."""
         if context.segment.is_type("select_clause"):
+            # Scan the select clause's children. Look for three things in sequence:
+            # 1. SELECT keyword (required)
+            # 2. Newline (optional)
+            # 3. Select modifier, e.g. "DISTINCT" (optional)
+            select_keyword = None
+            newline_between = None
+            select_modifier = None
+            for idx, child in enumerate(context.segment.segments):
+                if select_keyword is None:
+                    if child.is_type("keyword") and child.raw.lower() == "select":
+                        select_keyword = child
+                if (
+                    select_keyword
+                    and newline_between is None
+                    and select_modifier is None
+                    and child.is_type("newline")
+                ):
+                    newline_between = child
+                    newline_idx = idx
+                if select_modifier is None and child.is_type("select_clause_modifier"):
+                    select_modifier = child
+                    break
+
             # Does the select clause have modifiers?
-            select_modifier = context.segment.get_child("select_clause_modifier")
             if not select_modifier:
                 return None  # No. We're done.
-            select_modifier_idx = context.segment.segments.index(select_modifier)
 
-            # Does the select clause contain a newline?
-            newline = context.segment.get_child("newline")
-            if not newline:
+            # Does select clause contain a newline between SELECT and the modifiers?
+            if not newline_between:
                 return None  # No. We're done.
-            newline_idx = context.segment.segments.index(newline)
-
-            # Is there a newline before the select modifier?
-            if newline_idx > select_modifier_idx:
-                return None  # No, we're done.
 
             # Yes to all the above. We found an issue.
 
@@ -61,7 +76,12 @@ class Rule_L041(BaseRule):
             ]
             fixes = [
                 # E.g. "\n" -> " DISTINCT\n.
-                LintFix("edit", newline, replace_newline_with),
+                LintFix("delete", newline_between),
+                LintFix(
+                    "create",
+                    context.segment.segments[newline_idx + 1],
+                    replace_newline_with,
+                ),
                 # E.g. "DISTINCT" -> X
                 LintFix("delete", select_modifier),
             ]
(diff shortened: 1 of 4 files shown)


src/sqlfluff/rules/L041.py
```diff
@@ -35,21 +35,36 @@ class Rule_L041(BaseRule):
     def _eval(self, context: RuleContext) -> Optional[LintResult]:
         """Select clause modifiers must appear on same line as SELECT."""
         if context.segment.is_type("select_clause"):
+            # Scan the select clause's children. Look for three things in sequence:
```
barrywhart, line 38: The previous version of this rule would get confused by newlines in unexpected places (i.e. **before** the `SELECT`), so I replaced the three individual segment searches with this loop, which **only** triggers if there's a newline **between** `SELECT` and `DISTINCT`.

src/sqlfluff/rules/L041.py
```diff
@@ -61,7 +76,12 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
             ]
             fixes = [
                 # E.g. "\n" -> " DISTINCT\n.
-                LintFix("edit", newline, replace_newline_with),
```
barrywhart, line 64: Unrelated fix, but... I have recently found several inappropriate uses of `"edit"` fixes. Edits replace a single existing segment with a new segment of the same type, and the specified segments as children. This is **very often a mistake**, because you're adding child nodes rather than sibling nodes. Changing the structure of the parse tree often violates the assumptions implicit in the dialect (i.e. the segment and grammar definitions), and is prone to causing very weird interactions (i.e. **bugs**) together with other rules.

tunetheweb:
Is the title right?

> Fix bug where L016 places newlines in the parse structure that confuse L016

Should that be:

> Fix bug where L016 places newlines in the parse structure that confuse other rules

barrywhart:
Oh, sorry. Yes, I should've said "L041" or "other rules". I'll change it to "other rules"; that'll be more readable for the change log.

tunetheweb (approved):
LGTM

A few questions (which might be beyond the scope of this change):

Does rule L041 handle multiple new lines (either in old code or new)?

Or does it not implicitly but does as we do up to 10 iterations I believe? In which case should it?

Should we have a test case for this?

barrywhart:
Actually, I worded it a bit differently. Wanted to make it clear that I was fixing L041 (the "victim"). From a deeper perspective, I **think** L016 is the culprit, i.e. its fixes leave the parse tree in a state you'd never see in a "fresh" parse tree (i.e. parsed directly from the file). See this [comment](https://github.com/sqlfluff/sqlfluff/issues/1808#issuecomment-959074153) on #1808 for some related ideas.

barrywhart:
Good question about multiple newlines. I'll look into that. Pretty sure at the moment, the rule would have to run repeatedly.

barrywhart:
I looked at the multiple newline case. I'm going to skip that for this PR, for the following reasons:
* The stated goal of this rule is to put `SELECT` and the modifier on the same line. It technically doesn't care about blank lines. If there **are** multiple blank lines, it won't remove them, because after the first fix, the rule is happy because `SELECT` and the modifier are on the same line. To do more here risks scope creep in the rule.
* Removing multiple newlines is trickier than it seems, because of comments. If we aren't careful, we can end up moving `DISTINCT` to the end of a comment line, which "buries" it and corrupts the SQL (because it becomes part of a comment).

tunetheweb:
So it does handle multiple newlines? It just doesn’t delete them all when fixing? That seems fine with me if so.

barrywhart:
Yes, it seemed to work fine.

PR #1916, opened 2021-11-16
New Rule L053: Remove outer brackets from top-level statements.

### LINKED ISSUE #1899: Statements surrounded by brackets still parse
If a statement is wrapped in brackets I would expect it to fail to parse. However, the statement still parses fine and throws no error.
e.g. 
```SQL
(SELECT foo
FROM bar)
```

## Expected Behaviour
Parse error as invalid SQL.

## Observed Behaviour
Screen-Recording-2021-11-13-at-22 47 45

## Steps to Reproduce
Lint the example query. It also parses INSERT/UPDATE/WITH statements wrapped in brackets.

## Dialect
ANSI (although issue is the same regardless of selected dialect)

## Version
`python 3.9.8`
`sqlfluff, version 0.8.1` (HEAD of main branch)

## Configuration
Default configuration.

jpy-git:

Fixes #1899. This PR adds a rule to remove outer brackets on top-level SQL queries. Refer to corresponding issue for discussion.
image

No. Small tweak to L050 to fix issue but has no downstream impact.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L053.py
+++ b/src/sqlfluff/rules/L053.py
@@ -67,7 +67,7 @@ class Rule_L053(BaseRule):
                 [
                     segment
                     for segment in context.segment.segments
-                    if segment.name not in bracket_set
+                    if segment.name not in bracket_set and not segment.is_meta
                 ],
             )
         ]
--- a/src/sqlfluff/rules/L053.py
+++ b/src/sqlfluff/rules/L053.py
@@ -46,10 +46,14 @@ class Rule_L053(BaseRule):
         """Top-level statements should not be wrapped in brackets."""
         # We only care about bracketed segements that are direct
         # descendants of a top-level statement segment.
+        if context.dialect.name == "tsql":
+            top_level_parent_stack_types = ["file", "batch", "statement"]
+        else:
+            top_level_parent_stack_types = ["file", "statement"]
         if not (
             context.segment.is_type("bracketed")
             and [segment.type for segment in context.parent_stack]
-            == ["file", "statement"]
+            == top_level_parent_stack_types
         ):
             return None
 
--- a/src/sqlfluff/rules/L053.py
+++ b/src/sqlfluff/rules/L053.py
@@ -46,14 +46,14 @@ class Rule_L053(BaseRule):
         """Top-level statements should not be wrapped in brackets."""
         # We only care about bracketed segements that are direct
         # descendants of a top-level statement segment.
-        if context.dialect.name == "tsql":
-            top_level_parent_stack_types = ["file", "batch", "statement"]
-        else:
-            top_level_parent_stack_types = ["file", "statement"]
         if not (
             context.segment.is_type("bracketed")
-            and [segment.type for segment in context.parent_stack]
-            == top_level_parent_stack_types
+            and [
+                segment.type
+                for segment in context.parent_stack
+                if segment.type != "batch"
+            ]
+            == ["file", "statement"]
         ):
             return None
 
--- a/src/sqlfluff/rules/L053.py
+++ b/src/sqlfluff/rules/L053.py
@@ -58,7 +58,7 @@ class Rule_L053(BaseRule):
             return None
 
         # Replace the bracketed segment with it's
-        # children, exluding the bracket symbols.
+        # children, excluding the bracket symbols.
         bracket_set = {"start_bracket", "end_bracket"}
         fixes = [
             LintFix(
--- a/src/sqlfluff/rules/L053.py
+++ b/src/sqlfluff/rules/L053.py
@@ -7,84 +7,45 @@ from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 
 @document_fix_compatible
 class Rule_L053(BaseRule):
-    """Lone statements should not be wrapped in brackets.
+    """Top-level statements should not be wrapped in brackets.
 
     | **Anti-pattern**
-    | The content in file begins with newlines or whitespace, the ^ represents the beginning of file.
+    | A top-level statement is wrapped in brackets.
 
     .. code-block:: sql
        :force:
 
-        ^
+        (SELECT
+            foo
+        FROM bar)
 
-        SELECT
-            a
-        FROM foo
+        -- This also applies to statements containing a sub-query.
 
-        -- Beginning on an indented line is also forbidden,
-        -- (the • represents space).
-
-        ••••SELECT
-        ••••a
-        FROM
-        ••••foo
+        (SELECT
+            foo
+        FROM (SELECT * FROM bar))
 
     | **Best practice**
-    | Start file on either code or comment, the ^ represents the beginning of file.
+    | Don't wrap top-level statements in brackets.
 
     .. code-block:: sql
        :force:
 
-
-        ^SELECT
-            a
-        FROM foo
-
-        -- Including an initial block comment.
-
-        ^/*
-        This is a description of my SQL code.
-        */
         SELECT
-            a
-        FROM
             foo
+        FROM bar
 
-        -- Including an initial inline comment.
+        -- Likewise for statements containing a sub-query.
 
-        ^--This is a description of my SQL code.
         SELECT
-            a
-        FROM
             foo
+        FROM (SELECT * FROM bar)
     """
 
-    targets_templated = True
-
-    @staticmethod
-    def _potential_template_collision(context: RuleContext) -> bool:
-        """Check for any templated raw slices that intersect with source slices in the raw_stack.
-
-        Returns:
-            :obj:`bool` indicating a preceding templated raw slice has been detected.
-        """
-        templated_file = context.segment.pos_marker.templated_file
-        for segment in context.raw_stack:
-            source_slice = segment.pos_marker.source_slice
-            raw_slices = templated_file.raw_slices_spanning_source_slice(source_slice)
-            if any(
-                raw_slice
-                for raw_slice in raw_slices
-                if raw_slice.slice_type == "templated"
-            ):
-                return True
-
-        return False
-
     def _eval(self, context: RuleContext) -> Optional[LintResult]:
-        """Files must not begin with newlines or whitespace."""
-        # If parent_stack is empty we are currently at FileSegment.
-        [segment.type for segment in context.parent_stack] == ["file", "statement"]
+        """Top-level statements should not be wrapped in brackets."""
+        # We only care about bracketed segements that are direct
+        # descendants of a top-level statement segment.
         if not (
             context.segment.is_type("bracketed")
             and [segment.type for segment in context.parent_stack]
@@ -92,6 +53,8 @@ class Rule_L053(BaseRule):
         ):
             return None
 
+        # Replace the bracketed segment with it's
+        # children, exluding the bracket symbols.
         bracket_set = {"start_bracket", "end_bracket"}
         fixes = [
             LintFix(
@@ -104,5 +67,5 @@ class Rule_L053(BaseRule):
                 ],
             )
         ]
-        # return None
+
         return LintResult(anchor=context.segment, fixes=fixes)
--- a/src/sqlfluff/rules/L053.py
+++ b/src/sqlfluff/rules/L053.py
@@ -1,4 +1,4 @@
-"""Implementation of Rule L050."""
+"""Implementation of Rule L053."""
 from typing import Optional
 
 from sqlfluff.core.rules.base import BaseRule, LintResult, LintFix, RuleContext
@@ -6,8 +6,8 @@ from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 
 
 @document_fix_compatible
-class Rule_L050(BaseRule):
-    """Files must not begin with newlines or whitespace.
+class Rule_L053(BaseRule):
+    """Lone statements should not be wrapped in brackets.
 
     | **Anti-pattern**
     | The content in file begins with newlines or whitespace, the ^ represents the beginning of file.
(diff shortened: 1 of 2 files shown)


WittierDinosaur:
It's worth checking if this has the desired effect in T-SQL, as the high level structure is different

jpy-git:
@WittierDinosaur as in for the same queries, or is the query structure different?

jpy-git:
@WittierDinosaur good shout, just checked and the TSQL dialect has an extra BatchSegment before the top-level statements. I've added the extra logic plus unit-tests 👍

jpy-git:

src/sqlfluff/rules/L050.py
```diff
@@ -101,8 +101,10 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
             # Non-whitespace segment.
             (context.segment.name not in whitespace_set)
             # We want first Non-whitespace segment so
-            # all preceding segments must be whitespace.
+            # all preceding segments must be whitespace
+            # and at least one is not meta.
             and all(segment.name in whitespace_set for segment in context.raw_stack)
+            and not all(segment.is_meta for segment in context.raw_stack)
```
barrywhart, line 107: Can you add a test for this change?

src/sqlfluff/rules/L053.py
```diff
@@ -0,0 +1,75 @@
...
+            foo
+        FROM (SELECT * FROM bar)
+    """
+
+    def _eval(self, context: RuleContext) -> Optional[LintResult]:
+        """Top-level statements should not be wrapped in brackets."""
+        # We only care about bracketed segements that are direct
+        # descendants of a top-level statement segment.
+        if context.dialect.name == "tsql":
+            top_level_parent_stack_types = ["file", "batch", "statement"]
```
barrywhart, line 50: I wonder if we could get this kind of information from the dialect, rather than storing it on the rules. Rules that are dialect specific will be harder to maintain.

src/sqlfluff/rules/L053.py
```diff
@@ -0,0 +1,75 @@
...
+            top_level_parent_stack_types = ["file", "statement"]
+        if not (
+            context.segment.is_type("bracketed")
+            and [segment.type for segment in context.parent_stack]
+            == top_level_parent_stack_types
+        ):
+            return None
+
+        # Replace the bracketed segment with it's
+        # children, exluding the bracket symbols.
```
barrywhart, line 61:
```suggestion
        # children, excluding the bracket symbols.
```

src/sqlfluff/rules/L050.py
jpy-git: So for this one I discovered it needed adding as part of this PR. The issue was the bracket would have a meta character before it at the start of the file. and L050 would try and delete it, but then it would come back (and so on...).

So I guess the unit test for this change is the unit tests i've added for this rule? Haven't found another place where it comes up. I could add the same test query to L050.yml if you want though?

src/sqlfluff/rules/L053.py
jpy-git: Is there another rule where we do something similar?

tunetheweb: Do we even need this to be t-sql specific? It seems "batch" is only use din t-sql so why not just include this as the default `top_level_parent_stack_types` and remove this if statement completely?

src/sqlfluff/rules/L053.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+
+    def _eval(self, context: RuleContext) -> Optional[LintResult]:
+        """Top-level statements should not be wrapped in brackets."""
+        # We only care about bracketed segements that are direct
+        # descendants of a top-level statement segment.
+        if context.dialect.name == "tsql":
+            top_level_parent_stack_types = ["file", "batch", "statement"]
+        else:
+            top_level_parent_stack_types = ["file", "statement"]
```
tunetheweb, line 52: Why do we need `"statement"` in here? Shouldn't the top level statement always be `"file"` (or `"batch"` for t-sql)?

jpy-git: @tunetheweb basically we need to find the first bracketed segment and say are its parents segments what we expect to find if it's the at the top level (syntactically speaking).

so in non-tsql the parse tree looks like:
FileSegment
|_StatementSegment
   |_BracketedSegment

and for tsql the parse tree gives:
FileSegment
|_BatchSegment
   |_StatementSegment
      |_BracketedSegment

So I'm basically just looking at the parents and seeing if it's what we'd expect

src/sqlfluff/rules/L053.py
jpy-git: I guess because I'm comparing this to the parent stack for an exact match to prove I'm at the top level

jpy-git: I could maybe change the filter logic to something like:
```python
if not (
    context.segment.is_type("bracketed")
    and [segment.type for segment in context.parent_stack if segment.type != "batch"]
    == ["file", "statement"]
):
    return None
```
so effectively just ignoring the batch.

It's not perfect as it still references batch (and therefore tsql) and potentially less clearly than before 🤷 Not sure of another way to check that the brackets are at the top-level?

src/sqlfluff/rules/L053.py
tunetheweb: Ah I've got you. It's because the BracketedStatement is part of a statement right.

So does the code below, basically say we have to right up the tree (until we get to the end - the `FileSegment`), and if we find any other type (other than those three), then it's not a top level segment so find to have brackets?

```
        if not (
            context.segment.is_type("bracketed")
            and [segment.type for segment in context.parent_stack]
            == top_level_parent_stack_types
        ):
            return None
```

src/sqlfluff/rules/L053.py
tunetheweb: Could you do what I suggested here: https://github.com/sqlfluff/sqlfluff/pull/1916#discussion_r751810281

So, for example this:

```sql
(SELECT * FROM table1)
UNION
(SELECT * FROM table2)
```
Is this:

```
file
|_statement
   |_set_expression
      |_bracketed
      |_set_operator
      |_bracketed
```

Therefore as it has a `set_expression` in it's parent tree back to `file` so it's not a top level bracket.

Whereas this:

```sql

(SELECT * FROM table1)
```
Is this:

```
file
|_newline
|_statement
   |_bracketed
```

So from `bracketed` back to `file` is `bracketed->statement->file` so it only contains the allowed types. So it is a top level statement.

Note I deliberately included a newline to show it's a parallel tree and so not part of the statement.

WDYT?

jpy-git: or could do `{segment.type for segment in context.parent_stack}.is_subset({"file", "batch", "statement"})` which serves the same purpose and is a bit more concise (If we don't care about ordering/counts)

src/sqlfluff/rules/L053.py
jpy-git: Exactly right, although because of the list I'm being even more specific and saying I expect the top_level_parent_stack_types to be in the exact order and number as in the list i'm comparing too 😄

src/sqlfluff/rules/L053.py
tunetheweb: Exactly. Can't see why that wouldn't work, can you?

src/sqlfluff/rules/L053.py
tunetheweb: As per [other conversation above](https://github.com/sqlfluff/sqlfluff/pull/1916#discussion_r751813853) that even more specific might be unnecessary. Resolving this, so we can continue conversation in that thread.

src/sqlfluff/rules/L053.py
jpy-git: @tunetheweb I think that makes sense, that's how the code works already. I guess there's a question around whether we should also be able remove brackets around those statements in the set expression?

tunetheweb: > I guess there's a question around whether we should also be able remove brackets around those statements in the set expression?

I don't think so. It's not a top level segment - it's two bracketed sub segments with an unbracketed `UNION` in between.

tunetheweb: In fact I’m not even sure we should count `batch` in this. Is that for this use case?

```sql
BEGIN
  (SELECT 1)
END
```

I don’t think the rule should flag that as not top level either.

I think keep it simple just to File and Statement and that’s sufficient.

jpy-git: > Exactly. Can't see why that wouldn't work, can you?

I think if we know that we're looking for 'file->(batch)->statement->brackets' exactly then we might as well be sure that our parent chain matches that.

The set method would concern me if we could match 'file->(batch)->statement->statement->statement->brackets' which, although I can't visualise what that is, probably wouldn't be top level

jpy-git: > Exactly. Can't see why that wouldn't work, can you?

I think if we know that we're looking for 'file->(batch)->statement->brackets' exactly then we might as well be sure that our parent chain matches that.

The set method would concern me if we could match 'file->(batch)->statement->statement->statement->brackets' which, although I can't visualise what that is, probably wouldn't be top level

jpy-git: @tunetheweb actually the begin/end case you mention above has parents like 'file->(batch)->begin_end_block->statement->brackets' so wouldn't be touched by this rule, which is what we expect

image

jpy-git: Whereas 'file->(batch)->statement->brackets' covers the top level case we want to catch:
image

tunetheweb: Ah OK for T-SQL everything is in a "batch". Gotcha.

OK then I agree 'file->(batch)->statement->brackets' is the way to go, and no explicit override for T-SQL.

jpy-git: perfect, I've updated the logic to remove the tsql reference. I've also added extra unit tests to validate that we aren't getting the false positives discussed 👍

tunetheweb (approved):

src/sqlfluff/rules/L050.py
barrywhart: This sounds like a case where you need a test that runs multiple rules. Tests like this are structured a bit differently, and they can be added to a subdirectory of `test/fixtures/linter/autofix`. Can you do that? That sounds better than simply adding the **same** test to L050.yml, because it sounds like it's the **interaction** of the rules that needs testing.

test/fixtures/rules/std_rule_cases/L053.yml
```diff
@@ -0,0 +1,67 @@
...
+  fail_str: |
+    (SELECT foo
+    FROM (select * from bar))
+  fix_str: |
+    SELECT foo
+    FROM (select * from bar)
+
+test_pass_set_statement_brackets:
+  pass_str: |
+    (SELECT 1)
```
barrywhart, line 26: It's tempting to say the rule should remove these parentheses as well. Thoughts?

barrywhart (changes requested):

src/sqlfluff/rules/L050.py
jpy-git: @barrywhart Actually this issue was caused by the list comprehension in this rule

originally it was:
```python
[segment for segment in context.segment.segments if segment.name not in bracket_set],
```
essentially strips out the brackets but not the metas between the brackets and the inner statement (thats were the bad interaction with L050 described above was coming from).

I since changed that list comprehension to:
```python
[segment for segment in context.segment.segments if segment.name not in bracket_set and not segment.is_meta]
```
i.e. also exclude the metas, and just extract the wrapped statement.

so this interaction doesn't actually exist between the rules anymore so not sure adding it as a test would really validate anything?

jpy-git: @barrywhart had a look at `test/fixtures/autofix` and for this case it would literally just be a repeat of the existing unit test, since L050 is (rightly) not even triggered anymore:
image

I think it still makes sense to keep this adjustment in as it fixed the potential issue before my list comprehension fix above, but alternatively can just remove the change to this file as it technically wouldn't break any unit tests

test/fixtures/rules/std_rule_cases/L053.yml
jpy-git: @barrywhart I think we could go either way on this.

This might be an example of a top level query (it doesn't parse in sqlfluff but is valid in actual sql) that would be covered by this rule.
```SQL
(
(SELECT 1)
UNION
(SELECT 1)
)
```

if we were to cover the lower-level selects as well (which i agree is a nice to have) we would also have to cover things like
```SQL
SELECT *
FROM (
    (SELECT 1)
    UNION
    (SELECT 1)
)
```

I think if we were to do that we would have to start including a bunch of conditions to the rule like 
```
- if we find brackets whose direct parent is a set operator then remove
- if we find brackets whose direct parent is a begin/end statement then remove etc
```
Which we can definitely do, but maybe we consider those as an enhancement at a later date to this rule given the rule currently fulfils its stated aim and it would just be a matter of adding additional conditions?

jpy-git: are there other cases that you would want removed?

tunetheweb: My view is this is out of scope:

```sql
(SELECT 1)
UNION
(SELECT 1)
```

These are not top level brackets. And it's going to get much more complicated if we start trying to include those.

Plus the above example is good practice in a lot of cases to clearly differentiate between the two sub-clauses.

This SQL, however, has no value in the brackets - and that was what we were trying to address (or at least that was my understanding):

```sql
(SELECT 1)
```

jpy-git: exactly, think thats the pragmatic aproach 👍

src/sqlfluff/rules/L050.py
barrywhart: Now that I know the full context, it's fine to leave it as is. Testing decisions often involve nuance, and things evolve as you work on them. 😅 

As I often say, if something like this comes up again, we can always add a test then. It's so hard to guess how likely that is... 1%? 50% 🤷

test/fixtures/rules/std_rule_cases/L053.yml
barrywhart: Sounds good to me! I hope I'm not being a nuisance -- sometimes I ask questions just to get people's thoughts, whether or not I have a strong opinion myself. As you say, scope creep can be a real problem -- often it's better to get a first version and enhance later as needed. There's plenty of other important work to do on the project, I'm sure.

Thanks for the excellent work -- writing rules is one of the trickiest areas of SQLFluff to learn.

barrywhart (approved):
Good to merge once the build finishes. Nice work! 🙌

jpy-git:
Thanks both!

test/fixtures/rules/std_rule_cases/L053.yml
jpy-git: @barrywhart It's honestly great, the feedback is super insightful and having the discussions in the PR means we can refer back to this for future developments/issues 😄 
Thanks!

PR #1917, opened 2021-11-16
New Rule L054: Inconsistent column references in GROUP BY/ORDER BY clauses.

jpy-git:

This PR adds a new rule to prevent the use of implicit positional references in GROUP BY clauses.

Screen-Recording-2021-11-16-at-20 55 35

N.B. this rule does not provide an automatic fix (reasons discussed below)

No

### Pull Request checklist

  - Other.


jpy-git:
The reasoning for not providing an automatic fix is as follows:

We could provide a fix by back-checking the preceding SELECT statement in the simple case e.g.
columns in a valid order with no typecasting functions or aliasing (refer to `initial fix logic` commit if you want to see this)
```SQL
SELECT
    fake_column_1,
    fake_column_2,
    sum(fake_value) AS sum_value
FROM fake_table
GROUP BY
    1, 2;
```

However, due to the nature of the grammar definitions and the lack of interaction between the SELECT and GROUP BY clauses, the parser cannot raise a parse error for invalid column/positional reference combinations. As such attempting to provide a fix may end up causing more confusion to the end user and therefore I deemed it sensible to just raise the lint error and allow the user to correct.

Examples are:
1. Differing number of SELECT columns and GROUP BY columns e.g.
```SQL
SELECT
    fake_column_1,
    fake_column_2,
    fake_column_3,
    sum(fake_value) AS sum_value
FROM fake_table
GROUP BY
    1, 2;
```
or
```SQL
SELECT
    fake_column_1,
    sum(fake_value) AS sum_value
FROM fake_table
GROUP BY
    1, 2, 3;
```
should be a parse error. 1st case we could replace cols 1 & 2 but that wouldn't actually fix the query. 2nd case we could fix col 1, but 2 and 3 aren't possible group by columns.

2. Wildcards in select clause
```SQL
SELECT
    *,
    fake_column_1,
    fake_column_2,
    sum(fake_value) AS sum_value
FROM fake_table
GROUP BY
    1, 2;
```
We can't infer the number/ordering of columns in the source of the wildcard expression so shouldn't try.

3. Aggregate function out of order
```SQL
SELECT
    fake_column_1,
    sum(fake_value) AS sum_value
    fake_column_2,
FROM fake_table
GROUP BY
    1, 2;
```
We can fix col 1 but col 2 is supposedly referring to the aggregating function which is not valid. Even if we just tried to sub in for 1 and leave 2, this would get messed up by the reordering from L034 and could lead to us substituting in `fake_column_2` for 2 which the user may not be intending.

4. Repeated column references in GROUP BY
```SQL
SELECT
    fake_column_1,
    fake_column_2,
    fake_column_3,
    sum(fake_value) AS sum_value
FROM fake_table
GROUP BY
    1, fake_column_2, 2;
```
or 
```SQL
SELECT
    fake_column_1,
    fake_column_2,
    fake_column_3,
    sum(fake_value) AS sum_value
FROM fake_table
GROUP BY
    1, 2, 2;
```
again mistake by the user, but could cause more confusion if we try and start inserting the column names.

5. column reference vs alias. depending on the dialect you can refer to either the column or an alias in the group by clause
```SQL
SELECT
    fake_column_1,
    fake_column_2 AS alias_2,
    sum(fake_value) AS sum_value
FROM fake_table
GROUP BY
    1, 2;
```
I can see a lot of complaints here over whether we use fake_column_2 or alias_2 to sub for 2.

6. casting/formating functions vs aggregating functions:
```SQL
SELECT
    cast(fake_column_1 as varchar(255)) AS alias_1
    sum(fake_value) AS sum_value
FROM fake_table
GROUP BY
    1, 2;
```
we would have to infer for any functions whether or not they are an aggregating function or whether we can extract a column reference from the casting/aggregating function.

jpy-git:
Those are some of the reasons I felt it best to just highlight the linting issue and allow the user to fix.

This rule is one I use in my own team at work and the logic for just linting is concise and effective. So unless anyone has any objections I think we should not worry about an autofix for this rule.

We can always add the autofix part at a later date as well if we desire 😄

tunetheweb:
No massive opinion on whether to autofix or not, except to that I most of those issues aren’t problems with the rule itself but bad SQL to begin with. This rule is about not using positional parameters, not checking the validity of those positional parameters. So we could autofix what we could and leave the rest as is and think that would be fine. Anyway not a big deal.

However, I do have two a bigger questions (without having looked at the code at all - apologies!)

1)  Whether this rule should be configurable to allow or disallow positional GROUP BY params? Personally I don’t like it it and much prefer the explicitness of the column names (which I suspect is similar for you!). However some think the opposite and think SELECT columns should be in GROUP BY and ORDER BY order, and that this is simpler, avoids repetition and so is recommended.

So rather than take the opinionated route you have here, should you offer the reserve as a configurable option? Your preference can still be the default (mostly coz I agree with it 😁) but think it would be good to offer both options if it’s not to much effort.

2) Why only GROUP BY? Doesn’t the exact same not apply to ORDER BY? Could add an additional param for that, but imagine both would usually follow the same convention so that may not be necessary.

jpy-git:
@tunetheweb good points! I agree most of the possibilities would be the fault of the user entering bad SQL, my concern would just be making bad substitutions off the back of that. I'd rather their code fail because of them than something we change. My gut feeling is saying leave out autofix for this initial PR.

On point 1: my thought was that this rule is pretty much binary, you either allow positional params or you don't. So rather than having a config option to alter the behaviour of the rule you would just ignore the rule in your config if you didn't want it?

On point 2: haha I completely forgot that ORDER BY also allowed positional column references, I definitely agree that it should be included in this rule as the logic should be pretty similar.

tunetheweb:
> On point 1: my thought was that this rule is pretty much binary, you either allow positional params or you don't. So rather than having a config option to alter the behaviour of the rule you would just ignore the rule in your config if you didn't want it?

But what if someone preferred positional params and wanted to enforce them always?

Would they then have to implement a new rule on top of this? That just leads to possibility of clashes if both are (stupidly!) configured on?

To me it’s like preferring upper or lower case keywords - our rules allow both with a configurable param.

Or do we just say that if that was desired then someone would extend this rule then but for now we only offer enforcing no positional params?

jpy-git:
Ahh apologies I understand now 😄 Let me rephrase the wording of the rule then so it's a choice between consistent use of either explicit or positional references!

jpy-git:
@tunetheweb I've implemented the changes we discussed. 

There's now a config option `group_by_and_order_by_style` which can be set to either `explicit` (default) or `implicit` column references (see unit tests for examples). The rule works for both GROUP BY and ORDER BY clauses 😄

tunetheweb:
> @tunetheweb I've implemented the changes we discussed.
> 
> There's now a config option `group_by_and_order_by_style` which can be set to either `explicit` (default) or `implicit` column references (see unit tests for examples). The rule works for both GROUP BY and ORDER BY clauses 😄

No `consistent`? That might be a better default if it is possible to add. Less opinionated so less likely to break.

jpy-git:
@tunetheweb sure thing, I've changed the default to consistent 👍

jpy-git:

src/sqlfluff/core/default_config.cfg
```diff
@@ -43,6 +43,7 @@ comma_style = trailing
 allow_scalar = True
 single_table_references = consistent
 unquoted_identifiers_policy = all
+group_by_and_order_by_style = consistent
```
tunetheweb, line 46: Should this be a rule-specific config? Can't see it being needed by other rules.

src/sqlfluff/core/rules/config_info.py
```diff
@@ -103,6 +103,13 @@
             "Should alias have an explict AS or is implicit aliasing required?"
         ),
     },
+    "group_by_and_order_by_style": {
+        "validation": ["consistent", "implicit", "explicit"],
+        "definition": (
+            "Should GROUP BY/ORDER BY require explict column references "
+            "or are implicit positional column references required?"
```
tunetheweb, line 110: Should we make this more consistent with this:

```
    "single_table_references": {
        "validation": ["consistent", "qualified", "unqualified"],
        "definition": "The expectation for references in single-table select",
    },
```

```suggestion
            "The expectation for using explicit column name references "
            "or implicit position references"
```

Also is `implicit`/`explicit` the right values? Should it be `names`/`positions`?

test/fixtures/rules/std_rule_cases/L054.yml
```diff
@@ -0,0 +1,301 @@
+rule: L054
```
tunetheweb, line 1: By `consistent` I was more thinking this should be caught:

```sql
    SELECT
        foo,
        bar,
        sum(baz) AS sum_value
    FROM fake_table
    ORDER BY
        foo, bar
    GROUP BY
        1, 2;
```

Is that caught?

Should we add a test case for this? And vice versa (positional ORDER BY, named GROUP BY)?

tunetheweb (commented):
Nice code!

Do have some (more!) observations/questions - sorry!

jpy-git:
> Nice code!
> 
> Do have some (more!) observations/questions - sorry!

No worries, they're all sensible 😄

src/sqlfluff/rules/L054.py
```diff
@@ -0,0 +1,111 @@
...
+
+    config_keywords = ["group_by_and_order_by_style"]
+
+    def _eval(self, context: RuleContext) -> Optional[LintResult]:
+        """Inconsistent column references in GROUP BY/ORDER BY clauses."""
+        # Config type hints
+        self.group_by_and_order_by_style: str
+
+        # We only care about GROUP BY/ORDER BY clauses.
+        if context.segment.name not in {"GroupByClauseSegment", "OrderByClauseSegment"}:
```
barrywhart, line 83: Rules usually use `is_type()` instead, something like:
```
context.segment.is_type("groupby_clause", "orderby_clause")
```

I think they're functionally identical, but I'd prefer we change this to be consistent with other rules.

src/sqlfluff/core/rules/config_info.py
jpy-git: So I think either are fine, you explicitly refer to the columns by name or implicitly by postition. Could be swayed either way if you have a strong preference?

src/sqlfluff/rules/L054.py
```diff
@@ -0,0 +1,111 @@
...
+        child_segment_names = [segment.name for segment in context.segment.segments]
+        disallowed_segment_names = {
+            "explicit": "numeric_literal",
+            "implicit": "ColumnReferenceSegment",
+        }
+
+        if self.group_by_and_order_by_style == "consistent":
+            # If consistent naming then raise lint error if both
+            # numeric literal and column reference segments are present.
+            if all(
```
barrywhart, line 96: I'm struggling to understand how this code works -- I _think_ I understand, but it seems like a pretty indirect way of doing it. Can you think of a clearer way to express this logic?

src/sqlfluff/rules/L054.py
```diff
@@ -0,0 +1,111 @@
...
+
+        # We only care about GROUP BY/ORDER BY clauses.
+        if context.segment.name not in {"GroupByClauseSegment", "OrderByClauseSegment"}:
+            return None
+
+        # Look at child segments to detect the presence of disallowed segments.
+        child_segment_names = [segment.name for segment in context.segment.segments]
+        disallowed_segment_names = {
+            "explicit": "numeric_literal",
+            "implicit": "ColumnReferenceSegment",
```
barrywhart, line 90: `ORDER BY` and `GROUP BY` also support expressions. We should make sure the rule handles these correctly, e.g.:
```
select * from foo order by c1 * c1;
select * from foo order by power(c1, 2);
select * from foo order by 1 / c1;
```

src/sqlfluff/rules/L054.py
jpy-git: sure it basically just says if the ORDER BY/GROUP BY clause contains both numeric and column reference values then it's inconsistent

src/sqlfluff/core/rules/config_info.py
tunetheweb: OK I'm fine with names as is. Do think we should change comment though 1) to be more consistent and 2) as kinda of covers consistent use case more.

src/sqlfluff/rules/L054.py
tunetheweb: I thought it was pretty clear. And was one of the reasons for me "nice code" comment.

Can you explain more what you're struggling with @barrywhart ? Maybe I'm assuming something that's not true!

src/sqlfluff/rules/L054.py
tunetheweb: Would be good to add a test case for this. I was assuming it would be ignored as neither numeric nor column reference is a direct child but would be good to confirm.

jpy-git: @barrywhart Sure I can account for those, I think surely an expression must contain explicit column references, as `1 / 1` or `power(1, 2)` couldn't be valid with positional references. So can simplify the logic by considering an expression as explicit?

src/sqlfluff/rules/L054.py
jpy-git: will do

test/fixtures/rules/std_rule_cases/L054.yml
jpy-git: good spot, I'll look at using memory to remember the other clauses.

src/sqlfluff/rules/L054.py
barrywhart: I think so yes (i.e. expressions are explicit).

src/sqlfluff/rules/L054.py
barrywhart: It took me a while to figure out that we were checking if all segment name categories appeared in the list, that was wrong. The code is correct. It would read more clearly to me if it were using sets, and a set with size > 1 was an error.

This is not a big worry for me -- good to know that it's readable for you. I do like the functional style. 👍

test/fixtures/rules/std_rule_cases/L054.yml
jpy-git: Added this in and working well :)

src/sqlfluff/core/rules/config_info.py
jpy-git: I've updateed the message in the latest commit

src/sqlfluff/core/default_config.cfg
jpy-git: moved to the rule specific section

src/sqlfluff/rules/L054.py
jpy-git: used the types now as you said, I still use the names in constructing `column_reference_category_set` as the numeric_literal segment type is `raw` which is unhelpful. I've made a note in the code

src/sqlfluff/rules/L054.py
jpy-git: I've updated the code flow so should hopefully be a lot clearer now :)

src/sqlfluff/rules/L054.py
jpy-git: @barrywhart I've added the unit tests for explicit expressions and works fine for the `consistent`/`explicit` config options.

However, if a user were to set the config option as implicit (so they require all positional references) they would get unfixable lint errors if they were to use an expression as it is explicit and can't be made implicit. This isn't an issue with the code as the logic is consistent, so my thought is if they want to use expressions they shouldn't choose to force implicit group by references.

jpy-git: Basically they are setting inappropriate configs for their use case if they're using expressions and should consider using `consistent` instead, so don't see this as an issue 👍

tunetheweb (commented):
LGTM. And very thorough test cases.

@barrywhart do you want to look over the set and memory stuff before merging? As that's all a. bit beyond me 😀

tunetheweb (approved):

barrywhart:
I'll review this again. It has merge conflicts, so it'll require some updates before merging.

src/sqlfluff/rules/L054.py
```diff
@@ -0,0 +1,134 @@
...
+            return None
+
+        # Look at child segments and map column references to either the implict or explicit category.
+        # N.B. segment names are used as the numeric literal type is 'raw', so best to be specific with the name.
+        column_reference_category_map = {
+            "ColumnReferenceSegment": "explicit",
+            "ExpressionSegment": "explicit",
+            "numeric_literal": "implicit",
+        }
+        column_reference_category_set = {
```
barrywhart, line 93: Nice! 💯

src/sqlfluff/rules/L054.py
```diff
@@ -0,0 +1,134 @@
...
+            for segment in context.segment.segments
+            if segment.name in column_reference_category_map
+        }
+
+        if self.group_by_and_order_by_style == "consistent":
+            # If consistent naming then raise lint error if either:
+
+            if len(column_reference_category_set) > 1:
+                # 1. Both implicit and explicit column references are found in the same clause.
+                return LintResult(anchor=context.segment)
```
barrywhart, line 104: I notice we're not returning `memory` when the rule finds an issue with the code. Other rules such as [L003](https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/rules/L003.py) *do* return `memory` in this case, so I wonder if this is a bug.

jpy-git:
@barrywhart the merge conflicts were just because L052 and L054 were both adding to the config files. Should be good for review now 👍

barrywhart (commented):
I noticed a possible bug with the use of memory. Can you update this and/or confirm that it's okay "as is" (possibly adding more tests)?

barrywhart (changes requested):
Marking as "request changes" because I found a potential bug that should be looked into. If we determine it's not a bug, I'm happy to approve and merge the PR.

src/sqlfluff/rules/L054.py
jpy-git: @barrywhart you were correct, not returning memory for all LintResults could cause a bug in the case:
- `consistent` setting
- first group by/order by clause is explicit.
- subsequent group by/order by clauses are implicit.
- first subsequent group by/order by would raise lint error but not return memory
- therefore can't detect any subsequent additional errors.

I have added the memory to the LintResults and added an extra test case to check that multiple errors on `consistent` setting are caught 😄

jpy-git:
@barrywhart resolved the issue raised in code review and added an extra unit test (details discussed in response to review comment)

barrywhart (approved):

barrywhart:
The changes look good!

I was thinking about the `memory` thing -- now that the rules take a `context` object with `memory` attached, it seems strange to require the rule to return it. I'll create an issue about removing that requirement.

barrywhart:
>you would still need some sort of mechanism to update the memory attribute of the global context object.
You should be able to just modify it "in place" on the context object. From an OO point of view, some people may say that's bad, but the context is barely an object in the OO sense, just a convenient, loose set of public fields.

jpy-git:
seems sensible, worth pulling something together to look at :)

barrywhart:
I created an issue about it: #1943. If you're interested in working on it, that'd be great! I think it's cool and valuable to take lessons from specific work (e.g. new rules or dialect changes) and turn that into more general, "cross-cutting" improvements to the core.

PR #1923, opened 2021-11-17
L043 refinement: Add autofix for common use of CASE to fill NULL values.

### LINKED ISSUE #1797: New rule: COALESCE instead of CASE/IF when possible
In all dialects that have `COALESCE` code such as `CASE WHEN num_cats IS NULL THEN 0 ELSE num_cats END` are bad practices because they are very long and need to be simplified as `COALESCE(num_cats, 0)` which is much shorter and a lot more readable.

jpy-git:

This PR enhances L043 to fix #1797. I have also tidied up the docstring and added the corresponding unit tests.

L043 should now flag and fix more cases of unnecessary use of CASE. 

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L043.py
+++ b/src/sqlfluff/rules/L043.py
@@ -184,7 +184,7 @@ class Rule_L043(BaseRule):
                 coalesce_arg_1 = condition_expression
                 coalesce_arg_2 = KeywordSegment("false")
                 coalesce_arg_1_idx = condition_expression_idx
-                preceding_not = (then_expression.raw_upper == "FALSE")
+                preceding_not = then_expression.raw_upper == "FALSE"
 
                 fixes = self._coalesce_fix_list(
                     context,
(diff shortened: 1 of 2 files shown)


jpy-git:

src/sqlfluff/rules/L043.py
```diff
@@ -1,154 +1,261 @@
...
-                            KeywordSegment("not"),
-                            WhitespaceSegment(),
-                        ]
-                    )
-                # Add coalesce and parenthesis
-                edits.extend(
+                coalesce_arg_1 = condition_expression
+                coalesce_arg_2 = KeywordSegment("false")
+                coalesce_arg_1_idx = condition_expression_idx
+                preceding_not = True if then_expression.raw_upper == "FALSE" else False
```
barrywhart, line 187:
```suggestion
                preceding_not = (then_expression.raw_upper == "FALSE")
```

barrywhart (approved):
Looks good to me.

This rule has always been one of the scarier ones to me, because I worry about the risk of "false positives", like if it triggers for a corner case we didn't consider. For example, I think an early version of this rule assumed (incorrectly) that if a case statement used booleans and an else, that one boolean would be false and the other true. Can you think of any potential "gotchas" with these changes, and perhaps we should add more tests to catch them?

src/sqlfluff/rules/L043.py
jpy-git: @barrywhart 🤦 wow I definitely had a brain lapse there, thanks!

jpy-git:
@barrywhart so I added some checks to make sure we only operate on CASE statements with a single WHEN and also added a check that the IS NULL condition is the only condition. Iet me make some extra unit tests for cases we think should pass this lint unaffected for extra peace of mind 👍

jpy-git:
@barrywhart ive added additional unit test to validate my checks in the code so this should be good to go 🚀

PR #1860, opened 2021-11-09
Tox improvements: Streamline development/testing environments.

jpy-git:
### Brief summary of the change made

Primary goal is to improve `tox` setup for the project:
- Remove unnecessary installation of requirements in `requirements.txt` and editable package in `testenv`. `tox` already does this when you create the tox environment (it gets package details from `setup.py`/`setup.cfg`). If you want to create a virtual environment with everything pre-installed for you, just use:
```shell
tox -e py38 --devenv .venv && source .venv/bin/activate
```
- Above now means that in conjunction with #1851 we are now no longer installing the sqlfluff package and it's requirements in static analysis environments, which will cut down testing time.
- Remove previous use of `usedevelop = true` for `testenv`, we shouldn't be doing this for unit tests as we want to be testing as if we had downloaded the package. The `--devenv` option gives you a virtual environment with a editable package install for development, but we now use the `sdist` for unit testing with `tox -e py38` etc.
-  Moved `requests` into `requirements_dev.txt`, it's used in util.py so just include as a dev requirement for consistency. 
- Updated CONTRIBUTING.md to highlight the use of `tox --devenv` to quickly get started with development work.

### Are there any other side effects of this change that we should be aware of?
No impact on package, just development/testing improvements listed above.

### Pull Request checklist
- [X] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `python test/generate_parse_fixture_yml.py` or by running `tox` locally).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

Happy to discuss/provide further clarification if needed 😄

PR #1928, opened 2021-11-18
Add missing chardet install in setup.py

### LINKED ISSUE #1925: Add chardet to setup.py installation requirements
## Expected Behaviour

Able to run sqlfluff after `pip install sqlfluff==0.8.1`

## Observed Behaviour
```
import chardet
ModuleNotFoundError: No module named 'chardet'
```

## Steps to Reproduce

`pip install sqlfluff==0.8.1` on a system that doesn't have `chardet` installed.

## Version
sqlfluff version 0.8.1
python version 3.6.2

jpy-git:

Fixes #1925. chardet was missing from setup.py. 

The reason we likely missed this in unit tests was because in tox.ini we install requirements.txt (which had chardet), followed by a developer install of sqlfluff, rather than installing the package and its requirements via just setup.py. This is something I've already resolved in #1860 so just need that merged in to prevent this in the future.

No, fixes missing requirement

### Pull Request checklist

  - Other.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -208,9 +208,15 @@ jobs:
     - name: Run the version test
       run: |
         sqlfluff --version
-    - name: Run a simple select parse test
+    - name: Run a simple select parse test via stdin
       run: |
         echo "select 1" | sqlfluff parse -
-    - name: Run a simple select lint test
+    - name: Run a simple select lint test via stdin
       run: |
         echo "select 1" | sqlfluff lint -
+    - name: Run a simple select parse test via file
+      run: |
+        sqlfluff parse <(echo "select 1")
+    - name: Run a simple select lint test via file
+      run: |
+        sqlfluff lint <(echo "select 1")
--- a/setup.py
+++ b/setup.py
@@ -121,6 +121,8 @@ setup(
         "tblib",
         # For handling progress bars
         "tqdm",
+        # To get the encoding of files.
+        "chardet",
     ],
     entry_points={
         "console_scripts": [


tunetheweb:
Hmmmm #1843 was supposed to prevent this happening. Any ideas why that didn't pick this up this time?

jpy-git:
@tunetheweb because `pip-test-pull-request` only checks parse/lint of a statement via stdin. `chardet` is used to detect character encoding of a file so those tests don't touch the bit of the code that tries to `import chardet`

tunetheweb:
OK cause we're using stdin for our test. Which I specifically choose as wouldn't have access to the repo.

Wonder should we enhance that as part of this to also add a test like this:

```yml
    - name: Run a simple select lint test with a file
      run: |
        echo "select 1" > test.sql
        sqlfluff lint test.sql
```

WDYT?

jpy-git:
yeah makes sense!

or could do `sqlfluff lint <(echo "select 1")` which mimics file input

jpy-git:
maybe do one via std in and one via file

jpy-git:
that plus #1860 should put us in a pretty good position to catch these in the future

tunetheweb:
Cool can you add to this PR then? Will have another look at #1860

jpy-git:
got those extra pip test added :)

barrywhart (approved):

barrywhart:
Good catch! This has probably been broken for months. 😅

jpy-git:
We should look at removing `requirements.txt` at some point as `tox` handles all the requirements from `setup.py`, so `requirements.txt` is unnecessary and likely to cause confusion.

Similar for `requirements_dev.txt` as those dependencies can just go in their own tox environment. (i.e run tox -e flake8) to run flake8 in it's own environment

barrywhart:
I think the txt files are useful for ensuring stable CI builds. A new dependent package may break SQLFluff, but it shouldn't break the build.

One technique I've found very useful is to have two flavors of the build, one using just setup.py and one using txt files. If both break, it's a bug in our code. If only the former one breaks, that's an issue with a dependency. Saves a lot of "red herring" issues...

jpy-git:
> One technique I've found very useful is to have two flavors of the build, one using just setup.py and one using txt files. If both break, it's a bug in our code. If only the former one breaks, that's an issue with a dependency. Saves a lot of "red herring" issues...

but surely if you just run `tox -e py` it installs the deps first from setup.py so if there was an error in the dependencies install you'd just see it in the log?
The only places requirements.txt is used in the ci is `check-yaml-files-generation.yml` and `.readthedocs.yml`  both of which should be replaced with tox for consistency

tunetheweb:
We've never convinced @barrywhart to use `tox` 😁

PR #1930, opened 2021-11-18
New Rule L056: 'SP_' prefix should not be used for user-defined stored procedures

### LINKED ISSUE #1745: New TSQL rule: don't create stored procedures with sp_ prefix
https://docs.microsoft.com/en-us/sql/t-sql/statements/create-procedure-transact-sql?view=sql-server-ver15#procedure_name

The sp_ prefix is used by Microsoft to identify system procedures.  Using the sp_ prefix affects performance slightly since SQL Server performs an unnecessary check against the system schema.  It could also break processes if there is a naming conflict.

We should be able to create a linter for TSQL that identifies and flags the creation of stored procedures with this issue.

Anti-pattern:
```
CREATE PROCEDURE dbo.sp_pull_data
AS
SELECT
    [ID],
    [DataDate],
    [CaseOutput]
FROM table1
```
Pattern:

```
dbo.pull_data
```
or
```
dbo.usp_pull_data
```

jpy-git:

This PR aims to prevent users in T-SQL from creating stored procedures that clash with the system defined stored-procedures.
Fixes #1745 

Screenshot from T-SQL docs of why this rule is required:
image

No

### Pull Request checklist

  - Other.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -208,9 +208,15 @@ jobs:
     - name: Run the version test
       run: |
         sqlfluff --version
-    - name: Run a simple select parse test
+    - name: Run a simple select parse test via stdin
       run: |
         echo "select 1" | sqlfluff parse -
-    - name: Run a simple select lint test
+    - name: Run a simple select lint test via stdin
       run: |
         echo "select 1" | sqlfluff lint -
+    - name: Run a simple select parse test via file
+      run: |
+        sqlfluff parse <(echo "select 1")
+    - name: Run a simple select lint test via file
+      run: |
+        sqlfluff lint <(echo "select 1")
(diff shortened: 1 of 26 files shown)


tunetheweb (changes requested):
This shows as this:

```
L:   1 | P:  26 | L056 | 'SP\_' prefix should not be used for user-defined stored
                       | procedures in T-SQL.
```

Is that escape supposed to be there?

Also could you post a screenshot of the Rules page to see if it has the same issue? Details on how to build in `docs/README.md`

jpy-git:
@tunetheweb So the escape character is required by sphinx otherwise the docbuild fails.
(see https://github.com/sphinx-doc/sphinx/issues/4350)
Including the backslash makes the docs work:
image

That's annoying that the backslash now shows up in the docstring, but the solution is to just pass that message to the LintResult via the description parameter and now both are working:
image

tunetheweb (approved):

jpy-git:
@tunetheweb this should be good to merge in now 🚀

PR #1945, opened 2021-11-20
Official Docker image: Dockerfile and Github Actions workflow

### LINKED ISSUE #1874: Official Docker image
I'd love to be able to use SQLFluff for workloads running on Kubernetes. There are a few unofficial SQLFluff images [floating around on Dockerhub](https://hub.docker.com/search?q=sqlfluff&type=image) but nothing official.

Are there plans to maintain an official docker image somewhere? I'd be happy to help with the effort if so.

jpy-git:
This PR aims to introduce the Dockerfile and Github Actions workflow necessary to provide an Official Docker image for SQLFluff. Fixes #1874. 

Note for Admin (@alanmcruickshank): 
- Could you create a DockerHub user called `sqlfluff`. 
- Then in DockerHub `Account Settings > Security` create an access token to use in the Github Action.
- Then in the SQLFluff GitHub repository please add two secrets, `DOCKERHUB_USERNAME`: `sqlfluff` and `DOCKERHUB_TOKEN`: `<token you created in DockerHub>`.
- Then happy days 😄 

The action is setup (similar to PyPi publish) such that when we publish a new SQLFluff release:
- The `amd64` image is built.
- This image is then tested to ensure it is working.
- Then we build the `arm64` image (for M1 mac users) and push the images for both architectures to DockerHub.
- The images are tagged with both latest (so that by default users pull the latest release) and the specific version number from the release (so users can use a specific version if they require).

To use the image users can then run:
`docker run --rm -it -v <local_mount_path>:<container_mount_path> sqlfluff/sqlfluff lint <containter_mount_path>/<files to lint>`
or for a specific version:
`docker run --rm -it -v <local_mount_path>:<container_mount_path> sqlfluff/sqlfluff:<version_number> lint <containter_mount_path>/<files to lint>`
(this also works with fix/parse/etc if you replace lint in the lines above)

For testing I have uploaded the image to my personal DockerHub, so if you want to give it a quick go whilst the official user is being set up:
```console
echo 'SELECT * FROM test_table;' > test.sql
docker run --rm -it -v $PWD:/sql jpydocker/sqlfluff lint /sql/test.sql
```
image


jpy-git:
One point for discussion is where you want me to create the README/instructions info.
I want to get your opinions before I start documenting.

As you can see below DockerHub has fields for a short description and a README:
image
I would suggest that the short description is just something like "Official SQLFluff Docker image" and for the README there are a couple options:
- We could consider https://github.com/christian-korneck/update-container-description-action to try and upload the GitHub readme to DockerHub. This doesn't work with DockerHub access tokens (which isn't ideal), so if we want to use this then you will also have to add the password for the DockerHub account to GitHub secrets as `DOCKERHUB_PASSWORD`. Additionally our readme contains images and github badges which likely wouldn't come through on the docker repo.
- Manually copy the GitHub readme (minus images/badges) in DockerHub repo. This seems like a potential hassle to remember if we change/update the Github readme in the future.
- We have a simple README in DockerHub that describes basic actions and refers the user to the relevant section in the GitHub readme or docs. This is my personal preference.

Then in terms of where you want me to put the info on pulling the image/mounting/running commands:
- Is there a certain section of the docs I should put it in or do I need to create a new page?
- Do you want it mentioned in the GitHub README (maybe just a mention and link to relevant doc section)?

.github/workflows/publish-sqlfluff-docker-image-to-dockerhub.yaml
```diff
@@ -0,0 +1,65 @@
...
+      # Get the version of latest release in
+      # order to tag published Docker image.
+      - name: Get latest release name
+        id: latest_release
+        uses: pozetroninc/github-action-get-latest-release@master
+        with:
+          repository: ${{ github.repository }}
+
+      # Setup.
+      - name: Set up QEMU
```
barrywhart, line 25: Why is this needed? Perhaps for ARM support? Worth mentioning the reason in a comment?

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY README.md .
+
+# Install SQLFluff in virtual environment to
+# avoid potential clashes with system dependencies.
+RUN python -m venv .venv \
+    && . .venv/bin/activate \
+    && pip install --upgrade pip setuptools wheel \
+    && pip install .
+
+# Create unpriveleged user.
```
barrywhart, line 18:
```suggestion
# Create unprivileged user.
```

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+USER 5000
```
barrywhart, line 19: What does this do? It doesn't appear to create a user if I understand correctly. Isn't it switching to a pre-existing user with UD 5000?

.github/workflows/publish-sqlfluff-docker-image-to-dockerhub.yaml
jpy-git: Exactly QEMU is the virtualisation software required for building images for multiple platforms. So in our case allows us to build both amd and arm via buildx. I'll add a comment

Dockerfile
jpy-git: So by default the docker containter runs as root which is convenient for development, however it's recommended to run as a non-root user for security (not really a concern for us but still good to follow best practices).

uid 5000 is in the range for normal users in debian, I just picked that number as that was what was used in the original issue so I assume that's a convention. You can see below that the container is successfully running as uid 5000:
image
Given the use case of this container there's no point creating a username etc and assigning it to uid 5000.

I'll change the the comment to say "Switch to non-root user" for clarity

barrywhart: Cool. I didn't know this was possible. It's weird, but it's Linux, and if it works, that's good enough for me.

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    && . .venv/bin/activate \
+    && pip install --upgrade pip setuptools wheel \
+    && pip install .
+
+# Switch to non-root user.
+USER 5000
+
+# Set SQLFluff command as entry point for image.
+ENTRYPOINT ["/app/.venv/bin/sqlfluff"]
+CMD ["--help"]
```
ccurro, line 23:
```suggestion
ENV VIRTUAL_ENV .venv
RUN python -m venv $VIRTUAL_ENV
ENV PATH $VIRTUAL_ENV/bin:$PATH
RUN pip install --upgrade pip setuptools wheel
RUN pip install .

# Switch to non-root user.
USER 5000

# Set SQLFluff command as entry point for image.
ENTRYPOINT ["sqlfluff"]
CMD ["--help"]
```

By activating the virtual environment through `ENV` statements, you should be able to avoid the fully qualified pathname in `ENTRYPOINT` and can split up the `pip` commands over multiple `RUN` statements. 

Further, you could push the virtual environment creation up towards the top of the Dockerfile (above the `COPY` statements), along with `pip install --upgrade` command which would better take advantage of docker layer caching. Were someone doing incremental development with the docker images, this would save some build time. Though, I admit, I don't think it would be terribly substantial time savings.

Feel free to keep the way as is, if intended/preferred of course.

ccurro: Following up on this further - I think there's potential for substantial build time savings. You should be able to cache the dependencies between builds if none of the application code in `src` has changed. 

It would look something like this:

```Dockerfile
WORKDIR /app

ENV VIRTUAL_ENV .venv
RUN python -m venv $VIRTUAL_ENV
ENV PATH $VIRTUAL_ENV/bin:$PATH
RUN pip install --upgrade pip setuptools wheel

COPY setup.py .

RUN python setup.py egg_info
RUN pip install -r *.egg-info/requires.txt

COPY src ./src
COPY MANIFEST.in .
COPY README.md .

# continue with rest of file
```

The gotcha here is that it looks like `setup.py` depends on `src`...so it may take extra effort to realize this correctly. And maybe that makes it not worth the time savings :shrug: 

I'll note that GitHub Actions supports docker layer caching, so it can save time there as well if configured correctly. Maybe that would (over time) have a meaningful impact on total free minutes used on GitHub Actions. I'm new to the project so I don't know how concerned the group is about that sort of stuff.

See: https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md

ccurro: See https://github.com/jpy-git/sqlfluff/pull/1 for a functional version of `Dockerfile` akin to what I'm suggesting

jpy-git: @ccurro amazing thanks for the input! 😄 Will have a play around with the alternate venv activation, and agree that it makes sense to have the initial venv setup before copying over package.

jpy-git: On the second point, for context this image will be built on release of a new sqlfluff version, which is roughly every 1-2 weeks.
These new releases will always be for enhancements/new features/bugfixes, all of which mean changes to code in the src folder so unfortunately the method shown above for caching dependencies won't work.
However, we do have a requirements.txt file that I can use in the dependencies step to include them in cached layers. I think the ideal solution for this is to move from setup.py to setup.cfg in the future as we can then easily extract the dependency metadata from there as a single source of truth.
I'll take a look at the GHA caching mentioned as well. When I was testing the GHA build was only taking ~4 mins and as I said before will only be run roughly weekly so not a huge need to increase speed/complexity.

jpy-git: @ccurro I've made the changes following the discussion above:
- Virtual environment is created at the start and prepended to the path.
- Install requirements separately to take advantage of improved layer caching. (Currently from requirements.txt but I am going to look at adding setup.cfg to extract requirements from)
- Incorporate inline caching into GitHub action workflow. first build is still 4 mins (since there's no cache), but then subsequent builds (with no change to requirements) take 1 min.

Think the image is looking very nice now, thanks 😄

jpy-git:
@barrywhart @tunetheweb any preferences on the documentation point?

requirements.txt
```diff
@@ -1,18 +1,28 @@
+# Core
```
jpy-git, line 1: @barrywhart @barrywhart I went to use `requirements.txt` to take advantage of layer caching but found it was missing `typing_extensions`. I've copied the deps from `setup.py`.
Another example of the requirements getting out of sync so will pull together a PR to move `setup.py -> setup.cfg` and remove `requirements.txt`.

alanmcruickshank:
@jpy-git - account is up and running and secrets are in the repository as described. 😄 

https://hub.docker.com/u/sqlfluff

jpy-git:
@alanmcruickshank amazing thank you!

Dockerfile
ccurro: Looks great to me! :+1:

barrywhart:
I prefer a simple DockerHub README, perhaps linking to these SQLFluff doc sections:
* [CLI reference](https://docs.sqlfluff.com/en/stable/cli.html)
* [Configuration](https://docs.sqlfluff.com/en/stable/configuration.html)

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Create virtual environment.
+ENV VIRTUAL_ENV .venv
+RUN python -m venv $VIRTUAL_ENV
+ENV PATH $VIRTUAL_ENV/bin:$PATH
+RUN pip install --upgrade pip setuptools wheel
+
+# Install requirements seperately
+# to take advantage of layer caching.
+COPY requirements.txt .
+RUN pip install --upgrade -r requirements.txt
```
jpy-git, line 15: Once #1960 is resolved I can replace this line with:

```Dockerfile
COPY setup.cfg .
RUN pip install --upgrade -r <(python -c "import configparser; c = configparser.ConfigParser(); c.read('setup.cfg'); print(c['options']['install_requires'])")
```

jpy-git: This will ensure that we install packages from a single source of truth and is consistent with if we were to install the package directly (but with added benefit of layer caching).

barrywhart (approved):
It looks good to me. Do we need @alanmcruickshank to take a look and/or set up some things so new releases will automatically publish an image?

jpy-git:
@barrywhart: So @alanmcruickshank has already created an official sqlfluff user for us and provided the necessary credentials in Github secrets for us to start uploading the images via the release workflow. 

I'm just pulling together a quick draft of the docker README that @alanmcruickshank will need to add to the docker repo and then we should be good to go. Will let you know when that's ready 🚀

jpy-git:
# Official SQLFluff Docker Image

## The SQL Linter for Humans

**SQLFluff** is a dialect-flexible and configurable SQL linter. Designed with ELT applications in mind, **SQLFluff** also works with Jinja templating and dbt. **SQLFluff** will auto-fix most linting errors, allowing you to focus your time on what matters.

This repository contains Docker images of releases in our [GitHub](https://github.com/sqlfluff/sqlfluff) repository. The images are created via GitHub Actions to ensure the latest releases are uploaded to DockerHub.

## Basic Usage

Pull the docker image via:
```console
docker pull sqlfluff/sqlfluff:<version_tag>
```

SQLFluff commands can then be run via:
```console
docker run -it --rm -v $PWD:/sql sqlfluff/sqlfluff:<version_tag> lint test.sql
```
(lint sub-command can be exchanged with fix/parse/etc as required)

For convenience it is recommended to create the following alias in your shell configuration:
```console
alias sqlfluff='docker run -it --rm -v $PWD:/sql sqlfluff/sqlfluff:latest'
```
Commands can then be run in much the same way as the standard Python CLI:
```console
sqlfluff lint test.sql
```
 
# Additional information

For further information on SQLFluff please refer to our official documentation:

CLI Reference: https://docs.sqlfluff.com/en/stable/cli.html

Configuration: https://docs.sqlfluff.com/en/stable/configuration.html

jpy-git:
^The proposed DockerHub README.md

For the short description I suggest `Official SQLFluff Docker Image`

jpy-git:
@alanmcruickshank in order to be able to add the README and short description the repo has to be created by uploading an image first.

I've just tried temporarily changing this PR workflow to build on pull request so we can have 0.8.2 in there for you to add the descriptions.
image

but got a "username and password required error".
image

It works fine in my branch with my docker creds so worth checking secrets are correctly named and contain correct values:
image

tunetheweb:
Forks don't get access to the main repo's secrets.

Otherwise any cheeky hacker could just fork a repo, open a PR to `echo $SECRET` and get them.

This should not be an issue for real releases as, once it's merged to `main` the action will be run on release. I see you also allowed it to be run on `workflow_dispatch` so can run this manually after it's merged to get that first version up there.

jpy-git:
Makes sense, and yes the workflow dispatch should be a handy option to run it 👍 

In that case I think the plan is let's merge this in, run it manually via `workflow_dispatch` to upload 0.8.2 and then Alan can add the readme/short description once that's done 😄

Dockerfile
jpy-git: Implemented this

jpy-git:
@barrywhart thanks for merging 😄 When you get a chance could you manually trigger the docker build workflow?

I don't have permissions to trigger but there should be a run workflow button like below:
image

barrywhart:
Ok, I ran it just now.

barrywhart:
https://github.com/sqlfluff/sqlfluff/actions/runs/1495713512

jpy-git:
Works perfectly (https://hub.docker.com/r/sqlfluff/sqlfluff/tags) 🥳 
image

jpy-git:
@alanmcruickshank when you get a chance add could you add the short description and README mentioned earlier to the DockerHub repo (https://hub.docker.com/r/sqlfluff/sqlfluff).

Also I think there's an option to add a picture to the repo so may be worth adding one of the sqlfluff logos if possible.

Other than that I think we're good and these should start updating automatically on releases going forwards 😄

PR #1957, opened 2021-11-22
Speed up CI

pwildenhain:
See [official GitHub docs](https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows) and this [blog post](https://medium.com/ai2-blog/python-caching-in-github-actions-e9452698e98d) to follow what I've done for caching dependencies.

I think this will save us a few mins on the run time, but our test suite by itself does take a while to run. This is where [`pytest-xdist`](https://pypi.org/project/pytest-xdist/) comes in to do the heavy lifting 🏋🏻 

@tunetheweb I know you initially set this up for us, so it would be great to get your input specifically


.github/workflows/ci-tests.yml
```diff
@@ -26,13 +26,20 @@ jobs:
...
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: '3.10'
+        python-version: ${{ env.PYTHON_VERSION }}
+    - name: Cache dependencies
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.pythonLocation }}
+        key: ${{ env.PYTHON_VERSION }}-${{ runner.os }}-build-${{ hashFiles('tox.ini') }}
```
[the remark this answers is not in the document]
jpy-git: these static linting packages should really be installed in their relevant tox section, to avoid needlessly installing in the unit tests. So feel free to make that change and keep the hash as is if you want. If not we can do that later :)

.github/workflows/ci-tests.yml
```diff
@@ -91,13 +110,20 @@ jobs:
...
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: '3.10'
+        python-version: ${{ env.PYTHON_VERSION }}
+    - name: Cache dependencies
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.pythonLocation }}
+        key: ${{ env.PYTHON_VERSION }}-${{ runner.os }}-build-${{ hashFiles('setup.py') }}
```
jpy-git, line 126: This line is fine, but in line 130 remove `pip install tqdm` as it's installed as part of the package anyway.

.github/workflows/ci-tests.yml
```diff
@@ -114,13 +140,20 @@ jobs:
...
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: '3.8'
+        python-version: ${{ env.PYTHON_VERSION }}
+    - name: Cache dependencies
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.pythonLocation }}
+        key: ${{ env.PYTHON_VERSION }}-${{ runner.os }}-build-${{ matrix.dbt-version }}-${{ hashFiles('tox.ini') }}
```
jpy-git, line 156: Include `${{ hashFiles('setup.py') }}` and `${{ hashFiles('requirements_dev.txt') }}` and `${{ hashFiles('plugins/sqlfluff-templater-dbt/setup.py') }}` in hash key

.github/workflows/ci-tests.yml
```diff
@@ -143,7 +178,12 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: '3.8'
+        python-version: ${{ env.PYTHON_VERSION }}
+    - name: Cache dependencies
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.pythonLocation }}
+        key: ${{ env.PYTHON_VERSION }}-${{ runner.os }}-build-${{ hashFiles('tox.ini') }}
```
jpy-git, line 186: Include `${{ hashFiles('setup.py') }}` and `${{ hashFiles('requirements_dev.txt') }}` in hash key

.github/workflows/ci-tests.yml
```diff
@@ -176,7 +218,12 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: '3.8'
+        python-version: ${{ env.PYTHON_VERSION }}
+    - name: Cache dependencies
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.pythonLocation }}
+        key: ${{ env.PYTHON_VERSION }}-${{ runner.os }}-build-${{ matrix.dbt-version }}-${{ hashFiles('tox.ini') }}
```
jpy-git, line 226: Include `${{ hashFiles('setup.py') }}` and `${{ hashFiles('requirements_dev.txt') }}` in hash key

jpy-git (changes requested):

pwildenhain:
Looks like `python-xdist` is cutting the test time in half 🥳 since we're running with two processors instead of one. I have no idea why the plug test failed. Any insight from the team (or @alanmcruickshank who created the plug-in) would be helpful

jpy-git:
couple notes on the hash keys, we want to invalidate the caches if someone updates the requirement files so will need to include those files in the hash key :)

jpy-git:
@pwildenhain FYI we've moved the requirements from setup.py to setup.cfg in #1960 so will need to replace setup.py in the hash keys with setup.cfg.

Additionally we should consider replacing all pip installs in the workflows with `pip install --upgrade --upgrade-strategy eager <things to install>`. 

The motivation being that the current implementation will correctly invalidate the cache if requirements are updated in one of the files, however, given that we don't update these requirements often there is a risk that our cache will go stale and contain out of date packages. 

The above method will check if there is a newer package available and if not it installs from the cache, else it downloads the newer package. Think its a good compromise between caching whilst still keeping our test dependencies representative of what the average user will have 👍 

Just need to investigate how to achieve the same for the dependencies downloaded via tox.

chwiese:
[As per pip's documentation](https://pip.pypa.io/en/stable/user_guide/#resolver-changes-2020), it does currently not do a good job ensuring a consistent environment:

> This also means that, when you run a pip install command, pip only considers the packages you are installing in that command, and may break already-installed packages.

The combination of [poetry](https://python-poetry.org/) and tox might work better, as described [here](https://levelup.gitconnected.com/python-a-toxic-work-environment-tox-poetry-b8200563c474). You could then simply cache the poetry venv. [Here](https://github.com/marketplace/actions/install-poetry-action) is a Github action with an example workflow yml file that ties it together.

jpy-git:
@chwiese 
I would argue against something like changing package manager just for caching. Poetry is a bigger change that imo would make the project less accessible than sticking with pip. To be honest I think the truth is we aren't running actions frequently enough to benefit hugely from caching. 

@pwildenhain 
The big benefit of this PR is with the multi-processed pytest aspect as that is the main bottleneck. 
I'd suggest we remove the cache aspect of the PR and that can be re-evaluated at a later date. 

Additionally looking at typical runtimes the dbt tests only take <2mins before multiprocessing (remember the jobs in the matrix are run in parallel) so if that's not working then again let's remove them from this PR and look at them at a later date.

The main python unit tests take ~10 mins so this PR without the above already reduces the runtime by a massive 5 mins so let's focus down this PR and get it moving 💪

pwildenhain:
Makes a lot of sense @jpy-git -- caching saves seconds while doubling processors saves minutes. ill make those changes early next week.

FWIW I like @chwiese idea about poetry -- I use it a lot for personal development, but it is a barrier to entry for contributing since it's a 3rd party tool, where  as tox is really just for CI. As the project grows in complexity that trade off might become worth it

pwildenhain:
Ok I think I'm tapped out here 🏳️ I have no idea why adding more testing processors would mess with the plugin tests.

I was able to recreate the error locally, and confirm that it's the parallel processing which is the issue, but just have no clue ❓ where to go from here. The error is only for one of the tests, which makes it even weirder ☹️ It fails on `plugins/sqlfluff-templater-dbt/test/templater_test.py::test__templater_dbt_handle_exceptions[exception_connect_database.sql-dbt tried to connect to the database]` but passes everything else

I wonder if our best option is to try and not do parallel processing for the `dbt` tests -- though I'm not 💯 sure how to do that . I know you can use [`-o`](https://docs.pytest.org/en/6.2.x/reference.html#command-line-flags) to override the settings in a `.ini` file for `pytest`, but I'm not sure how to do that with `tox`.

tox.ini
```diff
@@ -41,7 +41,7 @@ commands =
     # Clean up from previous tests
     python util.py clean-tests
     # Run tests
-    pytest -vv -rs --cov=sqlfluff --cov-report=xml {posargs: test}
+    pytest -n 2 -vv -rs --cov=sqlfluff --cov-report=xml {posargs: test}
```
jpy-git, line 44: @pwildenhain we have posargs on the tox tests so lets set this line back to no multitasking by default
```suggestion
    pytest -vv -rs --cov=sqlfluff --cov-report=xml {posargs: test}
```

and then in `.github/workflows/ci-tests.yml` line 64 (the tox tests for the standard python unit tests) pass `-n 2` via posargs:
i.e.
```
tox -e ${{ env.TOXENV }} -- -n 2
```

This will use the multitasking on the python tests only and leave the dbt ones as they were before, which should hopefully let us get the speed boost on the main tests without impacting the dbt ones 😄

jpy-git: FYI posargs (the `--`) works by replacing the contents of `{posargs: test}` (where `test` is the default value for no posargs).
So in this case its like running `pytest -vv -rs --cov=sqlfluff --cov-report=xml -n 2 test` for the standard tests
and `pytest -vv -rs --cov=sqlfluff --cov-report=xml plugins/sqlfluff-templater-dbt` for the dbt tests 👍

pwildenhain:
@jpy-git once you join as a maintainer you should be able to review and merge 👍🏻

jpy-git (approved):

jpy-git:
@pwildenhain thanks! 😄  I'll merge it in now!

PR #1965, opened 2021-11-23
TSQL: Spaces allowed in comparison operators

### LINKED ISSUE #1767: TSQL: comparison operators can be split by whitespace
For example:
```
	SELECT 
		CASE 
			WHEN 1 >			= 0 THEN '1'
			ELSE '0'
		END
```

## Expected Behaviour

Code snippet parses.

## Observed Behaviour

Code snippet does not parse.

## Dialect

TSQL

jpers36:
Fixes #1767

TSQL, for whatever reason, allows multisymbol comparison operators to be separated by spaces.  This change reflects that functionality.

+NotSegment to support splitting multisymbol operators
+Sequences within ComparisonOperatorGrammar to represent the various multisymbol operators
+Test cases within select.sql
Updated YML for select.sql


test/fixtures/dialects/tsql/select.yml
```diff
@@ -47,6 +47,15 @@ file:
               - expression:
                   literal: "'True'"
               - keyword: WHEN
+              - expression:
+                - literal: '1'
+                - comparison_operator: '>'
+                - comparison_operator: '='
```
jpy-git, line 53: can see that these are parsed as two separate identifiers rather than a single `>=`. Might be worth adding some test cases to any rules that could potentially rely on logic that looks at the comparison operator (thinking L043).

jpy-git (commented):
Just a note as this review is probably better suited to one of the grammar experts @WittierDinosaur @tunetheweb .

Also, def worth looking at adding a rule in the future to remove this whitespace if you're interested. Should be fairly simple to detect consecutive comparison operators, replace the first comparison operator segment with the combined operator and delete the other segments.

test/fixtures/dialects/tsql/select.yml
jpy-git: L007 is also one to check for this

barrywhart: Also maybe L006 and L049

barrywhart (approved):
Looks good to me. Would you like to look at the additional tests @jpy-git suggested prior to merging this?

jpers36:
@barrywhart Yes, I think it would be good to look at the other tests.  It might not be until Monday, though, what with the holiday.

barrywhart:
That's fine. I'm resting this week, too. 🦃

test/fixtures/dialects/tsql/select.yml
tkachenkomaria244: @jpy-git , @barrywhart I wonder if such an issue was created (to add a rule to fix such cases with "> =") as we faced with same trouble in our repo?

jpy-git: @tkachenkomaria244 no rule has been created yet

tkachenkomaria244: @jpy-git Josef, I've submitted an issue then: [#1997](https://github.com/sqlfluff/sqlfluff/issues/1997)

jpy-git: perfect thanks!

Issue #1812, opened 2021-11-03
Should all rules be enabled by default?


tunetheweb:
Some thoughts from [another thread](https://github.com/sqlfluff/sqlfluff/issues/1687#issuecomment-946158751), that I think are worth pulling into it's own issue for discussion:

> I'm in two minds about disabling rules by default.
> 
> On one hand, there's some rules I disagree with (this being one, but also L028, L031, L032, L034, L035, L037, and L043) and think I'm probably not alone with some of those, and usually defaults should be the sensible options used by most.
> 
> On the other hand, having them all on by default, makes you look at a rule, evaluate it, and then decide if it's one for you or not - that is, it's an explicit decision to turn it off. I think that would happen less if we turned them off by default - i.e. people wouldn't evaluate whether to turn it on, and instead would ignore the non-default rules unless it was something they felt strongly about and wanted to enforce. And maybe that's OK? But maybe that's also a missed opportunity to look at these rules and see if you should be using them? L031 makes a lot of sense for small databases. Just not convinced it scales well.
> 
> And the other issue is, that the impression I get, is that some of the users don't realise it's OK to turn off rules. Most linters (flake8/black being the exception!) need customising to your preferences.

pwildenhain:
Could we potentially get the best of both worlds here? Where we have only the most stable, most common sense rules enabled by default, but provide some sort of flag that allows users to run *all* the rules e.g. `--run-all-rules` so they could see for themselves the difference between the "default" and "all" and decide which rules they want to whitelist.

I know we've also discussed in past issues (#201) on having severity levels for our lints. These conversations feel like they are circling around the same concept/problem

WittierDinosaur:
I think while we're in Beta we can still change this kind of behaviour, as long as it's announced and in the release notes. Also, separately, we are at 50/ 51 rules now. While we're in Beta, maybe we should start grouping them. eg. C01, C02, for core, D01,D02 for dialect dependant, T01, T02, for templaters, etc

WittierDinosaur:
As long as we don't use the LXX namespace, we can also keep back-compatible aliases for them

jpy-git:
I like @WittierDinosaur's idea of starting to group rules by the rule name prefix (C01, T03, etc.). 

I think if we were to do that then we could adjust the include/exclude rule options in the CLI/config to accept globs (or regex):
e.g.
```console
sqlfluff lint test.sql --rules C*
``` 
to just include core rules
or
```console
[...]

Issue #1938, opened 2021-11-19
How to disable a rule via the config?


yb66:
## URL or Command Line Option with the issue being reported

The documentation does not include something like "How to disable a rule". There is [Ignoring line ranges](https://docs.sqlfluff.com/en/stable/rules.html#ignoring-line-ranges) and within the configuration of some rules there are things like `force_enable` but nothing about *how to do that in the config*. Or at least, nothing *obvious* (I'm a new user, if things are to be found then they need to be obvious).

I've done site searches of the docs and scoured them but not found this.

Is it `exclude_rules`? I found it while writing this. There's a line hidden in [CLI Arguments](https://docs.sqlfluff.com/en/stable/configuration.html#cli-arguments). I guess it's that.

## Issue/Suggested Improvement

Give at least one example, preferably in a section with an appropriate title, of how to disable a rule *within the .sqlfluff config*, and make it *very clear*. For all I know, one of the lines in the [default config](https://docs.sqlfluff.com/en/stable/configuration.html#default-configuration) is disabling a rule but how would I know? It's almost a hundred lines long and the only candidate has a "None" next to it.

Regards,
iain

jpy-git:
@yb66 this already exists in the configuration section of the documentation:
![image](https://user-images.githubusercontent.com/80432516/142600111-a40a420d-3a5c-4167-ab84-f1f25728812f.png)

I think the `Configuration` page is a sensible place to put it. That section does quite clearly state that the config file is an alternative place to put command line options, so the place to look would be the `CLI Reference` page or just do `sqlfluff lint --help` if you want to understand what options you have?

On the default config point, I think `exclude_rules=None` is pretty clear that no rules are disabled by default.

No harm in creating a PR if you think you have a clearer layout though?

tunetheweb:
I think a link to that section from the rules page would help make this clearer:

Somewhere around this section: https://docs.sqlfluff.com/en/stable/rules.html#inline-ignoring-errors

jpy-git:
@tunetheweb ahh, those `Inline Ignoring Errors` and `Ignoring line ranges` are duplicated in the configuration page. 
I can see why this could be confusing to someone looking at the rules page as put those sections there implies you would expect the rest of the config info there.
I would propose remove those two sections from the rules page and then replace it with the link to the `CLI Arguments` section in `Configuration`. That way the user has one clear place to look for config info.

tunetheweb:
Hmmm.... I'm in two minds.

[...]

PR #1972, opened 2021-11-24
Rule globs

### LINKED ISSUE #1445: fix CLI - exclude rules by pattern
_Note: This is for general enhancements to the project. Please use the Bug report template instead to raise parsing/linting/syntax issues for existing supported dialects_

Hi there,
I setup a custom rules plugin and it works as expected with `lint`. But if I try to run `fix` (both with pre-commit), sqlfluff stuck in a loop (maybe tries to fix the custom rule). 

It would be nice if we could exclude rules by pattern or maybe by plugin or exclude all custom rules.

E.g.

```
sqlfluff fix --exclude-rules=MYPKG_*
# or
sqlfluff fix --exclude-rules-package=MYPKG
# or
sqlfluff fix --exclude-custom-rules
```

BR
Timo

jpy-git:

Fixes #1445. This PR provides the mechanism for rules to be included/excluded by globs in either the config files or via the CLI.
image

As well as resolving the linked issue, the big benefit of this will be to allow users to more easily bulk disable groups of rules 
via the rule prefix, as I discussed in #1812. This is essentially what is happening in the #1445 as well just for plugins. Personally think this is a very elegant solution. Also this use case is what motivates the use of globs rather than regex as the expected use is simply `--rules "Core_*"` or `--exclude-rules "Some_Plugin_*"` (or whatever prefixes we decide on), so keeps things nice and simple.

No

### Pull Request checklist

  - Other.

### DIFF
--- a/examples/01_basic_api_usage.py
+++ b/examples/01_basic_api_usage.py
@@ -22,7 +22,7 @@ result = sqlfluff.fix(my_bad_query, dialect="bigquery")
 # result = 'SELECT  *, 1, blah AS  foo  FROM mytable\n'
 
 # We can also fix just specific rules.
-result = sqlfluff.fix(my_bad_query, rules="L010")
+result = sqlfluff.fix(my_bad_query, rules=["L010"])
 # result = 'SELECT  *, 1, blah AS  fOO  FROM myTable'
 
 # Or a subset of rules...
--- a/setup.cfg
+++ b/setup.cfg
@@ -107,6 +107,7 @@ sqlfluff =
 sqlfluff =
     config.ini
     core/default_config.cfg
+    py.typed
 
 [sqlfluff_docs]
 stable_version = 0.8.2
--- a/src/sqlfluff/__init__.py
+++ b/src/sqlfluff/__init__.py
@@ -3,13 +3,21 @@ import sys
 import pytest
 
 # Expose the public API.
-from sqlfluff.api import lint, fix, parse, list_rules, list_dialects  # noqa: F401
+from sqlfluff.api import lint, fix, parse, list_rules, list_dialects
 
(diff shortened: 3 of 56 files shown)


tunetheweb (commented):
I've a few questions:
1. Does this allow `exclude_rules = L001, L05*` and if so should we add a test?
2. Should we update our documentation for this?
3. Should we take this as an opportunity to move to the more inclusive "allowlist/denylist" syntax? At least for the code touched here?

jpy-git:
1. Yes it does, will add that unit test in!
2. Sounds like a plan. I think it makes sense to address this as part of what we discussed in #1938. So let's make it a separate PR and address both at the same time.
3. Agree that would be a nice change 😄 Let me assess how much much the terms are used and then can replace them all at once (may be worth separate PR if there are loads)

An additional question I'm wondering myself is if this glob stuff also applies to inline comments. I think it does but let's resolve #1985 first and then I'll investigate 👍

barrywhart (approved):

barrywhart:
Looks like a dbt test failed in one of the builds?

jpy-git:
@barrywhart it didn't fail, the GHA runner lost internet connection so the install download failed. Should work fine once re-run 👍

jpy-git:
@tunetheweb Follow up on your questions
- 1. I've added the extra unit test
- 2. will be separate pr
- 3. I've had a look and will make a separate PR for this otherwise the diffs will look a mess 👍 
- 4. These globs don't apply to inline comments at the moment because the inline comment code has no interaction with the rule set and just checks if one of the supplied rules matches an error raised. Will make a follow up PR to extend this 👍 

LGTM 🚀

tunetheweb (approved):

PR #1976, opened 2021-11-25
Support __init__.py for library_path

Tonkonozhenko:
In our SQL files, we have jinja with non-primitive data types e.g. datatime. This PR allows setting such variable not in [sqlfluff:templater:jinja:context] section but in `__init__.py` file

I believe it should not cause any side effects.

### Pull Request checklist

  - Other.


jpy-git:
@Tonkonozhenko please can you include a description/motivation for the change in the pr description?

barrywhart:
Are there existing docs we could update to explain this new behavior?

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -116,7 +116,10 @@ def _extract_libraries_from_config(self, config):
             spec = importlib.util.spec_from_file_location(module_name, file_path)
             lib = importlib.util.module_from_spec(spec)
             spec.loader.exec_module(lib)
-            libraries[module_name] = lib
+            if module_name == '__init__':
```
jpy-git, line 119: Add a comment here describing what you're doing

jpy-git:
Agree with @barrywhart 

I think you need to include a small example/description of how this is used in the docs, so that others are able to understand and take advantage of the feature 👍

barrywhart:
This PR needs a test, also (code coverage).

Tonkonozhenko:
@jpy-git @barrywhart added tests and docs

jpy-git:
awesome thanks I can see what you're trying to add. I have a couple thoughts on further improvements for this so let me pull together some stuff for discussion this evening 👍

docs/source/configuration.rst
```diff
@@ -374,6 +374,19 @@ to use them for templated. In the above example, you might define a file at
         return "GROUP BY 1,2"
 
 
+If you need root-level functions, you can defined them in __init__.py.
```
barrywhart, line 377:
```suggestion
If you need root-level functions, you can define them in __init__.py.
```

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -116,7 +116,13 @@ def _extract_libraries_from_config(self, config):
             spec = importlib.util.spec_from_file_location(module_name, file_path)
             lib = importlib.util.module_from_spec(spec)
             spec.loader.exec_module(lib)
-            libraries[module_name] = lib
+            if module_name == "__init__":
+                # Allows settings root-level functions via __init__.py
```
barrywhart, line 120:
```suggestion
                # Allows setting root-level functions via __init__.py
```

barrywhart (commented):
Looks good to me, other than a couple of small typos. I'll delay approving so @jpy-git has a chance to review.

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -116,7 +116,13 @@ def _extract_libraries_from_config(self, config):
             spec = importlib.util.spec_from_file_location(module_name, file_path)
             lib = importlib.util.module_from_spec(spec)
             spec.loader.exec_module(lib)
-            libraries[module_name] = lib
+            if module_name == "__init__":
```
jpy-git, line 119: So my main thoughts around this PR were around the use of `__init__.py` to store these variables. `__init__.py` has special meaning in Python as in general it indicates the folder to be treated as a package.

So seeing a folder structure like
image
sort of implies to the user that when they specify in their config
```
[sqlfluff:templater:jinja]
library_path=libs
```
they are importing a package called libs
i.e.
```sql
SELECT 56
FROM {{ libs.foo.schema }}.{{ libs.foo.table("xyz") }}
WHERE {{ libs.bar.equals("x", 23) }} and {{ libs.root_equals("y", 42) }}
```
rather than a series of individual python modules within the libs folder as we are currently doing.

jpy-git: Implying package like structure also then implies the inclusion of submodules as well:
e.g.
image
suggests this is possible:
```sql
SELECT 56
FROM {{ libs.submodule.baz.table("xyz") }}
```
which we don't support.

Tonkonozhenko: We already have root-level functions and variables and I believe it's the best place where we can put such macros and variables

Tonkonozhenko: Won't nested modules be supported automatically?

jpy-git: So I gave some thought to redefining the library to expect a package to expand on the thoughts above.

code would be something like:
```python
def _extract_libraries_from_config(self, config):
        
        library_path = config.get_section(
            (self.templater_selector, self.name, "library_path")
        )
        if not library_path:
            return {}

        libraries = {}
        # We need to determine if the provided library is a folder of
        # python modules or a package (i.e. contains an __init__.py).
        # If it is a package then we can import the library as such,
        # else we import the individual modules.
        package_name = library_path.split("/")[-1]
        init_path = os.path.join(library_path, "__init__.py")
        
        if os.path.exists(init_path):
            # Library is a package.
            spec = importlib.util.spec_from_file_location(package_name, init_path)
            module = importlib.util.module_from_spec(spec)
            spec.loader.exec_module(module)
            libraries[package_name] = module

            # Recursively walk submodules and add them to libraries.
            for loader, module_name, _ in pkgutil.walk_packages(
                module.__path__, prefix=f"{package_name}."
            ):
                libraries[module_name] = loader.find_module(module_name).load_module(
                    module_name
                )

        return libraries
```
i.e. import the libs module and then recursively import all of it's modules and submodules.
- Namespace would look something like `{"libs": ..., "libs.foo": ..., "libs.bar": ..., "libs.submodule": ..., "libs.submodule.baz": ...}` 
- However if you try and reference nested modules in Jinja you can't do it. Reason being in `{{ libs.foo.some_func() }}` Jinja tries to search for the attribute `foo` in `libs` whic isn't there, rather than the library path equivalent in Python of doing `import libs.foo; libs.foo.some_func()`. So basically it's not possible and Jinja doesn't support package like libraries (nested or unnested). So we stick with the importing current importing individual modules method (of which `__init__.py` is just considered a standard module which we just apply some unique logic to).

jpy-git: > Won't nested modules be supported automatically?

No it doesn't, if you read the lines around your change it:
- lists the content for the files in the supplied directory path
- if the files is a `.py` file load as a standalone module (or if it's a `__init__.py` load the non-private attributes within the module`)
- ignore directories. therefore any subfolders/modules are not included and I don't think you can supply multiple paths to `library_path` in the config.

jpy-git: @Tonkonozhenko @barrywhart

So basically in principle I agree with the change, as from my tests above I've verified that a package based approach isn't possible for Jinja.

My main question is:
Are we happy with using `__init__.py` for this custom logic given the potential for user confusion described above. Or could we consider a custom name like `__libroot__.py` to be more clear that this is non-standard logic?

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -116,7 +116,13 @@ def _extract_libraries_from_config(self, config):
...
+                # Allows setting root-level functions via __init__.py
```
jpy-git, line 120:
```suggestion
                # Allows setting root-level functions via library __init__.py
```
__init__.py is common so be clear which one we're talking about

docs/source/configuration.rst
```diff
@@ -374,6 +374,19 @@ to use them for templated. In the above example, you might define a file at
         return "GROUP BY 1,2"
 
 
+If you need root-level functions, you can define them in __init__.py.
```
jpy-git, line 377:
```suggestion
If you need library root-level functions, you can define them in the library's __init__.py.
```
Again worth being specific here as these terms are common in the scope of a larger project that could have multiple packages/__init__s and as shown in the test case the function might be root-level for the library but not necessarily in the root directory of the repository.

docs/source/configuration.rst
```diff
@@ -374,6 +374,19 @@ to use them for templated. In the above example, you might define a file at
...
+
+    SELECT {{ custom_sum('foo', 'bar') }} FROM baz
+
+`sqlfluff_libs/__init__.py`:
+
+.. code-block:: python
+
+    def custom_sum(a: str, b: str) -> str:
+        return a + b
+
```
jpy-git, line 389: There's a bit in this section of the docs that says
> This will pull in any python modules from that directory and allow sqlfluff to use them for templated.

Maybe just add a slight caveat to reflect what we discussed about this not importing modules from subfolders of the supplied path:
e.g.
 > This will pull in any python modules from that directory (excluding subdirectories) and allow sqlfluff to use them in templates.

Think even if we stick with `__init__.py` as our module for root functions these changes should at least make the behaviour clear to people reading the docs 👍

src/sqlfluff/core/templaters/jinja.py
Tonkonozhenko: @jpy-git WDYT about such an approach for nested modules?
```python
        libraries = {}

        for loader, module_name, is_pkg in pkgutil.walk_packages([library_path]):
            _module = loader.find_module(module_name).load_module(module_name)

            if '.' in module_name:
                *modules, last = module_name.split('.')
                module = reduce(lambda res, path_part: res[path_part], modules, libraries)
                setattr(module, last, _module)
            else:
                libraries[module_name] = _module

        return libraries
```

Tonkonozhenko:
@jpy-git @barrywhart 

the final code to support root-level functions and nested modules in jinja:
```python
class JinjaTemplater(PythonTemplater):
    class Libraries:
        pass

        ...

    def _extract_libraries_from_config(self, config):
        library_path = config.get_section(
            (self.templater_selector, self.name, "library_path")
        )
        if not library_path:
            return {}

        libraries = JinjaTemplater.Libraries()

        is_lib_module = os.path.exists(f'{library_path}/__init__.py')
        walk_path = f'{library_path}/..' if is_lib_module else library_path

        for loader, module_name, is_pkg in pkgutil.walk_packages([walk_path]):
            _module = loader.find_module(module_name).load_module(module_name)

            if '.' in module_name:
                *modules, last = module_name.split('.')
                module = reduce(lambda res, path_part: getattr(res, path_part), modules, libraries)
                setattr(module, last, _module)
            else:
                setattr(libraries, module_name, _module)

        if is_lib_module:
            libraries = getattr(libraries, os.path.basename(library_path))

        return {k: v for k, v in libraries.__dict__.items() if not k.startswith("_")}
```

jpy-git:
Nice, let me have a quick play around with this 👍

jpy-git:
@Tonkonozhenko Nice work, I think the implementation above of mocking the package namespace is looking pretty nice 👍 Lets get it added to the PR so we can polish and get things moving 🚀 

Things to consider whilst adding this in:
- small docstring on `Libraries`. Something like `Mock namespace for user-defined Jinja library`
- use `os.path.join` for the paths rather than f strings. reason for best practice:
image
- A few small descriptive comments in `_extract_libraries_from_config`. Its a very metaprogramming approach so will be useful to have the commentary for future developers
- In terms of tests we currently just have the one in `test/fixtures/templater/jinja_j_libraries` that tests the package approach. Let's add a submodule to this test as well to validate it works as expected. Also let's add another libraries folder to validate the code when no `__init__.py` is present. Have ad-hoc verified these but good to get them in the unit tests 😄 
- Update the docs section to describe the logic new logic. If `__init__.py` detected the library modules are loaded including any submodules.

Anything else we can tidy up once it's in GitHub, but this is shaping up nicely 💪

docs/source/configuration.rst
```diff
@@ -374,6 +374,33 @@ to use them for templated. In the above example, you might define a file at
         return "GROUP BY 1,2"
 
 
+If __init__.py is detected in the library path, modules are loaded including
+any submodules, else individual .py modules are loaded.
+
+.. code-block:: jinja
+
+    SELECT {{ custom_sum('foo', 'bar') }}, {{ foo.bar.another_sum('foo', 'bar') }} FROM baz
```
jpy-git, line 382:
```suggestion
    SELECT
        {{ custom_sum('foo', 'bar') }},
        {{ foo.bar.another_sum('foo', 'bar') }}
    FROM baz
```
Let's split this onto multiple lines as it looks better in the docs

jpy-git (commented):
Couple small doc tweaks and linting issues (below) to address then we're good to go
image

Tonkonozhenko:
@jpy-git fixed the comments

jpy-git (approved):
@Tonkonozhenko well done for working through all these changes ⭐ LGTM 🚀

Tonkonozhenko:
Thanks for the help @jpy-git @barrywhart

PR #1979, opened 2021-11-26
Fix L009 FileSegment child + new create_before/create_after edit types

### LINKED ISSUE #1978: Nuisance: L052 (require_final_semicolon = True) does not reliably add `;` if L009 is on and EOF line does not exist already
If:
- Rule 009 is not excluded (default)
- There is no single trailing newline (as rule 009 requires)
- Rule 052 parameter `require_final_semicolon = True` (not default)

Then `sqlfluff fix` adds newline, but does not add a semicolon (`;`).

Related experiments:
- Value of another 052's parameter `semicolon_newline = True/False` does not affect the mechanics
- While ```[sqlfluff] exclude_rules = L009``` makes `;` appear always as expected

## Query
```
SELECT
    a
FROM foo
```

## Expected Behaviour
```
SELECT a
FROM foo;

```

## Observed Behaviour
```
SELECT a,
FROM foo

```

## Steps to Reproduce
I believe any setup will have such error if 
```
[sqlfluff:rules:L052] 
```
added to the `.sqlfluff`

## Dialect
- bigquery

## Version
- sqlfluff, version 0.8.2

## Configuration
```
[sqlfluff]
templater = jinja
dialect = bigquery
# exclude_rules = L009

[sqlfluff:rules:L052] 
```

jpy-git:

This PR fixes #1978. L009 and L052 were both being applied as edits on the final segment of the file, causing L052 to not be applied. This is similar to the behaviour described by @barrywhart in #1808.

The original reasoning for this approach, even though it leads to an incorrect parse tree, was because there wasn't a better way to discover where these segments should be placed , however this can actually be easily fixed for L009

Observing an existing final new line parse (for all dialects), we see that it is a child of the file segment, hence we can carry out the edit on the child of the file segment when fixing L009.

image

fixing L052 to better place the final semicolon in the parse tree is more of a challenge (reasoning discussed below) but this fix corrects the issue at hand, as well as L009 in general.

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -438,6 +438,32 @@ class BaseRule:
 
     # HELPER METHODS --------
 
+    def is_final_segment(self, context: RuleContext) -> bool:
+        """Is the current segment the final segment in the parse tree."""
+        if len(self.filter_meta(context.siblings_post)) > 0:
+            # This can only fail on the last segment
+            return False
+        elif len(context.segment.segments) > 0:
+            # This can only fail on the last base segment
+            return False
+        elif context.segment.is_meta:
+            # We can't fail on a meta segment
+            return False
+        else:
+            # We know we are at a leaf of the tree but not necessarily at the end of the tree.
+            # Therefore we look backwards up the parent stack and ask if any of the parent segments
+            # have another non-meta child segment after the current one.
+            child_segment = context.segment
+            for parent_segment in context.parent_stack[::-1]:
+                possible_children = [
+                    s for s in parent_segment.segments if not s.is_meta
+                ]
+                if len(possible_children) > possible_children.index(child_segment) + 1:
+                    return False
+                child_segment = parent_segment
+
+        return True
+
     @staticmethod
     def filter_meta(segments, keep_meta=False):
         """Filter the segments to non-meta.
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -113,7 +113,12 @@ class LintFix:
     """
 
     def __init__(self, edit_type, anchor: BaseSegment, edit=None):
-        if edit_type not in ["create", "edit", "delete"]:  # pragma: no cover
+        if edit_type not in (
+            "create_before",
+            "create_after",
+            "edit",
+            "delete",
+        ):  # pragma: no cover
             raise ValueError(f"Unexpected edit_type: {edit_type}")
         self.edit_type = edit_type
         if not anchor:  # pragma: no cover
@@ -153,7 +158,7 @@ class LintFix:
 
         Removing these makes the routines which process fixes much faster.
         """
-        if self.edit_type == "create":
+        if self.edit_type in ("create_before", "create_after"):
             if isinstance(self.edit, BaseSegment):
                 if len(self.edit.raw) == 0:  # pragma: no cover TODO?
                     return True
@@ -166,7 +171,7 @@ class LintFix:
     def __repr__(self):
         if self.edit_type == "delete":
             detail = f"delete:{self.anchor.raw!r}"
-        elif self.edit_type in ("edit", "create"):
+        elif self.edit_type in ("edit", "create_before", "create_after"):
             if hasattr(self.edit, "raw"):
                 new_detail = self.edit.raw  # pragma: no cover TODO?
             else:
--- a/src/sqlfluff/rules/L009.py
+++ b/src/sqlfluff/rules/L009.py
@@ -81,27 +81,9 @@ class Rule_L009(BaseRule):
         for this rule, we discard the others into the kwargs argument.
 
         """
-        if len(self.filter_meta(context.siblings_post)) > 0:
-            # This can only fail on the last segment
+        # We only care about the final segment of the parse tree.
+        if not self.is_final_segment(context):
             return None
-        elif len(context.segment.segments) > 0:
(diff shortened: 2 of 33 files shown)


jpy-git:
future todo:
L052 should be fixed in a similar manner to L009, and for non-tsql dialects the approach could be similar as shown below:

image

However the tricky thing here is that tsql handles semicolons differently and imo the semicolon grammar looks pretty messy here, making it hard to use better logic than we currently have. 
It seems to me that the appropriate place for the final semicolon to be in the tsql dialect is as a child of BatchSegment (rather than FileSegment). However looking at the parse we can see the final semicolon being part of the FromClauseSegment. 

image

maybe some work needs to be done around tidying up these semicolons on the grammar side to be consistent and then we can look at refining the rule

jpy-git:
Failing tests are coming from this condition not being sufficient to determine end of file, will try and come up with a better approach

image

jpy-git:
Yeah basically the old method in the screenshot wasn't a very reliable way to detect end of tree. I've implemented a proper algorithm that looks back up the parent stack to search for later branches. Think it's quite a nice implementation 👍

jpy-git:
The edit was a sensible idea, in terms of getting the correct parse tree, however where it ran into issues was that it spans the entire underlying source slice, which for templated files results in the fix being thrown away as unfixable. 

What we essentially need is an `edit_type` like `create` but that creates AFTER the provided segment (tbh this is something we should definitely have anyway). Therefore I have renamed `create` to `create_before` and added the new `create_after` edit type. This allows us to get the correct parse tree without touching templated raw_slices. Also will be useful for future rules.

jpy-git:
@barrywhart finally gotten to the bottom of this issue for L009 and updated the `create` edit type to be more useful as `create_before`/`create_after`, so this is ready for review now.

src/sqlfluff/rules/L009.py
```diff
@@ -91,15 +91,17 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
-            # So this looks like the end of the file, but we
-            # need to check that each parent segment is also the last.
-            # We do this with reference to the templated file, because it's
-            # the best we can do given the information available.
-            file_len = len(context.segment.pos_marker.templated_file.templated_str)
-            pos = context.segment.pos_marker.templated_slice.stop
-            # Does the length of the file equal the end of the templated position?
-            if file_len != pos:
-                return None
+            # We know we are at a leaf of the tree but not necessarily at the end of the tree.
```
barrywhart, line 94: Is this the same code that was added to L052? I wonder if this should be turned into a reusable function, e.g. on the rule context or base rule classes.

test/rules/std_L009_L052_test.py
```diff
@@ -0,0 +1,23 @@
+"""Tests the python routines within L009 and L052."""
+from sqlfluff.core import FluffConfig
+from sqlfluff.core import Linter
+
+
+def test__rules__std_L009_and_L052_interaction() -> None:
```
barrywhart, line 6: Nitpick: Could we avoid creating a custom test here by creating a test in `test/fixtures/linter/autofix/` instead? Those tests can run multiple rules.

barrywhart (commented):
Looks good overall. A few questions.

Thanks for adding "create_after". I had been thinking we needed this, but hadn't yet encountered a "must" case for adding it. 👍🏽

src/sqlfluff/rules/L009.py
jpy-git: Moved the logic to `BaseRule` and it looks much neater now 👍

test/rules/std_L009_L052_test.py
jpy-git: @barrywhart I'm struggling with this a bit. I can see that the autofix `test-config.yml` accepts multiple rules, but this test also requires a custom config setting to enable final semi-colons and there aren't any other examples of this in the autofix tests. 
tried similar to the std rule test syntax but that is unrecognised and raises an error. 
Can you let me know the syntax or if these tests even allow the custom config option?

I'll take a look in the code myself, but if it's not there then we can stick with custom test.

jpy-git: okay it's not possible to add a custom config to these tests so lets leave the custom python test for now.
image

If we want to add more detailed yaml config to these tests then that can be a separate PR 👍

barrywhart:
Thanks for looking into the custom config for those autofix tests -- I wasn't sure if that was supported or not. Sorry for sending you towards a dead end. It's may be worth creating an issue about adding that capability later.

jpy-git:
@barrywhart no worries, I'll create the issue for autofix config extension 👍

src/sqlfluff/core/rules/base.py
```diff
@@ -433,6 +438,32 @@ def _process_lint_result(res):
 
     # HELPER METHODS --------
 
+    def is_final_segment(self, context: RuleContext) -> bool:
```
barrywhart, line 441: Nice! I should look some time -- I think L003 may be able to use your function instead of the similar function [here](https://github.com/sqlfluff/sqlfluff/blob/253cfaaf058e364336234cc43f3775224de41dbd/src/sqlfluff/rules/L003.py#L235).

barrywhart (approved):
Great work!

src/sqlfluff/core/rules/base.py
jpy-git: awesome, looks like it should be a pretty straight swap! 😄

PR #1980, opened 2021-11-26
Postgres: New rule to fix deprecated syntax inside "CREATE TRIGGER"

kevinmarsh:

The `CREATE TRIGGER` [documentation in Postgres states](https://www.postgresql.org/docs/14/sql-createtrigger.html) that
> In the syntax of `CREATE TRIGGER`, the keywords `FUNCTION` and `PROCEDURE` are equivalent ... **The use of the keyword `PROCEDURE` here is historical and deprecated**.

This flags up `EXECUTE PROCEDURE` and can auto-fix to `EXECUTE FUNCTION`.

I've based a lot of the rule logic on the most recent rule (`L056`) but feedback welcome as this is my first rule contribution.

This [was _only_ changed in Postgres 11](https://www.postgresql.org/docs/release/11.0/), so this `LintError` and `LintViolation` is really only applicable to Postgres >=11. If there's a way to only target certain versions of the language or if the project is commited to support pre-Postgres 11 versions please let me know.

### Pull Request checklist

  - Other.


jpy-git:
@kevinmarsh so the logic for the rule checks out but I do question the need for a separate rule for each deprecated segment? I can see there was a brief discussion in the slack channel with @tunetheweb and @barrywhart.

My thought is:
- Add additional kwargs `is_deprecated: bool = False` and `deprecated_replace: Optional[BaseGrammar]: None` to `BaseGrammar`. (similar to how we have kwargs for `optional: bool = False`)
- you could then edit the postgres `CreateTriggerStatementSegment` execute part like:
image
- Redefine this rule as something like "Disallow deprecated segments" and the logic can then be super simple and generic:
```python
if context.segment.is_deprecated and context.segment.deprecated_replace:
    return LintResult(
        context.segment,
        fixes=[
            LintFix(
                "edit",
                anchor=context.segment,
                edit=context.segment.deprecated_replace,
            )
        ],
    )
```

Think would be very simple to implement/maintain and doesn't need to impact any parsing logic 👍

barrywhart:
+1 for making this rule generic.

jpy-git:
Refining my previous suggestion further:

- Add additional kwarg `deprecated_message: Optional[str]: None` to `BaseGrammar`. (similar to how we have kwargs for `optional: bool = False`). 
- you could then edit the postgres `CreateTriggerStatementSegment` execute part like:
image
- Redefine this rule as something like "Disallow deprecated segments" and the logic can then be super simple and generic:
```python
if context.segment.deprecated_message is not None:
    return LintResult(
        context.segment,
        description=context.segment.deprecated_message,
    )
```

So we don't provide an autofix because although possible to fix for simple replaced keyword, a deprecation could very likely impact a whole segment and it's children and we don't want to have to implement all the custom logic to re-write these. 

Instead we just flag the grammar segment as deprecated with a message describing the deprecation and suggested fix (you can also provide a link to the source of the deprecation info). Then the user can fix the deprecation if they want 👍

barrywhart (changes requested):
Marking as "request changes", following up the discussion about making the rule generic.

kevinmarsh:
I'm going close this, thanks for reviewing but I don't think I'm going to have time to think about it further this week. In principle I agree that it'd be great to have generic `Deprecated` wrappers around `Grammers` but I think it needs some more thought about 
- how autofix can be implemented
- guidance for how to ignore a specific deprecation rule (if we wanted to have something nicer than just inline ignores...)
- how to best communicate the motivation/source to end users
- how to handle differing versions of languages (it's a consideration but that's outside the scope of this particular change) 

I'll create an issue about the above bringing together some comments from above and from slack

barrywhart:
I don't think autofix is necessary for handling deprecation. It's a much more complex issue, because if something is deprecated, switching ot the alternative may be a bit complex, there may be a number of alternatives, etc. etc. More of a "nice to have", perhaps for certain simple cases.

jpy-git:
Agree with @barrywhart , think there's merit in not trying to over-engineer this.

- Don't autofix, as it only applies to simplest case.
- Guide/communicate via the message raised by the rule.
- The dialects in SQLFluff aren't exact matches to the true dialect, they're added to as we go. Trying to maintain every version of every dialect is IMO unfeasible (maybe one day once dialects are more complete). This rule essentially can just provide advice for users on deprecated segments and then they can make a value judgement on if they care about the deprecation.

PR #1986, opened 2021-11-27
Add config CLI argument to lint, fix, and parse

### LINKED ISSUE #1244: add sqlfluff CLI option to pass the configuration files
I would like to have possibility to pass sqlfluff configuration files via CLI option (instead of current predefined list of directories to look up for config files) with similar inheritance/override logic we have now.

The use case for which I need it:
I have multiple python projects in my company, and they all use a python module I've created which contains all the common configuration and functionality used by all my projects. So now I want to put most the sqlfluff configuration (like the rule configurations, excluded rules, etc) in the shared module and inherit from there, as those configs will be the same for all my projects, but for example, jinja context variables are project-specific, so I don't want to put them in the config file in the shared python module. I want to have those jinja context variables per-project and have them within the project directory. 
And when running the sqlfluff command from the project, I can for example list 2 config files - one from the project and one from the installed python module.

Does this make sense?

jpy-git:

Fixes #1244.
This PR adds a command line argument `--config` to the CLI `lint`, `fix`, and `parse` commands. 

The user can specify an additional config file such that the override logic is (later files override earlier ones):
- `default_config.cfg` (package default)
- `<user_appdir_configs>`
- `<user_homedir_configs>`
- `<user_homedir_configs>`
- `<configs_between_cwd_and_sql_path>`
- `<extra_config_file>` (this PR)
- Any command line arguments

This behaviour was voted on in the Slack channel: https://sqlfluff.slack.com/archives/C01GPHR8J0G/p1637859667129900

If no config files are included in the default config search paths then this argument can be used to specify a single off-path config file which will also be useful for several applications.
(Will investigate in a future PR adding `--ignore-local-config` option to allow only `--config` to be considered)

N.B. This PR focuses on the CLI, I will make a subsequent PR to flow the changes through to the Simple API as well.

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -177,7 +177,6 @@ def core_options(f: Callable) -> Callable:
     f = click.option(
         "--rules",
         default=None,
-        # short_help='Specify a particular rule, or comma separated rules, to check',
         help=(
             "Narrow the search to only specific rules. For example "
             "specifying `--rules L001` will only search for rule `L001` (Unnecessary "
@@ -189,7 +188,6 @@ def core_options(f: Callable) -> Callable:
     f = click.option(
         "--exclude-rules",
         default=None,
-        # short_help='Specify a particular rule, or comma separated rules to exclude',
         help=(
             "Exclude specific rules. For example "
             "specifying `--exclude-rules L001` will remove rule `L001` (Unnecessary "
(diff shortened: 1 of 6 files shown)


tunetheweb (commented):
LGTM but how do we know that the extra config is being used? Tests don't seem to check that.

Also test/fixtures/cli/extra_config_tsql.sql is empty which looks a little odd.

Maybe we could fix both concerns by adding some T-SQL specific SQL that fails in ANSI to check the dialect has indeed been changed as per the extra config?

jpy-git:
@tunetheweb weird that's exactly what the test was! Not sure how it didn't commit that sql but will add it back in now 😄

jpy-git:
@tunetheweb I've re-added that missing unit test back in 👍

tunetheweb (approved):

PR #1987, opened 2021-11-27
BigQuery: Allow keywords in column reference components

tunetheweb:
### Brief summary of the change made
Fixes #1981 

This is valid BigQuery even though `current` is a keyword:

```sql
-- current is a reserved word but keywords are allowed as part of a nested object name
SELECT
  table1.current.column,
  table1.object.current.column,
  table1.object.nested.current.column,
FROM
  table1
```

Docs: https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical

> A reserved keyword must be a quoted identifier if it is a standalone keyword or the first component of a path expression. It may be unquoted as the second or later component of a path expression.

### Are there any other side effects of this change that we should be aware of?
Only handles Column References. There would be more places where this should be implemented but think this is the main one.

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `python test/generate_parse_fixture_yml.py` or by running `tox` locally).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

Issue #1981, opened 2021-11-26
Error: "pop from an empty set"


jamiet-msm:
## Expected Behaviour

No runtime error occurs

## Observed Behaviour

Runtime error

```default
CRITICAL   [L054] Applying rule L054 threw an Exception: 'pop from an empty set' 
Traceback (most recent call last):
  File "/opt/conda/envs/py3/lib/python3.8/site-packages/sqlfluff/core/rules/base.py", line 324, in crawl
    res = self._eval(
  File "/opt/conda/envs/py3/lib/python3.8/site-packages/sqlfluff/rules/L054.py", line 112, in _eval
    column_reference_category_set.pop()
KeyError: 'pop from an empty set'
== [/dataplatform/google/runtime/modules/CommonDataModel/bigquery-views/cdm_processing.FctEnergyAppStatusUpdateProcessing.sql] FAIL
L:   7 | P:   1 | L054 | Unexpected exception: 'pop from an empty set'; Could you
                       | open an issue at
                       | https://github.com/sqlfluff/sqlfluff/issues ? You can
                       | ignore this exception for now, by adding '--noqa: L054'
                       | at the end of line 7
L:  17 | P:   1 | L054 | Unexpected exception: 'pop from an empty set'; Could you
                       | open an issue at
                       | https://github.com/sqlfluff/sqlfluff/issues ? You can
                       | ignore this exception for now, by adding '--noqa: L054'
                       | at the end of line 17
L:  27 | P:   1 | L054 | Unexpected exception: 'pop from an empty set'; Could you
                       | open an issue at
                       | https://github.com/sqlfluff/sqlfluff/issues ? You can
                       | ignore this exception for now, by adding '--noqa: L054'
                       | at the end of line 27
```

## Steps to Reproduce

Here is the content of the file referred to in the error message (cdm_processing.FctEnergyAppStatusUpdateProcessing.sql):

```sql
/* jscpd:ignore-start */
SELECT
    t.body.application.id AS ApplicationId,
    t.body.cashbackStatus.electricity.current.status AS ElectricityCashBackStatus, --noqa
    t.body.cashbackStatus.gas.current.status AS GasCashBackStatus --noqa
FROM `${datalake_project_id}.${name_prefix_sans_dashes}_event_service_presentation.applicationStatusUpdate_1_229_v2` AS t
GROUP BY  --noqa: PRS
    t.body.application.id,
    t.body.cashbackStatus.electricity.current.status, --noqa
[...]

PR #1988, opened 2021-11-27
Adds extra check to L054 to avoid weird error messages

tunetheweb:

Follow on from #1981 and #1987

Don't try to pop an empty column set in L054 to avoid weird error messages. Shouldn't really happen if there are no parsing errors and no `--noqa` (so took me a while to reproduce a test case!) but good check to have I think.

@jpy-git can you review too please since you added this rule?

Don't think so

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L054.py
+++ b/src/sqlfluff/rules/L054.py
@@ -96,6 +96,10 @@ class Rule_L054(BaseRule):
             if segment.name in column_reference_category_map
         }
 
+        # If there are no column references then just return
+        if not column_reference_category_set:
+            return LintResult(memory=context.memory)
+
         if self.group_by_and_order_by_style == "consistent":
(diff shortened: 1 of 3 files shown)


jpy-git:
@tunetheweb so the code change is perfect! ⭐ 

You're getting an error in the unit tests though. Reason being that it seems to raise if a parse error is detected at any point during the test, even if it's later suppressed by the inline comment.

jpy-git:
so may need a custom test or could be included as a parsing test and validate that no exception is raised

tunetheweb:
> so may need a custom test or could be included as a parsing test and validate that no exception is raised

Fixed! all tests passing now.

jpy-git:
Nice one, I'm happy with the change 👍

WittierDinosaur (approved):

PR #1990, opened 2021-11-27
Add config option to specify line_ending_export_style for fixed files

### LINKED ISSUE #797: Feature request: Enable use of unix-style line endings when running sqlfluff fix command on Windows
In our environment, we use Windows PCs for development, but the company standard is that SQL files have unix-style line endings (LF).  We use bash under WSL a lot, and set Windows editors to use LF only.  However, the only python we have available is on Windows. When running sqlfluff fix under python on Windows, the file gets changed from LF to Windows-style line endings CRLF.

I understand that this is python default behavior, but it would be nice to be able to over-ride the default and choose LF only, all the time.

jpy-git:

This PR fixes #797.
We provide a config option for the line ending to be used when exporting fixed sqlfluff files.
This defaults to the autodetect (lf for linux/mac, crlf for windows) or can be explicitly specified.

Example of getting crlf output on a mac:
Screen-Recording-2021-11-27-at-22 28 19

N.B. this is not a rule, python imports using universal newlines (all treated as \n) so has no impact on parsing/rule evaluation (also it's not like the users can do anything about it themselves if we flag it). This is only applied when the fixed file is written so non-fixed files will be unimpacted (if a more universal conversion to unix-style lines is required then using a program like `dos2unix` in their CI would be recommended).

No

### Pull Request checklist

  - Other.


jpy-git:
FYI investigating extending this fix to all files supplied to the fix command so hold off on review for now

src/sqlfluff/core/linter/linter.py
```diff
@@ -302,6 +303,35 @@ def _warn_unfixable(code: str):
...
 
+    @staticmethod
+    def get_line_ending_export_style(config: FluffConfig) -> Optional[str]:
+        """Get the line_ending_export_style to be used to export the fixed file.
+
+        Options are:
+        - autodetect: Use system default line ending.
+        - lf: Unix-style line ending.
+        - crlf: Windows-style line ending.
+        """
```
WittierDinosaur, line 314: You might want to add in Mac class style endings, which are just cr

jpy-git: @WittierDinosaur good point, it's just `cr: \r` so easy to add 👍

jpy-git:
Closing this PR. Reasoning being that this method although simple will only impact fixed files and could lead to confusion where users think their newlines will be fixed for files with no other linting issues.
I investigated extending this method to apply to all files supplied to fix but without fairly significant changes to how errors are flagged/cli messages are logged this is not possible.
The good news is I think I have a good approach to implement this via a rule which will follow shortly 👍

PR #1991, opened 2021-11-28
L044 should handle nested CTEs

### LINKED ISSUE #1984: L044: CTE in CTE doesn't pass linter
## Observed Behaviour
Rule `L044` fails when there's a CTE inside another CTE.

## Steps To Reproduce

Example SQL:
```sql
with a as (
    with b as (
        select 1
    )

    select * from b
)

select * from a
```

Linter output:
```
== [example/test.sql] FAIL
L:   9 | P:   1 | L044 | Query produces an unknown number of result columns.
```

[Link](https://online.sqlfluff.com/fluffed?sql=H4sIACUNoWEC_yvPLMlQSFRILFbQ4OVSAIJykEASkgAIFKfmpCaXKBhCRDR5uSAMqLCWQlpRfq5CEi8XWAZVNJGXCwBVidBWZAAAAA%3D%3D&dialect=snowflake) to SQLFluff Online output.

## Expected Behaviour
Linter rule `L044` should `PASS` on this, as there's just 1 column expected in the output.

## Debug Output
When ran with `-vvvvv`, here's the output:

```
DEBUG      [L044] Analyzing query: select * from a
DEBUG      [L044] Wildcard: * has target a
DEBUG      [L044] Analyzing query: select 1
DEBUG      [L044] Analyzing query: select * from b
DEBUG      [L044] Wildcard: * has target b
DEBUG      [L044] Query target b is external. Generating warning.
DEBUG      [L044] !! Violation Found: 'Query produces an unknown number of result columns.'
```

barrywhart:

Fixes #1984
### Pull Request checklist

  - Other.


src/sqlfluff/core/rules/analysis/select_crawler.py
```diff
@@ -29,15 +40,40 @@ def gather(
         cls, segment: BaseSegment, dialect: Dialect
     ) -> Dict[Optional[str], List["SelectCrawler"]]:
         """Find top-level SELECTs and CTEs, return info."""
-        queries = defaultdict(list)
+        queries: DefaultDict[Optional[str], List[SelectCrawler]] = defaultdict(list)
+        cls._process_select_statements(segment, dialect, queries)
+        return dict(queries)
+
+    @classmethod
+    def _process_select_statements(
```
barrywhart, line 48: New utility function, now used in two places (the new place handles nested CTEs)

src/sqlfluff/core/rules/analysis/select_crawler.py
```diff
@@ -29,15 +40,40 @@ def gather(
...
-            select_name = cls._get_name_if_cte(select_statement, segment)
-            queries[select_name].append(SelectCrawler(select_statement, dialect))
+            select_statement_info = cls._get_cte_info(select_statement, segment)
+            if select_statement_info:
+                queries[select_statement_info.select_name].append(
+                    SelectCrawler(select_statement, dialect, select_statement_info.cte)
+                )
+
+    @classmethod
+    def gather_nested_ctes(
```
barrywhart, line 67: New function, similar to the existing `gather()` function, but this one handles nested CTEs.

src/sqlfluff/core/rules/analysis/select_crawler.py
```diff
@@ -149,15 +189,18 @@ def get_wildcard_info(self) -> List[WildcardInfo]:
         return buff
 
     @staticmethod
-    def _get_name_if_cte(
+    def _get_cte_info(
```
barrywhart, line 192: Rework and rename this to also support nested CTEs

src/sqlfluff/rules/L044.py
```diff
@@ -102,11 +102,24 @@ def _analyze_result_columns(
...
                                 self._analyze_result_columns(
                                     queries.pop(wildcard_table), dialect, queries
                                 )
                             else:
+                                # Is it a nested CTE?
+                                if select_info.cte:
+                                    nested_queries = SelectCrawler.gather_nested_ctes(
+                                        select_info.cte.segments[-1], dialect
+                                    )
+                                    if wildcard_table in nested_queries:
```
barrywhart, line 115: Note the similarity of these lines to line 104 above. The block above handles CTEs at the same level, while this new block handles nested CTEs.

test/fixtures/rules/std_rule_cases/L044.yml
```diff
@@ -327,3 +327,46 @@ test_fail_unknown_number_of_result_columns_18:
...
+  pass_str: |
+    with a as (
+        with b as (
+            select 1 from c
+        )
+
+        select * from b
+    )
+
+    select * from a
```
jpy-git, line 342: for completeness, maybe add extra test cases:

```sql
with a as (
    with b as (
        select * from c
    )

    select 1 from b
)

select * from a
```

```sql
with a as (
    with b as (
        select * from c
    )

    select * from b
)

select 1 from a
```

jpy-git: don't think we need a test for it, but also checked 3 levels of nested with statements and it works perfectly 👍

barrywhart: I'll add those. 👍🏽

jpy-git: bit of a weird one but worth looking at if you get the chance, the rule seems to get tripped up if you mix nesting and same level ctes:
image

jpy-git: Another really weird edge case that also trips up L045:
image

barrywhart: Good test cases! I'm not sure what's happening in the first one. In the last one, I think L044 does not "see" CTEs defined at an outer scope, only the same level. I'll look into that and add those two cases.

I'll also create an issue about updating L045. Its design does not consider nested CTEs, either. In working on that other rule, I may go back and rework parts of L044 as well, if I see benefits to doing that.

barrywhart: Created this issue: #1999

jpy-git: perfect thanks 😄

barrywhart: After taking a bit closer look at the two new test cases, I think I'll go ahead and rework `SelectCrawler` now. Thus far, I had added some of the nested CTE logic directly in L044, but it has bugs, and if I've got to fix it anyway, I may as well improve the design as well. 👍🏽

jpy-git: makes sense!

had a couple more ideas for unit tests for L044 that we don't have in L044.yml yet, both of these pass currently but worth adding them 😄 

```sql
with a as (
    select 1 from b
)

select * from (
    select * from a
);
```

```sql
with b as (
    select 1 from c
)

select * from (
    with a as (
        select * from b
    )

    select * from a
);
```

src/sqlfluff/core/rules/analysis/select_crawler.py
```diff
@@ -1,126 +1,43 @@
...
 from cached_property import cached_property
 
 from sqlfluff.core.dialects.common import AliasInfo
 from sqlfluff.core.dialects.base import Dialect
-from sqlfluff.core.parser.segments.base import BaseSegment
+from sqlfluff.core.parser import BaseSegment
 from sqlfluff.core.rules.analysis.select import get_select_statement_info
 
 
+class QueryType(Enum):
```
barrywhart, line 14: This module has been heavily reworked, not only to support nested CTEs, but also:
* Improved object model that matches the structure of SQL better
* Streamlined the API of `SelectCrawler`

src/sqlfluff/rules/L045.py
```diff
@@ -44,25 +43,20 @@ class Rule_L045(BaseRule):
     """
 
     @classmethod
-    def _visit_sources(
```
barrywhart, line 47: This PR is focused on fixing L044, which involved major changes to `SelectCrawler`. Here, we update L045 so it works with the updated `SelectCrawler`. This PR does not address #1999, "L045 should handle nested CTEs".

barrywhart:
This is ready for review now, @jpy-git, @tunetheweb. The changes are significant; a thorough review would be cool but possibly not necessary, as the tests are really thorough. I'm probably most interested in a review of the design: the new classes, the APIs of those classes, etc.

jpy-git:
@barrywhart awesome work! I should have time this evening to review 😄

barrywhart:
Great, thanks!

src/sqlfluff/core/rules/analysis/select_crawler.py
```diff
@@ -148,16 +65,170 @@ def get_wildcard_info(self) -> List[WildcardInfo]:
...
+
+        def append_query(query):
+            if queries:
+                query.parent = queries[-1]
+            queries.append(query)
+            pop_queries_for.append(path[-1])
+
+        cte_name = None
+        for event, path in SelectCrawler.visit_segments(segment):
+            in_with = queries and queries[-1].query_type == QueryType.WithCompound
```
jpy-git, line 166: @barrywhart quick question for my personal curiosity :)

If we are at the top level of this recursion (i.e. the statement segment) then queries is an empty list. Therefore `in_with` evaluates to False, which is expected.

My question is: The `queries[-1]` on its own would be an index error, yet the whole expression still evaluates successfully to False. In Python does it not bother evaluating the second part of an `and` expression if the first part evaluates to false (and hence the expression can't possibly be True)?

barrywhart: That's right. When a language behaves this way, it's referred to as "short-circuit" evaluation, and it's a really convenient feature, because otherwise, you'd often have to chain together `if` statements to get the same effect.

Short-circuit evaluation is pretty common in programming languages. I know that C and C++ have it. I think Java does. For other languages, I'm not sure.

barrywhart: More detail here: https://www.geeksforgeeks.org/short-circuiting-techniques-python/

jpy-git: @barrywhart amazing, thanks for the link! Makes total sense, I was aware of the generator uses but will def be taking advantage of the boolean version going forwards 😄

src/sqlfluff/core/rules/analysis/select_crawler.py
```diff
@@ -148,16 +65,170 @@ def get_wildcard_info(self) -> List[WildcardInfo]:
...
+            if event == "start":
```
jpy-git, line 167: This logic all looks good, just thinking it might be nice to have a few small comments on the various conditions in this method for readabilty as it's quite dense?

Sort of like you did in `L044._analyze_result_columns`
image

barrywhart: Sure, happy to do that.

test/fixtures/rules/std_rule_cases/L044.yml
```diff
@@ -327,3 +327,128 @@ test_fail_unknown_number_of_result_columns_18:
...
+
+    a as (
+      with b as (
+        select * from c
+      )
+
+      select * from b
+    )
+
+    select * from a
```
jpy-git, line 454: Really tried to come up with some strange unit tests to break it but it's looking very robust!

A couple to add for completeness (all behave as expected):

- Pass: JOIN inside CTE with unqualified *
```sql
with cte as (
    select
        *
    from
        t1
    inner join t2
)

select a, b from cte;
```

- Pass: Known number of columns in two join subqueries.
```sql
select
    *
from (
    select a
    from foo
) t1
inner join (
    select b
    from bar
) t2;
```

- Fail: Two join subqueries, one with unknown number of columns.
```sql
select
    *
from (
    select *
    from foo
) t1
inner join (
    select b
    from bar
) t2;
```

jpy-git (commented):
@barrywhart this is really impressive stuff! Happy to approve once the comments and extra unit tests are added 🚀

barrywhart:
Thanks, @jpy-git, for the thorough review and the many suggestions for tests throughout this work. I added comments to that function, as well as the additional tests you suggested.

It's possible that we'll discover bugs in this rule at some point, but I feel pretty good that the design is solid, and _hopefully_ anything we missed can be addressed without significant rework.

While this code is complicated, I think it's a good investment, potentially a good direction for the project -- most of the rules in SQLFluff are either formatting or they take a very localized look at the code. Rules like this suggest the possibility of more sophisticated rules that can find more subtle bugs. I feel like the project is still very young, and who knows what it will grow into over the next few years. 💕💯

jpy-git (approved):
@barrywhart Amazing think it's looking fantastic, LGTM! 🚀 

Bit of blue sky thinking for future work, I know you're planning on updating L045 with this new SelectCrawler logic, but I'm also curious if it could have some useful application for L042 (Join/From clauses should not contain subqueries. Use CTEs instead) as well?

Since the select crawler can effectively resolve the ordering of nested subqueries and WITH statements, maybe we could expand the scope of L042 to detect these subqueries and nested structures and suggest to the user to use non-nested CTEs instead. Potentially there could be scope to apply a fix to the rule as well?

Can create an issue for it, but food for thought 😄

barrywhart:
@jpy-git: It would be fun to try that, definitely! An issue sounds like a good idea. Are you interested in working on it?

One possible corner case I can think of already: correlated subqueries. In that case, I don't think it could be converted to a CTE.

jpy-git:
I'll create the issue and then we can list out any potential edge cases, worth a look for sure 👍

PR #1994, opened 2021-11-28
Add encoding CLI argument

### LINKED ISSUE #654: Add a command-line option for file encoding (was: File encoding is not respected)

jpy-git:

This PR resolves #654.
The logic for specifying the encoding for reading/writing files has already been mostly implemented. This PR just adds the command line option.

No, extends the already existing config option.

### Pull Request checklist

  - Other.


src/sqlfluff/cli/commands.py
```diff
@@ -209,6 +209,13 @@ def core_options(f: Callable) -> Callable:
             "the standard configuration files. N.B. cfg format is required."
         ),
     )(f)
+    f = click.option(
```
barrywhart, line 212: Is this PR still in progress? I would expect to see code in `commands.py` that uses the `encoding` value, but I don't. (I should note that I don't fully understand how `@core_options` works.)

jpy-git: @barrywhart the PR is done, it's quite a simple one.

The core options get applied as a decorator to lint, fix, and parse. It is passed to these functions via `**kwargs`:
image

the `kwargs` then gets applied to the config in `get_config` via `overrides`:
image

and that's it, the rest has already been done in prior config work 👍

barrywhart (approved):
Thanks for the explanation! I thought there might be some magic there. 😛

PR #1985, opened 2021-11-26
Fix issue with inline ignores not respecting comment lines

jpy-git:
### Brief summary of the change made

Fixes #1977. Inline ignores weren't being processed on comment lines leading to issues like seen in the issue and the screenshot below. The fix is very simple.

image

### Are there any other side effects of this change that we should be aware of?
No

### Pull Request checklist
- [X] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `python test/generate_parse_fixture_yml.py` or by running `tox` locally).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #2002, opened 2021-11-29
Enable glob rule matching for inline comments

jpy-git:

This PR follows on from #1985 in which I added the ability to include/exclude rules using globs in the config. This PR allows rules to be included/excluded via globs in inline comments.

e.g. 
```sql
SELECT
    col_a a -- noqa: L01*
FROM foo
```

No

### Pull Request checklist

  - Other.


jpy-git:
@barrywhart @tunetheweb  could you take a look at this please, it extends the glob rule matching work in  #1985 to also apply to inline ignores.

src/sqlfluff/core/linter/linter.py
```diff
@@ -225,7 +226,11 @@ def _parse_tokens(
         return parsed, violations
 
     @staticmethod
-    def parse_noqa(comment: str, line_no: int):
+    def parse_noqa(
+        comment: str,
+        line_no: int,
+        rule_set: List[BaseRule],
```
barrywhart, line 232: Nitpick: For this and the other modified functions, should we consider passing in `List[str]` rather than `List[BaseRule]`, since that's needed? This is mostly a pedantic concern about excessive coupling, as they call it.

From [Wikipedia](https://en.wikipedia.org/wiki/Coupling_(computer_programming)):
>Stamp coupling (data-structured coupling)
>Stamp coupling occurs when modules share a composite data structure and use only parts of it, possibly different parts (e.g., passing a whole record to a function that needs only one field of it).
In this situation, a modification in a field that a module does not need may lead to changing the way the module reads the record.

src/sqlfluff/core/linter/linter.py
```diff
@@ -267,7 +272,21 @@ def parse_noqa(comment: str, line_no: int):
                             )
                     rules: Optional[Tuple[str, ...]]
                     if rule_part != "all":
-                        rules = tuple(r.strip() for r in rule_part.split(","))
+                        # Rules can be globs therefore we compare to the rule_set to expand the globs.
+                        unexpanded_rules = tuple(
+                            r.strip() for r in rule_part.split(",")
+                        )
+                        possible_rule_codes = [r.code for r in rule_set]
+                        expanded_rules = []
```
barrywhart, line 280: Is it possible we'll end up with multiple occurrences of the same rule? If so, should we filter out duplicates before returning?

E.g. `L0*5` and `L01*` will both match "L015".

test/core/linter_test.py
```diff
@@ -371,27 +372,52 @@ def test__linter__encoding(fname, config_encoding, lexerror):
     assert lexerror == (SQLLexError in [type(v) for v in result.get_violations()])
 
 
+# noqa tests require a rule_set, therefore we construct dummy rule set for glob matching.
+dummy_rule_set = [BaseRule(f"L{i:03d}", "dummy_rule") for i in range(1, 100)]
```
barrywhart, line 376: Will this require updating if/when the number of rules in SQLFluff exceeds 100? Should we try and avoid that, e.g. by building this dummy set based on the actual set of rules?

test/core/linter_test.py
```diff
@@ -371,27 +372,52 @@ def test__linter__encoding(fname, config_encoding, lexerror):
...
-        ("", None),
-        ("noqa", NoQaDirective(0, None, None)),
-        ("noqa?", SQLParseError),
-        ("noqa:", NoQaDirective(0, None, None)),
-        ("noqa:L001,L002", NoQaDirective(0, ("L001", "L002"), None)),
-        ("noqa: enable=L005", NoQaDirective(0, ("L005",), "enable")),
-        ("noqa: disable=L010", NoQaDirective(0, ("L010",), "disable")),
-        ("noqa: disable=all", NoQaDirective(0, None, "disable")),
-        ("noqa: disable", SQLParseError),
+        ("", None, dummy_rule_set),
```
barrywhart, line 382: Nitpick: It appears that all the cases of this test (`test_parse_noqa `) use the same rule set. If so, should we consider omitting it from the fixture and just using `dummy_rule_set` inside the test function? Or do we expect that future tests may use something other than `dummy_rule_set`?

src/sqlfluff/core/linter/linter.py
jpy-git: Think I just had it there for convenience when I was adding the additional logic to this function but definitely makes sense to clean up the coupling as you suggest 👍

src/sqlfluff/core/linter/linter.py
jpy-git: It's possible, but I thought since we later just check if the raised rule is `in` the rules list it wasn't a big issue. However, I think you're right in terms of a design decision as it means the results from this function are better defined 👍

test/core/linter_test.py
jpy-git: yeah I agree, I did actually initially do it like that but writing functions that reference global variables never sits well with me so I included it as an explicit argument to the function. tbh you're right though because it's just a test case, not a function that would ever get called elsewhere. Force of habit on my behalf 😛

src/sqlfluff/core/linter/linter.py
barrywhart: Cool. One reason I think this is worth avoiding (i.e. fixing) is that it could somehow cause the same rule to be run twice -- if not today, maybe in a future version.

jpy-git: makes sense, I've added an added some logic to remove dups and a test rule to verify it 👍

jpy-git:
@barrywhart implemented the feedback 😄

src/sqlfluff/core/linter/linter.py
```diff
@@ -267,7 +272,25 @@ def parse_noqa(comment: str, line_no: int):
...
+                                if r not in expanded_rules:
+                                    expanded_rules.append(r)
+                            else:
+                                expanded_rules.extend(
+                                    [
+                                        x
+                                        for x in fnmatch.filter(rule_codes, r)
+                                        if x not in expanded_rules
+                                    ]
+                                )
```
tunetheweb, line 292: Bit worried about hardcoding the list of special error types.

Shouldn't we just say if the unexpended rule contains a `*` then expand? If not don't expand?

test/core/linter_test.py
```diff
@@ -623,6 +667,9 @@ def test_linter_noqa():
         col_q q, --Inline comment --noqa: L012
         col_r r, /* Block comment */ --noqa: L012
         col_s s # hash comment --noqa: L012
+        , col_t t --noqa: L01*
+        , col_u u -- Some comment --noqa: L01*
+        , col_v v -- We can ignore both L012 and L019 -- noqa: L01[29]
```
tunetheweb, line 672: Why are we mixing trailing and leading commas here?

tunetheweb (commented):
Suggested some changes but in general looks good and am happy with this. Let me know what you think on the comments and can then decide whether to merge as is or not.

test/core/linter_test.py
jpy-git: This is so it triggers both L012 and L019, otherwise we're not really testing that the glob matching ignores multiple rules.

src/sqlfluff/core/linter/linter.py
jpy-git: sure, glob matching actually has more patterns than just `*` (e.g. `L01[12]` matches L011 and L012) so we can't just filter on that. 

let me change this section to 
```python
expanded_rule = [
    x
    for x in fnmatch.filter(rule_codes, r)
    if x not in expanded_rules
]
expanded_rules.extend(expanded_rule) if expanded_rule else expanded_rules.append(r)
```

That way we expand if we can else we assume the user is referencing a special error type.

test/core/linter_test.py
tunetheweb: Ah that wasn’t obvious to me. Could we expand the comment to note that? Or add another comment line above these three lines?

jpy-git: Will do 👍

jpy-git:
@tunetheweb I've updated the logic to not explicitly reference the special error types 😄

tunetheweb (approved):
LGTM!

Issue #1304, opened 2021-08-20
Running indentation fixes adds whitespace to various code segments


tunetheweb:
As a follow on from #1302 and #1303 (and likely related to #1149 and possibly others) I have noticed that when running `sqlfluff fix` on SQL which requires indentation fixes, the extra white space is added to the code, rather than as addition whitespace segments.

So the `function_name_identifier` of `UNNEST` can become `••••UNNEST` (where • represents a space).

This can then cause unexpected consequences when the next rule runs in that test as suddenly the parsed tree is different than what it should be.

## Expected Behaviour

Whitespace should not be added to other identifiers/segments as this can break expectations.

## Observed Behaviour

Whitespace is added other identifiers/segments, resulting in the next rule in that run potentially not running properly.

## Steps to Reproduce

Add the below print statement to `src/sqlfluff/core/rules/analysis/select.py`:

```py
    for function_name in table_expr.recursive_crawl("function_name"):
        print('LOOK AT THIS:%s:' % function_name.raw.lower())
        # Other rules can increase whitespace in the function name, so use strip to remove
        # See: https://github.com/sqlfluff/sqlfluff/issues/1304
        if function_name.raw.lower() in dialect.sets("value_table_functions"):
            return True
    return False
```

Create a test.sql file with this:

```sql
SELECT
    category,
    value
FROM
    table1,
UNNEST(1, 2, 3) AS category
```

Then run this command:

```
sqlfluff fix --dialect bigquery test.sql --fixed-suffix FIXED --force --rules L003,L025
```

And note that even though we are only printing the function name, the spaces are included:

```
==== finding fixable violations ====
LOOK AT THIS:    unnest:
LOOK AT THIS:    unnest:
LOOK AT THIS:    unnest:
...etc.
````

I've confirmed it's not a bigquery or function_name specific issue as using this SQL:

```sql
SELECT
    category,
    value
FROM
    table1,
table2
```
And adding this to the `_has_value_table_function`:

```py
    for table_ref in table_expr.recursive_crawl("table_reference"):
        print("LOOK AT THIS:%s:" % table_ref.raw)
```

Then running this command:

```
sqlfluff fix test.sql --fixed-suffix FIXED --force --rules L003,L025
```

Leads to the same issue:

```
==== finding fixable violations ====
LOOK AT THIS:table1:
LOOK AT THIS:    table2:
LOOK AT THIS:table1:
LOOK AT THIS:    table2:
== [test.sql] FAIL
[...]

Issue #1012, opened 2021-04-29
Option to automatically re-parse the SQL after "fix"


barrywhart:
We've seen some bugs in the past where a lint fix doesn't work correctly, and the resulting SQL is invalid (won't even parse). Should we add an option to `fix` that re-parses the output SQL files? This won't catch every possible kind of bug, but it would likely catch some bugs and avoid trouble for users.

This is similar to #813, but far more concrete and less ambitious. Maybe we could do this first.

Inspired by a [discussion](https://sqlfluff.slack.com/archives/C01GX4L213J/p1619590699133000) in Slack, where a user saw something similar to this behavior but doesn't yet have a repro case.

```
SELECT DISTINCT a.id as new_id
```

was "fixed" to become:
```
SELECT a.id as new_id DISTINCT
```

tunetheweb:
Can this be closed now with your recent work to revalidate SQL @barrywhart

barrywhart:
Ah yes, good call.

PR #2004, opened 2021-11-30
L008 refactor

### LINKED ISSUE #2001: Combination of L008 and L019 causes L008 triple count
## Expected Behaviour
Running `sqlfluff lint` on 
```sql
SELECT
    col_a AS a
    ,col_b AS b
FROM foo;
```
should raise one L008 (Commas should be followed by a single whitespace unless followed by a comment.) and one L019 (Found leading comma. Expected only trailing.). 

## Observed Behaviour
L008 is raised 3 times:
image

## Steps to Reproduce
`sqlfluff lint` on the query above.

## Dialect
ansi

## Version
sqlfluff, version 0.8.2 (HEAD of main branch)

## Configuration
default config

jpy-git:

This PR fixes #2001. The column reference following the comma had multiple child segments, therefore each child segment would detect the same comma and re-trigger the rule. I fixed this by using memory to keep a log of which commas have previously been fixed and added a test case to verify the single counting.

No

### Pull Request checklist

  - Other.


jpy-git:
@barrywhart could you take a look at this please, it uses memory to log fixed commas so we don't double flag them

barrywhart:
I can believe that this works, but it doesn't quite feel like a "fix" to me. My instinct is that there's a design flaw in the rule. Is there a way to write the rule so that it only checks for the issue for certain segments, so that by design, it only detects the error once? E.g. the first segment after the comma or the first raw segment after the comma?

In particular, [this code](https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/rules/L008.py#L46L47) seems odd to me: IIUC, rather than processing the segment that was passed by the linter, it looks at another segment related to that one.

Would love your thoughts on this. I can take a closer look if my advice seems too vague or wrongheaded (always a possibility!!).

jpy-git:
@barrywhart makes sense, let me have a look at a more comprehensive fix and get back to you on this one tomorrow 👍

src/sqlfluff/rules/L008.py
```diff
@@ -34,45 +36,75 @@ class Rule_L008(BaseRule):
...
-                return LintResult(
-                    anchor=cm1,
-                    fixes=[LintFix("edit", context.segment, [ins, context.segment])],
-                )
-
-        if len(context.raw_stack) < 2:
+    def _eval(self, context: RuleContext) -> Optional[LintResult]:
+        """Commas should be followed by a single whitespace unless followed by a comment."""
+        # We only care about commas.
+        if not context.segment.name == "comma":
```
jpy-git, line 79: rule is evaluated relative to the comma so no weird double counting

jpy-git (commented):
@barrywhart this is ready for review now 😄 Hopefully a much more robust implementation of L008

src/sqlfluff/rules/L008.py
```diff
@@ -34,45 +36,75 @@ class Rule_L008(BaseRule):
...
         """
-        if len(context.raw_stack) < 1:
-            return None
+        subsequent_whitespace = None
+        # Get all raw segments and find position of the current comma within the list.
+        file_segment = context.parent_stack[0]
+        # Raw stack is appropriate as the only segments we can care about are
+        # comma, whitespace, newline, and comment, which are all raw.
+        # Using the raw_segments allows us to account for possible unexpected
+        # parse tree structures resulting from other rule fixes.
```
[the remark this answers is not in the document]
barrywhart: It sounds like you may be seeing an occurrence of #1304. I would really like to get all these rule bugs fixed (I suspect many of them involve misuse of `LintFix` type `edit`) and add a linter check to prevent such bugs in the future.

Other related issues:
* #1304: Another apparent rule bug
* #1012: Vague idea about a general way of detecting all such issues

src/sqlfluff/rules/L008.py
```diff
@@ -34,45 +36,76 @@ class Rule_L008(BaseRule):
...
-        after the problem site. NB: We need at least two segments behind us for this to work.
+        Return a tuple of both the trailing whitespace segment and the
+        first non-whitespace segment discovered.
         """
-        if len(context.raw_stack) < 1:
-            return None
+        subsequent_whitespace = None
+        # Get all raw segments and find position of the current comma within the list.
+        file_segment = context.parent_stack[0]
+        raw_segments = file_segment.get_raw_segments()
```
barrywhart, line 51: Could you use `context.raw_stack` here instead? I think it may provide the same thing.

src/sqlfluff/rules/L008.py
```diff
@@ -34,45 +36,76 @@ class Rule_L008(BaseRule):
...
+            if s.is_meta:
+                continue
+            elif s.is_type("whitespace"):
+                # Capture the whitespace segment.
+                subsequent_whitespace = s
+            else:
+                # We've found a non-whitespace (and non-meta) segment.
+                # Therefore return the stored whitespace segment
+                # and this segment for analysis.
+                return subsequent_whitespace, s
```
barrywhart, line 69: Is it possible (and do we care) if `subsequent_whitespace` is `None` here?

src/sqlfluff/rules/L008.py
```diff
@@ -34,45 +36,76 @@ class Rule_L008(BaseRule):
...
+
+        # If we find ourselves here it's all
+        # whitespace (or nothing) to the end of the file.
+        # Shouldn't be possible to have a comma at the end of a file
```
barrywhart, line 73: I think it's possible in BigQuery, which allows trailing commas, e.g.:
```
SELECT 
    1,
    2,
```

src/sqlfluff/rules/L008.py
barrywhart: Nitpick: Perhaps it'd be clearer to say `if context.segment.name != "comma":` ?

barrywhart:
Looks good!  A few small questions, just about ready for merge.

src/sqlfluff/rules/L008.py
jpy-git: will do!

src/sqlfluff/rules/L008.py
jpy-git: @barrywhart `context.raw_stack` is different from `file_segment.get_raw_segments()`

`context.raw_stack` is a tuple of previously visited raw segments whereas `file_segment.get_raw_segments()` gives a list of all raw segments (including ones AFTER the current segment). This is important as we look ahead of the comma to see if there's whitespace so I think this is the correct approach

src/sqlfluff/rules/L008.py
jpy-git: It is possible, for example the comma is the last character in the file. In this case the conditions in `_eval` ignore the segment so we don't really care about it. L009 can handle adding in the final newline in that case 👍

src/sqlfluff/rules/L008.py
jpy-git: Ah makes sense, I always wondered why L038 was a rule rather than just not parsing. Let me update the comment here and add a unit test for that bigquery example 😄

src/sqlfluff/rules/L008.py
barrywhart: Oh, of course... makes sense! 👍🏽

jpy-git:
@barrywhart Implemented the feedback 😄

barrywhart (approved):
Looks great! 🎉

jpy-git:
@barrywhart I won't steal your merge this time! 😄

barrywhart:
Whee!! Now my weekend is complete.

src/sqlfluff/rules/L008.py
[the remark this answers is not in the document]
jpy-git: @barrywhart so I think a lot of these issues aren't necessarily an `edit` thing, you would get a similar issue with `create_before` or `create_after` as they all work in a similar way. 

Reading through the `apply_fixes` logic (https://github.com/sqlfluff/sqlfluff/blob/0624a848057dcbf020dd07fe3999605e5f2f973e/src/sqlfluff/core/parser/segments/base.py#L938), a fix is applied by:
- Search down the parse tree, at each level appending each segment you find to segment buffer.
- If you find the anchor segment (of the fix not the result) then either
edit: add the new segments to the buffer, but not the anchor.
create_before: add the new segment to the buffer and then the anchor segment
create_after: add the anchor segment to the buffer and then add the new segment
- then subsequent segments are added to the buffer.

The key thing to note here is that all fixes segments are inserted at the SAME level as the fix anchor_segment. The reason we seem to get a lot of issue like the one above is some logic (in this case I think it's L016) finds a segment to analyse and determine if a fix needs to be applied and then naively just applies the fix at the same tree level even if not appropriate.

So in this case L016 probably determines that it can split a line by looking at the first raw segment after a space, which is the function_name_identifier, and then inserts the newline and whitespace before that. Something like `LintFix("create_before", <function_name_identifier_segment>, [NewlineSegment(), WhitespaceSegment()])`, when in fact from the parse tree the appropriate fix is actually `LintFix("create_before", <expression_segment>, [NewlineSegment(), WhitespaceSegment()])`.
So there needs to be bit of extra logic/consideration to ensure the fix is actually placed in the correct place in the parse tree rather than to just get the output appearing correct in the simple case. That being said it is hard sometimes to implement.

A good example of this is the recent PR we did for #1979, in which we detect the linting error at the final code segment of the tree but apply the fix at the FileSegment in order to get the correct parse structure.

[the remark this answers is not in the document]
jpy-git: ^That's basically the cause of these bugs, however, given that it's sometimes hard to create logic to determine the correct fix placement (especially when the parse grammar itself can be inconsistent) I do think that #1012 is an interesting idea.

If you were, for example, to take the messed up post-fix parse tree structure shown above and re-parse the updated `<file_segment>.raw` to then supply to subsequent fix-rounds you would maintain the fixes whilst giving subsequent rules the correct parse tree. (guess it would be between loops in `lint_fix_parsed`)

It might not be the most efficient but could be more practical given that we are unlikely to have completely well defined grammar anytime soon?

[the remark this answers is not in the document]
barrywhart: I have also been wondering if we could run the relevant `match()` functions to detect and complain if a rule does something invalid to the parse tree. I think I need to look at some of the bugs more closely to decide if it's reasonable to flag/fix them.

Re-parsing is an alternative -- it should definitely work, but it feels like cheating, and in order to be 100% sure, we'd potentially need to re-parse after every fix (or every rule run)? It seems like it'd be useful to know which rules have these issues so we only re-parse after running a problematic rule.

PR #2019, opened 2021-12-01
Fix inline comment interactions with L052

### LINKED ISSUE #2014: L052 - Same line semicolon snaps it onto comments
## Expected Behaviour

sqlfluff fix
before:
```
SELECT a
FROM b --comment
;
```

after:
```
SELECT a
FROM b; --comment
```

## Observed Behaviour

after:
```
SELECT a 
FROM b --comment;
```

## Steps to Reproduce

sqlfluff fix the before block

## Dialect

ansi

## Version
0.8.2

## Configuration
Default - but crucially L052 `semicolon_newline = False`

jpy-git:

Fix #2014 . This PR addresses the linked issue as well as better defining the interaction of L052 with inline comments.

Unit tests show all the various permutations of L052 comment interactions.

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L052.py
+++ b/src/sqlfluff/rules/L052.py
@@ -48,7 +48,11 @@ class Rule_L052(BaseRule):
 
     @staticmethod
     def _handle_preceding_inline_comments(pre_semicolon_segments, anchor_segment):
-        """Adjust pre_semicolon_segments and anchor_segment to not move preceding inline comments."""
+        """Adjust pre_semicolon_segments and anchor_segment to not move preceding inline comments.
+
+        We don't want to move inline comments that are on the same line
+        as the preceding code segment as they could contain noqa instructions.
+        """
         # See if we have a preceding inline comment on the same line as the preceding segment.
         same_line_comment = next(
             (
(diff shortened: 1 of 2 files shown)


WittierDinosaur:
Probably worth adding test cases for multiline comments too:
```
SELECT *
FROM foo /* multiline
comment
*/
;
```

WittierDinosaur:
Ah, good question. I don't think it matters too much, as long as we're consistent. Maybe get some extra opinions? @tunetheweb @barrywhart

WittierDinosaur:
I think I prefer the former

jpy-git:
I like the first one too and that one is easier to implement. I don't really want to add the extra options as I feel we'd be adding edge cases to edge cases, so unless anyone strongly disagrees lets go for option 1 👍

jpy-git:
Added the handling and tests for block comments

barrywhart:
Can you add two additional "block comment" test cases for the case where the multiline comment is already on a separate line? (These are almost identical to the two existing tests; I just moved the comment down.)

```
    SELECT foo
    FROM bar
    /* multiline
    comment
    */
    ;
```

```
    SELECT foo
    FROM bar
    ;
    /* multiline
    comment
    */
```

test_fail_newline_trailing_block_comment:
  fail_str: |
    SELECT foo
    FROM bar; /* multiline
    comment
    */

jpy-git:
@barrywhart got those added 👍

WittierDinosaur (approved):

PR #2020, opened 2021-12-01
Fix config.get for iterable sections

jpy-git:

Was using config.get method and realised it should take an iterable of str to handle nested config sections but the current implementation only works for a single section str.
Used the already existing get_sections method to handle this.

No

### Pull Request checklist

  - Other.


src/sqlfluff/core/config.py
```diff
@@ -615,7 +615,11 @@ def get(
         self, val: str, section: Union[str, Iterable[str]] = "core", default: Any = None
     ):
         """Get a particular value from the config."""
-        return self._configs[section].get(val, default)
```
jpy-git, line 618: would only work for if session a str

src/sqlfluff/core/config.py
```diff
@@ -615,7 +615,11 @@ def get(
...
+        section_dict = self.get_section(section)
```
jpy-git, line 618: handles nested config sections

src/sqlfluff/core/config.py
```diff
@@ -615,7 +615,11 @@ def get(
...
+        if section_dict is None:
+            return default
```
jpy-git, line 620: returns default if the section doesn't exist. Could alternatively raise a KeyError if preferred. Fine either way

jpy-git:
@barrywhart could you take a look at this please, it's a quick bugfix for the config.get method

barrywhart (approved):

Issue #2015, opened 2021-12-01
Create/Drop Policy Support


mariugul:
## Expected Behaviour
Should be able to create and drop policies using the PostgreSQL dialect.

`CREATE POLICY name ON table_name []...`
[Create Policy Documentation](https://www.postgresql.org/docs/10/sql-createpolicy.html)

`DROP POLICY [ IF EXISTS ] name ON table_name [ CASCADE | RESTRICT ]`
[Drop Policy Documentation](https://www.postgresql.org/docs/10/sql-droppolicy.html)

**Expected**:
`All Finished 📜 🎉!  `

## Observed Behaviour

```bash
$ sqlfluff lint --dialect postgres policy.sql 
== [policy.sql] FAIL                                                                              
L:   6 | P:   1 |  PRS | Line 6, Position 1: Found unparsable section: 'CREATE
                       | POLICY p1 ON person'
L:   7 | P:   1 |  PRS | Line 7, Position 1: Found unparsable section: 'DROP
                       | POLICY p1 ON person'
All Finished 📜 🎉!
```

## Steps to Reproduce

**policy.sql**:
```sql
CREATE TABLE person (
    id serial,
    name text
);

CREATE POLICY p1 ON person;
DROP POLICY p1 ON person;
```

**Lint policy.sql**:
```bash
sqlfluff lint --dialect postgres policy.sql 
```

## Dialect
`postgres`

## Version
`sqlfluff, version 0.8.2`
`Python 3.8.10`
`pip 20.0.2 `

## Configuration
```
--dialect postgres
```

PR #2021, opened 2021-12-01
Grammar: Postgres CREATE POLICY

jpy-git:

Partially addresses #2015 (can add drop policy afterwards). Adds the CREATE POLICY statement to the postgres dialect.

docs here: https://www.postgresql.org/docs/14/sql-createpolicy.html

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_postgres_keywords.py
```diff
@@ -203,7 +203,7 @@ def get_keywords(keyword_list, keyword_type):
     ("CURRENT_TIME", "reserved"),
     ("CURRENT_TIMESTAMP", "reserved"),
     ("CURRENT_TRANSFORM_GROUP_FOR_TYPE", "not-keyword"),
-    ("CURRENT_USER", "reserved"),
+    ("CURRENT_USER", "non-reserved"),
```
jpy-git, line 206: `USING (ename=current_user)` is common in CREATE POLICY statements.
I needed to set this to non-reserved in order to get the expression above to parse. Let me know if this is incorrect/there's a better method 😄

WittierDinosaur: If you want to overwrite behaviour that is specified in the docs - please can you add it to the postgres_nondocs_keywords lower down. They take precedence, and I like to keep parity with the docs wherever possible

jpy-git: @WittierDinosaur thanks, moved the keyword down to `postgres_nondocs_keywords` now so this should be good to go 🚀

WittierDinosaur (approved):

PR #2028, opened 2021-12-02
TSQL (ASA): Fix CTAS with WITH statement

jpers36:
CTASes with a WITH statement may not parse properly due to ANSI's WithCompoundStatementSegment's use of StartsWith, which is too greedy for TSQL with its optional delimiters.  This PR adds a TSQL-specific definition without StartsWith.

Added a test case as well.


tunetheweb (approved):
LGTM

PR #2031, opened 2021-12-03
TSQL (ASA): Allow for table identifier in DELETE clause

jpers36:
TSQL (Azure Synapse Analytics) allows for a table to be specified as part of a DELETE clause in order to support JOINs within a DELETE statement's WHERE clause.

+TableReferenceSegment ref at the right place in TSQL's DeleteStatementSegment
+Test case


tunetheweb (approved):
LGTM

PR #2034, opened 2021-12-04
Linguist: Include test/** in language statistics to better reflect use of SQL

### LINKED ISSUE #2032: Add SQL as a SQLFluff repository language
Github lists the repository's languages as 99.9% Python, 0.1% Dockerfile.  Is there a way we can get SQL represented?

jpy-git:

Fixes #2032

GitHub uses [Linguist](https://github.com/github/linguist) to determine language statistics in the repo.
Our language stats currently show 99.9% Python and 0.1 Dockerfile but no SQL.

On further investigation it appears Linguist has a list of vendored file paths that are excluded from the statistics and one of those is test/fixtures, which is where all of our SQL is stored.(https://github.com/github/linguist/blob/c8ca48856b1998681cdf1c1234145532006f95d8/lib/linguist/vendor.yml#L317)

Linguist provides the ability to override default behaviour via the .gitattributes file (https://github.com/github/linguist/blob/master/docs/overrides.md)

Given the projects focus on SQL we feel it would be nice to include the SQL files in our language statistics as it indicates to potential contributors that SQL knowledge and understanding is a valuable skill that they can bring to the project.

Even with these SQL files included Python should still (rightly so) be the main language by far.

No, but I think the Linguist language detection job is fairly low priority so may take a while to update.

### Pull Request checklist

  - Other.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,9 @@
 # Isolate binary files in case the auto-detection algorithm fails and
 # marks them as text files (which could brick them).
 *.{png,jpg,jpeg,gif,webp,woff,woff2} binary
+
+# Linguist was excluding our test suite from the repo language statistics
+# and as a result the repo indicated that no SQL is present in the repo.
+# Information on overrides can be found here:
+# https://github.com/github/linguist/blob/master/docs/overrides.md
+test/** linguist-vendored=false


tunetheweb (approved):
Nice find

PR #2037, opened 2021-12-05
Fix parameters for Snowflake create tasks statement

chwiese:

Fixes options for the CREATE TASK statement in the Snowflake dialect:
- One can only specify the `WAREHOUSE` parameter _or_ the `USER_TASK_MANAGED_INITIAL_WAREHOUSE_SIZE` parameter, but not both
- The parameters can be in arbitrary order, except for `AFTER` which must be last
- The value for `USER_TASK_MANAGED_INITIAL_WAREHOUSE_SIZE` is a warehouse size, not a quoted literal. Warehouse sizes themselves are not identifiers, so I am changing the type of that RegexParser.
- The quotes around the warehouse sizes are optional, except for the ones with hyphens, which must be quoted. The same rules apply the `CREATE WAREHOUSE` statements, even if that is not clearly documented there.

None.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -1653,55 +1654,52 @@ class CreateTaskSegment(BaseSegment):
...
+                Sequence(
+                    Ref("ParameterNameSegment"),
+                    Ref("EqualsSegment"),
+                    OneOf(
+                        Ref("BooleanLiteralGrammar"),
+                        Ref("QuotedLiteralSegment"),
+                        Ref("NumericLiteralSegment"),
+                    ),
                 ),
+                delimiter=Ref("CommaSegment"),
```
jpy-git, line 1690: This is the default delimiter so we don't need it here

jpy-git (commented):
@chwiese very small tweak to remove the unnecessary default delimiter, but otherwise perfect and will approve and merge after once that line is removed ⭐

src/sqlfluff/dialects/dialect_snowflake.py
chwiese: Nice, I just removed it.

jpy-git (approved):
LGTM, well done @chwiese 🚀 🌟

Issue #2029, opened 2021-12-02
T-SQL: Semi-colon unparseable for various statements


skykasko:
## Expected Behaviour

SQLFluff should parse [USE][] statements in T-SQL that end with a semicolon.

[USE]: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/use-transact-sql

## Observed Behaviour

```shell
$ sqlfluff lint --dialect tsql use.sql
== [use.sql] FAIL
L:   1 | P:  23 |  PRS | Line 1, Position 23: Found unparsable section: ';'
All Finished 📜 🎉!
```

## Steps to Reproduce

Create `use.sql`, taken from the [example][] in the Microsoft documentation:

```sql
USE AdventureWorks2012;
GO

```

Run `sqlfluff lint --dialect tsql use.sql`.

Note that no parsing error is reported if the semicolon is removed. The presence or absence of the `GO` command makes no difference.

[example]: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/use-transact-sql#examples

## Dialect

`tsql`

## Version

- SQLFluff: 0.8.2
- Python: 3.9.9

jpy-git:
Think this is a bigger issue as something like `DROP SCHEMA sdfa;` also won't parse at the semi-colons. I imagine there are a lot more, especially for grammar inherited from ANSI.

The semi-colon implementation is a bit weird for T-SQL.

For every other dialect the statement terminator `;` is on the same parse depth as the statement segment itself:
![image](https://user-images.githubusercontent.com/80432516/144527446-426261f0-4f09-4296-bac9-3f6594fdd0d9.png)
and seems to be generic and not referenced in any of the statements themselves.

However, for T-SQL it seems to be at a random depth in the parse tree (same level as FROM in selectable):
![image](https://user-images.githubusercontent.com/80432516/144527572-eda6bfa8-4afc-44d3-8f07-a2ddb1eae1bb.png)
and is explicitly referenced in the grammar in various places:
![image](https://user-images.githubusercontent.com/80432516/144527698-3216b8cc-06cb-4f70-91ef-d0789f934d65.png)

Seems to me like the semi-colon segment in T-SQL should be similar to ANSI and parsed at the same level as StatementSegment.

jpy-git:
Initial thoughts are the BatchSegment T-SQL needs to be treated more similarly to FileSegment in ANSI.

[...]

PR #2044, opened 2021-12-05
TSQL: TRY/CATCH

### LINKED ISSUE #1541: Try-Catch in TSQL
TSQL support error handling with `begin try`, `end try`, `begin catch` and `end catch` (https://docs.microsoft.com/en-us/sql/t-sql/language-elements/try-catch-transact-sql?view=sql-server-ver15).

Would be nice to support this also in sqlfluff, with correct indentation and linebreaks.

jpers36:
Fixes #1541

Adding TRY/CATCH logic to TSQL.  This block pair allows for exceptions to be caught and handled within TSQL.  Added a test case as well.

As part of this change, I tweaked the way that BeginEndSegment is used.  BeginEndSegment was always referenced as an alternative to StatementSegment and vice versa, so I collapsed BeginEndSegment into StatementSegment.  This was to simplify TRY/CATCH's implementation, which is very similar to BeginEndSegment.  This caused trivial changes to a few BEGIN/END test cases.


src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -2104,15 +2100,49 @@ class BeginEndSegment(BaseSegment):
         Ref("DelimiterSegment", optional=True),
         Indent,
         AnyNumberOf(
-            OneOf(
-                Ref("BeginEndSegment"),
+            Ref("StatementSegment"),
```
jpy-git, line 2103: Do we not also want a `Ref("DelimiterSegment", optional=True),` after this line, like we've added to TryCatchSegment?

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -2104,15 +2100,49 @@ class BeginEndSegment(BaseSegment):
...
+    """A `TRY/CATCH` block pair.
+
+    https://docs.microsoft.com/en-us/sql/t-sql/language-elements/try-catch-transact-sql?view=sql-server-ver15
+    """
+
+    type = "try_catch"
+    match_grammar = Sequence(
+        "BEGIN",
+        "TRY",
+        Indent,
```
jpy-git, line 2122: In `BeginEndSegment` it's possible to have semicolons like:
```sql
BEGIN;
    SELECT 1;
END;
```
is it also possible to have semicolons after `BEGIN TRY`, `END TRY`, `BEGIN CATCH`, `END CATCH`?

jpy-git (commented):
@jpers36 just a couple questions to check, but otherwise looking good 😄

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -2124,9 +2154,9 @@ class BatchSegment(BaseSegment):
     match_grammar = OneOf(
         # Things that can be bundled
         AnyNumberOf(
-            OneOf(
-                Ref("BeginEndSegment"),
+            Sequence(
                 Ref("StatementSegment"),
```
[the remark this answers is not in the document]
jpers36: Sure.  Since introduction of the BatchSegment, StatementSegments are commands that can be bundled together 1-X times within a batch.  A CREATE PROCEDURE statement is different in TSQL in that it must be the only statement within a batch.  I believe there are other statements which behave similarly but I haven't yet looked for a comprehensive list.

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -2124,9 +2154,9 @@ class BatchSegment(BaseSegment):
...
+                Ref("DelimiterSegment", optional=True),
```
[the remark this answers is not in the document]
jpers36: Nope, not fixed.  The problem there is that we're still relying on the ANSI standard UseStatementSegment which leverages StartsWith in its match_grammar definition.  We have no reliable delimiter in TSQL to split StatementSegments before we initiate the StartsWith match, so StartsWith is too greedy and grabs the semicolon for inclusion within the UseStatementSegment match.  But the parse_grammar has no DelimiterSegment, plus there's no way for the match_grammar to hand the semi-colon back to the parent segment.

The fix for #2029 will be to override ANSI with a TSQL-specific UseStatementSegment that uses ANSI's parse_grammar definition as its match_grammar definition.

src/sqlfluff/dialects/dialect_tsql.py
jpers36: Yes, good catch!

src/sqlfluff/dialects/dialect_tsql.py
jpers36: I just tested and it's possible to have semicolons after ```BEGIN TRY```, ```BEGIN CATCH```, and ```END CATCH```, but not after ```END TRY```.  I've added optional DelimiterSegments after ```BEGIN TRY``` and ```BEGIN CATCH```.  The one after ```END CATCH``` will be handled by the parent segment.

src/sqlfluff/dialects/dialect_tsql.py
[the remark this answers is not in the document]
jpy-git: makes sense, thanks for clarifying 😄

src/sqlfluff/dialects/dialect_tsql.py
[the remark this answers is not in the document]
jpy-git: No worries, thanks for checking 👍

src/sqlfluff/dialects/dialect_tsql.py
jpy-git: perfect thanks for adding those

jpy-git (approved):
Thanks for adding adding this @jpers36, looks great! LGTM 🚀

PR #2052, opened 2021-12-06
Fix moved PROFILES_DIR in dbt 1.0.0

### LINKED ISSUE #2048: dbt v1.0.0
Sorry, very boring enhancement request… [dbt v1.0.0](https://github.com/dbt-labs/dbt-core/releases/tag/v1.0.0) is out, but sqlfluff-templater-dbt v0.8.2 is not compatible with it.

```
ImportError: cannot import name 'PROFILES_DIR' from 'dbt.config.profile'
```

thomasaarholt:
dbt 1.0.0 is fresh of the press and includes a big number of things that should speed up sqlfluff, like --partial-parse by default.

I noticed that this particular setting has changed, so I've suggested a fix.

Fixes #2048.

There may be more things that have changed with dbt 1.0 that I am unaware of.

I'm happy to add a test if deemed necessary.
### Pull Request checklist

  - Other.


plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -11,7 +11,12 @@
 from functools import partial
 
 from dbt.version import get_installed_version
-from dbt.config.profile import PROFILES_DIR
+
+try:
+    from dbt.flags import PROFILES_DIR # dbt >= 1.0
+except:
```
jpy-git, line 17:
```suggestion
except ImportError:
```
don't use bare except

jpy-git (commented):
@thomasaarholt couple linting errors to take care of:
image
(you can check these locally with `tox -e linting,mypy`)

@tunetheweb @barrywhart Think we'll want a step added to the CI to test these dbt-core v1.0.0 vs older version changes as there may well be more and we don't want to break sqlfluff for those using pre v1.0.0 dbt-core. Any opinions on how we do this?

barrywhart:
>Think we'll want a step added to the CI to test these dbt-core v1.0.0 vs older version changes

@jpy-git: Probably. TBH, I don't use dbt (except when developing and testing SQLFluff), so my thoughts on dbt testing may not be particularly helpful. I don't think @tunetheweb uses it, either. Do you use it? This could be a great topic for the next maintainers' meeting -- I know a few of the other maintainers are dbt users.

Is it possible to get version-specific PyPI download stats for dbt? If so, that could provide some guidance when to drop support for older versions of dbt.

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -11,7 +11,12 @@
 from functools import partial
 
 from dbt.version import get_installed_version
-from dbt.config.profile import PROFILES_DIR
+
+try:
```
barrywhart, line 15: Nitpick: Should we consider checking the dbt version rather than blindly trying to import it? It seems more readable and maintainable.

jpy-git: Was thinking this too.

dbt has a version tuple we can get from the package (see below):
https://github.com/sqlfluff/sqlfluff/blob/86b553d49bab5fa4a0bb6fd775802b308b7a9bdd/plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py#L40-L42

You can refer here for an example of using a version tuple to handle imports:
https://github.com/sqlfluff/sqlfluff/blob/86b553d49bab5fa4a0bb6fd775802b308b7a9bdd/src/sqlfluff/__init__.py#L8-L12

jpy-git:
@barrywhart I'm not a dbt user myself either, so let's check with the others 👍  I imagine, what with it being a major version change, there'll be a couple places where changes will be required.

I couldn't see an option for version specific download stats but can have a more in depth look later. According to the GitHub, dbt-core 1.0.0 has only been out 3 days so don't think we can really drop earlier version support anytime soon.

With dbt related issues over the next couple weeks it may be worth checking what version people are using and whether they still experience them when using dbt < 1.0.0. Good to figure out what issues are existing and which are caused by this major version change.

tunetheweb:
> > Think we'll want a step added to the CI to test these dbt-core v1.0.0 vs older version changes
> 
> @jpy-git: Probably. TBH, I don't use dbt (except when developing and testing SQLFluff), so my thoughts on dbt testing may not be particularly helpful. I don't think @tunetheweb uses it, either. Do you use it? This could be a great topic for the next maintainers' meeting -- I know a few of the other maintainers are dbt users.
> 
> Is it possible to get version-specific PyPI download stats for dbt? If so, that could provide some guidance when to drop support for older versions of dbt.

If we're adding v1.0.0-specific stuff we should definitely add that version to `tox.ini` as part of this PR so it's tested. The DBT tests are very light so probably fine to add another one. Though agree it would be nice to drop older versions at some point.

We should also increment `after_n_builds` in `.github/codecov.yml` so it doesn't report back too early.

But other than that advice, I don't use (or know) DBT at all so can't really help beyond that.

tunetheweb:
I'm more curious if DBT 1.0.0 will just work in our CI once we add it. Or if it'll require more code changes to get it to complete.

jpy-git:
let me create a PR adding the tox.ini to update to this PR and we can take a look

jpy-git:
@thomasaarholt I've created a PR (https://github.com/thomasaarholt/sqlfluff/pull/1) to your branch to add the necessary unit tests for dbt 1.0.0. Could you merge that in when you get the chance and then we can trigger the tests to assess whether this PR fixes dbt 1.0.0 or if more work is required. Thanks 😄

jpy-git:
Ahh so we've got 24 failing tests for dbt 1.0.0, although most of them are coming from the fact that that the tests can't find the default profile so thats probably the next thing to look into:
image

There's also a deprecation warning for this package:
image
which is referring to `plugins/sqlfluff-templater-dbt/test/fixtures/dbt/dbt_project/packages.yml`
Think the values just need changing to `dbt-labs/dbt_utils` and 0.8.0
(https://github.com/dbt-labs/dbt-utils)
I can see the deprecation warning on the other dbt versions as well so maybe the deprecation bit can be split into a seperate PR first to check it works with other versions and if so we can get that merged down.

jpy-git:
I've checked if the deprecated `fishtown-analytics/dbt_utils` switch was compatible with the existing codebase and it wasn't a straight switch so closed that PR to be incorporated with the larger dbt 1.0.0 compatibility work.

jpy-git:
@thomasaarholt Given that the dbt 1.0.0 unit tests are failing, it appears that the changes required to achieve compatibility are going to be non-trivial and are probably going beyond the scope of this PR.

I'm conscious that this could be a lot of work, so would you be interested in working on getting the code updated so all the unit tests pass?

Otherwise I think it would be best for us to close this PR and pass it over to one of the maintainers who are more familiar with the `sqlfluff-dbt-templater` codebase to look at when they have time.

Appreciate all your help on this so far though! 😄

jpy-git:
@alanmcruickshank is going to take a look at dbt 1.0.0 compatibility so I'm going to close this PR 👍

PR #2054, opened 2021-12-06
TSQL: INSERT INTO

### LINKED ISSUE #2045: TSQL: INSERT INTO SELECT functionality
https://docs.microsoft.com/en-us/sql/t-sql/statements/insert-transact-sql?view=sql-server-ver15

## Expected Behaviour

The following code parses:
```
BEGIN
  INSERT INTO HumanResources.NewEmployee   
      SELECT EmpID, LastName, FirstName, Phone,   
             Address, City, StateProvince, PostalCode, CurrentFlag  
      FROM EmployeeTemp;  
END
```

## Observed Behaviour
Code does not parse.

jpers36:
Fixes #2045

The ANSI InsertStatementSegment definition uses StartsWith(), which is generally incompatible with the way that we handle TSQL's optional delimitation.  This PR overrides with a TSQL-specific implementation of InsertStatementSegment not using StartsWith(). 

Added a test case proving that this fixes 2045.

Also fixed a trivial typo in ANSI documentation.

### DIFF
--- a/test/fixtures/dialects/tsql/insert_statement.sql
+++ b/test/fixtures/dialects/tsql/insert_statement.sql
@@ -28,3 +28,11 @@ SELECT
 	[SOURCE]
 FROM 
    STAGE.ECDC_CASES
+GO
+
+BEGIN
+  INSERT INTO HumanResources.NewEmployee   
+      SELECT EmpID, LastName, FirstName, Phone,   
+             Address, City, StateProvince, PostalCode, CurrentFlag  
+      FROM EmployeeTemp;  
+END
--- a/test/fixtures/dialects/tsql/insert_statement.sql
+++ b/test/fixtures/dialects/tsql/insert_statement.sql
@@ -36,3 +36,11 @@ BEGIN
              Address, City, StateProvince, PostalCode, CurrentFlag  
       FROM EmployeeTemp;  
 END
+
+GO
+
+INSERT INTO HumanResources.NewEmployee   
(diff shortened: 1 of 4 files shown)


test/fixtures/dialects/tsql/insert_statement.sql
```diff
@@ -28,3 +28,11 @@ SELECT
...
 FROM 
    STAGE.ECDC_CASES
+GO
+
+BEGIN
+  INSERT INTO HumanResources.NewEmployee   
+      SELECT EmpID, LastName, FirstName, Phone,   
+             Address, City, StateProvince, PostalCode, CurrentFlag  
+      FROM EmployeeTemp;  
+END
```
jpy-git, line 38: could we add another test case without the BEGIN END block as well?

```sql
INSERT INTO HumanResources.NewEmployee   
    SELECT
        EmpID,
        LastName,
        FirstName,
        Phone,   
        Address, City,
        StateProvince,
        PostalCode,
        CurrentFlag  
    FROM EmployeeTemp;
```

jpy-git (commented):
extra unit test, otherwise looks good 😄

test/fixtures/dialects/tsql/insert_statement.sql
jpers36: Done!

jpy-git (approved):
Thanks for adding the extra test, the PR is looking good! LGTM! 🥇

PR #2061, opened 2021-12-06
CLI: Add --ignore-local-config flag

jpy-git:

This is a follow up to #1986 (The code changes are near identical, just for a different flag).
This PR adds a `--ignore-local-config` flag to the CLI that serves 2 main purposes:
1. Allow a user to run a command with the default config. Useful for debugging whether issues are config specific.
2. When used in conjunction with the `--config` flag the user is able to use a custom config without the local config override behaviour.

The idea was from a discussion with @tunetheweb when making the original PR.
image

No

### Pull Request checklist

  - Other.


jpy-git:
@barrywhart @tunetheweb this option allows for non-override behaviour when specifying the config via the CLI 👍 

Looking to get this + the config simple API option finished before the next release

tunetheweb (approved):
LGTM!

Does seem a lot of repetition in these config changes. Wonder if there is any refactor possible to avoid having to make so many similar lines to so many places if this changes again in future? But outside of scope for this PR so happy to approve it.

jpy-git:
@tunetheweb yeah it's definitely something to look into. Most of the lines in this and #1986 are because they're specifically to do with loading the config. Most other config changes should be fairly lightweight 😄

PR #2067, opened 2021-12-08
Add basic pre-commit config

jpy-git:

This PR adds a basic pre-commit config to the project. This allows developers to run the linting/typing checks automatically at commit time. I know I'm not the only one who opens a PR only to realise the CI failed because I forgot to run black, so this is a nice convenience 😄

An important thing to note is that this isn't intended to provide extra checks compared to the CI, just as an easy way to automate these checks locally 👍 

I've also added a section to `CONTRIBUTING.md` to let people know how to get the pre-commit hooks setup.

N.B. for those not familiar with pre-commit, it only runs on staged files so is a quick check.

No, it's optional (but very convenient for devs to start using).

### Pull Request checklist

  - Other.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,28 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.0.1
+    hooks:
+    -   id: end-of-file-fixer
+    -   id: trailing-whitespace
+-   repo: https://github.com/psf/black
+    rev: 21.12b0
+    hooks:
+    -   id: black
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.910-1
+    hooks:
+    -   id: mypy
+        args: [--ignore-missing-imports]
+        additional_dependencies: [
+            types-toml,
+            types-pkg_resources,
+            types-chardet,
+        ]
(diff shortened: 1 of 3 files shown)


jpy-git:
@barrywhart @tunetheweb think this is a sensible add to the project 😄

jpy-git:
Also added a tox pre-commit section to simplify the install 😄

barrywhart (approved):
Looks great! I have often used pre-commit locally, even for SQLFluff, but it's cool to make it a standard part of the configuration. 💪🏽

I'll delay merging in case someone else wants to review.

tunetheweb (approved):
LGTM - though must admit I've not used pre-commit before.

Is there any reason NOT to add it to our dev dependencies? Is the main reason to keep it optional? And also avoid CI using this (though rare for that to commit)?

Maybe we merge as is, but in future consider adding to dev dependencies so it's on by default?

jpy-git:
@tunetheweb so because it's in `tox.ini` adding it to `requirements_dev.txt` is redundant + you would still need to run the `pre-commit install` step to set up the hooks.

If you run `tox -e pre-commit -- install` it handles the package install + hook install. You can't install pre-commit hooks by default, the user needs to actively install it. I think the tox method is the cleanest.

The CI doesn't need pre-commit as they're just running the same linting checks, just via different methods.

PR #2069, opened 2021-12-08
Bug fix of L054 for Snowflake and Exasol

tunetheweb:

Fixes GROUP BY grammar type for snowflake and Exasol to be consistent with other dialects and to allow L054 to work on these grammars

The clause does change from type `group_by_clause` to `groupby_clause` so if anyone depends on that using API then they will need to update. But it's more consistent so we should make this change IMHO.

### Pull Request checklist

  - Other.


jpy-git (approved):
Good spot! 💪

PR #2073, opened 2021-12-08
Generate substitution placeholders for Jinja "{{ }}" sections

### LINKED ISSUE #2066: Parse tree: Add placeholders to mark template substitutions
When using Jinja templating the various templating constructs (e.g. if, for loops, etc) are stored in the parse tree as `[META] placeholder`
image

This is very useful when it comes to developing rules as it allows for for us to account for the existence of these in our logic and prevents unintentional rule flagging/potential corruptions.

The DBT templater on the other hand will often omit these constructs (namely `{{ config(materialized='view') }}` and `{% endmacro %}`) from the parse tree. This means that it's impossible for us to properly assess and fix lints such as L009 if the file contains one of these constructs.

e.g. in #2046 a file ends with
```
ORDER BY rl.RunID

{% endmacro %}
```
and parses to
```
[L:  8, P:  1]      |                        identifier:                           'RunID'
[L:  8, P: 44]      |                    [META] dedent:
[L:  8, P:  1]      |    newline:                                                  '\n'
[L:  8, P:  1]      |    newline:                                                  '\n'
[L:  8, P: 44]      |    newline:                                                  '\n'
```
so from the perspective of L009 the file ends with 3 trailing newlines and hence raises the lint error. However, this query should really parse to something like:
```
[L:  8, P:  1]      |                        identifier:                           'RunID'
[L:  8, P: 44]      |                    [META] dedent:
[L:  8, P:  1]      |    newline:                                                  '\n'
[L:  8, P:  1]      |    newline:                                                  '\n'
[L:  8, P:  1]      |    [META] dedent:
[L:  8, P:  1]      |    [META] placeholder:                             [Type: 'macro_end', Raw: '{% endmacro %}']
[L:  8, P: 44]      |    newline:                                                  '\n'
```
This would allow us to properly account for these templating constructs in the rules, like we are able to do when using Jinja.

When encountering this same issue for L052 ([here](https://github.com/sqlfluff/sqlfluff/pull/1840#issuecomment-962682902)) we were able to implement some basic logic to at least abandon the lint/fix if one of these potential invisible dbt constructs were detected, but it's not a full fix as we would have to start implementing this sort of thing in all rules and prevents us from fixing a lot of situations for dbt that would be simple if given the correct parse tree.

## Dialect
ANSI

## Version
SQLFluff 0.8.2 (HEAD of main branch)
Python 3.9.9
Impacts all dbt versions.

## Configuration
Default config

barrywhart:
- [x] Fix the "initial blank line before empty substitution" bug
- [x] Update `TemplateSegment` to allow empty `source_str`, update `substitution_end` instances to specify blank string
- [x] Test where substitution (blank or not) occurs at the beginning and end of a file, also with 1 or 2 blank lines before or after. Add test(s) that rules L050 and L009 work correctly in these cases.
- [x] If the template references an undefined variable, then the placeholder gets double added within the unparseable section.
- [x] Failing L003 tests -- these are mostly failing because L003 tries to indent/align the placeholder blocks. Needs updating to account for these substitution ones (probably just ignore them).
- [ ] Try adding the "empty substitution" segments as a post-processing step. Trying to include them in "source only slices" is affecting position markers and, ultimately, the behavior of `Linter.remove_templated_errors()`, which is breaking one or more tests (details in this [PR comment](https://github.com/sqlfluff/sqlfluff/pull/2073#issuecomment-991778761)).

Fixes #2066

### Pull Request checklist

  - Other.


barrywhart:
Test case:
```
{{- foobar() -}}

{{ generate_table('mytable') }}
```

`.sqlfluff`:
```
[sqlfluff]
templater = jinja
dialect = bigquery
ignore_templated_areas = True

[sqlfluff:templater:jinja]
load_macros_from_path=macros

nothing=""

a_macro_def = {% macro my_macro(something) %}{{something}} + {{something * 2}}{% endmacro %}
generate_table = {% macro generate_table(table) %}
    SELECT * FROM {{table}}
    {% endmacro %}
```

`macros/macro_`.sql`:
```
{% macro foobar() %}{% endmacro %}
```

Parse output:
```
app@1a01a52ad2f7:~/src$ sqlfluff parse test.sql
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    [META] placeholder:                                       [Type: 'substitution_start', Raw: '{{- foobar() -}}']
[L:  1, P:  1]      |    newline:                                                  '\n'
[L:  3, P:  1]      |    [META] placeholder:                                       [Type: 'substitution_end', Raw: ' ']
[L:  3, P:  1]      |    [META] placeholder:                                       [Type: 'substitution_start', Raw: "{{ generate_table('mytable') }}"]
[L:  3, P:  1]      |    statement:
[L:  3, P:  1]      |        select_statement:
[L:  3, P:  1]      |            select_clause:
[L:  3, P:  1]      |                keyword:                                      'SELECT'
[L:  3, P: 32]      |                [META] indent:
[L:  3, P:  1]      |                whitespace:                                   ' '
[L:  3, P:  1]      |                select_clause_element:
[L:  3, P:  1]      |                    wildcard_expression:
[L:  3, P:  1]      |                        wildcard_identifier:
[L:  3, P:  1]      |                            star:                             '*'
[L:  3, P:  1]      |            whitespace:                                       ' '
[L:  3, P: 32]      |            [META] dedent:
[L:  3, P:  1]      |            from_clause:
[L:  3, P:  1]      |                keyword:                                      'FROM'
[L:  3, P:  1]      |                whitespace:                                   ' '
[L:  3, P:  1]      |                from_expression:
[L:  3, P:  1]      |                    [META] indent:
[L:  3, P:  1]      |                    from_expression_element:
[L:  3, P:  1]      |                        table_expression:
[L:  3, P:  1]      |                            table_reference:
[L:  3, P:  1]      |                                identifier:                   'mytable'
[L:  3, P: 32]      |                    [META] dedent:
[L:  3, P:  1]      |    newline:                                                  '\n'
[L:  3, P: 32]      |    [META] placeholder:                                       [Type: 'substitution_end', Raw: ' ']
[L:  3, P: 32]      |    newline:                                                  '\n'
```

jpy-git:
Nice! Let me have a play around with this 😄

src/sqlfluff/core/parser/lexer.py
```diff
@@ -371,6 +394,27 @@ def elements_to_segments(
                     elif source_only_slice.source_idx >= source_slice.start:
                         so_slices.append(source_only_slice)
 
+                # TODO: I don't think this code will generate substitution_start
+                # and substitution_end if a templated block evaluates to the empty
+                # string. Is this actually possible, and if so, is it an issue?
+                end_substitution_slice()
```
jpy-git, line 400: is there a chance where the template variable is the last thing in the file and therefore we don't get back here to close it?

jpy-git: or is that covered by the `end_substitution_slice()` at line 577?

src/sqlfluff/core/parser/lexer.py
```diff
@@ -371,6 +394,27 @@ def elements_to_segments(
                     elif source_only_slice.source_idx >= source_slice.start:
                         so_slices.append(source_only_slice)
 
+                # TODO: I don't think this code will generate substitution_start
+                # and substitution_end if a templated block evaluates to the empty
+                # string. Is this actually possible, and if so, is it an issue?
```
jpy-git, line 399: Curious to see this for `{{ config(materialized='view') }}` in dbt as thats the main example of a block that evaluates to an empty string and I have a feeling it gets excluded from the source slices

jpy-git:
```
[L:  1, P:  1]      |    [META] placeholder:                                       [Type: 'substitution_start', Raw: '{{- foobar() -}}']
[L:  1, P:  1]      |    newline:                                                  '\n'
[L:  3, P:  1]      |    [META] placeholder:                                       [Type: 'substitution_end', Raw: ' ']
```
Not sure why a newline gets inserted between these placeholders given that the macro is defined as `{% macro foobar() %}{% endmacro %}`

Also the use of `{{-` raises an interesting question for formatting as it removes the leading whitespace and newlines before the bracket, which means from the parse tree we would be trying to format the without knowledge of those newlines.

src/sqlfluff/core/parser/lexer.py
```diff
@@ -335,8 +335,31 @@ def elements_to_segments(
...
+            if last_substitution_slice is not None:
+                # Ending an old slice.
+                segment_buffer.append(
+                    TemplateSegment(
+                        pos_marker=PositionMarker.from_point(
+                            source_slice.start,
+                            element.template_slice.start,
+                            templated_file,
+                        ),
+                        source_str=" ",  # Dummy value
```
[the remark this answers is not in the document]
barrywhart: Yes. Currently, there's a check in `TemplateSegment.__init__` which prevents `None` or empty string values. I could remove that, but for now (since this is a prototype or POC), I just used `" "`.

src/sqlfluff/core/parser/lexer.py
barrywhart: 👍 I'll try that now. I tried some of my own blank macros or values, but testing the real dbt thing will be the 🥇 standard.

barrywhart: The dbt test works, too. 🎉🥳

My intuition is that it's _possible_ for a substitution to return a blank value, but I have yet to figure out _how_.

Parse output:
```
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    [META] placeholder:                                       [Type: 'substitution_start', Raw: "{{ config(materialized='view') }}"]
[L:  1, P:  1]      |    newline:                                                  '\n'
[L:  2, P:  1]      |    [META] placeholder:                                       [Type: 'substitution_end', Raw: ' ']
[L:  2, P:  1]      |    newline:                                                  '\n'
[L:  3, P:  1]      |    statement:
[L:  3, P:  1]      |        select_statement:
[L:  3, P:  1]      |            select_clause:
[L:  3, P:  1]      |                keyword:                                      'SELECT'
[L:  3, P:  7]      |                [META] indent:
[L:  3, P:  7]      |                whitespace:                                   ' '
[L:  3, P:  8]      |                select_clause_element:
[L:  3, P:  8]      |                    wildcard_expression:
[L:  3, P:  8]      |                        wildcard_identifier:
[L:  3, P:  8]      |                            star:                             '*'
[L:  3, P:  9]      |            whitespace:                                       ' '
[L:  3, P: 10]      |            [META] dedent:
[L:  3, P: 10]      |            from_clause:
[L:  3, P: 10]      |                keyword:                                      'FROM'
[L:  3, P: 14]      |                whitespace:                                   ' '
[L:  3, P: 15]      |                from_expression:
[L:  3, P: 15]      |                    [META] indent:
[L:  3, P: 15]      |                    from_expression_element:
[L:  3, P: 15]      |                        table_expression:
[L:  3, P: 15]      |                            table_reference:
[L:  3, P: 15]      |                                identifier:                   'foo'
[L:  3, P: 18]      |                    [META] dedent:
[L:  3, P: 18]      |    newline:                                                  '\n'
```

src/sqlfluff/core/parser/lexer.py
barrywhart: Yes, line 577 should cover it. I haven't explicitly tested this behavior, but that's the reason I wrote it this way -- i.e. creating the `end_substitution_slice()` function and calling it both "mid-stream" as well as at the end.

jpy-git:
Interestingly if I insert a single newline at the top of the example file you sent the first place holder disappears
image

but doesn't happen if I don't have the `-` signs in the brackets
image

src/sqlfluff/core/parser/lexer.py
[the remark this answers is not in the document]
jpy-git: makes sense 👍

src/sqlfluff/core/parser/lexer.py
jpy-git: How exactly does it know when to add the end_substitution_slice in? I get that it uses the last_substitution_slice to know it hasn't closed yet, but how does it know that all the substitution contents have been added before it gets added?

src/sqlfluff/core/parser/lexer.py
jpy-git: Nice, that's pretty promising!

barrywhart:
I'll look into the missing placeholder test case you found. My guess is that this is the "empty substitution" case we were wondering about. As currently written, the placeholders are triggered by the contents **after** templating, so if substitution emits no characters, it is "invisible" to this code. Shouldn't be hard to fix -- in fact, the pre-existing code already handles this for other Jinja code like `{% if %}` -- the code about "source-only" slices.

barrywhart:
Looking at the missing placeholder test. This is cool. You can see that the second slice (the `templated` one) has length 0 in the templated file. This suggests my guess about the cause was correct. Looking closer now.
```
(Pdb) pp templated_file.sliced_file
[TemplatedFileSlice(slice_type='literal', source_slice=slice(0, 1, None), templated_slice=slice(0, 0, None)),
 TemplatedFileSlice(slice_type='templated', source_slice=slice(1, 17, None), templated_slice=slice(0, 0, None)),
 TemplatedFileSlice(slice_type='literal', source_slice=slice(17, 19, None), templated_slice=slice(0, 0, None)),
 TemplatedFileSlice(slice_type='templated', source_slice=slice(19, 50, None), templated_slice=slice(0, 23, None)),
 TemplatedFileSlice(slice_type='literal', source_slice=slice(50, 51, None), templated_slice=slice(23, 24, None))]
```

barrywhart:
I've made some progress handling the empty substitution case, but it's turning up a subtle bug (possibly a pre-existing one) that causes the segments to be generated in the wrong order.

Notice below how the substitution segments appear before the newline, but they should appear **after**. (I haven't committed the corresponding code changes yet -- will wait until I have a fix.)

```
[L:  2, P:  1]      |file:
[L:  2, P:  1]      |    [META] placeholder:                                       [Type: 'substitution_start', Raw: '{{- foobar() -}}']
[L:  4, P: 32]      |    [META] placeholder:                                       [Type: 'substitution_end', Raw: ' ']
[L:  4, P: 32]      |    newline:                                                  '\n'
[L:  4, P:  1]      |    [META] placeholder:                                       [Type: 'substitution_start', Raw: "{{ generate_table('mytable') }}"]
```

barrywhart:
I committed and pushed the code to handle empty template substitutions. The bug I mentioned above is still present. I think it's a bug in the `JinjaTracer` class, a different area of the code than this PR. I'll probably work on that separately (new issue & PR).

jpy-git:
Another thought I was to check placeholder templates and it looks like it already adds its own placeholder to the tree:
*test.sql*
```sql
SELECT
    a,
    b,
    c
FROM
    :tablename
```
*.sqlfluff*
```
[sqlfluff]
templater = placeholder

param_style=colon
tablename=sometable
```
On both 0.8.2 and this PR I get this parse tree.
image

It doesn't have a closing tag so maybe we should make that consistent with this work in the future (think its out of scope for this PR but worth highlighting)

barrywhart:
@jpy-git: About this:

>Also the use of `{{-` raises an interesting question for formatting as it removes the leading whitespace and newlines before the bracket, which means from the parse tree we would be trying to format the without knowledge of those newlines.

AFAIK, this should work correctly. I made some improvements to handle Jinja ["whitespace control"](https://jinja.palletsprojects.com/en/3.0.x/templates/#whitespace-control) shortly before adding the `JinjaTracer` class. Together, those two improvements should have gotth us very close to 100% "fidelity" / alignment between actual Jinja behavior and SQLFluff's understanding of that behavior. The "leading newline" bug you discovered is the first discrepancy that I'm aware of, and I definitely plan to look into that soon!!

barrywhart:
@jpy-git: About this:
>Another thought I was to check placeholder templates and it looks like it already adds its own placeholder to the tree

From the info you provided, this looks like a bug. It's creating a `block_start` segment, which (IIRC) is intended for block-oriented control structures like Jinja's `{% if %}` and `{% for %}` blocks. In a sense, two things need to happen:
* If we **weren't** doing this "substitution" work, the placeholder template probably shouldn't create **any** placeholder segments, since these are substitutions, not blocks. It certainly shouldn't be creating a `block_start` without a `block_end`.
* Since we **are** doing this "substitution" work, the placeholder template should create `substitution_start` and `substitution_end` segments, i.e. the same as Jinja. The design of SQLFluff in this area is a little messy, in that 90% of the time, SQLFluff is used with dbt or Jinja, which are both Jinja underneath. Other templaters in SQLFluff are often not that well tested, and it's unclear whether they can have their own block types or must use the same names as the Jinja and dbt templaters. (In practice, the latter is true, but is that "good" design?)

jpy-git:
@barrywhart so my question on `{{-` was more around whether it's beneficial to us to acknowledge Jinja whitespace control in SQLFluff. 

For example:
We have a file with 100 leading newlines and then an element like `{{- some_variable -}}`. We then render this template and get a parse tree with just the template variable, but none of the leading 100 newlines. We then lint this and say there's no formatting issue with the file. We then replace the template and end up with a "linted" file still with 100 newlines. My question is: "From the user perspective is that a cleanly formatted file?"

The user never sees the rendered whitespace when using SQLFluff so why should we respect it? My opinion would be replace instances of `{{-` with `{{` when parsing as parse-wise it makes no real difference and will allow us to properly asses the formatting of these templating blocks and produce in a nicely formatted output file for the user.

They can still have the `{{-` there for when they go and use that template, but from the formatting perspective I think respecting it within SQLFluff is actually unhelpful for producing a nice looking file for review?

Just a thought 😄

barrywhart:
@jpy-git: About this:
>Also looking at the failing tests it looks like they're mostly coming from L003. It already has some logic to try and align the existing placeholder blocks so will probably need to add some extra logic to account for these substitution ones. I can investigate further after we do the 0.9.0 release.

If you have time to help, that's really great -- thanks! Feel free to create PRs against this PR and we'll review and merge.

Please don't feel any pressure, though -- if you're busy or would rather do other things, I'll work through this stuff. Now that we've gotten into this work and are seeing some of the issues popping up, this feels like about 2 person weeks of part-time work. Not trivial, but not huge, either.

I've gone through your notes and tried to capture them in a "TODO" checklist in the PR description. That should help ensure we don't forget anything.

barrywhart:
@jpy-git: Re this:
>my question on `{{-` was more around whether it's beneficial to us to acknowledge Jinja whitespace control in SQLFluff.

It's a good point. There's a lot of nuance here, in the sense that:
* The recent work on whitespace control got us to the point where the behavior "made sense", i.e. it was explainable.
* SQLFluff rules operate on the parse tree. That's mostly "post template" code, although with placeholders, we're gradually exposing more of the "pre-templated" code as well.

You make a good point. I _think_ the right broader goal is for SQLFluff to provide enough information that rules can operate the way they want: i.e. they can operate purely on the rendered code, purely on the raw code, or a mix. L003 is an example of a rule that does some of both: it indents both raw Jinja tags as well as rendered code.

In order to achieve what you're suggesting, can you start by thinking about:
* Is that information available at present? (Probably yes, at least in a clunky way through the `raw_sliced` and `sliced_file` properties on the `TemplatedFile` object.)
* Is it available in a way that's friendly for rule writers? Perhaps not. If not, how should it be provided? Through Python functions? By adding more stuff to the parse tree, similar to this substitution placeholder work? Thus far, we seem to be doing okay by adding more info to the parse tree, and it hasn't been too disruptive for existing rules. Will that always be the case? If it starts becoming disruptive, should we consider other approaches that are less disruptive to existing rules? Any ideas there?

I don't have firm ideas or opinions here. I think this is a pretty interesting area of the project, because if we do this well, we can continue building more rules, and more **useful** rules.

Another area I'm personally interested in is, what's the potential for building rules that have a deeper understanding of SQL? The `SelectCrawler` class and the associated L044 and L045 rules are one area where we've made some initial progress. Is it possible to do more? I think so!

What are some good use cases for making further progress in this area? At my previous job (which I quit in November), I prototyped a rule that would enforce security restrictions (privacy filtering) on queries. That's an interesting area. Perhaps there are other similar things, maybe involving column-level lineage.

jpy-git:
@barrywhart Re: the whitespace discussion. I'll create an issue to capture these thoughts and can try and expand on the idea using your suggestions! Just got me thinking as I hadn't encountered how the `{{-` was treated in SQLFluff before 😄

jpy-git:
> I've gone through your notes and tried to capture them in a "TODO" checklist in the PR description. That should help ensure we don't forget anything.

Awesome thanks, just to capture my initial thoughts on the L003 issue from a cursory glance:

- L003 searches for the first non-whitespace (or indent) element on a line. It then looks to see if there was an indent earlier on the line to determine whether or not there should be an indent there.
- The motivation for not excluding all placeholders is that it tries to align jinja if/for/comment blocks etc.
- For a non-substituted query there is an indent before the first non-whitespace (from_expression_element) so no lint raised as expected:
image
- For a substituted query the from_expression and its indent are within the substitution placeholder (correctly so), but the I believe the rule is detecting the substitution placeholder as the first non-whitespace in the line (as that's how it would handle correctly handle a block) and since that doesn't have an indent in front it gets dedented.
image
- Basically think we just need to ignore substituted placeholders from the logic 👍

jpy-git:
> From the info you provided, this looks like a bug. It's creating a `block_start` segment, which (IIRC) is intended for block-oriented control structures like Jinja's `{% if %}` and `{% for %}` blocks

makes sense, I think maybe add it to the checklist to look at once we have this working for Jinja/dbt as it would be good to have consistency from a rule standpoint 👍

barrywhart:
@jpy-git: I created a separate PR for the issue with `PlaceholderTemplater` creating `block_start` slices. The fix is simple -- the slice type should be `templated` instead, matching the behavior of all the other templaters. With this change, the substitution PR will now see the `templated` slices and do the right thing.

#2085

barrywhart:
@jpy-git: I think I got the "initial blank line" issue fixed. It wasn't actually a bug in `JinjaTracer`, it was a subtle bug in how the new code in `elements_to_segments()` was detecting and handling empty vs non-empty "templated" (substitution) slices.

Hopefully (and this is a big 🤞), that's the last **tricky** bug in this PR. Still a few days work left, I think, but perhaps no more 🤯, not even a little 🧨.

barrywhart:
Just a few remaining test failures now.

At least one of them is tricky. A test is failing with an L039 error, when there shouldn't be one. The rule itself is behaving the same as it did in `main`, but after the error is returned, `Linter.remove_templated_errors()` is filtering it out in `main` but not in the PR.

For some reason, `PositionMarker.is_literal()` is returning `False` in `main` but `True` in the PR for the "anchor" of the L039 error. That anchor segment is either inside or close to a templated region, so I think the newly added segments are impacting that somehow. The variable in the templated region is undefined, which _may_ also have an impact.

barrywhart:
@jpy-git: I've hit a big complication in this idea and wanted to get your thoughts.

Here's a template (source and rendered) from one of the tests. Note that the variable `col_a` is undefined or blank.

```
(Pdb) templated_file.source_str
'\n    SELECT {{ col_a }} AS a -- noqa: TMP,PRS\n    FROM foo;\n   
(Pdb) templated_file.templated_str
'\n    SELECT  AS a -- noqa: TMP,PRS\n    FROM foo;\n        '
(Pdb) pp elements
[TemplateElement(raw='\n', template_slice=slice(0, 1, None), matcher=<RegexLexer: newline>),
 TemplateElement(raw='    ', template_slice=slice(1, 5, None), matcher=<RegexLexer: whitespace>),
 TemplateElement(raw='SELECT', template_slice=slice(5, 11, None), matcher=<RegexLexer: code>),
 TemplateElement(raw='  ', template_slice=slice(11, 13, None), matcher=<RegexLexer: whitespace>),
 TemplateElement(raw='AS', template_slice=slice(13, 15, None), matcher=<RegexLexer: code>),
 TemplateElement(raw=' ', template_slice=slice(15, 16, None), matcher=<RegexLexer: whitespace>),
 TemplateElement(raw='a', template_slice=slice(16, 17, None), matcher=<RegexLexer: code>),
 TemplateElement(raw=' ', template_slice=slice(17, 18, None), matcher=<RegexLexer: whitespace>),
 TemplateElement(raw='-- noqa: TMP,PRS', template_slice=slice(18, 34, None), matcher=<RegexLexer: inline_comment>),
 TemplateElement(raw='\n', template_slice=slice(34, 35, None), matcher=<RegexLexer: newline>),
```

I have been working on modifying `elements_to_segments()` to add markers to mark substitutions. Note the `elements` array above. These are tokens from the lexer. `elements[3]` is a string with two spaces. Although those two spaces are one _token_, they are not contiguous in the raw source -- they are separated by the empty `{{ col_a }}` substitution. There's no "correct" place to add the empty `substitution_start` and `substitution_end` tokens -- we'd have to split the existing whitespace token to do so. Also note, this could happen with things other than whitespace. Imagine a template that does this:
```
SELECT
    c{{- col_name_middle -}n
```

If `col_name_middle` is blank, the column name is `cn`, and we definitely can't split that into two strings `c` and `n` without breaking things downstream -- i.e. if the parser sees two strings, that's not going to parse correctly as a column name.

Any thoughts on this?Now that I think about it, non-empty substitutions will have the same issue. In the same example above, the substitution provides the middle part of the column name token, not an entire token. Any design that assumes templating provides individual tokens won't work.

In general, template substitutions and tokens have a "many to many" relationship, i.e. a single substitution can provide:
* No tokens
* Part of a token: As above, or even weirder: `SELECT c{{- tp1 -}}d{{- tp2 -}}e{{- tp3 -}}e FROM foo`
* One token: `SELECT {{ col_name }}`
* Many tokens: ``SELECT * FROM foo {{ where_clause }}` where `where_clause` is "WHERE c1 < 10"

jpy-git:
@barrywhart interesting points!

No tokens:
Is this not a similar possibility for the `{{ config(materialized='view') }}` that triggered the thinking around this PR. So I know we verified that we are now getting substitution tokens for something like `\n{{ config(materialized='view') }}\n`, but does your point also mean we aren't getting a substitution token for something like `\s{{ config(materialized='view') }}\s`?
I feel like we would want to have that whitespace split to be able to asses it for removal in rules like L001, L009, L050?

One token:
Think this one's fine as we just put substitution tokens before and after. i.e. `SELECT <sub_start>my_col<sub_end>`

Many tokens:
Same as one token. i.e. `SELECT * FROM foo <sub_start>WHERE c1 < 10<sub_end>`

Part of token:
This one's interesting. My initial reaction is if a token contains a template within it then the substitution placeholders should expand to contain the full token. Reasoning being the same as why we would consider the type of that token to be templated.
If we had an empty string and got `SELECT cn` I would consider `cn` templated and not try to apply any rules to it and risk messing up the contained template, and that template could even contain multiple tokens/spaces/newlines/logic etc so should be off limits. So I would expect the substitution placeholders in this case to be something like `SELECT <sub_start>cn<sub_end>` or `SELECT <sub_start>can<sub_end>` if we subbed in `a`.
Same would apply if it intersected multiple tokens. i.e. `SELECT foo FR{{- something -}}ar` could evaluate to something like `SELECT foo <sub_start>FROM bar<sub_end>.

We can view the purpose of these substitution placeholders as a way of encoding the bounds of applicable segments for rules in the parse tree rather than necessarily a 1-to-1 mapping to it's position in the raw?

barrywhart:
How would you handle this?

```
SELECT abc<sub_start>my_col_part1<sub_end>def<sub_start>my_col_part2<sub_end>ghi<sub_start>my_col_part2<sub_end>jkl
```

The column name is one token. It starts and ends with a literal, it contains 3 substitutions with literals in between. So the lexer token boundaries don't align with the template boundaries at all. Essentially, the templater output string is sliced completely differently, depending on whether you're looking at it from a "template slice" perspective versus a "lexer/parser segment" perspective.

95% of the time, templates are used in nice, simple ways, but it's the question of what to do in the tricky cases that potentially makes it unworkable as a general approach.

Perhaps we could design a partial solution that doesn't cover every case, but covers some of the more common cases.  I think the most common cases are:
* One substitution -> one or more entire tokens
* Empty substitution with whitespace before and after
* One token that's a mixture of literals and substitutions

If the parse tree provided markers for some of these, but ignored the more complex cases, would that provide enough benefit for rule writers (i.e. making it significantly simpler to write those rules) that it's worthwhile? Also, is there potentially another way to provide this information other than new segments? For example, a function that tells you if a parse segment includes substitutions. (Let your imagination run wild here, mostly just pointing out there may be other approaches, and if we avoid changing the parse structure, we have a lot more freedom and a lot less risk of tricky bugs affecting other areas of the code.)

Note that the "full" information will always be available through `TemplatedFile`.

jpy-git:
Well I was just thinking that was `SELECT <sub_start>abcmy_col_part1defmy_col_part2ghimy_col_part2jkl<sub_end>`? I don't think any rule would ever care deeper than does this token contain a template?

I think in a way that is what `.is_template` already does, since in the example above the segment for `abcmy_col_part1defmy_col_part2ghimy_col_part2jkl` would likely have `segment.is_templated == True`.

barrywhart:
What if we provided a new function(s)  that could handle this case?

trim_left()
Input: a whitespace segment that is made up of multiple templated slices
Action: Splits the segment into slices, removes any leading whitespace literal slices
Output: New segment(s) that can be passed to LintFix to remove the leading whitespace

We could do the same for trailing whitespace.

Thoughts? Alternatives?

jpy-git:
@barrywhart think that's fine given it covers the most likely L052 and L009 use cases IF the invisible sub is within whitespace, BUT what about `\n{{ config(materialized='view') }}\n`?
The invisible sub is between the newlines and wouldn't get seen if we had that at the start of the file?

That's why in the [original check](https://github.com/sqlfluff/sqlfluff/blob/8ccb55943d539278840296bcca9a6aad9baec670/src/sqlfluff/rules/L050.py#L65) in L050 I check over a range of whitespace/newline segments if there's a gap in the raw slices hinting at the missing sub.

So maybe the function should be similar to what you describe but can be a method on the base rule instead and take in a list of segments. It would return the segments before the first detected template (splitting the whitespace segment if that's where it's found)?

barrywhart:
I can't recall, what does `\n{{ config(materialized='view') }}\n` look like in the parse tree? I was assuming it was a single whitespace segment, but is it actually several segments, a combination of whitespace + newlines?

Your idea sounds good. (I'm not partial to any particular idea, just suggesting ideas to hopefully trigger discussion.)

barrywhart:
@jpy-git: It sounds like we're moving towards making parts of your L050 code more generic and moving them to `BaseRule`. Would you like me to do that, or would you like to? I'm happy either way, and having two people collaborate on code often leads to a better design. Let me know! I'm planning to remain unemployed for at least another month, so I should have some time available. 🙏🏽🤪

jpy-git:
@barrywhart I'll take a look and create the PR 😄

barrywhart:
This was an experiment, and it turns out the core concept is flawed. We're thinking about other approaches to simplify rules that don't involve changing the parse tree.

PR #2079, opened 2021-12-09
dbt 1.0.0 compatability

### LINKED ISSUE #2048: dbt v1.0.0

alanmcruickshank:
Fixes #2048.

- New test cases for dbt 1.0.0 are in.
- Fix bug with `PROFILES_DIR` location.
- Fix test fixture for different output with new dbt_utils
- Fix test for database connection failure which now raises an exception


alanmcruickshank:
I think the current error comes from trying to set the profiles dir using:
```
self.dbt_config = DbtRuntimeConfig.from_args(
            DbtConfigArgs(
                project_dir=self.project_dir,
                profiles_dir=self.profiles_dir,
                profile=self._get_profile(),
            )
        )
```

I'm not sure how we should do that instead though. I'll set up a thread on the dbt slack.

NiallRees:
> I think the current error comes from trying to set the profiles dir using:
> 
> ```
> self.dbt_config = DbtRuntimeConfig.from_args(
>             DbtConfigArgs(
>                 project_dir=self.project_dir,
>                 profiles_dir=self.profiles_dir,
>                 profile=self._get_profile(),
>             )
>         )
> ```
> 
> I'm not sure how we should do that instead though. I'll set up a thread on the dbt slack.

https://github.com/sqlfluff/sqlfluff/pull/2079/files#diff-58e23bef1bd10006044251ce2394cbd69d284b4881be6929fe48dd2b509f5eb1R92-R96 has solved that one!

barrywhart:
I have a likely fix for the dbt_utils version issue. I'll be creating a new PR (building off this PR) momentarily...

plugins/sqlfluff-templater-dbt/test/templater_test.py
```diff
@@ -70,10 +72,24 @@ def test__templater_dbt_templating_result(
         fname=os.path.join(project_dir, "models/my_new_project/", fname),
         config=FluffConfig(configs=DBT_FLUFF_CONFIG),
     )
-    assert (
-        str(templated_file)
-        == open("plugins/sqlfluff-templater-dbt/test/fixtures/dbt/" + fname).read()
+    template_output_folder_path = (
```
barrywhart, line 75: This code is pretty complex -- I have a hard time following it. Could we make the additions clearer, perhaps with some comments?

plugins/sqlfluff-templater-dbt/test/templater_test.py
```diff
@@ -70,10 +72,24 @@ def test__templater_dbt_templating_result(
...
+        "plugins/sqlfluff-templater-dbt/test/fixtures/dbt/templated_output/"
     )
+    dbt_version_specific_fixture_folder = {(1, 0): "dbt_utils_0.8.0"}.get(
+        DBT_VERSION_TUPLE
+    )
+    fixture_path = None
+    if dbt_version_specific_fixture_folder:
+        version_specific_path = (
+            Path(template_output_folder_path)
```
barrywhart, line 84: I notice we're mixing use of `Path` and `os.path` in the same function. Think we could make it consistent within the function?

plugins/sqlfluff-templater-dbt/test/templater_test.py
```diff
@@ -70,10 +72,24 @@ def test__templater_dbt_templating_result(
...
+            / dbt_version_specific_fixture_folder
+            / fname
+        )
+        if os.path.exists(version_specific_path):
+            fixture_path = version_specific_path
+    if not fixture_path:
+        fixture_path = template_output_folder_path + fname
+    assert str(templated_file) == open(fixture_path).read()
```
barrywhart, line 92: Following up the above comment about `Path` vs `os.path`, if we were using `Path` here, we could use [the `read_text()` function](https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text), which opens, reads, and closes the file in one step.

plugins/sqlfluff-templater-dbt/test/templater_test.py
```diff
@@ -296,6 +311,49 @@ def test__templater_dbt_handle_exceptions(
...
+    # as dbt throws an error if a node fails to parse while computing the DAG
+    os.rename(src_fpath, target_fpath)
+    try:
+        _, violations = dbt_templater.process(
+            in_str="",
+            fname=target_fpath,
+            config=FluffConfig(configs=DBT_FLUFF_CONFIG),
+        )
+    except Exception as e:
+        if DBT_VERSION_TUPLE == (1, 0):
```
barrywhart, line 336: Seems like this might be more "future proof":
```suggestion
        if DBT_VERSION_TUPLE >= (1, 0):
```

plugins/sqlfluff-templater-dbt/test/templater_test.py
```diff
@@ -296,6 +311,49 @@ def test__templater_dbt_handle_exceptions(
...
+            # In dbt 1.0.0, connection failures raise an exception
+            assert str(e).startswith(
+                "Runtime Error\n  connection never acquired for thread"
+            )
+        else:
+            raise (e)
+    finally:
+        get_adapter(dbt_templater.dbt_config).connections.release()
+        os.rename(target_fpath, src_fpath)
+    if DBT_VERSION_TUPLE != (1, 0):
```
barrywhart, line 346:
```suggestion
    if DBT_VERSION_TUPLE < (1, 0):
```

barrywhart (dismissed):
A few small change suggestions and thoughts about style. Let me know what you think -- I don't want to be so picky that we delay getting 1.0 compatibility. Also ping me if you'd like me to address some of the suggested changes, since I have a time zone advantage at this time of day. 😊

NiallRees:
> A few small change suggestions and thoughts about style. Let me know what you think -- I don't want to be so picky that we delay getting 1.0 compatibility. Also ping me if you'd like me to address some of the suggested changes, since I have a time zone advantage at this time of day. 😊

Please do if you feel like it! Calling it a night here!

barrywhart:
@NiallRees: Ok, the "test tweaks" PR is ready for review. #2088

NiallRees (approved):

Peeyush-Now:
Is this still broken. trying to use sqlfluff with jinja templater but still complains about not finding "cannot import name 'PROFILES_DIR' from 'dbt.config.profile'"

PR #2080, opened 2021-12-09
Simple API config path

### LINKED ISSUE #1274: Add a parameter to provide a .sqlfluff configuration file when using the .fix or .lint API methods

jpy-git:

Fixes #1274. Builds upon my CLI config path additions #1986 and #2061.
Adds an option to pass a config_path to the Simple API. This doesn't have the same override behaviour as the CLI as it doesn't make sense in the context of the API. The config_path specified is the config that is used.

Also changes the API to create a config object rather than passing its args directly to the linter. This should help decouple the linter and the API and also make it easy to update the simple API with additional arguments from the CLI in the future.

No

### Pull Request checklist

  - Other.


src/sqlfluff/api/simple.py
```diff
@@ -38,7 +77,13 @@ def lint(
...
         :obj:`list` of :obj:`dict` for each violation found.
     """
-    linter = Linter(dialect=dialect, rules=rules, exclude_rules=exclude_rules)
+    cfg = get_simple_config(
+        dialect=dialect,
+        rules=rules,
+        exclude_rules=exclude_rules,
+        config_path=config_path,
+    )
+    linter = Linter(config=cfg)
```
jpy-git, line 86: we can hopefully remove those dialect/rules/exclude_rules arguments to the linter in the future as they should be part of the config. i.e. decouple linter and simple API 👍

jpy-git:
@barrywhart @tunetheweb config_path option for the Simple API 😄

src/sqlfluff/api/simple.py
tunetheweb: That would mean you'd have to have a config file to pass. Might be overkill for some simple options. For example I presume https://online.sqlfluff.com/ uses the simple API and just needs to specify the dialect.

But it is duplication which is annoying.

BTW what happens if you specify a dialect AND a config file? Which ones wins out? Or should that error and it should be either/or but not both?

tunetheweb:
Any idea what's up with docs as the parameter name doesn't appear (and also not for exclude_rules):

image

It appears fine on production for `exclude_rules`:

image

src/sqlfluff/api/simple.py
jpy-git: @tunetheweb you can still pass a dialect to the Simple API (and the dialect would overwrite that in the config as you'd expect)? 
The Simple API itself is unchanged from a user perspective:
```python
res = sqlfluff.lint(
    dialect="snowflake",
    rules=["L001", "L002"],
    config_path="some/path/to/config/.sqlfluff"
)
```

I'm specifically talking about `Linter` in the comment which doesn't need those arguments 😄

jpy-git: as in we don't need to be able to do `Linter(dialect="snowflake")` since the simple api was the only thing using that. Everything else does `Linter(config=cfg)`.

tunetheweb: Ah Gothca.

Yes In agree explicit dialect or rule should override config, so that's good, but I wonder if we need to expose this in the docs though so others know that's the way it's implemented?

jpy-git:
@tunetheweb Interesting let me look at that. Also need to see where I'm missing coverage.

jpy-git:
@tunetheweb updated the docstrings if you want to take another look?

Still working on coverage in the meantime

tunetheweb:
Docs look better now.

src/sqlfluff/api/simple.py
```diff
@@ -1,11 +1,49 @@
...
+    config_path: Optional[str] = None,
+) -> FluffConfig:
+    """Get a config object from simple API arguments."""
+    # Check the requested dialect exists and is valid.
+    try:
+        dialect_selector(dialect)
+    except SQLFluffUserError as err:
+        raise SQLFluffUserError(f"Error loading dialect '{dialect}': {str(err)}")
+    except KeyError:
+        raise SQLFluffUserError(f"Error: Unknown dialect '{dialect}'")
```
tunetheweb, line 27: Suggest you just Pragma no cover these:

```suggestion
    except SQLFluffUserError as err:  # pragma: no cover
        raise SQLFluffUserError(f"Error loading dialect '{dialect}': {str(err)}")
    except KeyError:  # pragma: no cover
        raise SQLFluffUserError(f"Error: Unknown dialect '{dialect}'")
```

tunetheweb (commented):
Found 4 of your lines.

The other two lines are due to this in config.py:

image

Guess the simple  api tests were covering that before, but not you instantiate a config object with those overrides so no longer directly call this.

So guess you need to either remove that functionality completely or replace that test coverage with complex API tests?

src/sqlfluff/api/simple.py
jpy-git: thanks I've made a test for the second error, but will use pragma for the first one as that's not easy to find a test for 😄 

I'm just investigating whether we can remove those config.py lines at this point 👍

tunetheweb: I wonder if passing an incorrect file location will trigger the first?

But I’m absolutely fine with pragma no covers for exceptions. They are, by definition, for unexpected cases. Of course if it’s easy to add a case then do, but not essential.

But I just like to keep the 100% cover for a few reasons:
1. Makes it obvious when you do miss cover you should have!
2. Removes noise when % rounding errors look like changes that aren’t cause total number of lines change but number of not cover lines don’t.
3. I’m a perfectionist! 😀

But the pragma no cover is fine to use to achieve that if it’s the right call rather than killing yourself making a complex test case that adds more complexity than real benefit.

PR #2081, opened 2021-12-09
Grammar: adds support for CREATE/ALTER/DROP DATABASE for Postgres dialect

### LINKED ISSUE #2017: Alter Database Unparsable Section
## Expected Behaviour
Using the postgres [ALTER DATABASE](https://www.postgresql.org/docs/14/sql-alterdatabase.html) should work.

## Observed Behaviour
```bash
$ sqlfluff lint --dialect postgres alter.sql
== [alter.sql] FAIL                                                          
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'ALTER
                       | DATABASE postgres SET password_enc...'
All Finished 📜 🎉!
```

## Steps to Reproduce

**alter.sql**:

```sql
ALTER DATABASE postgres SET password_encryption TO 'scram-sha-256';
```

```bash
sqlfluff lint --dialect postgres alter.sql
```

## Dialect
`postgres`

## Version
`sqlfluff, version 0.8.2`
`Python 3.8.10`
`pip 20.0.2`

## Configuration
`--dialtect postgres`

derickl:
Adds full support for [CREATE | ALTER | DROP] DATABASE statements.

Also fixes #2017 

None

### Pull Request checklist

  - Other.


jpy-git (commented):
@derickl a couple things for CREATE/DROP whilst I'm still reviewing ALTER 👍

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1555,6 +1555,159 @@ class DropMaterializedViewStatementSegment(BaseSegment):
...
+                Sequence("WITH", optional=True),
+                AnyNumberOf(
+                    Sequence("ALLOW_CONNECTIONS", Ref("BooleanLiteralGrammar")),
+                    Sequence(
+                        "CONNECTION",
+                        "LIMIT",
+                        Ref("NumericLiteralSegment"),
+                    ),
+                    Sequence("IS_TEMPLATE", Ref("BooleanLiteralGrammar")),
+                ),
```
jpy-git, line 1660:
```suggestion
                AnyNumberOf(
                    Sequence("ALLOW_CONNECTIONS", Ref("BooleanLiteralGrammar")),
                    Sequence(
                        "CONNECTION",
                        "LIMIT",
                        Ref("NumericLiteralSegment"),
                    ),
                    Sequence("IS_TEMPLATE", Ref("BooleanLiteralGrammar")),
                    min_times=1,
                ),
```
slight nuance here, we can have any number of these but we must have at least one of them so we use the min_times argument to AnyNumberOf

jpy-git (commented):
A couple more suggestions for ALTER DATABASE.

It's looking great so far! 🙌

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1555,6 +1555,159 @@ class DropMaterializedViewStatementSegment(BaseSegment):
...
+
+    type = "create_database_statement"
+
+    match_grammar = StartsWith(Sequence("CREATE", "DATABASE"))
+
+    parse_grammar = Sequence(
+        "CREATE",
+        "DATABASE",
+        Ref("DatabaseReferenceSegment"),
+        Sequence("WITH", optional=True),
```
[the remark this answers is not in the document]
derickl: I've grep'd through the source code and haven't come across this usage pattern `Ref("WITH", optional=True)` but `Ref.keyword("WITH", optional=True)` does exist.

When does one choose `Ref.keyword` over `Sequence`? Would `Ref.keyword` be for `reserved` keywords?

derickl:
Based on this https://github.com/sqlfluff/sqlfluff/pull/2081#discussion_r766207381

I've grep'd through the source code and haven't come across this usage pattern `Ref("WITH", optional=True)` but `Ref.keyword("WITH", optional=True)` does exist. 

What would be the difference between `Ref("WITH", optional=True)` and `Sequence("WITH", optional=True)`?

When does one choose `Ref.keyword` over `Sequence`? Would `Ref.keyword` be for `reserved` keywords?

derickl:
Thank you for the review @jpy-git!

tunetheweb:
> Based on this [#2081 (comment)](https://github.com/sqlfluff/sqlfluff/pull/2081#discussion_r766207381)
> 
> I've grep'd through the source code and haven't come across this usage pattern `Ref("WITH", optional=True)` but `Ref.keyword("WITH", optional=True)` does exist.
> 
> What would be the difference between `Ref("WITH", optional=True)` and `Sequence("WITH", optional=True)`?
> 
> When does one choose `Ref.keyword` over `Sequence`? Would `Ref.keyword` be for `reserved` keywords?

A good question and one I’ve wondered in the past!

As far as I can tell there is no real difference and it comes down to personal preference.

In theory using `Ref.keyword` might be slightly faster as the Sequence version effectively wraps the implied `Ref.keyword` in a (short) `Sequence` So it’s an extra layer to go through. In practice I would be VERY surprised if that created any real difference.

Personally I have a slight preference for Sequence (despite the above) for a few reasons:
- it seems more recognisable as we use Sequenxe a lot.
- It allows other grammar to more easily be added to this section (e.g. if you want to add a `NOT` before it).
- It two less characters and all one word so easier to type!

But honestly it doesn’t make a big difference. Maybe we should standardise on a preferred style but when I looked before they were both used quite a bit and I’ll bet it’s got worse since then.

test/fixtures/dialects/postgres/postgres_create_database.sql
```diff
@@ -0,0 +1,28 @@
...
+CREATE DATABASE db OWNER = user_name;
+CREATE DATABASE db WITH OWNER = user_name;
+CREATE DATABASE db TEMPLATE template_name;
+CREATE DATABASE db TEMPLATE = template_name;
+CREATE DATABASE db WITH TEMPLATE = template_name;
+CREATE DATABASE db ENCODING 'UTF8';
+CREATE DATABASE db ENCODING = 'UTF8';
+CREATE DATABASE db LOCALE 'en_US.UTF-8';
+CREATE DATABASE db LOCALE = 'en_US.utf8';
+CREATE DATABASE db LC_COLLATE 'en_US.UTF-8' LC_CTYPE 'en_US.UTF-8';
```
jpy-git, line 12: could we have a unit test with the order of `LC_COLLATE` and `LC_CTYPE` switched to cover the point we discussed in the previous review?

test/fixtures/dialects/postgres/postgres_create_database.sql
```diff
@@ -0,0 +1,28 @@
...
+CREATE DATABASE db TABLESPACE = new_tablespace;
+CREATE DATABASE db WITH TABLESPACE = new_tablespace;
+CREATE DATABASE db ALLOW_CONNECTIONS true;
+CREATE DATABASE db ALLOW_CONNECTIONS = true;
+CREATE DATABASE db CONNECTION LIMIT 10;
+CREATE DATABASE db CONNECTION LIMIT = 10;
+CREATE DATABASE db IS_TEMPLATE true;
+CREATE DATABASE db IS_TEMPLATE = true;
+CREATE DATABASE db IS_TEMPLATE = true CONNECTION LIMIT 10 ALLOW_CONNECTIONS = true;
+CREATE DATABASE db WITH IS_TEMPLATE = true CONNECTION LIMIT 10 ALLOW_CONNECTIONS = true;
```
jpy-git, line 28: Wonder if we should have one test case with all of the options specified to really test the grammar?
It should work but good to check 😄

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1555,6 +1555,166 @@ class DropMaterializedViewStatementSegment(BaseSegment):
...
+
+    match_grammar = StartsWith(Sequence("ALTER", "DATABASE"))
+
+    parse_grammar = Sequence(
+        "ALTER",
+        "DATABASE",
+        Ref("DatabaseReferenceSegment"),
+        OneOf(
+            Sequence(
+                Sequence("WITH", optional=True),
```
jpy-git, line 1653:
```suggestion
                Ref.keyword("WITH", optional=True),
```
To be consistent with CREATE/DROP

test/fixtures/dialects/postgres/postgres_alter_database.sql
```diff
@@ -0,0 +1,29 @@
+ALTER DATABASE db;
+ALTER DATABASE db ALLOW_CONNECTIONS true;
+ALTER DATABASE db WITH ALLOW_CONNECTIONS true;
+ALTER DATABASE db CONNECTION LIMIT 10;
+ALTER DATABASE db WITH CONNECTION LIMIT 10;
+ALTER DATABASE db IS_TEMPLATE true;
+ALTER DATABASE db WITH IS_TEMPLATE true;
+ALTER DATABASE db IS_TEMPLATE true ALLOW_CONNECTIONS true;
+ALTER DATABASE db WITH IS_TEMPLATE true ALLOW_CONNECTIONS true;
```
jpy-git, line 9: Maybe a test case with ALLOW_CONNECTIONS, CONNECTION LIMIT, and IS_TEMPLATE all specified together?

jpy-git (commented):
Looks good, couple final things to add then we should be good to get this going 👍

jpy-git:
@derickl good spot on the `Ref` vs `Ref.keyword`! 

Basically Ref.keyword is the syntax for explicitly specifying to search for the string in the keywords. You can actually think of any of the keywords as being wrapped by Ref.keyword
e.g.
```sql
parse_grammar = Sequence(
    Ref.keyword("CREATE"),
    Ref.keyword("DATABASE"),
...
```
but for shorthand you can exclude `Ref.keyword` in these cases. 

However, in the specific example of the WITH in this PR we need a way to tell the parser that it is optional. That's why we pass the keyword via the more explicit `Ref.keyword` segment so that we can also pass an `optional=True` argument.

I agree that Sequence could work but think Ref.keyword is more accurate to the logic we're trying to express (the maintainers can have a discussion around consistency outside of this PR).

test/fixtures/dialects/postgres/postgres_create_database.sql
derickl: I missed that!

test/fixtures/dialects/postgres/postgres_create_database.sql
derickl: Will do

jpy-git (approved):
Awesome work @derickl and really robust test cases, LGTM 🚀 🌟

derickl:
Thanks for the great review @jpy-git

derickl:
Thank you for sharing your thoughts [here](https://github.com/sqlfluff/sqlfluff/pull/2081#issuecomment-990708401)  @tunetheweb. I hope we converge to some consistency down the road.

tunetheweb:
> I hope we converge to some consistency down the road.

Or maybe not if @jpy-git and I can't even agree! 🤣

PR #2082, opened 2021-12-10
Update parse method of Simple API to output JSON parse tree

### LINKED ISSUE #2062: Simple API: parse command
So the purpose of the Simple API is to provide a clear way for users to interact with sqlfluff via code.

In the case of the `lint` and `fix` commands the action of the functions are pretty well defined, however, for the `parse` command we return a ParsedString named tuple.
There's nothing strictly wrong with this per se, but was just wondering if there was a discussion to be had around whether this is currently providing users with too much access to the sqlfluff internals (as shown below) rather than a more well defined API?

image

jpy-git:

Fixes #2062. Updated parse method of Simple API to output the parse tree in JSON format. This keeps the output of the Simple API simple and decoupled from the internal workings of SQLFluff.

Updated the docs and unit tests accordingly.

Changes the output of the parse method of the Simple API. However, we're about to release a new minor release soon anyway so should be fine.

### Pull Request checklist

  - Other.


jpy-git:
@tunetheweb @barrywhart pretty happy with how the outputs of the Simple API are looking now 😄

test/api/simple_test.py
```diff
@@ -178,21 +179,16 @@ def test__api__fix_string_specific_exclude():
...
-        col_ref.raw for col_ref in parsed.tree.recursive_crawl("column_reference")
-    ]
-    assert col_refs == ["blah"]
-    # Check we can get table from it
-    tbl_refs = [
-        tbl_ref.raw for tbl_ref in parsed.tree.recursive_crawl("table_reference")
-    ]
-    assert tbl_refs == ["myTable"]
+
+    # Check we a JSON object is returned.
```
jpy-git, line 183:
```suggestion
    # Check a JSON object is returned.
```

jpy-git:
Just seen there's some test for the example files that I'll need to update tomorrow, but the rest is good for review

jpy-git:
@tunetheweb resolved the coverage issues on this 👍

barrywhart (approved):
Nice!

tunetheweb:
I see we lost example 3 (since we can no longer call `get_table_references`) but I wonder if we need a simple example to show how to equivalent can now be accomplished with the JSON?

Because we've basically gone from this in our docs:

```py
#  -------- PARSING ----------
# NOTE: sqlfluff is still in a relatively early phase of its
# development and so until version 1.0.0 will offer no guarantee
# that the names and structure of the objects returned by these
# parse commands won't change between releases. Use with care
# and keep updated with the changelog for the project for any
# changes in this space.

parsed = sqlfluff.parse(my_bad_query)

# Get the structure of the query
structure = parsed.tree.to_tuple(show_raw=True, code_only=True)
# structure = ('file', (('statement', (('select_statement', (('select_clause', (('keyword', 'SeLEct'), ...

# Extract certain elements
keywords = [keyword.raw for keyword in parsed.tree.recursive_crawl("keyword")]
# keywords = ['SeLEct', 'as', 'from']
tbl_refs = [tbl_ref.raw for tbl_ref in parsed.tree.recursive_crawl("table_reference")]
# tbl_refs == ["myTable"]
```

to this:

```py
#  -------- PARSING ----------

# Parse the given string and return a JSON representation of the parsed tree.
parse_result = sqlfluff.parse(my_bad_query)
# parse_result = {'file': {'statement': {...}, 'newline': '\n'}}
```

Appreciate this is more to do with telling someone how to use Python to manipulate JSON so could be argued it's not really in SQLFluff remit to do that, but at same time the original example showed how to crawl the `table_reference` segment (which IS a SQLFluff concept) and the new, reduced example, doesn't really give that guidance to users.

WDYT?

tunetheweb:
Ooops took too long to write this (got caught on a call) and now merged. Still let me know what you think.

barrywhart:
Ohhh, that's a valuable example. We used the SQLFluff API at my last job for this exact purpose -- to determine table references, then use that info to automatically create table lineage diagrams.

jpy-git:
We can reverse out the change if you want? Do we want to provide an example of how to do this with json or do you want the original method (i.e. full revert)?

tunetheweb:
Oh no keep the change. JSON example is what I mean.

jpy-git:
Awesome, let me pull together some examples tonight/tomorrow 😄

PR #2096, opened 2021-12-11
Grammar: Adds support for ALTER VIEW statement for Postgres dialect

derickl:
### Brief summary of the change made
Adds support for parsing `ALTER VIEW` statements for the Postgres dialect
Closes #2094 

### Are there any other side effects of this change that we should be aware of?
None

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `python test/generate_parse_fixture_yml.py` or by running `tox` locally).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #2097, opened 2021-12-11
Release 0.9.0

jpy-git:
Get ready for release 0.9.0!

In terms of non-draft PRs I believe it is only #2096 that I need to review for @derickl.

2 quick outstanding things that would be good to sort before the release are:
- [x] I need to update the simple API docs as discussed in #2082.
- [x] @alanmcruickshank the Docker image is still undocumented so would you be able to update the README in DockerHub with the proposed README [here](https://github.com/sqlfluff/sqlfluff/pull/1945#issuecomment-976016598) and the short description with the suggestion [here](https://github.com/sqlfluff/sqlfluff/pull/1945#issuecomment-976018225). Also if possible to add the SQLFluff logo as the image icon that would be nice. Thanks 😄 

I believe I've correctly updated the CHANGELOG.md as described in CONTRIBUTING.md, but given it's my first time running the release lmk if I've missed anything or if we want anything changed!


CHANGELOG.md
```diff
@@ -9,6 +9,96 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Note: Changes are now automatically tracked in [GitHub](https://github.com/sqlfluff/sqlfluff/releases) and will be copied in here on each release (please remember to update the issues and contributors to links!). There is no need to manually edit this file going forward.
 -->
 
+## [0.9.0] - 2021-12-11
```
tunetheweb, line 12: Reminder to change this to whatever date you end up releasing this.

CHANGELOG.md
```diff
@@ -9,6 +9,96 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+## What’s Changed
+
+This release brings about several great new additions including:
+- dbt 1.0.0 compatibility.
+- CLI and Simple API paramaters to provide custom paths to config files.
+- Refinement to Simple API (**BREAKING CHANGE**).
```
tunetheweb, line 19: Probably good to explain what the breaking change is:

```suggestion
- Refinement to Simple API to return parse output on JSON rather than internal SQLFluff object (**BREAKING CHANGE**).
```

CHANGELOG.md
```diff
@@ -9,6 +9,96 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- An [Official SQLFluff Docker Image](https://hub.docker.com/r/sqlfluff/sqlfluff).
+- Grammar improvements across various dialects.
+- A new rule to check for non-alphanumeric values in identifiers.
```
tunetheweb, line 22:
```suggestion
- A new rule (L057) to check for non-alphanumeric values in identifiers.
```

tunetheweb (approved):
Looks good!

Some small comments and suggestions:

CHANGELOG.md
barrywhart: Did you mean "to return parse output in JSON"?

CHANGELOG.md
```diff
@@ -9,6 +9,96 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+* Add --disable-noqa option to CLI and config [#2043](https://github.com/sqlfluff/sqlfluff/pull/2043) [@jpy-git](https://github.com/jpy-git)
+* TSQL: TRY/CATCH [#2044](https://github.com/sqlfluff/sqlfluff/pull/2044) [@jpers36](https://github.com/jpers36)
+* enabled arrays support in `declare` and `set` statements for `bigquery` dialect [#2038](https://github.com/sqlfluff/sqlfluff/pull/2038) [@KulykDmytro](https://github.com/KulykDmytro)
+* L008 refactor [#2004](https://github.com/sqlfluff/sqlfluff/pull/2004) [@jpy-git](https://github.com/jpy-git)
+* Support __init__.py for library_path [#1976](https://github.com/sqlfluff/sqlfluff/pull/1976) [@Tonkonozhenko](https://github.com/Tonkonozhenko)
+* L052: Redefine semi-colon newline to multiline newline [#2022](https://github.com/sqlfluff/sqlfluff/pull/2022) [@jpy-git](https://github.com/jpy-git)
+* Grammar: Remove hash inline comment from Postgres [#2035](https://github.com/sqlfluff/sqlfluff/pull/2035) [@jpy-git](https://github.com/jpy-git)
+* Enable glob rule matching for inline comments [#2002](https://github.com/sqlfluff/sqlfluff/pull/2002) [@jpy-git](https://github.com/jpy-git)
```
barrywhart, line 49:
```suggestion
* `noqa` enhancement: Enable glob rule matching for inline comments [#2002](https://github.com/sqlfluff/sqlfluff/pull/2002) [@jpy-git](https://github.com/jpy-git)
```

CHANGELOG.md
```diff
@@ -9,6 +9,96 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+* TSQL (ASA): Allow for table identifier in DELETE clause [#2031](https://github.com/sqlfluff/sqlfluff/pull/2031) [@jpers36](https://github.com/jpers36)
+* TSQL (ASA): Fix CTAS with WITH statement [#2028](https://github.com/sqlfluff/sqlfluff/pull/2028) [@jpers36](https://github.com/jpers36)
+* Grammar: Parse multiple grants [#2023](https://github.com/sqlfluff/sqlfluff/pull/2023) [@jpy-git](https://github.com/jpy-git)
+* Add tsql nested block comment support and add regex package [#2027](https://github.com/sqlfluff/sqlfluff/pull/2027) [@jpy-git](https://github.com/jpy-git)
```
barrywhart, line 53:
```suggestion
* Add tsql nested block comment support and add regex package dependency [#2027](https://github.com/sqlfluff/sqlfluff/pull/2027) [@jpy-git](https://github.com/jpy-git)
```

CHANGELOG.md
```diff
@@ -9,6 +9,96 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+* Grammar: Add complete datetime units to postgres dialect [#2025](https://github.com/sqlfluff/sqlfluff/pull/2025) [@jpy-git](https://github.com/jpy-git)
+* Grammar: Postgres CREATE POLICY [#2021](https://github.com/sqlfluff/sqlfluff/pull/2021) [@jpy-git](https://github.com/jpy-git)
+* Speed up CI [#1957](https://github.com/sqlfluff/sqlfluff/pull/1957) [@pwildenhain](https://github.com/pwildenhain)
+* Add support for Snowflake create/alter SQL and js UDF [#1993](https://github.com/sqlfluff/sqlfluff/pull/1993) [@chwiese](https://github.com/chwiese)
+* Add encoding CLI argument [#1994](https://github.com/sqlfluff/sqlfluff/pull/1994) [@jpy-git](https://github.com/jpy-git)
+* TSQL: Spaces allowed in comparison operators [#1965](https://github.com/sqlfluff/sqlfluff/pull/1965) [@jpers36](https://github.com/jpers36)
+* Add Snowflake schema options [#1950](https://github.com/sqlfluff/sqlfluff/pull/1950) [@chwiese](https://github.com/chwiese)
+* Rule globs [#1972](https://github.com/sqlfluff/sqlfluff/pull/1972) [@jpy-git](https://github.com/jpy-git)
```
barrywhart, line 63:
```suggestion
* CLI/`.sqlfluff` enhancement: Rule globs [#1972](https://github.com/sqlfluff/sqlfluff/pull/1972) [@jpy-git](https://github.com/jpy-git)
```

CHANGELOG.md
```diff
@@ -9,6 +9,96 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+* Allow column references in IN statement [#1971](https://github.com/sqlfluff/sqlfluff/pull/1971) [@tunetheweb](https://github.com/tunetheweb)
+* Remove config.ini in favor of setup.cfg [#1966](https://github.com/sqlfluff/sqlfluff/pull/1966) [@jpy-git](https://github.com/jpy-git)
+* Convert sqlfluff-templater-dbt setup.py to setup.cfg [#1963](https://github.com/sqlfluff/sqlfluff/pull/1963) [@jpy-git](https://github.com/jpy-git)
+* Official Docker image: Dockerfile and Github Actions workflow [#1945](https://github.com/sqlfluff/sqlfluff/pull/1945) [@jpy-git](https://github.com/jpy-git)
+* Move package metadata to setup.cfg [#1960](https://github.com/sqlfluff/sqlfluff/pull/1960) [@jpy-git](https://github.com/jpy-git)
+
+## 🐛 Bug Fixes
+
+* Fix tsql block comment close [#2095](https://github.com/sqlfluff/sqlfluff/pull/2095) [@jpy-git](https://github.com/jpy-git)
+* Fix PlaceHolderTemplater slice_type for substitutions [#2085](https://github.com/sqlfluff/sqlfluff/pull/2085) [@barrywhart](https://github.com/barrywhart)
```
barrywhart, line 76:
```suggestion
* Fix PlaceholderTemplater slice_type for templated code (substitutions) [#2085](https://github.com/sqlfluff/sqlfluff/pull/2085) [@barrywhart](https://github.com/barrywhart)
```

barrywhart (approved):
A few minor suggestions around wording. Looks good overall.

CHANGELOG.md
tunetheweb: Yup. Updated.

CHANGELOG.md
tunetheweb: If changing this we should update the PR #2002 as the draft release page is fully refreshed each time a PR merge is merged.

alanmcruickshank:
@jpy-git - README and short description added ✅ . I haven't managed to find any settings or documentation on how to add an icon to the repository. I've got the sqlfluff icon on the user which I think is all I can do without paying for an account or asking docker really nicely.

If anyone has any other ideas on how to get an icon on it - then I'm all ears!

jpy-git:
@alanmcruickshank awesome thanks, the readme looks great.

It looks like to get an icon on the image itself we'd need to apply to make SQLFluff a "Docker Official Image": https://docs.docker.com/docker-hub/official_images/

Reading through those docs it seems like the process is fairly involved so maybe that's something we can leave for when we come out of beta 😊

tunetheweb:
I notice the Dockerfile doesn’t include the dbt templater plug-in. Should it? Or is it fair to presume dbt users will all be running it locally as they’ll all be Python users anyway?

jpy-git:
@tunetheweb we'll need to build a second docker image for the dbt templater but that's extension work

tunetheweb:
Surely both need to exist in the same docker image to be usable? You can't use the templater without SQLFluff.

So thew question is do you need two docker images:
- A SQLFluff only one
- A SQLFluff and dbt templater one

Or should we just have one? In which case does it need the dbt templater?

To me the only reason to have two would be to keep the image lighter, but don't think the dbt templater adds that much to the image size (though it does need bits of dbt top so maybe it would).

Anyway agree for now let's go with just the SQLFluff one. As I said above, it may well be a bit overkill having dbt within docker if most dbt users are python users so won't use the Docker image anyway.

jpy-git:
You would just have the templater image inherit from the non-templater image

I.e.
```dockerfile
FROM sqlfluff/sqlfluff:latest

<dbt templater extras here>
```

alanmcruickshank:
One thing that could make this relatively easy would be to maintain two seperate _tags_ within the same repo rather than two separate docker repositories (a little like how the python image maintains different versions and different debian versions as different tags). I agree that this doesn't need to stop us releasing and it could be follow-on work. It likely would involve some additional machinery anyway to get credentials into the container anyway.

barrywhart:
I'm not familiar with using Docker images this way. How will users provide their .sqlfluff, dbt_project.yaml, and other files? Will they use the image directly or inherit from our image?

jpy-git:
Let me make an issue to capture these initial thoughts and we can move the discussion there

tunetheweb:
> I'm not familiar with using Docker images this way. How will users provide their .sqlfluff, dbt_project.yaml, and other files? Will they use the image directly or inherit from our image?

Basically you normally mount your local drive in the image. Something like this:

```
docker container run -it --rm -v ~/myproject:/myproject --entrypoint="sqlfluff lint /myproject"
```

Docker runs the command and outputs the result.

So this would be useful for T-SQL developers for example, that normally might not use python and don't want to manage a python environment and all that entails to keep it working and up to date. If they have Docker installed they can just use our Docker image to save them having to run it through a local python install.

PR #2100, opened 2021-12-12
TSQL: "LANGUAGE" is not a reserved keyword

jpers36:
In TSQL, "LANGUAGE" is not yet a reserved keyword.  It's a valid name for a column.  This PR moves "LANGUAGE" from TSQL's reserved keyword list to its unreserved keyword list.  It also adds a test case element.

### DIFF
--- a/src/sqlfluff/dialects/dialect_tsql_keywords.py
+++ b/src/sqlfluff/dialects/dialect_tsql_keywords.py
@@ -114,7 +114,6 @@ RESERVED_KEYWORDS = [
     "JOIN",
     "KEY",
     "KILL",
-    "LANGUAGE",
     "LEFT",
     "LIKE",
     "LINENO",
@@ -284,6 +283,7 @@ UNRESERVED_KEYWORDS = [
     "KEEPFIXED",
     "KEEPIDENTITY",
     "LABEL",  # Azure Synapse Analytics specific, reserved keyword but could break TSQL parsing to add there
+    "LANGUAGE",
     "M",
     "MAX_DURATION",
     "MAX_GRANT_PERCENT",
--- a/test/fixtures/dialects/tsql/select.sql
+++ b/test/fixtures/dialects/tsql/select.sql
@@ -25,6 +25,7 @@ SELECT
 	'TSQL' AS 's escaping quotes test',
 	'',
 	'''',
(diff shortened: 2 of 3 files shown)


jpy-git:
Makes sense and agree that we should not stop this from parsing.

Reading through the [documentation](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/reserved-keywords-transact-sql?view=sql-server-ver15) it's marked as one of the ODBC Reserved Keywords, which aren't reserved for T-SQL itself but are reserved for ODBC function calls so recommended against using. It's also in the FUTURE keywords section.

Looking at that list there are a lot of other keywords that we have in our Reserved list that may also be currently possible to use as column names in T-SQL. Would you be able to investigate if this also affects those?

Also if we are to make these non-reserved should we have some mechanism to warn users against using them, given that it could lead to their code breaking in the future?

tagging @WittierDinosaur as I know he's done some more advanced separating of keywords in the Postgres dialect so might have some good opinions on how to handle these 😄

jpers36:
L029 checks for unreserved keywords being used as identifiers.  Is that the sort of mechanism you mean, or is there something additional you would recommend?

jpy-git:
Ah nice I think that covers the warnings then 😄 

Although worth noting that by default L029 has `unquoted_identifiers_policy = aliases` which wouldn't cover column names unless that was changed to `unquoted_identifiers_policy = all` in the users config. I assume the motivation being that you don't always have the ability to manage upstream naming conventions so harsh to flag the user for that 👍 

Could you test some of those other ODBC Reserved keywords that we have as Reserved (like AND, BEGIN, GO, FROM) and see if we can also use those as column names in T-SQL (and if we were to move them would it cause any issues with parsing in SQLFluff)?

jpers36:
Sure!  I'll probably get to that tomorrow, unless we're planning to get 0.9 out before then.

jpy-git:
awesome thanks! tomorrow should be fine, I'm still working through a couple reviews before the release 😄

jpy-git:
@jpers36 hows progress going on this?

jpers36:
@jpy-git this is ready for review.  I tested all keywords listed as reserved to see whether they could be instantiated as column identifiers, and then adjusted the lists based on what I discovered.

PR #2107, opened 2021-12-12
autofill params for placeholder template

danicc097:

Add an optional config flag to fill missing custom regex query params for the placeholder template

None

### Pull Request checklist

  - Other.


src/sqlfluff/core/templaters/placeholder.py
```diff
@@ -144,6 +147,17 @@ def process(
...
             last_literal_length = span[0] - last_pos_raw
+
+            if (
+                context.get("autofill_missing_params")
+                and param_name not in context.keys()
+            ):
+                templater_logger.info(
+                    f"Autofilling missing parameter {param_name} with {AUTOFILL_PARAMS['string']}"
+                )
+                # TODO generate from types if exist
```
jpy-git, line 158: This is a fairly big issue for this PR:

The current implementation works for the specific case that the we are substituting in an object (should be a string I guess but doesn't have quotes atm). 
image

But this is going to cause parsing errors whenever a string is not appropriate:
image
This is misleading to the user that there's something wrong with their query rather than that they've just forgotten to add a variable.

Or what if the intended substitution type is not clear or a mix or includes complicated structure and this substitution could cause the query to be interpreted differently or have a parsing error:
image
image

The big issue is templating happens before parsing, as how are you supposed to parse code without knowing what the code is, but then in your case if you haven't parsed the code then how can you interpret what type is implied?

That's the motivation for us having users provide their own values as a template could feasibly hold anything and only they'll 100% know what is intended.

test/core/templaters/placeholder_test.py
```diff
@@ -250,6 +251,26 @@ def test__templater_custom_regex():
...
+        )
+    )
+    outstr, _ = t.process(
+        in_str="SELECT bla FROM blob WHERE missing_param = @missing_param AND explicit_param = @explicit_param",
+        fname="test",
+        config=FluffConfig(),
+    )
+    assert (
+        str(outstr)
+        == f"SELECT bla FROM blob WHERE missing_param = {AUTOFILL_PARAMS['string']} AND explicit_param = explicit_param"
```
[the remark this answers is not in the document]
jpy-git: or maybe you mean it to be unquoted but then `__string__` probably isn't a good name

src/sqlfluff/core/templaters/placeholder.py
jpy-git: definitely can't be merged in as is given the potential parse issues.

jpy-git (commented):
I like the idea, but think there are some questions around how feasible this is?

@barrywhart might be worth getting your opinion too

barrywhart:
I'd like to see an explanation of what need this PR is intended to address. What kinds of parameters will this be useful for, and which will it not be useful for? Is there a reason SQLFluff can't use the same parameter values that will be used in production, or configured with dummy values?

This info will be really helpful for understanding and reviewing the PR. I think it'll also be useful to add to the documentation.

danicc097:
I've fixed some issues as pointed out by @jpy-git, and added some basic support for Postgres data types.

@barrywhart my specific use case, for instance, is using [sqlc](https://github.com/kyleconroy/sqlc) with Postgres. Variables use `@` prefix, so I'd have queries like:

```sql
-- name: UpdateUserById :one
UPDATE
  "users"
SET
  "password" = CASE WHEN @password_do_update::boolean THEN
    @password
  ELSE
    "password"
  END,
  "salt" = CASE WHEN @salt_do_update::boolean THEN
    @salt
  ELSE
    "salt"
  END,
  "username" = CASE WHEN @username_do_update::boolean THEN
    @username
  ELSE
    "username"
  END,
  "email" = CASE WHEN @email_do_update::boolean THEN
    @email
  ELSE
    "email"
  END
WHERE
  "id" = @id
RETURNING
  *;
```

creating dummy values in the config seems unnecesary in these cases, except common things like "id", etc.  Regarding usefulness, if a user opts in to autofill parameters it will look for the Postgres "::" syntax (else notify it will be replaced with text) and parse simple builtin types and arrays. Its just a matter of extending the recognized types with dummy values in the templater. Of course, as it is now, the implementation is limited to Postgres (else it will always autofill a string); I don't know if for other databases it's worth both the verbosity of typing and effort to make its type parser in the templater, maybe someone has more experience in other dialects

barrywhart:
This is an interesting idea, but as written, I don't think it's appropriate for merge into the public SQLFluff package. The placeholder templater is designed to be generic, but this adds two kinds of specific code:
* Postgres-specific code: The list of default values by type
* Code that is specific to your use case: I.e. the assumption that the regex may define a `param_type` field

I can think of some ways to do this more generically, but for now, I would recommend creating a custom templater for this. SQLFluff supports adding new templaters as [plugins](https://docs.sqlfluff.com/en/stable/developingplugins.html). For example, the dbt templater is written as a plugin.

jpy-git:
Have to agree with @barrywhart. It only applies to the specific case of people using the postgres + the placeholder template + also cast all of their variables (even though that doesn't apply if they're trying to insert a table name or a function name etc with the template) and risks causing issues/confusion for the general user not using those.

Makes sense to add this as a custom templater as @barrywhart describes. You can create that plugin separately for your use case and do whatever you want with it (N.B. you would create it in your own repo not the public one). You can basically copy a lot of the placeholder template logic but with these bits added in.

@barrywhart could you create an issue to capture any initial thoughts around a generic approach to this? 😄

Closing this PR

barrywhart:
Created an issue to capture some related ideas: #2108

PR #2111, opened 2021-12-13
fix delimited identifier parsing for spark3

### LINKED ISSUE #2110: Delimited identifiers in spark3 do not parse
Spark SQL uses backticks to quote identifiers (and calls them "delimited"), but Sqlfluff's spark3 dialect fails to parse these (example below). 

I believe this is a [straightforward fix](https://github.com/mcannamela/sqlfluff/commit/7d93e3db07d33f546761eff206807317d8674402#diff-d5586dfc7929eaa1d7a0f25e90a12e0e672d6eda58eaed0bd0c7f5e230118f26R123) for most cases (it will still fail if you attempt to use backticks to identify a CTE, i.e. ```with `a` as (...)```).

I have this fix plus tests [on a branch](https://github.com/mcannamela/sqlfluff/tree/mc-fix-spark-delimited-identifier) that I will PR if it's acceptable. 

## Expected Behaviour
Per the [spark docs](https://spark.apache.org/docs/latest/sql-ref-identifier.html#delimited-identifier), within backticks any character goes for identifiers. 

## Observed Behaviour
Parsing fails due to backtick-delimited identifier

## Steps to Reproduce
Use a backtick-delimited identifier in any part of the query. 

Ex:
```
select 1 as `delimited_identifier`
```

Result of `sqlfluff parse`:
```
== [fails_linting.sql] FAIL                                                                                                                                                     
L:   1 | P:   9 |  PRS | Line 1, Position 9: Found unparsable section: ' as
                       | `delimited_identifier`'
All Finished 📜 🎉!
root@0d408ccb11a1:/dbt# sqlfluff parse fails_linting.sql 
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    base:
[L:  1, P:  1]      |                keyword:                                      'select'
[L:  1, P:  8]      |                    literal:                                  '1'
[L:  1, P:  9]      |                    unparsable:                               !! Expected: 'Nothing...'
[L:  1, P:  9]      |                        whitespace:                           ' '
[L:  1, P: 10]      |                        raw:                                  'as'
[L:  1, P: 12]      |                        whitespace:                           ' '
[L:  1, P: 13]      |                        raw:                                  '`delimited_identifier`'
[L:  1, P: 35]      |            [META] dedent:
[L:  1, P: 35]      |    newline:                                                  '\n'

==== parsing violations ====
L:   1 | P:   9 |  PRS | Line 1, Position 9: Found unparsable section: ' as
                       | `delimited_identifier`'
```

## Dialect
spark3

## Version
Python 3.8.12
sqlfluff, version 0.8.2
dbt, installed version: 1.0.0

## Configuration
```
[sqlfluff]
verbose = 0
nocolor = False
dialect = spark3
templater = jinja
rules = None

recurse = 0
output_line_length = 80
runaway_limit = 10
ignore_templated_areas = True
```

mcannamela:
Fixes #2110. 

This PR replaces the trim character for `QuotedIdentifierSegment` in the spark3 dialect with backtick, per the [Spark docs on legal identifiers](https://spark.apache.org/docs/latest/sql-ref-identifier.html#delimited-identifier) and my own testing (using Databricks as the spark provider).  Both ANSI and non-ANSI mode have the same behavior:
```
--works
select 1 as `delimited identifier`;

--fails
select 1 as "delimited identifier";
```

Identifiers surrounded by double quotes will now (rightly I think) fail to parse. 

### Pull Request checklist

  - Other.


test/fixtures/dialects/spark3/select_regular_identifier_without_leading_digits.sql
```diff
@@ -0,0 +1,2 @@
+select 1 as plain_identifier_with_no_tricks_1;
```
mcannamela, line 1: Surprisingly, plain identifiers in spark can start with digits like `0_is_a_legal_identifier`, and these also fail to parse. However, since they parse when delimited and making that worked seemed like it would be a bigger impact than just fixing the quote character, I stopped investigating.

jpy-git: this should parse according to the definition in ANSI, but I can see an issue with the current regex so I'll look at fixing this. Can you remove this file from the PR as it's not relevant to the back tick identifiers?

test/fixtures/dialects/spark3/select_delimited_identifier_with_escaped_backticks.sql
```diff
@@ -0,0 +1 @@
+select `delimited``with escaped` from `some``table`;
```
jpy-git, line 1: So looking in the [docs](https://spark.apache.org/docs/latest/sql-ref-identifier.html#delimited-identifier):
image

It looks like in spark that the example above is a single identifier containing backticks in its name: `delimited``with escaped`
But if you refer to how this is being parsed it is being interpreted as a column `delimited` aliased as `with escaped`:
image

I'm not super familiar with Spark SQL myself so can you clarify a couple things for me:
- Am I correct in thinking that this should be a single `delimited``with escaped` identifier? If so then we'll need a more advanced Regex lexer for back_quote in Spark.
- If yes to the above, what does an aliased back-tick quoted identifier look like in Spark SQL without an AS (if that's possible)? Is this possible 
```
`delimited``with escaped` `alias`
```
I assume it's the fact that those escaped backticks are sequential that differentiates them

test/fixtures/dialects/spark3/select_regular_identifier_without_leading_digits.yml
```diff
@@ -0,0 +1,32 @@
+# YML test files are auto-generated from SQL files and should not be edited by
+# hand. To help enforce this, the "hash" field in the file must match a hash
+# computed by SQLFluff when running the tests. Please run
+# `python test/generate_parse_fixture_yml.py`  to generate them after adding or
+# altering SQL files.
+_hash: 7d052fb396242780a8bbfcdd82e1b3c6e8cfe53e5db297e79997723a9c493ad1
```
jpy-git, line 6: Also remove this file as it's not relevant to the back ticks

jpy-git (commented):
@mcannamela Always great to have spark contributors, it's a fairly new dialect in SQLFluff 😄 
Couple points to address re: the escaped back ticks.

test/fixtures/dialects/spark3/select_regular_identifier_without_leading_digits.sql
jpy-git: The leading numeric issue will be fixed by #2114 so thanks for highlighting this 👍

test/fixtures/dialects/spark3/select_delimited_identifier_with_escaped_backticks.sql
mcannamela: Ah good point, I didn't notice how it actually parsed. I'll caveat here by saying I am also no expert in this dialect, so I'm just reading the docs and verifying what happens for test queries as I go. 

You are correct that ``` `delimited``with escaped` ``` should be one identifier that contains a backtick; it will output as ``delimited`with escaped``. 

Your example with the alias is also legal. For instance this query will execute under both ANSI mode and not:
```
with a as (
select 1 as `b``c`
)
select `b``c` `!d!`, `b``c` as `?e` from a;
```
The result is: 
image

I don't immediately see where I would accommodate this in sqlfluff. 

Can you point me in the right direction or would it be better to keep this PR uncomplicated and raise a separate issue for handling escaped backticks? Per the docs:
> Use ` to escape special characters (e.g., `).

makes me wonder whether there are not other things that should be getting escaped. 

So far I have no examples of a backtick-containing identifier "in the wild", but the need for delimited identifiers has cropped up because of e.g. some naughty systems generating csv's with field names like "ctr, %".

jpy-git: Makes sense, you've got to question the design of whoever thought using backticks to escape backticks within backticks was nice design lol 😅 

So in terms of getting this to parse we're going to need to overwrite regex pattern in the back_quote lexer in the ANSI dialect ([here](https://github.com/sqlfluff/sqlfluff/blob/9504fc28af3142f3e4915b0adf5e31746df06e43/src/sqlfluff/dialects/dialect_ansi.py#L85)). The pattern currently matches anything that's not a backtick within 2 backticks.

For Spark we'll need to come up with some regex pattern that matches anything between 2 backticks provided the closing backtick is not immediately preceded or followed by another backtick.

So if you copy the linked ANSI back_quote lexer and add it into Spark [here](https://github.com/sqlfluff/sqlfluff/blob/9504fc28af3142f3e4915b0adf5e31746df06e43/src/sqlfluff/dialects/dialect_spark3.py#L45) then you can start playing around with the regex pattern to match these back_quote containing escaped backticks.

jpy-git: In terms of what that pattern would be it's hard to say without playing around myself. 

My initial thoughts would be something like
```
`.*(?<!`)`(?!`)
```
anything between 2 backticks, with negative lookbehind + negative lookahead to check that the closing backtick is not surrounded by another backtick. Can't promise that works but a good place to start.

We use the [regex](https://pypi.org/project/regex/) package so you should have all of the standard python re notations + the additional regex ones at your disposal.

jpy-git: (lmk if you're struggling with it though as I'm happy to have a play with the pattern tomorrow. Worst case we can merge as is, given the standard back_quote regex covers 99% of cases, and I'll deal with this edge case later 😄 )

mcannamela: Seriously. Ok I'll take a look, thanks!

jpy-git: realise my initial pattern suggestion is too greedy. It would capture all of
```
`delimited ``identifier` `alias`
```
rather than just the first possible match.

Making the `.*` non-greedy (i.e. `.*?`) seems to work from my initial testing:
image

so patten would be:
```
`.*?(?<!`)`(?!`)
```

mcannamela: heh thanks I am shamefully bad at regex, but that works afaict

mcannamela: ah although this also appears to match when it ought to fail i believe:
```
`delimited```with escaped`
```

jpy-git: nice, looks good 😄 Could you add some more unit tests to this file to verify that the pattern works for aliases (it does but good to have the tests in case someone breaks it in the future)

```sql
SELECT `delimited ``identifier` `alias` FROM `some``table` `some``alias`;
SELECT `delimited ``identifier` AS `alias` FROM `some``table` AS `some``alias`;
```

Then should be good to go 👍

mcannamela: ok, though what do you think of:
```
`.*?(?<![^`]`)`(?!`[^`])
```
to patch the false positive with the triple backticks in my previous comment?

Those are good suggestions I'll add them now.

jpy-git: ah nice test case, let me take a look

jpy-git: So both the original regex and the patched one you suggested also can't handle a valid column like this
image

jpy-git: Or this too
image

jpy-git: Okay think I have it: 
```
`([^`]|``)*`
```
opens the backtick quote, greedily matches all the double backtick qoutes and then closes on the next quote. Seems to be able to handle all the test cases, even with multiple escaped backticks, and doesn't parse odd numbers of subsequent ticks 👍 

image

Might as well chuck these tests in too 😄 
```sql
select ```delimited` `alias` FROM `some``table` `some````alias`;
select `delimited``` `alias` FROM `some``table` ```some````alias```;

SELECT `delimited ``identifier` `alias` FROM `some``table` `some``alias`;
SELECT `delimited ``identifier` AS `alias` FROM `some``table` AS `some``alias`;
```

mcannamela: hahaha i could see when i added a third backtick that things could get out of hand
i'll make the amendments and take this out of draft
thanks for your help!

jpy-git: haha it's a pandora's box! Thanks for your patience working through this 😄

test/fixtures/dialects/spark3/select_delimited_identifier_with_escaped_backticks.yml
```diff
@@ -0,0 +1,123 @@
...
+          - raw: table
+          - raw: '` `'
+          - raw: some
+          - raw: '``'
+          - raw: '``'
+          - raw: alias
+          - raw: "`;\nselect `"
+          - raw: delimited
+          - raw: '```'
+          - raw: '`alias`'
```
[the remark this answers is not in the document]
mcannamela: woops, good catch, pushed that just now

jpy-git:
@mcannamela looks like you need to run black to format the code

jpy-git (approved):
Well done @mcannamela, thanks for working through all the edge cases! LGTM 🚀 🌟

PR #2114, opened 2021-12-14
Stop numeric literal from splitting valid naked identifiers.

jpy-git:

The current numeric literal regex is matching the start of valid naked identifier segments and causing them to be split which results in parsing errors (Initially highlighted [here](https://github.com/sqlfluff/sqlfluff/pull/2111#discussion_r768046391)):
image

This PR adds a negative lookahead assertion to the numeric literal regex to check that we aren't at the start of naked identifier:
image

N.B. Since the negative lookahead comes after the exponential matching we don't need to worry about this impacting the parsing of those (evidenced by continued parsing of `select_numeric_literal_exponential_format.sql` unit test) 😄 

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -85,8 +85,12 @@ ansi_dialect.set_lexer_matchers(
         RegexLexer("back_quote", r"`[^`]*`", CodeSegment),
         # See https://www.geeksforgeeks.org/postgresql-dollar-quoted-string-constants/
         RegexLexer("dollar_quote", r"\$(\w*)\$[^\1]*?\$\1\$", CodeSegment),
+        # Numeric literal matches integers, decimals, and exponential formats,
+        # with a negative lookahead assertion to check it is not part of a naked identifier.
         RegexLexer(
-            "numeric_literal", r"(\d+(\.\d+)?|\.\d+)([eE][+-]?\d+)?", CodeSegment
+            "numeric_literal",
+            r"(?>\d+(\.\d+)?|\.\d+)([eE][+-]?\d+)?(?![a-zA-Z_])",
+            CodeSegment,
         ),
         RegexLexer("not_equal", r"!=|<>", CodeSegment),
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -85,8 +85,12 @@
         RegexLexer("back_quote", r"`[^`]*`", CodeSegment),
         # See https://www.geeksforgeeks.org/postgresql-dollar-quoted-string-constants/
         RegexLexer("dollar_quote", r"\$(\w*)\$[^\1]*?\$\1\$", CodeSegment),
+        # Numeric literal matches integers, decimals, and exponential formats,
```
barrywhart, line 88: This looks like a reasonable fix. I wonder, though -- often in the past, we have fixed similar issues by changing the order of definitions, so the lexer or parser checks for the more general thing first. Could we potentially fix this instead by reordering this and the "competing" token definition? If so, that may be a simpler fix.

jpy-git: We probably can, but doesn't that just risk someone swapping them again in the future and reintroducing the error? 

Plus not sure how that ordering is impacted if someone were to patch the naked identifier segment in an inheriting dialect? Would it get dropped below the numeric literal from ansi and cause this issue?

barrywhart: >doesn't that just risk someone swapping them again in the future and reintroducing the error?

Once we have tests, that would catch it if someone swaps them again.

>Plus not sure how that ordering is impacted if someone were to patch the naked identifier segment in an inheriting dialect? Would it get dropped below the numeric literal from ansi and cause this issue?

Offhand, I don't know.

I'm happy to approve and merge this, because it doesn't seem like a big deal either way. I did want to mention the ordering option, if nothing else for potential future fixes. For a while, I wasn't aware of it, then Niall mentioned it for something I was working on. In some cases, ordering is a nice solution because it may reduce complexity and improve speed ("first match wins" versus making a pattern more complex to avoid false matches).

barrywhart (approved):

Issue #1805, opened 2021-11-01
GROUPS/USER Query Errors in Redshift


tdstark:
Currently, any queries which CREATE/ALTER USER/GROUP permissions in Redshift throw errors. These will likely need to be added to the Redshift dialect.

## Expected Behaviour
The following four queries throw errors:
`CREATE USER test CREATEUSER PASSWORD :TEST_PASSWORD;`
`CREATE GROUP test_group;`
`ALTER GROUP test_group ADD USER new_user;`
`ALTER USER power_user WITH SYSLOG ACCESS UNRESTRICTED;`

## Observed Behaviour
Each one of these statements throws errors.

## Steps to Reproduce
Run `sqlfluff lint` while pointing at a .sql script with the aformentioned queries.

## Dialect
Redshift

## Version
`v0.7.1`

## Configuration
Default configuration

tunetheweb:
This is a great first issue for anyone looking to start contributing to SQLFluff: https://github.com/sqlfluff/sqlfluff/wiki/Contributing-Dialect-Changes

PR #2115, opened 2021-12-14
Grammar: Add CREATE USER/GROUP statement to Redshift dialect

jpy-git:

Partially addresses #1805. Adds the CREATE USER/CREATE GROUP statement to the Redshift dialect.

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -17,6 +17,7 @@
 )
 
 from sqlfluff.core.dialects import load_raw_dialect
+from sqlfluff.core.parser.grammar.greedy import StartsWith
```
tunetheweb, line 20: For ANSI this is added to the `sqlfluff.core.parser` import above. Any reason this needs it's own one here?

Though actually I don't think you need this, based on two comments I made below:

src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -293,9 +294,93 @@ class StatementSegment(BaseSegment):
...
+    """
+
+    type = "create_user"
+
+    match_grammar = StartsWith(
+        Sequence("CREATE", "USER"),
+    )
+    parse_grammar = Sequence(
+        "CREATE",
+        "USER",
```
tunetheweb, line 321: @WittierDinosaur and I [had a big discussion on this](https://github.com/sqlfluff/sqlfluff/pull/1877#discussion_r746851959) and came to the conclusion that `match_grammar`'s that start with the same as `parse_grammar`'s doesn't add much and so there's no real benefit for this (at least I think that's where we landed!).

So I'd suggest simplifying this to:

```suggestion
    match_grammar = Sequence(
        "CREATE",
        "USER",
```

src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -293,9 +294,93 @@ class StatementSegment(BaseSegment):
...
+    """
+
+    type = "create_group"
+
+    match_grammar = StartsWith(
+        Sequence("CREATE", "GROUP"),
+    )
+    parse_grammar = Sequence(
+        "CREATE",
+        "GROUP",
```
tunetheweb, line 376: Ditto here:

```suggestion
    match_grammar = Sequence(
        "CREATE",
        "GROUP",
```

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_redshift.py
jpy-git: Ahh sorry, VScode does auto-import for me when I use and unimported class/function and tends to choose the source file to import from (rather than the __init__.py in src/sqlfluff/core/parser).

As you said, it's unnecessary though so I'll remove it 👍

src/sqlfluff/dialects/dialect_redshift.py
jpy-git: @tunetheweb makes sense, I only did it because it's a very common pattern in the existing dialect. Will remove from here.

Should I make a separate PR to start removing this from the rest of the grammar as well to tidy things up?

tunetheweb: I actually checked a few statements in the Redshift dialogue and saw it wasn’t used on the ones I checked. Was going to let you away with it if it was the standard in this dialect. Guess you and I happened to check different statements! But then wouldn’t StartsWith already exist in the dialect? 🤔

@WittierDinosaur what do you reckon? Clean it up, or you still convinced this is the way to go and eventually you might make it work (in which case I’d suggest not cleaning it up but also not encouraging it’s use in case we never do that).

jpy-git: I would've looked at a similar segment from another dialect, but the question still stands for those other dialects 😄

jpy-git:
@tunetheweb removed those StartsWith match grammars 👍

tunetheweb (approved):

PR #2121, opened 2021-12-16
postgres: Add datatypes

### LINKED ISSUE #2109: postgres: Implement explicit datatypes
## Expected Behaviour
SqlFluff should post an error message if the `create table` statement has a column definition using `datetime` as data type. The Postgres dialect has no such data type.

## Observed Behaviour
No error message shown.

## Steps to Reproduce
Use the following Sql:
```sql
create table a(
  b datetime not null
);
```

## Dialect
Postgres

## Version
- SqlFluff 0.8.2
- Python 3.9.7
 

## Configuration
No config at all.

kayman-mk:
Add all numeric data types used by the Postgres dialect, e.g. int, float, numeric, serial.

Closes #2109 

I changed the Redshift dialect (inherits from Postgres) to use the old definition from the Ansi dialect. Later we have to check what part of the code should be moved to Ansi as a base for other dialects. But due to the inheritance this will change all other dialects too.

### Pull Request checklist

  - Other.


kayman-mk:
Ok, problems were related to my local environment. No I get the same error as in the pipeline.

kayman-mk:
@jpy-git Sorry, not ready for review

jpy-git:
@kayman-mk sorry that accidently pressed the ready for review button 👍

kayman-mk:
@jpy-git Do you have an idea how to solve these problems?

1. The parse tree looks strange now:
```
          data_type_identifier:
            keyword: integer
```
Should be like the old one:
```
- data_type_identifier: integer
```

I know that this is due to the segment I use. The solution would be to place the data types within the DatatypeIdentifierSegment. But as this definition gets huge I'd like to move it into a separate class. But it is a RegExParser and not a segment and I have no idea how to do that. I didn't find any examples in the code.

2. The Redshift dialect inherits Postgres. But Redshift still has to use the old ANSI definition. I tried to overwrite the DatatypeIdentifierSegment but it is not working as it is a RegExParser and not a BaseSegment.
```
redshift_dialect.replace(DatatypeIdentifierSegment=ansi_dialect.get_segment("DatatypeIdentifierSegment"))

  File "/mnt/sqlfluff/src/sqlfluff/dialects/dialect_redshift.py", line 49, in <module>
    redshift_dialect.replace(DatatypeIdentifierSegment=ansi_dialect.get_segment("DatatypeIdentifierSegment"))
  File "/mnt/sqlfluff/src/sqlfluff/core/dialects/base.py", line 205, in get_segment
    if not issubclass(self._library[name], BaseSegment):  # pragma: no cover
TypeError: issubclass() arg 1 must be a class
```

kayman-mk:
Have a working solution now, but is the parse tree correct (data_type section)?

```
create table a(a smallint);
file:
- statement:
    create_table_statement:
    - keyword: create
    - keyword: table
    - table_reference:
        identifier: a
    - bracketed:
      - start_bracket: (
      - column_reference:
          identifier: a
      - data_type:
          keyword: smallint
...
```

kayman-mk:
@jpy-git #2122 is an alternative implementation which has a nicer parse tree. So which one do you prefer?

jpy-git:
@kayman-mk I'll review this evening

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -402,7 +402,34 @@ class DatatypeSegment(BaseSegment):
...
+                    OneOf("FLOAT", "FLOAT4", "FLOAT8"),
+                    Bracketed(Ref("NumericLiteralSegment"), optional=True),
+                ),
+                # numeric types [precision ["," scale])]
+                Sequence(
+                    OneOf("DECIMAL", "NUMERIC"),
+                    Bracketed(
+                        Delimited(Ref("NumericLiteralSegment")),
+                    ),
+                    optional=True,
```
jpy-git, line 431: This optional is in the wrong place. if you refer to the parse tree you can see the current grammar doesn't handle a lone DECIMAL/NUMERIC
```suggestion
                    Bracketed(
                        Delimited(Ref("NumericLiteralSegment")),
                    optional=True,
                    ),
```

src/sqlfluff/dialects/dialect_postgres_keywords.py~
```diff
@@ -0,0 +1,973 @@
+"""Keywords in the Postgres Dialect.
```
jpy-git, line 1: This looks like a backup copy of dialect_postgres_keywords, please remove. General note: you need to be more conscious of what you're committing in these PRs

src/sqlfluff/dialects/dialect_postgres.py
jpy-git: In general this is the right approach though and the parse tree looks fine 👍 

Couple notes:
- DataTypeIdentifier: FYI This is a general regex matcher for things potential datatypes. The purpose of this is to catch user defined data types. It should exclude Reserved Keywords, but doesn't atm due to the type grammar not being fully implemented as we discussed in the original issue.
- Brackets after data types: if you refer to line 456 you can see that there is a bracketed segment, which I think was originally put in as a general way to get brackets after a datatype. But means that weird types can parse:
image
Basically we'll need to be more specific with the other datatypes as to which can and can't have brackets after them (like you did for FLOAT)

There aren't that many other types in https://www.postgresql.org/docs/14/datatype.html so it's easier to add them all in this PR and we can do the tidy up once rather than splitting this up.

src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -57,6 +60,17 @@ class DatePartFunctionNameSegment(BaseSegment):
...
+class DatatypeIdentifierSegment(BaseSegment):
+    """A scalar or aggregate function.
+
+    Revert back to the ANSI definition to support ignore nulls
+    """
+
+    type = "function"
+    match_grammar = ansi_dialect.get_segment("FunctionSegment").match_grammar.copy()
+
+
```
jpy-git, line 73: This is wrong, please remove. Provided description of DatatypeIdentifierSegment in earlier comment.

src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -44,6 +44,9 @@
 redshift_dialect.sets("bare_functions").update(["current_date", "sysdate"])
 
 redshift_dialect.replace(WellKnownTextGeometrySegment=Nothing())
+# TODO: for the time use the ANSI definition and not the updated one from Postgres as not all
+#       data types are supported by Redshift
+redshift_dialect.replace(DatatypeSegment=ansi_dialect.get_segment("DatatypeSegment"))
```
jpy-git, line 49: Fine with this 👍

jpy-git (commented):
for the most part good, couple things that need removing, and we should get the rest of the types added here to clear up some issue that are arising with half doing this

src/sqlfluff/dialects/dialect_postgres_keywords.py~
kayman-mk: Same idea. Sorry for that.

kayman-mk:
Everything fine from my side but Jinja complains about failing tests. I adjusted the yml files (`data_type_identifier` replaced by `keyword`) but no luck.

@jpy-git Any ideas appreciated. Is there a command to regenerate the yml files for Jinja?

test/fixtures/templater/jinja_f/jinja.yml
```diff
@@ -111,7 +111,7 @@ file:
                   end_square_bracket: "]"
               keyword: AS
               data_type:
-                data_type_identifier: FLOAT64
```
jpy-git, line 114: If you look at the `.sqlfluff` file for this test it's using ANSI dialect so we shouldn't change this for jinja_f yet. Then the tests will pass. (you were right to change it for jinja_g_macros though as that uses postgres 👍 )

test/fixtures/templater/jinja_f/jinja.yml
```diff
@@ -140,7 +140,7 @@ file:
...
+                keyword: FLOAT64
```
jpy-git, line 143: set this back to `data_type_identifier` as this test is ANSI

jpy-git (commented):
@kayman-mk you didn't need to change `data_type_identifier` in `jinja_f` as that test isn't using postgres. The tests will pass once that's fixed

jpy-git (approved):
Looks great, thanks for adding this @kayman-mk! 💪 LGTM 🚀 ⭐

jpy-git:
@kayman-mk I can't merge your PR as it's out of date with the base branch. Can you update your branch please?

When you create a PR there's a tickbox to allow edits by maintainers that's on by default. It's best to leave that ticked for future PRs so we can avoid the hassle when merging:
image

PR #2124, opened 2021-12-16
Add ability to parse multiple GO/semicolon delimiters

### LINKED ISSUE #2009: TSQL: PRS error: multiple GO statements are not parsed
Please add possibility to have multiple GO statements parsed in the TSQL scripts.

## Expected Behaviour
no parsing errors

## Observed Behaviour
```
CREATE TABLE [sch1].[table1] (
    [_id]         BIGINT         NOT NULL,
    [_date_from]     DATETIME       NULL,
    [_date_to]       DATETIME       NULL,
    [load_id]                 INT            NULL
);
GO

GO

GO
CREATE NONCLUSTERED INDEX [idx_table1_date_to]
    ON [dm_p3m].[position]([_id] ASC, [_date_to] ASC);


GO
CREATE CLUSTERED INDEX [idx_table1_date_from]
    ON [dm_p3m].[position]([_id] ASC, [_date_from] ASC);


```
```
==== parsing violations ====
L:   7 | P:   3 |  PRS | Line 7, Position 3: Found unparsable section:
                       | '\n\nGO\n\nGO\nCREATE NONCLUSTERED INDEX [idx_...'
```

## Dialect
TSQL

## Version
Python 3.9
SQLFluff 0.8.2

jpy-git:

Fixes #2009. Consecutive GO segments separating batch segments were not parsing and likewise for consecutive semicolon segments separating statement segments. These are both valid syntactically, consecutive delimiters in this case are just separating empty statemtents:
image

This is fixed by setting the delimiter to one or more of itself. It gives the correct parse tree and works in the tests.

Removed an old test that was checking that consecutive semicolons raise a parse error, as they shouldn't raise a parse error.

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/parser/helpers.py
+++ b/src/sqlfluff/core/parser/helpers.py
@@ -28,7 +28,7 @@ def check_still_complete(
     initial_str = join_segments_raw(segments_in)
     current_str = join_segments_raw(matched_segments + unmatched_segments)
 
-    if initial_str != current_str:
+    if initial_str != current_str:  # pragma: nocover
         raise SQLParseError(
             f"Could not parse: {current_str}",
             segment=unmatched_segments[0],
--- a/src/sqlfluff/core/parser/helpers.py
+++ b/src/sqlfluff/core/parser/helpers.py
@@ -28,7 +28,7 @@ def check_still_complete(
     initial_str = join_segments_raw(segments_in)
     current_str = join_segments_raw(matched_segments + unmatched_segments)
 
-    if initial_str != current_str:  # pragma: nocover
+    if initial_str != current_str:  # pragma: no cover
         raise SQLParseError(
             f"Could not parse: {current_str}",
             segment=unmatched_segments[0],
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -509,7 +509,7 @@ class FileSegment(BaseFileSegment):
     # going straight into instantiating it directly usually.
     parse_grammar = Delimited(
         Ref("StatementSegment"),
-        delimiter=Ref("DelimiterSegment"),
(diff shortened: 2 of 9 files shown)


jpy-git:
@barrywhart should be good to review 👍 
Some reason an unrelated mypy error is being raised for L003.py in all PRs today so investigating that.

jpy-git:
Fixed the mypy error in #2125

src/sqlfluff/core/parser/helpers.py
```diff
@@ -28,7 +28,7 @@ def check_still_complete(
     initial_str = join_segments_raw(segments_in)
     current_str = join_segments_raw(matched_segments + unmatched_segments)
 
-    if initial_str != current_str:
+    if initial_str != current_str:  # pragma: nocover
```
barrywhart, line 31: Was this code previously covered by a test? Is it possible this `if` could be removed?

barrywhart:
There's a linter error:
```
./test/core/parser/parse_test.py:67:1: D103 Missing docstring in public function
```

barrywhart (approved):
The changes look good. One question about a "no cover" that was added. Also (noted elsewhere), the linter is complaining about one of the tests.

src/sqlfluff/core/parser/helpers.py
jpy-git: @barrywhart so it was originally being hit by the test I removed:
image
which shouldn't have been a test in the first place.

It's one of those awkward cases where this isn't getting touched anywhere else other than this random test case, and I can't immediately think of a situation that would trigger it, but think it's probably safest to to leave it in there as it may catch some other parsing issues that aren't in the tests?

barrywhart: That makes sense -- I was just curious. 👍🏽

jpy-git: @barrywhart no worries, took the pragma(tic) approach 👍 Want to do the honours merging? 😛

Issue #1799, opened 2021-10-31
New rule: Nested CASEs with conditions on the same column


iajoiner:
When conditions in iterative `CASE` statements are on the same column they should be simplified without any cost i.e. 

```sql
CASE
  WHEN species = 'Cat' THEN 'Meow'
  ELSE
  CASE
     WHEN species = 'Dog' THEN 'Woof'
  END
END as sound
```

is really just

```sql
CASE
    WHEN species = 'Cat' THEN 'Meow'
    WHEN species = 'Dog' THEN 'Woof'
END as sound
```

jpy-git:
Is there ever a situation where nested CASE statements are needed?

Since the CASE statement is effectively an IF/ELIF/ELSE structure you could reduce all nested case statements?
i.e.
```SQL
CASE
  WHEN species = “Cat” THEN “Meow”
  ELSE
  CASE
     WHEN number_of_legs = 8 THEN “<spider_sounds>”
  END
END as sound
```
is equivalent to
```SQL
CASE
  WHEN species = “Cat” THEN “Meow”
  WHEN number_of_legs = 8 THEN “<spider_sounds>”
END as sound
```

So we can make a rule to stop the use of nested case statements and replace them as above.

barrywhart:
Nested `CASE` statements could be used as an alternative to `AND`, e.g.:
```
CASE
    WHEN number_of_sides = 4 THEN
        CASE
        WHEN side_lengths = "same" THEN
            CASE
            WHEN angle = 90 THEN "square"
            ELSE "rhombus"
        ELSE
            CASE
            WHEN angle = 90 THEN "rectangle"
            ELSE "parallelogram"
        END
    END
END
```

barrywhart:
I've got a preliminary implementation of this rule in a [branch](https://github.com/barrywhart/sqlfluff/tree/bhart-issue_1799_simplify_nested_cases). It builds on some unmerged work (PR #2126), so this new rule won't be completed until that other PR reviewed and merged. 

The approach I took for implementing it is to look for `CASE`/`ELSE` where the thing inside the `ELSE` is itself a `CASE` statement. IIUC, it should always be safe to "unnest" (i.e. move) the `WHEN` clauses of the inner `CASE` to the end of the outer `CASE`.

PR #2126, opened 2021-12-16
Simplify rule creation by adding a functional API to RuleContext

barrywhart:
Makes progress on #1808 

* Updates many existing rules to use the new classes (see changes to `rules.rst` for the current list).
* Updates `L041` to use `awesome-pattern-matching` package

These changes generally make the rule code:
* Shorter
* Less complex
* More declarative & functional

This style of code should be easier for heavy SQL users and mathematicians to author. 👍🏽💪🏽 

Reviewers may also wish to look at [this branch](https://github.com/barrywhart/sqlfluff/tree/bhart-issue_1799_simplify_nested_cases), which implements a new rule (suggested in #1799) using the new API. I thought it'd be a good test case to create a new rule, not just rewrite existing ones. Note that it's a potentially pretty tricky rule, but it lacks some of the typical "code smells" of past tricky rules, e.g. complex loops, lots of `if` statements, heavy use of segment indices.

### Pull Request checklist

  - Other.


barrywhart:
@jpy-git, @tunetheweb: Please take a preliminary look at this. I'm not looking for a typical "pre-merge" review. This is more of a design review:
* Does this look like a good start towards making it easier to write rules?
* Do you see any areas where the API might be improved? 
* Are there other rules we should look at updating to use this new API, as a kind of "beta test"? I can update them, or I'd be happy for you to update them instead.

Note that some of the functions on the new classes are just stubs (they raise `NotImplementedError`). If/when we get to a more proper review, I can implement these functions or remove them for now (i.e. add them later when needed).

src/sqlfluff/core/rules/base.py
```diff
@@ -221,6 +228,34 @@ class RuleContext:
...
+    @cached_property
+    def segment(self) -> "Segments":
+        """Returns a Segments object for context.segment."""
+        # Import here to avoid circular dependency.
+        from sqlfluff.core.rules.surrogates import Segments
+
+        return Segments(self.context.templated_file, self.context.segment)
+
+    @cached_property
+    def raw_stack(self) -> "Segments":
```
jpy-git, line 252: worth having a parent_stack method as well

src/sqlfluff/core/rules/base.py
```diff
@@ -221,6 +228,34 @@ class RuleContext:
...
+    """Provides access to surrogate objects that simplify writing rules."""
+
+    def __init__(self, context: RuleContext):
+        self.context = context
+
+    @cached_property
+    def segment(self) -> "Segments":
+        """Returns a Segments object for context.segment."""
+        # Import here to avoid circular dependency.
+        from sqlfluff.core.rules.surrogates import Segments
```
jpy-git, line 247: Feels a little weird to have to do this. As far as I can see the only thing we're importing from this file in `sqlfluff.core.rules.surrogates` is `LintFix`. Feels like `LintFix` and `LintResult` should be in their own file anyway. Would moving those to a separate file stop this?

src/sqlfluff/core/rules/surrogates/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Modules in this directory provide surrogates.
+
+In this case, the "surrogates" are classes that provide a higher-level API for
+writing rules than working with the related, lower-level classes.
+"""
+from sqlfluff.core.rules.surrogates.segments import Segments  # noqa: F401
```
jpy-git, line 6: you want a
```python
__all__ = ("Segments",)
```
here to avoid F401

src/sqlfluff/core/rules/surrogates/segments.py
```diff
@@ -0,0 +1,93 @@
...
+    #     select_if=Optional[Sequence[Predicate]],
+    #     loop_while=Optional[Sequence[Predicate]],
+    # ) -> "Segments":
+    #     """Retrieve range/subset."""
+    #     raise NotImplementedError
+
+    def delete(self, *predicates: Predicate) -> Sequence["LintFix"]:
+        """Return LintFix objects to delete segments that satisfy predicates."""
+        cp = _CompositePredicate(*predicates)
+        return [LintFix("delete", s) for s in self.segments if cp(s)]
```
jpy-git, line 69: Would we also have `create_before`, `create_after`, and `edit` methods? They all anchor to a single segment and add something new so don't really make as much sense if there are multiple segments in this class?

Curious to see what the plan is here as it would be good to be consistent across the different LintFix types. If we didn't add those as methods to this class then this method name should probably be more explicit like `get_delete_fixes` so that it's clear in the rule code that this just wraps the delete fix type when we have mixes of the different types.

i.e.
```
edit_fix = LintFix("edit", <some_segment>, [<edit_segments>,...])
delete_fix = segments.get_delete_fixes()
```
is more clear when mixing

src/sqlfluff/rules/L050.py
```diff
@@ -91,34 +71,29 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+            and not segment.all(lambda s: s.is_expandable)
             # It is possible that a template segment (e.g. {{ config(materialized='view') }})
             # renders to an empty string and as such is omitted from the parsed tree.
             # We therefore should flag if a templated raw slice intersects with the
             # source slices in the raw stack and skip this rule to avoid risking
             # collisions with template objects.
-            if self._potential_template_collision(context):
-                return None
-
+            and not raw_stack.raw_slices.any("templated")
```
jpy-git, line 93: Interesting that this doesn't seem to be triggering the dbt test case with the leading invisible `{{ config(materialized="view") }}` that was causing me issues when I first made this rule. But if it works then that's great haha 🤷

jpy-git: Oh nevermind, read it as any templated segments in the raw stack rather than raw slices as you'd added. So it encapsulates the same logic I originally had. Nice!

jpy-git (commented):
@barrywhart Like it a lot! 😎  Added my initial thoughts/questions. See if there's anything you think we should change off the back of that and then I can maybe try applying this to L009 to get a better idea of the user experience 👍

src/sqlfluff/core/rules/base.py
barrywhart: Probably. The circular dependency is between RuleContext -> Segments -> LintFix.

src/sqlfluff/core/rules/base.py
barrywhart: 👍  Easy to add.

src/sqlfluff/core/rules/surrogates/segments.py
barrywhart: I haven't thought that far ahead re: the other edit types. Was just experimenting to see what it "felt? like to have a method create the "delete" `LintFix` objects rather than having to explicitly create it yourself. I think it's an improvement, so I hope the other ones are a natural fit as well. I think **one** of the usability issues in the current way of doing things is that you have to create and work with various unrelated objects. It feels disjointed, verbose,  and it disrupts your "flow" when writing rules.

If we can connect these together so you're mostly working with one object plus others that it returns, it's a simpler, more unified experience.

src/sqlfluff/rules/L050.py
barrywhart: Yes, exactly the same, just... shorter code. 🎉

src/sqlfluff/core/rules/surrogates/segments.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # def select(
+    #     self,
+    #     start_seg: Optional[BaseSegment] = None,
+    #     stop_seg: Optional[BaseSegment] = None,
+    #     select_if=Optional[Sequence[Predicate]],
+    #     loop_while=Optional[Sequence[Predicate]],
+    # ) -> "Segments":
+    #     """Retrieve range/subset."""
+    #     raise NotImplementedError
```
jpy-git, line 64: I assume this acts like a filter so I can say something like "return me the whitespace segments before next code segment" using something like:
```python
raw_stack = context.surrogates.raw_stack
leading_whitespace = raw_stack.select(select_if=[attr("is_whitespace")], loop_while=[attr("is_code")])
```

jpy-git: One thing the above example got me thinking about is that it's hard to negate these predicates.

So if I want to select NOT whitespace I can't do
`raw_stack.select(select_if=[not attr("is_whitespace")])`
because that's not a callable.
And we don't have a `is_not_whitespace` attribute.

So I would have to make my own function to move the not within the function:
`raw_stack.select(select_if=[lambda x: not x.is_whitespace])`
which would work, but feels less natural than a list comprehension

src/sqlfluff/rules/L050.py
```diff
@@ -91,34 +72,29 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
             # Non-whitespace segment.
-            (context.segment.name not in whitespace_set)
+            not segment.all(*whitespace_types)
             # We want first Non-whitespace segment so
             # all preceding segments must be whitespace
             # and at least one is not meta.
-            and all(segment.name in whitespace_set for segment in context.raw_stack)
-            and not all(segment.is_meta for segment in context.raw_stack)
+            and raw_stack.all(*whitespace_types)
+            and not raw_stack.all(attr("is_meta"))
```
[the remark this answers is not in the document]
barrywhart: Makes sense, yeah. I initially wrote this using `lambda`. We can eventually build out more of this. The segment classes don't change much. Realistically, we may need more documentation about the segment types and properties for both the old or new ways of writing rules.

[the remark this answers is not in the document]
barrywhart: One issue around the properties is that properties vary by segment type; they don't all exist on `BaseSegment` or `RawSegment`.

To me, an "MVP" of this work will eliminate a lot of low-level coding (loops, conditionals, helper functions) for simple to medium complicated rules. Definitely open to adding other things that help with autocomplete, etc., possibly in future PRs.

src/sqlfluff/core/rules/base.py
barrywhart: Making this change now. Seems to fix the circular dependency -- great suggestion!

src/sqlfluff/core/rules/base.py
barrywhart: Adding now.

src/sqlfluff/core/rules/surrogates/segments.py
barrywhart: It should be easy to add a `not()` function to wrap and negate another predicate. I'll start a checklist in the PR description with a note to add this. 👍

src/sqlfluff/core/rules/surrogates/segments.py
barrywhart: Thinking more about this. As you say, the other edit types only apply to a single segment. Therefore, they don't make sense as instance methods. We could add them as class or static methods for completeness. Think that's worth doing now? Or wait and think about it more?

src/sqlfluff/rules/L050.py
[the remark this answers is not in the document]
barrywhart: The "common predicates" idea sounds good. I'll add something like that for `is_meta` and `is_expandable`. Those appear to be the only two boolean properties on `BaseSegment`. I'd like to think about it more before doing anything for the subclasses.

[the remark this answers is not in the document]
jpy-git: makes sense, I think it makes sense to add the not and attr predicates there too 👍

src/sqlfluff/core/rules/surrogates/segments.py
jpy-git: tbh I feel like they don't belong in that class. To me segments feels like it should just be for filtering and querying lists of segments.

So maybe we should just let LintFix take in Segments type?
e.g.
```python
raw_stack = context.surrogates.raw_stack
LintFix("delete", s.select(select_if=[cp.is_whitespace], loop_while=[cp.not(cp.is_whitespace)])
```
allows you to filter segments easily and then LintFix delete is still consistent with the other methods.

jpy-git: I think an easy and useful way to do this is to make `Segments` iterable and return BaseSegments.

So
```python
[s for s in Segments(<input_segments>).select(<some_filter_condition>)]
```
easily gets you back to a list of Base/Raw segments
or more concisely
```python
list(Segments(<input_segments>).select(<some_filter_condition>)): List[BaseSegment]
```

and you can just type LintFix input as `Iterable[BaseSegment]` to avoid worrying about manually converting back to segments in rule code.

jpy-git: Basically keeping it a filter language. It can still have logic to add in extra segments etc but just not build the LintfFixes

src/sqlfluff/core/rules/surrogates/segments.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def last(self, *predicates) -> Optional[BaseSegment]:
+        """Returns the last segment (if any) that satisfies the predicates."""
+        cp = _CompositePredicate(*predicates)
+        for s in reversed(self):
+            if cp(s):
+                return s
+        return None
+
+    def apply(self, fn: Callable[[BaseSegment], Any]) -> List[Any]:
```
jpy-git, line 76: Do we maybe want a `reversed` method?

```python
def reversed(self) -> "Segments":
    return reversed(self)
```
```python
raw_stack = context.surrogates.raw_stack
reversed_raw_stack = raw_stack.reversed()
```
^ Feel like that's a fairly common thing someone may do.
(N.B. tuples are immutable so this returns a new instance of Segments i.e. `id(reversed_raw_stack) != id(raw_stack)`, which I like)

jpy-git: Another thing that would be worth checking:
Since we inherit from tuple I assume we also inherit the __add__ method?
So can we do something like `segments_a + segments_b`? 
Could be nice

jpy-git: following on from above, I know I've done the thing before where I've made a "complete_stack" which could be something like the raw_stack + plus the current segment. Could we also have an append method for something like:
```python
raw_stack = context.surrogates.raw_stack
complete_stack = raw_stack.append(context.segment)
```
I know tuple doesn't have an append method itself, but you can effectively mock it with __add__
```python
def append(self, segment):
    return self + self.__class__(segment)
```

src/sqlfluff/core/rules/surrogates/segments.py
```diff
@@ -0,0 +1,130 @@
...
+                    child_segments.append(child)
+        return Segments(self.templated_file, *child_segments)
+
+    def first(self, *predicates) -> Optional[BaseSegment]:
+        """Returns the first segment (if any) that satisfies the predicates."""
+        cp = _CompositePredicate(*predicates)
+        for s in self:
+            if cp(s):
+                return s
+        return None
```
jpy-git, line 66: feel like by default if no predicates are defined then this should just return the first segment in the tuple?

src/sqlfluff/core/rules/surrogates/segments.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def last(self, *predicates) -> Optional[BaseSegment]:
+        """Returns the last segment (if any) that satisfies the predicates."""
+        cp = _CompositePredicate(*predicates)
+        for s in reversed(self):
+            if cp(s):
+                return s
+        return None
```
jpy-git, line 74: same comment here as for first

src/sqlfluff/rules/L013.py
```diff
@@ -42,33 +43,36 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+                # No fixes, because we don't know what the alias should be,
+                # the user should document it themselves.
+                if self.allow_scalar:  # type: ignore
+                    # Check *how many* elements there are in the select
+                    # statement. If this is the only one, then we won't
+                    # report an error.
+                    num_elements = sum(
+                        e.is_type("select_clause_element")
+                        for e in context.parent_stack[-1].segments
+                    )
```
jpy-git, line 70: could this be converted to use the new API:
```python
immediate_parent = context.surrogates.parent_stack.last()
num_elements = len(immediate_parent.children("select_clause_element"))
```
(provided `last` is made to work as suggested in previous comment)

Also think `len` should work here as we inherit from tuple which is quite intuitive.

jpy-git (commented):
A couple more random thoughts to add, shaping up nicely!

Will have a more in depth look at the edited rules tomorrow 😄

src/sqlfluff/core/rules/surrogates/segments.py
jpy-git: related LintFix improvement in #2133

src/sqlfluff/core/rules/surrogates/segments.py
barrywhart: That should happen already, because `_CompositePredicate.__call__()` returns `True` if there are no predicates. `return None` is for the case where no predicate is satisfied. I'll add a comment to clarify this.

src/sqlfluff/core/rules/surrogates/segments.py
barrywhart: I'll add the `reversed()` method now. I'll have to check on `add()`. We probably inherit it, but I'm not 100% sure it'll work correctly. From my experience with operator overloading in other languages, it seems likely that either `my_tuple + segments` or `segments + my_tuple` will return a bare `tuple`, not a `Segments` object.

`append()` will take some thought. With lists, it modifies the list **in place**. As written above, it looks like it would return a new `Segments` object. I don't think we'll be able to modify the base tuple in place, because tuples are immutable. It may be possible to inherit from `list` instead.

I'll add these to the "to do" list.

src/sqlfluff/rules/L013.py
barrywhart: Good idea! That _almost_ works. I tested this, and the issue is, the first line returns a `BaseSegment`, not a `Segments` object, then the second line fails because we're working with the wrong type of object.

Perhaps we should provide an alternate version of `last()` and `first()` which returns a `Segments`. Or I could change the existing functions? It seems like both versions are useful in certain situations. 🤔

src/sqlfluff/core/rules/surrogates/segments.py
jpy-git: yeah I know tuples are immutable but we don't need to modify in place.

assigning to a new Segments instance is fine:
`complete_stack = raw_stack.append(context.segment)`

src/sqlfluff/rules/L013.py
jpy-git: so my suggestion here would be to have `last()` and `first()` return a single length `Segments` instance here for consistency (no match to predicates is just a zero length `Segments` instance).
Same goes for `apply()` it should return a `Segments` instance: `return Segments([fn(s) for s in self])`

You can already get a list of base segments via list(Segments) or iterating over them if someone explicitly wants them.

src/sqlfluff/core/rules/surrogates/segments.py
barrywhart: I agree that'll work. I think it would be a confusing API because the name matches the similar function on `list`, which does modify in place. Imagine the poor rule developer: they assume it modified the object in place, so they didn't assign the result to a variable. The result is simply discarded, and they are left with the unmodified object. 🤯

I think the right solution is to inherit from `list` instead. I'll make that change when I have a chance. 👍

src/sqlfluff/rules/L013.py
barrywhart: Ok, I changed `first()` and `last()` to return a `Segments` instead, and I updated L013 with your suggested change, which works now. 🎉 

We can't change `apply()` as you suggest, because `apply()` returns the result of calling an arbitrary function on the segments, so this may return strings, numbers, whatever; almost certainly **not** segment objects.

The name `apply()`. comes from a very similar Pandas function: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.apply.html

jpy-git: ahh right makes sense

It's not exactly like pandas because pandas would still return a `pd.Series` object type. It's just the dtype within the column that gets changed rather than the data structure.
i.e.
```python
import pandas as pd

a = pd.DataFrame({"col1": [0, 1, 2]})  # pd.DataFrame
a["col1"] # pd.Series, with `int` as it's `dtype`
a["col1"].apply(lambda x: str(x))  # pd.Series, with `str` as it's `dtype`, but still a pd.Series
```

But makes sense in our case to not return a `Segments` object here

src/sqlfluff/rules/L009.py
```diff
@@ -85,44 +86,39 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
-        if len(eof_newline_segments) == 1:
+        # Include current segment for complete stack and reverse.
+        parent_stack: Segments = context.surrogates.parent_stack
+        raw_stack: Segments = context.surrogates.raw_stack
+        complete_stack = raw_stack.append(context.segment)
+        reversed_complete_stack = complete_stack.reversed()
+
+        # Find the trailing newline segments.
+        trailing_newlines = reversed_complete_stack.select(
+            select_if=["newline"], loop_while=[lambda x: x.is_whitespace or x.is_meta]
```
jpy-git, line 97: Forgot to mention in my PR for this, but maybe this also highlights that we want `and()` & `or()` predicates as well, like we're suggesting with `not()`?

So we can write this line as:
```
select_if=["newline"], loop_while=[or(segpred.is_whitespace, segpred.is_meta)]
```

barrywhart: WIll do!

src/sqlfluff/core/rules/functional/segments.py
```diff
@@ -0,0 +1,140 @@
...
+            return False
+
+        for p in self.other:
+            if isinstance(p, type):
+                # Check segment class
+                if not isinstance(segment, p):  # pragma: no cover
+                    return False
+            elif not p(segment):  # Arbitrary function
+                return False
+        return True
```
jpy-git, line 164: @barrywhart just a general thought from playing around further with the predicates:

We've added in the `not_()`, `and_()`, `or_()` predicates to allow us to combine multiple function predicates. However, predicates also allows types to be supplied as strings. The issue this raises for me is that we still can't combine a str type predicate and a function predicate using these boolean predicates.
e.g. can't do something like `segpred.and_("object_reference", segpred.is_expandable)`.

Think we should refine our definition of predicates to solve this and simplify things a bit:
1. make predicates only take in `Callable[[BaseSegment], bool]`.
2. replace the types logic with a common predicate:
```python
def is_type(
    *seg_type: str
) -> Callable[[BaseSegment], bool]:  # pragma: no cover
    """Returns a function that determines if the segment is of the supplied types."""
    def _(segment: BaseSegment):
        return segment.is_type(*seg_type)
    return _
```
(use a closure to generate the predicate)
3. the types can easily be combined with other predicates and keeps the API consistent: `segpred.and_(segpred.is_type("object_reference"), segpred.is_expandable)`

Let me know what you think, but I quite like the consistency 👍

jpy-git: We can also drop the composite predicates classes in favour of this as well

barrywhart: Sounds good to me. In my early commits, I didn't think through the usage really carefully, was just "sketching" something to try out.

src/sqlfluff/core/rules/functional/segments.py
```diff
@@ -0,0 +1,118 @@
...
+    def __add__(self, segments) -> "Segments":
+        return Segments(self.templated_file, *super().__add__(segments))
+
+    def __radd__(self, segments) -> "Segments":
+        return Segments(self.templated_file, *list(segments).__add__(list(self)))
+
+    def all(self, predicate: Optional[Callable[[BaseSegment], bool]] = None) -> bool:
+        """Do all the segments match?"""
+        for s in self:
+            if predicate and not predicate(s):
```
jpy-git, line 33:
```suggestion
            if predicate is not None and not predicate(s):
```
small nitpick here: when comparing to None as a default argument you should use the explicit `predicate is not None` (stops any weirdness if the user puts `predicate = 0` or `predicate = []` etc as they are falsey also).

There's a couple of other predicates that need this changing also

src/sqlfluff/core/rules/functional/raw_file_slices.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(self, templated_file: TemplatedFile, *raw_slices: RawFileSlice):
+        super().__init__(raw_slices)
+        self.templated_file = templated_file
+
+    def all(self, *predicates: Callable[[RawFileSlice], bool]) -> bool:
+        """Do all the raw slices match?"""
+        for s in self:
+            if predicates and not any(p(s) for p in predicates):
+                return False
+        return True
```
jpy-git, line 27:
```suggestion
    def all(self, *predicates: Optional[Callable[[RawFileSlice], bool]] = None) -> bool:
        """Do all the raw slices match?"""
        for s in self:
            if predicates is not None and not any(p(s) for p in predicates):
                return False
        return True
```
this should be made consistent with `segments.py`

jpy-git (commented):
some small tweaks to make

src/sqlfluff/core/rules/functional/segments.py
barrywhart: Will do!

src/sqlfluff/core/rules/functional/raw_file_slices.py
barrywhart: Will do!

alanmcruickshank (commented):
I haven't reviewed the docs of this - just the intent of what we're trying to achieve and I think it's great. Very interesting, but should also be very powerful and expressive. I agree that to many rules that this will be a more compact way of writing what's going on.

Two things we should keep an eye on while doing this:
- docs. Which it looks like we are doing.
- performance. It would be really easy to introduce unintentional slowdowns while introducing the functional approach - I don't think it _needs_ to, but it would be very easy by accident. We should keep an eye on runtime effects of the changes in the process.

barrywhart:
@alanmcruickshank: A note about performance... the classes inherit from `list`. We initially inherited from `tuple`, and I still wonder if that'd be better, because tuples are read only, making the objects safer to cache and reuse across rules.

@jpy-git: Do you have a strong feeling for `list` versus `tuple`? I'm tempted to change it back to `tuple`.

jpy-git:
@barrywhart I'm fine with tuple. the only reason it was added was because we need to append to Segments: https://github.com/sqlfluff/sqlfluff/pull/2126#discussion_r771763076

barrywhart:
Ok, I switched back to `tuple`, which allows the potential for caching the `Segments` instances across rules. The `Segments` class supports `__add__` and `__radd__` (in other words, `+` with itself and with tuples), which provides roughly equivalent functionality for tuples. Note that tuples are by definition read only, so `append()` (which, by convention, modifies a list-like collection in place) is not supported.

src/sqlfluff/rules/L025.py
```diff
@@ -2,6 +2,7 @@
 
 from sqlfluff.core.rules.base import LintFix, LintResult
 from sqlfluff.core.rules.doc_decorators import document_fix_compatible
+from sqlfluff.core.rules.functional import Segments
```
jpy-git, line 5:
```suggestion
from sqlfluff.core.rules.functional import Segments
import sqlfluff.core.rules.functional.segment_predicates as sp
```

src/sqlfluff/rules/L025.py
```diff
@@ -62,22 +63,18 @@ def _lint_references_and_aliases(
...
-                            # Stop once we reach an other, "regular" segment.
-                            break
+                to_delete = (
+                    Segments(*alias.from_expression_element.segments)
+                    .reversed()
+                    .select(
+                        start_seg=alias.alias_expression,
+                        # Stop once we reach an other, "regular" segment.
+                        loop_while=lambda seg: seg.name in ("whitespace", "newline")
+                        or seg.is_meta,
```
jpy-git, line 74:
```suggestion
                        loop_while=sp.or_(sp.is_whitespace, sp.is_meta),
```

jpy-git (commented):
@barrywhart think we can use the functional API here?

src/sqlfluff/rules/L025.py
barrywhart: I'll make this change. There are some subtle differences with some of the whitespace checks (spaces & tabs, or also newlines), but I confirmed this works. 👍

jpy-git: Yeah is_whitespace includes newlines 😊

barrywhart:
@jpy-git, @alanmcruickshank, @tunetheweb: This is ready for review now. I could do more, but there's already a lot here, and I'd like another look before investing more time in any particular direction.

If this looks good, I'd say my biggest interest going forward is to explore more of the possibilities of the `awesome-pattern-matching` package. I think if we learn how to use this well and potentially simplify how to use it in SQLFluff, it plus the new stuff in `functional/` will be a really big improvement in the rule writing experience.

docs/source/rules.rst
```diff
@@ -33,14 +33,77 @@ Implementation
...
+* `L025`
+* `L031`
+* `L035`
+* `L036`
+* `L038`
+* `L042`
+* `L043`
+* `L047`
+* `L050`
+* `L052`
```
alanmcruickshank, line 65: I'm not sure it makes sense to keep this list given how many you've converted. It will just end up out of date. If we want to still give some measure of scale, but without having to maintain the list, it might be better to say something like: `...at 2021-12-30, 17 rules use this method.`

docs/source/rules.rst
```diff
@@ -33,14 +33,77 @@ Implementation
...
-evaluate the position of the given segment in relation to its neighbors,
-and that the segment which finally "triggers" the error, should be the one
-that would be corrected OR if the rule relates to something that is missing,
-then it should flag on the segment FOLLOWING, the place that the desired
-element is missing.
+.. autoclass:: sqlfluff.core.rules.functional.raw_file_slices
+   :members:
+
+.. autoclass:: sqlfluff.core.rules.functional.raw_file_slice_predicates
+   :members:
```
alanmcruickshank, line 77: These autodoc commands don't seem to render properly at the moment: 
image

alanmcruickshank (changes requested):
Code looks great - docs aren't building properly at the moment but otherwise love it 👍

docs/source/rules.rst
barrywhart: Will do.

docs/source/rules.rst
barrywhart: Fixes incoming...

barrywhart:
@alanmcruickshank: Thanks for checking the docs! I hadn't built and previewed those yet. Should be good now.

docs/source/rules.rst
alanmcruickshank: @barrywhart - can you push your most recent changes? I can't see them on the branch. I'll give them a spin when they're up.

barrywhart:
@alanmcruickshank: Okay, the fixes are up now. (Earlier, I had accidentally pushed them to the wrong branch.)

jpy-git:
@barrywhart I'm conscious that things are still getting added to this PR and we still haven't reviewed the awesome-pattern-matching dependency.

Looking at the repo it seems like quite a small project?
image
Only 700 downloads in the past month: https://pypistats.org/packages/awesome-pattern-matching

This issue [here](https://github.com/scravy/awesome-pattern-matching/issues/3) mentions there may be a lot of overlap with the matching introduced in Python 3.10
(relevant PEP: https://www.python.org/dev/peps/pep-0622/)
So maybe there's value in seeing if there's an available backport for that instead?

I feel like maybe it's best to introduce the functional API and it's rule changes/documentation in this PR, as there are no concerns there and it's been pretty thoroughly reviewed, and separate the pattern matching stuff into it's own PR afterwards. They're not really related, other than both touching rules, and would be good to have any discussion on apm separated for future reference. I can see that the functional API has been linked to several issues so we should get that merged to unblock them.

barrywhart:
@jpy-git: That's definitely valid feedback. I'm happy to remove the pattern matching stuff for now.

I did take a look at the new built-in pattern matching in Python 3.10, and there is a backport (although the backport looks new and kind of unofficial). From my evaluation, the built-in pattern matching is less powerful, e.g. it doesn't support patterns with both leading and trailing "don't care" sections. It's also unclear if it supports matching across multiple levels of a tree structure.

OTOH:
* awesome-pattern-matching is definitely immature, e.g. current lack of support for nested multi-segment patterns (I asked about it here: https://github.com/scravy/awesome-pattern-matching/issues/8#issuecomment-1003528733)
* It's also unclear how often we'd need the additional power of awesome-pattern-matching. Until it matures a bit, I often find try more of a distraction than a boon. *If* you get something working, the code looks nice, but you usually encounter some odd behavior (bugs) along the way.

TL;DR I'll remove it from this PR. I may keep tinkering with it, possibly contributing to it. Among the packages out there, it comes closest to my dream of "regex-like power applied to arbitrary objects, plus support for patterns that span levels of a tree". (AFAICT, most of them aren't even trying to do one or both of these.) Part of the appeal is that the patterns closely resemble SQLFluff dialect definitions, so could be very approachable and intuitive for contributors.

jpy-git:
Makes sense, I def like the concept so it's worth investigating further 😄

barrywhart:
@jpy-git: Ok, I've removed all references to and usage of `awesome-pattern-matching`. Ready for a closer review?

I created an issue, #2221, to revisit the package in the future, and for reference, that issue also includes the rewritten rules and documentation that was removed from this PR.

mypy.ini
```diff
@@ -23,3 +23,6 @@ ignore_missing_imports = True
 
 [mypy-tblib.*]
 ignore_missing_imports = True
+
+[mypy-apm.*]
+ignore_missing_imports = True
```
jpy-git, line 28: remove

jpy-git (commented):
small mypy cleanup otherwise good 👍

mypy.ini
barrywhart: Done!

jpy-git (approved):
Looks good! 👍 Think we can get this merged if no one else has any objections? 🚀

barrywhart:
I went ahead and merged. I hope that was okay? 😬 @alanmcruickshank

PR #2127, opened 2021-12-16
Fix mypy type raise in L003

barrywhart:

With the new 0.920 release of mypy today it is flagging a new type issue in L003.

The issue is actually a valid one to raise: A tuple of BaseSegment is being appended to a tuple of RawSegment to create a complete raw stack similar to the complete stack I use in L050 or L009, however due to the naming convention I used those rules I could easily type the complete stack. In this case since it appends to the already defined raw stack you can't define the type at that assignment.

A full fix would require some variable renaming and slight rework of L003 so given that this issue is blocking all unrelated PRs I've just inline ignored it.

N.B. If using an earlier version of mypy this issue is not raised, but it will now raise for unused type ignore. Please update your mypy to avoid this: `tox -e mypy -r`, `pip install --upgrade mypy`, or `tox -e pre-commit -- install` depending on which you're using.

No, unblocks other PRs.

### Pull Request checklist

  - Other.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
     hooks:
     -   id: black
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.910-1
+    rev: v0.920
     hooks:
     -   id: mypy
         args: [--ignore-missing-imports]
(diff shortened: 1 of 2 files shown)


jpy-git (approved):
Looks good 👍

PR #2128, opened 2021-12-16
Normalise input newlines

### LINKED ISSUE #797: Feature request: Enable use of unix-style line endings when running sqlfluff fix command on Windows

jpy-git:

Fixes #797. After investigation (described in the issue) it seems most appropriate to not handle multiple newlines and instead add step to ensure all newline inputs to sqlfluff are normalised to unix-style line endings for consistency. Users wanting newline conversion should consider the easy to use [mixed-line-ending](https://github.com/pre-commit/pre-commit-hooks/blob/master/pre_commit_hooks/mixed_line_ending.py) pre-commit hook or the `dos2unix` command. 

Currently we load from files using universal newlines, but we don't do this for test input. Therefore I have added a conversion function to the linter to ensure consistency regardless of how the SQL is supplied.

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -16,7 +16,7 @@ from typing import (
 )
 
 import pathspec
-
+import regex
 from tqdm import tqdm
 
 from sqlfluff.core.errors import (
@@ -115,6 +115,11 @@ class Linter:
         file_config.process_raw_file_for_config(raw_file)
         # Return the raw file and config
         return raw_file, file_config, encoding
+
+    @staticmethod
+    def _normalise_newlines(string: str) -> str:
+        """Normalise newlines to unix-style line endings."""
+        return regex.sub(r"\n|\r\n|\r", "\n", string)
 
     @staticmethod
     def _lex_templated_file(
@@ -605,6 +610,11 @@ class Linter:
 
         # Start the templating timer
         t0 = time.monotonic()
+
+        # Newlines are normalised to unix-style line endings (\n).
+        # The motivation is that Jinja normalises newlines during templating and
+        # we want consistent mapping between the raw and templated slices.
+        in_str = self._normalise_newlines(in_str)
 
         if not config.get("templater_obj") == self.templater:
             linter_logger.warning(
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -119,7 +119,7 @@ class Linter:
     @staticmethod
     def _normalise_newlines(string: str) -> str:
         """Normalise newlines to unix-style line endings."""
-        return regex.sub(r"\n|\r\n|\r", "\n", string)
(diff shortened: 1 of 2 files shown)


tunetheweb:
Isn’t there a concern here that `sqlfluff fix` will now effectively rewrite line endings and lead to massive diffs every time a file needs a fix? At least until the code base is converted to unix style newlines in entirety?

Note I’m not talking about our code base (that I already converted) but any other users.

jpy-git:
@tunetheweb python already automatically does this when it reads in files so it's no different from what we already do, I'm just making it consistent for the API and CLI stdin.

jpy-git:
also when python writes out a file it automatically converts them back to the appropriate line ending for the system

jpy-git:
@barrywhart follow up PR to my post in #797

tunetheweb:
> @tunetheweb python already automatically does this when it reads in files so it's no different from what we already do, I'm just making it consistent for the API and CLI stdin.

I'm confused. If python already automatically does this, then why's there even an issue? Sorry if I'm missing the obvious here...

jpy-git:
@tunetheweb because it only does it when reading in from a file (i.e. `open("<some_sql_path>", "r") as f`), but not when we read the sql from either stdin in the CLI or via the string input to the API. So we should normalise those too for consistency

tunetheweb:
OK. Thanks for explaining.

src/sqlfluff/core/linter/linter.py
```diff
@@ -116,6 +116,11 @@ def _load_raw_file_and_config(
         # Return the raw file and config
         return raw_file, file_config, encoding
 
+    @staticmethod
+    def _normalise_newlines(string: str) -> str:
+        """Normalise newlines to unix-style line endings."""
+        return regex.sub(r"\n|\r\n|\r", "\n", string)
```
tunetheweb, line 122: Do we need this first one? Seems to be replacing `"\n"` with `"\n"`?

```suggestion
        return regex.sub(r"\r\n|\r", "\n", string)
```

tunetheweb (approved):
One small question but other than that this looks good.

src/sqlfluff/core/linter/linter.py
jpy-git: strictly no, but thought it was clearer to read as replacing all newlines with the unix newline (i.e. normalise), although in this case it's not like we'll ever change the newline replace style so happy either way.

I'll let you commit the suggestion and merge if you want

PR #2130, opened 2021-12-17
Lint spaces in qualified names

### LINKED ISSUE #1754: New Rule suggestion: flag spaces in qualified names
TSQL technically allows for spaces between the elements of a qualified name.  This is confusing and bad practice.  I assume it is also confusing and bad practice in other dialects, if it's even allowed.

We should be able to fix this automatically.

Anti-pattern:
```
SELECT [thistable] . [col]
FROM [thisdatabase] .     [thisschema]
   . [thistable]
```

Pattern:
```
SELECT [thistable].[col]
FROM [thisdatabase].[thisschema].[thistable]
```

jpers36:
Fixes #1754 

This rule change adds another type of unnecessary whitespace to L039.  Any whitespace or newline segments found within a reference segment type are removed.

Per @jpy-git 's suggestion this has been implemented as a change to an existing rule, but this could be spun out into its own rule if that's preferred.

Anti-pattern:

```
SELECT [thistable] . [col]
FROM [thisdatabase] .     [thisschema]
   . [thistable]

```
Pattern:

```
SELECT [thistable].[col]
FROM [thisdatabase].[thisschema].[thistable]
```


src/sqlfluff/rules/L039.py
```diff
@@ -66,4 +66,32 @@ def _eval(self, context: RuleContext) -> Optional[List[LintResult]]:
...
+                seg.is_type("object_reference")
+                or seg.is_type("table_reference")
+                or seg.is_type("schema_reference")
+                or seg.is_type("database_reference")
+                or seg.is_type("index_reference")
+                or seg.is_type("extension_reference")
+                or seg.is_type("column_reference")
+                or seg.is_type("sequence_reference")
+                or seg.is_type("trigger_reference")
+            ):
```
jpy-git, line 80:
```suggestion
            if seg.is_type("object_reference"):
```
all these segments inherit from `ObjectReferenceSegment`, so the other conditions are redundant

jpy-git (commented):
One thing that #2137 highlighted is that the changes to this rule are triggering warnings in some situations.

For example, in the unit test in `test/rules/std_L003_L036_L039_combo_test.py` if you try fixing the query you'll see these warnings:
image

As far as I can see there's nothing in this that should be touched by your addition but it is. Can you investigate please?

query here:
```
WITH example AS (
        SELECT my_id,
            other_thing,
            one_more
        FROM
            my_table
    )

    SELECT *
    FROM example
```

jpers36:
@jpy-git I looked into it just now, and it's due to #1304 as well.  The real fix would be to get that issue fixed.  The workaround would be to extend the workaround I added so that it ignores multiple leading whitespace segments and not just the first segment.  It shouldn't be hard, but the first solutions that come to mind are all inelegant.

src/sqlfluff/rules/L039.py
```diff
@@ -66,4 +66,22 @@ def _eval(self, context: RuleContext) -> Optional[List[LintResult]]:
                         )
                 prev_newline = False
                 prev_whitespace = None
+
+            if seg.is_type("object_reference"):
+                # This variable is a workaround to avoid removing new indents added at the beginning of a segment by L003.
+                # See Github issue #1304: https://github.com/sqlfluff/sqlfluff/issues/1304
+                # It represents the question: are we parsing through leading whitespace in this loop?
+                leading_whitespace = True
```
jpers36, line 74: @jpy-git adding this variable to track whether we're still in leading whitespace is my proposed approach.

jpy-git (approved):
@jpers36 Looks good and your suggestion to handle the leading whitespace seems pragmatic 👍 

Re: the incorrect parse tree issue causing those leading whitespaces in the object reference, we're well aware of the issue and it's a fairly complicated one. I'm doing some initial investigation atm. We can come back and tidy up the extra logic here in the future once that's fixed 😄 

Nice work, LGTM! 🚀

PR #2141, opened 2021-12-19
Remove support for Python 3.6 (it's "end of life" December 23, 2021)

barrywhart:

Removes support for Python 3.6.

### Pull Request checklist

  - Other.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -43,7 +43,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.6', '3.7', '3.8', '3.9', '3.10' ]
+        python-version: [ '3.7', '3.8', '3.9', '3.10' ]
     name: Python ${{ matrix.python-version }} tests
     steps:
     - uses: actions/checkout@v2
--- a/docs/source/gettingstarted.rst
+++ b/docs/source/gettingstarted.rst
@@ -28,7 +28,7 @@ which should give you a sensible read out and not an error.
 .. code-block:: bash
 
     $ python --version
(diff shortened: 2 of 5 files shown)


tunetheweb (approved):
I see another couple of references:

image

Or we happy to leave those in there and continue to support 3.6 unofficially?

Also FYI, in `.github/codecov.yml` we have this:

```
  notify:
    after_n_builds: 11
```

As of this PR:
- 4 linux standard runs are eligible 3.7 - 3.10
- 5 linux DBT runs (dbt018 - dbt100)
- 1 windows standard run runs
- 1 windows dbt run

Windows Linux is usually last to finish and think we now need it as a bit of coverage depends on that.

So think this is OK to leave. Or could reduce to 10 to get early feedback of code coverage but then you sometimes get false negatives of missing coverage.

barrywhart:
Last night, I removed those 2 other 3.6 references, and some tests were failing -- some weird pickling error in the progress bar tests. I'm guessing it was the stuff in `runner.py`. So I'd prefer to leave that for now if that's okay.

barrywhart:
I'm good with leaving it at 11. False positives are 😿

jpy-git:
> I'm good with leaving it at 11. False positives are 😿

Was wondering why we kept getting coverage errors before all the tests had finished since we added dbt100. Glad this will start working properly again 🎉

tunetheweb:
Yeah it’s annoying you have to remember to change this. But annoying. We used to have more headroom when Windows standard wasn’t required so didn’t matter it was the last to complete. I wonder if there’s a better way than just setting the number and telling codecov. Would be nice if we could say at least I’ve standard, dbt and windows test has to run.

PR #2144, opened 2021-12-20
Hive: add INTERVAL syntax

juhoautio:

The ANSI dialect of sqlfluff allows INTERVAL syntax like this:

```sql
SELECT INTERVAL 1 YEAR;
```

That is also supported since Hive 2.2.0, but in Hive 1.x that is not valid. The universal Hive syntax is like this:

```sql
SELECT INTERVAL '1' YEAR;
```

This PR adds that and the other [documented syntax alternatives of INTERVAL in Hive](https://cwiki.apache.org/confluence/display/hive/languagemanual+types#LanguageManualTypes-Intervals) to the dialect.

See the new test cases & Hive documentation for the full scope.

I'm not sure how to check, but maybe a quoted literal should be also accepted by ANSI dialect? What about the other "extensions" that Hive syntax has (are they extensions or actually just implementing ANSI syntax)?

### Pull Request checklist

  - [ ] **(not applicable)** `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - [x] `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `python test/generate_parse_fixture_yml.py` or by running `tox` locally).
  - [ ] **should something be added?** Full autofix test cases in `test/fixtures/linter/autofix`.
- [ ] **(not applicable)** Added appropriate documentation for the change.
- [ ] **please advise?** Created GitHub issues for any relevant followup/future enhancements if appropriate.


test/fixtures/dialects/hive/select_interval.sql
```diff
@@ -0,0 +1,39 @@
...
+       INTERVAL '2' HOUR +
+       INTERVAL '3' MINUTE +
+       INTERVAL '4' SECOND +
+       INTERVAL '5' NANO;
+
+SELECT INTERVAL 1 DAY;
+
+-- This interpretation may be too simplistic
+-- Original example was: INTERVAL (1+dt) DAY
+SELECT INTERVAL (1+2) DAY;
```
jpy-git, line 27: This isn't being parsed by the IntervalExpressionSegment, it's being interpreted as a function:
image

(same thing happens if you do `INTERVAL (1+dt) DAY`

jpy-git: To get both of those queries to parse we should have a bracketed expression in the grammar:
image

But the functions still seems to get matched first. @tunetheweb any idea on how we stop function from matching the `INTERVAL` keyword? (it's already a reserved keyword)

jpy-git (commented):
Looks good, need to move the datetime units and figure out how to get the expression examples to not parse as functions

src/sqlfluff/dialects/dialect_hive.py
```diff
@@ -554,3 +554,36 @@ class InsertStatementSegment(BaseSegment):
             ),
         ),
     )
+
+
+@hive_dialect.segment(replace=True)
+class IntervalExpressionSegment(BaseSegment):
+    """An interval expression segment.
+
+    Full Apache Hive `INSERT` reference here:
```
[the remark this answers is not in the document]
juhoautio: > This comment line looks wrong, should say something about intervals not insert

Sorry about that! Fixed.

src/sqlfluff/dialects/dialect_hive.py
```diff
@@ -554,3 +554,36 @@ class InsertStatementSegment(BaseSegment):
...
+    # "Add timeunit aliases to aid portability / readability"
+    date_time_aliases = [
+        "SECONDS",
+        "MINUTES",
+        "HOURS",
+        "DAYS",
+        "WEEKS",
+        "MONTHS",
+        "YEARS",
+    ]
```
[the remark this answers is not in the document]
juhoautio: Done. Awesome that you also pointed the suitable line for that, I was about to start wondering what would be the right position.

test/fixtures/dialects/hive/select_interval.sql
juhoautio: I moved IntervalExpressionSegment to be in order before FunctionSegment in ansi_dialect. That seems to make parser try to match on INTERVAL before function and it seems like it may have done the trick. Please let me know..

jpy-git: Ahh nice one, that's really good thinking! Looks good to me now 💪

jpy-git: Would you be able to add in a unit test for the `INTERVAL (1+dt) DAY` example as well to see how that parses?

jpy-git:
@juhoautio you'll need to run black to get the linting tests to pass. I'd recommend using the pre-commit hook to automatically run it at commit time.

install with: `tox -e pre-commit -- install`
and then it will run black/mypy/flake8 at commit time for you 😄 

Once that's done we should be good to approve and merge 👍

test/fixtures/dialects/hive/select_interval.sql
juhoautio: That works too and I was switching to just have that. But I got some git trouble when trying to push because of https://github.com/sqlfluff/sqlfluff/pull/2144/commits/93e7b3c4c32932ef5468743800a4e18e7e0dea3d, oops :) But coming soon. There's also a fix for the NANO parsing.

juhoautio:
> you'll need to run black to get the linting tests to pass

Yeah sorry about that, I also noticed the failed CI check, and will also push that fix.

> I'd recommend using the pre-commit hook to automatically run it at commit time.

I was about to ask about the pre-commit hooks: I enabled them with `tox -e pre-commit -- install` but the mypy hook always fails for me:

```
$ mypy --version
mypy 0.920

$ mypy src/sqlfluff --ignore-missing-imports
src/sqlfluff/__init__.py:12: error: Module "importlib" has no attribute "metadata"
src/sqlfluff/__init__.py:12: error: Name "metadata" already defined (by an import)
Found 2 errors in 1 file (checked 154 source files)
```

I presume it works for you? I wonder what I might be doing wrong.

While at it, the contribution doc mentioned this as "kick that last one off just before opening a PR":

```
tox -e cov-init,dbt018-py38,cov-report-dbt -- plugins/sqlfluff-templater-dbt
```

Any idea if that contains the same stuff the the pre-commit hooks? It doesn't matter much though – I didn't even run that this time (takes long). In any case I think pre-commit hook can be more strongly advised to be enabled (I can PR that after), but first I need to somehow fix the mypy error.

test/fixtures/dialects/hive/select_interval.yml
```diff
@@ -121,10 +121,10 @@ file:
               literal: "'4'"
               date_part: SECOND
           - binary_operator: +
-          - keyword: INTERVAL
-          - literal: "'5'"
-          alias_expression:
-            identifier: NANO
```
juhoautio: This was wrong. Now this is also recognized as date_part.

jpy-git:
@juhoautio ahh I bet you're using python3.7 right? I think I saw this mypy issue before but only when I was using 3.7 and didn't get a chance to add the type ignore comment.

Could you add a `  # type: ignore` comment to the end of line 12 in `src/sqlfluff/__init__.py` and let me know if that fixes it?
(basically we have to use the backport of metadata for python3.7 and it confuses mypy a bit so it's a false positive)

jpy-git:
> tox -e cov-init,dbt018-py38,cov-report-dbt -- plugins/sqlfluff-templater-dbt

so this one is just for running the dbt tests but we don't touch that in this PR so no need 👍 

basically I just use the pre-commit hook for formatting and linting and `tox -e py39` (or whatever your python version) for running the unit tests (the rest I just leave for the CI checks).

test/fixtures/dialects/hive/select_interval.yml
jpy-git: Nice spot, yeah it's good to check things parse as expected to avoid later unexpected issues in rules/formatting 👍

jpy-git (approved):
@juhoautio, Great work LGTM 🚀

PR #2147, opened 2021-12-20
Fix mypy error on python 3.7

juhoautio:

This fixes mypy (which is also part of this project's pre-commit hooks) when developing locally with python 3.7.

Thanks @jpy-git for the help in https://github.com/sqlfluff/sqlfluff/pull/2144#issuecomment-998304477.

The error before this change:

```
$ mypy src/sqlfluff --ignore-missing-imports
src/sqlfluff/__init__.py:12: error: Module "importlib" has no attribute "metadata"
src/sqlfluff/__init__.py:12: error: Name "metadata" already defined (by an import)
Found 2 errors in 1 file (checked 154 source files)
```

With this fix:
```
$ mypy src/sqlfluff --ignore-missing-imports
Success: no issues found in 154 source files
```

Now I was able to run the pre-commits:
```
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
black....................................................................Passed
mypy.....................................................................Passed
flake8...................................................................Passed
doc8.................................................(no files to check)Skipped
```

No.

### Pull Request checklist

N/A, but let's just see the CI checks ;)

### DIFF
--- a/src/sqlfluff/__init__.py
+++ b/src/sqlfluff/__init__.py
@@ -1,4 +1,3 @@
-# mypy: warn_unused_ignores=False
 """Sqlfluff is a SQL linter for humans."""
 import sys
 import pytest
@@ -7,10 +6,10 @@ import pytest
 from sqlfluff.api import lint, fix, parse, list_rules, list_dialects
 
 # Import metadata (using importlib_metadata backport for python versions <3.8)
-if sys.version_info < (3, 8, 0):
-    import importlib_metadata as metadata
+if sys.version_info >= (3, 8):
+    from importlib import metadata
 else:
-    from importlib import metadata  # type: ignore
+    import importlib_metadata as metadata
 
 __all__ = (
     "lint",
--- a/src/sqlfluff/__init__.py
+++ b/src/sqlfluff/__init__.py
@@ -1,3 +1,4 @@
+# mypy: warn_unused_ignores=False
 """Sqlfluff is a SQL linter for humans."""
 import sys
 import pytest
--- a/src/sqlfluff/__init__.py
+++ b/src/sqlfluff/__init__.py
@@ -9,7 +9,7 @@ from sqlfluff.api import lint, fix, parse, list_rules, list_dialects
 if sys.version_info < (3, 8, 0):
     import importlib_metadata as metadata
 else:
-    from importlib import metadata
+    from importlib import metadata  # type: ignore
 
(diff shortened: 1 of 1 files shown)


jpy-git (dismissed):
Nice, LGTM! 🚀 ⭐

jpy-git:
@juhoautio remembered why I didn't fix this before:
image
the line doesn't get triggered in python >=3.8 so instead raises a `type: ignore` unused error 😢 

If you've run black in your other PR mypy will parse so won't block you and I'll have a think about this

juhoautio:
@jpy-git looks like mypy nowadays supports [per-file configuration options inside files](https://mypy.readthedocs.io/en/stable/inline_config.html), so I set `# mypy: warn_unused_ignores=False` on the file (pushed the commit). I suppose it can be ok to ignore that check on the rare cases ie. files that have version specific code.

jpy-git:
@juhoautio according to mypy it shouldn't struggle with this: https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks

which is bascially what we have anyway, out of curiosity could we try rewriting the impacted lines like:
```python
if sys.version_info >= (3, 8):
    from importlib import metadata
else:
    import importlib_metadata as metadata
```
to match those mypy docs and see if it satisfies it for python 3.7 without the ignore?

jpy-git:
> @jpy-git looks like mypy nowadays supports [per-file configuration options inside files](https://mypy.readthedocs.io/en/stable/inline_config.html), so I set `# mypy: warn_unused_ignores=False` on the file (pushed the commit). I suppose it can be ok to ignore that check on the rare cases ie. files that have version specific code.

Yeah I'm fine with this 👍 Could you double check if my above suggestion makes any difference for you, if not we can go with this approach.

juhoautio:
Wow, indeed your suggestion in https://github.com/sqlfluff/sqlfluff/pull/2147#issuecomment-998329796 works! I also saw that somewhere and I was sure I tried it without success previously but must've made some mistake there.

jpy-git:
lol that's so weird, it's basically the same as my original code but swapped round 🤷 I'll make a mypy issue for it as it seems like if it can handle one it should handle the other. But glad this works now 🎉

jpy-git (approved):
Thanks @juhoautio! Second times the charm, LGTM 🚀 ⭐

PR #2155, opened 2021-12-21
TSQL: function parsing/linting

jpers36:
SQLFluff currently does not parse the internals of a function.  This means that SQLFluff will not fully lint a function.  This PR will add a proper grammar for functions.

Dialect Changes:
-Added a ReturnStatementSegment to represent the return segment
-Added FunctionOptionSegment to represent a function's optional WITH clause
-Added ExecuteAsClauseSegment to represent an EXECUTE AS clause
-Removed FunctionDefinitionGrammar, since it's redundant with ProcedureDefinitionGrammar
-Added ReturnStatementSegment reference to StatementSegment
-Updated CreateFunctionStatement: moved parse_grammar logic to match_grammar, added FunctionOptionSegment, replaced FunctionDefinitionGrammar with ProcedureDefinitionGrammar
-Updated ProcedureDefinitionGrammar description to note that it's used for functions as well

Keyword Changes:
-Added INLINE and CALLER unreserved keywords

Test changes:
-Added create_function.sql test case and YML
-Fixed parsing for three other function test cases
-Added L003 test case with TSQL function


tunetheweb (approved):
LGTM

Issue #2134, opened 2021-12-18
Result of `fix` changes on subsequent runs (fix is non-deterministic)


juhoautio:
## Expected Behaviour

Running `sqlfluff fix my_query.sql -f` modifies query format – but running it again on the same file should not make any further changes.

## Observed Behaviour

Summary: The result changes considerably also on the 2nd run. On further runs the result keeps changing with a minor whitespace change.

The behavior is demonstrated in this branch: https://github.com/juhoautio/sqlfluff/commits/non_deterministic_fix_bug

<details>

   - 1st run: https://github.com/juhoautio/sqlfluff/commit/bc60c5767245be92b1fc7ff1f5666253e7ea7af5
   - 2nd run: https://github.com/juhoautio/sqlfluff/commit/148c91d81808a845a4e068d864e5a418906d4acd
   - 3rd run: https://github.com/juhoautio/sqlfluff/commit/befff642abeb6a5f4de05ccbce9ea38203e01a9f
   - 4th run: https://github.com/juhoautio/sqlfluff/commit/d9bcbf6c984fa0c7138b99009f5898a38681cbdc
</details>

Console output for each command is included below.

### 1st run of `sqlfluff fix`

<details>

```
$ python -m sqlfluff fix my_query.sql -f
==== finding fixable violations ====
== [my_query.sql] FAIL
L:   3 | P:  20 | L010 | Keywords must be consistently upper case.
L:   4 | P:  20 | L010 | Keywords must be consistently upper case.
L:   5 | P:  23 | L010 | Keywords must be consistently upper case.
L:   7 | P:  16 | L008 | Commas should be followed by a single whitespace unless
                       | followed by a comment.
L:   9 | P:   2 | L003 | Indentation not consistent with line #8
L:   9 | P:   2 | L022 | Blank line expected but not found after CTE closing
                       | bracket.
L:  10 | P:   6 | L003 | Indentation not hanging or a multiple of 4 spaces
L:  11 | P:  10 | L003 | Indentation not hanging or a multiple of 4 spaces
L:  12 | P:  10 | L003 | Indentation not hanging or a multiple of 4 spaces
L:  13 | P:  10 | L003 | Indentation not hanging or a multiple of 4 spaces
L:  13 | P:  10 | L030 | Function names must be consistently lower case.
L:  14 | P:   6 | L003 | Indentation not hanging or a multiple of 4 spaces
[...]

PR #2161, opened 2021-12-21
Set _works_on_unparsable to False by default

jpy-git:

Partially addresses unparseable related issues in #2134. Previously we tried to apply fixes to unparseable segments unless specifically specified otherwise. However, you cannot expect reasonable application of rules to a parse tree that contains an unparseable segment as it won't obey logical structure. 

Fixes will be a bit more picky as to what they apply to.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -265,7 +265,7 @@ class BaseRule:
     """
 
     _check_docstring = True
-    _works_on_unparsable = True
+    _works_on_unparsable = False
     targets_templated = False
 
     def __init__(self, code, description, **kwargs):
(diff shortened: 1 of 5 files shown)


tunetheweb:
I think we should review the rules that this CAN be True for, and turn them on - rather than turn off for all:
- L001 (unnecessary trailing space) files like it doesn't need the line to be parsable to work
- L002 (mixed tables and whitespaces) feels similar
- L004 (incorrect indentation type) if it's a straight replacement
- L005 (no whitespace before commas)
- L008 (single whitespace after comma)
- L009 (files end with one newline)
- L016 (line is too long)
- L039 (unnecessary whitespace)
- L046 (Jinja tags should have a single white space either side of them) - I presume this is acted on the raw unparsed file anyway
- L050 (files should not begin with newlines or whitespaces)

The amount of lower-numbered ones maybe explains why the default was True! though agree the default should be false, should it explicitly be set to True for the above rules?

jpy-git:
@tunetheweb I've added a pretty strong suggestion to the linked issue so would recommend reading that.

PR #2158, opened 2021-12-21
Linter Test Name Duplication

jpers36:
Two of the L003 linter tests have duplicate names.  I've adjusted one to properly match the test case.

Two of the L014 linter tests have duplicate names.  I've adjusted one to properly match the test case.

Two of the L043 linter tests are exactly the same.  I've removed one.

PR #2162, opened 2021-12-21
Add yamllint to project

tunetheweb:

This PR adds yamllint support to the project including:
- CI checks
- Precommit checks
- Cleanup of existing yamllint failures

This helps us avoid issues like #2158 where we had duplicate test names (in which case only the later test is actually run by pytest!).

At the moment I have turned off some rules from the [standard yamllint config](https://yamllint.readthedocs.io/en/stable/configuration.html#default-configuration):
- Turned off `brackets` as was complaining about spaces that I kinda liked
- Turned off `document-start` as not all our YAML files begin with `---` (maybe they should?)
- Turned off some of the `indentation` config as couldn't figure out how to make it work nicely with our auto-generated YAML files
- Turned off `line-length` as some of our lines are quite long (maybe we should shorten?)
- Turned off one `truthy` checks for keys so it's only applied to values, and added some overrides for those external config files that don't use `true` and `false`.

I also standardised our YAML Boolean values to `true` and `false` (previously was mostly `True` and `False` but with a few instances of lowercase). This is a bit inconsistent with Python (which demands Title case), but seems to be more standard to be lowercase in YAML and lots of other YAML files in our repo use lowercase.

Open to discussion if we want to turn more of these on?

### Pull Request checklist

  - Other.


jpy-git:
whoops marked the wrong pr as draft 😅

test/fixtures/dialects/ansi/insert_b.yml
```diff
@@ -1,19 +0,0 @@
-# Insert Statements
```
jpy-git, line 1: why did this file get deleted?

.yamllint
```diff
@@ -0,0 +1,29 @@
...
+  hyphens: enable
+  key-duplicates: enable
+  key-ordering: disable
+  line-length: disable
+  new-line-at-end-of-file: enable
+  new-lines: enable
+  octal-values: disable
+  quoted-strings: disable
+  trailing-spaces: enable
+  truthy: disable
```
jpy-git, line 29:
```suggestion
  truthy:
    allowed-values: ['True', 'False']
```
We shouldn't disable this, If we want capitalised True/False as we currently use then set that

.yamllint
```diff
@@ -0,0 +1,29 @@
+---
+
+yaml-files:
+  - '*.yaml'
+  - '*.yml'
+  - '.yamllint'
+
+rules:
+  braces: enable
```
jpy-git, line 9: a lot of these options are in the default config: https://yamllint.readthedocs.io/en/stable/configuration.html#default-configuration

We should only include non-default rules, otherwise its a hassle to maintain as it looks like you've put them in here for a reason (also makes it hard to review).

jpy-git (commented):
@tunetheweb couple comments

test/fixtures/dialects/ansi/insert_b.yml
tunetheweb: It’s not used. There is no equivalent sql file and tests pass without it. I figured it was an old file that was no longer needed. It also fails yamllint at the mo but didn’t investigate how to fix as not needed.

.yamllint
tunetheweb: Fair point. Wasn’t sure if a config file was in addition to, or a complete replacement (so defaults no longer used at all). Will test some more.

tunetheweb: Ah see in that it extends. Cool. Will remove the defaults.

.yamllint
KulykDmytro: let me check on my PR whether this works in form of `true` / `false`

KulykDmytro: works in "native yaml" bool value
https://github.com/sqlfluff/sqlfluff/pull/2154/checks

tunetheweb: Yup just changed in https://github.com/sqlfluff/sqlfluff/pull/2162/commits/2cafb66cbb822479e1ae686995e8a498ba24c46e. As discussed on slack:
> Just change them all to lowercase in that PR after fighting with too many exceptions for other (non-SQLFluff) files to allow us to keep True with initial uppercase
> It’s a bit inconsistent since Python insists on True and now standardising on true for YAML but seems the simplest

.yamllint
tunetheweb: OK I had to add `extends: default` and now it extends :-)

jpy-git:
@tunetheweb So I've cloned this PR and run the `tox -e yamllint` and got a bunch of errors:
image

These are generated by tox and the dbt unit tests. Think we need an equivalent of .gitignore to ignore these files otherwise it'll cause issues for others trying to use pre-commit.

jpy-git:
The folders to ignore are `.tox` and `plugins/sqlfluff-templater-dbt/test/fixtures/dbt/dbt_project/dbt_packages/dbt_utils`

tunetheweb:
I added these:

```
ignore: |
  .tox/
  .venv/
  dbt_modules/
  dbt_packages/
```

Could you do a `git pull` and try again? I don't have dbt installed so didn't see this.

jpy-git:
it's working for me now thanks!

jpy-git (approved):
@tunetheweb Looks good, LGTM 🚀 ⭐

KulykDmytro:
🚀

PR #2167, opened 2021-12-21
L010 to apply to date_part (capitalization policy for time units)

juhoautio:

Before this change the new tests failed as follows:

`L010_test_fail_date_part_inconsistent_capitalisation`:
```sql
Expected :'SELECT dt + INTERVAL 2 DAY, INTERVAL 3 HOUR'
Actual   :'SELECT dt + INTERVAL 2 day, INTERVAL 3 HOUR'
```

`L010_test_fail_date_part_capitalisation_policy_lower`:
```sql
Expected :'select dt + interval 2 day, interval 3 hour'
Actual   :'select dt + interval 2 day, interval 3 HOUR'
```

`L010_test_fail_date_part_capitalisation_policy_upper`:
```sql
Expected :'SELECT dt + INTERVAL 2 DAY, INTERVAL 3 HOUR'
Actual   :'SELECT dt + INTERVAL 2 day, INTERVAL 3 HOUR'
```

So the keyword capitalization rule was not applied on the time units. The original case was kept (it wasn't touched).

I thus suggest that `Rule_L010: Inconsistent capitalisation of keywords` must be applied to `date_part`, because the time units are also keywords. But please advise if the definition of keyword should be thought as something more narrow for Rule_L010.

Another possibility could be to add a new rule, something similar as `Rule_L040: Inconsistent capitalisation of boolean/null literal`. But I think it would be a more common use case to use a different capitalization rule on the booleans. The desired treatment of time units would supposedly be to follow the setting of other keywords.

This change may touch quite many existing files of users if there are time units because capitalization rules haven't reached those before.

### Pull Request checklist

  - Other.

### DIFF
--- a/test/fixtures/linter/autofix/hive/001_date_part_capitalisation/.sqlfluff
+++ b/test/fixtures/linter/autofix/hive/001_date_part_capitalisation/.sqlfluff
@@ -1,0 +1,2 @@
+[sqlfluff:rules:L010]
+capitalisation_policy = upper
--- a/test/fixtures/linter/autofix/hive/001_date_part_capitalisation/.sqlfluff
+++ b/test/fixtures/linter/autofix/hive/001_date_part_capitalisation/.sqlfluff
@@ -1,2 +1,0 @@
-[sqlfluff:rules:L010]
-capitalisation_policy = upper
--- a/test/fixtures/rules/std_rule_cases/L010.yml
+++ b/test/fixtures/rules/std_rule_cases/L010.yml
@@ -35,3 +35,26 @@ test_fail_capitalisation_policy_capitalise:
     rules:
       L010:
         capitalisation_policy: capitalise
+
+test_fail_date_part_inconsistent_capitalisation:
+  # Test that time unit capitalization is fixed
+  fail_str: SELECT dt + interval 2 day, interval 3 HOUR
+  fix_str: SELECT dt + INTERVAL 2 DAY, INTERVAL 3 HOUR
+
+test_fail_date_part_capitalisation_policy_lower:
+  # Test that capitalization policy is applied on time units
+  fail_str: SELECT dt + interval 2 day, interval 3 HOUR
+  fix_str: select dt + interval 2 day, interval 3 hour
+  configs:
+    rules:
+      L010:
+        capitalisation_policy: lower
+
+test_fail_date_part_capitalisation_policy_upper:
+  # Test that capitalization policy is applied on time units
+  fail_str: SELECT dt + interval 2 day, interval 3 HOUR
+  fix_str: SELECT dt + INTERVAL 2 DAY, INTERVAL 3 HOUR
+  configs:
+    rules:
+      L010:
+        capitalisation_policy: upper
--- a/test/fixtures/rules/std_rule_cases/L010.yml
+++ b/test/fixtures/rules/std_rule_cases/L010.yml
@@ -58,3 +58,7 @@ test_fail_date_part_capitalisation_policy_upper:
     rules:
       L010:
         capitalisation_policy: upper
+
+test_pass_date_part_consistent_capitalisation:
+  # Test that correctly capitalized time units are left unchanged
+  pass_str: SELECT dt + INTERVAL 2 DAY, INTERVAL 3 HOUR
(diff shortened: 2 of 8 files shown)


test/rules/std_fix_auto_test.py
```diff
@@ -1,6 +1,6 @@
 """Automated tests for fixing violations.
 
-Any files in the /tests/fixtures/linter/autofix directoy will be picked up
+Any files in the test/fixtures/linter/autofix directory will be picked up
```
juhoautio, line 3: A separate PR about this little fix, or maybe not?

jpy-git (commented):
@juhoautio makes sense, just need to move those unit tests and add a couple more test variations and then it should be good to go 👍

test/fixtures/linter/autofix/hive/001_date_part_capitalisation/.sqlfluff
```diff
@@ -0,0 +1,2 @@
+[sqlfluff:rules:L010]
+capitalisation_policy = upper
```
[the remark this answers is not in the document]
juhoautio: > put your test cases in this file instead please: https://github.com/sqlfluff/sqlfluff/blob/main/test/fixtures/rules/std_rule_cases/L010.yml

Thanks! That is much nicer!

> Autofix is reserved for more complex interactions between multiple rules

I wonder if there could be some high level doc about how all the tests are organized. For dialects there's https://github.com/sqlfluff/sqlfluff/wiki/Contributing-Dialect-Changes#adding-test-cases-for-your-changes in wiki but I suppose I'm missing the big picture. Not that I necessarily need to know.. but you seem to have it all figured out so just a suggestion if it makes sense to write something down :)

Maybe something could be added in doctoring of yaml_test_cases_test.py, too. This seems kinda wrong: `"""Runs the rule test cases."""` – for sure there are also other kinds of tests related to rules.

test/rules/std_fix_auto_test.py
juhoautio: Well, I suppose this PR will be squashed so the commit message / git blame would be misleading for those. There were actually 2 cases like that so I created a separate PR after all https://github.com/sqlfluff/sqlfluff/pull/2179

juhoautio:
@jpy-git thanks again for the super quick & helpful review.

Does this need to be mentioned in L010 doc? IMHO maybe not and this is rather a bug fix than a change in behavior.

I have updated the PR description as well.

test/fixtures/rules/std_rule_cases/L010.yml
```diff
@@ -35,3 +35,26 @@ test_fail_capitalisation_policy_capitalise:
     rules:
       L010:
         capitalisation_policy: capitalise
+
+test_fail_date_part_inconsistent_capitalisation:
+  # Test that time unit capitalization is fixed
+  fail_str: SELECT dt + interval 2 day, interval 3 HOUR
+  fix_str: SELECT dt + INTERVAL 2 DAY, INTERVAL 3 HOUR
```
jpy-git, line 42: Could you add a pass unit test as well please?
```
test_pass_date_part_consistent_capitalisation:
  # Test that correctly capitalized time units are left unchanged.
  pass_str: SELECT dt + INTERVAL 2 DAY, INTERVAL 3 HOUR
```
(I'm sure it works but always good to check we don't raise false positives in the tests)

jpy-git (commented):
@juhoautio just one more unit test to add. I agree with you that the date parts should just be considered as keywords in L010, so no need for additional docs 👍

test/fixtures/linter/autofix/hive/001_date_part_capitalisation/.sqlfluff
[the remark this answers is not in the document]
jpy-git: Yeah I wasn't fully clear on where to put test cases when I first started so happy to discuss if you make a separate issue for it?

test/fixtures/rules/std_rule_cases/L010.yml
juhoautio: Added (with small change: removed `.` from end of comment to match file's current file).

Btw there were no other `pass_str` tests in `L010.yml` so this is the first one. I suppose you noticed.

jpy-git: Yeah I saw, some of these tests are fairly old. The more recent rules have more in-depth test cases so best to be thorough with anything new we add. No harm in building out the test cases of old rules in a separate PR if you wanted though 👍

jpy-git (approved):
@juhoautio Nice work, LGTM! 🚀 ⭐

PR #2170, opened 2021-12-22
TSQL: Remove dependency on ANSI keyword lists

jpers36:
Up to this point, the TSQL dialect has incorporated the ANSI dialect's keyword list into its own keyword list.  The upside of this has been the ability to rely on ANSI's definitions even when those definitions include references to keywords not existent in TSQL.  The downside is that some non-keywords may flag as keywords when parsing TSQL.

This PR aims to rip off the bandage.  This is driven by an issue where I have a stored procedure that can't parse because the parser incorrectly flags ROWS as a reserved keyword.  So this PR clears the dialect keyword list before applying TSQL's list, and incorporates all changes needed to make that work.

TSQL keyword changes:
- [x] Added a number of unreserved keywords that are actually used by TSQL but were previously not in the list

TSQL dialect changes:
- [x] Clear ANSI keywords from list
- [x] Add BaseExpressionElementGrammar to override ANSI and remove IntervalExpressionSegment reference
- [x] Add IsClauseGrammar to override ANSI and remove NAN reference
- [x] Add LikeGrammar to override ANSI and remove RLIKE/ILIKE
- [x] Add FunctionContentsGrammar to override ANSI and remove SEPARATOR
- [x]  Remove IntervalExpressionSegment reference from Expression_D_Grammar
- [x] remove CreateExtensionStatementSegment, CreateModelStatementSegment, DropModelStatementSegment, and DescribeStatementSegment from StatementSegment
- [x] Add CreateSequenceOptionsSegment to properly represent TSQL functionality
- [x] Adjust ColumnConstraintSegment to remove invalid AUTO_INCREMENT and UNSIGNED options, and remove reference to invalid CommentClauseSegment
- [x] Remove CreateExtensionStatementSegment, CreateModelStatementSement, DropModelStatementSegment, OverlapsClauseSegment Nothing() stubs, replacing them with removals elsewhere
- [x] Add MLTableExpressionSegment Nothing() stub to remove ML keyword; add TO DO to possibly represent PREDICT statment here
- [x] Rename PartitionByClause to PartitionClauseSegment to match ANSI dialect naming convention; fix type name to match as well
- [x] Add AlterTableSegment to override ANSI and remove MODIFY
- [x] Add SetOperatorSegment to override ANSI and remove MINUS
- [x] Add WindowSpecificationSegment to override ANSI and remove the unsupported functionality to name a window; this needs to be done because PARTITION's status change from a reserved keyword to an unreserved keyword otherwise confuses this definition

Test cases:
- [x] Added TSQL non-keywords to select test case
- [x] Added more detailed case to sequence test case
- [x] Added ALTER table test cases
- [x] Minor adjustments due to the PartitionByClause/PartitionClauseSegment change


src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -1577,6 +1662,40 @@ class OverlapsClauseSegment(BaseSegment):
...
 
+@tsql_dialect.segment(replace=True)
+class DescribeStatementSegment(BaseSegment):
+    """A `Describe` statement.
+
+    Not present in T-SQL.
+    """
+
+    type = "describe_statement"
+    match_grammar = Nothing()
```
jpy-git, line 1673: Rather than have a stub I think you can just remove this from the StatementSegment like this:
image

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -41,8 +41,8 @@
 tsql_dialect = ansi_dialect.copy_as("tsql")
 
 # Should really clear down the old keywords but some are needed by certain segments
```
jpy-git, line 43: Do we need this comment now?

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -1577,6 +1662,40 @@ class OverlapsClauseSegment(BaseSegment):
...
+
+@tsql_dialect.segment(replace=True)
+class CommentClauseSegment(BaseSegment):
+    """A comment clause.
+
+    Not present in T-SQL.
+    """
+
+    type = "comment_clause"
+    match_grammar = Nothing()
```
jpy-git, line 1684: - Remove reference to this on line 1966 in the CREATE TABLE segment.

- Remove reference to this on line 1293 in the CREATE TABLE segment.

These are the only segments in ANSI where this is referenced so and they are both overwritten in TSQL so if we remove the references listed above then no need to add this stub.

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -1577,6 +1662,40 @@ class OverlapsClauseSegment(BaseSegment):
...
+@tsql_dialect.segment(replace=True)
+class MLTableExpressionSegment(BaseSegment):
+    """An ML table expression.
+
+    Not present in T-SQL.
+    TO DO: Consider whether this segment can be used to represent a PREDICT statement.
+    """
+
+    type = "ml_table_expression"
+    match_grammar = Nothing()
```
jpy-git, line 1696: This ones only referenced in `FromExpressionSegment` in ANSI:
image

It's fairly simple so I guess maybe we could just copy this to TSQL but without MLTableExpressionSegment?

Happy to leave as a stub for now whilst you investigate the PREDICT statement implementation 👍

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -1867,6 +1986,55 @@ class CreateTableStatementSegment(BaseSegment):
     parse_grammar = match_grammar
 
 
+@tsql_dialect.segment(replace=True)
+class AlterTableStatementSegment(BaseSegment):
+    """An `ALTER TABLE` statement.
+
+    https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-ver15
+    Overriding ANSI to remove TSQL non-keywords MODIFY, FIRST
+    TODO: Flesh out TSQL-specific functionality
```
jpy-git, line 1995: Happy to leave like this for now, looks like it'll require pretty beefy syntax to complete this grammar 👍

jpy-git (commented):
@jpers36 looks great, couple points to clear up 👍

src/sqlfluff/dialects/dialect_tsql.py
jpers36: @jpy-git thanks for the syntax help!  I moved a couple more stubs to this design now that I know how.

src/sqlfluff/dialects/dialect_tsql.py
jpers36: @jpy-git removed, thanks!

src/sqlfluff/dialects/dialect_tsql.py
jpers36: @jpy-git removed, thanks!

src/sqlfluff/dialects/dialect_tsql.py
jpers36: @jpy-git I've added issue #2199 to cover the PREDICT implementation.

src/sqlfluff/dialects/dialect_tsql.py
jpers36: @jpy-git I submitted #2201 to cover the grammar completion

jpy-git: fab thanks 😄

src/sqlfluff/dialects/dialect_tsql.py
jpy-git: thanks! 👍

jpy-git (approved):
@jpers36 really great stuff, thanks for driving the T-SQL syntax coverage forwards 💪 LTGM 🚀 ⭐

PR #2171, opened 2021-12-23
L027 Fix self referring column alias edge case

### LINKED ISSUE #2169: [L027] Not working properly when identifier aliases with same name
It works as expected for the following minimal SQL query
```mysql
SELECT DISTINCT
    foo,
    bar
FROM
    a LEFT JOIN b ON a.id = b.id
```

It works as expected for the following minimal SQL query
```mysql
SELECT DISTINCT
    foo AS f,
    bar AS b
FROM
    a LEFT JOIN b ON a.id = b.id
```

It doesn't work as expected for the following minimal SQL query when the alias name is same as the identifier
```mysql
SELECT DISTINCT
    foo AS foo,
    bar AS bar
FROM
    a LEFT JOIN b ON a.id = b.id
```

## Expected Behaviour
```
== [bug.sql] FAIL
L:   2 | P:   5 | L027 | Unqualified reference 'foo' found in select with more
                       | than one referenced table/view.
L:   3 | P:   5 | L027 | Unqualified reference 'bar' found in select with more
                       | than one referenced table/view.
All Finished 📜 🎉!
```

## Observed Behaviour
```
All Finished 📜 🎉!
```

## Steps to Reproduce
Copy the samples to scripts and trigger `sqlfluff lint buggy.sql`

## Dialect
Didn't provide anything explicitly but am assuming it's mysql as I've been running them on `*.sql` files.

## Version
```
sqlfluff, version 0.9.0
Python 3.9.7
```

## Configuration
No configuration used

jpy-git:

Fixes #2169. Previously `get_select_statement_info` only returned a list of column alias names without information about the select clause element that they correspond to. I've added a `ColumnAliasInfo` named tuple (like `AliasInfo` for table aliases) to capture this information and supply it to rule L027.

Now we can check that the condition `r.raw not in col_alias_names` in L027, that is intended to allow unqualified column names if it is actually an alias for another column, does not compare a column reference to its own alias.

No

### Pull Request checklist

  - Other.


test/fixtures/rules/std_rule_cases/L027.yml
```diff
@@ -73,3 +73,30 @@ test_ignore_value_table_functions:
...
+    FROM
+        a LEFT JOIN b ON a.id = b.id
+
+test_pass_column_and_alias_same_name_1:
+  pass_str: |
+    SELECT
+        a.foo AS foo,
+        b.bar AS bar
+    FROM
+        a LEFT JOIN b ON a.id = b.id
```
tunetheweb, line 92: Should we add similar test cases for t-sql and mysql since you had to explicitly add extra code for them?

Any ideas how we can ensure no one forgets to add a `get_alias` for any future dialect that implements it's own `SelectClauseElementSegment`?

jpy-git: @tunetheweb I've added the extra unit tests.

For the `get_alias` in future dialects point, I think that's something we'd pick up in code review, since I'd expect the reviewer to look at the segment being overwritten? Don't think we have anything special in place for the existing similar `get_eventual_aliases`/`AliasInfo` cases.

tunetheweb: Yeah couldn’t think of any systematic way of checking but thought worth asking in case wiser minds could.

tunetheweb (approved):

PR #2189, opened 2021-12-23
Lexer: add non-breaking spaces to whitespace

jpers36:
The lexer can't lex non-breaking spaces.  This PR will add non-breaking spaces to the whitespace lexer.

### DIFF
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -58,7 +58,7 @@ ansi_dialect = Dialect("ansi", root_segment_name="FileSegment")
 
 ansi_dialect.set_lexer_matchers(
     [
-        RegexLexer("whitespace", r"[\t ]+", WhitespaceSegment),
+        RegexLexer("whitespace", r"[\t\xc2\xa0 ]+", WhitespaceSegment),
         RegexLexer(
             "inline_comment",
             r"(--|#)[^\n]*",
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -58,7 +58,7 @@ ansi_dialect = Dialect("ansi", root_segment_name="FileSegment")
 
 ansi_dialect.set_lexer_matchers(
     [
-        RegexLexer("whitespace", r"[\t\xc2\xa0 ]+", WhitespaceSegment),
+        RegexLexer("whitespace", r"[^\S\r\n]+", WhitespaceSegment),
         RegexLexer(
             "inline_comment",
             r"(--|#)[^\n]*",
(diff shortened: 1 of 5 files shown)


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -58,7 +58,7 @@
 
 ansi_dialect.set_lexer_matchers(
     [
-        RegexLexer("whitespace", r"[\t ]+", WhitespaceSegment),
+        RegexLexer("whitespace", r"[\t\xc2\xa0 ]+", WhitespaceSegment),
```
jpy-git, line 61: @jpers36 I don't doubt this works but feels a bit inelegant. Doing a bit of research I saw a comment that the regex `\s` also matches non-breaking spaces, however it also matches newlines.

A good post [here](https://stackoverflow.com/questions/3469080/match-whitespace-but-not-newlines) suggests we can use this double negative regex pattern `[^\S\r\n]` to do `\s` but not newlines.
Could you try that please and see if that also solves the issue?

jpers36: @jpy-git that works, thanks!

jpy-git: Nice!

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -58,7 +58,7 @@
 
 ansi_dialect.set_lexer_matchers(
     [
-        RegexLexer("whitespace", r"[\t ]+", WhitespaceSegment),
+        RegexLexer("whitespace", r"[^\S\r\n]+", WhitespaceSegment),
```
jpy-git, line 61: Could we also add a comment here to the effect of:
```
# Match all forms of whitespace except newlines and carriage returns:
# https://stackoverflow.com/questions/3469080/match-whitespace-but-not-newlines
# This pattern allows us to also match non-breaking spaces (#2189).
```
so people don't question our choice. 😄  

Also the whitespace pattern gets used in a couple different places so we should update these too:
https://github.com/sqlfluff/sqlfluff/blob/a5378224e2f1d0441137645179dfeb076dccc642/src/sqlfluff/dialects/dialect_ansi.py#L79

https://github.com/sqlfluff/sqlfluff/blob/a5378224e2f1d0441137645179dfeb076dccc642/src/sqlfluff/dialects/dialect_tsql.py#L123

https://github.com/sqlfluff/sqlfluff/blob/a5378224e2f1d0441137645179dfeb076dccc642/test/core/parser/lexer_test.py#L86

https://github.com/sqlfluff/sqlfluff/blob/a5378224e2f1d0441137645179dfeb076dccc642/test/core/parser/lexer_test.py#L88

jpy-git (commented):
@jpers36 Couple more places to update and add the comment, but otherwise looks good 👍

src/sqlfluff/dialects/dialect_ansi.py
jpers36: @jpy-git  got them!

jpy-git (approved):
@jpers36 looks great, thanks for adding this! LGTM 🚀

PR #2188, opened 2021-12-23
ALTER GROUP fix to accommodate quoted objects

tdstark:
### Brief summary of the change made
This is a fix for issue #2186. `NakedIdentifierStatement` will be changed to `ObjectReferenceSegment` to accommodate quoting.

### Are there any other side effects of this change that we should be aware of?
No

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

Issue #2186, opened 2021-12-23
ALTER GROUP statements are unparsable


tdstark:
The following statement is unparsable:
`ALTER GROUP test_group ADD USER "test.user";`

## Expected Behaviour
The aforementioned statement should be parsable.

## Observed Behaviour
The aforementioned statement is not parsable.

## Dialect
ANSI/Redshift

## Version
`0.9.0`

## Configuration
Default configuration

jpy-git:
@tdstark when I originally made the PR to add this statement the docs weren't very clear as to whether or not the user/groups could be quoted: https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_GROUP.html

Could you quickly confirm for me if the group name can also be quoted please?
```sql
ALTER GROUP "test.group" ADD USER "test.user";
```

tdstark:
> @tdstark when I originally made the PR to add this statement the docs weren't very clear as to whether or not the user/groups could be quoted: https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_GROUP.html
> 
> Could you quickly confirm for me if the group name can also be quoted please?
> 
> ```sql
> ALTER GROUP "test.group" ADD USER "test.user";
> ```

I ran a quick test and users and groups can both be quoted. I think it's as easy as changing `NakedIdentifierSegment` to `ObjectReferenceSegment`.

jpy-git:
@tdstark makes sense, you want to take it? (happy to do it if not 😄)

tdstark:
> @tdstark makes sense, you want to take it? (happy to do it if not 😄)

I've got it!

jpy-git:
This was fixed by @tdstark s PR #2188 💪 

Closing

PR #2190, opened 2021-12-23
Modifying Redshift USER/GROUP Statements To Use `ObjectReferenceSegment`

tdstark:
This PR is an extension of this [PR](https://github.com/sqlfluff/sqlfluff/pull/2188) and this [issue](https://github.com/sqlfluff/sqlfluff/issues/2186). Redshift allows for quoted objects to be used for both group and user names, and this PR will modify ALTER/CREATE USER/GROUP statements to allow for this.

I’ve also added the other missing ANSI keywords as part of this PR to the RedShift dialect so that the real error is shown as it confused me at first as SQLFluff threw unrelated errors as it try to use other (unrelated) ANSI segments in an attempt to resolve this segment.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_redshift_keywords.py
```diff
@@ -202,6 +202,7 @@
 ATTACH
 ATTRIBUTE
 ATTRIBUTES
+AUTO_INCREMENT
```
jpy-git, line 205: Why are you adding in these keywords in this PR, there's no mention of motivation and they aren't used in the CREATE/ALTER USER/GROUP grammar or tests?

If they are intended please elaborate on why or provide examples of their usage in the CREATE/ALTER USER/GROUP unit tests.

If they're unrelated but you want to add them then make a separate PR 👍

src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -317,7 +317,7 @@ class CreateUserSegment(BaseSegment):
     match_grammar = Sequence(
         "CREATE",
         "USER",
-        Ref("NakedIdentifierSegment"),
+        Ref("ObjectReferenceSegment"),
```
jpy-git, line 320: Agree with these changes but you should add some additional unit tests where these segments are quoted.

jpy-git (commented):
@tdstark Needs some extra unit tests and a question on why the reserved keywords are included. Otherwise looks good so far 👍

src/sqlfluff/dialects/dialect_redshift_keywords.py
tdstark: @jpy-git Myself and @tunetheweb had a discussion a while back about this on Slack. It seems that in order for Redshift to be able to parse ALTER USER or ALTER GROUP statements, those need to be added to the dialect.  Otherwise, I run into Keyword errors. I'm happy to add these in another PR though, but it seems to me that this is a blocker. Should I spin this off into another PR?

jpy-git: My point is that they aren't referenced in any of the documentation for these statements:
- https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_USER.html
- https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_USER.html
- https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_GROUP.html
- https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_USER.html

If you were to remove them from this PR none of the existing unit tests would fail. That means if someone were to, for example, come along in the future and tidy up the list of redshift keywords then they could delete these because they're not in the [docs](https://docs.aws.amazon.com/redshift/latest/dg/r_pg_keywords.html) and they wouldn't know that they're valid/used since there's no tests that would fail if they get removed.

Basically fine with you including them in this PR if they actually get used in some of these statements and I understand that sometimes the docs can be incomplete. All you need to do is just put some simple unit tests in the relevant statement where they actually get used (same applies if you split this into a separate PR).

tunetheweb: These are not needed by RedShift dialect. They can be helpful due to the way SQLFluff works and dialect inheritance.

So basically if a grammar is not recognised, SQLFluff will eventually try to go through all the segments trying to make a match. This means it will also go through all the segment sin the ANSI dialect RedShift inherits from, and reach "ANSI" segments that don't exist in RedShift. Then rather than saying a segment is unparsable it will say a more cryptic message like "Grammar refers to 'Reference_autoincrementKeywordSegment' which was not found in the redshift dialect`.

So choices are:
1. Add all keywords in a dialect that it is inherited from (ANSI usually) as unreserved keywords. No real downside to this except, as @jpy-git points out someone might remove them thinking they are not needed. Maybe a comment will help, but haven't really done that for other dialects (though maybe should).
2. Override all non-supported ANSI segments in each dialect so unsupported Keywords are never tried to be used. This is probably the "right" answer but a bit unrealistic IMHO.
3. Do nothing and just deal with the somewhat cryptic error message (which should only happen in unparsable segments anyway).

Personally I prefer option 1. We shouldn't add unit tests for them, but maybe should add a comment.

tunetheweb: Found the slack conversation for this: https://sqlfluff.slack.com/archives/D02L1N7EATB/p1635725867011800

tunetheweb (approved):
I'm good with this to merge as is.

src/sqlfluff/dialects/dialect_redshift_keywords.py
jpy-git: Most dialects just inherit their keywords from ANSI for that reason to avoid this, so not sure why redshift explicitly clears them [here](https://github.com/sqlfluff/sqlfluff/blob/212432f5d9830566a23479769d770eaeb7774317/src/sqlfluff/dialects/dialect_redshift.py#L33) if we then have to then manually re-add the ANSI keywords into the redshift_keywords. Makes the Redshift keywords a bit messy if we want to implement option 2 in the future.
(@jpers36 recently implemented option 2 for T-SQL in #2170 so it's a good thing to aim for as a dialect gets more complete)

jpy-git: Also these keywords are def not relevant to the title/description of this PR so should be separate with details on motivation (can't see the slack chat btw) in the future.

I'm not fussed either way so feel free to merge if you're happy 👍

tunetheweb: Inheriting keywords from ANSI can have other unintended side effects (especially for ANSI reserved keywords that shouldn’t be reserved in this dialect) but yes that is another option. The reserved list of keywords in ANSI is quite small anyway (and mostly reserved in other dialects too) and don’t _think_ there is a problem with a reserved keyword also being unreserved.

Anyway, I’ve added a comment to the “Are there any other side effects…” section of the initial comment to explain why these are being added despite only being tangentially related to this PR.

PR #2192, opened 2021-12-24
Grammar: Add SET sql_log_bin statement to MySQL dialect

jpy-git:

Pretty simple, just need to add a "ON|OFF"/"TRUE|FALSE" keyword option to the existing `SET` statement to parse correctly. i.e. "ON|OFF"/"TRUE|FALSE" should match keyword capitalisation not variable capitalisation.

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -316,6 +316,14 @@ class TableConstraintSegment(BaseSegment):
         name="declared_variable",
         type="variable",
     ),
+    BooleanDynamicSystemVariablesGrammar=OneOf(
+        # Boolean dynamic system varaiables can be set to ON/OFF, TRUE/FALSE, or 0/1:
```
tunetheweb, line 320: 0/1 is not covered here. Guessing not needed in this Grammar but should we at least add a test case to show this works anyway?

jpy-git: yeah this grammar is just to get these parsing as keywords and therefore have the correct capitalisation. Best to let 0/1 parse in the same way as the rest of the integers 👍 

I've added an extra unit test

tunetheweb (approved):
LGTM.

Might want to update that first comment and AND and OR as doesn’t seem to match what you ended up implementing?

jpy-git:
> Might want to update that first comment about AND and OR as doesn’t seem to match what you ended up implementing?

Ah yep that was a typo in the PR description sorry, fixed it now 👍

PR #2197, opened 2021-12-24
Grammar: Fix multi-character comparison operators

### LINKED ISSUE #1997: Implement fix rule for comparison operators with space inside
TSQL and Oracle (ANSI dialect) allow having comparison operators written with spaces inside (such as ">  =", "<    =").
Please add a possibility in SQLFluff to fix such cases and remove these whitespaces inside, i.e. to fix from "> =" to ">=".

## Dialect
TSQL, ANSI

### LINKED ISSUE #2209: L006/TSQL: rule wants to split certain compound comparison operators
This is caused by #1965.

## Expected Behaviour

L006 does not flag/fix the following code:

```
select *
from test
where 1 != 1
```

## Observed Behaviour

L006 changes the code to
```
select *
from test
where 1 ! = 1
```

## Dialect

tsql

## Version

latest main

jpy-git:

Partially addresses #1997. Closes #2209. T-SQL allows for spaces in its comparison operators (introduced in #1965), however operators with and without spaces were essentially treated separately in this implementation making it hard to generally reference them in rules and also has introduced some new parsing errors detailed [here](https://github.com/sqlfluff/sqlfluff/issues/1997#issuecomment-995364941). 

This PR makes the multi-character comparison operator logic consistent between T-SQL and ANSI (inspired by the BitwiseLShiftSegment segment in ANSI ([here](https://github.com/sqlfluff/sqlfluff/blob/cb5c7f9448595a62f31aa44bb7301a8f92702fbb/src/sqlfluff/dialects/dialect_ansi.py#L1622)).

This has a couple nice properties:
- Groups underlying individual characters together.
- ANSI can use `allow_gaps=False` option to stop spaces between the individual characters, and T-SQL is the same but with `allow_gaps=True`.
- Rules can reference the comparison operator segments via their name as before, but now this also encompasses those operators with spaces inside.
- Removes a lot of redundant grammar from T-SQL.
- Will be very easy for me to update rule L039 to apply fixes to comparison operator segments containing whitespace and resolve #1997 🎉 
- I was also looking at extending the comparison operators in MySQL to add the additional ones in and this will also be useful for those (e.g. `<=>`, `&&`, ...).

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -83,7 +83,7 @@
         RegexLexer("atsign_literal", r"@[a-zA-Z_][\w]*", CodeSegment),
         RegexLexer("dollar_literal", r"[$][a-zA-Z0-9_.]*", CodeSegment),
     ],
-    before="not_equal",
+    before="like_operator",
```
jpy-git, line 86: not_equal segment is now a multi-character segment rather than a lexer so just inserting before the next item down here 👍

src/sqlfluff/rules/L049.py
```diff
@@ -38,6 +38,9 @@ def _eval(self, context: RuleContext) -> LintResult:
         # Context/motivation for this rule:
         # https://news.ycombinator.com/item?id=28772289
         # https://stackoverflow.com/questions/9581745/sql-is-null-and-null
+        if not context.segment.is_type("expression"):
+            return LintResult()
```
jpy-git, line 42: Added this as L049 can only apply to expression segments so no point doing the loops and various checks below for non expression segments 👍

test/fixtures/dialects/ansi/arithmetic_a.yml
```diff
@@ -20,7 +20,9 @@ file:
               - binary_operator: '*'
               - literal: '3'
               end_bracket: )
-          - comparison_operator: '>='
+          - comparison_operator:
```
barrywhart, line 23: I notice the grouping segment and the items inside it have the same type. Is this potentially tricky for rule writers?

barrywhart (commented):
Looks good overall. One question about name "overloading".

test/fixtures/dialects/ansi/arithmetic_a.yml
jpy-git: @barrywhart 
I would definitely want a lone `>` to have the same type as the `>=` grouping.
I guess the easy solution in rules is just say find a segment of type `comparison_operator` whose parent is not also a `comparison_operator` and infer it's kind via the name.
(it's a shame we don't store a parent attribute on the segments (like a linked list) rather than the context parent stack method because then you could just say `if segment.is_type("comparison_operator") and not segment.parent.is_type("comparison_operator")`, as well as doing any other required traversing without requiring context)

I think its a better solution that the previous situation where we had previously for T-SQL:
```
- comparison_operator: '>'
- whitespace: ' '
- comparison_operator: '='
```
since we had no obvious grouping or programatic way to refer to this as `greater_than_equal_to`.

Is there an easy way to overwrite the type of the segments within the grouping, but not the lone `>`? 

If not potentially something like this would work:
```python
RawGreaterThanSegment=StringParser(
    ">", SymbolSegment, name="raw_greater_than", type="raw_comparison_operator"
),
RawEqualsSegment=StringParser(
    "=", SymbolSegment, name="raw_equals", type="raw_comparison_operator"
)
...
@ansi_dialect.segment()
class EqualsSegment(BaseSegment):
    """equals operator."""

    type = "comparison_operator"
    name = "equals"
    match_grammar = Ref("RawEqualsSegment")

@ansi_dialect.segment()
class GreaterThanSegment(BaseSegment):
    """Greater than operator."""

    type = "comparison_operator"
    name = "greater_than"
    match_grammar = Ref("RawGreaterThanSegment")

@ansi_dialect.segment()
class GreaterThanOrEqualToSegment(BaseSegment):
    """Greater than or equal to operator."""

    type = "comparison_operator"
    name = "greater_than_equal_to"
    match_grammar = Sequence(
        Ref("GreaterThanSegment"), Ref("EqualsSegment"), allow_gaps=False
    )
```
That way `segment.type = "comparison_operator"` only matches the grouped segments and also makes single character comparison operators consistent AST-wise:
```
# Equals
- comparison_operator:
    - raw_comparison_operator: '='
# Greater Than
- comparison_operator:
    - raw_comparison_operator: '>'
# Greater Than or Equal To
- comparison_operator:
    - raw_comparison_operator: '>'
    - raw_comparison_operator: '='
```
Main question there is if you think that's too verbose vs just including a condition in the rules?

barrywhart: I don't have a strong opinion either way. This feels like a classic tradeoff -- make X more complex to make Y less complex. On the one hand, SQLFluff has many dialects, but on the other hand, it also has lots of rules. I'm also unsure how many rules will be specifically dealing with comparison operators.

I'm okay leaving it "as is" for now. We can adjust later if needed as we develop and maintain rules.

Your proposal for how to do it using the grammar looks like it would work well.

barrywhart (approved):

test/fixtures/dialects/ansi/arithmetic_a.yml
jpy-git: yeah I think that's fair enough, the raw suggestion shouldn't be too hard to implement but lets hold off until we actually have a need 👍

jpy-git:
@barrywhart I'll do a quick check of the rules where we reference the comparison operators to see if we should add that "is parent also a comparison operator check" to be on the safe side tomorrow before we merge this 😄

test/fixtures/dialects/ansi/arithmetic_a.yml
jpy-git: Going to look at this as I think it's worth keeping it simple on the rules side

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -245,29 +243,20 @@
...
-    ),
-    GreaterThanOrEqualToSegment=StringParser(
-        ">=", SymbolSegment, name="greater_than_equal_to", type="comparison_operator"
+    RawNotSegment=StringParser(
+        "!", SymbolSegment, name="raw_not", type="raw_comparison_operator"
     ),
-    LessThanOrEqualToSegment=StringParser(
-        "<=", SymbolSegment, name="less_than_equal_to", type="comparison_operator"
+    RawEqualsSegment=StringParser(
+        "=", SymbolSegment, name="raw_equals", type="raw_comparison_operator"
```
jpy-git, line 253: @barrywhart I've added the raw segments as we discussed [here](https://github.com/sqlfluff/sqlfluff/pull/2197#discussion_r775085427). Now `.is_type("comparison_operator")` will unambiguosly refer to the combined comparison operators.

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1657,13 +1645,76 @@ class CaseExpressionSegment(BaseSegment):
...
+@ansi_dialect.segment()
+class GreaterThanOrEqualToSegment(BaseSegment):
+    """Greater than or equal to operator."""
+
+    type = "comparison_operator"
+    name = "greater_than_equal_to"
+    match_grammar = Sequence(
+        Ref("RawGreaterThanSegment"), Ref("RawEqualsSegment"), allow_gaps=False
+    )
+
```
jpy-git, line 1684: You can see the actual comparison operators are built from the raw_comparison_operators and ensures consistent AST for all comparison operators.

test/fixtures/dialects/ansi/create_table_column_constraint.yml
```diff
@@ -129,12 +130,15 @@ file:
...
+              - comparison_operator:
+                - raw_comparison_operator: '>'
+                - raw_comparison_operator: '='
               - literal: '18'
               - binary_operator: AND
               - column_reference:
                   identifier: City
-              - comparison_operator: '='
+              - comparison_operator:
+                  raw_comparison_operator: '='
```
jpy-git, line 141: good example of consistent AST. rules can just refer to `is_type("comparison_operator")` and use the name to decide whether its greater than, less than as before 👍

jpy-git (commented):
@barrywhart this is ready for review now.

N.B. it changes a lot of the yaml files since, for consistency, EqualsSegment now parses like:
```
- comparison_operator:
    raw_comparison_operator: '='
```

The only thing that's changed since your previous review is adding the raw segments and I've highlighted relevant code in this review 😄

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -405,6 +359,65 @@ class StatementSegment(ansi_dialect.get_segment("StatementSegment")):  # type: i
     parse_grammar = match_grammar
 
 
+@tsql_dialect.segment(replace=True)
+class GreaterThanOrEqualToSegment(BaseSegment):
+    """Greater than or equal to operator.
+
+    N.B. Patching to add !< and
+    to allow spaces between operators.
```
barrywhart, line 367: Is this part of the comment still accurate?
>to allow spaces between operators.

(Maybe so, wasn't sure since I don't see anything about `allow_gaps` here.)

barrywhart (approved):
Looks good. One question about whether a docstring is still accurate.

src/sqlfluff/dialects/dialect_tsql.py
jpy-git: Yeah its the lack of the `allow_gaps` operator here that allows the gaps since its True by default. The relevant T-SQL test fixtures pass 😄 

example:
image

PR #2210, opened 2021-12-29
Remove oyaml in favour of pyyaml

jpy-git:

Currently we have `oyaml` as a dependency in order to preserve the ordering of keys in yaml (useful for comparing test results). However, all `oyaml` does is provide a thin wrapper around `pyyaml` and `pyyaml` itself has provided this capability since v5.1.

Therefore we should look to move our dependency to the larger `pyyaml` project.

Also type stubs are available for `pyyaml` so I have added those to dev dependencies for better mypy coverage.

No

### Pull Request checklist

  - Other.


tunetheweb (approved):
LGTM, but might be good if someone with more python experience approves this.

@barrywhart any concerns?

Did find this which seems to back up what's being said here: https://github.com/wimglenn/oyaml/issues/22

barrywhart (commented):
Looks good. For sanity's sake, have we tried running the updated `test/generate_parse_fixture_yml.py` script to confirm the YAML files it generates are in fact identical to before?

tunetheweb:
We have a CI check for that which would have failed if it didn’t generate the exact same.

PR #2217, opened 2021-12-31
Snowflake: Split out `CREATE VIEW` into its own segment

### LINKED ISSUE #2000: Snowflake create view statement results in parse error
We define the column_list in all of our views and add comments to them. However when running sqlfluff with snowflake dialect we're getting a parse error.  

https://docs.snowflake.com/en/sql-reference/sql/create-view.html -> search for column_list

## Expected Behaviour
Parsing the SQL should be no issue

## Observed Behaviour
Console output when running sqlfluff lint .\test.sql --dialect snowflake
```
== [test.sql] FAIL
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: "create
                       | view dwh.vw_dummy \ncomment = 'The..."
```
When given this test.sql file: 
```
create view dwh.vw_dummy
comment = 'The purpose of this view' -- this doesnt result in a parse error
( -- this section does result in a parse error
    column1             comment 'Specs for column1'
,   column2             comment 'Specs for column2'
)
as
select foo
```

## Steps to Reproduce
Create test.sql with the contents above
run: `sqlfluff lint .\test.sql --dialect snowflake`

## Dialect
Snowflake

## Version

```
> sqlfluff.exe --version
sqlfluff, version 0.8.2
> py --version
Python 3.9.6
```

## Configuration
Default configuration

wong-codaio:

Snowflake dialect only: Create a segment specifically for Snowflake's `CREATE VIEW` statement.

* Removed `VIEW` from the general `CreateStatement` in Snowflake.
* Consolidated multiple test files around view creation into a single test file.
* Added test cases to support the existing as well as some additional functionality on CREATE VIEW.

fixes #2000

#### Context

* There is a `CreateStatementSegment` in Snowflake that aims to encapsulate all possibilities of `CREATE` statement in Snowflake. The statement however does not account for the specific needs that is targeted only for `CREATE VIEW` statement. 
* While it is possible to fix #2000 by adding extra clauses to the `CreateStatement` segment, it seems it's more appropriate to create a separate class for `CreateViewStatementSegment` - Which makes Snowflake more aligns what `dialect_ansi` does (Ansi also has a dedicated `CreateViewStatementSegment`.

* `CREATE VIEW` segment type has been changed - from `create_statement` to `create_view_statement` in Snowflake dialect.
   * If anyone has written custom rules targeting `CREATE VIEW` in Snowflake's dialect, and is using `create_statement` as a check in their matching logic, then it'll no longer work.

### Pull Request checklist

      * Refactored all CREATE VIEW test cases into a single `snowflake_create_view.sql` file.
   * Added a link to CREATE VIEW documentation on Snowflake's site.


wong-codaio:
@tunetheweb @Dikootje - Wanted to get how you guys feel about this change.

* I could fix #2000 by adding additional clauses in the `CreateStatement` segment, but it felt more appropriate to have a dedicated class to handle edge cases around `CREATE VIEW`, rather than adding more complex logic into the overloaded `CreateStatement` in Snowflake.

jpy-git:
> * I could fix [Snowflake create view statement results in parse error #2000](https://github.com/sqlfluff/sqlfluff/issues/2000) by adding additional clauses in the `CreateStatement` segment, but it felt more appropriate to have a dedicated class to handle edge cases around `CREATE VIEW`, rather than adding more complex logic into the overloaded `CreateStatement` in Snowflake.

I agree that it's a good to split into it's own segment, it better aligns with the docs and is easier to check/maintain. I imagine all the CREATE statements were originally combined to quickly get the basic statements parsing but as we develop them more we can split them out properly 👍

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -1772,7 +1772,6 @@ class CreateStatementSegment(BaseSegment):
             Sequence("NOTIFICATION", "INTEGRATION"),
             Sequence("SECURITY", "INTEGRATION"),
             Sequence("STORAGE", "INTEGRATION"),
-            "VIEW",
             Sequence("MATERIALIZED", "VIEW"),
             Sequence("SECURE", "VIEW"),
```
jpy-git, line 1776: Pretty sure this is covered by your create view statement now so can be removed?

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -1842,6 +1841,56 @@ class CreateStatementSegment(BaseSegment):
...
+    match_grammar = Sequence(
+        "CREATE",
+        Ref("OrReplaceGrammar", optional=True),
+        AnyNumberOf(
+            "SECURE",
+            "RECURSIVE",
+        ),
+        "VIEW",
+        Ref("IfNotExistsGrammar", optional=True),
+        Ref("ObjectReferenceSegment"),
```
jpy-git, line 1862: This is probably a holdover of the fact that there were multiple possible objects that could be being referenced in the generic CREATE statement.

In the rest of the codebase we tend to use `Ref("TableReferenceSegment")` to refer to view objects (since they are treated the same as tables in most rules).

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -1842,6 +1841,56 @@ class CreateStatementSegment(BaseSegment):
...
+            ),
+            Sequence(
+                Sequence("WITH", optional=True),
+                "ROW",
+                "ACCESS",
+                "POLICY",
+                Ref("NakedIdentifierSegment"),
+                "ON",
+                Bracketed(
+                    Delimited(Ref("NakedIdentifierSegment")),
```
jpy-git, line 1882:
```suggestion
                    Delimited(Ref("ColumnReferenceSegment")),
```

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -1842,6 +1841,56 @@ class CreateStatementSegment(BaseSegment):
...
+            Sequence(
+                Bracketed(
+                    Delimited(
+                        Sequence(
+                            Ref("ObjectReferenceSegment"),
+                            Ref("CommentClauseSegment", optional=True),
+                        ),
+                    ),
+                ),
+            ),
```
jpy-git, line 1873:
```suggestion
            Bracketed(
                Delimited(
                    Sequence(
                        Ref("ColumnReferenceSegment"),
                        Ref("CommentClauseSegment", optional=True),
                    ),
                ),
                optional=True,
            ),
```

jpy-git: The outer Sequence is unnecessary and you should use `ColumnReferenceSegment`

src/sqlfluff/dialects/dialect_snowflake.py
jpy-git: These are columns

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -1842,6 +1841,56 @@ class CreateStatementSegment(BaseSegment):
...
+            Sequence(
+                Sequence("WITH", optional=True),
```
jpy-git, line 1875:
```suggestion
                Ref.keyword("WITH", optional=True),
```

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -1842,6 +1841,56 @@ class CreateStatementSegment(BaseSegment):
     )
 
 
+@snowflake_dialect.segment(replace=True)
+class CreateViewStatementSegment(BaseSegment):
+    """A `CREATE VIEW` statement, specifically for Snowflake's dialect.
+
+    https://docs.snowflake.com/en/sql-reference/sql/create-view.html
```
jpy-git, line 1848: We're missing these options from the segment:
```
[ <col1> [ WITH ] MASKING POLICY <policy_name> [ USING ( <col1> , <cond_col1> , ... ) ]
           [ WITH ] TAG ( <tag_name> = '<tag_value>' [ , <tag_name> = '<tag_value>' , ... ] ) ]
  [ , <col2> [ ... ] ]
```
Any interest in including them in this PR? If not you should add a TODO note to the docstring to let people know it's incomplete.

test/fixtures/dialects/snowflake/snowflake_create_view.sql
```diff
@@ -0,0 +1,42 @@
...
+    col2 COMMENT 'some comment'
+)
+AS SELECT col1, col2 FROM src_table;
+
+CREATE OR REPLACE SECURE RECURSIVE VIEW IF NOT EXISTS view_with_all_implemented_features
+    COMMENT = 'table-level comment'
+    (
+        col1,
+        col2 COMMENT 'some comment'
+    )
```
jpy-git, line 37: Feels weird not defining the column list immediately after the view name. Could you confirm that this is valid syntax in snowflake?

Same question for the other options that we're allowing to parse before the column list?
e.g.
```sql
CREATE OR REPLACE VIEW IF NOT EXISTS view_with_all_implemented_features
    WITH TAG (foo = 'bar', hello = 'world')
    COPY GRANTS
    (
        col1,
        col2 COMMENT 'some comment'
    )
AS SELECT col1, col2 FROM src_table;
```

src/sqlfluff/dialects/dialect_snowflake.py
wong-codaio: 👍  Will add a note/TODO.

FYI - The complexity of column-level masking policy/tagging support is the `comma` that was added when multiple columns are involved, and per Snowflake's grammar, it doesn't specify that all column-level must be grouped together, making it difficult to craft the matching logic for it.

test/fixtures/dialects/snowflake/snowflake_create_view.sql
wong-codaio: I don't have access to the tag feature in my Snowflake instance, but I was able to verify that this syntax works.

```
CREATE OR REPLACE VIEW sandbox.test_view COMMENT = 'hello'
COPY GRANTS
(
  user_id, company_id COMMENT 'some comment'
)
AS SELECT id_user, id_workspace FROM sandbox.test_table;
```

src/sqlfluff/dialects/dialect_snowflake.py
wong-codaio: For my education: When should we use `Sequence(...)` vs `Ref.keyword(...)`?

src/sqlfluff/dialects/dialect_snowflake.py
wong-codaio: Added a TODO: https://github.com/sqlfluff/sqlfluff/pull/2217/files#diff-a8da7f7dc8f0e448950f6d5c2cd6cd5590026ea6da32cdc8f233e037bea947ecR1885

src/sqlfluff/dialects/dialect_snowflake.py
wong-codaio: Removed: https://github.com/sqlfluff/sqlfluff/pull/2217/files#diff-a8da7f7dc8f0e448950f6d5c2cd6cd5590026ea6da32cdc8f233e037bea947ecL1777

src/sqlfluff/dialects/dialect_snowflake.py
wong-codaio: Updated: https://github.com/sqlfluff/sqlfluff/pull/2217/files#diff-a8da7f7dc8f0e448950f6d5c2cd6cd5590026ea6da32cdc8f233e037bea947ecR1861

src/sqlfluff/dialects/dialect_snowflake.py
wong-codaio: Updated: https://github.com/sqlfluff/sqlfluff/pull/2217/files#diff-a8da7f7dc8f0e448950f6d5c2cd6cd5590026ea6da32cdc8f233e037bea947ecR1866

src/sqlfluff/dialects/dialect_snowflake.py
jpy-git: So you can think of `Ref.keyword` as the explicit way to refer to a keyword segment. Usually you don't need to specify it and can just write the keyword e.g. `"WITH"` but it's useful when you want to specify additional parameters e.g. an optional WITH keyword: `Ref.keyword("WITH", optional=True)`.

`Sequence` is a sequence of multiple segments e.g. you could specify an optional group of segments `Sequence("IF", "EXISTS", optional=True)` and the entire sequence is optional, you can't have one of the segments without the others.

It's not completely wrong to use `Sequence` here, it just makes more sense in this case to read it as an optional WITH rather than an optional sequence of length 1 😄

test/fixtures/dialects/snowflake/snowflake_create_view.sql
jpy-git: Nice thanks for confirming! 😄

jpy-git (approved):
@wong-codaio looks great, thanks! LGTM 🚀 ⭐

Dikootje:
Thanks guys!

PR #2218, opened 2022-01-01
Fix CLI arguments to allow for autocompletion

jpy-git:

click allows for users to optionally enable autocompletion for a command: https://click.palletsprojects.com/en/8.0.x/shell-completion/?highlight=shell

Currently the CLI is not using the correct click types for some arguments/options which prevents click autocompletion from working correctly. This PR resolves that.

No, zero impact on those who don't want to use it.

### Pull Request checklist

  - Other.


jpy-git:
@barrywhart @tunetheweb think this is a pretty neat addition 😄

src/sqlfluff/cli/commands.py
```diff
@@ -383,6 +422,35 @@ def dialects(**kwargs) -> None:
...
+    "--save-path",
+    default=None,
+    type=click.Path(),
+    help="Where to save the completion script. Defaults to ~/.config/sqlfluff",
+)
+@click.argument(
+    "shell_type",
+    default="bash",
+    type=click.Choice(["bash", "zsh", "fish"], case_sensitive=False),
+    # help="What shell to generate autocompletion script for.",
```
barrywhart, line 437: Did you mean to leave this commented out?

src/sqlfluff/cli/autocomplete.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # Ensure that parent config directory exists.
+    save_path = os.path.expanduser(save_path)
+    save_folder = os.path.dirname(save_path)
+    if not os.path.exists(save_folder):
+        os.makedirs(save_folder)
+
+    # Generate the autocompletion script.
+    subprocess.run(
+        [f"_SQLFLUFF_COMPLETE={shell_type.lower()}_source sqlfluff > {save_path}"],
```
barrywhart, line 43: What is this doing exactly? What does it it run?

src/sqlfluff/cli/commands.py
```diff
@@ -383,6 +422,35 @@ def dialects(**kwargs) -> None:
...
+    This saves an autocompletion shell script to ~/.config/sqlfluff that
+    the user can enable by adding the following line to their shell config:
+    . ~/.config/sqlfluff/sqlfluff-complete.{shell_type}
+
+    N.B. this is only available bash, zsh, and fish.
+    """
+    config = get_config(**kwargs)
+    autocomplete_dict = generate_autocomplete_script(shell_type, save_path)
+    click.echo(format_autocomplete(autocomplete_dict), color=config.get("color"))
+    _completion_message(config)
```
barrywhart, line 451: Where is this function defined?

src/sqlfluff/cli/commands.py
jpy-git: yeah you can't have help on a positional argument so will remove that

src/sqlfluff/cli/autocomplete.py
jpy-git: It literally just runs SQLFluff with an environment variable `_SQLFLUFF_COMPLETE` that tells `click` to produce the relevant shell autocompletion script:
image
the `> {save_path}` just saves the stdout to a file at the save path.

Docs: https://click.palletsprojects.com/en/8.0.x/shell-completion/?highlight=shell

src/sqlfluff/cli/commands.py
jpy-git: `_completion_message`?
It's just the usual `All Finished 📜 🎉!` message: https://github.com/sqlfluff/sqlfluff/blob/68f479a53cfcabe67379dfb96539e516bf80d3c8/src/sqlfluff/cli/commands.py#L744

src/sqlfluff/cli/autocomplete.py
barrywhart: Would it be possible (desirable?) to do this using a Python function call rather than `subprocess.run()`? There's probably a tradeoff here, just thinking that the `subprocess.run()` reads a bit cryptic, and could a Python call be more readable. 🤔

jpy-git: click doesn't provide an API to do it since it requires running via the CLI entry point to determine the relations between commands/sub-commands/args.

Relevant docs: https://click.palletsprojects.com/en/8.0.x/shell-completion/?highlight=shell#enabling-completion

Main reason for adding this as a sub-command to SQLFluff is to wrap the console command since I imagine the average non-technical SQLFluff user may be more comfortable that way.

That being said if we don't want the autocomplete sub-command we can just bin it and include a reference to the click shell completion section in the docs for those users who want to enable it. We should still keep the argument changes introduced in this PR as they are what tells autocompleter what to do for those who enable it.

barrywhart: I'm good with adding it, was just curious about options for implementing it. Your explanation makes sense. Think it'd be worth summarizing any bits of the above discussion as code comments?

jpy-git: I've reduced the scope of the PR to just fix CLI arguments/options so people can use the feature if they want. I'll make a later PR with a small doc note linking to click docs.

src/sqlfluff/cli/autocomplete.py
```diff
@@ -0,0 +1,17 @@
...
+from click.shell_completion import CompletionItem
+
+from sqlfluff import list_dialects
+
+
+def dialect_shell_complete(ctx, param, incomplete):
+    """Shell completion for possible dialect names.
+
+    We use this over click.Choice as we want to internally
+    handle errors messages and codes for incorrect/outdated dialects.
```
barrywhart, line 12:
```suggestion
    handle error messages and codes for incorrect/outdated dialects.
```

barrywhart (approved):
Looks good! One typo in a docstring, otherwise 💯 .

Issue #2046, opened 2021-12-06
dbt - L009: Files must end with a single trailing newline.


tekumara:
## Expected Behaviour
No error

## Observed Behaviour
```
L:  5 | P:  60 | L009 | Files must end with a single trailing newline.    
```

## Steps to Reproduce
A dbt model that calls a macro, eg:
```
{{ config ( materialized='incremental',
   transient=false,
    ) }}

{{ generate_table('mytable') }}

```

## Dialect
snowflake

## Version
sqlfluff, version 0.8.2
python 3.9

## Configuration
```
[tool.sqlfluff.core]
templater = "dbt"
sql_file_exts = ".sql,.sql.j2,.dml,.ddl"
dialect = "snowflake"
ignore_templated_areas = true

[tool.sqlfluff.indentation]
indented_joins = false
indented_using_on = true
template_blocks_indent = false

[tool.sqlfluff.templater.dbt]
project_dir = "."
profile = "snowflake"
```

tekumara:
Yep that's all.

Here's the tail of `sqlfluff parse`:
```
[L:  8, P:  1]      |                        identifier:                           'RunID'
[L:  8, P: 44]      |                    [META] dedent:
[L:  8, P:  1]      |    newline:                                                  '\n'
[L:  8, P:  1]      |    newline:                                                  '\n'
[L:  8, P: 44]      |    newline:                                                  '\n'
```

Here's the tail of the macro:
```
ORDER BY rl.RunID

{% endmacro %}

```

If I remove the newline before `{% endmacro %}` it produces two newlines in the parse, and still errors.

jpy-git:
Makes sense thanks! This is a known issue with the dbt templater.

In Jinja, the template lines are included as placeholders in the parse tree so it's possible to account for those
e.g.
![image](https://user-images.githubusercontent.com/80432516/145007044-7760e2cc-458f-46de-8468-fe1eac22a9f0.png)

However, as you can see from your parse tree, certain dbt templates don't show up in the parse tree. So as far as the rule can see the file finishes with 3 newlines. We encountered a similar error in L050 (no newlines at start of the file) https://github.com/sqlfluff/sqlfluff/pull/1840#issuecomment-962682902. I ended up adding some logic to infer the presence of those missing template blocks so I'll try and add some similar logic to L009.

I'll also create a separate issue to see if one of the dbt contributors can get those templates to show up in the parse tree as a more long term solution.

barrywhart:
[...]

PR #2220, opened 2022-01-01
Update BaseRule.discard_unsafe_fixes() to avoid touching templated code

### LINKED ISSUE #2117: With dbt templater, `sqlfluff fix <file>.sql` will overwrite lines if rule LO34 is triggered
## Expected Behaviour

`sqlfluff fix <file>.sql` should fix any formatting issues in place. 

## Observed Behaviour

`sqlfluff fix <file>.sql` will drop line and overwrite them with the code that is present in the macro, if rule LO34 is triggered and a macro exists in the code. This can occur when the macro contains parenthesis, if there is a `*` on a line other than the first one, or assumably other ways of triggering this rule that I am unaware of.

As a work around rule LO34 can be skipped with:

```
sqlfluff fix <file or dir> --exclude-rules L034
``` 

I believe this to be the root cause of the issues in:

* https://github.com/sqlfluff/sqlfluff/issues/1952
* https://github.com/sqlfluff/sqlfluff/issues/2056
* https://github.com/sqlfluff/sqlfluff/issues/1953

I also suspect it to be the cause of this one but don't have quite enough context for me to 100% convince myself:

* https://github.com/sqlfluff/sqlfluff/issues/2065

### Works as expected:

model.sql:
```
SELECT
   {{ normal() }}
  , line_two As line_two
FROM foo
```

macro.sql:
```
{% macro normal() %}
     col AS col
{% endmacro %}
```

`sqlfluff fix model.sql` makes model.sql:
```
SELECT
   {{ normal() }}
  , line_two AS line_two
FROM foo
```

### parenthesis wrapped (unexpected)

model.sql:
```
SELECT
   {{ par_wrap() }}
  , line_two As line_two
FROM foo
```

macro.sql:
```
{% macro par_wrap() %}
    ( col ) AS col
{% endmacro %}
```

`sqlfluff fix model.sql` makes model.sql:
```
SELECT
   {{ par_wrap() }}
  , ( col ) AS col
FROM foo
```

### in list (unexpected)

model.sql:
```
SELECT
   {{ in_list() }}
  , line_two As line_two
FROM foo
```

macro.sql:
```
{% macro in_list() %}
     col in ('var1', 'var2')
{% endmacro %}
```

`sqlfluff fix model.sql` makes model.sql:
```
SELECT
   {{ in_list() }}
  , col IN ('var1', 'var2')
FROM foo
```

A select * triggering Rule LO34

```
SELECT
   {{ normal() }}
  , line_two As line_two
  , *
FROM foo
```

macro.sql:
```
{% macro normal() %}
     col AS col
{% endmacro %}
```

`sqlfluff fix model.sql` makes model.sql:
```
SELECT
   {{ normal() }}
  , col AS col
  , second_line AS second_line
FROM foo
```

Similarly able to have this occur if the macro uses a sql functions that has parenthesis.

## Dialect

snowflake

## Version

sqlfluff: 0.9.0
sqlfluff-templater-dbt: 0.9.0
dbt:  0.21.1

I also tested this a few different versions of everything and it happened consistently.

## Configuration

```
[sqlfluff]
templater = dbt
dialect = snowflake

[sqlfluff:rules]
tab_space_size = 2
comma_style = leading

[sqlfluff:rules:L010]
capitalisation_policy = upper

[sqlfluff:templater:dbt]
profile = dbt_snowflake
```

(I believe the `templater = dbt` is the only part relevant for this bug but including all of it as is, just in case)

barrywhart:
Fixes #2117 

The linter will now be more stringent about not touching areas of templated code. The existing tests do a pretty good job of testing this, but it's possible we missed something.

### Pull Request checklist

  - Other.


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -155,7 +155,11 @@ def __eq__(self, other):
 
     def __hash__(self):
         return hash(
-            (self.__class__.__name__, self.raw, self.pos_marker.source_position())
+            (
```
barrywhart, line 158: Slightly unrelated fix, but this fixes an issue where segments lacking a `pos_marker` can't be hashed (thus can't be added to sets, etc.)

src/sqlfluff/core/rules/base.py
```diff
@@ -734,6 +795,7 @@ def valid_rule_name_regex(self):
         The second group captures the rule code.
 
         Examples of valid rule names:
+
```
barrywhart, line 798: Unrelated fix: Need a blank line here in order for Sphinx to render the bullet points correctly.

jpy-git:
@barrywhart just to be clear this doesn't have any impact on the invisible template sections that we discussed as part of https://github.com/sqlfluff/sqlfluff/issues/2046 or https://github.com/sqlfluff/sqlfluff/issues/2066 right?

src/sqlfluff/core/rules/base.py
```diff
@@ -668,6 +717,15 @@ def discard_unsafe_fixes(
                 lint_result.fixes = []
                 return
 
+        for fix in lint_result.fixes:
+            if fix.has_template_conflicts(templated_file):
+                linter_logger.info(
```
tunetheweb, line 722: Is `info` there right log level for this?

For a similar error we use `warning`:
https://github.com/sqlfluff/sqlfluff/blob/7fd452dbfb9ab8537b8ca6e92c9489ff04d6172b/src/sqlfluff/core/linter/linter.py#L329-L331

barrywhart:
@jpy-git: I don't think so. This fixes a class of issues where the entire set of fixes for a `LintResult` should be discarded if any of them touch templated code. The old behavior could end up applying some fixes but not others, which could corrupt the file. L034 is a common trigger for this, because it moves columns around. You can see this in the test case: two query columns get swapped. One of these fixes is okay, but the other isn't. If we allow one and not the other, it results in the behavior described in #2117 (and verified fixed by the test I added).

In this case, the templating outputs are non-empty strings.

src/sqlfluff/core/rules/base.py
barrywhart: I used `info()` because other checks in the same function use it. I think it's the correct level, because this is routine behavior: code is templated, and rules may try to do things that are legit in templated space, but when mapped back to raw space, the edits can't be applied. Users don't care about this as long as we behave correctly (i.e. disacarding the fixes that can't be applied).

src/sqlfluff/core/linter/linted_file.py
```diff
@@ -349,12 +349,12 @@ def fix_string(self) -> Tuple[Any, bool]:
                 filtered_source_patches.append(enriched_patch)
                 dedupe_buffer.append(enriched_patch.dedupe_tuple())
             # If it's ONLY templated then we should skip it.
-            elif "literal" not in local_type_list:
+            elif "literal" not in local_type_list:  # pragma: no cover
```
barrywhart, line 352: A couple areas in this file became "uncovered" because the upstream `has_template_conflicts()` check prevents hitting this code. We could potentially remove it, but TBH, I'd prefer leaving it in place, at least for now, as a second line of defense. (I'm not sure `has_template_conflicts()` eliminates all possibilities of hitting this code.)

jpy-git:
> e file. L034 is a common trigger for this, because it moves columns around. You can see this in the test case: two query columns get swapped. One of these fixes is okay, but the other isn't. If we allow one and not the other, it results in the behavior described in #2117 (and verified fixed by the test I added).

@barrywhart makes sense! Thanks for clarifying 😄

jpy-git (approved):
Looks good 👍

tunetheweb (approved):

PR #2223, opened 2022-01-02
Simplify nested cases

### LINKED ISSUE #1799: New rule: Nested CASEs with conditions on the same column

barrywhart:
Fixes #1799 

This is the first new rule implemented using the functional API. 🎉 

This rule works a bit differently than was initially suggested in the issue -- it doesn't check (nor does it need to) look at the _condition_ of the `WHEN` clause, because the `CASE`/`WHEN`/`ELSE` _structure_ alone is sufficient to allow simplification. It does simplify the example provided in the issue. 👍

### Pull Request checklist

  - Other.


barrywhart:
@jpy-git, @tunetheweb: This is ready for review. It uses the new API from the previous, large PR, but this should be a pretty straightforward PR: one new rule file plus the test YAML file.

src/sqlfluff/rules/L058.py
```diff
@@ -0,0 +1,89 @@
...
+from sqlfluff.core.rules.doc_decorators import document_fix_compatible
+from sqlfluff.core.rules.functional import sp
+
+
+@document_fix_compatible
+class Rule_L058(BaseRule):
+    """Nested CASE statements could be flattened.
+
+    | **Anti-pattern**
+    | In this example, the outer CASE's "ELSE" is another CASE.
```
tunetheweb, line 14:
```suggestion
    | In this example, the outer CASE's "ELSE" is an unnecessary other CASE.
```

test/fixtures/rules/std_rule_cases/L058.yml
```diff
@@ -0,0 +1,189 @@
...
+test_pass_1:
+  # The nested CASE is under a "WHEN", not an "ELSE".
+  pass_str: |
+    SELECT
+        c1,
+        CASE
+            WHEN species = 'Rat' THEN
+              CASE
+                  WHEN species = 'Dog' THEN 'Woof'
+              END
```
tunetheweb, line 12: I don't understand this example? Instead of returning 'squeak' we go into another case (also for species, but a different one, so this can never be true). I get it's just an example but should we pick a better one that makes sense?

```suggestion
            WHEN species = 'Rat' THEN
              CASE
                  WHEN colour = 'Black' THEN 'Growl'
                  WHEN colour = 'Grey' THEN 'Squeak'
              END
```

tunetheweb (approved):
A couple of minor suggestions

test/fixtures/rules/std_rule_cases/L058.yml
barrywhart: Ha, great point! I was in "robotic example mode." 🤪

jpy-git:
@barrywhart is the plan to add more cases in the future or is it just covering the ELSE case?

I get that some people would want to have nested case statements in when instead of more complicated conditions, but might be nice to have a config option to unnest the WHEN statements as well?

If it's just ELSE then "Nested CASE statements could be flattened" is a bit general for the description and should probably say it only covers ELSE. The average user will just see the short description rather than the examples when they get the linting violation in the CLI results?

barrywhart:
@jpy-git: It's a good point. I have no plan at the moment to add cases, but it's a possibility, based on user demand. Handling additional cases will require more care, as there's a greater risk of introducing bugs in the SQL. Think we should change the wording for now, unless and until the rule does more things?

jpy-git:
Yeah I think maybe just a quick PR to update the wording of the description and then we can assess expanding the rule in the future 😄

PR #1338, opened 2021-08-31
Postgres "ALTER TABLE" enhancement, and timestamp bug fix

WittierDinosaur:
The initial purpose of this PR was to enhance the functionality of ALTER TABLE is Postgres, to match https://www.postgresql.org/docs/13/sql-altertable.html

However, the PR also includes a bugfix, for the timestamp with time zone issue, as this was getting in the way of my testing. It also slightly expands the definition of Expression D in ANSI, to allow the following kind of expression:
`timestamp 'epoch' + foo_timestamp * interval '1 second'`

Fixes #1289, Fixes #1286, Fixes #1168, Fixes #1155

### Are there any other side effects of this change that we should be aware of?
Expands ANSI definitions of Expression_D, and timestamp with time zone (this is ANSI standard though). All current ANSI tests pass.

### Pull Request checklist
I have included a lot of test cases for alter table - though more may be required.
The only extra functionality needed is sequences, for which I haven't yet raised a ticket

PR #2227, opened 2022-01-03
Add L059 - Capitalization on Data Types

tdstark:
This will add rule L058 to test for capitalization on data types and will close [this issue/enhancement](https://github.com/sqlfluff/sqlfluff/issues/2228). It is built off of rule L010 and follows upper, lower, and capitalize conventions found in other capitalization rules.

### Pull Request checklist

  - Other.


tdstark:
Currently, this cannot resolve `SERIAL` data types because it is not a `data_type_identifier` but rather a `keyword`. If anyone has any suggestions about how to resolve this, please let me know.

tdstark:
@jpy-git here is the PR we were chatting about.

test/fixtures/rules/std_rule_cases/L059.yml
```diff
@@ -0,0 +1,77 @@
...
+
+test_fail_capitalisation_policy_keywords:
+  # Test cases where data types are keywords, not data_type_identifiers
+  # See: https://github.com/sqlfluff/sqlfluff/pull/2121
+  fail_str: CREATE TABLE table1 (account_id BIGINT, column_two serial);
+  fix_str: CREATE TABLE table1 (account_id BIGINT, column_two SERIAL));
+  configs:
+    rules:
+      L059:
+        capitalisation_policy: upper
```
jpy-git, line 77: Could you also add this as a test to L010 so we can see if that rule changes the capitalisation atm?

Something like:
```yml
test_pass_capitalisation_policy_data_type_keywords:
  # Test cases where data types are keywords and should therefore not be touched by L010
  # See: https://github.com/sqlfluff/sqlfluff/pull/2121
  fail_str: CREATE TABLE table1 (account_id bigint, column_two serial);
  configs:
    rules:
      L010:
        capitalisation_policy: upper
```

tdstark: I tried this and it fails as well.

jpy-git: yeah but you should still add it to the tests because:
a) It's convenient for the reviewers
b) How can we know that we've fixed L010s impact on data_type keywords if you don't add it?
c) How can we be sure that someone in the future will make a change that causes L010 to start messing with the data type keywords again after we've fixed them?

tdstark: @jpy-git Added!

tdstark: @jpy-git shouldn't the L010 test be this? 
```
test_pass_capitalisation_policy_data_type_keywords:
  # Test cases where data types are keywords but are handled by L059 instead of L010
  # See: https://github.com/sqlfluff/sqlfluff/pull/2121
  fail_str: create table table1 (account_id bigint, column_two serial);
  fix_str: CREATE TABLE table1 (account_id bigint, column_two serial);
  configs:
    rules:
      L010:
        capitalisation_policy: upper
```
My thought is that L010 would only be affecting the `create table` statement. Since L059 inherits L010, I'm assuming that the logic in L059 wouldn't be picked up by L010, correct?

src/sqlfluff/rules/L059.py
```diff
@@ -0,0 +1,47 @@
+"""Implementation of Rule L030."""
```
jpy-git, line 1:
```suggestion
"""Implementation of Rule L059."""
```

test/fixtures/rules/std_rule_cases/L010.yml
```diff
@@ -62,3 +62,13 @@ test_fail_date_part_capitalisation_policy_upper:
 test_pass_date_part_consistent_capitalisation:
   # Test that correctly capitalized time units are left unchanged
   pass_str: SELECT dt + INTERVAL 2 DAY, INTERVAL 3 HOUR
+
+test_pass_capitalisation_policy_data_type_keywords:
+  # Test cases where data types are keywords but are handled by L059 instead of L010
+  # See: https://github.com/sqlfluff/sqlfluff/pull/2121
+  fail_str: create table table1 (account_id bigint, column_two serial);
+  fix_str: CREATE TABLE table1 (account_id bigint, column_two serial);
```
jpy-git, line 70:
```suggestion
  pass_str: CREATE TABLE table1 (account_id bigint, column_two serial);
```

While both versions of this test show that L010 is not affecting serial, it is much clearer to ask the question "has anything changed?" vs "are some of the things that have changed not correct?"

test/fixtures/rules/std_rule_cases/L059.yml
```diff
@@ -0,0 +1,77 @@
...
+
+test_fail_capitalisation_policy_keywords:
+  # Test cases where data types are keywords, not data_type_identifiers
+  # See: https://github.com/sqlfluff/sqlfluff/pull/2121
+  fail_str: CREATE TABLE table1 (account_id BIGINT, column_two serial);
+  fix_str: CREATE TABLE table1 (account_id BIGINT, column_two SERIAL);
+  configs:
+    rules:
+      L059:
+        capitalisation_policy: upper
```
jpy-git, line 77: Interesting that serial is now behaving itself. What changed?

test/fixtures/rules/std_rule_cases/L059.yml
jpy-git: added an explanation in my review

test/fixtures/rules/std_rule_cases/L059.yml
tdstark: @jpy-git that was it!

jpy-git (approved):
Awesome work 😄 Will merge in once the tests have completed 🚀 ✨

tunetheweb:
Nice work both! Was going to look at the SERIAL thing but never got a chance. Glad you figured it out!

kevinmarsh:
Nice, thanks for this fix I was just looking for this functionality :+1:  I just tried this PR and it doesn't appear to be working for the `timestamp`/`timestamp with time zone` data type, eg.
```yaml
test_fail_timestamp_upper:
  fail_str: CREATE TABLE table1 (created_at timestamp);
  fix_str: CREATE TABLE table1 (created_at TIMESTAMP);
  configs:
    rules:
      L059:
        capitalisation_policy: upper
```

tdstark:
@tunetheweb @jpy-git Is there an elegant way to handle this?

jpy-git:
Ah makes sense, This is exactly the kind of thing I was concerned about in this comment: https://sqlfluff.slack.com/archives/C01GPHR8J0G/p1641244461224200?thread_ts=1641238813.220900&cid=C01GPHR8J0G

Looking deeper the SERIAL case we added to L010 was misleading because it's not actually an example of a Keyword, just another data type identifier so we haven't solved the keyword problem:
image

So what we need is another PR where we replace SERIAL with TIMESTAMP in the L010 example we added here so we can first confirm whether the fact that its a keyword confuses L010 and then asses how to fix L010 and extend L059.

Could one of you open that as a draft PR and we can investigate?

jpy-git:
> EDIT: think this is most likely due to [this change](https://github.com/sqlfluff/sqlfluff/pull/1338/files#diff-d8c55713457ca11da9c9dd7e107baf461825c91ad12fe4b636d630aab4b5ee12) where `TIMESTAMP` changes from a `data_type_identifier` to a `keyword`...

^ Good spot 😄 

For reference, it's not a mistake that these are keywords and actually that's correct, since `data_type_identifer` is really just meant to match user defined custom types and the rest are keywords. The reason why lots of data types match `data_type_identifer` is just because a lot of the data type grammar is yet to be implemented so allowing everything to be matched as `data_type_identifer` is convenient to get things parsing atm. So we need L059 to account for this case as it'll become more common in the future 👍

jpy-git:
I wonder if we should consider splitting out the data type keywords like we do with `date_part` keywords ([source](https://github.com/sqlfluff/sqlfluff/blob/ed3451c3c3080f7357e404c4b1bc404ffcd0a945/src/sqlfluff/dialects/dialect_ansi.py#L306-L313)) so that the parse tree can look more like:
```yml
data_type:
    data_part: TIMESTAMP
```
That way it wouldn't get matched as a keyword by L010 and we can just add `data_part` to L059. Plus nice to have these things defined in the grammar.

Not sure of the other implications of doing that though so something to play with.

@barrywhart any opinions on getting these keyword data types to not be matched by L010 and included in L059?

tunetheweb:
Alternatively should we just remove L059 and expand L010 to include date parts AND keywords? Either because they are keywords or date parts? Not sure why this needs to be two rules - would you ever want one but not that other?

jpy-git:
Personally I'm happy with that as well since I always capitalise these consistently with other keywords (I assume you mean data type keywords not `date_part`). That's basically the conclusion we came to with `date_part` in https://github.com/sqlfluff/sqlfluff/pull/2167 so if people are happy to take that approach then it would be just be a matter of adding `data_type_identifier` to L010 and removing this rule.

PR #2239, opened 2022-01-05
Fix L044 assertion failure: "SELECT *" with no "FROM" clause

barrywhart:

This fixes an assertion failure that appeared in a user's failure log in a [Slack chat](https://sqlfluff.slack.com/archives/C01GX4L213J/p1641331471426900?thread_ts=1641329507.421800&cid=C01GX4L213J) yesterday.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L044.py
+++ b/src/sqlfluff/rules/L044.py
@@ -120,7 +120,7 @@ class Rule_L044(BaseRule):
                         if isinstance(o, Query):
                             self._analyze_result_columns(o)
                             return
-                    assert False, "Should be unreachable"  # pragma: no cover
+                    raise RuleFailure(query.selectables[0].selectable)
 
     def _eval(self, context: RuleContext) -> Optional[LintResult]:
         """Outermost query should produce known number of columns."""
--- a/src/sqlfluff/rules/L044.py
+++ b/src/sqlfluff/rules/L044.py
@@ -120,6 +120,9 @@ class Rule_L044(BaseRule):
                         if isinstance(o, Query):
                             self._analyze_result_columns(o)
                             return
+                    self.logger.debug(
+                        f'Query target "{query.selectables[0].selectable.raw}" has no targets. Generating warning.'
+                    )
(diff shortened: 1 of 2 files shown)


src/sqlfluff/rules/L044.py
```diff
@@ -120,7 +120,7 @@ def _analyze_result_columns(self, query: Query):
                         if isinstance(o, Query):
                             self._analyze_result_columns(o)
                             return
-                    assert False, "Should be unreachable"  # pragma: no cover
+                    raise RuleFailure(query.selectables[0].selectable)
```
jpy-git, line 123: Do we want an accompanying `self.logger.debug(` message here?

What does `raise RuleFailure(query.selectables[0].selectable)` look like to us if we were trying to debug this?

jpy-git (commented):
@barrywhart small Q, otherwise makes sense 😄

src/sqlfluff/rules/L044.py
barrywhart: This is something specific to this rule -- this is how it reports lint issues. The rule itself catches this and returns a regular LintResult. (You'll see a very similar line 10 lines up in this file.) I designed it this way because this rule recourses deeply into the parse structure, and it seemed cleaner to raise an exception rather than returning the result step by step up the call stack the usual way.

barrywhart: For this reason, I don't think I log message is needed.

jpy-git: makes sense, the example 10 lines up has the log message before which was why I suggested it 😛

Happy to approve if you're happy it's sufficient 👍

jpy-git (approved):

src/sqlfluff/rules/L044.py
barrywhart: Ah, good point. I think some days I feel more "loggy" than others. I'll go ahead and add that, just a sec.

barrywhart: Log message added. 👍

Issue #2235, opened 2022-01-05
Syntax error when using DBT Jinja variable with Snowflake + DBT


Ben-A-Canva:
When linting a file containing the following section of code, the sqlfluff slicer includes the `alternate_code` segment which breaks the dbt compile run.

```
{%- set get_source_maximum_query %}
  select
    max(_partition_key) as max_partition_key
  from snowflake_db.{{ target.name }}__raw._test_table
{%- endset %}
```
The generated SQL is:
```
  select
    max(_partition_key) as max_partition_key
  from snowflake_db.659cfb6f9b26429aa2cf83619d3bf5f9 dev\x00__raw._test_table
```
Resulting in ```SQL compilation error: syntax error line 3 at position 21 unexpected '659'.``` In Snowflake's query history.

From debugging, i've found that the templater slicer is adding in the `alternate_code` which results in the above query.

## Expected Behaviour
The generated sql should be 
```
  select
    max(_partition_key) as max_partition_key
  from snowflake_db.dev__raw._test_table
```

## Dialect
Snowflake

## Version
sqlfluff, version 0.9.0
dbt version 1.0.0

## Configuration
```
[sqlfluff]
verbose = 2
nocolor = False
dialect = snowflake
templater = dbt
exclude_rules = L031,L034,L022,L051 
#L025,
recurse = 10
output_line_length = 500
runaway_limit = 10
ignore_templated_areas = True
encoding = 'utf-8'
format = 'human'
# Comma separated list of file extensions to lint.

# NB: This config will only apply in the root folder.
sql_file_exts = .sql,.sql.j2,.dml,.ddl

[sqlfluff:templater:dbt]
project_dir = src/app/transform
profiles_dir = src/app
profile = app

[sqlfluff:indentation]
indented_joins = False
template_blocks_indent = True

[sqlfluff:templater]
unwrap_wrapped_queries = False

[sqlfluff:templater:jinja:macros]
#Macros provided as builtins for dbt projects
dbt_ref = {% macro ref(model_ref) %}{{model_ref}}{% endmacro %}
dbt_source = {% macro source(source_name, table) %}{{source_name}}_{{table}}{% endmacro %}
dbt_config = {% macro config() %}{% for k in kwargs %}{% endfor %}{% endmacro %}
dbt_var = {% macro var(variable) %}item{% endmacro %}
dbt_is_incremental = {% macro is_incremental() %}False{% endmacro %}

[...]

PR #2240, opened 2022-01-05
Fix a bug handling Jinja "{% set %}" blocks with a templated block inside

barrywhart:

Makes progress on #2235 

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/templaters/slicers/tracer.py
+++ b/src/sqlfluff/core/templaters/slicers/tracer.py
@@ -5,7 +5,6 @@ This is a newer slicing algorithm that handles cases heuristic.py does not.
 
 import logging
 import regex
-import uuid
 from itertools import chain
 from typing import Callable, cast, Dict, List, NamedTuple, Optional
 
@@ -20,18 +19,6 @@ from sqlfluff.core.templaters.base import (
 
 # Instantiate the templater logger
 templater_logger = logging.getLogger("sqlfluff.templater")
-
-
-uuid_idx = 0
-
-
-def uuid4() -> str:
-    """Returns a unique ID (really just a counter for now)."""
-    global uuid_idx
-    result = uuid.uuid4().hex  # NOTE: We're discarding this value.
-    result = "{0:#0{1}x}".format(uuid_idx, 34)[2:]
-    uuid_idx += 1
-    return result
 
 
 class JinjaTrace(NamedTuple):
@@ -66,6 +53,7 @@ class JinjaTracer:
         self.env = env
         self.make_template: Callable[[str], Template] = make_template
         self.program_counter: int = 0
+        self.slice_id: int = 0
         self.raw_slice_info: Dict[RawFileSlice, RawSliceInfo] = {}
         self.raw_sliced: List[RawFileSlice] = self._slice_template()
         self.sliced_file: List[TemplatedFileSlice] = []
@@ -194,6 +182,12 @@ class JinjaTracer:
         if slice_type in ("literal", "templated"):
             self.source_idx += target_slice_length
 
+    def next_slice_id(self) -> str:
+        """Returns a unique ID (really just a counter for now)."""
+        result = "{0:#0{1}x}".format(self.slice_id, 34)[2:]
+        self.slice_id += 1
+        return result
+
     def _slice_template(self) -> List[RawFileSlice]:
         """Slice template in jinja.
 
@@ -227,10 +221,10 @@ class JinjaTracer:
             # Replace literal text with a unique ID.
             if elem_type == "data":
                 if set_idx is None:
-                    unique_alternate_id = uuid4()
+                    unique_alternate_id = self.next_slice_id()
                     alternate_code = f"\0{unique_alternate_id}_{len(raw)}"
                 else:
-                    unique_alternate_id = uuid4()
+                    unique_alternate_id = self.next_slice_id()
                     alternate_code = f"\0set{unique_alternate_id}_{len(raw)}"
                 result.append(
                     RawFileSlice(
@@ -319,7 +313,7 @@ class JinjaTracer:
                         # effects, but return a UUID.
                         if trimmed_content:
                             assert m_open and m_close
-                            unique_id = uuid4()
+                            unique_id = self.next_slice_id()
                             unique_alternate_id = unique_id
                             prefix = "set" if set_idx is not None else ""
                             open_ = m_open.group(1)
--- a/src/sqlfluff/core/templaters/slicers/tracer.py
+++ b/src/sqlfluff/core/templaters/slicers/tracer.py
@@ -83,8 +83,6 @@ class JinjaTracer:
                 if pos2 != -1
                 else trace_template_output[pos1 + 1 :]
             )
-            if not p:
-                continue
             is_set = p[:3] == "set"
             if is_set:
                 p = p[3:]
@@ -183,7 +181,7 @@ class JinjaTracer:
             self.source_idx += target_slice_length
 
     def next_slice_id(self) -> str:
-        """Returns a unique ID (really just a counter for now)."""
+        """Returns a new, unique slice ID."""
         result = "{0:#0{1}x}".format(self.slice_id, 34)[2:]
         self.slice_id += 1
         return result
--- a/src/sqlfluff/core/templaters/slicers/tracer.py
+++ b/src/sqlfluff/core/templaters/slicers/tracer.py
@@ -92,8 +92,8 @@ class JinjaTracer:
                 # in raw_str.
                 value = [m_id.group(1), int(m_id.group(3)), True]
             else:
-                # E.g. "adc15d2a41d14ead97411bce3fb55e32 a < 10". The characters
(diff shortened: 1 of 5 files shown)


barrywhart:
In the current state of this PR, `sqlfluff parse` on this file throws an exception:
```
{% set my_query %}
  select 1 from foobarfoobarfoobarfoobar_{{ "dev" }}
{% endset %}
{{ my_query }}
```

test/core/templaters/jinja_test.py
```diff
@@ -507,6 +507,23 @@ def test__templater_full(subpath, code_only, include_meta, yaml_loader, caplog):
                 (" BAR", "literal", 30),
             ],
         ),
+        (
+            """{% set my_query %}
+select 1 from foobarfoobarfoobarfoobar_{{ "dev" }}
```
barrywhart, line 512: Note the `{{ }}` block on this line inside the `{% set %}` block. This was previously not handled correctly and caused a runtime error.

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -70,9 +70,17 @@ def trace(self) -> JinjaTrace:
         trace_template = self.make_template(trace_template_str)
         trace_template_output = trace_template.render()
         # Split output by section. Each section has two possible formats.
-        for p in trace_template_output.split("\0"):
```
barrywhart, line 73: Previously, each output entry was terminated by a `\0` character. In order to handle the new case, it proved better to **prefix** each output entry with a `\0` instead, hence I refactored the loop.

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -157,6 +170,12 @@ def record_trace(self, target_slice_length):
         if slice_type in ("literal", "templated"):
             self.source_idx += target_slice_length
 
+    def next_slice_id(self) -> str:
```
barrywhart, line 173: This function replaces uses of `uuid.uuid4().hex`. There was really no need to use UUIDs, and changing to a deterministic counter makes the tests easier to understand and debug.

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -187,16 +206,14 @@ def _slice_template(self) -> List[RawFileSlice]:
         unique_alternate_id: Optional[str]
         alternate_code: Optional[str]
         for _, elem_type, raw in self.env.lex(self.raw_str):
-            # Replace literal text with a unique ID, except for "set"
-            # statements, which don't emit output and thus don't need this
-            # treatment.
+            # Replace literal text with a unique ID.
             if elem_type == "data":
                 if set_idx is None:
-                    unique_alternate_id = uuid.uuid4().hex
```
barrywhart, line 195: Here's where I moved the `\0` from the end of each chunk to the beginning.

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -139,16 +150,18 @@ def move_to_slice(self, target_slice_idx, target_slice_length):
                 candidates.sort(key=lambda c: abs(target_slice_idx - c))
                 self.program_counter = candidates[0]
 
-    def record_trace(self, target_slice_length):
```
barrywhart, line 142: I tweaked this function to optionally take a `slice_idx` so I could reuse it to process output from a `{% set %}` trace. Previously it always used `self.program_counter`.

test/core/templaters/jinja_test.py
tunetheweb: Might be good to link to the docs on this, as I didn't realise this was valid syntax, but looks like it is: https://jinja.palletsprojects.com/en/3.0.x/templates/#block-assignments

tunetheweb: Ah ignore me. See you added that comment in the code itself on line 314.

barrywhart: Done!

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -70,9 +70,17 @@ def trace(self) -> JinjaTrace:
...
-                continue
+        trace_entries = list(regex.finditer(r"\0", trace_template_output))
+        for match_idx, match in enumerate(trace_entries):
+            pos1 = match.span()[0]
+            try:
+                pos2 = trace_entries[match_idx + 1].span()[0]
+            except IndexError:
+                pos2 = len(trace_template_output)
+            p = trace_template_output[pos1 + 1 : pos2]
+            is_set = p[:3] == "set"
```
tunetheweb, line 81: This only covers the `{% set... %}` case but looking at https://jinja.palletsprojects.com/en/3.0.x/templates/ there are lots of similar cases (`for`, `call`, `macro`). Should we include them in this?

Or at least add a comment saying they are not supported for now?

Or where they already covered (I'd be really surprised if `for` was not covered already)?

test/core/templaters/jinja_test.py
barrywhart: No worries, it's often a good idea to document both the tests and the code, to help connect the dots.

src/sqlfluff/core/templaters/slicers/tracer.py
barrywhart: `{% for %}` is covered by the general block handling -- see the `block_types` lookup elsewhere in the file. Jinja is helping us with some of this by telling us the element type. TBH, I don't know if `call` and `macro` work. SInce I don't know for sure, I would rather not add comments about those.

barrywhart: Note that `tracer.py` is the primary (only) component used in the past few releases for Jinja and dbt template slicing, so it's fairly robust. Given the extreme amount of templating used in dbt projects, I think it's pretty robust, just can't say for sure. AFAIK, this issue was the first one raised since we switched over in late October/early November.

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -284,24 +301,28 @@ def _slice_template(self) -> List[RawFileSlice]:
                         # effects, but return a UUID.
```
tunetheweb, line 301: Can we update this comment since not using UUID's anymore?

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -89,10 +97,13 @@ def trace(self) -> JinjaTrace:
                 value = [m_id.group(0), p[len(m_id.group(0)) + 1 :], False]
```
tunetheweb, line 97: Just above this are two more UUID references in comments (I think). Should they be updated.

Sorry can't comment on unchanged lines.

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -89,10 +97,13 @@ def trace(self) -> JinjaTrace:
...
+            if not is_set:
+                self.move_to_slice(target_slice_idx, slice_length)
             else:
-                self.move_to_slice(target_slice_idx, len(str(content_info)))
+                # If we find output from a {% set %} directive, record a trace
+                # without reading or updating the program counter.
+                self.record_trace(slice_length, target_slice_idx)
         return JinjaTrace(
             self.make_template(self.raw_str).render(), self.raw_sliced, self.sliced_file
         )
```
tunetheweb, line 109: Just below this is another UUID reference.

tunetheweb (changes requested):
Can't say I honestly "get" all the code but happy with the test cases.

There's a few UUID references to clean up but after that should be good to merge.

barrywhart:
Ok, I believe I've updated/removed all the references to UUIDs. Thanks for the review!

tunetheweb (approved):

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -70,9 +70,17 @@ def trace(self) -> JinjaTrace:
         trace_template = self.make_template(trace_template_str)
         trace_template_output = trace_template.render()
         # Split output by section. Each section has two possible formats.
-        for p in trace_template_output.split("\0"):
-            if not p:
-                continue
+        trace_entries = list(regex.finditer(r"\0", trace_template_output))
```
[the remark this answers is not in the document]
barrywhart: Hmm, I tried this, and it wasn't working. Perhaps the objects in the list are different types? I think `finditer` returns match objects, while `findall` returns tuples of strings.

```
        trace_entries = regex.findall(r"\0", trace_template_output)
        for match_idx, match in enumerate(trace_entries):
>           pos1 = match.span()[0]
E           AttributeError: 'str' object has no attribute 'span'
```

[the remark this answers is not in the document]
jpy-git: ahh sorry you're right findall returns `list[AnyStr] | list[tuple[AnyStr, ...]]` so ignore me 😄

PR #2242, opened 2022-01-06
Docs: Make code blocks consistent

jpy-git:

In a lot of places in the docs we're marking code blocks as bash even though they contain command output as well as the console commands, resulting in weird highlighting:
image

image

This PR converts these blocks to text to avoid this.

Additionally I've made `cfg` blocks consistent and specified the blocks missing a marker.

No

### Pull Request checklist

  - Other.


docs/source/configuration.rst
```diff
@@ -522,7 +522,7 @@ You already know you can pass arguments (:code:`--verbose`,
 :code:`--exclude-rules`, etc.) through the CLI commands (:code:`lint`,
 :code:`fix`, etc.):
 
-.. code-block:: console
+.. code-block:: text
```
tunetheweb, line 525: Does this one make any difference?

tunetheweb (approved):

barrywhart (approved):

docs/source/configuration.rst
jpy-git: kind of becuase a `$` at the start isn't valid shell code and also it's consistent with every other code block so think we should leave as `text`?

tunetheweb: Fair enough.

Issue #2152, opened 2021-12-21
VALUES clause fails to parse in spark3


mcannamela:
## Expected Behaviour
`VALUES` clauses should parse, and correctly identify the subsequent literals. 

## Observed Behaviour
Parsing succeeds with one record and no alias, but treats it as a function call: `...VALUES (...);`
Parsing fails when using an alias: `...VALUES (...) as (...);`
Parsing fails when using multiple records: `...VALUES (...), (...)...(...);`

## Steps to Reproduce
See this example fail CI [here](https://github.com/mcannamela/sqlfluff/actions/runs/1607132440).

`sqlfluff parse` the following:
```
select a, b from values (1, 2);
select a, b from values (1, 2) as (a,b);
select a, b from values (1, 2), (3, 4);
```

result:
```
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    base:
[L:  1, P:  1]      |        select_statement:
[L:  1, P:  1]      |            select_clause:
[L:  1, P:  1]      |                keyword:                                      'select'
[L:  1, P:  7]      |                [META] indent:
[L:  1, P:  7]      |                whitespace:                                   ' '
[L:  1, P:  8]      |                select_clause_element:
[L:  1, P:  8]      |                    column_reference:
[L:  1, P:  8]      |                        identifier:                           'a'
[L:  1, P:  9]      |                comma:                                        ','
[L:  1, P: 10]      |                whitespace:                                   ' '
[L:  1, P: 11]      |                select_clause_element:
[L:  1, P: 11]      |                    column_reference:
[L:  1, P: 11]      |                        identifier:                           'b'
[L:  1, P: 12]      |            whitespace:                                       ' '
[L:  1, P: 13]      |            [META] dedent:
[L:  1, P: 13]      |            from_clause:
[L:  1, P: 13]      |                keyword:                                      'from'
[L:  1, P: 17]      |                whitespace:                                   ' '
[L:  1, P: 18]      |                from_expression:
[L:  1, P: 18]      |                    [META] indent:
[L:  1, P: 18]      |                    from_expression_element:
[L:  1, P: 18]      |                        table_expression:
[L:  1, P: 18]      |                            function:
[L:  1, P: 18]      |                                function_name:
[L:  1, P: 18]      |                                    function_name_identifier:  'values'
[L:  1, P: 24]      |                                whitespace:                   ' '
[L:  1, P: 25]      |                                bracketed:
[...]

PR #2245, opened 2022-01-06
spark3: parse some VALUES clauses

mcannamela:
Makes progress on #2152. 

`VALUES` clauses fail to parse in the spark3 dialect I believe because `TableExpressionSegment` does not include `ValuesClauseSegment`. I address that only for the spark3 dialect by including it, and also redefining some segments to address subsequent parsing failures. 

At present, I believe I am covering most simple cases of `VALUES` used in `FROM` clauses, but not any of the optional directives like`ORDER BY`, `LIMIT`, `OFFSET` that you can see in e.g. the [postgres docs](https://www.postgresql.org/docs/13/sql-values.html) (Spark SQL docs don't mention `VALUES` except in `INSERT` statements). I suspect this should all apply to the ansi dialect but I must admit I gave up trying to find the spec after not that much looking. 

Also, it only works reliably when the `VALUES` clause is in parentheses:
```
    select * from (values 1, 2) --works 
    select * from values 1, 2 --fails
```

I think this is because the `parse_grammar` of `FromClauseSegment` is `Delimited(Ref("FromExpressionSegment"))`, which splits on the first comma it finds rather than looking for the longest `FromExpressionSegment`. I don't see a good way around that. I could live with this, but I wish I had a way to hint to the user that this is what went wrong, since the output of `parse` does not make this at all obvious that just wrapping with parens will fix it. 

The only dialect tests that fail for spark3 are ones that I added, but since I do override `TableExpressionSegment` for spark3 I imagine there are tests we were relying on in ansi that we would want to make sure spark3 passes. Before going nuts with that, I wanted to see whether I shouldn't just be pushing all this down into the ansi dialect anyway. 

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -783,3 +783,117 @@ class JoinClauseSegment(BaseSegment):
...
     ).get_eventual_alias
+
+
+@spark3_dialect.segment(replace=True)
+class TableExpressionSegment(BaseSegment):
+    """The main table expression e.g. within a FROM clause."""
+
+    type = "table_expression"
+    match_grammar = OneOf(
+        Ref("ValuesClauseSegment"),
```
mcannamela, line 794: This had to come before `FunctionSegment`, or else it would try to interpret `values (1,2),(3,4)` as a function call

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -783,3 +783,117 @@ class JoinClauseSegment(BaseSegment):
...
+        SELECT * FROM VALUES (1,2) as (a, b);
+
+    """
+
+    type = "alias_expression"
+    match_grammar = Sequence(
+        Ref.keyword("AS", optional=True),
+        OneOf(
+            Sequence(
+                Ref("SingleIdentifierGrammar", optional=True),
```
mcannamela, line 822: the table alias is optional if you just want to alias columns, not sure if that's valid in ansi as well

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -783,3 +783,117 @@ class JoinClauseSegment(BaseSegment):
...
+                # Column alias in VALUES clause
+                Bracketed(Ref("SingleIdentifierListSegment"), optional=True),
+            ),
+            Ref("QuotedLiteralSegment"),
+        ),
+    )
+
+
+@spark3_dialect.segment()
+class DelimitedValues(BaseSegment):
```
mcannamela, line 832: I started with `DelimitedValues`, `ScalarValue`, and `TupleValue` inlined into `ValuesClauseSegment`, but factored them out for my sanity while debugging and don't see a particular reason to go back to the in-lined version.

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -783,3 +783,117 @@ class JoinClauseSegment(BaseSegment):
...
+        SELECT * FROM VALUES (1,2) as t (a,b);
+        SELECT * FROM (VALUES (1,2) as t (a,b));
+        WITH a AS (VALUES 1,2) SELECT * FROM a;
+
+    """
+
+    type = "values_clause"
+    match_grammar = StartsWith(Ref.keyword("VALUES"))
+
+    parse_grammar = Sequence(
```
mcannamela, line 895: I have to confess I do not understand the difference between `match_grammar` and `parse_grammar`, so this is just parroting a similar segment I saw elsewhere.

test/conftest.py
```diff
@@ -143,15 +144,15 @@ def yaml_loader():
...
+def generate_test_segments()->Callable[[List[str]], List[RawSegment]]:
     """Roughly generate test segments.
 
     This is a factory function so that it works as a fixture,
     but when actually used, this will return the inner function
     which is what you actually need.
     """
 
-    def generate_test_segments_func(elems):
+    def generate_test_segments_func(elems:List[str])->List[RawSegment]:
```
mcannamela, line 155: i ended up not using this fixture but i did want to ask, what is the point of this wrapper and using it as a fixture, rather than just calling the inner function? there's no state to deal with and nothing in the closure so it seems unnecessary

test/dialects/spark3_grammar_test.py
```diff
@@ -0,0 +1,97 @@
+"""Test some specific spark3 grammars."""
```
mcannamela, line 1: this was mainly for my learning/troubleshooting

test/dialects/spark3_grammar_test.py
```diff
@@ -0,0 +1,97 @@
...
+    [
+        (t[1], b, t[0])
+        for t in [
+            ("as t ( col1 , col2 )", "AliasExpressionSegment"),
+            ("as t \t ( col1 , col2 )", "AliasExpressionSegment"),
+            ("from t", "FromClauseSegment"),
+            ("from t as u", "FromClauseSegment"),
+            ("from values ( 1 , 2 )", "FromClauseSegment"),
+            ("from values 1", "FromClauseSegment"),
+            ("from values 1 , 2 ", "FromClauseSegment"),
```
mcannamela, line 41: this example is what ultimately allowed me to realize what why it wouldn't parse; i was getting very confused because `values 1,2` was correctly parsing as a `ValuesClauseSegment` and a `FromExpressionSegment`

test/dialects/spark3_grammar_test.py
```diff
@@ -0,0 +1,97 @@
...
+    with RootParseContext(dialect=fresh_spark3_dialect) as ctx:
+        ctx.logger.setLevel(logging.DEBUG)
+        with caplog.at_level(logging.DEBUG):
+            p = grammar.match(lexed_seg_list, parse_context=ctx)
+
+    if p.unmatched_segments:
+        logger = logging.getLogger()
+        logger.addHandler(logging.StreamHandler())
+        for _, _, message in caplog.record_tuples:
+            logger.warning(message)
```
mcannamela, line 95: for my debugging purposes. 

i wanted to ask, is there a good place that explains how to read the detailed parsing output?

just by staring at it i could see it was only picking up part of `from values 1,2` as the `FromExpressionSegment`, but beyond that i was having trouble reading it:

```
[PD:0  MD:1 ]	.None                                             	Ref.match OUT       	+   	[match=<MatchResult 1/9: 'from'>, seg="'from values 1 , 2'"]
[PD:0  MD:2 ]	..None                                            	Delim._look_ahead_match IN	    	[ls=7, seg='values 1 , 2']
[PD:0  MD:2 ]	..None                                            	Delim._look_ahead_match SI	    	[mq=[], sb=['VALUES', '', '1', '', ',', '', '2']]
[PD:0  MD:2 ]	..None                                            	Ref.match OUT       	+   	[match=<MatchResult 1/3: ','>, seg="', 2'"]
[PD:0  MD:2 ]	..None                                            	Delim._look_ahead_match SC	    	[bsm=(4, 1, <Ref: CommaSegment>)]
[PD:0  MD:4 ]	....FromExpressionSegment                         	OneOf.match PRN     	    	[ns=1, ps=1, ms=0, pruned=[<Ref: MLTableExpressionSegment>], opts=[<Ref: FromExpressionElementSegment>]]
[PD:0  MD:7 ]	.......FromExpressionElementSegment               	Ref.match OUT       	    	[match=<MatchResult 0/3: ''>, seg="'values 1'"]
[PD:0  MD:7 ]	.......FromExpressionElementSegment               	Optio.match PRN     	    	[ns=1, ps=1, ms=0, pruned=[<Bracketed: [<Ref: TableExpressionSegment>]>], opts=[<Ref: TableExpressionSegment>]]
[PD:0  MD:9 ]	.........TableExpressionSegment                   	OneOf.match PRN     	    	[ns=3, ps=1, ms=1, pruned=[<Bracketed: [<Ref: SelectableGrammar>]>], opts=[<Ref: ValuesClauseSegment>, <Ref: BareFunctionSegment>, <Ref: FunctionSegment>, <Ref: TableReferenceSegment>]]
[PD:0  MD:12]	............ValuesClauseSegment                   	Ref.match OUT       	+   	[match=<MatchResult 1/3: 'values'>, seg="'values 1'"]
[PD:0  MD:11]	...........ValuesClauseSegment                    	Start.match OUT     	++  	[match=<MatchResult 3/3: 'values 1'>, seg="'values 1'"]
[PD:0  MD:10]	..........ValuesClauseSegment                     	Value.match OUT     	++  	[match=<MatchResult 1/1: 'values 1'>, seg="'values 1'"]
[PD:0  MD:10]	..........TableExpressionSegment                  	Ref.match OUT       	++  	[match=<MatchResult 1/1: 'values 1'>, seg="'values 1'"]
[PD:0  MD:9 ]	.........TableExpressionSegment                   	OneOf.match OUT     	++  	[match=<MatchResult 1/1: 'values 1'>, seg="'values 1'"]
[PD:0  MD:8 ]	........TableExpressionSegment                    	Table.match OUT     	++  	[match=<MatchResult 1/1: 'values 1'>, seg="'values 1'"]
[PD:0  MD:8 ]	........FromExpressionElementSegment              	Ref.match OUT       	++  	[match=<MatchResult 1/1: 'values 1'>, seg="'values 1'"]
[PD:0  MD:7 ]	.......FromExpressionElementSegment               	Optio.match OUT     	++  	[match=<MatchResult 1/1: 'values 1'>, seg="'values 1'"]
[PD:0  MD:6 ]	......FromExpressionElementSegment                	Seque.match OUT     	++  	[match=<MatchResult 1/1: 'values 1'>, seg="'values 1'"]
[PD:0  MD:5 ]	.....FromExpressionElementSegment                 	FromE.match OUT     	++  	[match=<MatchResult 1/1: 'values 1'>, seg="'values 1'"]
[PD:0  MD:5 ]	.....FromExpressionSegment                        	Ref.match OUT       	++  	[match=<MatchResult 1/1: 'values 1'>, seg="'values 1'"]
[PD:0  MD:4 ]	....FromExpressionSegment                         	OneOf.match OUT     	++  	[match=<MatchResult 1/1: 'values 1'>, seg="'values 1'"]
[PD:0  MD:3 ]	...FromExpressionSegment                          	Condi.match OUT     	++  	[match=<MatchResult 1/1: ''>, seg="''"]
[PD:0  MD:3 ]	...FromExpressionSegment                          	Seque.match OUT     	++  	[match=<MatchResult 3/3: 'values 1'>, seg="'values 1'"]
[PD:0  MD:2 ]	..FromExpressionSegment                           	FromE.match OUT     	++  	[match=<MatchResult 1/1: 'values 1'>, seg="'values 1'"]
[PD:0  MD:2 ]	..None                                            	Ref.match OUT       	++  	[match=<MatchResult 1/1: 'values 1'>, seg="'values 1'"]
[PD:0  MD:2 ]	..None                                            	Delim._look_ahead_match IN	    	[ls=2, seg=' 2']
[PD:0  MD:2 ]	..None                                            	Delim._look_ahead_match SI	    	[mq=[], sb=['', '2']]
[PD:0  MD:2 ]	..None                                            	Delim._look_ahead_match SC	    	[bsm=None]
[PD:0  MD:4 ]	....FromExpressionSegment                         	OneOf.match PRN     	    	[ns=1, ps=1, ms=0, pruned=[<Ref: MLTableExpressionSegment>], opts=[<Ref: FromExpressionElementSegment>]]
[PD:0  MD:7 ]	.......FromExpressionElementSegment               	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:7 ]	.......FromExpressionElementSegment               	Optio.match PRN     	    	[ns=1, ps=1, ms=0, pruned=[<Bracketed: [<Ref: TableExpressionSegment>]>], opts=[<Ref: TableExpressionSegment>]]
[PD:0  MD:9 ]	.........TableExpressionSegment                   	OneOf.match PRN     	    	[ns=3, ps=2, ms=0, pruned=[<Ref: ValuesClauseSegment>, <Bracketed: [<Ref: SelectableGrammar>]>], opts=[<Ref: BareFunctionSegment>, <Ref: FunctionSegment>, <Ref: TableReferenceSegment>]]
[PD:0  MD:10]	..........TableExpressionSegment                  	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:11]	...........FunctionSegment                        	OneOf.match PRN     	    	[ns=1, ps=1, ms=0, pruned=[<Sequence: [<Sequence: [<Ref: DatePartFunctionNameSe...]>], opts=[<Sequence: [<Sequence: [<AnyNumberOf: [<Ref: Functio..., <Ref: PostFunctionGrammar [opt]>]>]]
[PD:0  MD:17]	.................DatePartFunctionNameSegment      	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:16]	................DatePartFunctionNameSegment       	Seque.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:15]	...............DatePartFunctionNameSegment        	DateP.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:15]	...............FunctionSegment                    	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:14]	..............FunctionSegment                     	AnyNu.match PRN     	    	[ns=1, ps=0, ms=0, pruned=[], opts=[<Ref: FunctionNameSegment>]]
[PD:0  MD:17]	.................FunctionNameSegment              	AnyNu.match PRN     	    	[ns=1, ps=0, ms=0, pruned=[], opts=[<Sequence: [<Ref: SingleIdentifierGrammar>, <Ref: DotSegment>]>]]
[PD:0  MD:19]	...................SingleIdentifierGrammar        	OneOf.match PRN     	    	[ns=2, ps=0, ms=0, pruned=[], opts=[<Ref: NakedIdentifierSegment>, <Ref: QuotedIdentifierSegment>]]
[PD:0  MD:20]	....................SingleIdentifierGrammar       	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:20]	....................SingleIdentifierGrammar       	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:19]	...................SingleIdentifierGrammar        	OneOf.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:19]	...................FunctionNameSegment            	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:18]	..................FunctionNameSegment             	Seque.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:17]	.................FunctionNameSegment              	AnyNu.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:17]	.................FunctionNameSegment              	OneOf.match PRN     	    	[ns=2, ps=0, ms=0, pruned=[], opts=[<Ref: FunctionNameIdentifierSegment>, <Ref: QuotedIdentifierSegment>]]
[PD:0  MD:18]	..................FunctionNameSegment             	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:18]	..................FunctionNameSegment             	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:17]	.................FunctionNameSegment              	OneOf.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:16]	................FunctionNameSegment               	Seque.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:15]	...............FunctionNameSegment                	Funct.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:15]	...............FunctionSegment                    	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:14]	..............FunctionSegment                     	AnyNu.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:13]	.............FunctionSegment                      	Seque.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:12]	............FunctionSegment                       	Seque.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:11]	...........FunctionSegment                        	OneOf.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:10]	..........FunctionSegment                         	Funct.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:10]	..........TableExpressionSegment                  	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:12]	............TableReferenceSegment                 	Delim._look_ahead_match IN	    	[ls=1, seg='2']
[PD:0  MD:12]	............TableReferenceSegment                 	Delim._look_ahead_match SI	    	[mq=[], sb=['2']]
[PD:0  MD:12]	............TableReferenceSegment                 	OneOf.match PRN     	    	[ns=1, ps=10, ms=0, pruned=[<Ref: OnKeywordSegment>, <Ref: AsKeywordSegment>, <Ref: UsingKeywordSegment>, <Ref: CommaSegment>, <Ref: CastOperatorSegment>, <Ref: StartSquareBracketSegment>, <Ref: StartBracketSegment>, <Ref: ColonSegment>, <Ref: DelimiterSegment>, <class 'sqlfluff.core.parser.segments.base.BracketedSegment'>], opts=[<Ref: BinaryOperatorGrammar>]]
[PD:0  MD:13]	.............BinaryOperatorGrammar                	OneOf.match PRN     	    	[ns=1, ps=3, ms=0, pruned=[<Ref: ArithmeticBinaryOperatorGrammar>, <Ref: StringBinaryOperatorGrammar>, <Ref: BooleanBinaryOperatorGrammar>], opts=[<Ref: ComparisonOperatorGrammar>]]
[PD:0  MD:14]	..............ComparisonOperatorGrammar           	OneOf.match PRN     	    	[ns=1, ps=9, ms=0, pruned=[<Ref: EqualsSegment>, <Ref: EqualsSegment_a>, <Ref: EqualsSegment_b>, <Ref: GreaterThanSegment>, <Ref: LessThanSegment>, <Ref: GreaterThanOrEqualToSegment>, <Ref: LessThanOrEqualToSegment>, <Ref: NotEqualToSegment_a>, <Ref: NotEqualToSegment_b>], opts=[<Ref: LikeOperatorSegment>]]
[PD:0  MD:15]	...............ComparisonOperatorGrammar          	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:14]	..............ComparisonOperatorGrammar           	OneOf.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:14]	..............BinaryOperatorGrammar               	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:13]	.............BinaryOperatorGrammar                	OneOf.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:13]	.............TableReferenceSegment                	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:12]	............TableReferenceSegment                 	OneOf.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:12]	............TableReferenceSegment                 	NonCo.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:12]	............SingleIdentifierGrammar               	OneOf.match PRN     	    	[ns=2, ps=0, ms=0, pruned=[], opts=[<Ref: NakedIdentifierSegment>, <Ref: QuotedIdentifierSegment>]]
[PD:0  MD:13]	.............SingleIdentifierGrammar              	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:13]	.............SingleIdentifierGrammar              	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:12]	............SingleIdentifierGrammar               	OneOf.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:12]	............TableReferenceSegment                 	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:11]	...........TableReferenceSegment                  	Delim.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:10]	..........TableReferenceSegment                   	Table.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:10]	..........TableExpressionSegment                  	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:9 ]	.........TableExpressionSegment                   	OneOf.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:8 ]	........TableExpressionSegment                    	Table.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:8 ]	........FromExpressionElementSegment              	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:7 ]	.......FromExpressionElementSegment               	Optio.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:6 ]	......FromExpressionElementSegment                	Seque.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:5 ]	.....FromExpressionElementSegment                 	FromE.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:5 ]	.....FromExpressionSegment                        	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:4 ]	....FromExpressionSegment                         	OneOf.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:3 ]	...FromExpressionSegment                          	Seque.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:2 ]	..FromExpressionSegment                           	FromE.match OUT     	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:2 ]	..None                                            	Ref.match OUT       	    	[match=<MatchResult 0/1: ''>, seg="'2'"]
[PD:0  MD:1 ]	.None                                             	Delim.match OUT     	    	[match=<MatchResult 0/2: ''>, seg="'values 1 , 2'"]
[PD:0  MD:0 ]	None                                              	Seque.match OUT     	    	[match=<MatchResult 0/9: ''>, seg="'from values 1 , 2'"]
```

test/fixtures/dialects/ansi/select_from_values.sql
```diff
@@ -0,0 +1,5 @@
+select * from values 1, 2;
```
mcannamela, line 1: readers may want to skip ahead to the [working cases](https://github.com/sqlfluff/sqlfluff/pull/2245/files#diff-65b86a0a7aee1334a536d1ee3a75631989ca84ca93754cccffbe9421c03f34f1R1)

these are still all split up from my troubleshooting

is there a way to document known statements that are legal but do not parse, like marking them as expected failures? or you do all that through issues?

src/sqlfluff/dialects/dialect_spark3.py
barrywhart: Sounds good. Someday, if we have rules looking at these specific kind of segment, they may benefit from making things more explicit. Just a few weeks ago, I refactored the `CASE` grammar to create segments for the various parts (`WHEN`, `ELSE`). It made a couple of rules **much** simpler.

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -783,3 +783,117 @@ class JoinClauseSegment(BaseSegment):
...
+    match_grammar = OneOf(
+        Ref("LiteralGrammar"),
+        Ref("BareFunctionSegment"),
+        Ref("FunctionSegment"),
+    )
+
+
+@spark3_dialect.segment()
+class TupleValue(BaseSegment):
+    """An element of a ``VALUES`` clause that has a multiple columns.
```
barrywhart, line 862:
```suggestion
    """An element of a ``VALUES`` clause that has multiple columns.
```

src/sqlfluff/dialects/dialect_spark3.py
barrywhart: `match_grammar` is like a "hint" to the parser to help it distinguish when a segment starts or stops. `parse_grammar` is a full grammar, i.e. describes the entire structure. Often, the `match_grammar` is not needed. Generally, you consider adding one if the parser is failing to parse some of the test SQL. If things are working okay like this, I would go ahead and keep it.

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -783,3 +783,117 @@ class JoinClauseSegment(BaseSegment):
...
+        WITH a AS (VALUES 1,2) SELECT * FROM a;
+
+    """
+
+    type = "values_clause"
+    match_grammar = StartsWith(Ref.keyword("VALUES"))
+
+    parse_grammar = Sequence(
+        Ref.keyword("VALUES"),
+        DelimitedValues,
```
barrywhart, line 897: Does this work? I thought you'd have to use `Ref("DelimitedValues")`.

test/conftest.py
barrywhart: I see no reason for it, either. Usually, you would do this if the outer function takes parameters and the inner function is a closure, using those parameters. Perhaps it was done for consistency with other fixtures? Or from not being aware of the "why" and copying from an example?

test/dialects/spark3_grammar_test.py
```diff
@@ -0,0 +1,97 @@
...
+
+from sqlfluff.core import FluffConfig, dialect_selector
+from sqlfluff.core.dialects import Dialect
+from sqlfluff.core.parser import Lexer, RawSegment
+from sqlfluff.core.parser.context import RootParseContext
+
+
+@pytest.fixture(scope="function")
+def fresh_spark3_dialect() -> Dialect:
+    """Expand the ansi dialect for use."""
```
barrywhart, line 15:
```suggestion
    """Expand the spark3 dialect for use."""
```

test/dialects/spark3_grammar_test.py
barrywhart: Nothing I'm aware of. @alanmcruickshank wrote the parser;  maybe he has something to add?

test/fixtures/dialects/ansi/select_from_values.sql
barrywhart: Yes, I would probably create issues for those.

barrywhart (commented):
About the parsing issues, you might try and get help from Barry Pollard. He has made a lot of dialect enhancements, as well as helped others with them.

These are some pretty significant additions to the dialect, so I think adding this stuff to the spark3 dialect makes sense, unless we know for certain that other databases support the exact same syntax. I'm _assuming_ these are extensions (i.e. not standard), so even if other databases support them, there maybe enough variation that leaving it in spark3 feels like a safe default for now. We can always move it later.

test/conftest.py
mcannamela: ok, i will leave it as is in this PR but would you be open to a separate PR to eliminate the wrapper if i get ambitious?

src/sqlfluff/dialects/dialect_spark3.py
mcannamela: it appears to, isn't the `Ref` only necessary for something you've potentially redefined?

barrywhart: Maybe. I have always used it for everything. @tunetheweb, @jpy-git: I remember talking about this recently. Did we know that direct class references work, at least sometimes?

mcannamela: https://github.com/sqlfluff/sqlfluff/blob/43b0fc29f26194f8812a40ed6e55af57c8240da0/src/sqlfluff/core/parser/grammar/base.py#L85 seems to anticipate that `BaseGrammar`s or `BaseSegment`s might show up

type hints in the right places might prevent the hapless (myself for instance) from getting lucky if this is not intended to work

src/sqlfluff/dialects/dialect_spark3.py
mcannamela: interesting, i had noticed `parse_grammar` not being defined but not seen a place where there was no `match_grammar`

but you're saying i could generally omit `match_grammar` and only define it if i run into trouble?

barrywhart: Maybe I had it backwards. I remembered you could omit one of them. So I think if you only provide `match_grammar`, it uses the same for both.

test/conftest.py
barrywhart: Sure!

src/sqlfluff/dialects/dialect_spark3.py
jpy-git: The reason you need Ref is because the SQL grammar can often be recursive so using Ref allows the grammar to be lazily loaded therefore avoiding circular references

jpy-git: So yes, best to just use it everywhere 👍

test/dialects/spark3_grammar_test.py
jpy-git: As far as I can see there's not really anything in this file that's not already being tested by the sql/yml tests so think we should remove this file when it comes to the non-draft PR.

(Nice work on the SQL tests btw, they look pretty thorough 😄)

test/conftest.py
jpy-git: I'm fine with you removing this fixture in a separate PR as well. For this PR can you revert this file to keep diffs focused please.

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -783,3 +783,118 @@ class JoinClauseSegment(BaseSegment):
...
```
jpy-git, line 898:
```suggestion
        Ref("DelimitedValues"),
```
See comment [here](https://github.com/sqlfluff/sqlfluff/pull/2245/files#r784352342) for motivation.

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -783,3 +783,118 @@ class JoinClauseSegment(BaseSegment):
...
+
+        VALUES 1,2 LIMIT 1;
+        SELECT * FROM VALUES (1,2) as t (a,b);
+        SELECT * FROM (VALUES (1,2) as t (a,b));
+        WITH a AS (VALUES 1,2) SELECT * FROM a;
+
+    """
+
+    type = "values_clause"
+    match_grammar = StartsWith(Ref.keyword("VALUES"))
```
jpy-git, line 894:
```suggestion
    match_grammar = StartsWith("VALUES")
```
`Ref.keyword` is implied here. The only time you need to use it is if you also need to pass an additional argument e.g. `Ref.keyword("VALUES", optional=True)`

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -783,3 +783,118 @@ class JoinClauseSegment(BaseSegment):
...
+
+    parse_grammar = Sequence(
+        Ref.keyword("VALUES"),
```
jpy-git, line 897:
```suggestion
        "VALUES",
```

src/sqlfluff/dialects/dialect_spark3.py
jpy-git: Makes sense, that's the right approach 👍

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -783,3 +783,118 @@ class JoinClauseSegment(BaseSegment):
...
+        Ref.keyword("AS", optional=True),
+        OneOf(
+            Sequence(
+                # maybe table alias
+                Ref("SingleIdentifierGrammar", optional=True),
+                # maybe column aliases
+                Bracketed(Ref("SingleIdentifierListSegment"), optional=True),
+            ),
+            Ref("QuotedLiteralSegment"),
+        ),
```
jpy-git, line 828:
```suggestion
        OneOf(
            Sequence(
                # maybe table alias
                Ref("SingleIdentifierGrammar", optional=True),
                # maybe column aliases
                Bracketed(Ref("SingleIdentifierListSegment")),
            ),
            # maybe table alias
            Ref("SingleIdentifierGrammar"),
            Ref("QuotedLiteralSegment"),
        ),
```
Not sure if there's a risk of something weird happening if both the table and column aliases are optional so maybe best to be more explicit about the options.

Also what is the situation in which we use `QuotedLiteralSegment`?
Not too familiar with spark3 myself but `SELECT * FROM VALUES (1,2) as 'I am just a string';` seems a bit odd. What does the string represent?

jpy-git (commented):
@mcannamela Couple things to look at, but this looks like it's shaping up pretty nicely so far 💪

src/sqlfluff/dialects/dialect_spark3.py
mcannamela: The `QuotedLiteralSegment` was copy-pasta  from [/src/sqlfluff/dialects/dialect_ansi.py#L820](https://github.com/mcannamela/sqlfluff/blob/84e61e00f6cfcbd8cd5b7e3f3ca100885f5c12f6/src/sqlfluff/dialects/dialect_ansi.py#L820). It struck me as odd as well. 

I'll play around with it in the spark dialect

src/sqlfluff/dialects/dialect_spark3.py
mcannamela: @jpy-git in that case is there a reason not to throw (or fail type checking) if you get anything other than a `Ref` when instantiating a grammar?

like is there ever a case where using `Ref` is forbidden?

test/dialects/spark3_grammar_test.py
mcannamela: it was...a process

these aren't completely redundant since you can verify that a snippet matches more than one grammar, and because of the xfails

my plan is to make issues for the stuff that's still broken, note the issue numbers below, and then come back here and make the tests pass. how would you feel about leaving those in place? 

of course i can also just maintain this file on a branch but my thought is that it could trim the learning curve for e.g. other folks trying to get into spark3 development

jpy-git:
We should hold off on further review on this for now as there are likely a couple things that might become redundant after we merge in #2404. Given the ANSI one affects all dialects and corrects other values related parsing issues, let's get that in first

jpy-git:
@mcannamela Will take another look at reviewing this now

test/fixtures/rules/std_rule_cases/L025.yml
```diff
@@ -166,3 +166,23 @@ test_fail_exasol_values_clause:
   configs:
     core:
       dialect: exasol
+
```
mcannamela, line 169: these rule tests copy-pasta from #2400 per @barrywhart's [suggestion](https://github.com/sqlfluff/sqlfluff/pull/2245#issuecomment-1018533160)

mcannamela:
@barrywhart I've copied tests from your PR as you suggest 

@jpy-git your PR allowed me to reuse the ValuesClauseSegment from ansi, which is great

I'm going to take it out of draft, ought not be more than a couple small items outstanding.

barrywhart:
@mcannamela: A couple of the test suites have failures. One is the normal Python tests.

Another is a new test added yesterday which runs the full `sqlfluff lint` on each of the parse test files in `test/fixtures/dialects/spark3` and the other sibling directories. This tends to catch two kinds of issues:
* Latent rule bugs
* Dialects that may do some things in a new or different way compared to other dialects, causing rules to fail.

Can you look at the first? I'll take an initial look at the second one. Many rules are failing, but the error message is the same for each, and I think several of them reuse some of the same underlying components, so it may be essentially the same error repeated multiple times.

barrywhart:
I have some possible fixes for the rule issues. I'll create a PR against your PR later today with those fixes.

tunetheweb:
> I have some possible fixes for the rule issues. I'll create a PR against your PR later today with those fixes.

Would a separate PR be better? Can then merge main into this branch after?

barrywhart:
@tunetheweb: Good idea! I created it as a standalone PR. It's similar to some of your fixes yesterday. #2488

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -1109,3 +1109,108 @@ class JoinClauseSegment(BaseSegment):
...
+        Ref.keyword("AS", optional=True),
+        OneOf(
+            Sequence(
+                # maybe table alias
+                Ref("SingleIdentifierGrammar", optional=True),
+                # maybe column aliases
+                Bracketed(Ref("SingleIdentifierListSegment")),
+            ),
+            # maybe table alias
+            Ref("SingleIdentifierGrammar"),
```
barrywhart, line 1140: Should we also support `SingleQuotedIdentifierSegment`, as seen in the ANSI dialect?

src/sqlfluff/dialects/dialect_spark3.py
barrywhart: Are all these new segments still needed in the Spark dialect, now that the ANSI dialect provides `ValuesClauseSegment`?

src/sqlfluff/dialects/dialect_spark3.py
tunetheweb: Exactly. You always need a `match_grammar` and it is used to parse the statement if no dedicated `parse_grammar` is provided. It's discussed here: https://docs.sqlfluff.com/en/stable/architecture.html#stage-3-the-parser

As @barrywhart says, `match_grammar` is useful if you can have a simple shortcut to identify the start and end of this grammar. It can then be parsed. Rather than trying to parse with no ending in sight, where it may overshot into the next statement for certain combinations.

Here you are only specifying the start (it starts with `"VALUES"`) so there is no benefit to splitting them out. There was some discussion as to whether that was quicker (i.e. did the parse need to load the whole, potentially complicated statement, then check first word?) but tests showed it's exactly the same.

So merge `match_grammar` and `parse_grammar` here into the one `match_grammar` definition with the full definition as no benefit in it being split.

test/dialects/spark3_grammar_test.py
tunetheweb: I would MUCH prefer these to be moved to SQL and YML files in `test/fixtures/dialects/spark3`.

There are several benefits to being in there:
- Easier to read and write
- We generate the parse tree in the YML file and commit this to git so can see any unexpected changes to this for future PRs - even if completely unrelated to this.
- We're testing whole SQL statements, rather than partial pieces, which might work individually but not together. In the real world we write SQL and not fragments.
- Allows us to run our [lint test](https://github.com/sqlfluff/sqlfluff/pull/2473).
- Deeper tests usually aren't required.

The downside is you can't be as specific in checks (it's a straight match or fail), so sometimes need Python tests for more complicated code, or python functions within segments. But they should not be used for SQL statements for above reasons.

tunetheweb: Oh and feel free to add commented out tests with TODOs in the SQLs for future PRs.

src/sqlfluff/dialects/dialect_spark3.py
mcannamela: ANSI provided `ValuesClauseSegment` before, it just didn't work because it was trying to parse it as a function call. So I was able to remove that segment from my changes, but I don't think that was the only issue.

 After merging recent changes, I tried:

1. comment out my new segments
2. run dialects tests, observed failures
3. `tox -e generate-fixture-yml`, maybe it parses but we have the wrong expectation
4. run dialects tests, still observe failures
5. uncomment some of my new segements; go to 1 until tests pass or all new segments are required 

Does that seem like it would determine whether the upstream changes "just work" for me?

src/sqlfluff/dialects/dialect_spark3.py
mcannamela: ahh thanks that clears things up, i'll rtfm

test/dialects/spark3_grammar_test.py
mcannamela: Commented tests with TODOs work for me.

Do you have thoughts on how to make it more obvious how to write this kind of test when needed during development? I found myself needing the fine-grained test to understand what was going wrong and having to do a fair amount of reading to understand how to set it up. Of course I now know right where this commit is and can always dig it up, but just thinking of the next casual dialect developer. (Totally fine if not I have what I need to adjust this PR)

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -1109,3 +1109,108 @@ class JoinClauseSegment(BaseSegment):
...
+
+    type = "alias_expression"
+    match_grammar = Sequence(
+        Ref.keyword("AS", optional=True),
+        OneOf(
+            Sequence(
+                # maybe table alias
+                Ref("SingleIdentifierGrammar", optional=True),
+                # maybe column aliases
+                Bracketed(Ref("SingleIdentifierListSegment")),
```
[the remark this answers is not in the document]
mcannamela: 👍  good catch, i'll verify

test/dialects/spark3_grammar_test.py
tunetheweb: I tend to start with small bits of SQL and then just use `sqlfluff parse` to look at the tree and comment/uncomment code until I get it to work. Not the most efficiint I agree but usually gets me there.

I do this in one of three ways:
- Command line for simple statements (`echo "select * from values 1, values 2;" | sqlfluff parse --dialect spark3 -`
- `test.sql` for more complex ones (`sqlfluff parse --dialect spark3 test.sql`). This is excluded from git so good place to test.
- From `test/fixtures` directory (test driven development! I should do this more).

That's usually sufficient for me but maybe some (particularly Python developers - which I am not), may prefer your approach.

Either way, there's good reason to move them to the standard ones, once you've figured it out as discussed above.

src/sqlfluff/dialects/dialect_spark3.py
mcannamela: hm, I am not grokking what that would be, do you know an example that would cover it?

`SingleIdentifierGrammar` should cover both naked and quoted identifiers, which in spark3 means single backticks

tunetheweb: In ANSI we have `SingleIdentifierGrammar` which is `NakedIdentifierSegment` (e.g. `barry`) or `QuotedIdentifierSegment` (e.g. `"barry"` in most dialects or with backticks in spark3).

However there is also a `SingleQuotedIdentifierSegment` (e.g. `'barry'`) which is only allowed to be used in some places, whereas `QuotedIdentifierSegment`s are allowed to be used in a lot more places. Therefore `SingleQuotedIdentifierSegment` it is not part of `SingleIdentifierGrammar` and needs to be explicitly added to the few places it's acceptable.

The Alias is one of those special places that allows this as well as normal quoted identifiers. Therefore it is explicitly added there in ANSI:

```py
    type = "alias_expression"
    match_grammar = Sequence(
        OneOf(
            Sequence(
                # Column alias in VALUES clause
                Bracketed(Ref("SingleIdentifierListSegment"), optional=True),
            ),
        ),
    )
```

So @barrywhart 's question is, should it be here too? If not, (because spark3 only allows backticks quotes even here) it might be good to add a comment to that effect as it's different from ANSI definition.

mcannamela: i see, let me verify the behavior and i'll either amend to cover the single quoted case or note the variance from ANSI

test/dialects/spark3_grammar_test.py
mcannamela: Thanks, TDD out of the fixtures directory is approximately what I did until I was stumped by the parser's terse output. It worked well for most of what I've done so far, though I feel I found the limits while developing this PR. 

I'll noodle on how to take the pain/learning curve out of temporarily standing up very specific Python tests.

src/sqlfluff/dialects/dialect_spark3.py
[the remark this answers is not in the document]
mcannamela: remedied

src/sqlfluff/dialects/dialect_spark3.py
mcannamela: it's not allowed, i've called this out in the docstring

mcannamela:
Thanks everybody for reviewing; I think I have addressed all feedback so assuming checks pass I'm hunting for ✅ s

src/sqlfluff/dialects/dialect_spark3.py
barrywhart: I think someone recently fixed the issue where it would parse as a function call. Was this occurring when there was exactly _one_ column?

@tunetheweb: Do you remember the history about this? IIRC, it was fixed in the last week or so.

barrywhart: Found it! See [this PR comment:](https://github.com/sqlfluff/sqlfluff/pull/2404#discussion_r790067975)
>This prevents the values clause from accidently being parsed as a function as described in #2406

tunetheweb: Going to merge in `main` here to see what it looks like.

tunetheweb (approved):
OK think this is good to merge. Thanks for all your work on it. Realise there's still more to do, but had a quick look and can't see a simple solution so let's merge as is and tackle later.

One suggestion to correct docstring back ticks that I'll merge in.

PR #2246, opened 2022-01-06
Add support for nested Jinja macros

### LINKED ISSUE #1077: Nested macros cause an error using Jinja templater and load_macros_from_path config
## Expected Behaviour
When running `sqlfluff lint` using the Jinja templater and `load_macros_from_path_config` setting, macros used within other macros should be parsed/linted as if they were being used in a top-level SQL file.

## Observed Behaviour
Calling a macro (e.g. `week_start_date(date)`) from within another macro (e.g. `school_year_start_date(date)`) causes linting to fail with an error message similar to the following:
```bash
== [test_model_broken.sql] FAIL
L:   0 | P:   0 |  TMP | Unrecoverable failure in Jinja templating:
                       | 'week_start_date' is undefined. Have you configured your
                       | variables?
```

Note that this error does NOT occur when calling `week_start_date(date)` directly from a SQL file (e.g. `test_model_working.sql` in the repo linked below).

## Steps to Reproduce
1. Clone `https://github.com/mjirv/sqlfluff-nested-macro-issue`
2. Run `sqlfluff lint .` in the top-level folder of that repo

## Version
- sqlfluff 0.5.6
- Python 3.7.3

## Configuration
```
[sqlfluff:templater:jinja]
load_macros_from_path=macros
```

barrywhart:

Fixes #1077 

This fixes a limitation with the `jinja` templater that several users have reported: it was impossible for a macro to invoke another macro stored in a separate file.

### Pull Request checklist

  - Other.


src/sqlfluff/core/templaters/jinja.py
```diff
@@ -232,6 +239,7 @@ def get_context(self, fname=None, config=None) -> Dict:
                 (self.templater_selector, self.name, "load_macros_from_path")
             )
             if macros_path:
+                env = self._get_jinja_env(macros_path)
```
jpy-git, line 242: makes sense, very elegant fix 😄

barrywhart:
@jpy-git, @tunetheweb: Can y'all review this? It's a pretty simple fix for an issue several users have reported over the past 8 months.

test/core/templaters/jinja_test.py
```diff
@@ -413,6 +413,14 @@ def assert_structure(yaml_loader, path, code_only=True, include_meta=False):
...
         print(p)
         raise RuntimeError(p[0][1])
+    # Enable this code to create a new test .yml file, i.e. when adding a new
+    # test case.
+    # if parsed:
+    #     r = parsed.as_record(code_only=True, show_raw=True)
+    #     if r:
+    #         with open(path + ".yml", "w") as f:
+    #             import yaml
+    #             yaml.dump(r, f, sort_keys=False)
```
jpy-git, line 423: As we mentioned on slack it would def be nice to get this included in `generate-fixture-yml` (or it's own script if you prefer). Could you make a small issue and copy this code into it and then remove from this PR so we keep the code here clean and then we can decide where to put this 😄

jpy-git (commented):
@barrywhart Really nice!

test/core/templaters/jinja_test.py
barrywhart: I created an issue: #2247. I'll remove this code.

barrywhart:
@jpy-git: Done! (Issue created for the YML thing, commented code removed.)

jpy-git (approved):
@barrywhart looks great, absolutely smashing it with the templater/corruption fixes atm! 👏 LGTM 🚀 ✨

jpy-git:
@barrywhart bit of a random question:
What happens if you recursively defined a jinja macro, does it blow up or cause any weird sql errors? 😛 

Real edge case stuff but just curious what sqlfluff would do
https://stackoverflow.com/questions/32771924/macro-recursion-in-jinja2

barrywhart:
@jpy-git: I haven't tried it, but I think it's likely to fail because of how the `JinjaSlicer` works. I _think_ this would be an instance of the thing I [mentioned in Slack yesterday](https://sqlfluff.slack.com/archives/G01G58WRJNR/p1641429286005200). If you think about it, Jinja is a weird programming environment, because almost everything you do is generating code. Normally, code generation is unusual. And since SQLFluff needs to understand where every output character came from, it needs to somewhat follow what's going on. It's pretty easy to hit the limits of what the slicer (and thus SQLFluff) can handle, but it's been unusual in practice, because the Jinja is usually not that complex. Any time you compute something with Jinja and then send the output through Jinja again, that's likely to fail. Someday, we may need to take a much deeper dive into the Jinja package to see if there's a more robust way to do this, but I hope not. 😅

jpy-git:
Makes sense! Realistically recursive Jinja is a 0.00001% situation and I really doubt there's any crossover between someone sensible enough to use a linter but insane enough to write SQL using a recursive Jinja template 🤣 Only a difficult reviewer like myself.

From what I've seen in the issues the `JinjaSlicer` seems to handle the standard situations pretty well plus the more difficult things you've fixed these last couple days so don't think there's a burning need to update it yet 😄 We can keep an eye on issues going forwards

barrywhart:
@jpy-git: Yes, and as I mentioned to the user in a recent issue, your linting config doesn't have to exactly match your production config, so even if you're doing something weird, you can potentially set up flags or use dummy config values to disable those bits during linting. People already do this in some ways (e.g. setting up a `.sqlfluff` file with non-production Jinja template values).

PR #2230, opened 2022-01-04
Fix connection issue in dbt 1.0.0

NiallRees:
Resolves #2113 

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #2254, opened 2022-01-07
Docker Compose environment for SQLFluff developers

### LINKED ISSUE #2248: Create a Docker Compose environment to help SQLFluff developers run the dbt templater tests locally
PR #2230 updates the dbt tests to require a Postgres server and a database connection.

This is currently not possible on M1 Macs due to the inavailability of the `psycopg2` database driver package not yet being available. Even on other platforms, requiring a database to be set up can be onerous -- what if you can't get the right version? Or what if you already have Postgres installed and are using it for something else and the passwords don't match, etc. etc.?

Let's create a `docker-compose.yml` for the dbt tests consisting of the configured postgres instance image and the python image with sqlfluff and dbt dependencies installed (both images amd64). That way we can just `docker compose up` our dbt dev environment which would make contributing to the dbt side of the project a lot easier for anyone. I prefer the more traditional approach with the main package, but the complexity for dbt development seems a good fit.

Relevant docs if you're not familiar: https://docs.docker.com/compose/
(disclaimer: this is not anything to do with official sqlfluff docker image, just using docker as a dev environment)

barrywhart:

Fixes #2248 

**If this PR looks reasonable, it would be cool to go ahead and merge it sooner rather than later, in order to potentially help with some of the work on the dbt connection PR, #2248.**  It's a "dev only" change -- i.e. there should be no risk to users, the CI builds, etc.

NOTE: This may be useful for anyone, but for now, the primary motivation for this work is to allow Mac M1 users to work on the dbt plugin, whose tests require installing a package (`psycopg2`) that is currently challenging to install on M1. Hence, the Docker environment explicitly uses an AMD64 Python image, which runs on Docker M1 using emulation. This works but is pretty slow.

Because of the slow emulation performance on M1, you probably don't want to use this environment to run **all** the SQLFluff tests (even **discovering** all the tests takes a pretty long time, must less running them!), but if you want to spin up the environment and verify things are mostly working, you can use do the following, which starts up the environment and runs the L003 tests. This runs pretty quickly.

```
docker/startup
docker/shell
py.test -v test/rules -k L003
```

Below, you can see a screenshot where I ran one of the dbt tests in Docker on an M1, using Niall's in-progress dbt PR (#2248). (In order to do this, I had to create a special temporary branch combining Niall's changes and mine.) I also had to manually change the host setting in `plugins/sqlfluff-templater-dbt/test/fixtures/dbt/profiles_yml/profiles.yml` from `localhost` to `postgres`.

---
image
---

### Pull Request checklist

  - Other.


jpy-git:
@barrywhart will take a proper look this eve. In the meantime would you be able to move these files into a `plugins/sqlfluff-templater-dbt/docker` folder so it's clear they're part of the dbt development and doesn't confuse with the official docker image stuff?

barrywhart:
@jpy-git: Yup, moved them just now. 👍

jpy-git (approved):
@barrywhart couple initial thoughts:
1) First off I was able to install this and run the dbt tests really easily which is awesome 🎉 
2) Is there a way we could get tox included in this so running the tests is consistent with non-docker development? Might need some tweaking on the docker image (N.B. running `pytest plugins/sqlfluff-templater-dbt/test` works fine in the meantime).
3) I know in this [thread](https://github.com/sqlfluff/sqlfluff/pull/2230/files#r779941667) I suggested that we could maybe use this in CI. However, on playing around it feels like this is a really nice approach for developing the latest master but not for handling the many various permutations of python versions and sqlfluff + sqlfluff-templater-dbt version permutations in the CI. In that case I think @NiallRees' use of a postgres docker image in the GHA makes more sense since tox handles all the permutations (just need to get it working for Windows).

The things I highlighted above are more considerations for improvements/future work. Given that this works currently and is a development tool that doesn't affect production code, I'm very happy to just merge this in for us to work on #2230 and we can iterate further on the design with addtional PRs as we refine the process 😄 

LGTM 🚀 ✨

barrywhart:
Thanks for the review, @jpy-git! I'm cool with adding tox support, maybe in a separate PR. TBH, I don't use tox locally, so I'm not familiar with it. I tend to choose and run tests for one Python version locally, then rely the CI build for the others.

Merging...

PR #2257, opened 2022-01-07
jy_cleanup

jpy-git:
some tidy up on #2230

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -45,7 +45,6 @@ commands =
     python {toxinidir}/util.py clean-tests
     # Run tests
     pytest -vv -rs --cov=sqlfluff --cov-report=xml {posargs: {toxinidir}/test}
-    dbt{017,018,019,020,021,100}: pytest -vv -rs --cov=sqlfluff --cov-report=xml {posargs: test} # Run the connection failure test in a separate invocation to reset the database connection
 
 [testenv:cov-init]
 setenv =
--- a/tox.ini
+++ b/tox.ini
@@ -44,8 +44,8 @@ commands =
     # Clean up from previous tests
     python {toxinidir}/util.py clean-tests
     # Run tests
-    pytest -vv -rs --cov=sqlfluff --cov-report=xml -m "not dbt_connection_failure" {posargs: {toxinidir}/test}
-    dbt{017,018,019,020,021,100}: pytest -vv -rs --cov=sqlfluff --cov-report=xml -m "dbt_connection_failure" {posargs: test} # Run the connection failure test in a separate invocation to reset the database connection
+    pytest -vv -rs --cov=sqlfluff --cov-report=xml {posargs: {toxinidir}/test}
+    dbt{017,018,019,020,021,100}: pytest -vv -rs --cov=sqlfluff --cov-report=xml {posargs: test} # Run the connection failure test in a separate invocation to reset the database connection
 
 [testenv:cov-init]
 setenv =
--- a/plugins/sqlfluff-templater-dbt/test/templater_test.py
+++ b/plugins/sqlfluff-templater-dbt/test/templater_test.py
@@ -287,11 +287,10 @@ def test__templater_dbt_templating_absolute_path(
 @pytest.mark.parametrize(
     "fname,exception_msg",
     [
-        pytest.param(
+        (
             "compiler_error.sql",
             "dbt compilation error on file 'models/my_new_project/compiler_error.sql', "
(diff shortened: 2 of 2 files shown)


.github/workflows/ci-tests.yml
```diff
@@ -203,7 +203,7 @@ jobs:
       # Do not set explicitly temp dir for dbt as causes problems
       # None of these test need temp dir set
       run: |
-        python -m tox -e dbt018-winpy -- plugins/sqlfluff-templater-dbt -m "not dbt_connection_failure and not needs_dbt_connection"
+        python -m tox -e dbt018-winpy -- plugins/sqlfluff-templater-dbt -m "not needs_dbt_connection"
```
jpy-git, line 206: dbt_connection_failure is mocked so never actually connects to db, let's try not excluding

tox.ini
```diff
@@ -44,8 +44,8 @@ commands =
     # Clean up from previous tests
     python {toxinidir}/util.py clean-tests
     # Run tests
-    pytest -vv -rs --cov=sqlfluff --cov-report=xml -m "not dbt_connection_failure" {posargs: {toxinidir}/test}
-    dbt{017,018,019,020,021,100}: pytest -vv -rs --cov=sqlfluff --cov-report=xml -m "dbt_connection_failure" {posargs: test} # Run the connection failure test in a separate invocation to reset the database connection
```
jpy-git, line 48: These are no longer necessary, just resetting

.github/workflows/ci-tests.yml
jpy-git: okay CI shows that this test still connects to the db, we just mock the return value to pretend it fails. Initial proposal is right. Removed.

tunetheweb: Good shout.

I also think I also think I added one too many:

image

Would you mind reverting that as part of this PR? Think I read it as "DB Error" instead of "DBT Error" and thought it was needed here but don't think it is.

jpy-git: no worries will do 😄

tox.ini
tunetheweb: I think these might be necessary, as otherwise it uses the DB connection already established for the previous linting tests.

tunetheweb: Oh maybe not if you've found the alternative reason for it failing above. Ignore me.

jpy-git: @barrywhart I've combined these two lines and now the mocked dbt function isn't happy. That makes me think that the mocking isn't currently removing the need for re-establishing the dbt connections?

jpy-git: but actually these test pass perfectly fine for me locally in the docker-compose environement 🤔

barrywhart: What do you mean by this?
>now the mocked dbt function isn't happy

barrywhart: I see that the dbt 1.0 tests are passing, but the older dbt versions are failing. That makes sense in terms of the mock, because the mock only comes into play for dbt 1.0. We can do one of two things:
* Only run the test if dbt version >= 1.0
*  Update the test to also mock & fail on older dbt versions

I suggest the former, because all this (the test, the PR, everything) is new for dbt 1.0 anyway.

tunetheweb: But why does it work in the original PR without these additional changes?

jpy-git: ah good shout thanks 😄

barrywhart: Ok, here's what I did (first 3 lines are new):
```
@pytest.mark.skipif(
    DBT_VERSION_TUPLE < (1, 0), reason="mocks a function that's only used in dbt >= 1.0"
)
@mock.patch("dbt.adapters.postgres.impl.PostgresAdapter.set_relations_cache")
@pytest.mark.dbt_connection_failure
def test__templater_dbt_handle_database_connection_failure(
    set_relations_cache, project_dir, dbt_templater  # noqa: F811
):
```

barrywhart: @tunetheweb: IIUC, in dbt < 1.0, a database connection was not required. And the version of the test in `main` **only** checks for (i.e. enforces) a connection failure if dbt version >= 1.0.

https://github.com/sqlfluff/sqlfluff/blob/main/plugins/sqlfluff-templater-dbt/test/templater_test.py#L352L360

barrywhart: @jpy-git: Note that I pushed the test change to Niall's branch, not yours. So your PR branch will need updating in order to pick up this change.

barrywhart (approved):

tunetheweb (approved):

PR #2258, opened 2022-01-07
TSQL: Add OFFSET unreserved keyword

jpers36:
OFFSET is a valid unreserved keyword for TSQL but is missing from the list.

https://docs.microsoft.com/en-us/sql/t-sql/queries/select-order-by-clause-transact-sql?view=sql-server-ver15#syntax

The TSQL dialect currently fails with parsing WITH(NOLOCK) query hints due to its check against ANSI FromExpressionElementSegment, which contains OFFSET.  This PR adds OFFSET as a unreserved TSQL keyword to remedy.

### Pull Request checklist


jpy-git (approved):
Makes sense, good spot 🚀 ✨

Issue #2251, opened 2022-01-07
L025 incorrectly removes column aliases


mcannamela:
First mentioned in https://github.com/sqlfluff/sqlfluff/issues/1462 

## Expected Behaviour
When a column or table alias is used in a query, it should not be removed. 

## Observed Behaviour
When column aliases are used but table aliases are not (e.g. `select a, b from t1 t2(a,b)`), the fix for L025 removes both the table and column aliases, corrupting the query. 

## Steps to Reproduce
test_fix.sql:
```
select
    col_1,
    col_2
from t1 as t(col_1, col_2)
```

```
root@38b5ef159c59:/dbt# sqlfluff parse --ignore-local-config --config .sqlfluff.lean test.sql && sqlfluff fix --ignore-local-config --config .sqlfluff.lean test.sql
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    base:
[L:  1, P:  1]      |        select_statement:
[L:  1, P:  1]      |            select_clause:
[L:  1, P:  1]      |                keyword:                                      'select'
[L:  1, P:  7]      |                [META] indent:
[L:  1, P:  7]      |                newline:                                      '\n'
[L:  2, P:  1]      |                whitespace:                                   '    '
[L:  2, P:  5]      |                select_clause_element:
[L:  2, P:  5]      |                    column_reference:
[L:  2, P:  5]      |                        identifier:                           'col_1'
[L:  2, P: 10]      |                comma:                                        ','
[L:  2, P: 11]      |                newline:                                      '\n'
[L:  3, P:  1]      |                whitespace:                                   '    '
[L:  3, P:  5]      |                select_clause_element:
[L:  3, P:  5]      |                    column_reference:
[L:  3, P:  5]      |                        identifier:                           'col_2'
[L:  3, P: 10]      |            newline:                                          '\n'
[L:  4, P:  1]      |            [META] dedent:
[L:  4, P:  1]      |            from_clause:
[L:  4, P:  1]      |                keyword:                                      'from'
[L:  4, P:  5]      |                whitespace:                                   ' '
[L:  4, P:  6]      |                from_expression:
[L:  4, P:  6]      |                    [META] indent:
[L:  4, P:  6]      |                    from_expression_element:
[L:  4, P:  6]      |                        table_expression:
[L:  4, P:  6]      |                            table_reference:
[L:  4, P:  6]      |                                identifier:                   't1'
[...]

Issue #2030, opened 2021-12-03
Incorrect reporting of "Alias is never used in SELECT statement"


jamiet-msm:
## Expected Behaviour
sqlfluff shouldn't raise any issues. Its complaining that alias `t` isn't used when in fact it is, its referenced in `TO_JSON_STRING()`

It is valid code 
![image](https://user-images.githubusercontent.com/79862277/144639463-5fc8ea87-47d3-4a70-bf29-509a70319a73.png)

## Observed Behaviour
> L:   4 | P:   6 | L025 | Alias 't' is never used in SELECT statement.

![image](https://user-images.githubusercontent.com/79862277/144609131-7fbccc88-b3aa-4740-b26d-11e70ea71afe.png)

## Steps to Reproduce

put this code into file.sql:

```sql
SELECT TO_JSON_STRING(t) AS json
FROM (
        SELECT 1 AS c
) AS t
```

run `sqlfluff lint --dialect bigquery path/to/file.sql`

## Dialect

bigquery

## Version

sqlfluff, version 0.8.2
Python 3.8.12

## Configuration

n/a

tunetheweb:
That's an interesting bit of SQL there. It's raw to pass a whole result set to a function, which is why I think rule L025 doesn't cater for it.

This works for example and doesn't throw a linting error:

```sql
SELECT TO_JSON_STRING(t.c) AS json
FROM (
        SELECT 1 AS c
) AS t
```

I'm presuming this is a cut down example, and there's a reason you don't do just that, but could you expand upon it a little more so we can decide if this is something L025 should support or not?

jpy-git:
@tunetheweb this is valid syntax (https://roboquery.com/app/syntax-to-json-string-function-bigquery)

TO_JSON_STRING takes in a result set and returns each row in a combined JSON format.

Parsing the query shows that `t` is being interpreted as a column reference (which is correct for most functions) rather than a table reference. Hence why it triggers L025.
![image](https://user-images.githubusercontent.com/80432516/144757265-607ecdfc-ea87-459e-8b19-fb23e74ff8f9.png)

Is there a way we could add to the bigquery grammar so that this specific `TO_JSON_STRING` function parses the `t` as a table reference? Potentially L025 will start behaving once it knows it's looking at a table reference.

jamiet-msm:
> I'm presuming this is a cut down example, and there's a reason you don't do just that, but could you expand upon it a little more so we can decide if this is something L025 should support or not?

[...]

PR #2273, opened 2022-01-10
L025 should look at subqueries

### LINKED ISSUE #1462: L025 removes alias which is used in subquery
_If this is a parsing or linting issue, please include a minimal SQL example which reproduces the issue, along with the `sqlfluff parse` output, `sqlfluff lint` output and `sqlfluff fix` output when relevant._

## Expected Behaviour
With fixing ~L031~ L025 the alias should replaced correctly also in subqueries.

## Observed Behaviour
A alias from the parent query is used in a subquery. After fixing the file, the alias of the parent was removed while the alias within the subquery remains.

## Steps to Reproduce
Fix this script:
```
WITH base AS (
    SELECT
        nvl(col1, col2) AS col1,
        col2,
        col3,
        col4,
        col5
    FROM
        SRC_PRD.MDM_EXTRACT
),
without_dup AS (
    SELECT
        DISTINCT col1,
        col2,
        col3,
        col4
    FROM
        base
)
SELECT
    col1,
    col2,
    col3,
    col4,
    (
        SELECT
            count(*)
        FROM
            base b
        WHERE
            a.col2 = b.col2
            AND b.col5 = 'X'
    ) AS X_CNT,
    (
        SELECT
            count(*)
        FROM
            base b
        WHERE
            a.col2 = b.col2
            AND b.col5 = 'Y'
    ) AS Y_CNT,
    (
        SELECT
            count(*)
        FROM
            base b
        WHERE
            a.col2 = b.col2
            AND b.col5 = 'Z'
    ) AS Z_CNT,
FROM
    without_dup a;
```

Alias `a` from `without_dup` is removed but lives in the where clause of the subselects.

## Dialect
exasol

## Version
sqlfluff, version 0.6.6
Python 3.8.12
## Configuration
default

barrywhart:

Fixes #1462

This is basically a rewrite of `L025`. Previously, it inherited from `L020`. Now, its implementation largely relies on `SelectCrawler`. A couple of other rules needlessly inherited from `L025` that could've simply inherited from `L020` -- I updated those, because after the rewrite, inheriting from `L025` no longer worked.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L025.py
+++ b/src/sqlfluff/rules/L025.py
@@ -1,15 +1,37 @@
 """Implementation of Rule L025."""
 
-from sqlfluff.core.rules.base import LintFix, LintResult
+from dataclasses import dataclass, field
+from typing import cast, List, Set
+
+from sqlfluff.core.dialects.base import Dialect
+from sqlfluff.core.rules.analysis.select import get_select_statement_info
+from sqlfluff.core.rules.analysis.select_crawler import (
+    Query as SelectCrawlerQuery,
+    SelectCrawler,
+)
+from sqlfluff.core.rules.base import (
+    BaseRule,
+    LintFix,
+    LintResult,
+    RuleContext,
+    EvalResultType,
+)
 from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 from sqlfluff.core.rules.functional import Segments
 import sqlfluff.core.rules.functional.segment_predicates as sp
-from sqlfluff.rules.L020 import Rule_L020
 from sqlfluff.core.dialects.common import AliasInfo
 
 
+@dataclass
+class L025Query(SelectCrawlerQuery):
+    """SelectCrawler Query with custom L025 info."""
+
+    aliases: List[AliasInfo] = field(default_factory=list)
+    tbl_refs: Set[str] = field(default_factory=set)
+
+
 @document_fix_compatible
-class Rule_L025(Rule_L020):
+class Rule_L025(BaseRule):
     """Tables should not be aliased if that alias is not used.
 
     | **Anti-pattern**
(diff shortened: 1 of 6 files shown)


src/sqlfluff/rules/L025.py
```diff
@@ -38,50 +59,78 @@ class Rule_L025(Rule_L020):
...
+            cls._analyze_table_aliases(cast(L025Query, child), dialect)
+
+    @classmethod
+    def _resolve_and_mark_reference(cls, query: L025Query, ref: str):
+        # Does this query define the referenced alias?
+        if any(ref == a.ref_str for a in query.aliases):
+            # Yes. Record the reference.
+            query.tbl_refs.add(ref)
+        elif query.parent:
+            # No. Recursively check the query's parent hierarchy.
```
barrywhart, line 113: This is the key area of the fix: Now we'll correctly resolve a reference to a table alias that was defined in an "ancestor" query.

src/sqlfluff/rules/L025.py
```diff
@@ -38,50 +59,78 @@ class Rule_L025(Rule_L020):
...
+            cls._resolve_and_mark_reference(cast(L025Query, query.parent), ref)
+
+    @classmethod
+    def _report_unused_alias(cls, alias: AliasInfo) -> LintResult:
```
barrywhart, line 117: This code is identical to lines 66-86 of the old implementation, so it should not require a close review.

tunetheweb:
@barrywhart does this also solve #2251 and #2030 ?

barrywhart:
It does not address #2251. That appears to be related to an unusual alias syntax supported by the Spark3 dialect. I just commented on that other issue to ask for more context.

It also does not address #2030. That issue, also, involves something dialect-specific -- in that case, BigQuery. This reminds me a bit of some previous challenges we've had with BigQuery, where its support of structures makes it difficult for SQLFluff to understand the query.

It would be good to fix those at some point. Lately, I've been mostly focused on issues where the user's SQL is corrupted or broken, and I've prioritizing things approximately as follows:
* Fix issues that cut across multiple rules (e.g. template-related corruption).
* Fix rule issues that affect the ANSI (or most) dialects.
* Rule-specific, dialect-specific issues: I have not prioritized these due to time constraints. Would be cool to get more people working on rules. 💪

src/sqlfluff/rules/L025.py
```diff
@@ -38,50 +59,78 @@ class Rule_L025(Rule_L020):
...
+                # Record the aliases.
+                query.aliases += select_info.table_aliases
+
+                # Look at each table reference; if its an alias reference,
+                # resolve the alias: could be an alias defined in "query"
+                # itself or an "ancestor" query.
+                for r in select_info.reference_buffer:
+                    for tr in r.extract_possible_references(level=r.ObjectReferenceLevel.TABLE):  # type: ignore
+                        # This function walks up the query's parent stack if necessary.
+                        # Reuses the existing "for r in references:" loop from L025.
```
tunetheweb, line 99: Should we remove this line?

```suggestion
```

tunetheweb (changes requested):
This looks quite nice. Need to find some time to dig into your rule work more.

One suggestion to remove a comment.

Also any performance difference (either better or worse) with this re-implementation? Any easy way of benchmarking that?

src/sqlfluff/rules/L025.py
barrywhart: Oh yes -- that's left over from my pseudocode.

tunetheweb (approved):
LGTM. Would still be nice to know if it's more or less performant but since we don't demand that for other PRs I'll not hold up approval for that.

barrywhart:
In the typical "nothing to do" case (where it's not the target segment type or the query defines no aliases), performance should be similar. When there _are_ aliases, it'll be slower, because now it's doing a recursive walk of the parse tree underneath in order to find subqueries. The `SelectCrawler` class, which does the work, is already used by L044 and L045, so I wouldn't expect any big surprises here.

At some point, if enough rules use `SelectCrawler` and we are worried about performance, we could look at ways to cache/reuse this "query tree" info across rules (i.e. if several rules want a `SelectCrawler` on the same query, avoid having each rule redo the work).

barrywhart:
I don't know if we have a typical way to benchmark performance. It would be easy to get precise timings from each of the rule test cases, but those are usually much smaller than real SQL. It could be cool if there were some real-world open-source SQL projects we could monitor how SQLFluff performs with.

barrywhart:
@tunetheweb: I removed the extraneous comment and answered your questions. Ready for another review when you have time. Thanks!

PR #2274, opened 2022-01-10
Optionally allow additional configurable characters in L057

tunetheweb:

Allow hyphens in L057 as quoted identifiers (same as we allow underscores), as a lot of these are flagging for a lot of my table names in BigQuery.

@jpers36 , any issues with this since you created L057? If you would prefer not to allow this, then happy to move to a BigQuery specific bit of L057 since that already exists.

None

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L057.py
+++ b/src/sqlfluff/rules/L057.py
@@ -45,6 +45,7 @@ class Rule_L057(BaseRule):
         "quoted_identifiers_policy",
         "unquoted_identifiers_policy",
         "allow_space_in_identifier",
+        "additional_allowed_identifiers",
     ]
 
     def _eval(self, context: RuleContext) -> Optional[LintResult]:
@@ -53,6 +54,7 @@ class Rule_L057(BaseRule):
         self.quoted_identifiers_policy: str
         self.unquoted_identifiers_policy: str
         self.allow_space_in_identifier: bool
+        self.additional_allowed_identifiers: str
 
         # Exit early if not a single identifier.
         if context.segment.name not in ("naked_identifier", "quoted_identifier"):
@@ -88,12 +90,16 @@ class Rule_L057(BaseRule):
             if identifiers_policy_applicable(
                 self.quoted_identifiers_policy, context.parent_stack
             ) and not (
-                identifier.replace("_", "").replace("-", "").isalnum()
+                identifier.replace("_", "")
+                .translate(str.maketrans("", "", self.additional_allowed_identifiers))
+                .isalnum()
                 or (
                     self.allow_space_in_identifier
                     and identifier.replace("_", "")
-                    .replace("-", "")
                     .replace(" ", "")
+                    .translate(
+                        str.maketrans("", "", self.additional_allowed_identifiers)
+                    )
                     .isalnum()
                 )
             ):
--- a/src/sqlfluff/rules/L057.py
+++ b/src/sqlfluff/rules/L057.py
@@ -48,6 +48,28 @@ class Rule_L057(BaseRule):
         "additional_allowed_identifiers",
     ]
 
+    @staticmethod
+    def _standard_cleanups(
+        identifier: str,
+        allow_space_in_identifier: bool,
+        additional_allowed_identifiers: str,
+    ) -> str:
+
+        # We always allow underscores so strip them out
+        identifier = identifier.replace("_", "")
+
+        # Set the identified minus the allowed characters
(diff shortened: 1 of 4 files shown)


tunetheweb:
Can we hold off merging this until @jpers36 gives his opinion please?

jpers36:
@tunetheweb maybe this is something that would best be handled in a config setting, such as either a flag or, more flexibly, a list of allowable non-alphanumerics.

tunetheweb:
> @tunetheweb maybe this is something that would best be handled in a config setting, such as either a flag or, more flexibly, a list of allowable non-alphanumerics.

Good idea done.

In theory, this would allow us to get rid of `allow_space_in_identifier` and handle that in that. Alternatively could keep that (for backwards compatibility and for ease of setting as not always obvious should use quotes). WDYT?

jpers36:
@tunetheweb makes sense to me to leave `allow_space_in_identifier ` for now for backwards compatibility.  We can always remove it later.

tunetheweb:
OK think this is ready for review now.

src/sqlfluff/rules/L057.py
```diff
@@ -62,7 +64,11 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
             # Evaluate unquoted identifiers.
             if identifiers_policy_applicable(
                 self.unquoted_identifiers_policy, context.parent_stack
-            ) and not (context.segment.raw.replace("_", "").isalnum()):
+            ) and not (
+                context.segment.raw.replace("_", "")
```
barrywhart, line 68: Nitpick: IIUC, we are ignoring occurrences of `_` as well as `additional_allowed_identifiers`, but doing so in two different ways. Think it might be clearer to incorporate the `_` into the other code somehow? One thought: set the default value of `additional_allowed_identifiers` to `_`. 

Also, it looks like this same (increasingly complex) code is repeated 3 times in the rule. Think we could avoid this duplication?

src/sqlfluff/rules/L057.py
```diff
@@ -53,6 +54,7 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
         self.quoted_identifiers_policy: str
         self.unquoted_identifiers_policy: str
         self.allow_space_in_identifier: bool
+        self.additional_allowed_identifiers: str
```
barrywhart, line 57: These aren't really "identifiers". Rather, they are additional _characters_ allowed in identifiers. Perhaps call it `additional_allowed_identifier_characters` for clarity?

barrywhart (changes requested):

src/sqlfluff/rules/L057.py
tunetheweb: I think I'd prefer to keep the underscores in there. To me they should always be allowed. Petty Python doesn't have an `isalnum` equivalent that include them.

Also means we can skip the slightly more complex additional logic if it's not set (which it isn't by default) which would be a slight performance gain... maybe.

However I have refactored the code to avoid duplicating it. Let me know what you think.

src/sqlfluff/rules/L057.py
tunetheweb: I've gone with `additional_allowed_characters` as it's a rule-specific config specific to this rule, so think it's obvious it's for identifiers.

barrywhart (approved):
Looks good! Thanks for the refactoring -- it is 💯.

tunetheweb:
> Looks good! Thanks for the refactoring -- it is 💯.

I had another go to make it even cleaner. WDYT?

barrywhart:
New version looks good, also. When I work on rule code, I am often unsure whether it's better to have one long function or several short ones. Lately, I tend to prefer several short ones, if it helps divide the code into logical sections. I think that makes it easier for maintainers, as it potentially lets them focus on the relevant area of the code without having to worry about the whole function. 🤷‍♂️

tunetheweb:
Yeah it's a judgement call.

I think short ones are good if we want to abstract away a specific bit of function. With this one here, we were basically abstracting away one line calls so having a function seemed overkill. Also a near-identical `replace` calls was left behind in main function (the BigQuery bit), so again didn't seem right to separate them. Not to mention there was the `if` in the function for the `allow_space_in_identifier` which seemed to pass a parameter for the sake of something that was better handled in the one `if` statement, rather than two.

All in all, I think this is better as just one function as pretty short (and a lot of comments, without which it would be shorter still). But any larger, I agree might want to split it up.

PR #2277, opened 2022-01-10
Redshift CREATE LIBRARY statements

rpr-ableton:

Parse `CREATE LIBRARY` statements in Redshift dialect.

One bit that is currently missing is how _region_ (optional) and _authorization_ (mandatory) can be swapped around in `CreateLibraryStatementSegment`. I couldn't find a clean way to get there :sweat_smile: Any pointers would be appreciated!

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -360,6 +399,42 @@ class CreateExternalTableAsStatementSegment(BaseSegment):
     )
 
 
+@redshift_dialect.segment()
+class CreateLibraryStatementSegment(BaseSegment):
+    """A `CREATE LIBRARY` statement.
+
+    As specified in https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_LIBRARY.html
+    TODO: Region sequence can be either before or after the authorization
```
tunetheweb, line 407: Added comment below on this.

```suggestion
```

src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -360,6 +399,42 @@ class CreateExternalTableAsStatementSegment(BaseSegment):
...
+        Sequence(
+            Ref("AuthorizationSegment", optional=False),
+            Sequence(
+                "REGION",
+                Ref.keyword("AS", optional=True),
+                Ref("QuotedLiteralSegment"),
+                optional=True,
+            ),
+            optional=True,
+        ),
```
tunetheweb, line 434: We don't nice way of doing this so normally just put in `AnyNumberOf` (which also removes the need for the Optional since it allows 0 instances so is optional by default):

```suggestion
        AnyNumberOf(
            Ref("AuthorizationSegment", optional=False),
            Sequence(
                "REGION",
                Ref.keyword("AS", optional=True),
                Ref("QuotedLiteralSegment"),
                optional=True,
            ),
        ),
```

This allows them to be added in any order. It does however also allow two AuthorizationSegments, or two Regions. So not ideal. But the best we have for now.

tunetheweb (changes requested):
Looks good! Added the info on how to workaround the remaining TODO.

rpr-ableton:
Thanks for the help @tunetheweb. I had gone in this direction at first but the conditions seemed too loose. If there's nothing stricter, that'll do just fine :slightly_smiling_face: 

I've applied your recommendations and added a test to reflect it.

test/fixtures/dialects/redshift/redshift_create_library.sql
```diff
@@ -0,0 +1,46 @@
...
+ACCESS_KEY_ID '<access-key-id>'
+SECRET_ACCESS_KEY '<secret-access-key>'
+SESSION_TOKEN '<temporary-token>'
+region 'us-east-1';
+
+create library lib1 language plpythonu
+from 'https://example.com/packages/lib1.0.3.zip';
+
+create or replace library lib1 language plpythonu
+from 'https://example.com/packages/lib1.0.3.zip';
```
tunetheweb, line 46: Add missing newline. BTW if your editor supports `.editorconfig` files (some need a plugin: https://editorconfig.org/#download), then it'll automatically fix things like.

```suggestion
from 'https://example.com/packages/lib1.0.3.zip';

```

tunetheweb (changes requested):
One small thing, then good to merge. Will try commit that, but might need YAML regeneration.

tunetheweb (approved):

PR #2278, opened 2022-01-10
L025 with 'bigquery' dialect: Correctly interpret calling functions with a table as a parameter

### LINKED ISSUE #2030: Incorrect reporting of "Alias is never used in SELECT statement"

barrywhart:
Fixes #2030 

BigQuery already had special code to deal with some ambiguity about which portion of an object reference was the table, schema, etc. To fix this issue (BigQuery functions that take a table as a parameter), we change it again, making it even more "accepting".

The tradeoff of doing it this way (which we were already doing in L026) is that rules using this function to interpret a reference are given things that _may or may not_ be table references. In this case, that works fine. Generally, any rule using this function must be designed so that it behaves reasonably in spite of the ambiguity. (Generally, rules should "fail safe", i.e. don't "fix" code unless they know for sure how to do it safely, and don't warn unless they're sure there's an issue.)

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -708,17 +708,33 @@ class ColumnReferenceSegment(ObjectReferenceSegment):  # type: ignore
     )
 
     def extract_possible_references(self, level):
-        """Extract possible references of a given level."""
+        """Extract possible references of a given level.
+
+        Overrides the parent-class function. BigQuery's support for things like
+        the following
+        - Functions that take a table as a parameter (e.g. TO_JSON_STRING)
+          https://cloud.google.com/bigquery/docs/reference/standard-sql/
+          json_functions#to_json_string
+        - STRUCT
+
+        mean that, without schema information (which SQLFluff does not have),
+        references to data are often ambiguous.
+        """
         level = self._level_to_int(level)
         refs = list(self.iter_raw_references())
         if level == self.ObjectReferenceLevel.SCHEMA.value and len(refs) >= 3:
             return [refs[0]]  # pragma: no cover
         if level == self.ObjectReferenceLevel.TABLE.value:
+            # One part: Could be a table, e.g. TO_JSON_STRING(t)
+            # Two parts: Could be schema+table or table+column.
+            # Three parts: Could be table+column+struct or schema+table+column.
             return refs[:2]
-        if level == self.ObjectReferenceLevel.OBJECT.value and len(refs) >= 3:
+        if (
+            level == self.ObjectReferenceLevel.OBJECT.value and len(refs) >= 3
+        ):  # pragma: no cover
             # Ambiguous case: The object (i.e. column) could be the first or
             # second part, so return both.
-            return [refs[1], refs[2]]  # pragma: no cover
+            return [refs[1], refs[2]]
         return super().extract_possible_references(level)
 
 
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -738,7 +738,7 @@ class ColumnReferenceSegment(ObjectReferenceSegment):  # type: ignore
             # Ambiguous case: The object (i.e. column) could be the first or
             # second part, so return both.
             return [refs[1], refs[2]]
-        return super().extract_possible_references(level)
+        return super().extract_possible_references(level)  # pragma: no cover
 
 
 @bigquery_dialect.segment()
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -717,7 +717,7 @@ class ColumnReferenceSegment(ObjectReferenceSegment):  # type: ignore
           json_functions#to_json_string
         - STRUCT
 
-        mean that, without schema information (which SQLFluff does not have),
+        means that, without schema information (which SQLFluff does not have),
         references to data are often ambiguous.
         """
         level = self._level_to_int(level)
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -726,8 +726,8 @@ class ColumnReferenceSegment(ObjectReferenceSegment):  # type: ignore
             return [refs[0]]  # pragma: no cover
         if level == self.ObjectReferenceLevel.TABLE.value:
             # One part: Could be a table, e.g. TO_JSON_STRING(t)
-            # Two parts: Could be schema+table or table+column.
-            # Three parts: Could be table+column+struct or schema+table+column.
+            # Two parts: Could be dataset+table or table+column.
+            # Three parts: Could be table+column+struct or dataset+table+column.
             return refs[:2]
         if (
             level == self.ObjectReferenceLevel.OBJECT.value and len(refs) >= 3
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -726,9 +726,12 @@ class ColumnReferenceSegment(ObjectReferenceSegment):  # type: ignore
             return [refs[0]]  # pragma: no cover
         if level == self.ObjectReferenceLevel.TABLE.value:
             # One part: Could be a table, e.g. TO_JSON_STRING(t)
-            # Two parts: Could be dataset+table or table+column.
-            # Three parts: Could be table+column+struct or dataset+table+column.
-            return refs[:2]
+            # Two parts: Could be dataset.table or table.column.
+            # Three parts: Could be table.column.struct or dataset.table.column.
+            # Four parts: dataset.table.column.struct
+            # Five parts: project.dataset.table.column.struct
+            # So... return the first 3 parts.
+            return refs[:3]
         if (
             level == self.ObjectReferenceLevel.OBJECT.value and len(refs) >= 3
         ):  # pragma: no cover
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -711,7 +711,7 @@ class ColumnReferenceSegment(ObjectReferenceSegment):  # type: ignore
         """Extract possible references of a given level.
 
         Overrides the parent-class function. BigQuery's support for things like
-        the following
+        the following:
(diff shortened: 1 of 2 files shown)


src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -708,19 +708,33 @@ class ColumnReferenceSegment(ObjectReferenceSegment):  # type: ignore
...
-            return [refs[0], refs[1]]
-        if level == self.ObjectReferenceLevel.OBJECT.value and len(refs) >= 3:
+        if level == self.ObjectReferenceLevel.TABLE.value:
+            # One part: Could be a table, e.g. TO_JSON_STRING(t)
+            # Two parts: Could be schema+table or table+column.
+            # Three parts: Could be table+column+struct or schema+table+column.
+            return refs[:2]
+        if (
+            level == self.ObjectReferenceLevel.OBJECT.value and len(refs) >= 3
+        ):  # pragma: no cover
```
barrywhart, line 734: Had to move the "no cover" because, by removing the ` and len(refs) >= 3` above, no we never hit this `if` statement.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -708,19 +708,33 @@ class ColumnReferenceSegment(ObjectReferenceSegment):  # type: ignore
...
         if level == self.ObjectReferenceLevel.SCHEMA.value and len(refs) >= 3:
             return [refs[0]]  # pragma: no cover
-        if level == self.ObjectReferenceLevel.TABLE.value and len(refs) >= 3:
-            # Ambiguous case: The table could be the first or second part, so
-            # return both.
-            return [refs[0], refs[1]]
-        if level == self.ObjectReferenceLevel.OBJECT.value and len(refs) >= 3:
+        if level == self.ObjectReferenceLevel.TABLE.value:
+            # One part: Could be a table, e.g. TO_JSON_STRING(t)
+            # Two parts: Could be schema+table or table+column.
```
tunetheweb, line 729: I would change schema to dataset. Schema means something else in BigQuery:

```suggestion
            # Two parts: Could be dataset+table or table+column.
```

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -708,19 +708,33 @@ class ColumnReferenceSegment(ObjectReferenceSegment):  # type: ignore
...
+            # Three parts: Could be table+column+struct or schema+table+column.
```
tunetheweb, line 730:
```suggestion
            # Three parts: Could be table+column+struct or dataset+table+column.
```

In theory there could be four parts (dataset.table.column.struct), or even five parts (project.dataset.table.column.struct) in which case the current query won’t pick this up.

Similarly I think it could be several structs (e.g. table.struct.struct).

I wonder is it better to look at first three parts instead of last two? That should always get the table.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -708,19 +708,33 @@ class ColumnReferenceSegment(ObjectReferenceSegment):  # type: ignore
     )
 
     def extract_possible_references(self, level):
-        """Extract possible references of a given level."""
+        """Extract possible references of a given level.
+
+        Overrides the parent-class function. BigQuery's support for things like
+        the following
```
jpy-git, line 714:
```suggestion
        the following:
```

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -708,19 +708,33 @@ class ColumnReferenceSegment(ObjectReferenceSegment):  # type: ignore
...
+        - Functions that take a table as a parameter (e.g. TO_JSON_STRING)
+          https://cloud.google.com/bigquery/docs/reference/standard-sql/
+          json_functions#to_json_string
+        - STRUCT
+
+        mean that, without schema information (which SQLFluff does not have),
```
jpy-git, line 720:
```suggestion
        means that, without schema information (which SQLFluff does not have),
```

jpy-git (approved):

test/fixtures/rules/std_rule_cases/L025.yml
```diff
@@ -102,3 +102,19 @@ test_pass_subselect_uses_alias_3:
...
+    SELECT TO_JSON_STRING(t)
+    FROM my_table
+
+test_bigquery_function_takes_table_parameter:
+  pass_str: |
+    SELECT TO_JSON_STRING(t)
+    FROM my_table AS t
+  configs:
+    core:
+      dialect: bigquery
```
tunetheweb, line 120: If you agree with above, should consider adding more test cases as described above.

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_bigquery.py
barrywhart: Will do

src/sqlfluff/dialects/dialect_bigquery.py
barrywhart: It's actually looking at the first 2, not the last 2. Will change it to the first 3, though. Good catch! 👍

test/fixtures/rules/std_rule_cases/L025.yml
barrywhart: Will do.

src/sqlfluff/dialects/dialect_bigquery.py
tunetheweb: Actually I’m having second thoughts about this now…

This rule only cares about alias’s that are unused. Therefore my suggestions of 4 and 5 aren’t valid as those are full table references and not alias references. So think you should revert back to what you originally had.

In fact is first part sufficient instead of first two parts?

barrywhart:
@tunetheweb: I added a couple of additional test cases, bringing it up to 3 tests. At first, I thought we may want 5 tests, but because this rule works with table _aliases_, the cases where the table is preceded by a dataset and/or project do not apply (because those contain _tables_, not _aliases_). Does this seem reasonable?

tunetheweb:
> @tunetheweb: I added a couple of additional test cases, bringing it up to 3 tests. At first, I thought we may want 5 tests, but because this rule works with table _aliases_, the cases where the table is preceded by a dataset and/or project do not apply (because those contain _tables_, not _aliases_). Does this seem reasonable?

Literally just figured that out myself and commented as such 😁 Does make me wonder why you need first two references though rather than just the first one?

src/sqlfluff/dialects/dialect_bigquery.py
barrywhart: I think your point was valid, in the sense that this function (on the BigQuery dialect side) deals with general object references -- it doesn't know that the caller is only interested in aliases. So I think this is reasonable for now, unless we want to get super precise and distinguish between callers asking for TABLE vs ALIAS. We _could_ do that, but it feels a bit like overengineering at this point, maybe?

tunetheweb: Ah fair point. This function potentially isn’t just for L025. Ok fine to leave as is then with three.

src/sqlfluff/dialects/dialect_bigquery.py
tunetheweb: I don’t get this. By removing part of the if statement how can we make it LESS likely the if statement is hit?

barrywhart:
@tunetheweb:
>Literally just figured that out myself and commented as such 😁 Does make me wonder why you need first two references though rather than just the first one?

If this function knew the caller wanted an alias, not just a table, the first alone would be fine. At this time, there's no way for L025 to request "alias". Instead it asks for table. We could tidy this up if you think it's important. This area of BigQuery is so "loosey goosy" 🦆  that I'm not sure it's worth the additional coding and testing effort/complexity at this time.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -708,19 +708,36 @@ class ColumnReferenceSegment(ObjectReferenceSegment):  # type: ignore
...
+            # So... return the first 3 parts.
+            return refs[:3]
+        if (
+            level == self.ObjectReferenceLevel.OBJECT.value and len(refs) >= 3
+        ):  # pragma: no cover
             # Ambiguous case: The object (i.e. column) could be the first or
             # second part, so return both.
-            return [refs[1], refs[2]]  # pragma: no cover
+            return [refs[1], refs[2]]
         return super().extract_possible_references(level)
```
tunetheweb, line 741: I think this might now need a pragma no cover to fix your coverage issue?

```suggestion
        return super().extract_possible_references(level)  # pragma: no cover
```

src/sqlfluff/dialects/dialect_bigquery.py
barrywhart: Because making the `if` statement less stringent, we are more likely to enter the `if` block, which returns from the function, thus the rest of the function is skipped.

barrywhart: In other words, previously the function only exited early for TABLE and a reference length 3 or greater. Now it exits early for **any** table reference.

tunetheweb: OK but moving the no cover from here:

```py
if if_test1 and test2:
   if_clause # pragma: no cover
other_statement
```

to this:


```py
if (
  if_test1
):  # pragma: no cover
   if_clause
other_statement
```

Doesn't really have any effect. The problem was the `other_statement` was no longer being hit, so that's where the pragma needed to be added. Which you did now in https://github.com/sqlfluff/sqlfluff/pull/2278/commits/90dd808d82cd3518df47dc3428b13f8d1f9fb697

So good to approve now. You could revert this pragma change now before merging if you wanted to remove the need for the bracket and extra lines (which I presume were only added to allow the pragma to fit in?).

tunetheweb (approved):

PR #2279, opened 2022-01-10
L052: Fix case where no preceding segments and mulitline

### LINKED ISSUE #2268: L052: Auto-fix appended extra `semicolon` in `CREATE TABLE` statement when `multiline_newline` is `True`
## Expected Behaviour

* Auto-fix of L052 should move `;` at the end of `CREATE TABLE` statement to the next line when `multiline_newline = True`.

## Observed Behaviour

* Auto-fix printed the following when `sqlfluff fix` is executed:
```
== [sample.sql] FAIL
L:   3 | P:   2 | L052 | Statements must end with a semi-colon.
==== fixing violations ====
```
* Auto-fix for L052 did not move `;` to the next line.
* Auto-fix for L052 appended multiple `;` after the `CREATE TABLE` statement.

## Steps to Reproduce

1. Create `sample.sql` with the following content:
```
CREATE TABLE my_table (
    id INTEGER
);
```
2. Override L052 to force semi colon of multi-line statement to be on the new line with the following override: (You can create a `.sqlfluff` with the content below.)
```
[sqlfluff:rules:L052]
multiline_newline = True
```
3. Run `sqlfluff fix sample.sql -f`

## Dialect

Ansi (Default - Not specific to any dialect.)

## Version

```
(.venv) ~/code/sqlfluff (main) $ sqlfluff fix sample.sql
(.venv) ~/code/sqlfluff (main) $ sqlfluff --version
sqlfluff, version 0.9.1
(.venv) ~/code/sqlfluff (main) $ python --version
Python 3.9.9
```

## Configuration
* Override L052 to force multiline statement's semicolon to always be in a new line.
```
[sqlfluff:rules:L052]
multiline_newline = True
```

## Output of `sqlfluff fix`

```
(.venv) ~/code/sqlfluff (main) $ sqlfluff fix sample.sql
== [sample.sql] FAIL
L:   3 | P:   2 | L052 | Statements must end with a semi-colon.
==== fixing violations ====
Attempting fixes...
Persisting Changes...
== [sample.sql] PASS
All Finished 📜 🎉!
(.venv) ~/code/sqlfluff (main) $
```

jpy-git:

Closes #2268. This change accounts for the possibility that we have to move the semicolon onto a newline and there are no segments between the semicolon and the newline (previous tested statements such as select always had a meta dedent segment that we could anchor on so this wasn't noticed).

No

### Pull Request checklist

  - Other.


src/sqlfluff/rules/L052.py
```diff
@@ -193,21 +193,38 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+                            LintFix.replace(
                                 anchor_segment,
-                                NewlineSegment(),
-                                SymbolSegment(raw=";", type="symbol", name="semicolon"),
-                            ],
-                        ),
-                        LintFix.delete(
-                            context.segment,
-                        ),
-                    ]
```
jpy-git, line 209: Usually anchor_segment != context.segment (since there was always a meta segment between the semicolon and the end of a select statment). Therefore these 2 fixes were anchored to different segments.

In the case highlighted in the linked issue, there were no meta statements between the semicolon and the closing bracket so anchor_segment = context.segment. This makes sense in the context of other cases in the rule and the gives the correct anchoring behaviour for the lint. But in this case we are trying to apply to LintFixes to the same anchor, which due to [this](https://github.com/sqlfluff/sqlfluff/blob/741e814a0bd6c10592d4683af0337f012dea3b1f/src/sqlfluff/core/parser/segments/base.py#L1014-L1016) line of code gets consumed on the replace and therefore the original semicolon never gets deleted, leading to the observed multiple semicolon behaviour.

In a way, the original logic isn't really wrong. The delete fix is just getting thrown away due to that linked TODO.

jpy-git: In the event that `anchor_segment == context.segment` we can apply the fix in one go so added a clause to handle that and avoid the overlapping fix anchors mentioned above.

src/sqlfluff/rules/L052.py
```diff
@@ -193,21 +193,38 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
                     anchor_segment = self._handle_trailing_inline_comments(
                         context, anchor_segment
                     )
-
-                    fixes = [
-                        LintFix.replace(
-                            anchor_segment,
-                            [
+                    fixes = []
+                    if anchor_segment == context.segment:
```
barrywhart, line 197: Nitpick: I think rules are often loose about this, but would it be more precise (and still correct) to instead say:
```suggestion
                    if anchor_segment is context.segment:
```

src/sqlfluff/rules/L052.py
```diff
@@ -193,21 +193,38 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+                                    SymbolSegment(
+                                        raw=";", type="symbol", name="semicolon"
+                                    ),
+                                ],
+                            )
+                        )
+                    else:
+                        fixes.extend(
+                            [
+                                LintFix.replace(
```
barrywhart, line 212: It looks like the two occurrences of `LintFix.replace()` are identical. Could we pull that part of the fix out of the `if` to avoid duplicating that code?

barrywhart (changes requested):

src/sqlfluff/rules/L052.py
jpy-git: They're not the same:
first one:
```python
LintFix.replace(
                                anchor_segment,
                                [
                                    NewlineSegment(),
                                    SymbolSegment(
                                        raw=";", type="symbol", name="semicolon"
                                    ),
                                ],
                            )
```

second one:
```python
LintFix.replace(
                                anchor_segment,
                                [
                                    anchor_segment,
                                    NewlineSegment(),
                                    SymbolSegment(
                                        raw=";", type="symbol", name="semicolon"
                                    ),
                                ],
                            )
```

Second one keeps the anchor segment since it's not the semicolon and shouldn't be overwritten

barrywhart (approved):
Looks good! Sorry for being slow with the follow-up review. 🎉

PR #443, opened 2020-09-24
BigQuery date part units

alanmcruickshank:
Fixes #387

PR #2283, opened 2022-01-11
Remove DATE, DATETIME and TIME from BigQuery DatePart

### LINKED ISSUE #2281: L010: SQLFluff 0.9.1 capitalizes columns named date when used as parameters in DATE_ADD
## Expected Behaviour
When run on a SQL file containing a line like DATE_ADD(date, INTERVAL 5 YEAR), SQLFluff should not prompt or change date to be capitalized because it is a variable name, not a SQL keyword, in this context.

## Observed Behavior
Starting in SQLFluff 0.9.1, for this function (and possibly others) DATE is treated like an SQL keyword rather than a column name.

## Steps to Reproduce
(1). Generate a SQL file of the form:
```sql
SELECT DATE_ADD(date, INTERVAL 5 YEAR) AS display_date
FROM
    model
```
(2). Set BigQuery as the dialect and run sqlfluff lint on this file.

Output is **L:   1 | P:  17 | L010 | Keywords must be upper case.**

## Dialect
BigQuery

## Version
SQLFluff version is 0.9.1, testing in 0.9.0 doesn't produce the same error
Python version is 3.8.1.0.
(I'm using DBT but the same issue occurs when not using dbt as a template).

## Configuration
```
[sqlfluff]
dialect = bigquery

[sqlfluff:rules]
max_line_length = 120
comma_style = leading

[sqlfluff:rules:L010]
capitalisation_policy = upper

[sqlfluff:rules:L030]
capitalisation_policy = upper
```

tunetheweb:

Fixes #2281 

This was added as part of #443 but can'd see why DATE and TIME were added as not part of https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#extract as referenced in the code so think they were added in error.

### Pull Request checklist

  - Other.


test/fixtures/dialects/bigquery/select_datetime.yml
```diff
@@ -26,15 +26,17 @@ file:
...
-              start_bracket: (
-              date_part: DATE
-              keyword: FROM
-              expression:
+            - start_bracket: (
+            - expression:
+                column_reference:
+                  identifier: DATE
+            - keyword: FROM
+            - expression:
```
jpy-git, line 34: @tunetheweb FYI I originally tried this same method but didn't submit because of this.

In this context in `EXTRACT` `DATE` is a vaild `date_part` (or maybe a `keyword`, don't know strictly which as the original approach may have been a bit of a hack) and should therefore be controlled by L010.
`EXTRACT(DATE FROM gmv.order_created_at)` converts the datetime value in the column into a `DATE`. So making this change would just introduce the opposite bug into `DATE` from `EXTRACT`.

Feels like the current `FunctionContentsGrammar` approach is too general to be applied to all functions. Interestingly in [ANSI](https://github.com/sqlfluff/sqlfluff/blob/80be738545d39e578c830b4347af3f44797d3833/src/sqlfluff/dialects/dialect_ansi.py#L3363-L3372) the `DATEADD` function seems to be separated out from the other functions but this is overwritten in bigquery.

tunetheweb: Actually this is not valid BigQuery syntax!

image

Will update the date to do a real datepart.

jpy-git: Cool thanks, let's see what the OP says.

jpy-git (approved):

tunetheweb:
Thanks for the approval @jpy-git . Will hold off merging this for a few days to see if OP comes back. If no answer in a few days then will go ahead and merge.

kdw2126:
Just as an FYI:  There are contexts where DATE should be capitalized within an extract statement such as in the context EXTRACT(DAY FROM DATE '2013-12-25') (from [here](https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#current_date)).  I think this is where the DATE, DATETIME, and TIME references came from because you can define the second part of the EXTRACT statement using a string if you precede with these three keywords.  

Hard to solve my issue while not capitalizing DATE in the first part of an EXTRACT statement.

tunetheweb:
That seems to work fine with this change @kdw2126 :

```
% echo "SELECT EXTRACT(DAY FROM DATE '2013-12-25')" | sqlfluff lint --dialect bigquery -
All Finished 📜 🎉!                                                                                                                      
% echo "SELECT EXTRACT(DAY FROM date '2013-12-25')" | sqlfluff lint --dialect bigquery -
== [stdin] FAIL                                                                                                                          
L:   1 | P:  25 | L010 | Keywords must be consistently upper case.
All Finished 📜 🎉!
```

I've no doubt there will be some examples where it's difficult to say whether it's a column name, or a keyword but luckily this example isn't one of them.

And then we're also back to best practice of not using KEYWORDS for column names where possible (Rule L029). Though admittedly I turn that rule off for my BigQuery queries, particularly as I do tend to use `date` as a column name. But do as I say and not as I do! 😉

kdw2126:
You are absolutely right @tunetheweb -- my mistake.  That will teach me to try to guess the consequences of code changes without testing first thing in the morning.  Appreciate the quick response to my issue and all of your work in this repo!

tunetheweb:
Not heard anything back as to why not to do this so gonna merge this.

Issue #2232, opened 2022-01-04
L10 rule in pyproject.toml is ignored


jan-benisek:
## Expected Behaviour

So I am configuring the linter using `pyproject.toml`. I want all key words UPPERCASE.
```
[sqlfluff.rules.L010]
capitalisation_policy = "upper"
```
When I run lint and fix
```
sqlfluff lint --dialect 'redshift' foo/bar/test.sql
sqlfluff fix --force --dialect 'redshift' foo/bar/test.sql
```

on this file:
```sql
select a+b  as foo,
c as bar from my_table;
```

I expect this:
```sql
SELECT
    c AS bar,
    a + b AS foo
FROM my_table;
```

## Observed Behaviour
The linter gives no warning about rule L10
```
sqlfluff lint --dialect 'redshift' foo/bar/test.sql
sqlfluff fix --force --dialect 'redshift' foo/bar/test.sql
```

And this is the result
```sql
select
    c as bar,
    a + b as foo
from my_table;
```

**IF** I use `.sqlfluff` with this:
```
[sqlfluff:rules:L010]
capitalisation_policy = upper
```
everything works as it should

## Steps to Reproduce
See above

## Dialect
None

## Version
sqlfluff, version 0.9.0

## Configuration
See above

jpy-git:
Pretty sure for `pyproject.toml` you should be using:
```
[tool.sqlfluff.rules.L010]
capitalisation_policy = "upper"
```

the `tool.` prefix is standard for `pyproject.toml` and there's also a note on it in the docs: https://docs.sqlfluff.com/en/stable/configuration.html

jan-benisek:
That worked, thanks!
The documentation is not very clear in this regard. Otherwise great job, starting to really like the library.

tunetheweb:
Any suggestions on how to make the documentation clearer @jan-benisek ?

jan-benisek:
@tunetheweb 
Firstly, I want to say they this is a great tool, the documentation is already better than average and I know that supporting open source is a tough work. I do not criticise, only trying to help you.

I do not know whether this is possible (and it is probably late), but having the same syntax (dots and "tools") for `.sqlfluff` and `pyproject.toml` would be great.

It would be nice to receive an error if certain line in `pyproject.toml` is ignored. In my case, this:
```
[sqlfluff.rules.L010]
capitalisation_policy = "upper"
```
Gave no error, which added to the confusion.

[...]

PR #2291, opened 2022-01-11
Configuration doc: add rule configuration section

juhoautio:

At the moment the configuration documentation doesn't explain how rules can be configured, expect for indirectly showing it as part of the [Default Configuration](https://docs.sqlfluff.com/en/stable/configuration.html#default-configuration). There is also nothing about `.sqlfluff` config in the [Rules Reference](https://docs.sqlfluff.com/en/stable/rules.html) (and it should stay that way, in my opinion).

It seems that rule configuration deserves a section of its own in this doc. So proposing to add it in this PR.

This was something that I struggled with a bit when getting started with sqlfluff: I wasn't sure where to find what are the allowed values for the `capitalisation_policy` of various rules (answer: [Rules Reference](https://docs.sqlfluff.com/en/stable/rules.html)). This piece of documentation should make it easier to find.

No

### Pull Request checklist

I built the docs locally as documented in https://github.com/sqlfluff/sqlfluff/tree/main/docs#sqlfluff---generating-the-document-website.

  - Other.

### Screenshot

Here are screenshots from the rendered page:

----

Näyttökuva 2022-1-19 kello 0 16 05

Näyttökuva 2022-1-19 kello 0 09 06


docs/README.md
```diff
@@ -3,7 +3,7 @@
 You can run the following steps to generate the documentation website:
 
 ```
-tox -e docbuild, doclint
+tox -e docbuild,doclinting
```
juhoautio, line 6: Having a space there was a bug, causing only docbuild to be run.

```
tox -e docbuild,doclint
```
->
```
ERROR: unknown environment 'doclint'
```
So also fixed that -> doclinting.

docs/source/configuration.rst
```diff
@@ -88,6 +88,37 @@ in projects which utilise a lot of complicated templating.
 
 .. _templateconfig:
 
+Rule Configuration
+------------------
+
+Rules can be configured with the `.sqlfluff` config files.
+
+Common rule configurations can be set in the *[sqlfluff:rules]* section.
```
juhoautio, line 96: In the code the common section is called "generic":

https://github.com/sqlfluff/sqlfluff/blob/7ffea4670c384f575dd7a59ed31da0b93ae8f800/src/sqlfluff/core/rules/base.py#L951-L954

But I'm leaning towards "common" as the user-facing term to use. Please advise :)

juhoautio:
DBT Plugin test failed with "Read timed out". I don't think that has anything to do with my changes.

docs/source/configuration.rst
tunetheweb: It’s called common in the default config too:
https://github.com/sqlfluff/sqlfluff/blob/7ffea4670c384f575dd7a59ed31da0b93ae8f800/src/sqlfluff/core/default_config.cfg#L37-L38

tunetheweb:
> DBT Plugin test failed with "Read timed out". I don't think that has anything to do with my changes.

Kicked it off again.

juhoautio:
Thanks, all checks passed now so ready for review.

docs/source/configuration.rst
```diff
@@ -88,6 +88,37 @@ in projects which utilise a lot of complicated templating.
...
+
+For example:
+
+.. code-block:: cfg
+
+    [sqlfluff:rules]
+    unquoted_identifiers_policy = all
+    max_line_length = 88
```
tunetheweb, line 104: I think it would be worth including all the common ones here to make it more obvious which ones go in this section:

```suggestion
    # Some rules can be configured directly from the config common to other rules.
   [sqlfluff:rules]
   tab_space_size = 4
   max_line_length = 80
   indent_unit = space
   comma_style = trailing
   allow_scalar = True
   single_table_references = consistent
   unquoted_identifiers_policy = all
```

docs/source/configuration.rst
```diff
@@ -88,6 +88,37 @@ in projects which utilise a lot of complicated templating.
...
+
+Rule specific configurations are set in the *[sqlfluff:rules:\*]* sections.
```
tunetheweb, line 106: I find that syntax a little confusing to read. What about this:

```suggestion
Rule specific configurations are set in subsections under each rule id.
```

docs/source/configuration.rst
```diff
@@ -88,6 +88,37 @@ in projects which utilise a lot of complicated templating.
...
+
+    [sqlfluff:rules:L010]  # Keywords
+    capitalisation_policy = upper
+
+All possible options for rule sections are documented in :ref:`ruleref`.
+
+For an overview of the most common rule configurations that you may want to
+tweak, see `Default Configuration`_ (and use :ref:`ruleref` to find the
+available alternatives).
+
```
tunetheweb, line 121: Could we add a section on how to disable rules? Both by explaining the `rules` and `exclude_rules` options in the core section as shown in`configuration.html#default-configuration` (maybe with some examples) and also by referencing inline ignore options with a link to `rules.html#inline-ignoring-errors`

tunetheweb (changes requested):
This looks great. I've a few suggestions to make it more comprehensive, and if you do those then I think we can also say this closes #1938

Also might be worth having a read of https://github.com/sqlfluff/sqlfluff/issues/2232 to see if any other improvements we can make in this PR to address some of the issues raised there? I've had another quick scan and think most of the issues raised about rules are covered here but second pair of eyes couldn't hurt!

docs/source/configuration.rst
juhoautio: I committed the suggestion, but in what sense are those "all the common ones"?

docs/source/configuration.rst
juhoautio: Is it just me or is it confusing to se that the subsections are "under" rule ids? Would this work: "..set in rule specific subsections"?

docs/source/configuration.rst
tunetheweb: I got them from the default config: https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/core/default_config.cfg

As everything needs to be added to that AFAIK I think that is them all

docs/source/configuration.rst
tunetheweb: Yup that works for me.

docs/source/configuration.rst
juhoautio: Unfortunately that means that it can be easy to miss adding it here if something new is added. But I suppose we don't promise here that the list would be comprehensive. It's just an example.

docs/source/configuration.rst
```diff
@@ -88,6 +109,79 @@ in projects which utilise a lot of complicated templating.
...
+For example, to enable :class:`L010 <sqlfluff.core.rules.Rule_L027>`:
+
+.. code-block:: cfg
+
+    [sqlfluff]
+    rules = L027
+
+^ **TODO** does this make sense only if overriding an exclusion from a higher level
+configuration? Or what does this conf mean? How is it interpreted if
+:code:`exclude_rules = None` but :code:`exclude_rules != None`?
```
juhoautio, line 173: Do you know how this is? If not, I could try to look it up but not feeling super confident :)

docs/source/configuration.rst
tunetheweb: Yeah we’ll try to keep it in sync but may will miss something. Still I think it’s good to start with what we know now.

docs/source/configuration.rst
```diff
@@ -88,6 +109,79 @@ in projects which utilise a lot of complicated templating.
...
+
+Additionally, some rules have a special :code:`force_enable` configuration
+option, which allows to enable the given rule even for dialects where it is
+disabled by default. The rules that support this can be found in the
+:ref:`ruleref`.
+
+The default values can be seen in `Default Configuration`_.
+
+See also: :ref:`inline_ignoring_errors` and `Ignoring Errors & Files`_.
+(**TODO** aren't those two pretty much overlapping?)
```
juhoautio, line 183: This is strange. Is it ok to just link to both at this point?

docs/source/configuration.rst
```diff
@@ -88,6 +109,79 @@ in projects which utilise a lot of complicated templating.
...
+
```
juhoautio, line 184: Should we also cross link back to the CLI args section from here?

docs/source/configuration.rst
tunetheweb: Someone looked this up a few days ago and exclude takes priority over rules if that’s what you’re asking.

I think it only makes sense to either to explicitly include some rules, or to include all and explicitly include some of them. Mixing the two doesn’t make sense IMHO.

juhoautio: > exclude takes priority over rules

Oh. So then `rules` only makes sense to be set if `exclude_rules = None`? If `exclude_rules` is set to any value, `rules` will be entirely ignored?

juhoautio: Is there a concept like this though: "overriding an exclusion from a higher level
configuration"? And is it possible to have `.sqlfluff` files in folder hierarchy that overwrite some particular configurations? If yes, then I could imagine `rules` to be set on highest level, but then one could exclude some of those rules with `exclude_rules` in the scope of some subdirectory. Or am I just imagining something that is not even a feature? :)

tunetheweb: Ah that’s true. Yes you could do that.

docs/source/configuration.rst
tunetheweb: Oh I don't think we should duplicate this content. One should just be a link to the other.

Not sure where is best lives? Under configuration or under rules? But since it's under rules already let's just add the link here and remove all the text.

tunetheweb: Oh it's already duplicated in both! Never noticed that before!

OK let's leave as is. I think just the one link to the one on this page is fine. Might be worth doing a review to ensure they are both saying the same thing and sync up if not.

docs/source/configuration.rst
tunetheweb: I don't think you need to cross link here. We're not cross linking to the configuration file for `rules` and `exclude_rules` so why cross link to CLI which has the same?

docs/source/configuration.rst
```diff
@@ -88,6 +109,79 @@ in projects which utilise a lot of complicated templating.
...
+For example, to enable :class:`L010 <sqlfluff.core.rules.Rule_L027>`:
+
+.. code-block:: cfg
+
+    [sqlfluff]
+    rules = L027
+
+^ **TODO** does this make sense only if overriding an exclusion from a higher
+level configuration? Or what does this conf mean? How is it interpreted if
+:code:`exclude_rules = None` but :code:`exclude_rules != None`?
```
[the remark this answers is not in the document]
tunetheweb: Or do you mean like this?:

```
Top-level folder .sqlfluffini: exclude_rules = None
      Sub-folder .sqlfluffini: exclude_rules != None
```

In that case the subfolder will override the top level folder config and for the queries in that solder folder they will exclude some rules. This can be handy if you know a folder can't handle a certain rule, but the rest of the project can.

Note also it's overridden completely as far as I know.

So, with this for example:

```
Top level folder .sqlfluffini: exclude_rules = L001`
      Sub Folder .sqlfluffini: exclude_rules = L002`
```

Queries in the sub folder would only exclude `L002` and would start to lint for `L001`. It would NOT exclude both because the closer config overrides, but is not cumulative.

tunetheweb:
Would love to get this merged @juhoautio if you have some time to address the feedback?

juhoautio:
@tunetheweb thanks for reminding, I'll work on this.

docs/source/configuration.rst
juhoautio: I was thinking, this has now become the main place to read about "how to enable or disable rules". For completeness, it could be worth pointing to the fact that rules can be also controlled via CLI args. That could be the best option in some situation. But of course it is almost obvious that user would check what `sqlfluff lint --help` has to offer. I don't have any real opinion on what should be done. Adding links everywhere can of lead into a bit bloated docs, yes. So we live without linking to CLI args from here?

docs/source/configuration.rst
```diff
@@ -88,6 +109,83 @@ in projects which utilise a lot of complicated templating.
...
+:ref:`ruleref`.
+
+The default values can be seen in `Default Configuration`_.
+
+See also: `Ignoring Errors & Files`_.
+
+..
+    TODO ^ There is also :ref:`inline_ignoring_errors` with overlapping
+    content. Might be worth reviewing to ensure they are both saying the same
+    thing and sync up if not.
```
juhoautio, line 187: I thought I'd leave this as a hidden comment:

Näyttökuva 2022-1-19 kello 0 09 06

Or it can be removed & create an issue about it?

docs/source/configuration.rst
tunetheweb: Yeah I think leave it off here (for the bloated doc reason you give). It's not really the recommended way to ignore errors or include/exclude rules IMHO, but more a short hand for quite testing of only specific rules for that particular run.

docs/source/configuration.rst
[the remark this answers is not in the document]
juhoautio: I replaced this paragraph with:

> If both :code:`exclude_rules` and :code:`rules` have non-empty value, then the
excluded rules are removed from the rules list. This allows for example
enabling common rules on top level but excluding some on subdirectory level.

I would say this is the reasonable level of detail. Thoughts?

docs/source/configuration.rst
tunetheweb: That looks good to me. I say include it.

docs/source/configuration.rst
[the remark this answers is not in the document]
tunetheweb: That works for me!

juhoautio:
Looks like everything so far is resolved. Anything else?

I updated the PR desc with fresh screenshots (is there any better way?).

tunetheweb:
Minor gripe, but this looks a little weird:

image

You see the way the `[sqlfluff:rules]` is displaying differently to the `[sqlfluff:rules:L010]  # Keywords`? Just looks a little off. Turns out this is due to the `# Keywords` comment. What do you think about moving that to the line below? And also cleaning up the default config similarly? To be honest the comments in the default config could do with a little expanding:

```
operator_new_lines = after

[sqlfluff:rules:L011]  # Aliasing
aliasing = explicit

[sqlfluff:rules:L012]  # Aliasing
aliasing = explicit
```

Would be nice to explain different between L007 and L010 and similarly L011 and L012.

Or shall we just merge this and fix that in a separate PR?

juhoautio:
I suppose it's easier if you could make those changes in a separate PR, especially considering that you may expand the comments in default conf. Is that ok?

Btw, isn't it basically a syntax highlighting bug though? :)

tunetheweb:
> Btw, isn't it basically a syntax highlighting bug though? :)

True! But you work with what you've got!

OK lets merge this. Thanks very much of this - improves the docs a lot!

tunetheweb (approved):

juhoautio:
Many thanks for reviewing!

tunetheweb:
FYI I raise #2347 for the things I raised above, plus a good few other things!

PR #2300, opened 2022-01-13
Fix #1292: nocolor and verbose can work in config files

### LINKED ISSUE #1292: verbose and nocolor options are ignored from config files
## Expected Behaviour

The [default configuration file](https://docs.sqlfluff.com/en/stable/configuration.html#default-configuration) has the following entries:

```
verbose = 0
nocolor = False
```

This suggests these options can be configured in a config file but this does not appear to be the case.

## Observed Behaviour

Changing those defaults does not change the verbosity nor whether color printing is used.

## Steps to Reproduce

- Create a .sqlfluff files with the following set:

```
[sqlfluff]
verbose = 1
nocolor = True
```

- Lint files.
- The verbosity is not used and the default color options are used.

## Version

Latest (0.6.3) and Python 3.8

## Configuration

```
[sqlfluff]
verbose = 1
nocolor = True
```

## More information

If you add the following to `def common_options` for both these config in [`src/sqlfull/cli/commands.py`](https://github.com/sqlfluff/sqlfluff/blob/00fca3a3af9d48902e35a6de8fd60aff32de2035/src/sqlfluff/cli/commands.py#L104):
```
        default=None,
```

Then it looks like these configs can be overridden by a config file in the current directory. Without this, it looks like the command line options are used (even if not set).

Is there a reason these are not set with this `default=None` option?

However even with this change, it looks like it's only honoured for the config file in the current directory that sqlfluff is run from. And if you specify a path (e.g. `sqlfluff lint /some/other/path`) then the configs are not used for these two settings (even though other settings are used for linting/parsing/fixing). I'm not sure if this is due in part due to the issue discussed in #556 ?

I'm not sure it makes sense to override some of these options anyway - e.g. does it make sense to have different verbosity or nocolor settings for different paths or should it be a global setting?

Maybe, rather than looking to fix this, these two options should these just be made command line options and removed from the example default config file so people don't presume they can override this (when currently they can't)?

cympfh:

- Fix #1292

The cause of this bug is that some options in Click (`--nocolor` and `--verbose`) don't have default values when not specified as CLI options. In original code, only `None` must be default value. This change should define default values for each options other than `None`.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -7,6 +7,7 @@ import logging
 import time
 from logging import LogRecord
 from typing import (
+    Any,
     Callable,
     Tuple,
     NoReturn,
@@ -299,8 +300,19 @@ def get_config(
                 )
             )
             sys.exit(66)
+
     # Instantiate a config object (filtering out the nulls)
-    overrides = {k: kwargs[k] for k in kwargs if kwargs[k] is not None}
+    def null(key: str, val: Any) -> bool:
+        if val is None:
+            return True
+        if key == "nocolor" and val is False:
+            return True
+        if key == "verbose" and val == 0:
+            return True
+        return False
+
+    overrides = {k: kwargs[k] for k in kwargs if not null(k, kwargs[k])}
+
     try:
         return FluffConfig.from_root(
             extra_config_path=extra_config_path,
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -7,7 +7,6 @@ import logging
 import time
 from logging import LogRecord
 from typing import (
-    Any,
     Callable,
     Tuple,
     NoReturn,
@@ -151,6 +150,7 @@ def common_options(f: Callable) -> Callable:
         "-v",
         "--verbose",
         count=True,
+        default=None,
         help=(
             "Verbosity, how detailed should the output be. This is *stackable*, so `-vv`"
             " is more verbose than `-v`. For the most verbose option try `-vvvv` or `-vvvvv`."
@@ -160,6 +160,7 @@ def common_options(f: Callable) -> Callable:
         "-n",
         "--nocolor",
         is_flag=True,
+        default=None,
         help="No color - if this is set then the output will be without ANSI color codes.",
     )(f)
 
@@ -300,19 +301,8 @@ def get_config(
                 )
             )
             sys.exit(66)
-
     # Instantiate a config object (filtering out the nulls)
-    def null(key: str, val: Any) -> bool:
-        if val is None:
-            return True
-        if key == "nocolor" and val is False:
-            return True
-        if key == "verbose" and val == 0:
-            return True
-        return False
-
-    overrides = {k: kwargs[k] for k in kwargs if not null(k, kwargs[k])}
-
+    overrides = {k: kwargs[k] for k in kwargs if kwargs[k] is not None}
     try:
         return FluffConfig.from_root(
             extra_config_path=extra_config_path,
--- a/test/cli/commands_test.py
+++ b/test/cli/commands_test.py
@@ -928,8 +928,8 @@ def test_cli_get_config():
     config = get_config(
         "test/fixtures/config/toml/pyproject.toml",
         True,
-        nocolor=False,
-        verbose=0,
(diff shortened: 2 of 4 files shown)


src/sqlfluff/cli/commands.py
```diff
@@ -299,8 +300,19 @@ def get_config(
...
+    def null(key: str, val: Any) -> bool:
+        if val is None:
+            return True
+        if key == "nocolor" and val is False:
+            return True
+        if key == "verbose" and val == 0:
+            return True
+        return False
+
+    overrides = {k: kwargs[k] for k in kwargs if not null(k, kwargs[k])}
```
jpy-git, line 314: The reason for using `None` as a default value in python is that it is a good way to indicate that a user has not specified an argument versus them explicitly specifying the default value.

If you read your test case, you explicitly pass in `nocolor=False` and receive out `nocolor==True` which is odd behaviour for a function (even if it is doing the correct thing from the CLI perspective):
```python
def test_cli_get_config():
    config = get_config(
        "test/fixtures/config/toml/pyproject.toml",
        True,
        nocolor=False,
        verbose=0,
    )
    assert config.get("nocolor") is True
    assert config.get("verbose") == 2
```

What should happen is if you pass in `nocolor=True` you receive out `True`, if you pass in `nocolor=False` you receive out `False`, and if you pass in `None` (the default) you receive out the value from your config (or the default config if not specified in user config).

test/cli/commands_test.py
```diff
@@ -924,6 +924,17 @@ def test_cli_disable_noqa_flag():
...
 
+def test_cli_get_config():
+    config = get_config(
+        "test/fixtures/config/toml/pyproject.toml",
+        True,
+        nocolor=False,
+        verbose=0,
+    )
+    assert config.get("nocolor") is True
+    assert config.get("verbose") == 2
```
jpy-git, line 935:
```suggestion
def test_cli_get_config():
    config = get_config(
        "test/fixtures/config/toml/pyproject.toml",
        True,
        nocolor=None,
        verbose=None,
    )
    assert config.get("nocolor") is True
    assert config.get("verbose") == 2
```
See previous review comment for reason

src/sqlfluff/cli/commands.py
jpy-git: Can we reset this section to as it was before and instead just simply add `default=None` to the CLI options:
```python
def common_options(f: Callable) -> Callable:
    """Add common options to commands via a decorator.

    These are applied to all of the cli commands.
    """
    f = click.version_option()(f)
    f = click.option(
        "-v",
        "--verbose",
        count=True,
        default=None,
        help=(
            "Verbosity, how detailed should the output be. This is *stackable*, so `-vv`"
            " is more verbose than `-v`. For the most verbose option try `-vvvv` or `-vvvvv`."
        ),
    )(f)
    f = click.option(
        "-n",
        "--nocolor",
        is_flag=True,
        default=None,
        help="No color - if this is set then the output will be without ANSI color codes.",
    )(f)

    return f
```

jpy-git (commented):
@cympfh Thanks for spotting the cause of this issue! 😄

Small tweak to make in terms of implementation but otherwise great

cympfh:
Thank you for your comments. Yeah, `default=None` works well!

test/cli/commands_test.py
```diff
@@ -924,6 +924,17 @@ def test_cli_disable_noqa_flag():
     assert r"L:   5 | P:  11 | L010 |" in raw_output
 
 
+def test_cli_get_config():
```
jpy-git, line 927:
```suggestion
def test_cli_get_default_config():
    """`nocolor` and `verbose` values loaded from config if not specified via CLI."""
```
Linting is failing since we need a docstring here.

(for future reference would recommend using the pre-commit hooks to help catch these at commit time: `tox -e pre-commit -- install` 😄)

jpy-git:
@cympfh FYI your PR is correct, the reason it's failing the unit test atm is that we weren't previously allowing `pyproject.toml` files in the `extra_config_path` parameter so it's being ignored. I've create a PR (#2305) to allow this and will unblock this PR once merged in 😄

jpy-git (approved):
All good, thanks @cympfh 😄 LGTM 🚀 ✨

PR #2307, opened 2022-01-14
Redshift dialect: add COPY and UNLOAD statements

rpr-ableton:

**Features**:
* Ensure that SQLFluff can parse `UNLOAD` statements as defined in documentation here: https://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html
* Ensure that SQLFluff can parse `COPY` statements as defined in documentation here: https://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html & https://docs.aws.amazon.com/redshift/latest/dg/r_COPY-parameters.html
* Update keywords
* Tests for both new statements (from the official AWS documentation)

Both statements are quite long. I've tried to break out some of logic into their own statements where it seemed appropriate but still not the cleanest state :sweat_smile: The tricky part is that most of the statement components can be in any order relatively to one another.

Let me know if there is anything to do in order to improve it!

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -99,6 +99,77 @@ class ColumnEncodingSegment(BaseSegment):
...
+@redshift_dialect.segment()
+class CompressionSegment(BaseSegment):
+    """Compression segment.
+
+    Indicates file compression.
+
+    As specified by: https://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-file-compression.html
+    """
+
+    type = "compression_segment"
```
barrywhart, line 111: Should we consider calling it `compression_type_segment` or just `compression_type`, since this segments specifies which type of compression to use?

barrywhart (approved):
One small question/suggestion. Looks good overall. 👍

rpr-ableton:
Good shout, I've updated the segment to reflect your suggestion.

Thanks for the review!

PR #2315, opened 2022-01-15
Full REFERENCES grammar for CREATE TABLE statement

### LINKED ISSUE #2237: Postgres dialect: parse error for "on delete" and "on update" in column constraint
## Expected Behaviour
Sqlfluff parses column constraints with "on delete" and "on update" clause.

## Observed Behaviour
Sqlfluff throws a parse error
```
                       | TABLE ktw_account.ktw_account_bin...'
```

## Steps to Reproduce
Parse this file:
```
CREATE TABLE ktw_account.ktw_account_binding (
  ppid VARCHAR(255) NOT NULL,
  provider PLATFORM NOT NULL,
  ktw_id VARCHAR(32) NOT NULL REFERENCES ktw_account.ktw_account_core(ktw_id)
  ON DELETE CASCADE ON UPDATE CASCADE,
  PRIMARY KEY (ppid, provider)
);

```

## Dialect
postgres
## Version
sqlfluff, version 0.9.0

## Configuration
```
[sqlfluff]
dialect = postgres

[sqlfluff:rules:L016]
ignore_comment_lines = True

```

jpy-git:

Closes #2237. The `REFERENCES` grammar for `ColumnConstraintSegment`s and `TableConstraintSegment`s in a `CREATE TABLE` statement.
I know this applies to:
- Postgres: https://www.postgresql.org/docs/14/sql-createtable.html (search REFERENCES)
- MySQL: https://dev.mysql.com/doc/refman/8.0/en/create-table.html (search reference_definition)

MySQL uses the ANSI `CREATE TABLE` definition and Postgres has it's own `CREATE TABLE` definition so I've updated the relevant places and added corresponding tests.

No

### Pull Request checklist

  - Other.


Zimle:
@jpy-git : I noticed that in your ReferenceDefinitionGrammar, sqlfuff allows any repetition of ON DELETE and ON UPDATE statement, e.g.
```sql
CREATE TABLE ktw_account_binding (
    ktw_id VARCHAR(32) NOT NULL REFERENCES ref_table(bla)
    ON DELETE RESTRICT ON DELETE CASCADE
);
```
would be okay for sqlfluff, whereas postgres is not okay with it. I am not sure about the scope of sqlfluff in the sense of: Should sqlfluff throw an parse error if the statement is invalid in the target language?
An alternative would be more lengthy:
```python
                OneOf(
                    Sequence(
                        Sequence(
                            "ON",
                            "DELETE",
                            Ref("ReferentialActionSegment"),
                            optional=True,
                        ),
                        Sequence(
                            "ON",
                            "UPDATE",
                            Ref("ReferentialActionSegment"),
                            optional=True,
                        ),
                    ),
                    Sequence(
                        Sequence(
                            "ON",
                            "UPDATE",
                            Ref("ReferentialActionSegment"),
                            optional=True,
                        ),  
                        Sequence(
                            "ON",
                            "DELETE",
                            Ref("ReferentialActionSegment"),
                            optional=True,
                        ),
                    ),
                    optional=True,
                ),
```
which allows either no UPDATE/DELETE, one of them or UPDATE ON foo DELETE ON bar or DELETE ON foo UPDAETE ON
bar.

jpy-git:
@Zimle this is a known limitation of sqlfluff atm and it's previously been decided on to use `AnyNumberOf` in these situations.

There's been talk of introducing an `AnySetOf` grammar that allows the segments in any order but a max of 1 time for each segment, which would be ideal.

In the meantime we should stick with the consistent agreed upon approach though.

jpy-git:
As described above we should stick with `AnyNumberOf` for this PR. I've created a follow up PR to implement an `AnySetOf` grammar that builds on this PR. So once this PR is merged in we can review that 👍

Zimle:
@jpy-git Many thanks for the info! I really appreciate the work that you are doing and I dont mean to be picky! But AnySetOf would also accept the non-parsable (at least postgres) term ```ON DELETE RESTRICT ON DELETE CASCADE```, as ```ON DELETE RESTRICT``` and ```ON DELETE CASCADE``` are different. For me, that's fine, but I just wanted to highlight the isssue. Sorry if I am mistaken or annoying.

jpy-git:
@Zimle hard to read if that's a question or a statement?

But for the sake of clarity, `AnySetOf` would not accept `ON DELETE RESTRICT ON DELETE CASCADE` because that's two lots of
```python
Sequence(
    "ON",
    "DELETE",
    Ref("ReferentialActionGrammar"),
)
```
which is the correct behaviour.

I've already done the work to implement it in jpy-git/sqlfluff#6

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -477,14 +477,43 @@
...
-    TableConstraintReferenceOptionGrammar=OneOf(
+    ReferentialActionGrammar=OneOf(
         "RESTRICT",
         "CASCADE",
         Sequence("SET", "NULL"),
         Sequence("NO", "ACTION"),
         Sequence("SET", "DEFAULT"),
     ),
     DropBehaviorGrammar=OneOf("RESTRICT", "CASCADE", optional=True),
+    ReferenceDefinitionGrammar=Sequence(
```
barrywhart, line 488: Think it'd be worth making this a segment instead? It could make future rules easier to write.

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -477,14 +477,43 @@
...
+        "REFERENCES",
+        Ref("TableReferenceSegment"),
+        # Foreign columns making up FOREIGN KEY constraint
+        Ref("BracketedColumnReferenceListGrammar", optional=True),
+        Sequence(
```
barrywhart, line 493: Similar question as above -- wondering whether these optional parts (`MATCH ...` and `ON ...`) would be better as separate segments.

barrywhart (approved):
Looks good. Possible question(s) about making more use of segments so the parse tree is easier to navigate. (I'm thinking of #2213, which helped simplify the `CASE` rules.)

src/sqlfluff/dialects/dialect_ansi.py
jpy-git: Hmm, I can't see these specific segments being referenced in a rule? `CASE`/`WHEN`/`THEN`/`ELSE` is pretty common, whereas this is a bit niche. I guess we could move things like this to a separate segment but feels like it might be overkill (at least for now).

Happy for a follow up PR to make them into segments if you want though

src/sqlfluff/dialects/dialect_ansi.py
jpy-git: Same thoughts as above here

src/sqlfluff/dialects/dialect_ansi.py
barrywhart: No worries, just thinking aloud. And yes, almost all the rules are focused on queries or DML.

barrywhart:
@jpy-git: Good to merge!

PR #2317, opened 2022-01-15
Fix Spark numeric literals

jpy-git:

Spark3 Numeric literals are a little more complicated than other dialects as they allow additional characters to specify the data type of the literal.
This PR adds the complete Spark3 numeric literal implementation ([docs](https://spark.apache.org/docs/latest/sql-ref-literals.html#numeric-literal)) and the relevant tests.
I've also provided a Pattern breakdown to explain what the regex is doing.

Side Notes:
- Hive allows similar (but slightly different) characters to specifiy data types, so I will follow up in a separate PR to implement that.
- Noticed that ANSI regex can't handle fractional numeric literals with a trailing `.` (e.g. `123.`). Not sure how many other dialects can handle this, but seems worth checking them to see if we should allow this in ANSI definition as well.

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -60,6 +60,42 @@
         # Ex: select `delimited `` with escaped` from `just delimited`
         # https://spark.apache.org/docs/latest/sql-ref-identifier.html#delimited-identifier
         RegexLexer("back_quote", r"`([^`]|``)*`", CodeSegment),
+        # Numeric literal matches integers, decimals, and exponential formats.
+        # https://spark.apache.org/docs/latest/sql-ref-literals.html#numeric-literal
+        # Pattern breakdown:
```
barrywhart, line 65: Nice work! I'm curious, could we potentially use Python "verbose regular expressions"? I have never used them, but IIUC, it would let us write the actual code in something similar to the format of your comments here.

https://www.geeksforgeeks.org/verbose-in-python-regex/

barrywhart (approved):

src/sqlfluff/dialects/dialect_spark3.py
jpy-git: yeah good shout, I've never used the flag before but will experiment 👍

jpy-git:
@barrywhart So I checked the `regex.VERBOSE` flag and it made nearly all the tests fail. I think it's just because we have lot's of unescaped `#`s and spaces in the lexers atm so I'm going to merge this PR as is and investigate that separately in the interest of keeping the diffs sensible 👍

barrywhart:
@jpy-git: Makes sense. To be clear, I was thinking maybe we could try it on one (or a handful) of the regular expressions, but I guess we'd have to add a parameter to do that. No big deal overall, but could be a minor readability improvement.

PR #2341, opened 2022-01-18
L059: Unnecessary quoted identifier

### LINKED ISSUE #1995: Restrict quoted alias
Would be great have a new rule to restrict/identified quoted aliases for column names.

when a view or table is created using quotes it needs to be referenced exactly with the name it was created respecting upper and lower case. for example, if we create the table 

`CREATE TABLE TEST AS SELECT 'aaaaa' AS "column1"`

we can query from this table using 

`SELECT "column1" FROM TEST`

but the following scripts will return an Invalid Identifier error 

```
SELECT column1 FROM TEST
SELECT COLUMN1 FROM TEST
```

jpy-git:

Closes #1995. Adds a new rule to lint and fix unnecessarily quoted identifiers.

These are quoted identifiers for which their contents are perfectly valid unquoted identifiers and aren't otherwise reserved keywords.

No

### Pull Request checklist

  - Other.


tunetheweb:
This LGTM. Just running it across a few of my code bases to see if it flags anything weird and then can approve.

tunetheweb (approved):
Very nice - flagged two real instances in my big code base, with no false positives.

PR #2349, opened 2022-01-19
Postgres: INSERT INTO table alias

jpy-git:

Postgres INSERT INTO statement allows the table to be aliased. I couldn't use the standard alias expression since the AS is non-optional in this case so created another segment with the alias_expression type.

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -2989,6 +2989,20 @@ class DropTriggerStatementSegment(BaseSegment):
...
+    """A reference to an object with an `AS` clause.
+
+    AS is not optional in this segment.
+    """
+
+    type = "alias_expression"
+    match_grammar = Sequence(
+        "AS",
+        Ref("SingleIdentifierGrammar"),
+    )
```
tunetheweb, line 3003: There is nothing in this to imply it's limited to `INSERT INTO` (except the segment name),and it seems fairly simple. Does it really need to be a separate segment?

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -3004,6 +3018,7 @@ class InsertStatementSegment(BaseSegment):
         "INSERT",
         "INTO",
         Ref("TableReferenceSegment"),
+        Ref("InsertIntoAliasExpressionSegment", optional=True),
```
tunetheweb, line 3021: Why not just this?:

```suggestion
        Sequence("AS", Ref("SingleIdentifierGrammar"), optional=True),
```

jpy-git: I was going to just do this, but the motivation for using a separate segment with the `alias_expression` type is that I think rules that analyse tables and aliases rely on it being a separate segment with the  `alias_expression` type in order to detect it.

jpy-git: And the existing `AliasExpressionSegment` segment was too permissive in this situation and led to parsing errors.

jpy-git: Can you think of a better name than `InsertIntoAliasExpressionSegment` as that might help? 😅

tunetheweb: That makes sense, and see you did comment to that effect in the PR description but I missed it.

Fancy adding a little comment to that effect to the `InsertIntoAliasExpressionSegment ` class in case anyone else wonders why this was done to in the future?

Would be nice if Sequence took and optional type argument (I presume it doesn't?) but to be honest if typing something then that's probably reason enough make it a separate segment.

tunetheweb: `AsAliasExpressionSegment`? That would also allow it to be used elsewhere if needed.

jpy-git: I've updated the segment name and description 😄

tunetheweb (approved):

PR #2355, opened 2022-01-19
Add support for Jinja import and include

### LINKED ISSUE #829: Are there plans to support Jinja `include` and `from` statements?
Currently, when trying to lint an sql file with one of these jinja statements:

```jinja
{% from 'my_macros.sql' import a_macro %}
```
or 
```jinja
{% include 'file.sql' %}
```
it will fail with the message: "no loader for this environment specified".

Unless I'm missing something, this seems to be caused by the use of a SandboxedEnvironment with, as the message says, no loader to load either _my_macros.sql_ or _file.sql_.

Are there any plans to support these? Or maybe I'm doing something wrong?

Thank you.

barrywhart:

Makes progress on #829

### Pull Request checklist

  - Other.


src/sqlfluff/core/templaters/jinja.py
```diff
@@ -200,10 +200,10 @@ def _crawl_tree(cls, tree, variable_names, raw):
                 line_pos=pos,
             )
 
-    @staticmethod
```
barrywhart, line 203: Lots of small changes in this file. The details are not interesting; basically refactoring so we always work with a Jinja environment that is aware of the `load_macros_from_path` config setting.

src/sqlfluff/core/templaters/python.py
```diff
@@ -175,7 +175,7 @@ def infer_type(s):
         except (SyntaxError, ValueError):
             return s
 
-    def get_context(self, fname=None, config=None) -> Dict:
```
barrywhart, line 178: Uninteresting changes updating the Python templater to align with the changes to `jinja.py`.

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -547,11 +547,10 @@ def make_template(in_str):
             [],
         )
 
-    @classmethod
```
barrywhart, line 550: Some refactoring of the templater API. See notes on `jinja.py` for more detail.

test/core/templaters/jinja_test.py
```diff
@@ -330,6 +331,7 @@ def test__templater_jinja_slices(case: RawTemplatedTestCase):
     """Test that Jinja templater slices raw and templated file correctly."""
     t = JinjaTemplater()
     templated_file, _ = t.process(in_str=case.instr, fname="test", config=FluffConfig())
+    assert templated_file
```
barrywhart, line 334: Makes mypy happy

test/core/templaters/jinja_test.py
```diff
@@ -347,8 +349,8 @@ def test__templater_jinja_slices(case: RawTemplatedTestCase):
 
     # Build and check the list of source strings referenced by "raw_sliced".
     previous_rs = None
-    actual_rs_source_list = []
-    for rs in templated_file.raw_sliced + [None]:
+    actual_rs_source_list: List[RawFileSlice] = []
```
barrywhart, line 352: Mypy stuff

jpy-git (commented):
@barrywhart So I think this looks good!

One thing that would be good to test is if the `include` or `from` can import from a subfolder of the search path.
i.e. specify `load_macros_from_path=macros`
and then try
```
{% include 'subfolder/include_comment.j2' %}
```

Motivation being this [comment](https://github.com/sqlfluff/sqlfluff/issues/829#issuecomment-1016468280) where the user does that.
(N.B. currently he has his macros path set to `dags\sql\macros\` which means jinja def won't know where to find `sql/macros/get_region.sql` since its not on the search path. But I'm hoping that setting `load_macros_from_path=dag` would be enough to solve that)

I can also see users maybe having these includes in a diff location to other macros (like in the example above) so is there any reason why load_macros_from_path can't be a comma-separated list of search paths, so they can import from multiple locations?

Once we've tested these things I think it's also worth adding to the docs a bit to clarify how it works to users.

barrywhart:
@jpy-git: I updated the PR to test including and importing from subdirectories. I'm 👍 for updating `load_macros_from_path` to support multiple entries and updating the docs, but I think that should be a separate PR because it requires changes to other areas of the code, e.g. `FluffConfig`, and this is already a fairly large PR.

I changed the PR description from "Fixes..." to "Makes progress on #829", so the linked issue will remain open.

test/core/templaters/jinja_test.py
```diff
@@ -551,14 +553,13 @@ def test__templater_jinja_slice_template(test, result):
 
 
 @pytest.mark.parametrize(
-    "raw_file,templated_file,override_context,result",
```
barrywhart, line 554: Cleaning up tech debt: The `templated_file` parameter in these test cases has not been used for the past several months (I think since `JinjaTracer` was added.)

jpy-git (approved):
Looks great @barrywhart! I've added a checklist to the original issue [here](https://github.com/sqlfluff/sqlfluff/issues/829#issuecomment-1017646218) for you to check off - This completes the first one 💪 LGTM 🚀 ✨

barrywhart:
Thanks!

PR #2356, opened 2022-01-19
whitespace concatenated string literals

### LINKED ISSUE #1488: Parser should handle newline-concatenated string literals in more situations
## Expected Behaviour

The following should parse successfully:

```sql
SELECT * FROM t_table WHERE c_column = 'foo'
'bar';

SELECT * FROM t_table WHERE c_column IN (
    'FOO'
    'BAR',
    'BAZ'
);

SELECT 'foo'
'bar'
'baz';
```

## Observed Behaviour

`sqlflull parse` fails with the following violations (full output attached):

```
==== parsing violations ====
L:   1 | P:  45 |  PRS | Line 1, Position 45: Found unparsable section: "\n'bar'"
L:   5 | P:   5 |  PRS | Line 5, Position 5: Found unparsable section: "'FOO'\n 'BAR',\n
                       | 'BAZ'"
L:  11 | P:   6 |  PRS | Line 11, Position 6: Found unparsable section: "\n'baz'"
```

[error_output.txt](https://github.com/sqlfluff/sqlfluff/files/7268779/error_output.txt)

## Dialect

This appears to currently happen in every dialect. In section 4.1.2.1 of the [Postgres docs](https://www.postgresql.org/docs/11/sql-syntax-lexical.html), it says:

> Two string constants that are only separated by whitespace with at least one newline are concatenated and effectively treated as if the string had been written as one constant.
> ...
>  (This slightly bizarre behavior is specified by SQL; PostgreSQL is following the standard.)

which makes me think this should be handled by the `ansi` dialect. I haven't been able to test with any other databases, but PostgreSQL will parse all of the example SQL (even the one with three contatenated strings).

## Version
Taken from latest `main` (0.6.6) on Python 3.8.3

jpy-git:

Closes #1488. Postgres and Redshift allow newline-concatenated string literals and MySQL allows whitespace+newline-concatenated string literals.

As described in this [comment](https://github.com/sqlfluff/sqlfluff/issues/1488#issuecomment-1016570918) it's possible to have comments between the string literals, therefore we solve this via grammar.

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -33,7 +33,10 @@
             r"(-- |#)[^\n]*",
             CommentSegment,
             segment_kwargs={"trim_start": ("-- ", "#")},
-        )
+        ),
+        RegexLexer(
+            "single_quote", r"(?s)('')+?(?!')|('.*?(?<!')(?:'')*'(?!'))", CodeSegment
+        ),
```
jpy-git, line 39: This just updates the quoted literal in mysql to match postgres. N.B. this matches the individual quoted literals, not the concatenated combination (for reason mentioned in PR description).

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -165,6 +168,17 @@
...
+        # MySQL allows whitespace-concatenated string literals (#1488).
+        # Since these string literals can have comments between them,
+        # we use grammar to handle this.
+        NamedParser(
+            "single_quote",
+            CodeSegment,
+            name="quoted_literal",
+            type="literal",
+        ),
+    ),
```
jpy-git, line 181: MySQL is very simple: Since it's string literals can be whitespace and/or newline spaced it's just `AnyNumberOf`.

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -203,12 +204,55 @@
...
+                name="quoted_literal",
+                type="literal",
+            ),
+            delimiter=NamedParser(
+                "newline",
+                NewlineSegment,
+                name="newline",
+                type="newline",
+            ),
+            allow_trailing=True,
```
jpy-git, line 223: Postgres is a little more complicated: quoted literals must have at least one newline between them (additional whitespace is fine) to be concatenated and therefore parse. This is done by delimiting via newline segments. The `allow_trailing` is because in this context we don't want to fail parsing if there's a newline after the final concatenated quoted literal, that's quite standard.

(N.B. `unicode_single_quote` and `escaped_single_quote` follow the same logic)

jpy-git: Also note redshift inherits this same logic and @rpr-ableton has kindly confirmed this behaviour for us [here](https://github.com/sqlfluff/sqlfluff/issues/1488#issuecomment-1016664179)

jpy-git:
Ahh spotted why that one statement wasn't parsing, one mo

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -165,6 +168,17 @@
             Ref("NumericLiteralSegment"),
         ),
     ),
+    QuotedLiteralSegment=AnyNumberOf(
+        # MySQL allows whitespace-concatenated string literals (#1488).
+        # Since these string literals can have comments between them,
+        # we use grammar to handle this.
+        NamedParser(
+            "single_quote",
```
barrywhart, line 176: I read somewhere that MySQL also supports double-quoted strings. Are we handling this?

https://dev.mysql.com/doc/refman/8.0/en/string-literals.html

barrywhart (commented):
Nice work! One question about support for double-quoted MySQL strings, otherwise 👍 .

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -168,6 +169,19 @@
...
+    # In general let the parser handle newlines and whitespace.
+    MultilineConcatenateNewline=NamedParser(
+        "newline",
+        NewlineSegment,
+        name="newline",
+        type="newline",
+    ),
+    MultilineConcatenateDelimiterGrammar=AnyNumberOf(
+        Ref("MultilineConcatenateNewline"), min_times=1, allow_gaps=False
+    ),
```
jpy-git, line 184: So the reason
```sql
SELECT 'foo'

'bar';
```
didn't parse before is that I needed to allow mulitple newlines as the delimiters.

I also had to define `MultilineConcatenateNewline` as the parser didn't like me using a `NamedParser` within the `AnyNumberOf`.

The `allow_gaps` is to force the delimiter to be greedy with multiple newlines (n.b. still allows whitespace, see unit tests) so all good.

jpy-git (commented):
Got the failing test working now 🎉

src/sqlfluff/dialects/dialect_mysql.py
jpy-git: I hadn't considered the double-quoted strings, let me take a look tomorrow and hopefully shouldn't be too hard to extend 👍

jpy-git: @barrywhart So double-quoted strings are disabled if `ANSI_QUOTES` SQL mode is enabled.

When I use the MySQL docker image ([source](https://hub.docker.com/_/mysql/)) to test this, it looks like this is enabled by default:
```
docker run --platform linux/amd64 --name some-mysql -e MYSQL_ROOT_PASSWORD=my-secret-pw -d -p 3306:3306 mysql
```
image

Given that by default (at least in docker) double quotes are still for quoted identifiers, and we don't support flags for different options, I would suggest we stick with the default behaviour.

We're currently parsing the double quote as a string literal but perhaps, given the above, we should also remove this and parse as a quoted identifier.
image

jpy-git: I've posted on #general on slack in case anyone disagrees with my approach above: https://sqlfluff.slack.com/archives/C01GPHR8J0G/p1642673594037700

barrywhart: Sounds good. Thanks for checking into this!

barrywhart (approved):

PR #2364, opened 2022-01-20
Exclude EMIT clauses from rule L013

### LINKED ISSUE #2361: EXASOL: L013 wrong result when using json_extract
In Exasol we could use the `json_extract()` function to extract multiple values out of a json string. The `emits` clause defines how the values are exposed. In such a case its not possible to use a alias.

## Expected Behaviour
L013 should not raise when using `json_extract()`

## Observed Behaviour
L013 raises.

## Steps to Reproduce
```sql
select
    somecol,
    json_extract(json_str, '$.AFIELD', '$.BFIELD') emits (cola char(1), colb char(1))
from abc;
```

## Dialect
Exasol

## Version
Include the output of `sqlfluff --version` along with your Python version.
If you are using dbt then please additionally include the dbt version.

sqlfluff v0.9.1

## Configuration
```
```

tunetheweb:

Fixes #2361 

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -3770,3 +3766,17 @@ class FileSegment(BaseFileSegment):
             allow_trailing=True,
         ),
     )
+
+
+@exasol_dialect.segment()
+class EmitsGrammar(BaseFileSegment):
+    """EMITS Grammar for JSON_EXTRACT for example.
```
tunetheweb, line 3773: I've left this with the same name but what's the difference between a"Grammar" and a "Sequence" out of interest? Should this be a Sequence?

src/sqlfluff/rules/L013.py
```diff
@@ -48,6 +48,14 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
         if segment.all(sp.is_type("select_clause_element")) and not children.any(
             sp.is_type("alias_expression")
         ):
+            # Ignore if it's a function with an EMITS clause as EMITS is equivalent to AS
+            if (
+                children.select(sp.is_type("function"))
+                .children()
+                .select(sp.is_type("emits_grammar"))
+            ):
+                return None
```
tunetheweb, line 57: @barrywhart my first rule change using your new rule functions! Is this the right way of using it?

Also trying to think of any queries that might not follow this. For example, if you wrapped the whole select clause in brackets:

```sql
(json_extract(json_str, '$.AFIELD', '$.BFIELD') emits (cola char(1), colb char(1)))
```

Then the above code would not catch it and L013 would flag but then not sure if that's actually OK and it SHOULD flag (I don't know Exasol so don't know how Emits works). Thoughts @sti0 ?

Maybe we should just search for ANY occurrence of `emits_grammar` in any children/grandchildren...etc. rather than just the direct children functions but not sure how easy that is (or even how accurate that would be and maybe above is actually right and rest SHOULD flag for L013)).

src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -3770,3 +3766,17 @@ class FileSegment(BaseFileSegment):
             allow_trailing=True,
         ),
     )
+
+
+@exasol_dialect.segment()
+class EmitsGrammar(BaseFileSegment):
```
jpy-git, line 3772:
```suggestion
class EmitsGrammar(BaseSegment):
```
`BaseFileSegment` should only be used by `FileSegment`

src/sqlfluff/dialects/dialect_exasol.py
jpy-git: `Sequence` is a type of grammar (as are `AnyNumberOf`, `AnySetOf`, `Delimited`, etc..).

So calling it `EmitsGrammar` made sense when it was written like:
```python
EmitsGrammar=Sequence(
        "EMITS",
        Bracketed(Ref("UDFParameterGrammar")),
    ),
```

This is now a `Segment` (inherits from `BaseSegment`) so we should change the name to `EmitsSegment` (or something similar).

src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -3770,3 +3766,17 @@ class FileSegment(BaseFileSegment):
...
+    """EMITS Grammar for JSON_EXTRACT for example.
+
+    In it's own class to give it a type to allow L013 to find it easily.
```
jpy-git, line 3775:
```suggestion
    In it's own segment to give it a type to allow L013 to find it easily.
```

src/sqlfluff/rules/L013.py
jpy-git: > Maybe we should just search for ANY occurrence of `emits_grammar` in any children/grandchildren...etc. rather than just the direct children functions but not sure how easy that is (or even how accurate that would be and maybe above is actually right and rest SHOULD flag for L013)).

@barrywhart On this point, think we should add the `recursive_crawl` method to `Segments` to handle this sort of case? Not necessarily suggesting it has to be used in this rule, but would be useful in general.

jpy-git (commented):
Couple suggestions, I'll let those more familiar with exasol comment on the fix itself.

src/sqlfluff/dialects/dialect_exasol.py
tunetheweb: Oops - good spot. Fixed

src/sqlfluff/dialects/dialect_exasol.py
tunetheweb: OK makes sense.

Though would be nice to be able to add a name attribute to Grammars without having to "upgrade" Grammars Segments just for that 😞

src/sqlfluff/rules/L013.py
jpy-git: ^pulling together a pr for recursive crawl

tunetheweb: Yeah I get how it works. So we can have:

```sql
select
    somecol,
    json_extract(json_str, '$.AFIELD', '$.BFIELD') emits (cola char(1), colb char(1))
from abc;
```

And that's basically equivalent to this:
```sql
select
    somecol,
    json_extract(json_str, '$.AFIELD') AS cola,
    json_extract(json_str, '$.BFIELD') AS colb
from abc;
```

The current fix caters for that.

However my question is, is below valid and equivalent?

```sql
select
    somecol,
    (json_extract(json_str, '$.AFIELD', '$.BFIELD') emits (cola char(1), colb char(1)))
from abc;
```

That is EMITS nested in another clause (brackets in this case)?

Because the current fix doesn't category that and ONLY works if the FUNCTION and EMITS is directly part of the SELECT clause.

But the more I think about it, the more it seems that brackets (or other nesting) aren't valid anyway, and EMITS probably DOES have to be part of the top level SELECT clause so think we're fine. And if it isn't we can always revisit later.

barrywhart: @tunetheweb: Usage of the new rule functions looks 💯.

@jpy-git: +1 for updating the `Segments` class to provide `recursive_crawl()`. My intuition (based on a quick scan of the EXASOL docs) is that we wouldn't need to use that here, but I'm not certain. If we're in doubt, seems better to omit it for now. 🤷‍♂️

barrywhart (approved):
Looks good! 🎉

src/sqlfluff/rules/L013.py
sti0: @tunetheweb thanks for the fix. I will look into the `(json_extract(json_str, '$.AFIELD', '$.BFIELD') emits (cola char(1), colb char(1)))` thing tomorrow. 

Only for your interesst: `json_extract()` must use `emits()`. This 
```sql
select
    somecol,
    json_extract(json_str, '$.AFIELD') AS cola,
    json_extract(json_str, '$.BFIELD') AS colb
from abc;
```
must be written like this:
```sql
select
    somecol,
    json_value(json_str, '$.AFIELD') AS cola,
    json_value(json_str, '$.BFIELD') AS colb
from abc;
```

PR #2371, opened 2022-01-20
Fix L026 false positive on "SELECT INTO" statement

### LINKED ISSUE #1830: Exasol L026 - Fails on SELECT INTO TABLE statement
In Exasol we could insert data into a table like this:

```
select * into table schm.tab from schm.tab2;
```

## Expected Behaviour
Rule L026 should pass

## Observed Behaviour
Rule was raised
```
L:   2 | P:  21 | L026 | Reference 'schm.tab' refers to table/view not found in
                       | the FROM clause or found in parent subquery
```

## Steps to Reproduce
Lint the statement above

## Dialect
exasol

## Version
sqlfluff 0.7.1
python 3.8.12
## Configuration
default

barrywhart:
Fixes #1830

### Pull Request checklist

  - Other.


src/sqlfluff/rules/L026.py
```diff
@@ -122,19 +122,33 @@ def _analyze_table_references(
                 # Try and resolve each reference to a value in query.aliases (or
                 # in an ancestor query).
                 for r in select_info.reference_buffer:
-                    # This function walks up the query's parent stack if necessary.
-                    violation = self._resolve_reference(
-                        r, self._get_table_refs(r, dialect), dml_target_table, query
-                    )
-                    if violation:
-                        violations.append(violation)
+                    if not self._should_ignore_reference(r, selectable):
```
barrywhart, line 125: Lines 126-131 are not changed; I simply moved them inside a new `if` statement.

src/sqlfluff/rules/L026.py
```diff
@@ -122,19 +122,33 @@ def _analyze_table_references(
...
 
         # Visit children.
         for child in query.children:
             self._analyze_table_references(
                 cast(L026Query, child), dml_target_table, dialect, violations
             )
 
+    @staticmethod
+    def _should_ignore_reference(reference, selectable):
+        if selectable.dialect.name == "exasol":
```
tunetheweb, line 141: Why just exasol? Surely this is common to many (all?) dialects?

T-SQL has this for example.

The other dialects SHOULD have type set up IMHO this but don't. Whether we want to change that as part of this PR, or just leave that for later is up to you, but either way I think we should remove this `if` statement.

tunetheweb (changes requested):

src/sqlfluff/rules/L026.py
jpy-git: Yeah I think removing it makes sense

barrywhart: Good catch! `SELECT INTO` sounded like it _might_ be generic, but I didn't think to check.

I'll remove the `if` statement, but would rather not make any dialect changes in this PR (scope creep). Want to create an issue for that, or should we wait until a user reports it?

tunetheweb: Happy to wait for reports.

L026 is disabled by default for some dialects anyway ("bigquery", "hive", "redshift").

PR #2373, opened 2022-01-20
TSQL: Add ASA Clustered Index functionality

jpers36:
Azure Synapse Analytics allows for tables to be created with clustered indexes.  This PR allows SqlFluff to parse this type of table definition.

None.

### Pull Request checklist

### DIFF
--- a/src/sqlfluff/dialects/dialect_tsql.py
+++ b/src/sqlfluff/dialects/dialect_tsql.py
@@ -2202,6 +2202,20 @@ class TableIndexClause(BaseSegment):
                 "COLUMNSTORE",
                 "INDEX",
             ),
+            Sequence(
+                "CLUSTERED",
+                "INDEX",
+                Bracketed(
+                    Delimited(
+                        Ref("ColumnReferenceSegment"),
+                        OneOf(
+                            "ASC",
+                            "DESC",
+                            optional=True,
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -2202,6 +2203,20 @@ class TableIndexClause(BaseSegment):
                 "COLUMNSTORE",
                 "INDEX",
             ),
+            Sequence(
```
jpy-git, line 2206:
```
CLUSTERED COLUMNSTORE INDEX ORDER (column [,...n])
```
Looking at the docs, this is also an option so we could probably add this now?

Think we'd just need to update the lines above this to:
```python
Sequence(
    "CLUSTERED",
    "COLUMNSTORE",
    "INDEX",
    Sequence(
        "ORDER",
        Bracketed(
            Delimited(
                Ref("ColumnReferenceSegment")
            ),
        ),
        optional=True
    ),
),
```

jpers36: @jpy-git good catch!  I'll update.

jpy-git (approved):
Thanks @jpers36! LGTM 🚀 ✨

PR #2400, opened 2022-01-21
L044, L045: Handle Exasol VALUES clause

### LINKED ISSUE #2362: L044 unexpected exception, L045 CTE not used?
<!--If this is a parsing or linting issue, please include a minimal SQL example which reproduces the issue, along with the `sqlfluff parse` output, `sqlfluff lint` output and `sqlfluff fix` output when relevant.-->

## Expected Behaviour

1. There should be no unexpected exception with L044
2. Both CTE are used. L045 should not raise.

## Observed Behaviour

```
CRITICAL   [L044] Applying rule L044 threw an Exception:                                                                                                                                                                                                              
  File "/home/vscode/workspace/_venv/sqlfluff_venv/lib/python3.8/site-packages/sqlfluff/core/rules/base.py", line 504, in crawl
    res = self._eval(
  File "/home/vscode/workspace/_venv/sqlfluff_venv/lib/python3.8/site-packages/sqlfluff/rules/L044.py", line 136, in _eval
    return self._analyze_result_columns(crawler.query_tree)
  File "/home/vscode/workspace/_venv/sqlfluff_venv/lib/python3.8/site-packages/sqlfluff/rules/L044.py", line 85, in _analyze_result_columns
    assert query.selectables
AssertionError
L:   1 | P:   1 | L044 | Unexpected exception: ; Could you open an issue at
                       | https://github.com/sqlfluff/sqlfluff/issues ? You can
                       | ignore this exception for now, by adding '--noqa: L044'
                       | at the end of line 1
L:   1 | P:   1 | L045 | Query defines a CTE (common-table expression) but does
                       | not use it.
```

## Steps to Reproduce
```sql
WITH base AS (
    SELECT
        c1 AS cx,
        c2
    FROM
        schm.tab
),
txt AS (
    VALUES
    (1,
        'v1'),
    (2,
        'v2')
    AS t
    (id,
        txt)
)
SELECT
    cb.cx,
    cb.c2,
    txt_de.txt AS de_txt
FROM
    base AS cb
LEFT JOIN txt AS txt_de ON
    cb.c2 = txt_de.id;
```

## Dialect
Exasol / Ansi

## Version
Include the output of `sqlfluff --version` along with your Python version.
If you are using dbt then please additionally include the dbt version.

sqlfluff v0.9.1
python v3.8.12


## Configuration
```
```

barrywhart:
Fixes #2362

Enhances `SelectCrawler` class to interpret `VALUES` clause in a CTE as an equivalent `SELECT`.

### Pull Request checklist

  - Other.

### DIFF
--- a/test/fixtures/rules/std_rule_cases/L044.yml
+++ b/test/fixtures/rules/std_rule_cases/L044.yml
@@ -541,3 +541,12 @@ test_pass_exasol_values_clause_cte_2:
     core:
       dialect: exasol
 
+test_pass_exasol_values_clause:
+  pass_str:
+    SELECT *
+    FROM (
+        VALUES (1, 2), (3, 4)
+    ) AS t(c1, c2)
+  configs:
+    core:
+      dialect: exasol
--- a/test/fixtures/rules/std_rule_cases/L044.yml
+++ b/test/fixtures/rules/std_rule_cases/L044.yml
@@ -541,28 +541,3 @@ test_pass_exasol_values_clause_cte_2:
     core:
       dialect: exasol
 
-test_pass_snowflake_values_clause_cte:
-  pass_str: |
-    WITH txt AS (
-        VALUES (1)
-        AS t (id)
-    )
-
-    SELECT *
-    FROM txt
-  configs:
-    core:
-      dialect: snowflake
-
-test_pass_postgres_values_clause_cte:
-  pass_str: |
-    WITH txt AS (
-        VALUES (1)
-        AS t (id)
-    )
-
-    SELECT *
-    FROM txt
-  configs:
-    core:
-      dialect: postgres
--- a/test/fixtures/rules/std_rule_cases/L044.yml
+++ b/test/fixtures/rules/std_rule_cases/L044.yml
@@ -515,7 +515,7 @@ test_snowflake_delete_cte:
     core:
       dialect: snowflake
 
-test_pass_exasol_values_clause_cte:
+test_pass_exasol_values_clause_cte_1:
   pass_str: |
     WITH txt AS (
         VALUES (1)
@@ -524,3 +524,32 @@ test_pass_exasol_values_clause_cte:
 
     SELECT *
     FROM txt
+  configs:
+    core:
+      dialect: exasol
+
+test_pass_exasol_values_clause_cte_2:
+  pass_str: |
+    WITH txt AS (
+        VALUES (1, 'foo')
+        AS t (id, name)
+    )
+
+    SELECT *
+    FROM txt
+  configs:
+    core:
+      dialect: exasol
+
+test_pass_snowflake_values_clause_cte:
+  pass_str: |
+    WITH txt AS (
+        VALUES (1)
+        AS t (id)
+    )
+
+    SELECT *
+    FROM txt
+  configs:
+    core:
+      dialect: snowflake
--- a/test/fixtures/rules/std_rule_cases/L044.yml
+++ b/test/fixtures/rules/std_rule_cases/L044.yml
@@ -514,3 +514,13 @@ test_snowflake_delete_cte:
   configs:
     core:
       dialect: snowflake
+
+test_pass_exasol_values_clause_cte:
+  pass_str: |
+    WITH txt AS (
+        VALUES (1)
+        AS t (id)
+    )
+
+    SELECT *
+    FROM txt
--- a/test/fixtures/rules/std_rule_cases/L044.yml
+++ b/test/fixtures/rules/std_rule_cases/L044.yml
@@ -553,3 +553,16 @@ test_pass_snowflake_values_clause_cte:
   configs:
     core:
       dialect: snowflake
+
+test_pass_postgres_values_clause_cte:
+  pass_str: |
+    WITH txt AS (
+        VALUES (1)
+        AS t (id)
+    )
+
+    SELECT *
+    FROM txt
+  configs:
+    core:
+      dialect: postgres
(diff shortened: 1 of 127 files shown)


test/fixtures/rules/std_rule_cases/L044.yml
```diff
@@ -514,3 +514,13 @@ test_snowflake_delete_cte:
...
+
+test_pass_exasol_values_clause_cte:
+  pass_str: |
+    WITH txt AS (
+        VALUES (1)
+        AS t (id)
+    )
+
+    SELECT *
+    FROM txt
```
jpy-git, line 526: This isn't valid snowflake/postgres/mysql
image
so maybe use
```yml
configs:
    core:
      dialect: exasol
```
in these test cases

test/fixtures/rules/std_rule_cases/L044.yml
```diff
@@ -514,3 +514,13 @@ test_snowflake_delete_cte:
   configs:
     core:
       dialect: snowflake
+
+test_pass_exasol_values_clause_cte:
+  pass_str: |
+    WITH txt AS (
+        VALUES (1)
```
tunetheweb, line 521: Can we have an example with multiple values?

test/fixtures/rules/std_rule_cases/L044.yml
jpy-git: This however is valid snowflake/postgres (although annoyingly doesn't parse in postgres atm) and also raises the same L044 issue:
- On main branch
image

- On your branch:
image

L044 still there and catastophic failures 😱 🔥

jpy-git: (I'll fix the postgres parsing for my example above since I know exactly where I've accidently introduced this, but that's not related to the above issue 😄 )

jpy-git: Came up with the equivalent MySQL test case but it won't parse so I've added it into issue #2402. I have a feeling if it could parse then it would likely be solved by your fix for the postgres/snowflake example so we can leave it for a later date for now.

barrywhart: Oops, I'll add the `exasol` config. I intended to do that, but forgot. Will also add the same test for `snowflake` dialect.

test/fixtures/rules/std_rule_cases/L044.yml
barrywhart: Can do. I will say, though, the new code doesn't look at that part at all, AFAIK.

test/fixtures/rules/std_rule_cases/L044.yml
barrywhart: I added a `postgres` test case as well.

barrywhart: @jpy-git: Think we could create a separate issue for your `SELECT *` test case? I agree it should be fixed, but I think it's less likely to occur in production code. Happy to look into it, but I'd rather address it later (potentially next release).

test/fixtures/rules/std_rule_cases/L044.yml
```diff
@@ -514,3 +514,55 @@ test_snowflake_delete_cte:
...
+        AS t (id)
+    )
+
+    SELECT *
+    FROM txt
+  configs:
+    core:
+      dialect: postgres
```
jpy-git, line 568:
```suggestion
```
This isn't valid snowflake/postgres so we shouldn't add it
image

test/fixtures/rules/std_rule_cases/L045.yml
```diff
@@ -172,3 +172,55 @@ test_snowflake_delete_cte:
...
+    WITH txt AS (
+        VALUES (1)
+        AS t (id)
+    )
+
+    SELECT *
+    FROM txt
+  configs:
+    core:
+      dialect: postgres
```
jpy-git, line 226:
```suggestion
```
Not valid syntax

test/fixtures/rules/std_rule_cases/L044.yml
jpy-git: Hmm it's kind of the same as the existing test case just in a subquery rather than a CTE, and we'd be introducing catastrophic errors in rules L025 + L026. So don't think we should be merging this before at least the L025 + L026 [errors](https://github.com/sqlfluff/sqlfluff/pull/2400#discussion_r789718402) are fixed.

Worst case this runs into next release.

barrywhart: I'll try and take a look this afternoon. I often have a couple of hours to work, then I don't always know when I'll have time again. 👶

barrywhart: @jpy-git: I found a parsing issue with certain uses of `VALUES`. I created an issue in case you want to look at it: #2406. For now, I'll focus on the ones that parse correctly.

barrywhart:
@jpy-git, @tunetheweb: I believe I've addressed the PR feedback, including the additional fixes.

The only failure(s) I'm aware of now are caused by issue #2406 (which I created today). That's apparently a dialect issue which I think we should address separately. No other changes to `SelectCrawler` or the rules should be necessary once that's fixed; things will "just work".

jpy-git (approved):
@barrywhart Looks good, I've tested and can confirm those L025/L026 error have gone too 👍 LGTM 🚀 ✨ 

As you said I think once we get that remaining Postgres/MySQL/Exasol Values grammar parsing, they should be handled by this rule in the exact same way. I'm looking at that in #2404 so will try and add extra unit tests to that to cover these statements

PR #2401, opened 2022-01-21
Postgres: Fix Values alias regression

jpy-git:

Fixes parsing issue noted [here](https://github.com/sqlfluff/sqlfluff/pull/2400/files#r789723089).
In #2353 I tidied up the postgres alias expression to remove quoted literal alias which are invalid. At the time I also thought the bracketed segment was also not valid, but have since realised I was wrong there! :)
This adds the values alias back in to the dialect.

N.B. I also checked the syntax that I spotted this issue in and it's not valid MySQL so I think the MySQL changes in #2353 are still valid.

No

### Pull Request checklist

  - Other.

### DIFF
--- a/test/fixtures/dialects/postgres/postgres_values_alias.sql
+++ b/test/fixtures/dialects/postgres/postgres_values_alias.sql
@@ -1,0 +1,4 @@
+select *
+from (
+    values (1, 2), (3, 4)
+) as t(c1, c2);
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -3055,6 +3055,7 @@ class AliasExpressionSegment(BaseSegment):
     match_grammar = Sequence(
         Ref.keyword("AS", optional=True),
         Ref("SingleIdentifierGrammar"),
+        Bracketed(Ref("SingleIdentifierListSegment"), optional=True),
     )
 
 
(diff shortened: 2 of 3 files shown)


test/fixtures/dialects/postgres/postgres_values_alias.sql
```diff
@@ -0,0 +1,4 @@
+select *
+from (
+    values (1, 2), (3, 4)
+) as t(c1, c2);
```
tunetheweb, line 4: is `as t (c1, c2)` valid (i.e. with a space)?

If not is there anyway for SQLFluff to prevent parsing that? Cause at the moment I think it will accept it when it maybe shouldn't?

barrywhart (approved):

tunetheweb (changes requested):
As discussed above, would like to consider that space issue some more before merging this.

test/fixtures/dialects/postgres/postgres_values_alias.sql
jpy-git: @tunetheweb yeah no spaces is fine 😄 
image

For general interest though, yes you can prevent SQLFluff from parsing gaps.
You would use the `allow_gaps=False` paramater to the grammar segment.
e.g. Hypothetically if this grammar didn't allow spaces (it does but let's imagine):
```python
match_grammar = Sequence(
        Sequence(
            Bracketed(Ref("SingleIdentifierListSegment"), optional=True),
            allow_gaps=False,
        ),
    )
```
😎

tunetheweb: Excellent. Approval coming up!

tunetheweb (approved):

PR #2405, opened 2022-01-21
L060: Use COALESCE instead of IFNULL or NVL.

### LINKED ISSUE #2392: New Rule: Use COALESCE over IFNULL or NVL
From Brooklyn Data co. style guide: https://github.com/brooklyn-data/co/blob/main/sql_style_guide.md#use-coalesce-instead-of-ifnull-or-nvl
```
Use coalesce instead of ifnull or nvl.
- coalesce is universally supported, whereas Redshift doesn't support ifnull and BigQuery doesn't support nvl.
- coalesce is more flexible and accepts an arbitrary number of arguments.
```

Should be pretty simple to implement

jpy-git:

Closes #2392. Adds a new rule to require the use of `COALESCE` over `IFNULL` or `NVL`. See linked issue for source + motivation. 

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L060.py
+++ b/src/sqlfluff/rules/L060.py
@@ -1,0 +1,62 @@
+"""Implementation of Rule L060."""
+
+from typing import Optional
+
+from sqlfluff.core.parser.segments.raw import CodeSegment
+from sqlfluff.core.rules.base import BaseRule, LintFix, LintResult, RuleContext
+from sqlfluff.core.rules.doc_decorators import document_fix_compatible
+
+
+@document_fix_compatible
+class Rule_L060(BaseRule):
+    """Use ``COALESCE`` instead of ``IFNULL`` or ``NVL``.
+
+    | **Anti-pattern**
+    | ``IFNULL`` or ``NVL`` are used to fill ``NULL`` values.
+
+    .. code-block:: sql
+
+        SELECT ifnull(foo, 0) AS bar,
+        FROM baz;
+
+        SELECT nvl(foo, 0) AS bar,
+        FROM baz;
+
+    | **Best practice**
+    | Use ``COALESCE`` instead.
+    | ``COALESCE`` is universally supported,
+    | whereas Redshift doesn't support ``IFNULL``
+    | and BigQuery doesn't support ``NVL``.
+    | Additionally ``COALESCE`` is more flexible
+    | and accepts an arbitrary number of arguments.
+
+    .. code-block:: sql
+
+        SELECT coalesce(foo, 0) AS bar,
+        FROM baz;
+
+    """
+
+    def _eval(self, context: RuleContext) -> Optional[LintResult]:
+        """Use ``COALESCE`` instead of ``IFNULL`` or ``NVL``."""
+        # We only care about function names.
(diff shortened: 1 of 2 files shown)


jpers36:
In TSQL, COALESCE and ISNULL (TSQL's version of it) perform slightly differently, [specifically around datatyping](https://www.mssqltips.com/sqlservertip/2689/deciding-between-coalesce-and-isnull-in-sql-server/), such that they can't be swapped one for the other 100% of the time. TSQL won't be affected by this change at the moment since ISNULL isn't checked, but are there any similar concerns in the dialects that this change will impact?

sti0 (approved):
Great rule. LGFM. 🎉

barrywhart (approved):

jpy-git:
> In TSQL, COALESCE and ISNULL (TSQL's version of it) perform slightly differently, [specifically around datatyping](https://www.mssqltips.com/sqlservertip/2689/deciding-between-coalesce-and-isnull-in-sql-server/), such that they can't be swapped one for the other 100% of the time. TSQL won't be affected by this change at the moment since ISNULL isn't checked, but are there any similar concerns in the dialects that this change will impact?

@jpers36 so while I agree there can be small performance differences between these similar functions, I think that sort of misses the spirit of a rule like this which is intended for code consistency and portability which is often more valued in Data Analytics/Science contexts than these small performance differences. I think users concerned about you can always just turn on or off depending on preference.

I didn't consider `ISNULL` since it wasn't in the Brooklyn Data co. guide but based off the definition/motivation we should probably include it in this rule as well.

jpers36:
> @jpers36 so while I agree there can be small performance differences between these similar functions, I think that sort of misses the spirit of a rule like this which is intended for code consistency and portability which is often more valued in Data Analytics/Science contexts than these small performance differences. I think users concerned about you can always just turn on or off depending on preference.

To be clear, my concerns are not performance-related, at least not in terms of speed. Switching ISNULL to COALESCE will break code at times.

jpy-git:
> To be clear, my concerns are not performance-related, at least not in terms of speed. Switching ISNULL to COALESCE will break code at times.

@jpers36 ah gotcha thanks for clarifying! That would likely explain why it isn't included in the original style guide then 😄 

Will merge as is then 👍

barrywhart:
@jpy-git: I'm trying to follow the discussion above. From your final comment:
>@jpers36 ah gotcha thanks for clarifying! That would likely explain why it isn't included in the original style guide then 😄
>Will merge as is then 👍

I thought you were going to either exclude this rule for T-SQL or for `ISNULL`, but it looks like we didn't. Did I misunderstand something? Thanks!

jpy-git:
@barrywhart `ISNULL` was never included in this PR? This rule (inline with the Brooklyn Data co style guide) changes `IFNULL` & `NVL`.

barrywhart:
Oh, got it! `IFNULL` vs `ISNULL` -- got confused by the similarity. 

🎉 💯

jpy-git:
> `IFNULL` vs `ISNULL` -- got confused by the similarity.

Classic confusing SQL naming haha 😅

PR #2215, opened 2021-12-29
Support `DROP` DDL statements for Spark3

R7L208:
### Brief summary of the change made

This PR adds code and test coverage for supported spark3 statements that `DROP` an object, ie database, table, function, etc..

### Are there any other side effects of this change that we should be aware of?
- `DropDatabseStatementSegment` was moved from Hive dialect to Ansi dialect.
  - Added `DropSchemaStatementSegment` to Hive dialect since the Ansi `DropDatabaseStatementSegment` does not match for `SCHEMA`.
  - Added `replace=True` parameter for `DropDatabaseStatementSegment` in `src/sqlfluff/dialects/dialect_postgres.py` to avoid conflict with the Ansi dialect
- Updates to `DropStatement` in Ansi dialect to better reflect usage, ie not matching `TableReferenceSegment` against a `USER`. I tried to extend `DropStatement` in Ansi to cover Spark3 test cases but ended up replacing it within the spark3 dialect. I created comments in the docstring of Ansi file to provide clarification of what was attempted and why it did not work.
- Created additional test cases for `test/fixtures/dialects/spark3/create_database.sql` and `test/fixtures/dialects/spark3/create_function.sql` to capture unparsable sections that I received when linting a project.
- Added `optional=True` for `ResourceLocationGrammar` in `CreateFunctionStatementSegment` as this should not be required syntas.
- 
### Pull Request checklist
- [X] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #2407, opened 2022-01-21
Exasol: Overhaul drop statements

sti0:

The Exasol `DROP` statements are defined stricter in a recent change (https://github.com/sqlfluff/sqlfluff/pull/2215). This PR finish that work. The `DropWithoutOptionsStatementSegment` was removed and splited in separate segments. I aligned the other drop segments as well. All `DROP` segments are put in place to the corresponding `CREATE/ALTER` segment.

Maybe if one uses the `DropWithoutOptionsStatementSegment` segment.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -2563,6 +2464,29 @@ class UserLDAPAuthSegment(BaseSegment):
...
+    is_ddl = False
+    is_dml = False
+    is_dql = False
+    is_dcl = True
+
+    match_grammar = Sequence(
+        "DROP",
+        "USER",
+        Ref("IfExistsGrammar", optional=True),
+        Ref("NakedIdentifierSegment"),
```
jpy-git, line 2485: Do you know if this can be a quoted identifier in Exasol?

In a lot of other dialects this is possible.
e.g. Postgres:
image

Same question applies to all the `DROP` segments with `NakedIdentifierSegment` in this PR?

(also I appreciate you didn't introduce them in this PR, but good opportunity to check since I'm not an exasol user 😄 )

src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -2616,6 +2540,20 @@ class ConsumerGroupParameterSegment(BaseSegment):
...
 
 
+@exasol_dialect.segment()
+class DropConsumerGroupSegment(BaseSegment):
+    """A `DROP CONSUMER GROUP` statement.
+
+    https://docs.exasol.com/sql/consumer_group.htm
+    """
+
+    type = "drop_consumer_group_statement"
```
jpy-git, line 2550: A lot of the other segments in the exasol dialect flag the statement type like this:
```
is_ddl = False
is_dml = False
is_dql = False
is_dcl = True
```
Should we also add it in here?

Also do you know if these get used anywhere? (haven't seen them in the other dialects)

jpy-git (commented):
@sti0 Looks good, couple quick things to confirm 😄

src/sqlfluff/dialects/dialect_exasol.py
sti0: I implemented them as I introduced the Exasol dialect. The idea behind this was to use this information after parsing for other programs (e.g. our deployment tool which now works with `sqlfluff` under the hood). 

It could be discussed if we should remove this attributes.

A consumer group fits in none of these attributes.

src/sqlfluff/dialects/dialect_exasol.py
sti0: Thanks for the hint. Its may bad pratice to use double quotes for identifier (my personal meaning) but its possible. I will look into this and fix the impacted locations. 

We should make this in another PR because there are more than the `DROP` statements impacted (e.g. create user/role...)

src/sqlfluff/dialects/dialect_exasol.py
jpy-git: Ah nice, thanks for clarifying! tbh I don't think there's any need to remove those attributes anytime soon was just curious 😄

src/sqlfluff/dialects/dialect_exasol.py
jpy-git: Yeah totally agree with separating the fix into a separate PR but just wanted to check 👍 

FWIW I'm also strongly against using quoted identifiers, but there are others less sane 😄

jpy-git (approved):
@sti0 Nice work and thanks for assisting with my questions 😄 LGTM 🚀 ✨

PR #2413, opened 2022-01-22
Add Oracle PROMPT statement

r0fls:

This adds support for the Oracle `PROMPT` statement.

No

### Pull Request checklist

  - Other.


r0fls:
Using the Regex was straightforward here. @tunetheweb if you can give me some pointers to use the other method, if preferred, I can go that route. Otherwise if this looks good I will add a comment to explain the Regex similar to the one here: https://github.com/sqlfluff/sqlfluff/pull/1423/files#diff-84df248c7619e658dbf10bea3cad532f9c0d17e06230f791f4cdd419721d2573R80-R91

src/sqlfluff/dialects/dialect_oracle.py
```diff
@@ -16,10 +18,22 @@
...
-    ["COMMENT", "ON", "UPDATE", "INDEXTYPE"]
+    ["COMMENT", "ON", "UPDATE", "INDEXTYPE", "PROMPT"]
 )
 
 
+oracle_dialect.insert_lexer_matchers(
+        [
+            RegexLexer(
+                "prompt_command",
+                r"PROMPT([^(\r\n)]|)+((?=\n)|(?=\r\n))?",
```
tunetheweb, line 29: Rather than the initial `|`, could you not just change the `+` to `*`?

```suggestion
                r"PROMPT([^(\r\n)])*((?=\n)|(?=\r\n))?",
```

test/fixtures/dialects/oracle/prompt.sql
```diff
@@ -0,0 +1,2 @@
+PROMPT this is an Oracle SQL newline delimited prompt statement
+SELECT job_id from employees;
```
tunetheweb, line 2: Since you specifically added a change for an empty PROMPT shall we add a test case for that?

This will also test that PROMPTS don't need to be semi-colon separated

```suggestion
PROMPT this is an Oracle SQL newline delimited prompt statement
PROMPT
PROMPT another prompt
SELECT job_id from employees;
```

tunetheweb (changes requested):
Change itself looks pretty good, but suggest a small number of changes.

Also the CI is failing because you need to run `black` to format your code.

Would recommend using the pre-commit hook described [here](https://github.com/sqlfluff/sqlfluff/blob/main/CONTRIBUTING.md#pre-commit-config) so this will just run automatically for you at commit time 👍

src/sqlfluff/dialects/dialect_oracle.py
jpy-git: +1 regex is greedy so the * should suffice

tunetheweb (approved):

test/fixtures/dialects/oracle/prompt.sql
r0fls: Whoops, sorry for the delay here. Thanks for picking up my slack @tunetheweb! I was skiing ⛷️ 😄

r0fls:
Thank you @tunetheweb @jpy-git for the speedy review, and sorry for the delayed updates: I was off skiing in the blue skies! ⛷️ You two are amazing 🙂

PR #2421, opened 2022-01-22
Misc grammar improvements for Snowflake

chwiese:

Fixes column definitions in the Snowflake CREATE EXTERNAL TABLE statement and adds the INTEGRATION parameter.
As a minor code improvement, I removed the `optional=True` parameters within the `AnyNumberOf()` segment, because they are redundant. The use of `AnySetOf()` is not strictly speaking correct, because `LOCATION` and `FILE_FORMAT` are required. These two parameters can however be in arbitrary order with the optional parameters and I do not know of a way to achieve that outside of `AnySetOf()`. See #2426.
I noticed that the top of the file still references the Postgres dialect, which is incorrect, so I included that as a drive-by fix.

During the code review, it became apparent that the dialect has several instances where AnyNumberOf is used, when it should be AnySetOf. When fixing that, I noticed that CreateStatementCommentSegment replicates CommentEqualsClauseSegment, so I removed it. 

No.

### Pull Request checklist

  - Other.


jpy-git:
`AnySetOf` is more appropriate if the options can't be repeated, basically the same as `AnyNumberOf` but you can't repeat the children.

Regarding those elements being required but also in any order I don't think we have anything to handle that atm so I think we're fine just using `AnySetOf`. Although we want to reflect the grammar where possible, It's more important that we get sensible statements parsing for linting purposes (we can assume people are using the relevant docs to understand syntax rather than reading SQLFluff grammar).

(you could open an issue about adding that feature to `AnyNumberOf`/`AnySetOf` as a future improvement 😄 )

chwiese:
Excellent suggestion! It did make me snowball a bit, because I realised there are several places in the Snowflake dialect where `AnyNumberOf` should be `AnySetOf`. When changing that, I saw that `CreateStatementCommentSegment` replicates `CommentEqualsClauseSegment`, so I removed it. These changes also required a reordering of the elements in the CREATE TASK statement, so that we actually hit the `USER_TASK_TIMEOUT_MS` keyword.

tunetheweb (approved):
LGTM. Though will rename it if you don’t mind.

chwiese:
> LGTM. Though will rename it if you don’t mind.

Not at all, seems like naming branches and PRs is not my strongest suit. 😅

PR #2423, opened 2022-01-22
"sqlfluff fix" should report any parse errors found

### LINKED ISSUE #2345: fix succeeds even though lint fails with PRS error
## Expected Behaviour

I assume fix is basically lint + fix (probably many do if they fix in a pre-commit, for example). If that assumption is reasonable, then `fix -f` should fail whenever `lint` would fail.

## Observed Behaviour

It's possible that `fix -f` succeeds when fix doesn't find anything to do, even though there are errors that would cause `lint` to fail. 

## Steps to Reproduce

Edit: since https://github.com/sqlfluff/sqlfluff/pull/2438 the `INTEGER` is detected. Replace it with any non-existing pseudo type to reproduce this issue again.

`lint_placeholders.sql`:
```sql
SELECT my_col
FROM my_schema.my_table
WHERE processdate = CAST(
        event_date AS INTEGER
    )
    AND eventtype = &{event_type}
```

```
$ sqlfluff lint lint_placeholders.sql
== [lint_placeholders.sql] FAIL
L:   4 | P:  70 |  PRS | Line 4, Position 70: Found unparsable section: ' AS
| INTEGER'
All Finished 📜 🎉!

$ echo $?
65
```

That is as expected, because `INTEGER` is not listed as a keyword in Hive dialect at the moment.

```
$ sqlfluff fix -f lint_placeholders.sql
All Finished 📜 🎉!

$ echo $?
0
```

That is unexpected. I was expecting this to also fail and complain about the PRS error.

## Dialect

Hive, but this issue doesn't seem to be dialect specific per se.

## Version

sqlfluff, version 0.9.1

Python 3.7.5

(not using dbt)

## Configuration

`.sqlfluff`:

```ini
[sqlfluff]
dialect = hive
templater = placeholder

param_style = ampersand
event_type='my_event'
```

barrywhart:

Makes progress on #2345 

At the moment we don't exit error on parsing failure as there's some debate amongst the maintainers whether this is the right thing to do.

### Pull Request checklist

  - Other.


src/sqlfluff/cli/commands.py
```diff
@@ -747,19 +748,22 @@ def fix(
...
-
-    if result.num_violations(types=SQLTemplaterError) > 0:
-        click.echo(
-            f"  [{result.num_violations(types=SQLTemplaterError)} templating errors "
-            "found]"
-        )
-        exit_code = 1
+    fatal_error_types = [
+        (
+            dict(types=SQLLintError, fixable=False),
```
tunetheweb, line 753: Is this a bit strict?

A couple of questions:
- If I have 5 errors and 4 can be fixed currently we benefit from this and have to fix the other one manually, but will this now mean we have to fix all 5 manually?
- - I presume it excludes rules that don’t implement a fix routine?

tunetheweb: Or will it still save the fixes it could do but exit error? If so that’s probably ok for the first point.

barrywhart: This doesn't change the **behavior** with respect to fixes, i.e. if it fixed them before, it will still fix them. All this does it print a message and change the exit code to "failure". TBH, this might be seen as a bit weird, since it made changes _and_ reported failure, but note this is what it already did with the other two error types (non-fixable `SQLLintError`, `SQLTemplaterError`). We're just adding a third type with the same behavior.

barrywhart: If you look at the old versus new code, you'll see what I mean. I took two blocks of near-identical code and turned them into a loop, eliminating the duplicate code while preserving the same behavior.

tunetheweb: Ok then I’m happy. The (very slight) refactoring threw me as only had a quick look on the phone.

Will not approve or merge yet and see if anyone else has opinions.

barrywhart: @tunetheweb: Understood. I couldn't resist the refactoring -- I generally have a "hard" limit of two copies of the same code. 😛

I tagged a few more people for review. I may be away from the computer today and tomorrow. Feel free to merge once you feel comfortable with the change. As I implied above, I don't think this is necessarily the best or right behavior, but it's probably the most straightforward change that's in line with current handling of other "errors".

jpy-git: > NOTE: If someone is using pre-commit, I'm pretty sure this will prevent them committing code with a parse error. Is that a good thing?

That would be pretty frustrating, it's not a users fault if the code is unparsable and they would likely remove the pre-commit hook because of this. SQLFluff should be lenient on unparsable files i.e. don't try and lint/fix them because it's not reliable but also don't block them from being committed.

There's always going to be a bit of a catch 22 in that, since we don't have complete grammar (and likely never will due to the complexity and variance between dialects), we can't tell if the code is unparsable because we can't parse it or if the user has actually tried to commit nonsense code. But I think the pragmatic approach is to assume generally the user is going to be writing sensible code (especially with SQL since people tend to write and test their code against a db as they go) and that it's likely SQLFluff who can't parse it, so we should just leave the unparsable code be.

tunetheweb: The counter argument is “why did SQLFluff allow me to commit bad code? What’s the point of having that as a pre-commit if it doesn’t warn about that?”

This is where force flag could be useful so user can choose whether to block commit or not.

> it's not a users fault if the code is unparsable

You’re assuming all unparsable code is due to SQLFluff errors or unimplemented logic. We have a biased view on that because we see bugs raised, but it could well be just bad SQL that IS the users “fault”. Those would be fixed by the user without being raised to us so we never see those. Hence the bias view we have.

tunetheweb: Sorry, missed your last paragraph where you explained that assumption. While I get that I think that’s quite a negative view of SQLFluff and if someone is going to use it in their code base (especially with an auto fix pre-commit) then they should ensure it has a good level of coverage for their code. That is, they should lint their whole code base, raise issue for missing functionality, open PRs if they want them added, add `noqa`‘s to their code in the meantime, and so unparsable code should be the rare exception rather than the norm that should be raised and handled with a force type option.

If it is not exception then they SHOULD disable the pre-commit IMHO and use SQLFluff fix with extreme care until they can identify and get fixed the main issues so it does become the exception.

Saying that, if we did go with the PR in its current state is this enough to make this a minor upgrade instead of a patch upgrade as we are changing behaviour here?

barrywhart: @tunetheweb: The behavior you're describing is what I would consider good behavior. However, I have worked with lots of data people (analysts, scientists) and a lot of them don't have strong coding or technical skills. They are often "over their heads" in terms of tools and coding. As such, we maintainers sometimes have to make some difficult judgement calls, balancing safety, usability, etc. Some of the absolute worst behavior I've seen occurs with data scientists is when they feel stuck (e.g. if `pre-commit` won't let them move ahead). They will throw every sort of discipline out the window just to keep moving, and they'll do it without letting anyone know or asking for help. As SQLFluff becomes more popular, the percentage of our users like this _increases_.

So I lean towards what I'd call "pragmatic safety": i.e., try and make the tool safe, but let the user know what's happening and try to avoid "blocking" them.

I posted a comment on the new issue I'd created describing what I think we should do: https://github.com/sqlfluff/sqlfluff/issues/2424#issuecomment-1019358328.

tunetheweb: I still don’t see the worst case here?

Is it that they override any pre-commit block, or remove pre-commit? Well I that’s no worse than not blocking the block in my opinion and at least they’re aware of it.

Or stop using SQLFluff cause it “cries wolf” too often? Well that’s a risk with any linter and will happen if they don’t understand the linter and it’s rules (or have someone to help manage this).

I come back to the use case of: “I installed a linter to prevent bad code getting in the code base, so it shouldn’t knowingly give up and allow bad code to be commuted without a very obvious warning and/or manual override.” Especially if it’s code that SQLFluff claims to have a rule to flag and prevent being committed.

Of course others might disagree with that approach and want a “best efforts” basis (which is what they have now) so failing but not blocking should potentially be an option, but so should failing and blocking IMHO. Which is the default is very much up for debate (and should be added to `.sqlfluffini` config file and comment line option).

barrywhart: In my previous job, worst case would be that they disabled or otherwise quit using it. Or they interrupted an engineer, telling them, this thing you told us to use is broken, so you need to get me past this.

The tool is a part of their job, but they have a bigger purpose in mind, so it's presumptuous of a tool to completely block them. I'm really intrigued by your suggestion to have parse errors block them, but that "noqa: PRS" would let them bypass the block. I'd like to think about that some more, but it sounds like a potentially great idea.  😎

tunetheweb: Well if they don’t see the value in the tool, and would rather turn it off cause it causes more issues than it solves, then they should strongly consider turning it off. Forcing a linter on people without buy in rarely works in my experience.

It’s also unrealistic IMHO Ato expect a linter to be fully automated and work seamlessly in the background. Also, (again IMHO) part of the value of a linter is to teach people better code style (black is more a code formatter than a linter so can run seamlessly in the background - not there is no mention of the word lint on its docs). SQLFluff is an opinionated linter that requires setup and understanding.

tunetheweb (approved):

PR #2430, opened 2022-01-22
Exasol: Fix `LOCAL.ALIAS` Syntax

sti0:

Currently the `LOCAL.ALIAS` syntax is parsed by `sqlfluff`. But it parses `LOCAL` as an identifier rather than a keyword. Furthermore there are things like `SELECT local.local FROM x` possible. This isn't valid code.

I added a new `LocalAliasSegment` to the `ansi` dialect as a hookpoint because it's necessary to add it to the `ExpressionSegment`. Also added various test cases for the different possibilities.

none

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -1706,6 +1706,7 @@ ansi_dialect.add(
                     Ref("DateTimeLiteralGrammar"),
                 ),
             ),
+            Ref("LocalAliasSegment"),
         ),
         Ref("Accessor_Grammar", optional=True),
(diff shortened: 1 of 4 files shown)


tunetheweb (approved):
LGTM!

Probably should consider adding this to rule L025's logic, but that wasn't in there anyway before so no worse off now.

sti0:
I don't think so. L025 meets table aliases. `LOCAL` only addresses columns. So this should be fine.

tunetheweb:
> I don't think so. L025 meets table aliases. `LOCAL` only addresses columns. So this should be fine.

OK good point!

PR #2434, opened 2022-01-23
T-SQL support default params and no RETURN value

### LINKED ISSUE #2370: return does not need a parameter
At least in T-SQL, `return` does not require a parameter: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/return-transact-sql?view=sql-server-ver15

## Expected Behaviour
```sql
return;
```
can be parsed

## Observed Behaviour
parsing fails

## Steps to Reproduce
1. Use `return` without a paramter somewhere.
2. Run `sqlfluff parse` on it

## Dialect
T-SQL

## Version
```
sqlfluff, version 0.9.1
Python 3.10.1
```

## Configuration
```
[sqlfluff]
dialect = tsql
```

### LINKED ISSUE #2372: tsql: Should be able to parse stored procedure declarations with default values for parameters
At least in T-SQL, you can define default values for a stored procedure's parameters (https://docs.microsoft.com/en-us/sql/relational-databases/stored-procedures/specify-parameters?view=sql-server-ver15#specifying-parameter-data-types). It would be nice if `sqlfluff` could parse and maybe even format them.

tunetheweb:

Fixes #2370 
Fixes #2372

I had to fix #2372 as part if #2370 as the example in the docs for no RETURN uses a default param 😁

Shouldn't be

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -1358,9 +1358,11 @@ class FunctionParameterListGrammar(BaseSegment):
     # Function parameter list
     match_grammar = OptionallyBracketed(
         Ref("FunctionParameterGrammar"),
+        Sequence(Ref("EqualsSegment"), Ref("ExpressionSegment"), optional=True),
```
fdw, line 1361: Out of curiosity (as I'm trying to learn more about `sqlfluff`): Why don't you add the new sequence to the `FunctionParameterGrammar`?

tunetheweb: There's usually many ways achieving the same thing in SQLFluff 😁

That would have worked too. And to be honest I didn't really consider it, but you got me curious so had a look...

`FunctionParameterGrammar` is not currently defined in T-SQL (meaning it picks it up from ANSI as that's the base of the T-SQL dialects) so would have meant either overriding that in T-SQL, or adding it to ANSI (which means identifying it all dialects support this).

Overriding it's fine, but since we already override `FunctionParameterListGrammar` this is less of a change. Possibly it's technically more accurate to put in the `FunctionParameterGrammar` (we already have the data type in that, so arguable an optional default value is part of that parameter definition).

I'd imagine it's _fairly common_, so maybe there is a case for adding it to ANSI but I know BigQuery for example doesn't support it.

Anyway, seems easier to just add it here. But happy to hear reasons for moving it if you have them?

fdw: Originally, it was just my gut feeling that the default value is part of the parameter definition and not of the list 😉
But I definitely understand your logic of not wanting to overwrite more than is necessary.

Regarding a general change in ANSI: I don't think a default value is part of ANSI (but I can never find a definition of ANSI syntax), as MySQL doesn't seem to support it at all and Postgre also supports a `DEFAULT` keyword instead of `=`.

tunetheweb: OK you've persuaded me :-)

Also think that cleans up `FunctionParameterGrammar` for T-SQL anyway, as ANSI has `ANY TYPE` but from a quick google that's not valid in T-SQL so T-SQL should have it's own definition anyway.

barrywhart (approved):
Great!

PR #2455, opened 2022-01-24
Exclude WINDOW clauses from L054

### LINKED ISSUE #2453: Remove WINDOW functions clauses from Rule L054
SQLFluff is saying I’m violating L054, cause I’ve got implicit ORDER BY clauses, but EXPLICIT ones inside ORDER BY of my WINDOW Functions. I'd suggest that Window functions clauses should be excluded from this rule since, although feasible, it's unnatural to use IMPLICIT clauses within them.

```sql
SELECT 
    field_1
    , field_2
    , SUM(field_3) as field_3_total
    , SUM(field_3) OVER (ORDER BY field_1) AS field_3_window_sum
FROM table
GROUP BY 1, 2
ORDER BY 1, 2
```

tunetheweb:

Fixes #2453

Nope

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L054.py
+++ b/src/sqlfluff/rules/L054.py
@@ -5,6 +5,7 @@ from sqlfluff.core.rules.base import BaseRule, LintResult, RuleContext
 from sqlfluff.core.rules.doc_decorators import (
     document_configuration,
 )
+import sqlfluff.core.rules.functional.segment_predicates as sp
 
 
 @document_configuration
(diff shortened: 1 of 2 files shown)


src/sqlfluff/rules/L054.py
```diff
@@ -88,6 +91,11 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
         if not context.segment.is_type("groupby_clause", "orderby_clause"):
             return None
 
+        # Ignore Windowing clauses
+        for parent in context.parent_stack:
```
barrywhart, line 95: I think you could simplify this a bit with the functional API (it's good at eliminating loops):
```
if context.functional.parent_stack.select(sp.is_type(*_self._ignore_types)):
    return LintResult()
```

barrywhart (approved):
Looks good! One suggestion for a slight simplification; totally optional.

src/sqlfluff/rules/L054.py
tunetheweb: Done. Also changed the LintResult to return Memory like the others.

barrywhart (approved):

PR #2459, opened 2022-01-24
Ignore words for various rules

tunetheweb:
### Brief summary of the change made

As raised on Slack 0.9.2 introduced a lot of keywords to Snowflake, including a lot of single word short alias's for dates (e.g. `d` for `day`).

This PR introduces the concept of `ignore_words` to allow the rules that check identifiers to be ignored for certain "allowed words". This is a good feature for feature to have anyway as allows users to handle bugs without waiting for the next release or turning off a rule entirely.

I've also excluded single letter chars from L026 and did some clean up in there for performance.

Adds an optional `ignore_words` parameter to the following rules:
- L010 - Inconsistent capitalisation of keywords.
- L014 - Inconsistent capitalisation of unquoted identifiers.
- L029 - Keywords should not be used as identifiers (rule inherits this from L010).
- L030 - Inconsistent capitalisation of function names (rule inherits this from L010).
- L040 - Inconsistent capitalisation of boolean/null literal (rule inherits this from L010).

I've also made some further changes to L029:
- Exit early if not an identifier
- Exit if 1 character identifier (sometimes these are keywords - like `d` in Snowflake, but people expect to be able to use them)
- Cache cleaned up `ignore_words` list (similar to how L010 caches it's config).

### Are there any other side effects of this change that we should be aware of?
Shouldn't be

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #1486, opened 2021-10-01
Better parsing of DATEADD function

jpers36:
### Brief summary of the change made
Fixes #1465 

ANSI additions:
+DateAddFunctionNameSegment representing special function DATEADD()
+Unreserved keywords
+DatePartClause representing the DatePart parameter to DATEADD()
Change FunctionSegment to include DATEADD special function

ANSI unreserved keywords additions:
DATEADD
QUARTER
WEEK
WEEKDAY

Exasol additions:
+unchanged FunctionSegment from ANSI to override changed FunctionSegment

PostgreSQL additions:
+unchanged FunctionSegment from ANSI to override changed FunctionSegment
+Code within keywords to explicitly remove DATEADD keyword

TSQL additions:
Adjust FunctionSegment overriding ANSI standard to add special DATEADD() structure
+DatePartClause override ANSI standard to include TSQL-specific abbreviations

TSQL unreserved keywords added:
    "D",
    "DATEADD",
    "DAY",
    "DAYOFYEAR",
    "DD",
    "DW",
    "DY",
    "HH",
    "HOUR",
    "M",
    "MCS",
    "MI",
    "MICROSECOND",
    "MILLISECOND",
    "MINUTE",
    "MM",
    "MONTH",
    "MS",
    "N",
    "NANOSECOND",
    "NS",
    "Q",
    "QQ",
    "QUARTER",
    "S",
    "SECOND",
    "SS",
    "W",
    "WEEK",
    "WEEKDAY",
    "WK",
    "WW",
    "YEAR",
    "Y",
    "YY",
    "YYYY",

Test case changes:
ansi/select_simple_h.yml updated to recognize DATEADD as a keyword
tsql/functions_a.sql and yml updated to include DATEADD test case

PR #2464, opened 2022-01-25
Cleanup date logics and other rule fixes

tunetheweb:

Following on from #2458 I decided to check what other simple statements cause exceptions to be thrown by linting all the text fixtures:

```
sqlfluff lint   --dialect ansi test/fixtures/dialects/ansi
sqlfluff lint --dialect bigquery   test/fixtures/dialects/bigquery
sqlfluff lint --dialect exasol   test/fixtures/dialects/exasol
sqlfluff lint --dialect hive   test/fixtures/dialects/hive
sqlfluff lint --dialect mysql   test/fixtures/dialects/mysql
sqlfluff lint --dialect oracle   test/fixtures/dialects/oracle
sqlfluff lint --dialect postgres   test/fixtures/dialects/postgres
sqlfluff lint --dialect redshift   test/fixtures/dialects/redshift
sqlfluff lint --dialect snowflake   test/fixtures/dialects/snowflake
sqlfluff lint --dialect spark3   test/fixtures/dialects/spark3
sqlfluff lint --dialect sqlite   test/fixtures/dialects/sqlite
sqlfluff lint --dialect teradata   test/fixtures/dialects/teradata
sqlfluff lint --dialect tsql   test/fixtures/dialects/tsql
```

This led to a few discoveries:
- A bug in select.py and select_crawler.py for `test/fixtures/dialects/exasol/SelectStatement.sql` (`SELECT INVALID FOREIGN KEY...` example)
- A bug in L010 for `test/fixtures/dialects/tsql/select_date_functions.sql`
- A bug in L044 for `test/fixtures/dialects/ansi/with_update_statement.sql`

The L010 bug required the most change. It basically looks at these types:

```py
        ("type", "keyword"),
        ("type", "binary_operator"),
        ("type", "date_part"),
        ("type", "data_type_identifier"),
```

and expects them to be identifiers.

For `date_part` this is defined twice - but differently. First of all it's a RegexParser - which means these are direct identifers:

```pr
    # Ansi Intervals
    DatetimeUnitSegment=SegmentGenerator(
        lambda dialect: RegexParser(
            r"^(" + r"|".join(dialect.sets("datetime_units")) + r")$",
            CodeSegment,
            name="date_part",
            type="date_part",
        )
    ),
```

And secondly in a class:

```py
@ansi_dialect.segment()
class DatePartClause(BaseSegment):
    """DatePart clause for use within DATEADD() or related functions."""

    type = "date_part"

    match_grammar = OneOf(
        "DAY",
        "DAYOFYEAR",
        "HOUR",
        "MINUTE",
        "MONTH",
        "QUARTER",
        "SECOND",
        "WEEK",
        "WEEKDAY",
        "YEAR",
    )
```

The former produces a part tree like this:

```
[L: 22, P: 14]      |                                date_part:                    'dayofyear'
```

That is, the `date_part` is directly the value.
The latter one produces a parse try like this:

```
[L: 22, P: 14]      |                                date_part:
[L: 22, P: 14]      |                                    keyword:                  'dayofyear'
```

That is, the `date_part` is a parent group containing a keyword of the value.

So we have one type meaning two different things. And that is confusing L010.

This also meant each date part was defined in two segments:
- One in `datetime_units` set (which then is used in `DatetimeUnitSegment`).
- Again in `DatePartClause`. This one also requires date parts to be Keywords - which we know causes problems like those in #2459 

This is all unnecessary duplication (triplication if you include the keywords) and, in fact was only a relatively recent change from #1486 in October 2021.

I've therefore basically reverted these unnecessary parts of that change and pointed everything that was using `DatePartClause` to `DatetimeUnitSegment` instead. This resolves the bug in L010 but also means we don't need dateparts to be keywords for those dialects that didn't want them in there (Snowflake, BigQuery, and T-SQL at least).

- Date parts are no longer keywords in Snowflake, Bigquery, and T-SQL - as they weren't back in October.

### Pull Request checklist

  - Other.


tunetheweb:
Pulled out the non-date changes into their own PRs for easier review

src/sqlfluff/rules/L044.py
```diff
@@ -83,7 +83,6 @@ def _analyze_result_columns(self, query: Query):
         # Recursively walk from the given query (select_info_list) to any
         # wildcard columns in the select targets. If every wildcard evdentually
         # resolves to a query without wildcards, all is well. Otherwise, warn.
-        assert query.selectables
```
barrywhart, line 86: Was there a particular test case causing this assertion to fail? Curious because, as annoying as assertion failures are, it has detected a couple of bugs and may detect future ones. Rather than remove it, perhaps we should log an error so problems don't go undetected?

tunetheweb: Sorry that's fixed in #2468 not #2470

barrywhart (approved):

tunetheweb:
> The procedure you describe, where you ran all the rules on all the test cases sounds like it could be a great automated check to run on all builds. What do you think?

Already wrote it - #2473 😁

PR #2470, opened 2022-01-25
Fix select_crawler issue with some Exasol statements

tunetheweb:
### Brief summary of the change made

Fixes #2469

(Note originally part of #2464 but pulled out into it's own PR for easier review)

### Are there any other side effects of this change that we should be aware of?
Shouldn't be

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

Issue #2469, opened 2022-01-25
L026 and L044 throw exception on Exasol statement


tunetheweb:
## Expected Behaviour

This should lint:

```
echo "SELECT first_name,name WITH INVALID FOREIGN KEY (nr) from T1 REFERENCING T2 (id);" | sqlfluff lint --dialect exasol -
```

## Observed Behaviour

```
 -
CRITICAL   [L026] Applying rule L026 threw an Exception: 'NoneType' object has no attribute 'recursive_crawl'                                                    
Traceback (most recent call last):
  File "/Users/barry/sqlfluff/src/sqlfluff/core/rules/base.py", line 519, in crawl
    res = self._eval(
  File "/Users/barry/sqlfluff/src/sqlfluff/rules/L026.py", line 91, in _eval
    self._analyze_table_references(
  File "/Users/barry/sqlfluff/src/sqlfluff/rules/L026.py", line 118, in _analyze_table_references
    select_info = selectable.select_info
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/functools.py", line 967, in __get__
    val = self.func(instance)
  File "/Users/barry/sqlfluff/src/sqlfluff/core/rules/analysis/select_crawler.py", line 42, in select_info
    return get_select_statement_info(
  File "/Users/barry/sqlfluff/src/sqlfluff/core/rules/analysis/select.py", line 35, in get_select_statement_info
    reference_buffer = list(sc.recursive_crawl("object_reference"))
AttributeError: 'NoneType' object has no attribute 'recursive_crawl'
CRITICAL   [L044] Applying rule L044 threw an Exception: 'NoneType' object has no attribute 'recursive_crawl'                                                    
Traceback (most recent call last):
  File "/Users/barry/sqlfluff/src/sqlfluff/core/rules/base.py", line 519, in crawl
    res = self._eval(
  File "/Users/barry/sqlfluff/src/sqlfluff/rules/L044.py", line 143, in _eval
    return self._analyze_result_columns(crawler.query_tree)
  File "/Users/barry/sqlfluff/src/sqlfluff/rules/L044.py", line 89, in _analyze_result_columns
    for wildcard in selectable.get_wildcard_info():
  File "/Users/barry/sqlfluff/src/sqlfluff/core/rules/analysis/select_crawler.py", line 82, in get_wildcard_info
    if not self.select_info:
[...]

Issue #2467, opened 2022-01-25
L044 throws exception when final statement has no SELECT


tunetheweb:
With this SQL:

```sql
WITH mycte AS (
    SELECT
        foo,
        bar
    FROM mytable1
)

UPDATE sometable
SET
    sometable.baz = mycte.bar
FROM
    mycte;
```

## Expected Behaviour

SQL Lints

## Observed Behaviour

```
% sqlfluff lint test.sql
CRITICAL   [L044] Applying rule L044 threw an Exception:                                                                                                         
Traceback (most recent call last):
  File "/Users/barry/sqlfluff/src/sqlfluff/core/rules/base.py", line 519, in crawl
    res = self._eval(
  File "/Users/barry/sqlfluff/src/sqlfluff/rules/L044.py", line 145, in _eval
    return self._analyze_result_columns(crawler.query_tree)
  File "/Users/barry/sqlfluff/src/sqlfluff/rules/L044.py", line 86, in _analyze_result_columns
    assert query.selectables
AssertionError
```

## Steps to Reproduce

As per above

## Dialect

All

## Version

0.9.2

## Configuration
Default

tunetheweb:
Fixed in #2468

PR #2473, opened 2022-01-25
Add linting of fixtures SQL for critical rules errors to tox

tunetheweb:

We have a wealth of example SQL files, and we test them for parsing in various environments. However we don't lint them.

We have had some recent issues (see #2464, #2467, #2469) whereby certain SQL caused rule code to throw a critical error, mostly because the rule hasn't been built to consider certain scenarios.

This PR adds a script to lint all the `test/fixtures/dialects` files, but only complain if a CRITICAL error is thrown by SQLFluff (normal SQLFluff linting errors are thrown aside as most test cases are not written with all the rules in mind - and shouldn't be IMHO).

It's a simple bash script (it's what I know best!) and changes to tox and our CI GitHub Action. We could change it to a python script (in line with the rest of the code), but this'll do for now. Pus I can of like the fact it's calling SQLFluff by command line as it would. I'll leave one of the more experienced Python developers to take a conversion to Python on if they want!

Note this cannot be merged until #2464, #2468, #2470 are all merged as currently it (correctly!) highlights errors, that those were raised to fix.

I also added a .sqlfluff config file to each dialect directory to automatically set the dialect for that directory. This is automatically taken care of in our test suite, but think it'll be handy anyway for when running sqlfluff outside the test suite (like we do here).

At the moment our Windows CI takes the longest to run, so will time this, but suspect (based on local testing on my slow Mac) it'll be quicker than that so shouldn't add to CI run times.

### Pull Request checklist

  - Other.


tunetheweb:
Confirmed it takes 4 mins to run, so won't hold up CI.

pwildenhain (commented):
Looks like a great addition

test/test-lint-rules.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+DIALECTS=($( ls -d test/fixtures/dialects/*/ ))
+for DIALECT in "${DIALECTS[@]}"
+do
+    echo "Testing $DIALECT SQL files lint without critical errors..."
+    OUTPUT=$(sqlfluff lint ${DIALECT})
```
barrywhart, line 7: Should we (or do we already) check the `sqlfluff` exit code here as well?

barrywhart (approved):
So beautiful, I'm almost in tears. Very nice addition!!

One small question/suggestion, otherwise 💯.

barrywhart:
@tunetheweb: It looks like this is not currently a "required" CI check. Can you make it required, or does @alanmcruickshank have to do that?

test/test-lint-rules.sh
tunetheweb: We can't. There are LOTS of linting errors (fixtures aren't written to follow all the rules). So it mostly exists with exit code 65. But that's not to say it must. A new dialect could be written with one perfect test case which will exit with 0.

So let's leave as unchecked.

tunetheweb:
> @tunetheweb: It looks like this is not currently a "required" CI check. Can you make it required, or does @alanmcruickshank have to do that?

Only @alanmcruickshank can do that. Let's discuss in this afternoon's meeting.

tunetheweb:
> Let's discuss in this afternoon's meeting.

Sorry next week's meeting :-)

barrywhart (approved):

NiallRees:
Would definitely be good to get this in Python so that it can be run by Windows users, but not a blocker!

PR #2489, opened 2022-01-26
Enforce parentheses for function definitions in T-SQL

### LINKED ISSUE #2472: T-SQL: CREATE FUNCTION needs parentheses even if there are no parameters
## Expected Behaviour
```sql
create or alter function test() returns int as
begin
  return 2;
end
```
is correct T-SQL and should work.

## Observed Behaviour
`sqlfluff` cannot parse the statement, as it either expect at least one parameter in the parentheses, or no parantheses at all (which is not valid T-SQL).

## Steps to Reproduce
See above

## Dialect
T-SQL

## Version
Latest `sqlfluff` from git
Python 3.10.1

## Configuration
```
[sqlfluff]
dialect = tsql
```

fdw:

In T-SQL, function definitions must have parentheses, even if they have no parameters.
On the other hand, procedures can have parentheses around parameters, but they don't need to.
To correctly lint this difference, they now use different segments.

Fixes #2472

I don't think so

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_tsql.py
+++ b/src/sqlfluff/dialects/dialect_tsql.py
@@ -1414,12 +1414,15 @@ class FunctionParameterListGrammar(BaseSegment):
 
     type = "function_parameter_list"
     # Function parameter list
-    match_grammar = OptionallyBracketed(
-        Ref("FunctionParameterGrammar"),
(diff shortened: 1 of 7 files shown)


tunetheweb (approved):
LGTM!

PR #2505, opened 2022-01-28
Spark3: Support for Grouping Sets, `CUBE` and `ROLLUP` in `GROUP BY` clause of `SELECT` statement

R7L208:

 - replace `GroupByClauseSegment` from ANSI to allow for `CUBE` and `ROLLUP`
 - created `WithCubeRollupClauseSegment` to handle the `WITH CUBE | WITH ROLLUP` syntax that occurs after a `GROUP BY`
 - Created `CubeRollupClauseSegment` based on Snowflake dialect to handle `CUBE | ROLLUP` inside of the `GROUP BY` clause
 - Created `GroupingSetsClauseSegment` based on Snowflake dialect to handle Grouping Sets within a `GROUP BY` clause
 - Created `GroupingExpressionList` for `CUBE`, `ROLLUP`, and Grouping Sets. Also based on existing class in Snowflake dialect.

Based on a quick search:
- Cubes appear in Exasol, Postgres, Snowflake, and now Spark as implemented segments but are also listed in Hive, Ansi and BigQuery as Keywords.
- Rollups are implemented in Snowflake, Exasol, and Spark.
- Grouping Sets are also implemented in Snowflake, Exasol, and Spark.

It appears that segments (`CubeRollupClauseSegment`, `GroupingSetsClauseSegment`, `GroupingExpressionList`) from Spark, Snowflake, and Exasol are all copy/paste.

I think it makes sense to move these segments and needed grammar into Ansi dialect but would appreciate a second opinion before I do the lift & shift.

`WithCubeRollupClauseSegment` is unique to Spark3 for now, so that could be kept in Spark3 dialect.

### Pull Request checklist

  - Other.


R7L208:
Seems that BigQuery only supports ROLLUP at the moment.
https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#group_by_clause

Hive Reference.
https://cwiki.apache.org/confluence/display/Hive/LanguageManual+GroupBy#LanguageManualGroupBy-GroupingSets,Cubes,Rollups,andtheGROUPING__IDFunction

Exasol Reference (but need to scroll to down for cube/rollup/grouping sets)
https://docs.exasol.com/sql/select.htm

Snowflake docs.
https://docs.snowflake.com/en/sql-reference/constructs/group-by-cube.html
https://docs.snowflake.com/en/sql-reference/constructs/group-by-rollup.html
https://docs.snowflake.com/en/sql-reference/constructs/group-by-grouping-sets.html

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -1086,6 +1085,143 @@ class SelectStatementSegment(BaseSegment):
...
+    """
+
+    type = "with_cube_rollup_clause"
+
+    match_grammar = StartsWith(
+        Sequence(
+            "WITH",
+            OneOf("CUBE", "ROLLUP"),
+        ),
+        terminator=OneOf(
```
tunetheweb, line 1141: Why do you need this? It’s a simple enough statement so shouldn’t be confused with the next statement (which is the usual reason for adding a terminator clause).

So can this `match_clause` and `parse_clause`s be merged into one simple `match_clause`?

tunetheweb (commented):
Good set of test case!

I do have one question though.

src/sqlfluff/dialects/dialect_spark3.py
R7L208: yes, I'll make that update. They can be merged

tunetheweb (approved):

PR #2506, opened 2022-01-28
Support NATURAL JOINS

### LINKED ISSUE #2497: Include NATURAL as a keyword for joins
Currently, NATURAL joins are unparsable. I'd suggest including it as an optional keyword that precedes INNER/LEFT/RIGHT/FULL.
A more complex thing to check is that NATURAL joins don't support a join condition (it's implicit), so ON or USING aren't valid words in this context.

tunetheweb:

Fixes #2497 

Updated ANSI, but did check all the other dialects (overrides for BigQuery and TSQL as they don't support this).

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1356,58 +1371,69 @@ class JoinClauseSegment(BaseSegment):
...
-                    "LEFT",
-                    "RIGHT",
-                ),
-                Ref.keyword("OUTER", optional=True),
-            ),
-            optional=True,
-        ),
-        Ref("JoinKeywords"),
-        Indent,
         Sequence(
```
tunetheweb, line 1379: Use Hide Whitespace to review this and it'll look at lot smaller as this section was indented to allow `OneOf` above for the NATURAL join section below.

image

src/sqlfluff/dialects/dialect_bigquery_keywords.py
```diff
@@ -58,7 +58,6 @@
 LIMIT
 LOOKUP
 MERGE
-NATURAL
```
tunetheweb, line 61: This was down as an unreserved word in BigQuery (I think I copied this from ANSI when setting up the BigQuery word list) but is not needed. Test case added to ensure it doesn't fail in future.

barrywhart (approved):

PR #2510, opened 2022-01-29
Refactor date part functions

### LINKED ISSUE #2508: L010: SQLFluff 0.9.1 (and later) capitalize columns named year/month when used as the second parameters in PARSE_DATE
## Expected Behaviour
When run on a SQL file containing a line like PARSE_DATE("%Y", year), SQLFluff should realize that year must be a variable rather than an SQL keyword this context and not request that it is capitalized.

## Observed Behavior
For this function year and month is treated like SQL keywords rather than column names when used in the second argument.

## Steps to Reproduce
(1). Generate a SQL file of the form:
```sql
SELECT
    PARSE_DATE("%Y%m", CONCAT(year, month)) AS date
FROM
    model
```
(2). Set BigQuery as the dialect and run sqlfluff lint on this file.

Output is:
```
L:   1 | P:  34 | L010 | Keywords must be upper case.
L:   1 | P:  40 | L010 | Keywords must be upper case.
```

## Dialect
BigQuery

## Version
SQLFluff version is 0.9.3 (although this is not an issue in SQLFluff 0.9.0).  Python version is 3.8.1.0.
(I'm using DBT but the same issue occurs when not using dbt as a template).

## Configuration
```
[sqlfluff]
dialect = bigquery

[sqlfluff:rules]
max_line_length = 120
comma_style = leading

[sqlfluff:rules:L010]
capitalisation_policy = upper

[sqlfluff:rules:L030]
capitalisation_policy = upper
```

tunetheweb:

Fixes #2508 

This PR refactors all date part functions with the following changes:
- BigQuery changes from datepart values being seen as date parts EXCEPT for date part functions (e.g. `myFunction(year)` will assume `year` is an identifier, while `date_add(year, my_date)` will assume `year` is a datepart. These are now inline with ANSI and the other dialects that already did this.
- Remove date functions as keywords and handle as set instead for all dialects. This allows them to follow function rules (L017 and L030) and also is now inline with non-date function names from a parse tree sense.
- Removes unnecessary function overrides for Postgres and Redhshift

As per above.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1044,6 +1054,9 @@ class FunctionSegment(BaseSegment):
     type = "function"
     match_grammar = OneOf(
         Sequence(
+            # Treat fucnctions which take date parts separately
```
barrywhart, line 1057:
```suggestion
            # Treat functions which take date parts separately
```

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -365,23 +387,44 @@ class FunctionSegment(BaseSegment):
...
-    match_grammar = Sequence(
+    match_grammar = OneOf(
         Sequence(
-            AnyNumberOf(
-                Ref("FunctionNameSegment"),
-                max_times=1,
-                min_times=1,
-                exclude=OneOf(
-                    Ref("ValuesClauseSegment"),
+            # Treat fucnctions which take date parts separately
```
barrywhart, line 392:
```suggestion
            # Treat functions which take date parts separately
```

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -365,23 +387,44 @@ class FunctionSegment(BaseSegment):
...
+            # So those functions parse date parts as DatetimeUnitSegment
+            # rather than identifiers.
+            Sequence(
+                Ref("DatePartFunctionNameSegment"),
```
barrywhart, line 396: Does the parser know (and does it need to know) which parameter(s) of these functions are date parts? I.e. parameter 1, parameter 2, whatever? Or will it accept a date part for *any* parameter of these functions?

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -365,23 +387,44 @@ class FunctionSegment(BaseSegment):
...
+                Bracketed(
+                    Delimited(
+                        Ref("DatetimeUnitSegment"),
+                        # Ref("IntervalExpressionSegment"),
```
barrywhart, line 400: Can this commented-out code be removed?

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -1907,6 +1912,9 @@ class FunctionSegment(BaseSegment):
     type = "function"
     match_grammar = OneOf(
         Sequence(
+            # Treat fucnctions which take date parts separately
```
barrywhart, line 1915:
```suggestion
            # Treat functions which take date parts separately
```

src/sqlfluff/dialects/dialect_bigquery.py
tunetheweb: No it does not need to know.

All this does is say any DatePart function has `"DatePartFunctionNameSegment"` as a higher precedence than the rest of the `"FunctionContentsGrammar"`.

Some functions (e.g. `EXTRACT`) expect it in first place, some (e.g. `DATE_TRUNC`) in second position, and some (e.g. `DATE_DIFF`) in the third position.

The important point is not to exclude any of those, but just to say "if you see something that looks like a datepart, then treat it as such, otherwise carry on and parse it as a `"FunctionContentsGrammar"`.

This is the way ANSI currently handles it.

This does, however, mean `SELECT EXTRACT(year, year) FROM table1` will incorrectly assume both are date parts (when actually it's only the first part and this really should be like `SELECT EXTRACT(year, t.year) FROM table1 t`). We could setup different grammar for each type (1st, 2nd, 3rd) to avoid this but think that adds a lot of complexity for little gain. Arguable in above instance the second `year` should be qualified with a table alias to make it clear.

src/sqlfluff/dialects/dialect_bigquery.py
tunetheweb: Yup. It's gone.

barrywhart (approved):

Issue #2512, opened 2022-01-30
Tests failing due to Black errors


tunetheweb:
## Expected Behaviour

Valid Python parses.

## Observed Behaviour

Lots of errors like this:

```
./setup.py:0:1: BLK999 Unexpected exception: unsupported operand type(s) for /: 'tuple' and 'str'
./util.py:0:1: BLK999 Unexpected exception: unsupported operand type(s) for /: 'tuple' and 'str'
```

## Steps to Reproduce

This is due to an incompatibility with the last Black (ironically the first "stable" version of Black) and Flake8-black as raised here: https://github.com/peterjc/flake8-black/issues/44

tunetheweb:
We froze the Black instance in #2513 but leaving this issue open as we should unfreeze (potentially when a decision is made on #2514 as older DBT versions seem to require an older version of Black).

PR #2513, opened 2022-01-30
Freeze Black due to incompatibility between 22.1 and flake8-black

tunetheweb:

Makes progress on #2512 
We will not close that as "fixed" since this should be a temporary change.

Nope.

### Pull Request checklist

  - Other.

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -2,7 +2,9 @@
 # code linting and formatting
 flake8
 flake8-docstrings
-flake8-black>=0.2.4
+# Black 22.1 has a fatal bug when used with flake8-black, so freeze it for now.
+black<=21.12b0
+flake8-black
 # documentation checks
 doc8
 Pygments
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -2,9 +2,7 @@
 # code linting and formatting
 flake8
 flake8-docstrings
-# Black 22.1 has a fatal bug when used with flake8-black, so freeze it for now.
-black<=21.12b0
-flake8-black
+flake8-black>=0.2.4
 # documentation checks
 doc8
 Pygments
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -3,7 +3,8 @@
 flake8
 flake8-docstrings
 # Black 22.1 has a fatal bug when used with flake8-black, so freeze it for now.
-black<=21.12b0flake8-black
+black<=21.12b0
+flake8-black
 # documentation checks
 doc8
 Pygments
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -2,7 +2,8 @@
 # code linting and formatting
 flake8
 flake8-docstrings
-flake8-black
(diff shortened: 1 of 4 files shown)


barrywhart (approved):

src/sqlfluff/rules/L003.py
```diff
@@ -882,15 +882,11 @@ def segment_info(idx: int) -> Tuple[str, Optional[str]]:
...
-                    ("placeholder", "compound"),
-                    ("placeholder", "block_mid"),
-                )
-                and segment_info(idx) == ("indent", None)
-            ):
+            if segment_info(idx - 1) in (
+                ("placeholder", "block_start"),
+                ("placeholder", "compound"),
+                ("placeholder", "block_mid"),
+            ) and segment_info(idx) == ("indent", None):
```
tunetheweb, line 889: OK this is an interesting one. Black 22.1.0 prefers this without the outer brackets.
Versions less than that prefer it with.

Unfortunately dbt needs an. older version of black so we can't pin it to 22.1.0 to ensure this doesn't revert in future.

We could pin to `<=21.12b0` but that feels kind of wrong and limiting for the future.

So currently it all works, but if a user has an older version of black (e.g. because they use dbt) then they might try to revert this change. That will fail in our CI and may lead to confusion.

But can't think of a better solution.

barrywhart: If we want, we could install Black in a separate Python environment from dbt. At my last job, I eventually identified and separated a number of tools in this way -- they didn't have to be in the same Python environment because they didn't `import` or otherwise call other packages, they just did "system" things such as reading from files, writing to the terminal etc. This included most of the linting kind of tools, but not the ones that ran automated tests (e.g. pytest).

tunetheweb: We already do for linting (tox sets up it's own env). The problem is dbt installs it. So if using that env it will use that black. And some people (including you!) don't use tox.

Then again dbt tests seem to be freezing anyway. So might might end up freezing black anyway!

tunetheweb: OK I'm sticking with pinning the older version of black for now. Only thing that seems to work.

I'll pin flake8-black for now too for when we unpin black.

barrywhart:
Looks like the dbt tests finished!

requirements_dev.txt
```diff
@@ -2,7 +2,11 @@
 # code linting and formatting
 flake8
 flake8-docstrings
-flake8-black
+# Black 22.1 has a fatal bug when used with flake8-black and dbt, so freeze it for now.
```
barrywhart, line 5: Nitpick: The comment says we're freezing versions, but we're not freezing (i.e. requesting a specific version). Rather, we're specifying minimum or maximum versions.

barrywhart:
@tunetheweb: The build succeeded! Shall we merge?

requirements_dev.txt
tunetheweb: Just fixed to "cap" and included a link to this issue.

PR #2521, opened 2022-01-30
Use the slower but safer "spawn" method to create multiprocessing.Pool

barrywhart:
@tunetheweb: I think this may avoid the issues you've described where the coverage check sometimes hangs. This PR changes the behavior both for tests and actual usage. It will be a bit slower spawning subprocesses, but since linting takes a long time (minutes), this probably isn't a big concern.

Is there some way we can test to confirm this change is helpful?

Relevant docs:
* https://stackoverflow.com/questions/43818519/what-is-the-meaning-of-context-argument-in-multiprocessing-pool-pool/43818829
* https://python.readthedocs.io/en/stable/library/multiprocessing.html#contexts-and-start-methods

### Pull Request checklist

  - Other.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -77,7 +77,7 @@ jobs:
         echo 'EOF' >> $GITHUB_ENV
     - name: Run the tests
       run: |
-        tox -e ${{ env.TOXENV }} -- -n 2 test
+        tox -e ${{ env.TOXENV }} -- test
     - name: Upload Coverage Report
       uses: codecov/codecov-action@v1
       with:
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -13,7 +13,7 @@ flake8-black>=0.2.4
 doc8
 Pygments
 # testing
-coverage==6.2
+coverage
 hypothesis
 pytest
 pytest-cov
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -13,7 +13,7 @@ flake8-black>=0.2.4
 doc8
 Pygments
 # testing
-git+https://github.com/nedbat/coveragepy@nedbat/fix-1310
+coverage
 hypothesis
 pytest
 pytest-cov
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -13,7 +13,7 @@ flake8-black>=0.2.4
 doc8
 Pygments
 # testing
-coverage
+git+https://github.com/nedbat/coveragepy@nedbat/fix-1310
 hypothesis
 pytest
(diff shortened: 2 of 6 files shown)


tunetheweb:
Once the run completes can we try unpinning coverage to see if it fixes that issue too?

barrywhart:
@tunetheweb: Any idea why I'm seeing "cancelled" for the "Rules critical errors tests"?! It's happened twice for this PR.

image

tunetheweb:
Usually implies another failure in that test group so no point continuing other tests. Though thought I’d put that in its own group…

barrywhart:
The check runs fine locally on my Mac M1.

image

tunetheweb:
I looked through all the past tests for the PR bad that check never failed. Sure you weren’t confusing it with something else?

barrywhart:
@tunetheweb: Oh sorry -- I meant that it failed twice for **this** PR (two separate commits).

barrywhart:
I temporarily disabled it. I want to test the coverage stuff, then can re-enable it with more logging if needed.

barrywhart:
I can understand "failing", but "cancelled" implies that a human did it, or that it triggered some sort of policy/security issue or ???

tunetheweb:
Oh it’s also cancelled if you submit a new commit as pointless completing that one o e. Then need to refresh this PR page to see latest one. Maybe it was that? Maybe that job cancels quicker than. The others so that’s why you noticed it?

barrywhart:
Hmm, I didn't think I had submitted a new commit, but maybe. I'll pay closer attention now.

barrywhart:
@tunetheweb: Looks like the 3.7 tests are still hanging, even with this change. It's odd, because AFAIK, I made the exact change mentioned in the "coverage" issue, i.e. using "spawn" model to create new processes. TBH, I have not used this process setting before, so I _could* have messed it up, but it seems pretty straightforward.

Since you said this mainly (only?) happens in Python 3.7, think we could use a newer Python version for checking coverage? The project has very little if any version-specific code, so there's not much additional benefit to checking coverage on _every_ supported version of Python.

tunetheweb:
We could to that. Feels less than satisfactory though.

It would be worth trying your branch with their latest patch. Replace this line in requirements_dev.txt:
`coverage`
with this:
`git+https://github.com/nedbat/coveragepy@nedbat/fix-1310`
Which will install their patched version of coverage instead.

barrywhart:
I'll leave this open in case we want to try more experiments. So far, **nothing** has helped. 😿

tunetheweb:
Closing this for now as Coverage are continuing to investigate.

PR #2526, opened 2022-01-31
Fix Quoted Literals for Postgres and Redshift affecting rule L039

### LINKED ISSUE #2495: Postgres: Support for multiline string literals causes incorrect parsing behaviour and false warnings from L039
## Expected Behaviour

I expect the following query to lint successfully for the Postgres dialect
```sql
SELECT
    a,
    CASE
        WHEN 1 THEN 'one'
        WHEN 2 THEN 'two'
        ELSE 'other'
    END AS b
FROM test;
```

## Observed Behaviour
```shell
sqlfluff lint -n sample.sql --dialect postgres
== [sample.sql] FAIL                                                                                                                                                         
L:   5 | P:   1 | L039 | Unnecessary whitespace found.
L:   6 | P:   1 | L039 | Unnecessary whitespace found.
L:   7 | P:   1 | L039 | Unnecessary whitespace found.
```
Running `sqlfluff fix -n sample.sql --dialect postgres` yields an output that doesn't pass linting.

From this [thread](https://sqlfluff.slack.com/archives/C01GX4L213J/p1643299089547419), it seems this is an unintended bug from #2356.

The above query lints successfully for the `ansi` dialect.

## Dialect
postgres

## Version
0.9.3
master

## Configuration
Default

### LINKED ISSUE #2516: Create String Constant Segment for Postgres DIalect
In our Postgres Dialect, A String Constant (defined [here](https://www.postgresql.org/docs/9.3/sql-syntax-lexical.html) in 4.1.2.1-4) can be represented in our grammar using the following:

```py
                OneOf(
                    Ref("QuotedLiteralSegment"),
                    Ref("DollarQuotedLiteralSegment"),
                ),
```

To best modularize and maintain string constants in our parser, it might make sense for us to:
1. Create a `StringConstantSegment` class in our Postgres Dialect that has a `match_grammar` that is defined as above.
2. Refactor implicit references to string constants in our Postgres Dialect to explicit references to the `StringConstantSegment` class.

_Originally posted by @tunetheweb in https://github.com/sqlfluff/sqlfluff/pull/2511#pullrequestreview-867061366_

tunetheweb:

Fixes #2495 - prevent final comma being swallowed up
Fixes #2516 - move DollarQuotedLiteralSegment into QuotedLiteralSegment

A potential nicer fix to #2495 would be to have an alternative to `allow_trailing` option, which doesn't error when there is a trailing delimiter, but recognised this is not part of this Grammar. But had a quick look and can't figure out how to do that.

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -240,15 +240,24 @@ postgres_dialect.replace(
         # Postgres allows newline-concatenated string literals (#1488).
         # Since these string literals can have comments between them,
         # we use grammar to handle this.
-        Delimited(
+        # Note we CANNOT use Delimited as it's greedy and swallows the
+        # last Newline - see #2495
+        Sequence(
             NamedParser(
                 "single_quote",
                 CodeSegment,
                 name="quoted_literal",
                 type="literal",
             ),
-            delimiter=Ref("MultilineConcatenateDelimiterGrammar"),
-            allow_trailing=True,
+            AnyNumberOf(
+                Ref("MultilineConcatenateDelimiterGrammar"),
+                NamedParser(
+                    "single_quote",
+                    CodeSegment,
+                    name="quoted_literal",
+                    type="literal",
+                ),
+            ),
         ),
         Delimited(
             NamedParser(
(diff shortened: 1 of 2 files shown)


barrywhart (approved):
Looks great!

jpy-git:
> A potential nicer fix to #2495 would be to have an alternative to `allow_trailing` option, which doesn't error when there is a trailing delimiter, but recognised this is not part of this Grammar. But had a quick look and can't figure out how to do that.

I actually tried this the other day and it worked fine in terms of not including but not failing on a trailing delimiter and when I tested the grammar on it's own it would work. But whenever I tried parsing some actual SQL it would still get included in the expression 🤷  Your solution works well though 👍

tunetheweb:
Glad I didn't work too hard on that then! 😁

Would be nice to look at `Delimted` at some point to stop being sooo Greedy (e.g. ignore commas in brackets), and also for this use case.

Issue #2390, opened 2022-01-21
sqlfluff-lint requires for `dbt deps` to be run when using `pre-commit`


robertdefilippi:
## Expected Behaviour

When using `sqlfluff lint` with `pre-commit`, the linter should go through changed files and show the issues with the file. However, an error is being thrown that `dbt deps` needs to be run. 

## Observed Behaviour

```
# editing a sql file, then attempting to commit to a repository

$ git commit "some commit message"

Trim Trailing Whitespace.................................................Passed
Check for added large files..............................................Passed
Check Yaml...........................................(no files to check)Skipped
Fix End of Files.........................................................Passed
Mixed line ending........................................................Passed
Detect Private Key.......................................................Passed
sqlfluff-lint............................................................Failed
- hook id: sqlfluff-lint
- exit code: 1

=== [dbt templater] Sorting Nodes...
02:07:34  [WARNING]: Deprecated functionality
The `source-paths` config has been renamed to `model-paths`. Please update your
`dbt_project.yml` configuration to reflect this change.
02:07:34  [WARNING]: Deprecated functionality
The `data-paths` config has been renamed to `seed-paths`. Please update your
`dbt_project.yml` configuration to reflect this change.
Traceback (most recent call last):
  File "/Users/robertdefilippi/.cache/pre-commit/repolmlyjftv/py_env-python3.8/bin/sqlfluff", line 8, in <module>
    sys.exit(cli())
  File "/Users/robertdefilippi/.cache/pre-commit/repolmlyjftv/py_env-python3.8/lib/python3.8/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/Users/robertdefilippi/.cache/pre-commit/repolmlyjftv/py_env-python3.8/lib/python3.8/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/Users/robertdefilippi/.cache/pre-commit/repolmlyjftv/py_env-python3.8/lib/python3.8/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/robertdefilippi/.cache/pre-commit/repolmlyjftv/py_env-python3.8/lib/python3.8/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #2530, opened 2022-01-31
Add additional documentation on dbt-adapter in pre-commit

robertdefilippi:
Given the comments on [this issue](https://github.com/sqlfluff/sqlfluff/issues/2390) I've added additional documentation on how you can customize adapters

None

### Pull Request checklist

No tests to write, as just the documentation was updated

### DIFF
--- a/docs/source/production.rst
+++ b/docs/source/production.rst
@@ -130,7 +130,8 @@ When trying to use the `dbt templater`_, uncomment the
 ``additional_dependencies`` to install the extras.
 This is equivalent to running ``pip install <dbt-adapter> sqlfluff-templater-dbt``.
 
-You're able to specify the verion of ``dbt-adapter`` used in `pre-commit`_ E.g. ``additional_dependencies : ['dbt-bigquery==1.0.0', 'sqlfluff-templater-dbt']``
+You can specify the verion of ``dbt-adapter`` used in `pre-commit`_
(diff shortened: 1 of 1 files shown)


docs/source/production.rst
```diff
@@ -130,6 +130,11 @@ When trying to use the `dbt templater`_, uncomment the
 ``additional_dependencies`` to install the extras.
 This is equivalent to running ``pip install <dbt-adapter> sqlfluff-templater-dbt``.
 
+You can specify the verion of ``dbt-adapter`` used in `pre-commit`_
+E.g. ``additional_dependencies : ['dbt-bigquery==1.0.0', 'sqlfluff-templater-dbt']``
```
tunetheweb, line 134: Looks a bit messy when I run the site:

image

```suggestion
You can specify the verion of ``dbt-adapter`` used in `pre-commit`_,
for example:

.. code-block:: yaml

   additional_dependencies : ['dbt-bigquery==1.0.0', 'sqlfluff-templater-dbt']
```

Which looks like this:

image

It's true it does overflow with a scrollbar but so does the section above. I think this looks better.

WDYT?

tunetheweb (changes requested):
Thanks for this!

I've suggested a small change to format it better.

docs/source/production.rst
robertdefilippi: Agreed. I did not think about doing a code-block like that.

tunetheweb (approved):

PR #2537, opened 2022-02-01
Allow L059 to be configured to always prefer quoted identifiers

### LINKED ISSUE #2527: L059: Allow prefered quoted for identifiers
The current rule L059 is raising an error when some unnecessary quoted identifier has been detected.

Although it is not recommended to quote identifiers when it's not needed, this is a convention I am using for automatic business script that are parsing my SQLs. Therefore, I would like to use the backtick as a mandatory wrapper for all identifiers, no matter if they need to be quoted.

I could ignore the rule L059, but I would prefer to let the user allow, through the configuration, which quote to use and have a configuration to actually force it everywhere.

I propose 2 configurations for the L059 rule:
- force_quote_identifier: [True,False] (Defaults: False) # Force every identifiers to be quoted, even when unnecessary if True.
- preferred_quote_identifier: (any character) (Default: ") # Preferred quote identifier to use.

#### Anti-pattern
```sql
SELECT `foo`, "bar", `sum`
FROM table_a;
```
#### Pattern
```sql
SELECT foo, bar, "sum"
FROM table_a;
```

#### sqlfluff.cfg
```cfg
[sqlfluff:rules:L059]
force_quote_identifier: True
preferred_quote_identifier: `
```
#### Anti-pattern
```sql
SELECT foo, "bar", "sum"
FROM table_a;
```
#### Pattern
```sql
SELECT `foo`, `bar`, `sum`
FROM table_a;
```

This doesn't break the current behavior of the L059 rule, but allows a better user-configuration for this rule.

What do you think about?

Thanks :+1:

niconoe-:
fixes #2527 by providing a configuration option on the L059 rule.

No side effects expected as the default values for the newly added configuration are respecting the previous L059 behavior.

### Pull Request checklist
- [x] Included test cases to demonstrate any code changes:  `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.

Even though I think I have fulfilled the requirements in terms of development, tests, documentation, please do not hesitate to help me ensuring this PR is OK, as I didn't wrote anything in python since a decade (literally), and I'm not 100% sure I respected the code of conduct (and if I failed, I'm sorry, I'll fix my mistakes).

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -143,3 +143,8 @@ unquoted_identifiers_policy = all
 quoted_identifiers_policy = all
 allow_space_in_identifier = False
 additional_allowed_characters = ""
+
+[sqlfluff:rules:L059]
+# Policy on quoted and unquoted identifiers
+force_quote_identifier = False
+preferred_quote_identifier = '"'
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -147,7 +147,6 @@ additional_allowed_characters = ""
 [sqlfluff:rules:L059]
 # Policy on quoted and unquoted identifiers
 force_quote_identifier = False
-preferred_quote_identifier = '"'
 
 [sqlfluff:rules:L062]
 # Comma separated list of blocked words that should not be used
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -146,7 +146,7 @@ additional_allowed_characters = ""
 
 [sqlfluff:rules:L059]
 # Policy on quoted and unquoted identifiers
-force_quote_identifier = False
+prefer_quoted_identifiers = False
 
 [sqlfluff:rules:L062]
 # Comma separated list of blocked words that should not be used
--- a/src/sqlfluff/core/rules/config_info.py
+++ b/src/sqlfluff/core/rules/config_info.py
@@ -144,12 +144,6 @@ STANDARD_CONFIG_INFO_DICT = {
             "Defaults to ``False``."
         ),
     },
-    "preferred_quote_identifier": {
-        "definition": (
-            "Expected character used to quote identifiers when necessary, "
-            "or when ``force_quote_identifier`` is evaluated to ``True``"
-        ),
-    },
     "blocked_words": {
         "definition": (
             "Optional comma-separated list of blocked words which should not be used "
--- a/src/sqlfluff/core/rules/config_info.py
+++ b/src/sqlfluff/core/rules/config_info.py
@@ -139,7 +139,7 @@ STANDARD_CONFIG_INFO_DICT = {
     },
     "force_quote_identifier": {
         "validation": [True, False],
-        "definition": ("If True, forces every identifier to be quoted. Defaults to False."),
+        "definition": ("If ``True``, forces every identifier to be quoted. Defaults to ``False``."),
     },
     "preferred_quote_identifier": {
         "definition": (
--- a/src/sqlfluff/core/rules/config_info.py
+++ b/src/sqlfluff/core/rules/config_info.py
@@ -137,10 +137,10 @@ STANDARD_CONFIG_INFO_DICT = {
             "in addition to alphanumerics (A-Z, a-z, 0-9) and underscores."
         ),
     },
-    "force_quote_identifier": {
+    "prefer_quoted_identifiers": {
         "validation": [True, False],
         "definition": (
-            "If ``True``, forces every identifier to be quoted. "
+            "If ``True``, requires every identifier to be quoted. "
             "Defaults to ``False``."
         ),
     },
--- a/src/sqlfluff/core/rules/config_info.py
+++ b/src/sqlfluff/core/rules/config_info.py
@@ -139,7 +139,10 @@ STANDARD_CONFIG_INFO_DICT = {
     },
     "force_quote_identifier": {
         "validation": [True, False],
-        "definition": ("If ``True``, forces every identifier to be quoted. Defaults to ``False``."),
+        "definition": (
+            "If ``True``, forces every identifier to be quoted. "
+            "Defaults to ``False``."
+        ),
     },
     "preferred_quote_identifier": {
         "definition": (
--- a/src/sqlfluff/core/rules/config_info.py
+++ b/src/sqlfluff/core/rules/config_info.py
@@ -137,6 +137,16 @@ STANDARD_CONFIG_INFO_DICT = {
             "in addition to alphanumerics (A-Z, a-z, 0-9) and underscores."
         ),
     },
+    "force_quote_identifier": {
+        "validation": [True, False],
+        "definition": ("If True, forces every identifier to be quoted. Defaults to False."),
+    },
+    "preferred_quote_identifier": {
+        "definition": (
+            "Expected character used to quote identifiers when necessary, "
+            "or when ``force_quote_identifier`` is evaluated to True"
+        ),
+    },
 }
 
 
--- a/src/sqlfluff/core/rules/config_info.py
+++ b/src/sqlfluff/core/rules/config_info.py
@@ -144,7 +144,7 @@ STANDARD_CONFIG_INFO_DICT = {
     "preferred_quote_identifier": {
         "definition": (
             "Expected character used to quote identifiers when necessary, "
-            "or when ``force_quote_identifier`` is evaluated to True"
+            "or when ``force_quote_identifier`` is evaluated to ``True``"
         ),
     },
 }
--- a/src/sqlfluff/rules/L059.py
+++ b/src/sqlfluff/rules/L059.py
@@ -90,7 +90,7 @@ class Rule_L059(BaseRule):
             )
 
         # Now we only deal with NOT forced quoted identifiers configuration
-        # (meaning force_quote_identifier=False).
+        # (meaning prefer_quoted_identifiers=False).
 
         # Extract contents of outer quotes.
         quoted_identifier_contents = context.segment.raw[1:-1]
--- a/src/sqlfluff/rules/L059.py
+++ b/src/sqlfluff/rules/L059.py
@@ -69,10 +69,10 @@ class Rule_L059(BaseRule):
         self.force_quote_identifier: bool
 
         if self.force_quote_identifier:
-            context_policy = ("naked_identifier")
+            context_policy = "naked_identifier"
             identifier_contents = context.segment.raw
         else:
-            context_policy = ("quoted_identifier")
+            context_policy = "quoted_identifier"
             identifier_contents = context.segment.raw[1:-1]
 
         # Ignore the segments that are not of the same type as the defined policy above.
--- a/src/sqlfluff/rules/L059.py
+++ b/src/sqlfluff/rules/L059.py
@@ -17,11 +17,11 @@ from sqlfluff.core.rules.doc_decorators import (
 class Rule_L059(BaseRule):
     """Unnecessary quoted identifier.
 
-    This rule will fail if the quotes used to quote an identifier are unnecessary,
-    or wrong considering the configuration.
+    This rule will fail if the quotes used to quote an identifier are (un)necessary
+    depending on the ``force_quote_identifier`` configuration.
 
-    By default, the quotes are optional, but required to double-quotes when identifier
-    is named as a keyword, or contains special characters.
+    When ``force_quote_identifier = False`` (default behavior), the quotes are 
+    unnecessary, except for reserved keywords and special charcters in identifiers.
 
     | **Anti-pattern**
     | In this example, a valid unquoted identifier,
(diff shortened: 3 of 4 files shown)


src/sqlfluff/rules/L059.py
```diff
@@ -6,36 +6,118 @@
...
+                return LintResult(
+                    context.segment,
+                    fixes=[
+                        LintFix.replace(
+                            context.segment,
+                            [CodeSegment(raw=expected, name="quoted_identifier", type="identifier")],
+                        )
+                    ],
+                    description=f"Missing quoted identifier {expected}.",
+                )
```
tunetheweb, line 101: So we're barring `naked_identifiers`? And forcing even them to be quoted?

That feels like a separate option to what I thought this was.

Should we split this into:
- `force_quote_identifiers` - replaces unquoted identifiers with quoted version (e..g. `foo` to `"foo"`)
- `force_quote_type` - Changes quotes to preferred type (e.g. `'foo'` ro `"foo"`)

Some might want the second without the first.

src/sqlfluff/rules/L059.py
```diff
@@ -76,4 +158,26 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+        # If this is a reserved keyword, or it is not a valid naked identifier,
+        # ensure the quote used is the preferred one.
+        expected = f"{quote}{quoted_identifier_contents}{quote}"
+        if expected != context.segment.raw:
+            # If the identifier contents contains any occurrence of the preferred quote,
+            # ignore this rule for now.
+            if quote in quoted_identifier_contents:
+                # TODO: if different levels of criticity can be set, this case should
+                #       trigger an `info` level.
+                return None
```
tunetheweb, line 170: Could this be moved way up to the top to exit early?

test/fixtures/rules/std_rule_cases/L059.yml
```diff
@@ -47,3 +79,197 @@ test_pass_special_symbols:
...
+test_pass_multiple_references_backtick_forced:
+  pass_str: |
+    SELECT `foo`
+    FROM `bar`.`baz`;
+  configs:
+    rules:
+      L059:
+        force_quote_identifier: True
+        preferred_quote_identifier: "`"
+
```
tunetheweb, line 227: It's very common in BigQuery to back tick the quote table so we should add that test case in:

```suggestion
test_pass_multiple_references_backtick_forced:
  pass_str: |
    SELECT `foo`
    FROM `bar`.`baz`;
  configs:
    rules:
      L059:
        force_quote_identifier: True
        preferred_quote_identifier: "`"

test_pass_multiple_references_backtick_forced:
  pass_str: |
    SELECT `foo`
    FROM `bar.baz`;
  configs:
    core:
      dialect: bigquery
    rules:
      L059:
        force_quote_identifier: True
        preferred_quote_identifier: "`"

```

tunetheweb (changes requested):
Very good work! However I do have quite a few comments and fear this won't work for some dialects as is.

src/sqlfluff/core/default_config.cfg
```diff
@@ -143,3 +143,8 @@ unquoted_identifiers_policy = all
 quoted_identifiers_policy = all
 allow_space_in_identifier = False
 additional_allowed_characters = ""
+
+[sqlfluff:rules:L059]
+# Policy on quoted and unquoted identifiers
+force_quote_identifier = False
+preferred_quote_identifier = '"'
```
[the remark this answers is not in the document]
niconoe-: I wanted to dissociate the 2 configurations because having a preferred quoting system doesn't mean you want every identifier to be quoted. This is actually the opposite looking at best practices: the better is to not quote identifiers unless you don't have the choice (special chars in fields, reserved keywords, ...).

Even if many dialects use backtick by default, the SQL-92 ANSI standard is using double quotes (and actually, even MySQL can switch to this default standard with `SET GLOBAL SQL_MODE=ANSI_QUOTES`.

Also, looking at the unit tests already defined in L059 before my changes, the double-quote was already the expected behavior, that's why I wanted to set it by default: doing it this way will break nothing from original behavior.

What I could try, if you find this too dangerous, is adding the capability of adding "default" as a value of `preferred_quote_identifier`. When this is set to "default" value, in the L059 process, I can define that `quote` is evaluated to the default quote usage depending on the dialect.
The only thing I'd miss for this is the knowledge about "what is the default quote identifier character for each dialect?". 
And doing this might make it harder to add new dialects, so I guess someone more focused on this repo than me should approve this "default" idea, first...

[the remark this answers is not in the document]
tunetheweb: I think default should be unset (or empty string). That way it's an explicit decision to force one quote type and default is not to do that.

And I'd rename the options, as mentioned below, to the following:

- `force_quote_identifiers` (default: `False`) - replaces unquoted identifiers with quoted version (e..g. foo to "foo")
- `force_quote_type` (default: unset) - Changes quotes to preferred type (e.g. 'foo' to "foo")

src/sqlfluff/rules/L059.py
```diff
@@ -6,36 +6,118 @@
...
 
     .. code-block:: sql
 
-        SELECT 123 as "foo"
+        SELECT 123 as "foo", 'sum'
 
     | **Best practice**
-    | Use unquoted identifiers where possible.
+    | If ``force_quote_identifier = False`` and ``preferred_quote_identifier = '"'`` (or unspecified, as these are the default)
+    | Use unquoted identifiers where possible, and use '"' whevener necessary.
```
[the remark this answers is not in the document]
niconoe-: > whevener?

Yes, I'm very good at doing typos, unfortunately. `Whenever` is better.

I added the last line for cases when another quote identifier has been used instead of the preferred one.

For instance, with `force_quote_identifier = False` and `preferred_quote_identifier = '"'`, if you have an SQL query like:
```sql
SELECT 123 AS `SELECT`;
```
then, the L059 rule should fail, as the expected query should be:
```sql
SELECT 123 AS "SELECT";
```

src/sqlfluff/rules/L059.py
```diff
@@ -6,36 +6,118 @@
...
+        self.force_quote_identifier: bool
+        self.preferred_quote_identifier: str
+        
+        quote = self.preferred_quote_identifier
+        
+        # If the configuration says to not force the identifier quoting, then
+        # simply care about the quoted identifiers.
+        context_policy = ("quoted_identifier")
+        if self.force_quote_identifier == True:
+            context_policy = ("naked_identifier", "quoted_identifier")
```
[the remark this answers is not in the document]
niconoe-: I didn't know about it, thanks.

I guess it was missing on previous behavior too, I'll add it on both cases. The name to use is `single_quoted_identifier`, isn't it?

src/sqlfluff/rules/L059.py
niconoe-: I don't know very well the segment categorizations. I thought `naked_identifiers` is a synonym for "unquoted identifier". This isn't the case?

My point of this feature is to actually have a configuration option to replaces unquoted identifiers (even if unnecessary) with their quoted version (using the quote defined into `preferred_quote_identifier`), but also changes quotes to preferred type using `preferred_quote_identifier` config.

If some want the second without the first, using `force_quote_identifier = False` is doing the job.

src/sqlfluff/rules/L059.py
```diff
@@ -6,36 +6,118 @@
...
+        
+        # We only care about quoted identifiers and possibly naked identifiers if 
+        # the configuration setting is forcing identifiers to be quoted.
+        if context.segment.name not in context_policy:
             return None
-
+        
+        # Manage simplest case of quoted identifiers must be forced first.
```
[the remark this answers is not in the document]
niconoe-:
```suggestion
        # Manage cases of quoted identifiers must be forced first.
```

src/sqlfluff/core/default_config.cfg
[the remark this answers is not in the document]
tunetheweb: Checking out BigQuery it's definitely more complicated. The two types of quotes are very different and cannot be interchanged:

```sql
--Valid - the literal a can be in double quotes, and alias can be in backticks
SELECT "a" AS `b`;

--Invalid - the `a` suggests an identifier from the FROM clause.
SELECT `a` AS `b`;

--Invalid - alias cannot be in quotes
SELECT "a" AS "b";

--Valid:
SELECT
  date aS `col1`, -- date column
  "date" aS `col2`, -- literal
  `date` aS `col3` -- date column (same as col1)
FROM `owner.schema.table` AS `tab1`;
```

So consolidating those to only backticks or only double quotes will break the SQL as they mean different things.

Default definitely has to be not set.

test/fixtures/rules/std_rule_cases/L059.yml
niconoe-: Done in : https://github.com/niconoe-/sqlfluff/commit/4f912bd15120e18b1f2d4c1f33c8cefe811b6f27

src/sqlfluff/rules/L059.py
tunetheweb: > I don't know very well the segment categorizations. I thought naked_identifiers is a synonym for "unquoted identifier". This isn't the case?

That is the case.

> My point of this feature is to actually have a configuration option to replaces unquoted identifiers (even if unnecessary) with their quoted version

Yes I got that, when reading the code. Was a bit shocked to be honest, but hey if that's what you want! 😁

But we need a way of preserving the existing L059 functionality (remove unnecessary quotes, but don't add them). Which I don't think your current config supports does it? To me that should be the default, to be backwards compatible, and any of your new asks, should be configurable extras for those that want it. Maybe that could be done by having the `force_identifiers_quote_type` to be not set? So it's ignored?

src/sqlfluff/core/default_config.cfg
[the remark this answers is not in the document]
niconoe-: Ok, I understand what you mean. I'll rework this using your proposal for the option names, and try to not mix both configuration expectations.

src/sqlfluff/rules/L059.py
niconoe-: > we need a way of preserving the existing L059 functionality (remove unnecessary quotes, but don't add them).

I was trying to do this actually. With the default configuration I provided (`force_quote_identifier = False`), then unnecessary quotes must be removed, and the L059 rule fails if they're not.
My main issue was to find the expected quote for identifiers when `force_quote_identifier = True` (or when the identifier is a reserved keyword/has special chars), which is why I added the configuration `preferred_quote_identifier` so that each user can define its own.

src/sqlfluff/rules/L059.py
niconoe-: Actually, on those lines, I know I'm in a context where I'm dealing with reserved keywords or identifiers with special chars. If I move this way up, it would means I also extend the context to any kind of identifier, and moving it will change the semantic.

src/sqlfluff/core/default_config.cfg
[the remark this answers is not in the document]
niconoe-: I don't understand this:
> `force_identifier_quote_type` (default: unset) - Changes quotes to preferred type (e.g. 'foo' to "foo")

If this config is blank (`""`), then it should either:
1. change quotes to blank (e.g. `'foo'` to `foo`). 
2. change quotes to the default one that is dialect dependent (e.g. `'foo'` to ``` `foo` ``` for MySQL, BigQuery, ..., but `'foo'` to `"foo"` for Ansi, ...).

If you wanna go for 1., then to me there's a conflict with configuration `force_quote_identifier = True`, as the identifier is expected to be quoted, but with no characters, so it doesn't make sense.
If you wanna go for 2., I think it will be definitively less dangerous, but I need to know which default quote is to use for every dialects (I might lack of knowledge about it).

src/sqlfluff/rules/L059.py
tunetheweb: But if an identifier contains the quote you want to use, you'll never change it will you? So don't see why you can't check that early and drop out?

src/sqlfluff/core/default_config.cfg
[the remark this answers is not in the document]
tunetheweb: I'd suggesting if it's unset we don't change quotes and leave them as is (i.e. current L059 functionality).

src/sqlfluff/rules/L059.py
niconoe-: Yep, definitely right.
Updated.

src/sqlfluff/core/default_config.cfg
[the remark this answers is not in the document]
niconoe-: > I'd suggesting if it's unset we don't change quotes and leave them as is (i.e. current L059 functionality).

Will try out this. Sounds good to me.

niconoe-:
@tunetheweb 
I'll work tomorrow on renaming `preferred_quote_identifier` by `force_identifier_quote_type` and allow blank, default value to blank, to be as close as possible from the behavior you described to me today.
I also have seen lots of tests and lints aren't passing, so I'll give it a look tomorrow as well. 

Thanks a lot for your participation

src/sqlfluff/core/default_config.cfg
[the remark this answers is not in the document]
niconoe-: @tunetheweb 
I have some issues in managing the `force_identifier_quote_type` when it's not empty (overwritten by user definition).

If I change your previous example to:
```sql
SELECT
  foo aS `col1`, -- date column
  "foo" aS `col2`, -- literal
  `foo` aS "col3" -- date column (same as col1)
```
if, `"foo"` is parsed as a segment of type `quoted_identifier`, it will be impossible to infer if this is a literal string or a quoted identifier `foo` with `"` as quote, just like it is for `"col3"` (I know this breaks for BigQuery, but it's allowed in lots of other dialects as default quote for SQL-92 ANSI).

I guess I'll just reduce the scope of this enhancement and simply deal with `force_quote_identifier = [True,False]`.
Default to `False` (current behavior on `main` branch). When `True`, the L059 rule will fail if there's some naked_identifiers.

About my check that should ensure which character is used to quote the identifier, that's sound currently impossible to me as long as we can't differentiate literal strings from quoted identifiers with `"`. I'll find another way using business methods that shouldn't be in the scope of SQLFluff tool (business oriented).

src/sqlfluff/rules/L059.py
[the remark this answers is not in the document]
niconoe-: _Resolved as wanted behavior has changed_

src/sqlfluff/rules/L059.py
niconoe-: _Resolved as wanted behavior has changed_

src/sqlfluff/core/default_config.cfg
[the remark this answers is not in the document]
niconoe-: _Resolved as wanted behavior has changed_

src/sqlfluff/rules/L059.py
```diff
@@ -28,14 +38,63 @@ class Rule_L059(BaseRule):
...
+
+    .. code-block:: sql
+
+        SELECT 123 as "foo" -- For ANSI, ...
+        -- or
+        SELECT 123 as `foo` -- For BigQuery, MySql, ...
+
     """
 
+    config_keywords = [
```
niconoe-, line 64: @tunetheweb [Linting tests](https://github.com/sqlfluff/sqlfluff/runs/5037642948?check_suite_focus=true) are outputting this error on this line:
```
./src/sqlfluff/rules/L059.py:64:24: BLK100 Black would make changes.
```

I don't understand what it means. 
Could you help me please?

tunetheweb: Just run black:
```
black src/sqlfluff/rules/L059.py
```

And it'll magically fix your issues.

I would guess it's saying you can fix this on one line, so black is asking why you are splittng it when you don't need it:

```py
config_keywords = ["force_quote_identifier"]
```

tunetheweb: Yikes! Not sure how you can code like that. You;'re a better developer than me! ;-)

Ran `black` for you there and committed.

niconoe-: I'm PHP oriented and my workspace is dedicated to PHP then, not Python :wink: . I usually don't work editing files 1 by 1 :laughing: .

Thanks for running `black` for me!

Hope this will be ok then.

src/sqlfluff/core/default_config.cfg
```diff
@@ -144,6 +144,10 @@ quoted_identifiers_policy = all
 allow_space_in_identifier = False
 additional_allowed_characters = ""
 
+[sqlfluff:rules:L059]
+# Policy on quoted and unquoted identifiers
+force_quote_identifier = False
```
tunetheweb, line 149:
```suggestion
# Policy on quoted and unquoted identifiers
prefer_quoted_identifiers = False
```

src/sqlfluff/core/rules/config_info.py
```diff
@@ -137,6 +137,13 @@
             "in addition to alphanumerics (A-Z, a-z, 0-9) and underscores."
         ),
     },
+    "force_quote_identifier": {
+        "validation": [True, False],
+        "definition": (
+            "If ``True``, forces every identifier to be quoted. "
+            "Defaults to ``False``."
+        ),
+    },
```
tunetheweb, line 146:
```suggestion
    "prefer_quoted_identifiers": {
        "validation": [True, False],
        "definition": (
            "If ``True``, requires every identifier to be quoted. "
            "Defaults to ``False``."
        ),
    },
```

src/sqlfluff/rules/L059.py
```diff
@@ -6,13 +6,23 @@
...
+@document_configuration
 @document_fix_compatible
 class Rule_L059(BaseRule):
     """Unnecessary quoted identifier.
 
+    This rule will fail if the quotes used to quote an identifier are (un)necessary
+    depending on the ``force_quote_identifier`` configuration.
+
+    When ``force_quote_identifier = False`` (default behavior), the quotes are
+    unnecessary, except for reserved keywords and special charcters in identifiers.
```
tunetheweb, line 24:
```suggestion
    When ``prefer_quoted_identifiers = False`` (default behaviour), the quotes are
    unnecessary, except for reserved keywords and special characters in identifiers.
```

src/sqlfluff/rules/L059.py
```diff
@@ -28,14 +38,61 @@ class Rule_L059(BaseRule):
 
         SELECT 123 as foo
 
+    When ``force_quote_identifier = True``, the quotes are always necessary, no
+    matter if the identifier is valid, a reserved keyword, or contains special
+    characters.
```
tunetheweb, line 43:
```suggestion
    When ``prefer_quoted_identifiers = True``, the quotes are always necessary, no
    matter if the identifier is valid, a reserved keyword, or contains special
    characters. Note due to different quotes this mode is not `sqlfluff fix` compatible.
```

src/sqlfluff/rules/L059.py
```diff
@@ -28,14 +38,61 @@ class Rule_L059(BaseRule):
...
+
+    .. code-block:: sql
+
+        SELECT 123 as "foo" -- For ANSI, ...
+        -- or
+        SELECT 123 as `foo` -- For BigQuery, MySql, ...
+
     """
 
+    config_keywords = ["force_quote_identifier"]
```
tunetheweb, line 64:
```suggestion
    config_keywords = ["prefer_quoted_identifiers"]
```

src/sqlfluff/rules/L059.py
```diff
@@ -28,14 +38,61 @@ class Rule_L059(BaseRule):
...
+
     def _eval(self, context: RuleContext) -> Optional[LintResult]:
         """Unnecessary quoted identifier."""
-        # We only care about quoted identifiers.
-        if context.segment.name != "quoted_identifier":
+        # Config type hints
+        self.force_quote_identifier: bool
+
+        if self.force_quote_identifier:
```
tunetheweb, line 71:
```suggestion
        self.prefer_quoted_identifiers: bool

        if self.prefer_quoted_identifiers:
```

src/sqlfluff/rules/L059.py
```diff
@@ -28,14 +38,61 @@ class Rule_L059(BaseRule):
...
+        # Manage cases of quoted identifiers must be forced first.
+        # Naked identifiers are _de facto_ making this rule fail as configuration forces
+        # them to be quoted.
+        # In this case, it cannot be fixed as quote to use is dialect and DBMS
+        # configuration dependent.
+        if self.force_quote_identifier:
+            return LintResult(
+                context.segment,
+                description=f"Missing quoted identifier {identifier_contents}.",
+            )
```
tunetheweb, line 91:
```suggestion
        # Manage cases of identifiers must be quoted first.
        # Naked identifiers are _de facto_ making this rule fail as configuration forces
        # them to be quoted.
        # In this case, it cannot be fixed as which quote to use is dialect dependent
        if self.prefer_quoted_identifiers:
            return LintResult(
                context.segment,
                description=f"Missing quoted identifier {identifier_contents}.",
            )
```

src/sqlfluff/rules/L059.py
```diff
@@ -28,14 +38,61 @@ class Rule_L059(BaseRule):
...
+
+        # Now we only deal with NOT forced quoted identifiers configuration
+        # (meaning force_quote_identifier=False).
```
tunetheweb, line 94:
```suggestion
        # Now we only deal with NOT forced quoted identifiers configuration
        # (meaning prefer_quoted_identifiers=False).
```

test/fixtures/rules/std_rule_cases/L059.yml
```diff
@@ -47,3 +80,189 @@ test_pass_special_symbols:
...
   pass_str: |
     SELECT 123 AS "SELECT"
+
+test_pass_column_reference_forced_ansi:
+  pass_str: |
+    SELECT 123 AS "foo";
+  configs:
+    rules:
+      L059:
+        force_quote_identifier: True
```
tunetheweb, line 90: Can we change all the other test cases like this?

```suggestion
test_pass_column_reference_prefer_quoted_ansi:
  pass_str: |
    SELECT 123 AS "foo";
  configs:
    rules:
      L059:
        prefer_quoted_identifiers: True
```

tunetheweb (changes requested):

test/fixtures/rules/std_rule_cases/L059.yml
niconoe-: Sure, will do!

niconoe-: Done in: https://github.com/niconoe-/sqlfluff/commit/a5e60fce2af064733ca36c2b5d68444714c37971

test/fixtures/rules/std_rule_cases/L059.yml
```diff
@@ -1,5 +1,38 @@
...
+# Pass      True    test_pass_column_reference_prefer_quoted_backticks
+# Fail      True    test_fail_column_reference_prefer_quoted_backticks
+# Pass      True    test_pass_table_reference_prefer_quoted_backticks
+# Fail      True    test_fail_table_reference_prefer_quoted_backticks
+# Pass      True    test_pass_multiple_references_prefer_quoted_backticks
+# Fail      True    test_fail_multiple_references_prefer_quoted_backticks
+# Pass      True    test_pass_whitespace_prefer_quoted_backticks
+# Pass      True    test_pass_special_symbols_prefer_quoted_backticks
+# Pass      True    test_pass_reserved_keyword_prefer_quoted_backticks
+
```
tunetheweb, line 35: Lets remove these lines - test case should be named to make them obvious:

```suggestion
```

tunetheweb (approved):
Pretty much there. One suggestion that I'll commit and then merge.

tunetheweb:
Thanks for working on this @niconoe- ! And for being patient with all the feedback.

test/fixtures/rules/std_rule_cases/L059.yml
tunetheweb: Oh - will remove.

PR #2507, opened 2022-01-29
Postgres: EXPLAIN ANALYSE allows British spelling

jpy-git:
### Brief summary of the change made

Just noticed that Postgres `EXPLAIN ANALZYE` works perfectly fine with the British spelling of `ANALYSE`.

### Are there any other side effects of this change that we should be aware of?
No

### Pull Request checklist
- [X] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #2540, opened 2022-02-01
Adds new rule to allow blocking of certain words

### LINKED ISSUE #362: Trailing commas in select statement should raise linting error
#354 allows parsing of trailing commas in select statements. This is a bigquery feature only (and so technically should raise a *parsing* error), but on balance I think it's more _useful_ that the query parses and then this can be raised as an issue.

I think even in dialects where this is supported, this should be regarded as an anti-pattern. We should have a linting rule which catches this.

### LINKED ISSUE #1800: New rule: database & schema restrictions
_Note: This is for general enhancements to the project. Please use the Bug report template instead to raise parsing/linting/syntax issues for existing supported dialects_
At some companies different databases & schemas serve different purposes. Some are secure. Others aren’t. Some are sandboxes not fit for production while others are for production. In order not to misuse data source it is a good idea to allow projects to prohibit usage of (or limit usage to) certain databases & schemas in their `.sqlfluff` file.

tunetheweb:

Fixes #362
Fixes #1800 
Follow on from #2507

This adds a very configurable new rule to allow blocking of certain terms. To quote from the rule docstring:

    This generic rule can be useful to prevent certain keywords, functions, or objects
    from being used. Only whole words can be blocked, not phrases.

    This block list is case insensitive.

    Example use cases:

    * We prefer ``BOOL`` over ``BOOLEAN`` and there is no existing rule to enforce
      this. We can add ``BOOLEAN`` to the deny list until such a rule is written to
      cause a linting error.
    * We have deprecated a schema/table/function and want to prevent it being used
      in future. We can add that to the denylist and then add a ``-- noqa: L062`` for
      the few exceptions that still need to be in the code base.

Can't think of any. Obviously people shouldn't block certain words (e.g. `SELECT`) but that's up to them.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L062.py
+++ b/src/sqlfluff/rules/L062.py
@@ -39,7 +39,11 @@ class Rule_L062(BaseRule):
 
     def _eval(self, context: RuleContext) -> Optional[LintResult]:
         # Config type hints
-        self.block_word_list: str
+        self.block_word_list: Optional[str]
+
+        # Exit early if no block list set
+        if not self.block_word_list:
+            return None
 
         # Only look at child elements
         # Note: we do not need to ignore comments or meta types
--- a/src/sqlfluff/rules/L062.py
+++ b/src/sqlfluff/rules/L062.py
@@ -80,7 +80,7 @@ class Rule_L062(BaseRule):
     def _init_blocked_words(self):
         """Called first time rule is evaluated to fetch & cache the blocked_words."""
         # Use str() in case bools are passed which might otherwise be read as bool
-        blocked_words_config = str(getattr(self, "blocked_words"))
+        blocked_words_config = getattr(self, "blocked_words")
         if blocked_words_config and blocked_words_config != "None":
             self.blocked_words_list = self.split_comma_separated_string(
                 blocked_words_config.upper()
@@ -88,5 +88,4 @@ class Rule_L062(BaseRule):
         else:
             self.blocked_words_list = []
 
-        blocked_words_list = self.blocked_words_list
-        return blocked_words_list
+        return self.blocked_words_list
--- a/src/sqlfluff/rules/L062.py
+++ b/src/sqlfluff/rules/L062.py
@@ -1,0 +1,58 @@
+"""Implementation of Rule L062."""
+
+from typing import Optional
+
+from sqlfluff.core.rules.base import BaseRule, LintResult, RuleContext
+from sqlfluff.core.rules.doc_decorators import document_configuration
+
+
+@document_configuration
+class Rule_L062(BaseRule):
+    """Block a list of configurable phrases from being used.
+
+    This generic rule can be useful to prevent certain keywords, functions, or objects
+    from being used. Only whole words can be blocked, not phrases.
+
+    This block list is case insensitive.
+
+    Example use cases:
+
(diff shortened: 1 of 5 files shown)


src/sqlfluff/rules/L062.py
```diff
@@ -0,0 +1,62 @@
...
+            return None
+
+        # Only look at child elements
+        # Note: we do not need to ignore comments or meta types
+        # or the like as they will not have single word raws
+        if context.segment.segments:
+            return None
+
+        if str(context.segment.raw_upper) in (
+            self.split_comma_separated_string(self.block_word_list.upper())
```
barrywhart, line 55: Performance concern: Think we could split the string once and store that value, rather than splitting it every time? It would be very similar to [this code](https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/rules/L029.py#L56L62) in rule L029.

barrywhart (changes requested):

src/sqlfluff/rules/L062.py
tunetheweb: Fixed. Meant to do this but forgot!

src/sqlfluff/rules/L062.py
```diff
@@ -0,0 +1,92 @@
...
+        blocked_words_config = str(getattr(self, "blocked_words"))
+        if blocked_words_config and blocked_words_config != "None":
+            self.blocked_words_list = self.split_comma_separated_string(
+                blocked_words_config.upper()
+            )
+        else:
+            self.blocked_words_list = []
+
+        blocked_words_list = self.blocked_words_list
+        return blocked_words_list
```
barrywhart, line 92: Nitpick: I'd replace these two lines with the single line:
```
return self.blocked_words_list
```

src/sqlfluff/rules/L062.py
```diff
@@ -0,0 +1,92 @@
...
+            # So we can cache them for next time for speed.
+            blocked_words_list = self._init_blocked_words()
+
+        # Only look at child elements
+        # Note: we do not need to ignore comments or meta types
+        # or the like as they will not have single word raws
+        if context.segment.segments:
+            return None
+
+        if str(context.segment.raw_upper) in blocked_words_list:
```
barrywhart, line 72: Is the surrounding `str( ... )` necessary? Based on the type hint [here](https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/core/parser/segments/base.py#L239) and the other implementation of this property [here](https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/core/parser/segments/raw.py#L99), it looks like it will always return a `str` without requiring a type conversion.

barrywhart (approved):
A couple of small suggestions. Looks good overall.

src/sqlfluff/rules/L062.py
tunetheweb: Fixed. Not sure why L029 did it that way. Well I added that, but it was copied from another rule (L010) that did it like that.

src/sqlfluff/rules/L062.py
tunetheweb: Looks like you're right. Removed.

Issue #2424, opened 2022-01-22
Improve how "sqlfluff fix" handles parse errors


barrywhart:
Currently, `sqlfluff fix` applies fixes even if there are parse errors. It merely prints a warning and exits with status code `1`.

We should make it "harder" to apply fixes when there are parse errors, because this can easily cause worse problems (i.e. introducing new errors into the SQL code, because rules often don't handle bad code very well, and this is likely to remain the case).

Some ideas:
* If not in "force" mode, mention there were parse errors when asking whether to apply fixes.
* Should we provide a way to let the user apply fixes to the files that parsed successfully, skipping only the files that had parse errors?
* Should we consider changing "force" mode to a "counter" option, so -f would fail on parse errors without applying fixes, but -ff would go ahead and apply fixes anyway?

Note: This issue is a related follow-on to #2345.

tunetheweb:
I think fix the files with no parsing errors, do not fix files with parsing errors (even if some fixes could potentially be applied to parsable bits of the file) and exit non-zero.

jpy-git:
> because rules often don't handle bad code very well, and this is likely to remain the case

It's not that rules don't handle unparsable code well due to a flaw in their logic or that they can be improved to, it's that unparsable code is a flat structure and therefore does not follow the same logical segment grouping as a normal tree and therefore a rule cannot be expected to interpret that, nor should it. It's the responsibility of the parser to group the segments not a rule, hence unparsable, and if the parser can't interpret the code then the rule shouldn't try to infer logic from it. That's why most linters/static analysis tools take a fail safe and fail early approach to unparsable code.

Note that even if you only disable `fix` for these unparsable segments you would still have `lint` raise the same nonsense lint results that can mislead users into looking for fixes that aren't there and wasting their time.
Take #2306, a good example of one of these corrupting fixes, even if you were to disable the fix in that case, a user could still run `sqlfluff lint` (as many of us do in general use and in pre-commit or CI) and see the nonsense errors that lead to the corrupted fix. They might run `fix` afterwards and wonder why it isn't being fixed and then try and fix it themselves manually even though their statement is fine and SQLFluff just can't parse it.

jpy-git:
> I think fix the files with no parsing errors, do not fix files with parsing errors (even if some fixes could potentially be applied to parsable bits of the file) and exit non-zero.

^ I agree that we shouldn't quit out if only 1 file doesn't parse.
[...]

PR #2546, opened 2022-02-02
Improve how "sqlfluff fix" handles templating and parse errors

### LINKED ISSUE #2345: fix succeeds even though lint fails with PRS error

barrywhart:
- [x] Fixing stdin should support same or similar behavior
- [x] Config file setting `fix_even_unparsable`
- [x] Simple API should support same or similar behavior as command line?
- [x] Multi-file tests
- [x] Documentation

Fixes #2345, #2424, #2251, #2134

### Pull Request checklist

  - Other.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -644,7 +644,6 @@ You might have arguments that you pass through every time, e.g rules you
     [sqlfluff]
     verbose = 1
     exclude_rules = L022,L027
-    fix_even_unparsable = True
 
 Note that while the :code:`exclude_rules` config looks similar to the
 above example, the :code:`verbose` config has an integer value. This is
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -644,6 +644,7 @@ You might have arguments that you pass through every time, e.g rules you
     [sqlfluff]
     verbose = 1
     exclude_rules = L022,L027
+    fix_even_unparsable = True
 
 Note that while the :code:`exclude_rules` config looks similar to the
 above example, the :code:`verbose` config has an integer value. This is
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -716,7 +716,7 @@ project would be:
     # Comments start with a hash.
 
     # Ignore anything in the "temp" path
-    /path/
+    /temp/
 
     # Ignore anything called "testing.sql"
     testing.sql
--- a/docs/source/production.rst
+++ b/docs/source/production.rst
@@ -108,7 +108,8 @@ SQLFluff comes with two `pre-commit`_ hooks:
 
 .. warning::
    For safety reasons, ``sqlfluff-fix`` by default will not make any fixes in
-   files that had templating or parse errors.
+   files that had templating or parse errors, even if those errors were ignored
+   using ``noqa`` or `--ignore``.
 
    Although it is not advised, you *can* tell SQLFluff to try and fix
    these files by overriding the ``fix_even_unparsable`` setting
--- a/docs/source/production.rst
+++ b/docs/source/production.rst
@@ -106,7 +106,7 @@ SQLFluff comes with two `pre-commit`_ hooks:
 * sqlfluff-lint: returns linting errors.
 * sqlfluff-fix: attempts to fix rule violations.
 
-.. note::
+.. warning::
    For safety reasons, ``sqlfluff-fix`` by default will not make any fixes in
    files that had templating or parse errors.
 
--- a/docs/source/production.rst
+++ b/docs/source/production.rst
@@ -108,16 +108,16 @@ SQLFluff comes with two `pre-commit`_ hooks:
 
 .. note::
    For safety reasons, ``sqlfluff-fix`` by default will not make any fixes in
-   files that had parse errors.
+   files that had templating or parse errors.
 
    Although it is not advised, you *can* tell SQLFluff to try and fix
-   unparseable files by overriding the ``fix_even_unparsable`` setting
+   these files by overriding the ``fix_even_unparsable`` setting
    in ``.sqlfluff`` config file or using the ``sqlfluff fix --FIX-EVEN-UNPARSABLE``
    command line option.
 
    *Overriding this behavior may break your SQL. If you use this override,
-   always be sure to review any fixes applied to files with parse errors to
-   verify they are okay.*
+   always be sure to review any fixes applied to files with templating or parse
+   errors to verify they are okay.*
 
 You should create a file named `.pre-commit-config.yaml`
 at the root of your git project, which should look
--- a/docs/source/production.rst
+++ b/docs/source/production.rst
@@ -112,7 +112,8 @@ SQLFluff comes with two `pre-commit`_ hooks:
 
    Although it is not advised, you _can_ tell SQLFluff to try and fix
    unparseable files by overriding the ``fix_even_unparsable`` setting
-   (``.sqlfluff`` or ``sqlfluff fix`` command line).
+   in ``.sqlfluff`` config file or using the ``sqlfluff fix --FIX-EVEN-UNPARSABLE``
+   command line option.
 
    *Overriding this behavior may break your SQL. If you use this override,
    always be sure to review any fixes applied to files with parse errors to
--- a/docs/source/production.rst
+++ b/docs/source/production.rst
@@ -118,7 +118,6 @@ SQLFluff comes with two `pre-commit`_ hooks:
    always be sure to review any fixes applied to files with parse errors to
    verify they are okay.*
 
-
 You should create a file named `.pre-commit-config.yaml`
 at the root of your git project, which should look
 like this:
--- a/docs/source/production.rst
+++ b/docs/source/production.rst
@@ -106,6 +106,19 @@ SQLFluff comes with two `pre-commit`_ hooks:
 * sqlfluff-lint: returns linting errors.
 * sqlfluff-fix: attempts to fix rule violations.
 
+.. note::
+   For safety reasons, ``sqlfluff-fix`` by default will not make any fixes in
+   files that had parse errors.
+
+   Although it is not advised, you _can_ tell SQLFluff to try and fix
+   unparseable files by overriding the ``fix_even_unparsable`` setting
+   (``.sqlfluff`` or ``sqlfluff fix`` command line).
+
+   *Overriding this behavior may break your SQL. If you do, be sure to review
+   any fixes applied to files with parse errors to verify they are sensible and
+   correct.*
+
+
 You should create a file named `.pre-commit-config.yaml`
 at the root of your git project, which should look
 like this:
--- a/docs/source/production.rst
+++ b/docs/source/production.rst
@@ -110,7 +110,7 @@ SQLFluff comes with two `pre-commit`_ hooks:
    For safety reasons, ``sqlfluff-fix`` by default will not make any fixes in
    files that had parse errors.
 
-   Although it is not advised, you _can_ tell SQLFluff to try and fix
+   Although it is not advised, you *can* tell SQLFluff to try and fix
    unparseable files by overriding the ``fix_even_unparsable`` setting
    in ``.sqlfluff`` config file or using the ``sqlfluff fix --FIX-EVEN-UNPARSABLE``
    command line option.
--- a/docs/source/production.rst
+++ b/docs/source/production.rst
@@ -114,9 +114,9 @@ SQLFluff comes with two `pre-commit`_ hooks:
    unparseable files by overriding the ``fix_even_unparsable`` setting
    (``.sqlfluff`` or ``sqlfluff fix`` command line).
 
-   *Overriding this behavior may break your SQL. If you do, be sure to review
-   any fixes applied to files with parse errors to verify they are sensible and
-   correct.*
+   *Overriding this behavior may break your SQL. If you use this override,
+   always be sure to review any fixes applied to files with parse errors to
+   verify they are okay.*
 
 
 You should create a file named `.pre-commit-config.yaml`
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -127,7 +127,7 @@ def fix(
     should_fix = True
     if not fix_even_unparsable:
         # If fix_even_unparsable wasn't set, check for templating or parse
-        # errors. Okay to fix if there aren't any.
+        # errors and suppress fixing if there were any.
         _, num_filtered_errors = result.check_templating_or_parse_errors()
         if num_filtered_errors > 0:
             should_fix = False
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -131,7 +131,6 @@ def fix(
         # If fix_even_unparsable wasn't set, check for templating or parse
         # errors and suppress fixing if there were any.
         _, num_filtered_errors = result.count_tmp_prs_errors()
-        result.discard_fixes_for_lint_errors_in_files_with_tmp_or_prs_errors()
         if num_filtered_errors > 0:
             should_fix = False
     if should_fix:
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -129,7 +129,7 @@ def fix(
         # If fix_even_unparsable wasn't set, check for parse errors. Okay to fix
         # if there are no parse errors.
         _, num_filtered_parse_errors = result.check_parse_errors()
-        if not num_filtered_parse_errors:
+        if num_filtered_parse_errors == 0:
             should_fix = True
     if should_fix:
         sql = result.paths[0].files[0].fix_string()[0]
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -129,8 +129,8 @@ def fix(
         # If fix_even_unparsable wasn't set, check for templating or parse
         # errors. Okay to fix if there aren't any.
         _, num_filtered_errors = result.check_templating_or_parse_errors()
-        if num_filtered_errors == 0:
-            should_fix = True
+        if num_filtered_errors > 0:
+            should_fix = False
     if should_fix:
         sql = result.paths[0].files[0].fix_string()[0]
     return sql
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -130,7 +130,8 @@ def fix(
     if not fix_even_unparsable:
         # If fix_even_unparsable wasn't set, check for templating or parse
         # errors and suppress fixing if there were any.
-        _, num_filtered_errors = result.mark_failed_files_unfixable()
+        _, num_filtered_errors = result.count_tmp_prs_errors()
+        result.mark_failed_files_unfixable()
         if num_filtered_errors > 0:
             should_fix = False
     if should_fix:
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -126,10 +126,10 @@ def fix(
     fix_even_unparsable = cfg.get("fix_even_unparsable")
     should_fix = True
     if not fix_even_unparsable:
-        # If fix_even_unparsable wasn't set, check for parse errors. Okay to fix
-        # if there are no parse errors.
-        _, num_filtered_parse_errors = result.check_parse_errors()
-        if num_filtered_parse_errors == 0:
+        # If fix_even_unparsable wasn't set, check for templating or parse
+        # errors. Okay to fix if there aren't any.
+        _, num_filtered_errors = result.check_templating_or_parse_errors()
+        if num_filtered_errors == 0:
             should_fix = True
     if should_fix:
         sql = result.paths[0].files[0].fix_string()[0]
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -131,7 +131,7 @@ def fix(
         # If fix_even_unparsable wasn't set, check for templating or parse
         # errors and suppress fixing if there were any.
         _, num_filtered_errors = result.count_tmp_prs_errors()
-        result.mark_failed_files_unfixable()
+        result.discard_fixes_for_lint_errors_in_files_with_tmp_or_prs_errors()
         if num_filtered_errors > 0:
             should_fix = False
     if should_fix:
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -97,6 +97,7 @@ def fix(
     rules: Optional[List[str]] = None,
     exclude_rules: Optional[List[str]] = None,
     config_path: Optional[str] = None,
+    fix_even_unparsable: Optional[bool] = None,
 ) -> str:
     """Fix a SQL string.
 
@@ -123,12 +124,13 @@ def fix(
     linter = Linter(config=cfg)
 
     result = linter.lint_string_wrapped(sql, fix=True)
-    fix_even_unparsable = cfg.get("fix_even_unparsable")
+    if fix_even_unparsable is None:
+        fix_even_unparsable = cfg.get("fix_even_unparsable")
     should_fix = True
     if not fix_even_unparsable:
         # If fix_even_unparsable wasn't set, check for templating or parse
         # errors and suppress fixing if there were any.
-        _, num_filtered_errors = result.check_templating_or_parse_errors()
+        _, num_filtered_errors = result.mark_failed_files_unfixable()
         if num_filtered_errors > 0:
             should_fix = False
     if should_fix:
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -123,8 +123,15 @@ def fix(
     linter = Linter(config=cfg)
 
     result = linter.lint_string_wrapped(sql, fix=True)
-    fixed_string = result.paths[0].files[0].fix_string()[0]
-    return fixed_string
+    fix_even_unparsable = cfg.get("fix_even_unparsable")
+    should_fix = fix_even_unparsable
+    if not should_fix:
+        _, num_filtered_parse_errors = result.check_parse_errors()
+        if not num_filtered_parse_errors:
+            should_fix = True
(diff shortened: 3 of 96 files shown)


docs/source/production.rst
```diff
@@ -106,6 +106,18 @@ SQLFluff comes with two `pre-commit`_ hooks:
 * sqlfluff-lint: returns linting errors.
 * sqlfluff-fix: attempts to fix rule violations.
 
+.. note::
+   For safety reasons, ``sqlfluff-fix`` by default will not make any fixes in
+   files that had parse errors.
+
+   Although it is not advised, you _can_ tell SQLFluff to try and fix
```
tunetheweb, line 113: What does the underline mean? Should it be in italics?

```suggestion
   Although it is not advised, you `can` tell SQLFluff to try and fix
```

docs/source/production.rst
```diff
@@ -106,6 +106,18 @@ SQLFluff comes with two `pre-commit`_ hooks:
...
+   unparseable files by overriding the ``fix_even_unparsable`` setting
+   (``.sqlfluff`` or ``sqlfluff fix`` command line).
```
tunetheweb, line 115:
```suggestion
   unparseable files by overriding the ``fix_even_unparsable`` setting
   in ``.sqlfluff`` config file or using the ``sqlfluff fix --FIX-EVEN-UNPARSABLE``
   command line option.
```

docs/source/production.rst
```diff
@@ -106,6 +106,18 @@ SQLFluff comes with two `pre-commit`_ hooks:
 * sqlfluff-lint: returns linting errors.
 * sqlfluff-fix: attempts to fix rule violations.
 
+.. note::
+   For safety reasons, ``sqlfluff-fix`` by default will not make any fixes in
+   files that had parse errors.
```
tunetheweb, line 111: Is it same for templating errors? Or just parse errors?

src/sqlfluff/api/simple.py
```diff
@@ -123,8 +123,17 @@ def fix(
     linter = Linter(config=cfg)
 
     result = linter.lint_string_wrapped(sql, fix=True)
-    fixed_string = result.paths[0].files[0].fix_string()[0]
-    return fixed_string
+    fix_even_unparsable = cfg.get("fix_even_unparsable")
+    should_fix = fix_even_unparsable
+    if not should_fix:
```
tunetheweb, line 128: I'm very confused by this.

1) Why do we need two variables (`fix_even_unparsable` and `should_fix`)?
2) `if not should_fix` is doing my head in. Any way to rephrase without the `not`? Especially as it seems to be a variable created just for this.

src/sqlfluff/api/simple.py
```diff
@@ -123,8 +123,17 @@ def fix(
...
+        # If fix_even_unparsable wasn't set, check for parse errors. Okay to fix
+        # if there are no parse errors.
+        _, num_filtered_parse_errors = result.check_parse_errors()
+        if not num_filtered_parse_errors:
+            should_fix = True
```
tunetheweb, line 133: Would this not be more readible stated as this?:

```suggestion
        if num_filtered_parse_errors == 0:
            should_fix = True
```

src/sqlfluff/core/default_config.cfg
```diff
@@ -27,6 +27,11 @@ disable_noqa = False
 # Comma separated list of file extensions to lint
 # NB: This config will only apply in the root folder
 sql_file_exts = .sql,.sql.j2,.dml,.ddl
+# By default, files won't be fixed if they contain parse errors (even if those
+# errors are ignored with "noqa"). This option allows overriding that behavior:
+# if set to True, fixes *will* be applied. USE WITH CAUTION: MAY APPLY SPURIOUS
+# OR UNNECESSARY FIXES. IT MAY BREAK OR CORRUPT YOUR SQL.
+fix_even_unparsable = False
```
tunetheweb, line 34: Could we make this smaller?

```suggestion
# Allow fix to run on files, even if they contain parsing errors
# Note altering this is NOT RECOMMENDED as can corrupt SQL
fix_even_unparsable = False
```

test/cli/commands_test.py
```diff
@@ -509,6 +511,149 @@ def test__cli__command__fix(rule, fname):
...
+                FROM my_schema.my_table
+                where processdate ! 3
+                """,
+                """SELECT my_col
+                FROM my_schema.my_table
+                where processdate != 3""",
+            ],
+            ["--force", "--fixed-suffix", "FIXED", "--rules", "L010"],
+            [
+                None,
```
tunetheweb, line 612: Should this be None? Or should it be the original sql?

tunetheweb (changes requested):

docs/source/configuration.rst
```diff
@@ -644,6 +644,7 @@ You might have arguments that you pass through every time, e.g rules you
     [sqlfluff]
     verbose = 1
     exclude_rules = L022,L027
+    fix_even_unparsable = True
```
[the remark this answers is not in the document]
barrywhart: I was unsure. Having it here makes people aware that it can be configured, but if we don't explain that it's a bad idea, it may encourage people to use it inappropriately. I'll remove it.

docs/source/production.rst
barrywhart: Oh yeah. Markdown confusion. Will fix.

docs/source/production.rst
barrywhart: Thus far, I've only touched the behavior for parse errors. Let me think about and experiment with templating errors. They should probably behave the same way.

src/sqlfluff/api/simple.py
barrywhart: Works for me!

src/sqlfluff/core/default_config.cfg
barrywhart: I often wonder how much we need to scare people. We'll go with the shorter phrasing for now. We can make it scarier later if people keep shooting themselves in the foot and reporting issues. 😅

test/cli/commands_test.py
barrywhart: `None` is a special value: It means that `fix` did nothing, i.e. it didn't even _write_ an output file. Technically, they are almost the same, but if we're using `--fixed-suffix`, then `None` makes it clearer that the "fixed" file shouldn't even be created. Same contents would suggest the file was written, but without changes. See the test body and you'll see what I mean -- with `None`, it literally checks that the output file is not present.

src/sqlfluff/api/simple.py
barrywhart: It's because there are not one but *two* ways that `should_fix` can be set to `True`, either initially, based on `fix_even_unparsable`, or inside the `if` statement. Basically, we "should fix" the file if either:
* `fix_even_unparsable` is set **OR**
* The file has no parse errors. The code inside the `if` checks for this.

There are other ways to structure this code that may be clearer, but I think they would involve duplicating some of the code.

tunetheweb: Well the way it’s currently written is a bit of an optimisation - it only calls the unparsable check if it needs to. So those using the force option (that we want to discourage) save that check.

Given that we want to discourage that force nature, I’m not really inclined to optimise for them, and would favour code readability over that.

So I would say do the check and then after, if no parse errors OR flag is set, do the fix.

WDYT?

src/sqlfluff/core/default_config.cfg
tunetheweb: I was thinking the big block made it stand out (which we don’t really want)! 😁 So tried to minimise it a bit.

test/cli/commands_test.py
tunetheweb: Grand. I just presumed that fix returned the original SQL if nothing to fix. Think the simple API does that. But probably makes more sense there. Anyway must be right as tests pass! So was more a question for my understanding.

src/sqlfluff/api/simple.py
barrywhart: I agree the current code looks odd when you initially see it, but I'm also trying to consider if you're a maintainer and you take a closer look, you're likely to ask, "Why is it doing all this useless work to count errors when it doesn't need to? Maybe I'm missing something. I'm afraid to touch this code."

I'm making a slight change that you may like better.

src/sqlfluff/api/simple.py
```diff
@@ -123,8 +123,17 @@ def fix(
...
-    fixed_string = result.paths[0].files[0].fix_string()[0]
-    return fixed_string
+    fix_even_unparsable = cfg.get("fix_even_unparsable")
+    should_fix = True
+    if not fix_even_unparsable:
+        # If fix_even_unparsable wasn't set, check for parse errors. Okay to fix
+        # if there are no parse errors.
+        _, num_filtered_parse_errors = result.check_parse_errors()
+        if num_filtered_parse_errors == 0:
+            should_fix = True
```
tunetheweb, line 133: Doesn’t this need to be `!=` and also setting it to false?

btw do like this better. Though personally think I would have gone for parse error check first and then see if needed. Even if it’s a slight waste of processing for those using the force flag.

docs/source/production.rst
```diff
@@ -106,6 +106,19 @@ SQLFluff comes with two `pre-commit`_ hooks:
 * sqlfluff-lint: returns linting errors.
 * sqlfluff-fix: attempts to fix rule violations.
 
+.. note::
```
tunetheweb, line 109:
```suggestion
.. warning::
```

docs/source/production.rst
barrywhart: Ok, I updated it to handle templater errors the same way. Need to review a little more closely to see if there's leftover code to clean up, but it should be working correctly and consistently now.

barrywhart:
I'm putting this back in draft mode because I "discovered" the existing `--ignore templating,parsing` command-line option. I think we should use that, rather than adding a new `--FIX-EVEN-UNPARSABLE` option.

@tunetheweb, @alanmcruickshank

src/sqlfluff/api/simple.py
barrywhart: Ah yes. Cuz it's already set to true. The risk of refactoring working code. 😉

tunetheweb:
> I'm putting this back in draft mode because I "discovered" the existing `--ignore templating,parsing` command-line option. I think we should use that, rather than adding a new `--FIX-EVEN-UNPARSABLE` option.

That’s a slightly different thing though. It basically means ignore them from linting. Your own means “fix what you can even if linting errors, whether ignored or not”.

barrywhart:
>That’s a slightly different thing though. It basically means ignore them from linting.
>Your own means “fix what you can even if linting errors, whether ignored or not”.

Are you sure? Look at [these messages](https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/cli/commands.py#L672L685) printed in `fix` when there are templater errors.  Specifically, this one:
>Use '--ignore templating' to attempt to fix anyway.

🤯

barrywhart:
@tunetheweb: Oh yes, you're right. `--ignore` is almost identical to `noqa`, but applied globally. 👍

src/sqlfluff/core/linter/linting_result.py
```diff
@@ -181,3 +184,34 @@ def tree(self) -> Optional[BaseSegment]:  # pragma: no cover
...
+                        types=types, filter_ignore=False
+                    )
+                    if num_errors:
+                        # File has errors. Discard all the fixes: they are
+                        # potentially unsafe.
+                        for violation in linted_file.violations:
+                            if hasattr(violation, "fixes"):
+                                violation.fixes = []  # type: ignore
+                        num_filtered_errors += linted_file.num_violations(types=types)
+        return total_errors, num_filtered_errors
```
barrywhart, line 217: This function is the main part of the PR. Every flavor of "fix" eventually uses it:
* `fix` command line
* `fix` command line for stdin (special case)
* Simple API `fix()` function

test/cli/commands_test.py
```diff
@@ -509,6 +511,162 @@ def test__cli__command__fix(rule, fname):
...
+                """SELECT my_col
+                FROM my_schema.my_table
+                WHERE processdate != 3""",
+            ],
+            1,
+        ),
+    ],
```
barrywhart, line 631: These test cases are worth a careful review. They should match with the behavior we discussed in the maintainers meeting and added to issue #2424:

`sqlfluff fix`:
- Only fix if no unparsable segments found in the file. If any are found, no fixes happen even if parsing errors are supressed using noqa or `--ignore` cli options. The only way to override this is with the `--FIX-EVEN-UNPARSABLE` option as above.
- If there are still parse errors after applying "noqa" directives, exit 1. Otherwise, allow exit 0.

src/sqlfluff/api/simple.py
```diff
@@ -123,8 +123,17 @@ def fix(
     linter = Linter(config=cfg)
 
     result = linter.lint_string_wrapped(sql, fix=True)
-    fixed_string = result.paths[0].files[0].fix_string()[0]
-    return fixed_string
+    fix_even_unparsable = cfg.get("fix_even_unparsable")
```
alanmcruickshank, line 126: For API usage, I think this should be an available kwarg on `fix()`, and not just accessible via the config.

src/sqlfluff/core/linter/linting_result.py
```diff
@@ -181,3 +184,34 @@ def tree(self) -> Optional[BaseSegment]:  # pragma: no cover
                 "path."
             )
         return self.paths[0].tree
+
+    def check_templating_or_parse_errors(self) -> Tuple[int, int]:
+        """Treat tiles with templating or parse error as unfixable.
+
+        NOTE! THIS FUNCTION HAS SIDE EFFECTS: IT DISCARDS LINT FIXES.
+
```
alanmcruickshank, line 192: I can see why this function has side effects. Ideally it wouldn't, but I think they way this function is used it's more of a pruning function or filter function on the current set of violations.

Given that it *does* have side effects however, I wonder if we should change the name. `check_<x>` makes it sound like a function which _doesn't_ have side effects.

test/cli/commands_test.py
alanmcruickshank: I think these test cases reflect what we intended 👍 .

alanmcruickshank (changes requested):
Two minor changes - but the bulk of this is spot on I think 👍

src/sqlfluff/cli/commands.py
```diff
@@ -678,11 +716,13 @@ def fix(
             )
             click.echo(
                 colorize(
-                    "Use '--ignore templating' to attempt to fix anyway.",
+                    "Use '--ignore templating' and '--fix-even-unparsable' to "
+                    "attempt to fix anyway.",
```
tunetheweb, line 720: Do you need both? I thought you only needed the latter and the former just controlled the exit code? Though not sure how to succinctly put that!

What about:

```suggestion
                    "Use --fix-even-unparsable' to attempt to fix the SQL anyway, ",
                    " and use '--ignore templating' to avoid outputing templating ",
                    " errors as well.",
```

src/sqlfluff/core/linter/linting_result.py
```diff
@@ -181,3 +184,32 @@ def tree(self) -> Optional[BaseSegment]:  # pragma: no cover
                 "path."
             )
         return self.paths[0].tree
+
+    def mark_failed_files_unfixable(self) -> Tuple[int, int]:
+        """Discard lint fixes for files with templating or parse errors.
```
tunetheweb, line 189: The function name and the description differs.

What about:

```suggestion
    def discard_unfixable_errors(self) -> Tuple[int, int]:
        """Discard lint fixes for files with templating or parse errors.
```

src/sqlfluff/core/linter/linting_result.py
```diff
@@ -181,3 +184,32 @@ def tree(self) -> Optional[BaseSegment]:  # pragma: no cover
...
+
+        Scan all LintedFiles:
+        - Files with none of these errors: No action
+        - Files with these errors (before filtering): Removes fixes from
+          SQLLintError objects, i.e. marks them "unfixable"
```
tunetheweb, line 194: I misread this that we are removing `SQLLintError`s. But actually what we're doing is removing `SQLParseError` and `SQLTemplaterError` from the list of linting errors and leaving the other `SQLLintError`s still. Maybe this could be clearer?

```suggestion
        - Files with these errors (before filtering): Removes fixes for
          `SQLParseError` and `SQLTemplaterError` errors (i.e. marks them
          "unfixable").
```

src/sqlfluff/core/linter/linting_result.py
```diff
@@ -181,3 +184,32 @@ def tree(self) -> Optional[BaseSegment]:  # pragma: no cover
...
+
+        Also returns the total number of these errors before and after
+        "ignore/noqa" filters are applied.
```
tunetheweb, line 197: I wonder if the number of errors after "ignore/noqa" filters are applied should be part of this function?

Seems separate (and short) so could be pulled out.

Just concerned there's potential for confusion between the removal (aka "filtering") of TMP/PRS errors, and the filtering of `ignore`/`--noqa` errors.

tunetheweb (commented):
Sorry, but lots of comments and questions.

It works and so happy with code from that perspective, but think it could be clearer to read, for future.

src/sqlfluff/api/simple.py
barrywhart: Done.

src/sqlfluff/core/linter/linting_result.py
barrywhart: Done.

src/sqlfluff/cli/commands.py
```diff
@@ -575,6 +576,28 @@ def lint(
         sys.exit(0)
 
 
+def _mark_failed_files_unfixable(lint_result: LintingResult) -> int:
+    """Discard lint fixes for files with templating or parse errors.
+
+    Returns 1 if there are any files with templating or parse errors after
+    filtering, else 0. (Intended as a process exit code.)
+    """
+    total_errors, num_filtered_errors = lint_result.mark_failed_files_unfixable()
```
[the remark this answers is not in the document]
barrywhart: One of them is a wrapper around the other, and they're in different modules. There's a similar (albeit exact) duplication of function names across the linter modules and classes, because of the layered structure of the linter (commands -> LintingResult -> LintedDir -> LintedFile:
* `get_violations()`
* `num_violations()`
* `persist_changes()`
* etc.

src/sqlfluff/cli/commands.py
```diff
@@ -575,6 +576,28 @@ def lint(
...
+    if total_errors:
+        click.echo(
+            colorize(f"  [{total_errors} templating/parsing errors found]", Color.red)
+        )
+        if 0 < num_filtered_errors < total_errors:
```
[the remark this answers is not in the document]
barrywhart: >What if there are no filtered errors at all but there are TMP/PRS errors?
`num_filtered_errors` is a variable that counts the number of filtered TMP and PRS errors, so I don't understand the question.

>Should it be 0 <= num_filtered_errors < ...? At which point do you even need this and should it just be num_filtered_errors < total_errors?
I originally didn't have the `0 <` part. I added it because I was thinking, if *all* the errors are filtered, why print an "error" message saying "0 errors after filtering", because that's _good_, right?

src/sqlfluff/cli/commands.py
barrywhart: Will do.

src/sqlfluff/core/linter/linting_result.py
barrywhart: It's a difficult function to name. The name you proposed is also inaccurate: we're not discarding _errors_, we're discarding _fixes_ for errors and keeping the actual errors.

How about `discard_fixes_for_lint_errors_in_files_with_templating_or_parsing_errors`?

src/sqlfluff/core/linter/linting_result.py
barrywhart: It's not removing fixes for SQLParseError and SQLTemplaterError errors. It's removing fixes for SQLLintError in *files* with SQLParseError and SQLTemplaterError. The SQLParseError and SQLTemplaterError objects are left untouched.

src/sqlfluff/core/linter/linting_result.py
barrywhart: Done.

src/sqlfluff/core/linter/linting_result.py
barrywhart: Renamed as `discard_fixes_for_lint_errors_in_files_with_tmp_or_prs_errors()`.

src/sqlfluff/cli/commands.py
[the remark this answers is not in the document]
barrywhart: Renamed both these functions. The names are more distinct now.

src/sqlfluff/cli/commands.py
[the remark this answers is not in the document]
barrywhart: I removed the `0 < ` from the `if`, so now the `no cover` shouldn't be needed, because existing tests will now hit this code. Instead, I changed it so the color of the output is red if there are still errors after filtering, green if there are none.

src/sqlfluff/cli/commands.py
```diff
@@ -575,6 +576,28 @@ def lint(
...
+            click.echo(
+                colorize(
+                    f"  [{num_filtered_errors} templating/parsing errors "
+                    f'remaining after "ignore"]',
+                    Color.red,
+                )
+            )  # pragma: no cover
```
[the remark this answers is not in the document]
barrywhart: As noted above, I changed the `if` statement a bit. Now this line will be executed by the existing tests.

src/sqlfluff/cli/commands.py
barrywhart: Technically, you're absolutely right that only `--fix-even-unparsable` is now needed. I was unsure of the best wording here, and thought that users may want to be aware of both options. It's a good point, though. I'll remove mention of `--ignore templating`.

barrywhart: Done.

src/sqlfluff/core/linter/linting_result.py
barrywhart: I'm changing the `hasattr(violation, "fixes")` to `isinstance(violation, SQLLintError)`. This will make it clearer what's happening. 👍

barrywhart:
@tunetheweb: Okay, I think I've addressed all (almost all?) of your feedback. Ready for another review. I removed the `#pragma: no cover` from the `click.echo()` after changing the code around it, so I think it'll be covered by existing tests. We'll know for sure once the build finishes. 🍿

src/sqlfluff/cli/commands.py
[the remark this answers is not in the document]
tunetheweb: Fair point about it just being output btw. Overlooked that. But good that it's not needed anyway now.

src/sqlfluff/core/linter/linting_result.py
tunetheweb: A function name almost as long as your branch names 😁

src/sqlfluff/api/simple.py
```diff
@@ -123,8 +124,19 @@ def fix(
...
+    should_fix = True
+    if not fix_even_unparsable:
+        # If fix_even_unparsable wasn't set, check for templating or parse
+        # errors and suppress fixing if there were any.
+        _, num_filtered_errors = result.count_tmp_prs_errors()
+        result.discard_fixes_for_lint_errors_in_files_with_tmp_or_prs_errors()
+        if num_filtered_errors > 0:
+            should_fix = False
+    if should_fix:
+        sql = result.paths[0].files[0].fix_string()[0]
```
tunetheweb, line 138: Sorry, I'm still having trouble with this (head now hurts!)

Why do we need to call `discard_fixes_for_lint_errors_in_files_with_tmp_or_prs_errors` here now?

If there are PRS or TMP errors (whether `noqa`ed or not) then we will set `should_fix = False` so won't fix, so no reason to discard.
If there are no PRS or TMP errors, then no reason to discard.

barrywhart: We don't. This is an optimization exposed by splitting the one function into two functions. Will change.

barrywhart: Done!

barrywhart:
>A function name almost as long as your branch names 😁
I was trying to avoid it, but it's hard to explain in a few words. 😛  Next time, `magic_code_xxx123()`. 🎉

tunetheweb (approved):
I think we got there!

barrywhart:
@tunetheweb: Whee! Onward and upward. 🚀

PR #2547, opened 2022-02-02
Remove requirement for Click>=8

### LINKED ISSUE #2543: Less strict click version
From what I get the pinned click requirement >= 8.0.0 was added to support autocompletion of the cli. Could that be made optional  and the version requirement be lowered. It makes it difficult for us to combine in env with other old apps. It might be an edge case, I don't know but we get hit by it through and old airflow -> flask version.  

Get that it's not the intention by we and I would assume at a lot of other users prefer to have sqlfluff as a lib and used in other cli's or apps.

tunetheweb:

Fixes #2543 

Not sure if this it the best way of doing this in Python, and don't think we can add test cases, so feel free to suggest improvements

Shouldn't be

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/cli/autocomplete.py
+++ b/src/sqlfluff/cli/autocomplete.py
@@ -1,14 +1,12 @@
 """autocompletion commands."""
 
-import click
-
 from sqlfluff import list_dialects
 
 # Older versions of click don't have shell completion
 # so handle for now, as version 8 still fairly recent
 shell_completion_enabled = True
 try:
-    completion = __import__("click.shell_completion", fromlist=[click])
+    completion = __import__("click.shell_completion", fromlist=["click"])
 except ImportError:
     shell_completion_enabled = False
 
--- a/src/sqlfluff/cli/autocomplete.py
+++ b/src/sqlfluff/cli/autocomplete.py
@@ -1,8 +1,16 @@
 """autocompletion commands."""
 
-from click.shell_completion import CompletionItem
+import click
 
 from sqlfluff import list_dialects
+
+# Older versions of click don't have shell completion
+# so handle for now, as version 8 still fairly recent
+shell_completion_enabled = True
+try:
+    completion = __import__("click.shell_completion", fromlist=[click])
+except ImportError:
+    shell_completion_enabled = False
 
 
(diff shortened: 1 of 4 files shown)


tunetheweb:
The alternative is we could just remove the dialect completion code. It's only a small thing, and that way we could just depend on the default completion for commands (which doesn't require ANY code on our side) and have people just have to type the dialect in full (shock! horror!).

WDYT?

src/sqlfluff/cli/autocomplete.py
```diff
@@ -1,9 +1,15 @@
...
 
-from click.shell_completion import CompletionItem
-
 from sqlfluff import list_dialects
 
+# Older versions of click don't have shell completion
+# so handle for now, as version 8 still fairly recent
+shell_completion_enabled = True
+try:
+    completion = __import__("click.shell_completion", fromlist=["click"])
```
barrywhart, line 9: You can use the normal.`import` command here -- no need to use the function.

barrywhart (approved):

src/sqlfluff/cli/autocomplete.py
tunetheweb: Oh yeah. Went round the houses as this was originally importing just one class where that doesn't work when the module doesn't exist. But works fine now just importing the module.

Fixed!

terzieff:
Hey, 

so I just tried to use SQLFluff for the first time, installed it and got hit with this right away (happens with whatever sqlflull method):
```>sqlfluff version
  File "/opt/conda/bin/sqlfluff", line 5, in <module>
    from sqlfluff.cli.commands import cli
  File "/opt/conda/lib/python3.9/site-packages/sqlfluff/cli/commands.py", line 415, in <module>
    type=click.Choice([ft.value for ft in FormatType], case_sensitive=False),
TypeError: __init__() got an unexpected keyword argument 'case_sensitive'
```

It looks like I have `click==6.7` and the issue looks similar to [this](https://github.com/bentoml/BentoML/issues/335) and [this](https://github.com/ledesmablt/spotify-cli/issues/9). So maybe it'd make sense to require a certain version of click?

PR #2553, opened 2022-02-03
Spark3: Add Direct File Query

### LINKED ISSUE #602: Issue with Databricks absolute path tables
We're using DBT with Databricks and tried to implement sqlfluff for linting. However, we're seeing that the following syntax is unsupported:

```sql
SELECT *
FROM delta.`/mnt/datalake/table`
```

Is this something that we want to support in sqlfluff? If so, I can try to come up with a PR.

R7L208:

- Adds support to directly query files of supported types.
- Fixes #602 

- Reorganized some grammar for Files to isolate File keywords that support this clause.
- Removed "XML" File format. No test cases and requires additional work before it will parse correctly.

- Linting fails due to special characters in identifiers (L057). These special keywords are required so would be an exception to L057. I've not yet worked on rules so would appreciate any guidance on where to get started on fixing this particular bug.

```bash
> sqlfluff lint --dialect spark3 test/fixtures/dialects/spark3/select_from_file.sql
== [test/fixtures/dialects/spark3/select_from_file.sql] FAIL
L:   6 | P:  14 | L057 | Do not use special characters in identifiers.
L:  13 | P:  14 | L057 | Do not use special characters in identifiers.
L:  20 | P:  10 | L057 | Do not use special characters in identifiers.
L:  27 | P:  11 | L057 | Do not use special characters in identifiers.
L:  34 | P:  11 | L057 | Do not use special characters in identifiers.
L:  41 | P:  11 | L057 | Do not use special characters in identifiers.
L:  48 | P:  11 | L057 | Do not use special characters in identifiers.
L:  55 | P:  17 | L057 | Do not use special characters in identifiers.
L:  62 | P:  17 | L057 | Do not use special characters in identifiers.
L:  69 | P:  10 | L057 | Do not use special characters in identifiers.
L:  76 | P:  12 | L057 | Do not use special characters in identifiers.
All Finished 📜 🎉!
```

```bash
[L: 76, P:  6]      |                from_expression:
[L: 76, P:  6]      |                    [META] indent:
[L: 76, P:  6]      |                    from_expression_element:
[L: 76, P:  6]      |                        table_expression:
[L: 76, P:  6]      |                            file_reference:
[L: 76, P:  6]      |                                keyword:                      'DELTA'
[L: 76, P: 11]      |                                dot:                          '.'
[L: 76, P: 12]      |                                identifier:                   '`/mnt/datalake/table`'
[L: 76, P: 33]      |                    [META] dedent:
[L: 76, P: 33]      |    statement_terminator:                                     ';'
[L: 76, P: 34]      |    newline:                                                  '\n'
```

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_spark3.py
+++ b/src/sqlfluff/dialects/dialect_spark3.py
@@ -1644,8 +1644,8 @@ class TableExpressionSegment(BaseSegment):
         Ref("ValuesClauseSegment"),
         Ref("BareFunctionSegment"),
         Ref("FunctionSegment"),
-        Ref("TableReferenceSegment"),
         Ref("FileReferenceSegment"),
+        Ref("TableReferenceSegment"),
         # Nested Selects
         Bracketed(Ref("SelectableGrammar")),
     )
--- a/src/sqlfluff/dialects/dialect_spark3.py
+++ b/src/sqlfluff/dialects/dialect_spark3.py
@@ -1663,5 +1663,5 @@ class FileReferenceSegment(BaseSegment):
     match_grammar = Sequence(
         Ref("DataSourcesV2FileTypeGrammar"),
         Ref("DotSegment"),
-        Ref("QuotedIdentifierSegment"),
+        Ref("QuotedLiteralSegment"),
     )
--- a/src/sqlfluff/dialects/dialect_spark3.py
+++ b/src/sqlfluff/dialects/dialect_spark3.py
@@ -1663,5 +1663,7 @@ class FileReferenceSegment(BaseSegment):
     match_grammar = Sequence(
         Ref("DataSourcesV2FileTypeGrammar"),
         Ref("DotSegment"),
-        Ref("QuotedLiteralSegment"),
+        # NB: Using `QuotedLiteralSegment` here causes `FileReferenceSegment`
+        # to match as a `TableReferenceSegment`
+        Ref("QuotedIdentifierSegment"),
     )
--- a/src/sqlfluff/dialects/dialect_spark3.py
+++ b/src/sqlfluff/dialects/dialect_spark3.py
@@ -275,7 +275,7 @@ spark3_dialect.add(
         "DBPROPERTIES", Ref("BracketedPropertyListGrammar")
     ),
     DataSourcesV2FileTypeGrammar=OneOf(
-        # https://github.com/apache/spark/tree/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2 # noqa: E501
+        # https://github.com/apache/spark/tree/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2  # noqa: E501
         # Separated here because these allow for additional
         # commands such as Select From File
         # https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-file.html
--- a/src/sqlfluff/dialects/dialect_spark3.py
+++ b/src/sqlfluff/dialects/dialect_spark3.py
@@ -207,7 +207,12 @@ spark3_dialect.replace(
 )
 
 spark3_dialect.add(
-    # Add Hive Segments TODO : Is there a way to retrieve this w/o redefining?
+    BinaryfileKeywordSegment=StringParser(
+        "BINARYFILE",
+        KeywordSegment,
+        name="binary_file",
+        type="file_format",
+    ),
     JsonfileKeywordSegment=StringParser(
         "JSONFILE",
         KeywordSegment,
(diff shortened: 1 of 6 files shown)


src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -270,18 +274,43 @@
...
         "CSV",
         "JSON",
         "PARQUET",
         "ORC",
-        "JDBC",
-        # Community Contributed Data Sources
+        # Separated here because these allow for additional commands
+        # Similar to DataSourcesV2
         "DELTA",  # https://github.com/delta-io/delta
-        "XML",  # https://github.com/databricks/spark-xml
```
tunetheweb, line 284: You've lost XML support but not re-added it elsewhere. Was that intentional?

tunetheweb (changes requested):
Looks good. One question and one nit.

R7L208:
@tunetheweb - I'll work on the review comments today. Thank you!

Do you have a suggestion or similar PR you could point me to for adjusting L057 since identifier is required in this case? It shouldn't raise the rule here.

src/sqlfluff/dialects/dialect_spark3.py
tunetheweb: Oh missed that you pointed this out above:

> Linting fails due to special characters in identifiers (L057). These special keywords are required so would be an exception to L057. I've not yet worked on rules so would appreciate any guidance on where to get started on fixing this particular bug.

Happy to resolve this as it's an intended change. Sorry I missed that.

tunetheweb:
> @tunetheweb - I'll work on the review comments today. Thank you!
> 
> Do you have a suggestion or similar PR you could point me to for adjusting L057 since identifier is required in this case? It shouldn't raise the rule here.

In L057 there is already and exception for BigQuery:

```py
            # BigQuery table references are quoted in back ticks so allow dots
            #
            # It also allows a star at the end of table_references for wildcards
            # (https://cloud.google.com/bigquery/docs/querying-wildcard-tables)
            #
            # Strip both out before testing the identifier
            if (
                context.dialect.name in ["bigquery"]
                and context.parent_stack
                and context.parent_stack[-1].name == "TableReferenceSegment"
            ):
                if identifier[-1] == "*":
                    identifier = identifier[:-1]
                identifier = identifier.replace(".", "")
```

You could do something similar for this:

```py
            # Allow slashes in file references
            if (
                context.parent_stack
                and context.parent_stack[-1].type == "file_reference"
            ):
                identifier = identifier.replace("/", "")
```

src/sqlfluff/rules/L057.py
```diff
@@ -89,6 +89,41 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+            ]
+
+            if (
+                context.dialect.name in ["spark3"]
+                and context.parent_stack
+                and context.parent_stack[-1].name == "FileReferenceSegment"
+            ):
+                for allowed in spark3_allowed_identifiers:
+                    if allowed in identifier:
+                        identifier = identifier.replace(allowed, "")
```
tunetheweb, line 125: Seems like a lot. Rather than allowing them piece by piece I wonder if just easier to ignore these fields from this rule:

```suggestion
            # Ignore file names as are expected to contain a lot of characters
            if (
                context.parent_stack
                and context.parent_stack[-1].name == "FileReferenceSegment"
            ):
                return None
```

src/sqlfluff/rules/L057.py
```diff
@@ -89,6 +89,22 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
                     identifier = identifier[:-1]
                 identifier = identifier.replace(".", "")
 
+            # Spark3 file references for direct file query
+            # are quoted in back ticks to allow for identfiers common
+            # in file paths and regex patterns for path globbing
+            # https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-file.html
+            #
+            # Path Glob Filters (done inline for SQL direct file query)
+            # https://spark.apache.org/docs/latest/sql-data-sources-generic-options.html#path-global-filter  # noqa: E501
```
tunetheweb, line 98: I don't think you need the noqa as long as the URL is the ONLY thing on that line. It's cleaver enough to know that URLs can't be split.

```suggestion
            # https://spark.apache.org/docs/latest/sql-data-sources-generic-options.html#path-global-filter
```

src/sqlfluff/rules/L057.py
```diff
@@ -89,6 +89,22 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+            #
+
+            if (
+                context.dialect.name in ["spark3"]
+                and context.parent_stack
+                and context.parent_stack[-1].name == "FileReferenceSegment"
+            ):
+                return None
```
tunetheweb, line 106: Could we add a test to test/fixtures/rules/std_rule_cases/L057.yml? The other test cases you already added are only tested for parsing, not for rules compliance (as most peoples test SQL cases probably wouldn't pass ALL the rules!)

tunetheweb (changes requested):

R7L208:
Yep, just added them and running pytest locally then will push

tunetheweb (approved):
LGTM

PR #2563, opened 2022-02-04
Simple API: Fix bug where omitted parameters still override .sqlfluff

### LINKED ISSUE #2560: Empty allow/exclude list overrides config in the API
## Expected Behaviour
Calling API's `sqlfluff.fix` or `sqlfluff.lint` with `config_path` argument pointing to `.sqlfluff` with `exclude_rules` set should return list of errors without rules listed in `.sqlfluff`.

## Observed Behaviour
`sqlfluff.lint` and `sqlfluff.fix` both uses [get_simple_config](https://github.com/sqlfluff/sqlfluff/blob/cd8aceb5107abaf3639c4c1e2f7cca25aa193cc3/src/sqlfluff/api/simple.py#L13) to get the `FluffConfig`. Not only `get_simple_config` calls `FluffConfig.from_root` with `ignore_local_config=True` argument, but also builds up `overrides` dictionary:
```py
overrides = {
    "dialect": dialect,
    "rules": ",".join(rules) if rules is not None else None,
    "exclude_rules": ",".join(exclude_rules) if exclude_rules is not None else None,
}
```
which gets higher priority than the config referenced by a `config_path` argument. In the effect, if no `rules` or `exclude_rules` get passed to `lint` or `fix` functions, `overrides` overrides those config settings with `None` values, effectively allowing for all the rules.

## Steps to Reproduce
Add .sqlfluff file to your project:
```ini
[sqlfluff]
exclude_rules = L027,L029
```
and then call Python function:
```py
import sqlfluff
import pathlib

sql = "SELECT TRIM(name) AS name FROM some_table"
lint_results = sqlfluff.lint(sql, config_path=str(pathlib.Path.cwd().joinpath(".sqlfluff")))
for lint_result in lint_results:
  if lint_result["code"] in ["L027", "L029"]:
    raise Exception(lint_result["code"])
```

## Dialect
--

## Version
sqlfluff, version 0.9.0

However, reading the freshest code on GitHub in [simple.py](https://github.com/sqlfluff/sqlfluff/blob/cd8aceb5107abaf3639c4c1e2f7cca25aa193cc3/src/sqlfluff/api/simple.py#L13) and [config.py](https://github.com/sqlfluff/sqlfluff/blob/cd8aceb5107abaf3639c4c1e2f7cca25aa193cc3/src/sqlfluff/core/config.py#L454) it looks like the problem persists.

barrywhart:
Fixes #2560

### Pull Request checklist

  - Other.

### DIFF
--- a/test/api/simple_test.py
+++ b/test/api/simple_test.py
@@ -1,6 +1,7 @@
 """Tests for simple use cases of the public api."""
 
 import json
+import pathlib
 
 import pytest
 
@@ -254,6 +255,41 @@ def test__api__config_path():
 
     # Compare JSON from parse to expected result.
     assert parsed == expected_parsed
+
+
+@pytest.mark.parametrize(
+    "exclude_sqlfluff,kwargs,expected",
+    [
+        (
+            # No override from API, so uses .sqlfluff value
+            "L027,L029",
+            {},
+            set(),
+        ),
+        (
+            # API overrides, so it uses that
+            "L027,L029",
+            dict(exclude_rules=["L027"]),
+            {"L029"},
+        ),
+    ],
+)
+def test__api__config_override(exclude_sqlfluff, kwargs, expected, tmpdir):
+    """Test that parameters to lint() override .sqlfluff correctly (or not)."""
+    tmp_path = pathlib.Path(str(tmpdir))
+    config_path = tmp_path / ".sqlfluff"
+    with tmpdir.as_cwd():
+        config_path.write_text(
+            """[sqlfluff]
+exclude_rules = {}""".format(
+                exclude_sqlfluff
+            )
+        )
+        sql = "SELECT TRIM(name) AS name FROM some_table"
+        lint_results = sqlfluff.lint(sql, config_path=str(config_path), **kwargs)
+        assert expected == {"L027", "L029"}.intersection(
+            {lr["code"] for lr in lint_results}
+        )
 
 
 def test__api__invalid_dialect():
--- a/test/api/simple_test.py
+++ b/test/api/simple_test.py
@@ -1,7 +1,6 @@
 """Tests for simple use cases of the public api."""
 
 import json
-import pathlib
 
 import pytest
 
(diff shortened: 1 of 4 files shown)


test/api/simple_test.py
```diff
@@ -256,6 +257,41 @@ def test__api__config_path():
...
+            """[sqlfluff]
+exclude_rules = {}""".format(
+                exclude_sqlfluff
+            )
+        )
+        sql = "SELECT TRIM(name) AS name FROM some_table"
+        lint_results = sqlfluff.lint(sql, config_path=str(config_path), **kwargs)
+        assert expected == {"L027", "L029"}.intersection(
+            {lr["code"] for lr in lint_results}
+        )
```
tunetheweb, line 292: Rather than writing to a temp dir (which sounds flakey, prone to conflcits, and not very scalable) could we not just have a test `.sqlfluff` config file in our code in our `tests` dir and use that? I believe we use that model elsewhere.

barrywhart: This test is also a little unusual in that it is made up of both SQL and config files but also the parameters that will be passed to the simple API (note that `expected` is a Python dictionary that we pass to a function).

tunetheweb: I’m more concerned with two tests running at the same time, and getting some sort of race condition as the file is written by one process and read by another.

This could even be within the one run if using our parallel processing technique.

`tmpdir` is an OS-agnostic way of getting access to the temp directory location, but it doesn’t create a unique dir/file within that AFAIK to avoid that situation.

Maybe we don’t need a new file for this and can just use one of the existing ones to avoid bloating the code base?

barrywhart: Ah. This isn't the same tmpdir you're thinking of. This is a purest "fixture" that creates a unique directory every time (it'll be a subdirectory of the "main" temp directory). So it's guaranteed safe from those kinds of race conditions.

You can see the docs here. PyTest has a bunch of nice built-in features like this that make life easier.

https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture

barrywhart: Ok, moved the `.sqlfluff` file to a directory in git -- no temp directories here. 🙏

tunetheweb (approved):

Issue #2500, opened 2022-01-28
Snowflake dialect: ALTER TABLE ADD COLUMN statements do not parse if multiple columns specified


erevear:
## Expected Behaviour

Potentially seems related to this change: [ALTER TABLE DROP COLUMN statements in Snowflake fail if dropping multiple columns ](https://github.com/sqlfluff/sqlfluff/issues/2344)
The DROP for multiple columns works now, however, ADD COLUMN does not, where it did previously

## Observed Behaviour

Parsing fails on ALTER TABLE ADD COLUMN with multiple columns

## Steps to Reproduce
Example SQL (add_column_example.sql)
```
ALTER TABLE MY_TABLE ADD COLUMN COLUMN_1 varchar, COLUMN_2 integer;
```

1. On the above SQL run `sqlfluff parse add_column_example.sql --dialect=snowflake` or 
`sqlfluff lint add_column_example.sql --dialect=snowflake --ignore linting`

2. See parsing error

## Dialect

Snowflake
## Version
Include the output of `sqlfluff --version` along with your Python version.
If you are using dbt then please additionally include the dbt version.
sqlfluff version is 0.9.3
python version 3.9.7

## Configuration
No, config  (aside from dialect), as this is a parsing issue

tunetheweb:
What is the parsing error? Including the parse tree will help get a quicker diagnosis.

erevear:
This is the parse tree and error I get when I run the parse command:
```
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    statement:
[L:  1, P:  1]      |        alter_table_statement:
[L:  1, P:  1]      |            keyword:                                          'ALTER'
[L:  1, P:  6]      |            whitespace:                                       ' '
[L:  1, P:  7]      |            keyword:                                          'TABLE'
[L:  1, P: 12]      |            whitespace:                                       ' '
[L:  1, P: 13]      |            table_reference:
[L:  1, P: 13]      |                identifier:                                   'MY_TABLE'
[L:  1, P: 21]      |            whitespace:                                       ' '
[L:  1, P: 22]      |            alter_table_table_column_action:
[L:  1, P: 22]      |                keyword:                                      'ADD'
[L:  1, P: 25]      |                whitespace:                                   ' '
[L:  1, P: 26]      |                keyword:                                      'COLUMN'
[L:  1, P: 32]      |                whitespace:                                   ' '
[L:  1, P: 33]      |                column_reference:
[...]

PR #2578, opened 2022-02-08
Snowflake ALTER TABLE: Add multiple columns

erevear:

Attempt at a fix for https://github.com/sqlfluff/sqlfluff/issues/2500

Moving the ALTER TABLE ADD COLUMN statement elements into a Delimited object in order to add more than one at a time.

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -960,47 +960,53 @@ class AlterTableTableColumnActionSegment(BaseSegment):
...
-                            "AUTOINCREMENT",
-                            "IDENTITY",
-                        ),
-                        OneOf(
-                            # ( <start_num>, <step_num> )
-                            Bracketed(
-                                Ref("NumericLiteralSegment"),
-                                Ref("CommaSegment"),
-                                Ref("NumericLiteralSegment"),
+                Sequence(
```
tunetheweb, line 963: You need to run `black` on this file. Guessing the indentation means some lines are too long and need to be moved down. `black` should be able to do this for you automatically.

test/fixtures/dialects/snowflake/snowflake_select_union.yml
```diff
@@ -19,7 +19,7 @@ file:
...
           select_clause_element:
             literal: '2'
-    - orderby_clause:
-      - keyword: ORDER
-      - keyword: BY
-      - literal: '1'
+        unparsable:
+        - keyword: ORDER
+        - keyword: BY
+        - raw: '1'
```
tunetheweb, line 25: This looks wrong. But I just tried your branch and this works. Think you need to rerun the generate script to correct this:

```
```

tunetheweb (changes requested):

test/fixtures/dialects/snowflake/snowflake_select_union.yml
erevear: Yes! I re-ran the generator and this was corrected, so the file is no longer in the change

src/sqlfluff/dialects/dialect_snowflake.py
erevear: Got it! I installed the pre commit hook and black reformatted the file through that, so it passes linting. Does that solve the issue here, or is there another step?

tunetheweb: Looks good now.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -960,47 +960,54 @@ class AlterTableTableColumnActionSegment(BaseSegment):
...
```
tunetheweb, line 963: I don’t think you need this Sequence as it only includes one component (the Delimited). Can we remove it?

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_snowflake.py
tunetheweb: Have made this change (and another unnecessary `Sequence`) so we can include this in 0.10.0.

tunetheweb (approved):

Issue #1296, opened 2021-08-17
"sqlfluff fix" fails with "Some operations failed... Some errors cannot be fixed or there is another error blocking it."


barrywhart:
-- If this is a parsing or linting issue, please include a minimal SQL example which reproduces the issue, along with the `sqlfluff parse` output, `sqlfluff lint` output and `sqlfluff fix` output when relevant. Finally, please add any labels that are related to the issue before submitting so that it can be addressed easily! --

## Expected Behaviour

File successfully fixed (or left untouched).

## Observed Behaviour

`sqlfluff fix` reports an error:
```
 sqlfluff fix sql/campaign_performance_queries/constraints/_postcondition_ignore_internal_accounts.sql
==== finding fixable violations ====
== [sql/campaign_performance_queries/constraints/_postcondition_ignore_internal_accounts.sql] FAIL
L:   9 | P:  59 | L016 | Line is too long
==== fixing violations ====
1 fixable linting violations found
Are you sure you wish to attempt to fix these? [Y/n] ...
Attempting fixes...
Persisting Changes...
== [sql/campaign_performance_queries/constraints/_postcondition_ignore_internal_accounts.sql] FAIL
Done. Some operations failed. Please check your files to confirm.
Some errors cannot be fixed or there is another error blocking it.
```

The exit code reports failure, which means that using `sqlfluff fix` in a pre-commit hook prevents a user from committing the file in this case. Two suggestions:
* Fix the underlying bug
* Consider whether `sqlfluff fix` should instead exit with status `0` in this case, because it's not clear the user can do anything about this.

## Steps to Reproduce

```
-- Checks to make sure that no internal accounts are included in the campaign_performance table.
SELECT user_id
FROM
    `{{gcp_project}}.{{dataset}}.campaign_performance`
WHERE
    user_id IN (
        SELECT user_id
        FROM
            `{{BI_INTERNAL_USERS}}` {{table_at_job_start}}
    )
```

## Version
SQLFluff: 0.6.3
Python: 3.7.9

## Configuration
```
[sqlfluff]
verbose = 0
nocolor = False
dialect = bigquery
templater = jinja
rules = None
exclude_rules = L019,L025,L027,L029,L031,L032,L034
recurse = 0

[sqlfluff:templater:jinja:macros]

# Some rules can be configured directly from the config common to other rules.
[sqlfluff:rules]
tab_space_size = 4
max_line_length = 120

# Some rules have their own specific config.
[...]

PR #2586, opened 2022-02-09
Prepare 0.10.0

tunetheweb:
We've lots of changes queuing up again so let's look at a 0.10.0 release.

Discussed with the maintainers and we have two breaking changes (dropping older DBT support, and changes on how `sqlfluff fix` works on files with parsing and templating errors) so we're not going for a 0.9.5, and don't feel we're quite ready for a 1.0.0 so that's why 0.10.0 was chosen.

Have tentatively put a release date in for Friday for now. Will keep an eye on pull requests and add them in as they are merged.

FYI: @iajoiner @rpr-ableton @jpy-git @R7L208 @erevear that all have open PRs.

Please comment below if you have other PRs in progress that are NOT YET in the pull request queue and think you'll make it and will try to hold off releasing until we've an update from them.


CHANGELOG.md
```diff
@@ -10,6 +10,61 @@ Note: Changes are now automatically tracked in [GitHub](https://github.com/sqlfl
...
 
 
+## [0.10.0] - 2022-02-11
+
+## Highlights
+
+Major changes include:
+
+* Dropping support of DBT < 0.2.0 ** BREAKING CHANGE **
+* `sqlfluff fix` no will no longer fix SQL containing parsing or templating errors ** BREAKING CHANGE **
```
barrywhart, line 20: Should we mention the new `--fix-even-unparsable` option here, in case people don't like the change?  I'm a little bit unsure about this, because, while some people may want it, advertising it may cause people to use it even though we'd rather they didn't. 🤔

tunetheweb: I say now. Those that really want it can find it in the docs, or raise an issue. Let's not advertise it.

barrywhart (approved):

barrywhart:
I'm working on issue #1296 and expect to have a PR for it in the next ~30 minutes.

tunetheweb:
Will wait for ya.

Then let's do this!!!

PR #2608, opened 2022-02-10
Add Oracle at and double at sign (execution symbol)

r0fls:
### Summary
This adds support for the Oracle `@` and `@@` syntax. This is only "loosely" supported at this time, namely it treats these like comments, so they aren't invalid, but won't really benefit from full on parsing/validation at this point. However, it's an improvement versus having these appear invalid altogether.

### DIFF
--- a/src/sqlfluff/dialects/dialect_oracle.py
+++ b/src/sqlfluff/dialects/dialect_oracle.py
@@ -31,19 +31,18 @@ oracle_dialect.insert_lexer_matchers(
             r"PROMPT([^(\r\n)])*((?=\n)|(?=\r\n))?",
             CommentSegment,
         ),
-        #RegexLexer(
-        #    "execution_sign",
-        #    r"^@@(?i).*\.sql(\s)(.*)$",
-        #    CodeSegment
-        #)
+        RegexLexer(
+            "double_execution_sign",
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_oracle.py
```diff
@@ -28,6 +30,16 @@
...
+        ),
+        RegexLexer(
+            "double_execution_sign",
+            r"@@([^(\r\n)])*((?=\n)|(?=\r\n))?",
+           CommentSegment, 
+        ),
+        RegexLexer(
+            "execution_sign",
+            r"@@([^(\r\n)])*((?=\n)|(?=\r\n))?",
+           CommentSegment,
```
tunetheweb, line 42: What's the difference between these two? Should one be just a single `@`?

r0fls: Yes, thank you

tunetheweb: That was going to be my next question! 😁

tunetheweb: Could we merge them both into the one test SQL file? Don’t see much value in them being separate when they’re basically there same?

r0fls: Done!

PR #2619, opened 2022-02-12
BigQuery UNPIVOT and PIVOT fixes

### LINKED ISSUE #2616: BigQuery: Issues With Parsing Quotation Marks Around String Literal Aliases Used in UNPIVOT Operations
## Expected Behaviour
Whenever valid BigQuery SQL code is run through ```sqlfluff lint```, it does not produce unexpected parse errors.   Since literals can either be quoted using single- or double-quotation marks according to BigQuery documentation, this suggests that changing string literal aliases from single to double quotation marks should not trigger parsing errors if the underlying SQL is still valid.

In addition, in contexts where variable names have to be given as string literals in BigQuery, SQLFluff should understand that these items need to be quoted and should not produce the L059 error.

## Observed Behaviour
When row_value_alias values used in an UNPIVOT operation in BigQuery (see definition [here](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#unpivot_operator)) are quoted using double-quotation marks rather than single-quotation marks a PARSE error is produced.

When row_value_alias values used in an UNPIVOT operation in BigQuery (see definition [here](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#unpivot_operator)) are quoted using single-quotation marks, SQLFluff suggests removing quotation marks around the alias even though they are required in this context.

## Steps to Reproduce
(1). Save the following (nonsense but valid) jinja/dbt SQL file:
```sql
    SELECT
        *
    FROM model
    UNPIVOT(
      (A, B)
      FOR year
      IN ((C, D) AS "year_2011", (E, F) AS "year_2012"))
```
(2). Set BigQuery as the dialect and jinja as the templater and run sqlfluff lint on this file.  You will get a parse error.

(3).  Change the file as follows and re-run and the parse error disappears, confirming the issue is with the double-quotation marks used as string literal names.
```sql
    SELECT
        *
    FROM model
    UNPIVOT(
      (A, B)
      FOR year
      IN ((C, D) AS 'year_2011', (E, F) AS 'year_2012'))
```

However, this now produces L059 errors:
- Unnecessary quoted identifier 'year_2011'.
- Unnecessary quoted identifier 'year_2012'.

These errors are also erroneous because these alias [must be string literals](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#rules_for_row_value_alias), meaning that they have to be quoted in this context.  Very confusing for BigQuery to use different rules for aliases for UNPIVOT operations but that's the case here unfortunately.

## Dialect
BigQuery

## Version
SQLFluff version is 0.10.0.  Python version is 3.8.10.
I'm using dbt 1.0.1 but the same issue occurs when no templater is used.

## Configuration
```
[sqlfluff]
dialect = bigquery

[sqlfluff:rules]
max_line_length = 120
comma_style = leading

[sqlfluff:rules:L010]
capitalisation_policy = upper

[sqlfluff:rules:L030]
capitalisation_policy = upper
```

tunetheweb:

Fixes #2616

Also fixes small issue with `PIVOT` statement not accepting multiple functions, that I noticed while looking if same fix needed to be applied there (it doesn't).

Nope

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -1103,7 +1103,6 @@ class FromUnpivotExpressionSegment(BaseSegment):
                                 ),
                             ),
                             Ref("UnpivotAliasExpressionSegment", optional=True),
-                            # Ref("AliasExpressionSegment", optional=True),
                         ),
                     ),
                 ),
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -1048,15 +1048,8 @@ class UnpivotAliasExpressionSegment(BaseSegment):
     match_grammar = Sequence(
         Ref.keyword("AS", optional=True),
         OneOf(
-            Sequence(
-                Ref("SingleIdentifierGrammar"),
-                # Column alias in VALUES clause
-                Bracketed(Ref("SingleIdentifierListSegment"), optional=True),
-            ),
-            OneOf(
-                Ref("SingleQuotedLiteralSegment"),
-                Ref("DoubleQuotedLiteralSegment"),
-            ),
+            Ref("SingleQuotedLiteralSegment"),
+            Ref("DoubleQuotedLiteralSegment"),
         ),
     )
 
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -1081,7 +1081,6 @@ class FromUnpivotExpressionSegment(BaseSegment):
                 Bracketed(
                     Delimited(Ref("SingleIdentifierGrammar")),
                     Ref("UnpivotAliasExpressionSegment", optional=True),
-                    # Ref("AliasExpressionSegment", optional=True),
                 ),
             ),
             Bracketed(
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -84,6 +84,13 @@ bigquery_dialect.add(
         type="literal",
         trim_chars=('"',),
     ),
+    SingleQuotedLiteralSegment=NamedParser(
+        "single_quote",
+        CodeSegment,
+        name="quoted_literal",
+        type="literal",
+        trim_chars=("'",),
+    ),
     DoubleQuotedUDFBody=NamedParser(
         "double_quote",
         CodeSegment,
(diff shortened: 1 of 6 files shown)


src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -84,6 +84,13 @@
         type="literal",
         trim_chars=('"',),
     ),
+    SingleQuotedLiteralSegment=NamedParser(
+        "single_quote",
+        CodeSegment,
+        name="quoted_literal",
+        type="literal",
+        trim_chars=("'",),
```
barrywhart, line 92: I'm not familiar with `trim_chars`. What does it do, and why is it needed here?

test/fixtures/dialects/bigquery/select_unpivot.sql
```diff
@@ -17,3 +17,11 @@ UNPIVOT(
   (first_half_sales, second_half_sales)
   FOR semesters
   IN ((Q1, Q2) AS 'semester_1', (Q3, Q4) AS 'semester_2'));
+
+SELECT
+    a AS 'barry'
```
barrywhart, line 22:
```suggestion
    a AS 'barry_squared'
```

Just joking! 🎉🤣

src/sqlfluff/dialects/dialect_bigquery.py
tunetheweb: Direct copy from other simile functions. It basicaly takes the starting quote and ending quote marks off IIRC leaving the actual value.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -1056,7 +1080,8 @@ class FromUnpivotExpressionSegment(BaseSegment):
                 "IN",
                 Bracketed(
                     Delimited(Ref("SingleIdentifierGrammar")),
-                    Ref("AliasExpressionSegment", optional=True),
+                    Ref("UnpivotAliasExpressionSegment", optional=True),
+                    # Ref("AliasExpressionSegment", optional=True),
```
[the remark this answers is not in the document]
tunetheweb: Of course. Dunno why I left that!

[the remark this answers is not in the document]
tunetheweb:
```suggestion
```

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -1078,7 +1103,8 @@ class FromUnpivotExpressionSegment(BaseSegment):
                                     min_delimiters=1,
                                 ),
                             ),
-                            Ref("AliasExpressionSegment", optional=True),
+                            Ref("UnpivotAliasExpressionSegment", optional=True),
+                            # Ref("AliasExpressionSegment", optional=True),
```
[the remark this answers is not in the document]
tunetheweb:
```suggestion
```

test/fixtures/rules/std_rule_cases/L059.yml
```diff
@@ -241,10 +241,24 @@ test_pass_datetime_redshift:
...
+      dialect: bigquery
+
 test_pass_quoted_identifier_keyword_tsql:
   pass_str: |
     SET QUOTED_IDENTIFIER ON
     GO
   configs:
     core:
       dialect: tsql
+
```
tunetheweb, line 264:
```suggestion
```

PR #2620, opened 2022-02-12
Ensure ordering of fix compatible and config in rules docs

zhongjiajie:

This patch try to strict the order of rule's hand
writing docstring, fix_compatible, note, Configuration
and example pattern.

* Rule's hand writing docstring(expect example pattern
  and note) is always in the fist place
* (Optional) Fix compatible in the second place
* (Optional) Note in the third place
* (Optional) Configuration in the fourth place
* Example pattern would always in the tail

It is compatible all orders of doc decorators

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/rules/doc_decorators.py
+++ b/src/sqlfluff/core/rules/doc_decorators.py
@@ -69,10 +69,15 @@ def document_configuration(cls, ruleset="std"):
         return cls
     # Add final blank line
     config_doc += "\n"
-    # Match `**Anti-pattern**`, then insert configuration before the first occurrences
-    pattern = re.compile("(\\s{4}\\*\\*Anti-pattern\\*\\*)", flags=re.MULTILINE)
 
-    cls.__doc__ = pattern.sub(f"\n{config_doc}\n\\1", cls.__doc__, count=1)
+    if "**Anti-pattern**" in cls.__doc__:
+        # Match `**Anti-pattern**`, then insert configuration before the first occurrences
+        pattern = re.compile("(\\s{4}\\*\\*Anti-pattern\\*\\*)", flags=re.MULTILINE)
+        cls.__doc__ = pattern.sub(f"\n{config_doc}\n\\1", cls.__doc__, count=1)
+    else:
+        # Match last `\n` or `.`, then append configuration
+        pattern = re.compile("(\\.|\\n)$", flags=re.MULTILINE)
+        cls.__doc__ = pattern.sub(f"\\1\n{config_doc}\n", cls.__doc__, count=1)
     return cls
 
 
--- a/src/sqlfluff/core/rules/doc_decorators.py
+++ b/src/sqlfluff/core/rules/doc_decorators.py
@@ -2,6 +2,7 @@
 
 from sqlfluff.core.rules.config_info import get_config_info
 from sqlfluff.core.rules.base import rules_logger  # noqa
+import re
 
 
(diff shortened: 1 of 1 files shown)


zhongjiajie:
A quick look about the different

zhongjiajie:
> A quick look at the different
> image

zhongjiajie:
PTAL @tunetheweb @barrywhart

src/sqlfluff/core/rules/doc_decorators.py
```diff
@@ -60,12 +69,16 @@ def document_configuration(cls, ruleset="std"):
         return cls
     # Add final blank line
     config_doc += "\n"
-    # Add the configuration section immediately after the class description
-    # docstring by inserting after the first line break, or first period,
-    # if there is no line break.
-    end_of_class_description = "." if "\n" not in cls.__doc__ else "\n"
 
-    cls.__doc__ = cls.__doc__.replace(end_of_class_description, "\n" + config_doc, 1)
+    if "**Anti-pattern**" in cls.__doc__:
```
barrywhart, line 73: If I understand correctly, this code reorders parts of the hand-written rule documentation. Could we instead change this to **detect** and **force** rule developers to write the docs in the desired order? I think that would be better, because I worry this code become more complex or fragile over time.

zhongjiajie: It not reorder hand-written rule documentation, it is about doc decorators `document_fix_compatible` and `document_configuration` place to hand-written rule documentation.
when we add decorator to rule:
* Both `document_fix_compatible` and `document_configuration` will insert the first line of docstring.
* But when we use different order of `document_fix_compatible` and `document_configuration`, it would insert fix flag and config in different place.

What more, this patch also change the place we inserted.
* Before that, we insert in the end of first line

```markdown
The first line docstring
<--- the place we will insert decorators
Addition information about first line

example
```

* After that, we insert before the example
```markdown
The first line docstring

Addition information about first line

<--- the place we will insert decorators

example
```

tunetheweb, line 73: I think we have another rules saying every rule MUST have an anti-pattern. So not sure we need this `if`/`else` but no harm.

tunetheweb (approved):
I'm happy with this.

It is a bit more regex than I'd like, so a little less clean than thew original code. But the order is better.

You happy to merge @barrywhart ?

src/sqlfluff/core/rules/doc_decorators.py
tunetheweb: Agreed. It's not re-ordering and it was always inserting these two extra pieces. Just being a bit more complicated about where to insert them now.

barrywhart (approved):
Looks good!

src/sqlfluff/core/rules/doc_decorators.py
zhongjiajie: Hi @tunetheweb, we have a specific rule in `Rule_016` which you could see in https://github.com/sqlfluff/sqlfluff/blob/88df89f4c5fbfedd250e6aabd1dc69710a499951/src/sqlfluff/rules/L016.py#L26. we would skip the docstring test. and the test as below https://github.com/sqlfluff/sqlfluff/blob/7da5c175e3647359a4ea6d0c192aa23e5e8549d4/test/core/rules/docstring_test.py#L13-L29

Issue #2617, opened 2022-02-11
L023 for CTE is unfixable with newline instead of space


juhoautio:
Lint & fix are both failing on this kind of pattern:

```sql
WITH cte AS
(
```

This can be fixed manually by replacing the newline after `AS` with a space, like so:

```sql
WITH cte AS (
```

It would be nice if sqlfluff can fix this automatically.

Or is this rule actually wrong? The "original query" (see below) looks ok to me.

The rule documentation for [L023](https://docs.sqlfluff.com/en/stable/rules.html#sqlfluff.core.rules.Rule_L023) demonstrates fixing `AS(` to `AS (`. But is it correct for `lint` to require that `(` should be on the same line?

Btw the error message is confusing  because there _is_ whitespace after the `AS` keyword in my example: there's a newline. The parse tree categorizes as `newline` while there are other parts that are `whitespace`. So the technical definition seems to be that newlines are not `whitespace`. But isn't it confusing to use that kind of definition in user-facing docs? To be exact, I suppose it can be called _horizontal whitespace_.

The lint error message should be improved for this lint rule if it remains as it is, even if it becomes auto-fixable.

## Steps to Reproduce

`myquery.sql`:

```sql
WITH cte AS
(
    SELECT m_rule_id
    FROM t
)

SELECT * FROM cte;
```

### fix attempt

```
sqlfluff fix -f myquery.sql
```

```
==== finding fixable violations ====
==== no fixable linting violations found ====
All Finished!
  [1 unfixable linting violations found]
```

### lint
```
sqlfluff lint myquery.sql
```

```
== [myquery.sql] FAIL
L:   1 | P:  10 | L023 | Single whitespace expected after 'AS' in 'WITH' clause.
All Finished!
```

### parse
```
sqlfluff parse myquery.sql
```
<details>

```
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    statement:
[L:  1, P:  1]      |        with_compound_statement:
[L:  1, P:  1]      |            keyword:                                          'WITH'
[L:  1, P:  5]      |            whitespace:                                       ' '
[L:  1, P:  6]      |            common_table_expression:
[L:  1, P:  6]      |                identifier:                                   'cte'
[L:  1, P:  9]      |                whitespace:                                   ' '
[L:  1, P: 10]      |                keyword:                                      'AS'
[L:  1, P: 12]      |                newline:                                      '\n'
[L:  2, P:  1]      |                bracketed:
[...]

PR #2623, opened 2022-02-12
L023 to also fix extra newlines in CTE

juhoautio:

fixes https://github.com/sqlfluff/sqlfluff/issues/2617

No.

Any amount of whitespace & newlines are now covered.

Caveats:
- At least a comment section between `AS` and `(` can result in an unfixable error.
- postgres dialect seems to have some extensions to the CTE syntax. Some of those dialect specific cases may not be still fixable after this change.

### Pull Request checklist

  - Other.

### DIFF
--- a/test/fixtures/rules/std_rule_cases/L023.yml
+++ b/test/fixtures/rules/std_rule_cases/L023.yml
@@ -1,6 +1,41 @@
 rule: L023
 
 test_fail_cte_no_space_after_as:
-  # Check fixing of single space rules
+  # Check fixing of single space rule when space is missing
   fail_str: WITH a AS(select 1) select * from a
   fix_str: WITH a AS (select 1) select * from a
+
+test_fail_multiple_spaces_after_as:
+  # Check fixing of single space rule on multiple spaces
+  fail_str: WITH a AS  (select 1) select * from a
+  fix_str: WITH a AS (select 1) select * from a
+
+test_fail_cte_newline_after_as:
+  # Check fixing of replacing newline with space
+  fail_str: |
+    WITH a AS
+    (
+      select 1
+    )
+    select * from a
+  fix_str: |
+    WITH a AS (
+      select 1
+    )
+    select * from a
+
+test_fail_cte_newline_and_spaces_after_as:
+  # Check stripping newlines and extra whitespace
+  fail_str: |
+    WITH a AS
+
+
+        (
+      select 1
+    )
+    select * from a
+  fix_str: |
+    WITH a AS (
+      select 1
+    )
+    select * from a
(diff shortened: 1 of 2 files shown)


test/fixtures/rules/std_rule_cases/L023.yml
```diff
@@ -1,6 +1,41 @@
...
+  # Check stripping newlines and extra whitespace
+  fail_str: |
+    WITH a AS
+
+
+        (
+      select 1
+    )
+    select * from a
+  fix_str: |
```
juhoautio, line 37: A general question about the `yaml_test_cases_test.py`.

IMHO it's confusing and possibly dangerous that even if the `fail_str` + `fix_str` comparison passes, it could be possible that `lint` would _not_ pass. I presume `fix` would also return non-zero exit code in such case. Would it make sense to also ensure here that there are no unfixable errors remaining?

For example, this yaml test passes at the moment:
```
test_fail_comment:
  # Check fixing of single space rule on when there's a comment before bracketed section
  fail_str: |
    WITH a AS -- comment
    (select 1) select * from a
  fix_str: |
    WITH a AS -- comment
    (select 1) select * from a
```

If I modify the `rules.py` to be like this:
```python
        if test_case.fix_str:
            assert res == test_case.fix_str
            # MY NEW CODE BELOW:
            assert_rule_pass_in_sql(
                test_case.rule,
                test_case.fix_str,
                configs=test_case.configs,
            )
```
Then the test case fails (like I would expect) as follows:

```
Errors Found: [<SQLLintError: rule L023 pos:(1, 8), #fixes: 0, description: Single whitespace expected after 'AS' in 'WITH' clause.>]
```

Maybe there can be legitimate cases where the point of test case is to just check that fix doesn't touch the query (or that it fixes it to some extent but not fully). But IMHO it should be explicitly defined by the test cases then that fix is expected to (also) return an error instead of succeeding.

tunetheweb: I’d say that’s a bad test, in that the `fix_str` shouldn’t be a copy of the `fail_str`. It should either be omitted (so it fails, but there’s an acceptance the rule can’t fix that scenario - this is what would be used for the rules that are not fixable), or it should be different.

But would be good to enforce this so no bad test cases can be added and “silently fail” as you state. If your proposed extra code does that then I think that would be a good extra addition. Maybe as a separate PR to this one?

juhoautio: > I’d say that’s a bad test, in that the fix_str shouldn’t be a copy of the fail_str.

True. Indeed I'm more concerned if there can be cases where some fixing does happen (so that `fix_str` & `fail_str` differ) but but the fixing is not enough to make the rule pass.

> Maybe as a separate PR to this one?

Sure thing!

barrywhart:
Can you look into the failing tests?

barrywhart (approved):

Issue #2337, opened 2022-01-17
Warn on unused inline ignores


jpy-git:
This is a nice feature that `mypy` offers via
```cfg
[mypy]
warn_unused_ignores = True
```

The idea is that occasionally when using mypy you can run into a false positive (usually due to some incorrect/incomplete stubs in typeshed) and as a result you can add an inline ignore (`# type: ignore`) to ignore it.

However, over time these issues get fixed and your inline ignore can become redundant. With the flag above mypy will raise a lint for these redundant inline ignores, therefore allowing the user to clean them up from their code over time.

There are pretty obvious parallels here with how users might use our inline ignores to ignore a parse error or a misbehaving rule that we fix in a subsequent release but they never clean up the old inline ignores from their codebase.

In terms of implementing this I think there are a couple discussion points:
- When we parse the `NoQaDirective` it will be given some attribute like `exhausted=False` and when we then mask the lints we mark off the `NoQaDirective` as exhausted if it actually corresponds to a detected lint. Then flag the rest as unused and raise a lint for them.
- Like mypy this would not be enabled by default, it's more of an advanced feature for those who want stricter linting or are maybe doing a one off cleanup. It can be a config option + CLI flag like in mypy.
- OPTIONAL: We currently have a couple different syntaxes for inline ignoring that we could possibly simplify in favour of the simpler `mypy`/`flake8` style (i.e. just the `-- noqa: L006` style ones rather than pylint style directives) ([docs](https://docs.sqlfluff.com/en/stable/rules.html#inline-ignoring-errors)). This one's not essential, just my opinion based on how I've used the inline ignores and could be a good chance to remove them if they're not used since it would simplify this work.

tunetheweb:
I like it!

One thing to consider is different versions of SQLFluff.

For example I use SQLFluff locally, but for some projects if use it via GitHub Super Linter in CI and that is usually a version or two behind (cause of our frequent release cadence and the fact they only update dependencies once a week or so).

So could get annoying if it warns on something I don’t want to remove yet.

Or maybe there could be a command line switch (I don’t want it in `.sqlfluffini` and that is shared between both) so can ignore those errors locally until CI catches up.

jpy-git:
@tunetheweb Yeah it's easy enough to have it in either config or as a CLI flag without much extra work. Both use cases are valid.

mypy offers a CLI flag as well for the same reason.
[...]

Issue #2180, opened 2021-12-23
Document how tests are organized


juhoautio:
As a contributor I would like to find a high level doc about how the tests are organized.

For dialects there is already some helpful documentation in the [Contributing Dialect Changes wiki page](https://github.com/sqlfluff/sqlfluff/wiki/Contributing-Dialect-Changes#adding-test-cases-for-your-changes).

Other than that, the big picture is missing. Is it worth documenting that for contributors? I think a very high level description would be helpful enough.

I haven't examined the repo in full, but based on https://github.com/sqlfluff/sqlfluff/pull/2167#discussion_r774253690 I have learned this:

- For "more complex interactions between multiple rules", run `std_fix_auto_test.py` which includes `fixtures/linter/autofix` 
- For pure unit testing of individual rules, run `yaml_test_cases_test.py` which includes `fixtures/rules/std_rule_cases`

PR #2624, opened 2022-02-12
Improve rule yaml tests: assert that fix_str passes the rule

juhoautio:
(Originally proposed in https://github.com/sqlfluff/sqlfluff/pull/2623#discussion_r805205211)

Enforce that `fix_str` passes or `violations_after_fix` is defined.

The rule tests for _fixing_ are written in YAML so that both `fail_str` and `fix_str` are set.

So far it has been asserted that:
- the rule fails on the `fail_str`
- fix formats the query so that the result matches the `fix_str`

However, it's possible that the `fix_str` value would still not pass the rule. In other words, if the value of `fix_str` would be also tested as `pass_str`, the test would fail. This is surprising. It means that some test cases may pass even though the fix is just partial. A special case of that is that nothing is fixed – the case where `fail_str` and `fix_str` have the same value.

**This PR** adds a new check: the `fix_str` must pass the rule. This prevents us from adding broken test cases for fix that pass silently.

Example output for a test that fails this new assertion:

> Errors Found: [<SQLLintError: rule L004 pos:(2, 4), #fixes: 0, description: Incorrect indentation type found in file. The indent occurs after other text, so a manual fix is needed.>]

> No violations_after_fix was provided, but the SQL after fix violates the rule. To accept a partial fix, violations_after_fix must be set.

There were only 2 cases left that would have not passed this new assertion. Both cases were already fixed in [L004.yml](https://github.com/sqlfluff/sqlfluff/pull/2718/files#diff-6bf6914958b1b802eaa41356518b010fa0222a8f9595d734a90961d464e4d323)

No.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/testing/rules.py
+++ b/src/sqlfluff/testing/rules.py
@@ -190,7 +190,7 @@ def rules__test_helper(test_case):
             assert_violations_before_fix(test_case, violations_before_fix)
         # If a `fixed` value is provided then check it matches
         if test_case.fix_str:
-            assert res == test_case.fix_str
+            assert res == test_case.fix_str, "result of fix doesn't match the fix_str"
             assert_rule_pass_in_sql(
                 test_case.rule,
                 test_case.fix_str,
--- a/src/sqlfluff/testing/rules.py
+++ b/src/sqlfluff/testing/rules.py
@@ -106,7 +106,7 @@ def assert_rule_pass_in_sql(code, sql, configs=None, msg=None):
     parsed = linter.parse_rendered(rendered, recurse=True)
     if parsed.violations:
         if msg:
-            print(msg)
+            print(msg)  # pragma: no cover
         pytest.fail(parsed.violations[0].desc() + "\n" + parsed.tree.stringify())
     print(f"Parsed:\n {parsed.tree.stringify()}")
 
--- a/src/sqlfluff/testing/rules.py
+++ b/src/sqlfluff/testing/rules.py
@@ -202,9 +202,9 @@ def rules__test_helper(test_case):
                     test_case.rule,
                     test_case.fix_str,
                     configs=test_case.configs,
-                    msg="No violations_after_fix was provided, but the SQL after fix"
-                    "violates the rule. To accept a partial fix, "
-                    "violations_after_fix must be set.",
+                    msg="The SQL after fix is applied still contains rule violations. "
+                    "To accept a partial fix, violations_after_fix must be set "
+                    "listing the remaining, expected, violations.",
                 )
         else:
             # Check that tests without a fix_str do not apply any fixes.
--- a/src/sqlfluff/testing/rules.py
+++ b/src/sqlfluff/testing/rules.py
@@ -191,6 +191,11 @@ def rules__test_helper(test_case):
         # If a `fixed` value is provided then check it matches
         if test_case.fix_str:
             assert res == test_case.fix_str
+            assert_rule_pass_in_sql(
+                test_case.rule,
+                test_case.fix_str,
+                configs=test_case.configs,
+            )
             if test_case.violations_after_fix:
                 assert_violations_after_fix(test_case)
         else:
--- a/src/sqlfluff/testing/rules.py
+++ b/src/sqlfluff/testing/rules.py
@@ -91,7 +91,7 @@ def assert_rule_fail_in_sql(code, sql, configs=None, line_numbers=None):
     return fixed, linted.violations
 
 
-def assert_rule_pass_in_sql(code, sql, configs=None):
+def assert_rule_pass_in_sql(code, sql, configs=None, msg=None):
     """Assert that a given rule doesn't fail on the given sql."""
     # Configs allows overrides if we want to use them.
     if configs is None:
@@ -105,6 +105,8 @@ def assert_rule_pass_in_sql(code, sql, configs=None):
     rendered = linter.render_string(sql, fname="<STR>", config=cfg, encoding="utf-8")
     parsed = linter.parse_rendered(rendered, recurse=True)
     if parsed.violations:
+        if msg:
+            print(msg)
         pytest.fail(parsed.violations[0].desc() + "\n" + parsed.tree.stringify())
     print(f"Parsed:\n {parsed.tree.stringify()}")
 
@@ -116,6 +118,8 @@ def assert_rule_pass_in_sql(code, sql, configs=None):
     lerrs = lint_result.violations
     print(f"Errors Found: {lerrs}")
     if any(v.rule.code == code for v in lerrs):
+        if msg:
+            print(msg)
         pytest.fail(f"Found {code} failures in query which should pass.", pytrace=False)
 
 
@@ -190,14 +194,18 @@ def rules__test_helper(test_case):
             assert_violations_before_fix(test_case, violations_before_fix)
         # If a `fixed` value is provided then check it matches
         if test_case.fix_str:
-            assert res == test_case.fix_str, "result of fix doesn't match the fix_str"
-            assert_rule_pass_in_sql(
-                test_case.rule,
-                test_case.fix_str,
-                configs=test_case.configs,
-            )
+            assert res == test_case.fix_str
             if test_case.violations_after_fix:
                 assert_violations_after_fix(test_case)
+            else:
+                assert_rule_pass_in_sql(
+                    test_case.rule,
+                    test_case.fix_str,
+                    configs=test_case.configs,
+                    msg="No violations_after_fix was provided, but the SQL after fix"
+                    "violates the rule. To accept a partial fix, "
+                    "violations_after_fix must be set.",
+                )
         else:
             # Check that tests without a fix_str do not apply any fixes.
             assert res == test_case.fail_str, (
--- a/src/sqlfluff/testing/rules.py
+++ b/src/sqlfluff/testing/rules.py
@@ -119,7 +119,7 @@ def assert_rule_pass_in_sql(code, sql, configs=None, msg=None):
     print(f"Errors Found: {lerrs}")
     if any(v.rule.code == code for v in lerrs):
         if msg:
-            print(msg)
+            print(msg)  # pragma: no cover
         pytest.fail(f"Found {code} failures in query which should pass.", pytrace=False)
 
 
--- a/test/fixtures/rules/std_rule_cases/L003.yml
+++ b/test/fixtures/rules/std_rule_cases/L003.yml
@@ -323,6 +323,7 @@ test_pass_indent_indent_bigquery:
 
 
 test_jinja_indent_templated_table_name_a:
+  # TODO partial fix - this requires some more attention
   fail_str: |
     -- This file combines product data from individual brands into a staging table
     {% for product in ['table1', 'table2'] %}
@@ -353,6 +354,7 @@ test_jinja_indent_templated_table_name_a:
 # Like test_jinja_indent_1_a but "FROM" table not initially
 # indented.
 test_jinja_indent_templated_table_name_b:
+  # TODO partial fix - this requires some more attention
   fail_str: |
     -- This file combines product data from individual brands into a staging table
     {% for product in ['table1', 'table2'] %}
--- a/test/fixtures/rules/std_rule_cases/L003.yml
+++ b/test/fixtures/rules/std_rule_cases/L003.yml
@@ -324,6 +324,9 @@ test_pass_indent_indent_bigquery:
 
 test_jinja_indent_templated_table_name_a:
   # TODO partial fix - this requires some more attention
+  # Errors Found: [<SQLLintError: rule L003 pos:(11, 1), #fixes: 0, description:
+  #  Indentation not consistent with line #9>, <SQLLintError: rule L003 pos:(11, 1),
+  #  #fixes: 0, description: Indentation not consistent with line #19>]
   fail_str: |
     -- This file combines product data from individual brands into a staging table
     {% for product in ['table1', 'table2'] %}
@@ -335,7 +338,7 @@ test_jinja_indent_templated_table_name_a:
       id
     FROM
       {{ product }}
-    {% if not loop.last -%} UNION ALL {%- endif %}
+    {% if not loop.last -%} UNION ALL {%- endif %} -- noqa: L003
     {% endfor %}
   fix_str: |
     -- This file combines product data from individual brands into a staging table
@@ -355,6 +358,9 @@ test_jinja_indent_templated_table_name_a:
 # indented.
 test_jinja_indent_templated_table_name_b:
   # TODO partial fix - this requires some more attention
+  # Errors Found: [<SQLLintError: rule L003 pos:(11, 1), #fixes: 0, description:
+  #   Indentation not consistent with line #9>, <SQLLintError: rule L003 pos:(11, 1),
+  #   #fixes: 0, description: Indentation not consistent with line #19>]
   fail_str: |
     -- This file combines product data from individual brands into a staging table
     {% for product in ['table1', 'table2'] %}
@@ -366,7 +372,7 @@ test_jinja_indent_templated_table_name_b:
       id
     FROM
     {{ product }}
-    {% if not loop.last -%} UNION ALL {%- endif %}
+    {% if not loop.last -%} UNION ALL {%- endif %} -- noqa: L003
     {% endfor %}
   fix_str: |
     -- This file combines product data from individual brands into a staging table
--- a/test/fixtures/rules/std_rule_cases/L003.yml
+++ b/test/fixtures/rules/std_rule_cases/L003.yml
@@ -323,10 +323,6 @@ test_pass_indent_indent_bigquery:
 
 
 test_jinja_indent_templated_table_name_a:
-  # TODO partial fix - this requires some more attention
-  # Errors Found: [<SQLLintError: rule L003 pos:(11, 1), #fixes: 0, description:
-  #  Indentation not consistent with line #9>, <SQLLintError: rule L003 pos:(11, 1),
-  #  #fixes: 0, description: Indentation not consistent with line #19>]
   fail_str: |
     -- This file combines product data from individual brands into a staging table
     {% for product in ['table1', 'table2'] %}
@@ -338,7 +334,7 @@ test_jinja_indent_templated_table_name_a:
       id
     FROM
       {{ product }}
-    {% if not loop.last -%} UNION ALL {%- endif %} -- noqa: L003
+    {% if not loop.last -%} UNION ALL {%- endif %}
     {% endfor %}
   fix_str: |
     -- This file combines product data from individual brands into a staging table
@@ -357,10 +353,6 @@ test_jinja_indent_templated_table_name_a:
 # Like test_jinja_indent_1_a but "FROM" table not initially
 # indented.
 test_jinja_indent_templated_table_name_b:
-  # TODO partial fix - this requires some more attention
-  # Errors Found: [<SQLLintError: rule L003 pos:(11, 1), #fixes: 0, description:
-  #   Indentation not consistent with line #9>, <SQLLintError: rule L003 pos:(11, 1),
-  #   #fixes: 0, description: Indentation not consistent with line #19>]
   fail_str: |
     -- This file combines product data from individual brands into a staging table
     {% for product in ['table1', 'table2'] %}
@@ -372,7 +364,7 @@ test_jinja_indent_templated_table_name_b:
       id
     FROM
     {{ product }}
-    {% if not loop.last -%} UNION ALL {%- endif %} -- noqa: L003
+    {% if not loop.last -%} UNION ALL {%- endif %}
     {% endfor %}
   fix_str: |
     -- This file combines product data from individual brands into a staging table
--- a/test/fixtures/rules/std_rule_cases/L004.yml
+++ b/test/fixtures/rules/std_rule_cases/L004.yml
@@ -81,7 +81,17 @@ indented_comments:
 
 indented_comments_default_config:
   # TODO partial fix - this requires some more attention
-  #  Not visible to bare eye but here tab indent is replaced with spaces
+  # This is because of TODO fix indents after text:
+  # https://github.com/sqlfluff/sqlfluff/pull/590#issuecomment-739484190
+  # Lint fails on the tabs that come before the comments (expects spaces)
+  #  Errors Found: [
+  #    <SQLLintError: rule L004 pos:(2, 7), #fixes: 0, description:
+  #      Incorrect indentation type found in file. The indent occurs after other text,
+  #      so a manual fix is needed.>,
+  #    <SQLLintError: rule L004 pos:(3, 15), #fixes: 0,
+  #      description: Incorrect indentation type found in file. The indent occurs after
+  #      other text, so a manual fix is needed.>]
+  # Not visible to bare eye but here tab indent is replaced with spaces
   fail_str: |
     SELECT
     	a,			-- Some comment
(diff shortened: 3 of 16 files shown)


barrywhart:
I like the idea, but some tests are failing, I assume because the new check has found some bugs in the rules. Can you look into those? Let us know what you find. If any of those bugs are tricky, we might consider creating separate GitHub issues for those, and fixing those first, before merging this PR.

juhoautio:
> some tests are failing
> Can you look into those? Let us know what you find.

I wrote about my first findings in the PR description ("Failing test cases when introducing this check"). Any comment on how to proceed with those 2 cases, for example?

I'm going to check all the different failures next.

> we might consider creating separate GitHub issues for those, and fixing those first, before merging this PR.

That should be the plan for cases that need some actual fixes (if there are any cases like that). The ones where the failure should be ignored are better to ignore in this PR (because may need to add a new accompanying mechanism to support ignoring).

barrywhart:
I'll try and look tomorrow. The big 🏈 game is starting soon and will keep me busy for the next few hours. 😁

tunetheweb:
I think in these chases the `fix_str` should be removed.

Bonus points if we could check where there is a `fail_str` but no `fix_str` then we should confirm that `fail_str` == final output. I.e. we implicitly check that those tests without a `fail_str` do no apply ANY fixes.

barrywhart:
I agree with @tunetheweb: If the rule is identifying an error but not fixing it, the test case should omit the `fix_str`, and the test framework should verify that no changes (i.e. fixes) were applied to the input `fail_str`.

juhoautio:
Thanks for the comments. Just letting you know that I'm working on it. I may open some small new PRs to fix individual cases first.

barrywhart:
>Regarding the cases with only fail_str remaining, how are they useful? To check that query is not modified at all.. but why does that matter?

These tests are useful because they verify that the rule finds issues. Many of the SQLFluff rules don't fix things; they simply tell the user there's a problem and expect them to fix it manually.

>Regarding the cases where there's a partial fix (TODO comments in the diff of this PR), what should I do about them? Isn't it accepted that sometimes a fix can be partial?

I'll look at these. I'm not aware of any rules that identify issues and only partly fix them, but maybe?

juhoautio:
> I'm not aware of any rules that identify issues and only partly fix them

Those cases could for example be such that running fix again settles the remaining violation(s).

test/fixtures/rules/std_rule_cases/L004.yml
```diff
@@ -80,6 +80,8 @@ indented_comments:
     FROM spam
 
 indented_comments_default_config:
+  # TODO partial fix - this requires some more attention
+  #  Not visible to bare eye but here tab indent is replaced with spaces
```
barrywhart, line 84: I tried to fix this manually from the command line with the following command (which generates more verbose linter output):
```
sqlfluff fix test.sql --fixed-suffix FIXED --force --rules L004 -vvvv
```
The following messages were printed, so this is a legitimate case where the rule **finds** an issue but cannot **fix** the issue. We can remove the `fix_str` from this test case.
```
DEBUG      [L004] !! Violation Found: 'Incorrect indentation type found in file. The indent occurs after other text, so a manual fix is needed.'                                                        
DEBUG      [L004] !! Violation Found: 'Incorrect indentation type found in file. The indent occurs after other text, so a manual fix is needed.'                                                        
```

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -323,6 +323,7 @@ test_pass_indent_indent_bigquery:
 
 
 test_jinja_indent_templated_table_name_a:
+  # TODO partial fix - this requires some more attention
```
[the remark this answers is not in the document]
juhoautio: I couldn't get that working

test/fixtures/rules/std_rule_cases/L004.yml
juhoautio: Is it ok that fix is partial?

If I remove the fix_str here then this new assertion will fail:
```python
            # check that tests without a fail_str do no apply ANY fixes
             assert res == test_case.fail_str, "fix should not have modified the query"
```

Also, if we don't assert the result with fix_str any more, test coverage goes down.

I proposed instead adding a new test param to cover these cases, for example:
```yaml
  fail_after_fix: True
```
That way we can assert that the modifications after fix are as expected AND that it's still able to catch the remaining unfixable violation.

tunetheweb: I think it's OK that the fix is partial. It's not ideal but at least the test confirms that's what happens and it's what;'s expected.

I presume it consistently "fixes" to the same SQL after `fix` is run from the original SQL.

test/fixtures/rules/std_rule_cases/L003.yml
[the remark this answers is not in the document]
barrywhart: Please go ahead and commit the change. I can help look into it. These YAML tests sometimes behave strangely because they don't run precisely the normal linter code. In the long run, it'd be cool to clean that up.

test/fixtures/rules/std_rule_cases/L004.yml
juhoautio: > I presume it consistently "fixes" to the same SQL after fix is run from the original SQL?

Yes. That is asserted by the existing code. What the existing code doesn't check is that the check still fails on the fix_str. I propose adding a new param to state that + code to assert it.

test/fixtures/rules/std_rule_cases/L003.yml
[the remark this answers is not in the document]
juhoautio: Actually I'm not entirely sure which line you mean by line 11. Do you mean this:

```
{{ product }} -- noqa: disable=L003
```

or this:

```
{% if not loop.last -%} UNION ALL {%- endif %} -- noqa: disable=L003 
```
?

I tried both but neither of them seemed to help.

test/fixtures/rules/std_rule_cases/L004.yml
barrywhart: My other PR (#2657) would not help in this case. The rule is not buggy -- it's behaving as expected. I reviewed a few of these TODOs today and found two cases:
* The rule finds an issue but does not generate a fix. This is fine!
* The rule finds an issue and generates a fix, but the core linter does not apply it. There _may_ be possible improvements to make here in the future, but it's working as designed for the moment (I don't want to try and change this now).

There are several potential paths forward:
* `fail_after_fix`: This could be a bool value, as proposed. Another option would be to list the specific violations. Should we consider that?
* Use `-- noqa` to suppress these messages. If we implement #2337, then this would effectively be the same as `fail_after_fix`.

Any thoughts on which of these is better? I think I prefer the second, because it leverages existing or future linter functionality, rather than making the SQLFluff **test framework** more complex.

juhoautio: Couldn't agree more – using noqa is a clever idea. However, does sqlfluff complain about redundant noqa comments? If not, then we would be still missing the assertion for catching the lint violation.

tunetheweb: I still prefer the (third?) option of going what did we have and confirming that’s the fix output. Even if a subsequent run might change it again.

To me that’s what happens currently and saves having to define and set new flags or noqa any code.

So the only changes needed are:
1. Confirm if we have a `fail_str` we should not have a `pass_str` nor a second `fail_str`. This is an easy mistake to make so would be good to capture this. A `fail_str` should either be on its own or have a single `fix_str`.
2. Similarly a `pass_str` should always be on its own.
3. If there is a `fail_str` but no `fix_str` then we should confirm the rule does not fix anything. If it does then it should be in the test in a `fix_str` IMHO and the fact it isn’t makes it an invalid test. As I say above this doesn’t have to be the “final fix” as would happen with multiple independent runs (though ideally it would be!) but should be deterministic of one run of the rule by itself.

I think those three things are valuable to have and would make the test suite more robust and less prone to errors that we see now. There’s value in that even if it’s not the full, deterministic, solution we’d all ideally like.

As I say, at some point in the future, we could take the next step and confirm the fix has settled and will not change on subsequent runs (I.e. is completely deterministic). But there’s a bit of work to be done first there before we can enforce that IMHO and not sure it’s worth adding flags to note these cases for now, while we work to get there, as would make the tests a bit more complex to understand and write for those that don’t know the full history here or what exactly those flags mean and why they are set on some tests and not others.

src/sqlfluff/testing/rules.py
```diff
@@ -141,4 +141,12 @@ def rules__test_helper(test_case):
...
-            assert res == test_case.fix_str
+            assert res == test_case.fix_str, "result of fix doesn't match the fix_str"
+            assert_rule_pass_in_sql(
+                test_case.rule,
+                test_case.fix_str,
+                configs=test_case.configs,
+            )
+        else:
+            # check that tests without a fix_str do no apply ANY fixes
+            assert res == test_case.fail_str, "fix should not have modified the query"
```
juhoautio, line 152: This assertion is straightforward. If a test fails on this assertion, it can be fixed by adding the `fix_str`. I could make a new PR first that only adds this assertion?

test/fixtures/rules/std_rule_cases/L004.yml
barrywhart: I'm happy with your suggestions, @tunetheweb: it's often good to split work into smaller pieces so we can make steady progress and avoid getting stuck for a long time on one issue. We can go ahead and create a new issue for the remaining work so we don't forget it.

@juhoautio: About your question:
>does sqlfluff complain about redundant noqa comments? If not, then we would be still missing the assertion for catching the lint violation.

It doesn't catch this now, but see issue #2337 mentioned earlier: it's a proposed enhancement for that exact change. Someone could start work on that whenever they have time. Are you interested in working on that? Once we implement that, I think we'd be ready to work on the **new** issue we were discussing, i.e. enforcing that the `fix_str` passes the linter.

src/sqlfluff/testing/rules.py
```diff
@@ -141,4 +141,12 @@ def rules__test_helper(test_case):
         )
         # If a `fixed` value is provided then check it matches
         if test_case.fix_str:
-            assert res == test_case.fix_str
+            assert res == test_case.fix_str, "result of fix doesn't match the fix_str"
+            assert_rule_pass_in_sql(
+                test_case.rule,
+                test_case.fix_str,
+                configs=test_case.configs,
+            )
```
juhoautio, line 149: This assertion was the original idea of this PR, but is tricky, because there are some tests where the fix result also fails against the rule. Need to find find a way to explicitly record the exceptional expectation for those cases.

test/fixtures/rules/std_rule_cases/L004.yml
juhoautio: I have some second thoughts about the generality of noqa for this purpose. Is it too theoretic or not, but if we want to ignore a remaining error on some row, we need to add the noqa comment already in the fail_str. If the remaining error is expected to happen on that same line, we won't be able to test that partial fix, because also the initial fix will be skipped due to the noqa marker on the line.

test/fixtures/rules/std_rule_cases/L003.yml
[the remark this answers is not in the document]
juhoautio: @barrywhart bumping this comment in case you missed it by accident ^

juhoautio:
How about this?

> Fixing the obvious cases in https://github.com/sqlfluff/sqlfluff/pull/2658

Could we get that our of the way first? :)

test/fixtures/rules/std_rule_cases/L003.yml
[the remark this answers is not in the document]
tunetheweb: `-- noqa: disable=L003` means disable L003 from this line and every subsequent line going forward (until either end of file or you see an `enable` equivalent).

That's probably overkill (you actually WANT the rule to run on all lines - except the troublesome ones we know we have an issue fixing).

`-- noqa: L003` means just disable L003 from this line, then carry on checking it again from future lines. I think that is the option your want.

test/fixtures/rules/std_rule_cases/L004.yml
barrywhart: >If the remaining error is expected to happen on that same line, we won't be able to test that partial fix

It's possible, but I think it will be unusual. If it does happen, in many cases, I think we could update the test case (possibly creating new ones) to test those errors separately.

I agree noqa probably isn't perfect, but it seems pretty close, so I'm in favor of trying it. If it doesn't work well, we can easily try something new, because the SQLFluff tests only affect developers, not users.

test/fixtures/rules/std_rule_cases/L003.yml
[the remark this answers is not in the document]
barrywhart: You're correct: I had the wrong `noqa` syntax!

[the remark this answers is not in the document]
barrywhart: @juhoautio: Please add it to this line: `{% if not loop.last -%} UNION ALL {%- endif %}`, then tag me once you push it. I will look and see why `noqa` is not working.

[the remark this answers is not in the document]
juhoautio: Pushed it https://github.com/sqlfluff/sqlfluff/pull/2624/commits/fe66a93e20a46d2b640e0b9524f94f1ff88774dc

test/fixtures/rules/std_rule_cases/L004.yml
```diff
@@ -80,6 +80,18 @@ indented_comments:
...
+  # https://github.com/sqlfluff/sqlfluff/pull/590#issuecomment-739484190
+  # Lint fails on the tabs that come before the comments (expects spaces)
+  #  Errors Found: [
+  #    <SQLLintError: rule L004 pos:(2, 7), #fixes: 0, description:
+  #      Incorrect indentation type found in file. The indent occurs after other text,
+  #      so a manual fix is needed.>,
+  #    <SQLLintError: rule L004 pos:(3, 15), #fixes: 0,
+  #      description: Incorrect indentation type found in file. The indent occurs after
+  #      other text, so a manual fix is needed.>]
+  # Not visible to bare eye but here tab indent is replaced with spaces
```
juhoautio, line 94: I had a look at this one. This seems to be a case that is consciously left for the user to fix by hand. It would be nice to be able to assert exactly this: that it applies some fixes but then lint fails with that particular error. This is exactly the type of case where ignoring with line level noqa won't cut it because the wrong kind of indent in the beginning of the line is expected to be fixed, but in the middle of the same line not.

Same for the other case in this file, just swapping space <-> tab.

juhoautio: I think in general it's definitely a good idea to start asserting that fix_str also always passes lint.

But this test would fail if that is enforced

I think the actual options are:

1. test these cases separately:
   - test with only fail_str with only unfixable content
   - test with fail_str & fix_str with only fixable content
2. extend the test framework to support testing both things in the same test case

With option 1 I feel like we're losing something because we can't test complicated situations where a bug would surface only when both fixable & unfixable sections are present in the source query.

src/sqlfluff/testing/rules.py
barrywhart, line 149: If you change it to this, it will use the "noqa" comments when checking the fix_str, but not when making the original fixes. 😎 
```suggestion
            linted = Linter(config=test_case.configs).lint_string(test_case.fix_str)
            assert not linted.get_violations()
```

test/fixtures/rules/std_rule_cases/L003.yml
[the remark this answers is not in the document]
barrywhart: See my comment elsewhere -- we need to change how `rules.py` is checking for remaining lint errors, after the fix.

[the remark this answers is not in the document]
barrywhart: If you prefer, I have a suggestion similar to your idea about `fail_after_fix`. I suggest we add a property called `failures_after_fix`, which specifically lists the failures remaining after fixes are applied, e.g.:
```
failures_after_fix:
- code: L003
  description: 'Indentation not consistent with line #9'
  line_no: 11
  line_pos: 1
- code: L003
  description: 'Indentation not consistent with line #19'
  line_no: 11
  line_pos: 1
```

I generated the YAML above with the following code:
```
print(yaml.dump([e.get_info_dict() for e in lerrs]))
```

Thoughts, @juhoautio and @tunetheweb?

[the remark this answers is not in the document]
juhoautio: Nice, that would satisfy the need for asserting errors that I had in mind. Looks good to me!

In the same fashion this could be an optional extension for the cases where there's only `fail_str`, if one wants to add stronger assertions. In that case it could mean for example adding:
```
failures:
- code: L003
  description: 'Indentation not consistent with line #9'
  line_no: 11
  line_pos: 1
```

> I generated the YAML above with the following code

Wow! Seems like it would be quite easy to also print that out in case the detected failures don't match with the expectation.

src/sqlfluff/testing/rules.py
juhoautio: I should probably take your word on it, but I can't see how that could work. The fixing seems to happen already before the code that you suggested to change, in `res = assert_rule_fail_in_sql(..`. I understood it so that noqa prevents any fixes to be done on the lines that have it.

But if I understood your idea correctly, that seems valid though: fix first so that noqa is ignored, then lint with noqa enabled.

But sorry, I still don't like it too much. I think it's not good to have a different meaning for noqa in tests than in runtime. I'm in favor of extending the testing framework.

I could imagine that addressing any complexity like this would be trivial if tests cases were written in python instead of using a custom yaml DSL. To me it seems that the only benefit of yaml is that it allows writing some things without surrounding quotes. But it makes it harder to run specific tests (especially in an IDE) and has limitations for what kind of things one is able to assert in the tests. But it would be nice to understand better what benefit the yaml approach brings, maybe I'm missing something.

test/fixtures/rules/std_rule_cases/L003.yml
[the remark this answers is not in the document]
juhoautio: @barrywhart @tunetheweb I can try to implement the strategy that's proposed in https://github.com/sqlfluff/sqlfluff/pull/2624/files#r807452356 if you'd want me to. On the other hand I'm wondering if test cases like this should be moved to be written in python. Could you suggest where I should place such tests? We could then even make the assertion error just say that hey, your test has still failures after some fixes happened, and in this type of case it's best to write the test in python because the yaml framework doesn't cover complicated cases like that. Thoughts?

src/sqlfluff/testing/rules.py
tunetheweb: Personally, as a non-Python developer, I like the simplicity of the current test suite.

We do have Python tests outside of the fixtures directory for more complex tests (e.g. testing for two rules together when they’ve caused issues in the past, or testing that specific error message messages are returned for those rules that alter their error messages).

But those are the rare exceptions and the YAML syntax to quickly be able to paste in SQL (rather than some string value containing that SQL) and simple configuration makes test cases very easy to add, and approachable for new users - especially those without Python skills.

test/fixtures/rules/std_rule_cases/L003.yml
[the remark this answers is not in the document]
barrywhart: I prefer to extend the testing framework. It seems like a simple enough extension, and it avoids potentially moving and translating tests from one format to another as the rule changes over time.

src/sqlfluff/testing/rules.py
barrywhart: I like the YAML tests because it's very easy to add them, very little "overhead" in terms of effort.

+1 for extending the testing framework to verify certain expected errors are present after fixing.

test/fixtures/rules/std_rule_cases/L003.yml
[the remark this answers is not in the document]
barrywhart: How many tests would we be moving to Python?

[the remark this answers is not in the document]
tunetheweb: FYI @juhoautio here’s where the Python test cases for rules live: https://github.com/sqlfluff/sqlfluff/tree/main/test/rules

src/sqlfluff/testing/rules.py
tunetheweb: We really should take a stab at this: https://github.com/sqlfluff/sqlfluff/issues/2180

And just realised that you raised this! 😁

juhoautio: > I like the YAML tests because it's very easy to add them, very little "overhead" in terms of effort.

Ok, I suppose I can appreciate that as well. But still, based on my experience on developing these tests I would have to disagree, because I haven't been able to find a nice way to execute specific rule yaml tests in my IDE. Being able to run & debug regular python pytests is way more useful for my productivity than the possibility to avoid some minimal clutter in the test case syntax. + I like the power of python & the freedom it gives for eg. documenting the test cases better.

> +1 for extending the testing framework to verify certain expected errors are present after fixing.

This is nice to hear, but I'm afraid that might take longer to accomplish, so I'm proposing to just migrate the few cases that remain to python based tests for now.

Here's the first one, I hope you're ok with this approach at least as an initial solution:
https://github.com/sqlfluff/sqlfluff/pull/2681

test/fixtures/rules/std_rule_cases/L003.yml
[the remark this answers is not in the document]
juhoautio: > How many tests would we be moving to Python?

It should be only the ones that I marked as TODO in the diff of this PR, so 4 affected files and just 8 test cases in total. My first PR covers 2 of them and I can work on the rest in the same fashion.

src/sqlfluff/testing/rules.py
juhoautio: Note to self: this assertion can be eventually enabled once all the existing cases that would fail on this have been migrated to assert the remaining violations explicitly as supported by https://github.com/sqlfluff/sqlfluff/pull/2718. 

The assertion to be added eventually is:

_if `violations_after_fix` is not given, `fix_str` must not have any remaining violations_

juhoautio:
Rebased etc. See the full PR diff: there isn't much to see any more :) I updated the PR description as well.

src/sqlfluff/testing/rules.py
```diff
@@ -193,6 +197,15 @@ def rules__test_helper(test_case):
...
             if test_case.violations_after_fix:
                 assert_violations_after_fix(test_case)
+            else:
+                assert_rule_pass_in_sql(
+                    test_case.rule,
+                    test_case.fix_str,
+                    configs=test_case.configs,
+                    msg="No violations_after_fix was provided, but the SQL after fix"
+                    "violates the rule. To accept a partial fix, "
+                    "violations_after_fix must be set.",
```
tunetheweb, line 207: This reads a little confusing. WDYT about this?:

```suggestion
                    msg="The SQL after fix is applied still contains rule violations. "
                    "To accept a partial fix, violations_after_fix must be set "
                    "listing the remaining, expected, violations.",
```

src/sqlfluff/testing/rules.py
```diff
@@ -116,6 +118,8 @@ def assert_rule_pass_in_sql(code, sql, configs=None):
     lerrs = lint_result.violations
     print(f"Errors Found: {lerrs}")
     if any(v.rule.code == code for v in lerrs):
+        if msg:
+            print(msg)
```
tunetheweb, line 122:
```suggestion
            print(msg)  # pragma: no cover
```

src/sqlfluff/testing/rules.py
```diff
@@ -105,6 +105,8 @@ def assert_rule_pass_in_sql(code, sql, configs=None):
     rendered = linter.render_string(sql, fname="<STR>", config=cfg, encoding="utf-8")
     parsed = linter.parse_rendered(rendered, recurse=True)
     if parsed.violations:
+        if msg:
+            print(msg)
```
tunetheweb, line 109:
```suggestion
            print(msg)  # pragma: no cover
```

tunetheweb (changes requested):

tunetheweb (approved):

PR #2628, opened 2022-02-13
Hitting the linter loop limit should be treated as an error

### LINKED ISSUE #1386: When linter loop limit is hit, should report an error/exception
When the loop limit is hit, `sqlfluff fix` logs a warning:
```
Loop limit on fixes reached
```

This is a pretty "quiet" notification that can mask issues. Suggest we change this:
* Increase the default loop limit from 10 to 25 iterations.
* Log an ERROR and/or raise an exception. If an exception, suggest (maybe?) raising one that is caught and included in the list of lint issues, not one that stops the `fix` run. Open to other ideas, but I think this is the right tradeoff between raising visibility of what is _likely_ an infinite loop in the linter run while still giving the user useful `fix` output.

Related to #1344, among others.

barrywhart:
Fixes #1386

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -484,7 +484,7 @@ class Linter:
             ignore_buff = []
 
         save_tree = tree
-        for loop in range(loop_limit):
+        for loop in range(loop_limit + 1):
             changed = False
 
             progress_bar_crawler = tqdm(
@@ -552,17 +552,21 @@ class Linter:
                 linter_logger.warning(f"Loop limit on fixes reached [{loop_limit}].")
 
                 # Discard any fixes for the linting errors, since they caused a
-                # loop. IMPORTANT: Unfixable linting errors also cause
-                # "sqlfluff fix" to exit with a "failure" exit code, which is
-                # the desired outcome in this situation.
+                # loop. IMPORTANT: By doing this, we are telling SQLFluff that
+                # these linting errors are "unfixable". This is important,
+                # because when "sqlfluff fix" encounters unfixable lint errors,
+                # it exits with a "failure" exit code, which is exactly what we
+                # want in this situation. (Reason: Although this is more of an
+                # internal SQLFluff issue, users deserve to know about it,
+                # because it means their file(s) weren't fixed.
                 for violation in initial_linting_errors:
                     if isinstance(violation, SQLLintError):
                         violation.fixes = []
 
-                # Return the original file. We do this because if the linter
-                # hit the loop limit, the file is probably messy, e.g. some of
-                # the fixes were applied repeatedly, possibly other strange
-                # things. We don't want the user to see this junk.
+                # Return the original parse tree, before any fixes were applied.
+                # Reason: When the linter hits the loop limit, the file is often
+                # messy, e.g. some of the fixes were applied repeatedly, possibly
+                # other weird things. We don't want the user to see this junk!
                 return save_tree, initial_linting_errors, ignore_buff
 
         if config.get("ignore_templated_areas", default=True):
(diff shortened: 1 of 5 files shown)


src/sqlfluff/core/linter/linter.py
```diff
@@ -542,8 +543,27 @@ def lint_fix_parsed(
                     "loops."
                 )
                 break
-        if fix and loop + 1 == loop_limit:
-            linter_logger.warning(f"Loop limit on fixes reached [{loop_limit}].")
+        else:
+            if fix:
```
barrywhart, line 547: The old `if` is considered a poor, hacky way to determine if the loop exited or not. I updated it to use `for/else` instead. Explanation: https://book.pythontips.com/en/latest/for_-_else.html#else-clause

src/sqlfluff/core/rules/base.py
```diff
@@ -555,39 +555,28 @@ def crawl(
...
-            if lerr:
-                if ignore_mask:
-                    filtered = LintedFile.ignore_masked_violations([lerr], ignore_mask)
-                    if not filtered:
-                        lerr = None
-                        ignored = True
-            if lerr:
-                new_lerrs.append(lerr)
-            if not ignored:
-                new_fixes.extend(res.fixes)
```
barrywhart, line 574: I moved the nested function `_process_lint_result()` to a separate method on `RuleBase` so the automated test can use it.

src/sqlfluff/core/linter/linter.py
```diff
@@ -483,7 +483,8 @@ def lint_fix_parsed(
         else:
             ignore_buff = []
 
-        for loop in range(loop_limit):
+        save_tree = tree
+        for loop in range(loop_limit + 1):
```
barrywhart, line 487: When I changed to a `for/else` loop (see the comment below for more info), I had to adjust the loop range in order to keep the same behavior -- one of the existing tests surfaced this subtle issue.

test/fixtures/rules/std_rule_cases/L019.yml
```diff
@@ -230,10 +230,10 @@ trailing_comma_move_past_several_comment_lines:
     global_actions_states
   configs:
     core:
-      # Set runaway_limit=1 to verify the fix only requires one pass. In an
+      # Set runaway_limit=2 to verify the fix only requires one pass. In an
       # earlier version, the comma before "SAFE_DIVIDE()" was being moved one
       # line per pass. Too lazy!
-      runaway_limit: 1
+      runaway_limit: 2
```
barrywhart, line 236: Now that we treat hitting the limit as an error, these tests need one additional loop (First iteration generates a fix, then the second iteration generates **no** fixes and the loop exits). One iteration would've not only caused the linter to report an error, but also to discard the fixes!)

barrywhart:
@juhoautio: Given there is work remaining to do on #2624 (i.e. rules that are currently causing the additional lint check to fail), I think your suggestion about the linter loop should be a separate issue after all. It's a great idea, but we shouldn't add it while there are still known issues.

src/sqlfluff/core/linter/linter.py
```diff
@@ -542,8 +543,31 @@ def lint_fix_parsed(
...
+                # because it means their file(s) weren't fixed.
+                for violation in initial_linting_errors:
+                    if isinstance(violation, SQLLintError):
+                        violation.fixes = []
+
+                # Return the original parse tree, before any fixes were applied.
+                # Reason: When the linter hits the loop limit, the file is often
+                # messy, e.g. some of the fixes were applied repeatedly, possibly
+                # other weird things. We don't want the user to see this junk!
+                return save_tree, initial_linting_errors, ignore_buff
```
alanmcruickshank, line 570: I like this approach 👍

alanmcruickshank (approved):
This looks good. Thanks @barrywhart

PR #2637, opened 2022-02-14
Draft release of 0.10.1

alanmcruickshank:
Draft release of 0.10.1


tunetheweb (approved):
LGTM - assuming you'll do a final sync up with latest merges before merging.

barrywhart (approved):

tunetheweb (changes requested):
This should not be merged until #2652 is merged as we've introduced a bug since 0.10.0 that we need to fix before releasing 0.10.1.

barrywhart (approved):
Looks good to me -- ready for release!

PR #2642, opened 2022-02-14
L048 - handle more statements and exclude casting operators

tunetheweb:

Fixes issue reported in Slack.

Nope

### Pull Request checklist

  - Other.


tunetheweb:
> Looks good for the ANSI case. Teradata and BigQuery both have custom versions of `cast_expression` with syntax that is quite different, and potentially should not allow the "exception" we're adding here. Thoughts? Should we add tests for those dialects? Possibly tweak the rule?

Thanks for spotting and I looked into this, which lead me into discovering that that L048 didn't work for a lot of cases - including the example case in the docs! And never has.

This is because it is based on L006 which checks for spaces in operators (e.g. `a + b`) so explicitly ignores segments with <= 2 children. However we need them for L046 (e.g. `SELECT '1343'AS col` or `SELECT DATE '2022-04-15'`).

So fixed that, and made my rule addition more restrictive to only allow casting expressions with a casting operator.

So think it's good for another review now.

src/sqlfluff/rules/L006.py
```diff
@@ -162,7 +167,7 @@ def _eval(self, context: RuleContext) -> EvalResultType:
                         LintResult(
                             anchor=before_anchor,
                             description="Missing whitespace before {}".format(
-                                before_anchor.raw[:10]
+                                before_anchor.raw
```
tunetheweb, line 170: I dunno why this was restricted before and causes incorrect messages to be printed.

For example for `SELECT DATE'2021-02-02'` this is printed:

```
== [test.sql] FAIL                                                                                                                  
L:   1 | P:  12 | L048 | Missing whitespace before '2021-02-0
All Finished 📜 🎉!
```

src/sqlfluff/rules/L006.py
```diff
@@ -189,7 +194,7 @@ def _eval(self, context: RuleContext) -> EvalResultType:
                         LintResult(
                             anchor=after_anchor,
                             description="Missing whitespace after {}".format(
-                                after_anchor.raw[-10:]
+                                after_anchor.raw
```
barrywhart, line 197: Was this an intentional change? Not sure why it was written the old way, but maybe there was some worry about the string being super long?

test/fixtures/rules/std_rule_cases/L048.yml
```diff
@@ -1,19 +1,29 @@
...
   # Test that brackets don't trigger it
   pass_str: "SELECT ('foo' || 'bar') as buzz"
 
-test_fail_simple:
+test_fail_as:
+  fail_str: |
+    SELECT
+        'foo'AS bar
+    FROM foo
+  pass_str: |
```
barrywhart, line 15: Should this be `fix_str`?

barrywhart (changes requested):

src/sqlfluff/rules/L006.py
tunetheweb: Yes. Same as above. Not sure why restricted, but not restricted anywhere else. Guess it could lead to very long output but still think clearer.

test/fixtures/rules/std_rule_cases/L048.yml
tunetheweb: Good catch! Fixed. Hopefully the enhancements to #2624 will catch these in future (FYI @juhoautio )

src/sqlfluff/rules/L006.py
barrywhart: I guess this **is** the lint result message, so it's user facing, unlike most strings in SQLFluff. If we were going to chop it, would probably be good to add "..." when it's too long, so you know something was omitted.

barrywhart (approved):
Nice work! Picking up code that never worked can be 🤯.

src/sqlfluff/rules/L006.py
tunetheweb: Indeed. But don't chop elsewhere (and probably should create a common function for that if we want to). I say leave as unchopped for now.

src/sqlfluff/rules/L006.py
tunetheweb: Yeah but not chopped in other rules. Probably should create a common function in BaseRule to do a nice chop if we wanted to. But out of scope for this change IMHO.

barrywhart: I agree. I was just blathering. 🤣

PR #2657, opened 2022-02-15
Core linter enhancement: Check for successful parse after applying fixes

### LINKED ISSUE #1304: Running indentation fixes adds whitespace to various code segments

barrywhart:
When it finds an issue, it logs a warning message to the terminal. Example:

image

Fixes #1304 

Adds extra runtime checks on all the fixes applied. If an issue is detected, a runtime error is raised. Several existing rules had issues detected by these checks, so they were updated. Tests are passing, but there's some risk of new, latent issues in those rules or in other rules, since the rules all operate against the same parse structure. This happened already: see rule L039, which began failing and had to be updated, not because of the parse check itself, but because of changes in how L003 behaves.

### Pull Request checklist

  - Other.


src/sqlfluff/core/rules/base.py
```diff
@@ -768,36 +770,6 @@ def discard_unsafe_fixes(
...
-        #   in which case the loop body contains template variables.
-        for block_id in fix_block_ids:
-            if block_id in block_info.literal_only_loops:
-                linter_logger.info(
-                    "      * Discarding fixes to literal-only loop: %s",
-                    lint_result.fixes,
-                )
-                lint_result.fixes = []
-                return
-
```
barrywhart, line 800: Both these checks are now obsolete. These were somewhat hacky safety mechanisms that were required when we were using the old heuristic slicer, which returned approximate results that often had errors. Now that `JinjaTracer` has replaced it, slice information is 100% accurate, and the linter has more robust mechanisms to avoid issues with fixes.

src/sqlfluff/rules/L039.py
```diff
@@ -75,18 +77,14 @@ def _eval(self, context: RuleContext) -> Optional[List[LintResult]]:
...
-                                    fixes=[LintFix.delete(child_seg)],
-                                )
+                        violations.append(
+                            LintResult(
+                                anchor=child_seg,
+                                fixes=[LintFix.delete(child_seg)],
                             )
-                    else:
-                        leading_whitespace = False
+                        )
```
barrywhart, line 87: The `leading_whitespace` logic was a workaround for issue #1304. This PR fixes that issue, so this code is no longer needed.

test/core/rules/config_test.py
```diff
@@ -26,10 +26,14 @@ class Rule_T001(BaseRule):
...
-        if context.segment.is_type("star"):
+        if context.segment.is_type("whitespace"):
             return LintResult(
                 anchor=context.segment,
-                fixes=[LintFix.create_before(context.segment, [NewlineSegment()])],
+                fixes=[
+                    LintFix.replace(
+                        context.segment, [WhitespaceSegment(context.segment.raw + " ")]
+                    )
+                ],
```
barrywhart, line 36: I had to update this dummy rule, because the old version corrupted the parse tree, which we now catch and report.

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -347,7 +347,7 @@ test_jinja_indent_templated_table_name_a:
             id
         FROM
             {{ product }}
-    {% if not loop.last -%} UNION ALL {%- endif %}
+        {% if not loop.last -%} UNION ALL {%- endif %}
```
barrywhart, line 350: The linter now returns nicer results for this test now that we removed the obsolete code from `discard_unsafe_fixes()`. Previously, the rules were generating these results, but some of the changes were discarded for "safety reasons".

src/sqlfluff/rules/L036.py
```diff
@@ -74,8 +74,7 @@ def _eval(self, context: RuleContext):
             if len(select_targets_info.select_targets) == 1:
                 return self._eval_single_select_target_element(
                     select_targets_info,
-                    context.functional.segment,
-                    context.parent_stack,
+                    context,
```
barrywhart, line 77: Rule L036 required heavy rework because it was previously leaving the parse tree in a pretty bad state. It was leaving trailing whitespace and comment child segments after deleting or moving other segments. To make the parser happy, these segments had to be relocated. The new `_fixes_for_move_after_select_clause()` function assists with this.

src/sqlfluff/core/templaters/base.py
```diff
@@ -180,39 +177,6 @@ def _find_slice_indices_of_templated_pos(
             raise ValueError("Position Not Found")
         return first_idx, last_idx
 
-    @cached_property
-    def raw_slice_block_info(self) -> RawSliceBlockInfo:
```
barrywhart, line 184: This function was only used by the code I removed from `discard_unsafe_fixes()`.

barrywhart:
@tunetheweb: I updated the description a bit. TBH, haven't thought about how to present these errors. From your perspective, what would be the desired behavior? Some possible options:
* Log it as a WARNING or CRITICAL and continue.
* Fail that file.
* Fail the run.

tunetheweb:
Maybe we should go with this initially?:

> Log it as a WARNING or CRITICAL and continue.

Just concerned that we don't know how prevent this is? Don't want to merge this and suddenly lots of people's runs fails when they didn't before. Of course some of those failures may be causing corruption, but many may not actually be causing an issue.

barrywhart:
@tunetheweb: I agree with that choice. It's not a "house on fire" situation if it happens, it's more that a rule is doing something strange (like the known L003 issue), which can cause other rules to get confused. I'm thinking we start as a warning, and we can ramp up to critical later, maybe even fall that file. The point is, do it gradually so it's not widely disruptive.

I think of this change as being similar to a check constraint in a database: prevent corruption to the shared data (in this case, the parse tree), so we know every rule works correctly, and we're not sloppily using one rule to "accidentally" test another.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1074,11 +1079,66 @@ def apply_fixes(self, fixes):
...
+                                "failed the match() parser check. "
+                                f"Result: {match_result!r} "
+                                f"Fixes: {fixes_applied!r}"
+                            )
+                if getattr(segment, "parse_grammar", None):
+                    try:
+                        # :HACK: Calling parse() corrupts the segment 'r'
+                        # in some cases, e.g. adding additional Dedent child
+                        # segments. Here, we work around this by calling
+                        # parse() on a "backup copy" of the segment.
```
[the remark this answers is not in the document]
barrywhart: It's _probably_ not an issue in normal circumstances -- during normal parsing, all the segments are new. In this case, we're parsing some segments and not others. Whatever the corruption is, I think it's more likely that the parser wasn't intended to be used this way. I'm okay with the current approach unless we encounter issues with it later; copying objects is fairly cheap and easy. 

@alanmcruickshank, any thoughts on this, as someone more familiar with the parser?

barrywhart:
@tunetheweb: I updated the behavior -- now it will log warnings rather than throwing errors.

tunetheweb (approved):
A good bit beyond my understanding so would be good to get other's opinions but can't spot anything obviously wrong.

Good work!

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1074,11 +1079,70 @@ def apply_fixes(self, fixes):
...
+                        seg.pos_marker = replace(
+                            seg.pos_marker,
+                            templated_file=self.pos_marker.templated_file,
+                        )
+                    match_result = segment.match(
+                        deepcopy(
+                            tuple([seg for seg in segment.segments if not seg.is_meta])
+                        ),
+                        parse_context,
+                    )
```
OTooleMichael, line 1106: Would it be possible insert / replace new segments as a special type/Class.

`FixRawSegement(raw="my_whole.new_unlexed AND unparsed_segement")`

And also run a quick Lex on these small segments? around here. 
Giving us a simpler and more correct API for edits.

barrywhart: @OTooleMichael: That's definitely something we can look at. Can you create a GitHub enhancement issue for this?

PR #2658, opened 2022-02-15
Fix mistakes in rule yaml tests

juhoautio:

Fixes some common mistakes in rule yaml tests.

Covers these type of mistakes:
- Successful fix, but `fix_str` missing to assert the result -> add the missing `fix_str`
- `pass_str` used instead of `fix_str`, ie. fix result would not be asserted -> change `pass_str` to `fix_str`
- Unfixable failure, but `fix_str` is used with the same value as `fail_str`. Doesn't make sense to do that -> remove `fix_str`.

These mistakes were revealed by the pending additional assertions to testing/rules.py in https://github.com/sqlfluff/sqlfluff/pull/2624/files

No.

But note that a yaml rule test with only `fail_str` should always mean that there's an unfixable error and the query is not modified as a result. Or at least how I understood, but there's an open question about it: https://github.com/sqlfluff/sqlfluff/pull/2624#issuecomment-1040524243. There are still some cases in the repo where there's a partial fix ie. the result after fix is modified query + raising an error. Those are not yet fixed in this PR, I plan to deal with those separately. 

### Pull Request checklist

  - Other.

### DIFF
--- a/test/fixtures/rules/std_rule_cases/L002.yml
+++ b/test/fixtures/rules/std_rule_cases/L002.yml
@@ -2,3 +2,4 @@ rule: L002
 
 test_fail_mixed_tabs_and_spaces:
   fail_str: "    \t    \t    SELECT 1"
+  fix_str: "                    SELECT 1"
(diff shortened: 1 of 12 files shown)


tunetheweb (approved):
LGTM. Nice clean up and catches.

PR #2663, opened 2022-02-15
Assert that fix_str is set

juhoautio:

Assert in yaml rule tests that if only `fail_str` is given, then the rule violation is detected but original query is not modified at all.

If this new check fails but some fixing is expected, then the test case can be passed by adding the missing `fix_str`.

So this just helps in keeping the test coverage better for fixing. 

This doesn't find any issues in the existing tests now, because those were just fixed in https://github.com/sqlfluff/sqlfluff/pull/2658.

No.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/testing/rules.py
+++ b/src/sqlfluff/testing/rules.py
@@ -144,5 +144,7 @@ def rules__test_helper(test_case):
             assert res == test_case.fix_str
         else:
             # Check that tests without a fix_str do not apply any fixes.
-            # If the fix is expected, add the missing fix_str in the test case.
-            assert res == test_case.fail_str
+            assert res == test_case.fail_str, (
+                "No fix_str was provided, but the rule modified the SQL. "
+                "Where a fix can be applied by a rule, a fix_str must be supplied in the test."
+            )
--- a/src/sqlfluff/testing/rules.py
+++ b/src/sqlfluff/testing/rules.py
@@ -145,6 +145,6 @@ def rules__test_helper(test_case):
         else:
             # Check that tests without a fix_str do not apply any fixes.
             assert res == test_case.fail_str, (
-                "No fix_str was provided, but the rule modified the SQL. "
-                "Where a fix can be applied by a rule, a fix_str must be supplied in the test."
+                "No fix_str was provided, but the rule modified the SQL. Where a fix "
+                "can be applied by a rule, a fix_str must be supplied in the test."
             )
--- a/src/sqlfluff/testing/rules.py
+++ b/src/sqlfluff/testing/rules.py
@@ -142,3 +142,7 @@ def rules__test_helper(test_case):
         # If a `fixed` value is provided then check it matches
         if test_case.fix_str:
             assert res == test_case.fix_str
+        else:
+            # Check that tests without a fix_str do not apply any fixes.
+            # If the fix is expected, add the missing fix_str in the test case.
+            assert res == test_case.fail_str


src/sqlfluff/testing/rules.py
```diff
@@ -142,3 +142,7 @@ def rules__test_helper(test_case):
         # If a `fixed` value is provided then check it matches
         if test_case.fix_str:
             assert res == test_case.fix_str
+        else:
+            # Check that tests without a fix_str do not apply any fixes.
+            # If the fix is expected, add the missing fix_str in the test case.
+            assert res == test_case.fail_str
```
tunetheweb, line 148: I wonder if we need an assertion error here to make it more obvious why we are asserting this?

Something like “Please supply a fix_str if the rule can apply a fix for this test”. Or something like that?

barrywhart: Or "No fix_str was provided, but the rule modified the SQL. Please add a fix_str to the test."

juhoautio, line 148: Indeed. How about this (info is in assert msg instead of comments):
```suggestion
            # check that tests without a fix_str do no apply ANY fixes
            assert res == test_case.fail_str, (
                "No fix_str was provided, but the rule modified the SQL. "
                "If the modification is expected, please add a fix_str to the test."
            )
```
I added "If the modification is expected" ^

tunetheweb: To be honest that reads a little off to me. I think it's better without it. Also noticed a slight typo in comment "do not" instead of "do no".

```suggestion
            # Check that tests without a fix_str do not apply any fixes.
            assert res == test_case.fail_str, (
                "No fix_str was provided, but the rule modified the SQL. "
                "Please add a fix_str to the test."
            )
```

juhoautio: I'm just hesitating to instruct simply to add a fix_str. Do you think it's so obvious/implicit that anyone knows that fix_str should be added only if the fix result is as desired? Maybe it is.

tunetheweb: Altered by suggestion above. WDYT?

juhoautio: I committed the suggestion by @tunetheweb in https://github.com/sqlfluff/sqlfluff/pull/2663/commits/311480392f4e99f0d1542d8d93d4a0389176b6f2

tunetheweb (approved):
LGTM! Thanks for working on this. Small actual code change but think it will help a lot to keep tests valid going forward.

src/sqlfluff/testing/rules.py
```diff
@@ -142,3 +142,9 @@ def rules__test_helper(test_case):
         # If a `fixed` value is provided then check it matches
         if test_case.fix_str:
             assert res == test_case.fix_str
+        else:
+            # Check that tests without a fix_str do not apply any fixes.
+            assert res == test_case.fail_str, (
+                "No fix_str was provided, but the rule modified the SQL. "
+                "Where a fix can be applied by a rule, a fix_str must be supplied in the test."
```
tunetheweb, line 149:
```suggestion
                "No fix_str was provided, but the rule modified the SQL. Where a fix "
                "can be applied by a rule, a fix_str must be supplied in the test."
```

tunetheweb (changes requested):
Oops. Looks like my suggested wording needs a little `black` love.

tunetheweb (approved):

Issue #2666, opened 2022-02-16
Parser fails when using additional clauses in MERGE conditions 


The-Loud:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Using the `MERGE` command with snowflake will fail to parse if a `WHEN NOT MATCHED AND` clause exists.

The error
```
==== parsing violations ====
L:  90 | P:  38 |  PRS | Line 90, Position 38: Found unparsable section: '\n\nWHEN NOT
                       | matched\n AND S.change_type...'

```
The lines that failed

```
[L: 90, P: 38]      |                        [META] dedent:
[L: 90, P: 38]      |                unparsable:                                   !! Expected: 'Nothing...'
[L: 90, P: 38]      |                    newline:                                  '\n'
[L: 91, P:  1]      |                    newline:                                  '\n'
[L: 92, P:  1]      |                    raw:                                      'WHEN'
[L: 92, P:  5]      |                    whitespace:                               ' '
[L: 92, P:  6]      |                    raw:                                      'NOT'
[L: 92, P:  9]      |                    whitespace:                               ' '
[L: 92, P: 10]      |                    raw:                                      'matched'
[L: 92, P: 17]      |                    newline:                                  '\n'
[L: 93, P:  1]      |                    whitespace:                               '    '
[L: 93, P:  5]      |                    raw:                                      'AND'
```

### Expected Behaviour

The file parses correctly for 

```
MERGE INTO foo
USING
\t ...
WHEN MATCHED 
AND bar = baz 
THEN SET ...

WHEN NOT MATCHED 
AND baz > bar 
THEN INSERT ...
```

### Observed Behaviour

The parser failed (see above)

### How to reproduce

Add an `AND` condition to `WHEN NOT MATCHED` to any `MERGE` statement

### Dialect

Snowflake

### Version

0.10.0

### Configuration

```
[sqlfluff]

# verbose is an integer (0-2) indicating the level of log output
verbose = 0

# Turn off color formatting of output
nocolor = False

# Supported dialects https://docs.sqlfluff.com/en/stable/dialects.html
# Or run 'sqlfluff dialects'
dialect = snowflake

# One of [raw|jinja|python|placeholder]
templater = jinja

# Comma separated list of rules to check, or None for all
rules = None

# Comma separated list of rules to exclude, or None
exclude_rules = None

# The depth to recursively parse to (0 for unlimited)
[...]

PR #2670, opened 2022-02-16
Snowflake dialect update for `MERGE INTO` predicates

The-Loud:
Fix for #2666 
Updated the snowflake dialect file to handle `AND` within the `WHEN NOT MATCHED` clause.

Updated snowflake merge dialect test to include an example.

### Pull Request checklist
- [x ] Please confirm you have completed any of the necessary steps below.

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_snowflake.py
+++ b/src/sqlfluff/dialects/dialect_snowflake.py
@@ -3251,8 +3251,6 @@ class MergeNotMatchedClauseSegment(BaseSegment):
             "WHEN",
             "NOT",
             "MATCHED",
-            Sequence("AND", Ref("ExpressionSegment"), optional=True),
-            "THEN",
         ),
     )
     parse_grammar = Sequence(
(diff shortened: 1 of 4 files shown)


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3251,13 +3251,15 @@ class MergeNotMatchedClauseSegment(BaseSegment):
             "WHEN",
             "NOT",
             "MATCHED",
+            Sequence("AND", Ref("ExpressionSegment"), optional=True),
             "THEN",
```
tunetheweb, line 3255: I would remove these two lines.

The point of a `match_grammar` being different than a `parse_grammar` is to allow that to match quickly without doing the full parse. I think the addition of `Expression` in this PR makes that a little more effort, for not much extra gain so think just matching on `WHEN NOT MATCHED` is sufficient.

```suggestion
```

tunetheweb (changes requested):
Looks great!

Normally would suggest not including other unrelated changes like the bench one, but I agree with your change so let's just include it.

I've had two small suggestions on the other changes.

The-Loud:
Thank you for the highlights! Updated with your recommended changes.

tunetheweb:
> Thank you for the highlights! Updated with your recommended changes.

You'll need to regenerate the YAMLs for my second suggestion. After that think we're good to merge.

tunetheweb (approved):

tunetheweb:
Thanks! Hopefully will have given you the bug to add more PRs 😁

PR #2680, opened 2022-02-17
Exasol: Allow CROSS joins

sti0:

Currently the `CROSS` keyword within a `JOIN` clause is handled as an alias of the previous table. This breaks code after formatting because `sqlfluff` fixes the `JOIN` condition to `INNER JOIN` and removes the `CROSS` because it wasn't used as an alias. To solve this I moved the `CROSS` keyword from unreserved to reserved to make it work.

image

```sql
-- before I executed fix
SELECT -1 * row_number() OVER() AS nummer
FROM sys.exa_sql_keywords
CROSS JOIN sys.exa_sql_keywords
UNION ALL
SELECT 0;

-- after I executed fix => broken sql
SELECT -1 * row_number() OVER() AS nummer
FROM sys.exa_sql_keywords INNER JOIN sys.exa_sql_keywords
UNION ALL
SELECT 0;
```

nope

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_exasol_keywords.py
+++ b/src/sqlfluff/dialects/dialect_exasol_keywords.py
@@ -78,6 +78,7 @@ RESERVED_KEYWORDS = [
     "CONVERT",
     "CORRESPONDING",
     "CREATE",
+    "CROSS",  # a unreserved keyword but needed to be reserved to make join clause work
     "CS",
     "CSV",
     "CUBE",
@@ -530,7 +531,6 @@ UNRESERVED_KEYWORDS = [
     "COVAR_POP",
     "COVAR_SAMP",
     "CREATED",
-    "CROSS",
     "CURDATE",
     "DATABASE",
     "DATE_TRUNC",
--- a/test/fixtures/dialects/exasol/SelectStatement.sql
+++ b/test/fixtures/dialects/exasol/SelectStatement.sql
@@ -121,3 +121,9 @@ SELECT a, b, c FROM x
 union
 SELECT a, b, c FROM y
 ORDER BY a;
+----
+SELECT -1 * row_number() OVER() AS nummer
+FROM sys.exa_sql_keywords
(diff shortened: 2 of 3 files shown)


tunetheweb:
BTW an alternative way of handling this, if that really isn't a keyword, is by using `exclude` in the table alias clause:

```sql
class FromExpressionElementSegment(BaseSegment):
    """A table expression."""

    type = "from_expression_element"
    match_grammar = Sequence(
        Ref("PreTableFunctionKeywordsGrammar", optional=True),
        OptionallyBracketed(Ref("TableExpressionSegment")),
        # https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays#flattening_arrays
        Sequence("WITH", "OFFSET", optional=True),
        OneOf(
            Sequence(Ref("AliasExpressionSegment"), Ref("SamplingExpressionSegment")),
            Ref("SamplingExpressionSegment"),
            Ref("AliasExpressionSegment"),
            optional=True,
            exclude=Ref.keyword("CROSS"),
        ),
```

The downside of this is that's not an exasol specific segment so would either need to be copied in there to add this, or added to ANSI (which could cause problems for dialects without the `CROSS` keyword), or added via another function (that can be overridden in other dialects if needs be):

```py
            exclude=Ref("NonTableAliasKeywords"),
```
Then again it seems all the dialects have `CROSS` as a reserved keyword except Exasol, so your solution is probably the easiest even if not strictly 100% correct (would you confirm it definitely isn't a reserved keyword btw?).

sti0:
Exasol has a system table with all keywords inside. They are flagged as reserved or unreserved. `CROSS` is marked as unreserved but this only means, it could be used e.g. as a column name. But maybe not as a table name. Same for other unreserved keywords (they can be used in context, but not used in another). We've done that same procedure for the `PARTITION` keyword.

So yes, I think its the easiest and maintainable way to move it to the reserved keywords.

tunetheweb (approved):

PR #2681, opened 2022-02-17
Migrate tricky L004 tests to python

juhoautio:

The yaml fixtures based test for rules (`yaml_test_cases_test.py`) doesn't support checking this type of case:
- Some fixes are made
- The fix result still fails the lint because of some remaining errors

In plain python it's easy to check anything we want.

These two tests for rule L004 are some of the rare cases that need to be migrated before https://github.com/sqlfluff/sqlfluff/pull/2624 can be merged.

No.

### Pull Request checklist

  - Other.


test/rules/std_L004_test.py
```diff
@@ -0,0 +1,68 @@
...
+        FROM spam
+    """
+    )
+    configs = {}
+    res = assert_rule_fail_in_sql(RULE, fail_str, configs)
+    assert res == fix_str
+    assert lint(res, configs) == 2 * [
+        "Incorrect indentation type found in file. "
+        "The indent occurs after other text, so a manual fix is needed."
+    ]
```
tunetheweb, line 36: Does this run all the rules, and then only assert that L004 is in the fail string? Or does it only run rule L004? It looks like it's the former.

Because the YAML version of this test only runs L004 (at least that was my understanding). It's a unit test of L004 rather than a full test of the linter.

As we know, strange things can sometimes happen when multiple rules are run together so we should isolate this just to test L004, especially when supplying a `fix_str` IMHO.

In this particular case that is the only rule failing, but that might not always be the case (if someone doesn't right as clean a test case, or a new rule comes along in future). As I say, I think it's cleaner to unit test L004. Which is something the YAML test suite does automatically (though with the downside that it can't be used for tests where we explicitly want to use multiple rules).

juhoautio: Fixed. You have a point :) But for example here, if there was an actual error, it would've been that the test could fail too easily if anything. And that's not so bad.

It may be a bigger problem that test framework is very limited and doesn't allow checking the specific thing that is expected to happen (like the YAML rule tests are), because it can lead to situations where a test case keeps succeeding because it caught _some_ violation, even if _the_ violation that the test was created to find is not caught any more.

On the other hand I can't immediately think of anything else missing from the YAML based tests than the method for asserting what exact violations were found. But is it it really going to be enough, or would it hit another wall soon 🤔  Anyhow, I should give at least a quick stab at the required extension for YAML to see if I hit any bigger obstacle.

tunetheweb (approved):

PR #2687, opened 2022-02-18
Spark3: Support for `LATERAL VIEW` clause

R7L208:

- ~~update `sqlfluff/core/parser/grammar/anyof.py` to allow `exclude` named argument to be an iterable~~
- created `LateralViewClauseSegment` in `sqlfluff/dialects/dialect_spark3.py` to match/parse the clause
- removed unneeded comments & TODOs
- Add exclusion for `LATERAL` keyword in `AliasExpressionSegment`
- Redefine `FromExpressionElementSegment` from `sqlfluff/dialects/dialect_ansi.py` to include `LateralViewClauseSegment`
- Updated L026 to be default off for Spark3 due to false positives in LATERAL VIEW (and structs are applicable here as well)

- ~~I added a try/except in `sqlfluff/core/parser/grammar/anyof.py` that will use `any()` to exclude segments then revert to the previously existing code block on a TypeError~~
- Updated docstring to reflect all dialects in the exclusion list for L026

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -1726,7 +1752,10 @@ class AliasExpressionSegment(BaseSegment):
             ),
             # just a table alias
             Ref("SingleIdentifierGrammar"),
-            exclude=Ref("JoinTypeKeywords"),
+            exclude=[
+                Ref.keyword("LATERAL"),
+                Ref("JoinTypeKeywords"),
+            ],
```
tunetheweb, line 1758: On the past we've just used this:

```suggestion
            exclude=OneOf(
                Ref.keyword("LATERAL"),
                Ref("JoinTypeKeywords"),
            ),
```

Does that negate the need for the iterable changes above? Or still think those are better?

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -1844,3 +1870,36 @@ class FileReferenceSegment(BaseSegment):
...
+
+    type = "from_expression_element"
+    match_grammar = Sequence(
+        Ref("PreTableFunctionKeywordsGrammar", optional=True),
+        OptionallyBracketed(Ref("TableExpressionSegment")),
+        AnyNumberOf(
+            Ref("LateralViewClauseSegment"),
+            min_times=1,
+            optional=True,
+        ),
```
tunetheweb, line 1890: Are these not contradictory? If it's optional isn't `min_times` actually `0`?

And then the optionality is implicit in `AnyNumberOf` I think.

```suggestion
        AnyNumberOf(Ref("LateralViewClauseSegment")),
```

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -1515,6 +1515,29 @@ class SamplingExpressionSegment(BaseSegment):
...
+    match_grammar = Sequence(
+        Indent,
+        "LATERAL",
+        "VIEW",
+        Ref.keyword("OUTER", optional=True),
+        Ref("FunctionSegment"),
+        AnyNumberOf(
+            Ref("AliasExpressionSegment", optional=True),
+            min_times=1,
+        ),
```
tunetheweb, line 1536: Are these not contradictory? If it's optional isn't `min_times` actually 0?

Could this not be simplified to:

```suggestion
        AnyNumberOf(Ref("AliasExpressionSegment")),
```

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -1726,7 +1749,10 @@ class AliasExpressionSegment(BaseSegment):
             ),
             # just a table alias
             Ref("SingleIdentifierGrammar"),
-            exclude=Ref("JoinTypeKeywords"),
+            exclude=OneOf(
+                Ref.keyword("LATERAL"),
```
tunetheweb, line 1753: You may not need the `Ref.keyword` now. You do need it, if it's the ONLY item in the `exclude` paramer as it needs a match, but think it will match agains the `OneOf` now without the need of that.

```suggestion
                "LATERAL",
```

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -1844,3 +1870,36 @@ class FileReferenceSegment(BaseSegment):
         # to match as a `TableReferenceSegment`
         Ref("QuotedIdentifierSegment"),
     )
+
+
+@spark3_dialect.segment(replace=True)
+class FromExpressionElementSegment(BaseSegment):
+    """A table expression.
+
+    Enhanced from Ansi to allow for `LATERAL VIEW` clause
```
tunetheweb, line 1879: Nit

```suggestion
    Enhanced from ANSI to allow for `LATERAL VIEW` clause
```

tunetheweb (changes requested):
A lot cleaner now! But spotted a few more things.

src/sqlfluff/dialects/dialect_spark3.py
tunetheweb: Did you miss this one @R7L208 ?

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -1515,6 +1515,28 @@ class SamplingExpressionSegment(BaseSegment):
...
+
+    match_grammar = Sequence(
+        Indent,
+        "LATERAL",
+        "VIEW",
+        Ref.keyword("OUTER", optional=True),
+        Ref("FunctionSegment"),
+        AnyNumberOf(
+            Ref("AliasExpressionSegment"),
+        ),
```
tunetheweb, line 1535: One last question, I promise!

Wouldn't this allow this?:

```sql
    LATERAL VIEW OUTER EXPLODE(ARRAY()) AS tbl_name AS c_age;
```

R7L208: Please keep asking 🙂  Updated (and will include CI updates as well) to fix this.

Using the below instead.

```python
        Ref("SingleIdentifierGrammar", optional=True),
        Ref("AliasExpressionSegment", optional=True),
```

I ran the query with `AS tbl_name AS c_age;` and it failed. The first `AS` isn't allowed like you raised.

R7L208: Tested the second `AS` and going to push one more commit then should be good to go

R7L208: Also removed two other test cases that were not valid so thank you for catching that initial case.

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -1515,6 +1515,35 @@ class SamplingExpressionSegment(BaseSegment):
...
+        # NB: AliasExpressionSegment is not used here for table
+        # or column alias because `AS` is optional within it
+        # (and in most scenarios). Here it's explicitly defined
+        # for when it is required and not allowed.
+        Ref("SingleIdentifierGrammar", optional=True),
+        Sequence(
+            "AS",
+            Ref("SingleIdentifierGrammar"),
+            optional=True,
+        ),
```
tunetheweb, line 1542: My read of [the above link](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html) is this should be:

```suggestion
        Sequence(
            "AS",
            Delimited(
                Ref("SingleIdentifierGrammar")
            ),
            optional=True,
        ),
```

Is that so? And if so should we add a test case for multiple aliases?

tunetheweb (approved):
Got there in the end!
😁

Issue #2436, opened 2022-01-23
Add 'reflow' capability for configurable, black-like deterministic formatting


NiallRees:
# Feature
Same functionality as black, but totally configurable. The 'reflow' command would produce a deterministic output for the same 'code_only' input, regardless of the non-code elements (whitespace, capitalisation, indentation).

## Benefits:
- Fully user configurable code style
- Deterministic output for queries with the same choice and order of code segments
- No need to maintain and create rules for formatting-only concerns, which is hard currently as we have to detect bad SQL, which has infinite possibilities. By letting the user decide what good SQL is, we only have to define how the SQL is rendered.
- The implementation suggested is purely additive in relation to the parse tree, mitigating issues where formatting rules can interfere with each other.

## Illustrative example
The user has a sql file with the following:
```
select a,b, 
c FROM schema.table TBL
```

The user has configured the following:
```
base:
  case: lower # render all base segments (BaseSegment) and children in lowercase
select_clause_elements:
  indent_from_parent: 1 # indents all of `select_clause_elements_segment` by one unit from parent segment (this grouping segment doesn't exist yet)
select_clause_element:
  pre_newline: true # adds a newline before `select_clause_element`
from_clause:
  pre_newline: true # adds a newline before `from_clause`
```
And so the output is:
```
select
    a,
    b,
    c
from schema.table tbl
```

## How
We're already embedding Indent and Dedent into the dialects, this feature builds upon that to add all configuration and functionality to render valid SQL from code-only elements.
1. We parse the SQL as usual
2. We remove all non-code segments from the parse tree (output here is what our dialect test fixture ymls look like).
3. Using this parse tree, we render the 'reflowed' SQL.

## We've talked for a while about the idea of separating rules by what they do. It seems we can clearly draw a line between:
1. Rules which care about query structure (choice and ordering of code segments)
2. Rules which just care about appearance (non-code segments, strictly whitespace, capitalisation and indentation).

This feature focuses strictly on the latter and would not add, remove or modify code segments beyond their capitalisation.

## Implementation suggestions
- Embed the rendering logic into the dialect grammars/segments for maximum maintainability and ease of contribution. This is deliberately moving away from the approach in [L003](https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/rules/L003.py) which is very difficult to maintain due to its aim of handling all indentation across every grammar and segment.
[...]

PR #2691, opened 2022-02-20
[WIP] Add reflow capability

NiallRees:
Initial work toward #2436.

Introduces a new 'reflow' capability for deterministically formatting an input statement.

Next steps before merging:
- Add user configurable config.
- (Unsure if this should be a requirement of merging) Produce complete config and logic to render valid SQL for all existing dialects, segments and grammars. The way to achieve this would be to assert that running reflow against all our parse fixtures results in a parseable output.

Other things to consider:
- Making this work with templated code (should be very similar to how we work with existing rule fixes)
- Comments are none code segments and so would currently be removed

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -9,6 +9,8 @@ from sqlfluff.core import (
     SQLFluffUserError,
 )
 
+from sqlfluff.core.parser.segments.base import BaseSegment
+from sqlfluff.core.parser.segments.raw import WhitespaceSegment, NewlineSegment
 
 def get_simple_config(
     dialect: Optional[str] = None,
@@ -172,6 +174,22 @@ def parse(
         return {}
     return parsed.tree.as_record(show_raw=True)
 
+def reflow_tree(segment: BaseSegment, last_inserted=None):
+    """Reflows the tree into one with consistent formatting."""
+    reflowed_segments = []
+    if segment.segments:
+        for segment in segment.segments:
+            if segment.is_code:
+                if segment.reflow_pre_space and (not last_inserted or last_inserted.type not in (WhitespaceSegment, NewlineSegment)):
+                    reflowed_segments.append(WhitespaceSegment())
+                    last_inserted = reflowed_segments[-1]
+                reflowed_segment, last_inserted = reflow_tree(segment, last_inserted=last_inserted)
+                reflowed_segments.append(reflowed_segment)
+        tuple(reflowed_segments)
+        return segment.__class__(tuple(reflowed_segments)), last_inserted
+    else:
+        return segment, segment
+
 
 def reflow(
     sql: str,
@@ -189,4 +207,6 @@ def reflow(
     # If we encounter any parsing errors, raise them in a combined issue.
     if parsed.violations:
         raise APIParsingError(parsed.violations)
-    return parsed.tree.reflow()
+
+    reflowed = reflow_tree(parsed.tree)
+    return reflowed.raw()
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -174,6 +174,7 @@ def parse(
         return {}
     return parsed.tree.as_record(show_raw=True)
 
+
 def reflow_tree(segment: BaseSegment, last_inserted=None):
     """Reflows the tree into one with consistent formatting."""
     reflowed_segments = []
@@ -185,8 +186,8 @@ def reflow_tree(segment: BaseSegment, last_inserted=None):
                     last_inserted = reflowed_segments[-1]
                 reflowed_segment, last_inserted = reflow_tree(segment, last_inserted=last_inserted)
                 reflowed_segments.append(reflowed_segment)
-        tuple(reflowed_segments)
-        return segment.__class__(tuple(reflowed_segments)), last_inserted
+        segment.segments = tuple(reflowed_segments)
+        return segment, last_inserted
     else:
         return segment, segment
 
--- a/src/sqlfluff/api/simple.py
+++ b/src/sqlfluff/api/simple.py
@@ -171,3 +171,22 @@ def parse(
     if parsed.tree is None:  # pragma: no cover
         return {}
     return parsed.tree.as_record(show_raw=True)
+
+
+def reflow(
+    sql: str,
+    dialect: str = "ansi",
+    config_path: Optional[str] = None,
+) -> Dict[str, Any]:
+    """Parse and reflow the tree in a consistent fashion."""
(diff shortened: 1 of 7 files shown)


barrywhart:
Can you explain this?
>Add default configuration and logic to produce valid SQL?

Can you explain how the SQL produced might be invalid? Are you thinking of cases like we've seen sometimes with the rules, where they (for example) place code at the end of a one-line comment, so it no longer parses?

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1293,6 +1294,7 @@ class SelectClauseElementSegment(BaseSegment):
     """An element in the targets of a select statement."""
 
     type = "select_clause_element"
+    reflow_pre_space = "true"
```
barrywhart, line 1297: Did you mean to make this a boolean value, as elsewhere in the file?
```suggestion
    reflow_pre_space = True
```

NiallRees: Certainly did!

NiallRees:
> Can you explain this?
> 
> > Add default configuration and logic to produce valid SQL?
> 
> Can you explain how the SQL produced might be invalid? Are you thinking of cases like we've seen sometimes with the rules, where they (for example) place code at the end of a one-line comment, so it no longer parses?

Apologies - was a hurried PR description. The SQL might be invalid if there isn't spacing, for example. Right now this actually gets rid of comments as they're 'not code'. But perhaps the shortest route to functioning here is to add a space before every segment? Then we can add additional config in future iterations. I have no idea if a single space before every segment will result in valid SQL though. I'll give it a try!

barrywhart:
I think a single space will work, except for inline comments -- you want a newline after those. A newline after every segment will almost certainly work, although it'll be ugly. (One possible corner case: Some dialects support multiple string literals, one per physical line, and treat those as a single string. It's basically the SQL equivalent of this Python behavior: https://docs.python.org/3/reference/lexical_analysis.html#string-literal-concatenation

src/sqlfluff/api/simple.py
```diff
@@ -171,3 +171,22 @@ def parse(
...
         return {}
     return parsed.tree.as_record(show_raw=True)
+
+
+def reflow(
+    sql: str,
+    dialect: str = "ansi",
+    config_path: Optional[str] = None,
+) -> Dict[str, Any]:
+    """Parse and reflow the tree in a consistent fashion."""
```
alanmcruickshank, line 181: I feel like this description should be explicit that it's reflowing _whitespace_ while leaving code untouched.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1245,6 +1245,34 @@ def edit(self, raw):
...
         raise NotImplementedError()
 
+    reflow_pre_space = False
+
+    @property
+    def pre_reflow_strs(self) -> str:
+        """TODO: Add logic for segments which should appear before this segment."""
+        str = ""
+        if self.reflow_pre_space:
+            str += " "
```
alanmcruickshank, line 1255: I don't know if this approach is going to work. We kind of want to have an idea of different _grades_ of gaps:
1. Places where we _recommend_ a break - regardless of the rest of the query (e.g. we might say that a `SELECT` or `FROM` should always be at the start of the line).
2. Places where there cloud be a space _or a newline_, and there are probably explicit rankings between them. To evaluate which we use - we would take the length of the line into account (and probably also the size of the indent).
3. Place where there should _only_ ever be a space and not a newline.

alanmcruickshank (commented):
I realise this is just an outline and I think the basics are good but we're probably going to need something a little smarter as a base version. I'm not sure there are many cases where just "there should be a space here" is going to be enough.

There's probably a relatively simple decision engine however which takes into account some properties of the segment, the current indent size and the rest of the segment to work out where to add breaks. There's a good chance however that it might have to be recursive though if we rely on delegation to subsegments, because the indentation of the parent segment depends on the indentation of it's children and visa versa.

tunetheweb:
Personally I'm a little confused where this fits in versus the rules (particularly L003). Is this intended to eventually be a replacement? An alternative? Complementary?

NiallRees:
I've flip flopped on whether this should be a rule or not. If it were a rule, the rule would be 'all SQL non-code formatting should adhere to the configuration'. That feels too broad for one rule, which led me down the path of creating a new entry point for this behaviour. I'm going to push some changes I made now (not working) which insert segments instead of strings.

src/sqlfluff/api/simple.py
```diff
@@ -171,3 +173,40 @@ def parse(
...
+    if segment.segments:
+        for segment in segment.segments:
+            if segment.is_code:
+                if segment.reflow_pre_space and (not last_inserted or last_inserted.type not in (WhitespaceSegment, NewlineSegment)):
+                    reflowed_segments.append(WhitespaceSegment())
+                    last_inserted = reflowed_segments[-1]
+                reflowed_segment, last_inserted = reflow_tree(segment, last_inserted=last_inserted)
+                reflowed_segments.append(reflowed_segment)
+        tuple(reflowed_segments)
+        return segment.__class__(tuple(reflowed_segments)), last_inserted
```
NiallRees, line 189: This doesn't work for the same reasons that we have all the 'fix' functionality for the rules

NiallRees:
I'm going to be away for a couple of weeks - I'm not at all precious over this if someone wants to pick it up

alanmcruickshank:
> Personally I'm a little confused where this fits in versus the rules (particularly L003). Is this intended to eventually be a replacement? An alternative? Complementary?

In my mind: this would not be a rule per-se, but it would be the mechanism that any fixing of "layout-type" violations would occur. The easiest way to expose that is initially to add it as a separate command to directly reflow the whole file, but in essence it could be used under the hood to selectively reflow particular sections of a file as required by the violation we're trying to fix.

tunetheweb:
OK so eventually L003 (and other whitespace rules) might use this to apply their fixes.

But initially it might be separate command. But the output would then still be compatible with L003 (and others) and wouldn't cause them to suddenly start flagging.

Some of that may require adding extra options to the whitespace rules - for example if we added some of the column alignment changes or the like that people have been asking for.

That about right?

barrywhart:
@NiallRees: Do you think you'll continue work on this? It's been a few months, just curious if we need to keep this PR open.

NiallRees:
> @NiallRees: Do you think you'll continue work on this? It's been a few months, just curious if we need to keep this PR open.

It makes me sad to say not right now :( hoping someone else will. Will close

PR #2718, opened 2022-02-25
Support for checking violations in YAML rule tests

juhoautio:
Credits & thanks to @barrywhart for this idea (https://github.com/sqlfluff/sqlfluff/pull/2624#discussion_r807452356) 🎉.

Adds a new optional feature for the YAML rule test cases (run with yaml_test_cases_test.py):
- `violations` or `violations_after_fix` can be defined in YAML to check for specific failures

Taking `violations_after_fix` into use for the tricky cases of L004. With that, the python test with the same coverage is deleted. In other words, this reverts/replaces https://github.com/sqlfluff/sqlfluff/pull/2681.

Also adding the `violations` assertion for one "fail_str only" case to demonstrate how that works.

No.

### Pull Request checklist

  - Other.


src/sqlfluff/testing/rules.py
```diff
@@ -120,6 +120,42 @@ def assert_rule_raises_violations_in_file(rule, fpath, violations, fluff_config)
...
+
+
+def assert_violations_before_fix(test_case, violations_before_fix):
+    """Assert that the given violations are found in the given sql."""
+    violation_info = [e.get_info_dict() for e in violations_before_fix]
+    try:
+        assert violation_info == prep_violations(test_case.rule, test_case.violations)
+    except AssertionError:
+        print("Actual violations:\n" + yaml.dump(violation_info))
+        raise
```
tunetheweb, line 138: Shouldn't this be covered by one of your tests below?

src/sqlfluff/testing/rules.py
```diff
@@ -120,6 +120,42 @@ def assert_rule_raises_violations_in_file(rule, fpath, violations, fluff_config)
...
+        line_numbers=test_case.line_numbers,
+    )
+    violation_info = [e.get_info_dict() for e in violations_after_fix]
+    try:
+        assert violation_info == prep_violations(
+            test_case.rule, test_case.violations_after_fix
+        )
+    except AssertionError:
+        print("Actual violations_after_fix:\n" + yaml.dump(violation_info))
+        raise
```
tunetheweb, line 156: Same here.

src/sqlfluff/testing/rules.py
```diff
@@ -135,38 +171,22 @@ def rules__test_helper(test_case):
...
-
-
-def lint(code: str, sql: str, configs: Optional[dict] = None) -> List[str]:
-    """Runs lint and returns the found violations."""
-    return [
-        v.description if isinstance(v, SQLLintError) else str(v)
-        for v in Linter(FluffConfig(configs=configs, overrides={"rules": code}))
-        .lint_string(sql)
-        .violations
-    ]
```
tunetheweb, line 172: Why are these three functions not needed anymore?

test/fixtures/rules/std_rule_cases/L004.yml
```diff
@@ -26,6 +26,12 @@ spaces_fail_custom_tab_space_size:
 spaces_fail_custom_tab_space_size_not_multiple:
   fail_str: "     \nSELECT 1"
 
+  violations:
+  - description: 'Incorrect indentation type found in file. The number of spaces is
+    not a multiple of '
+    line_no: 1
+    line_pos: 1
+
```
tunetheweb, line 34: Since we usually separate our test cases by two newlines, I always find it really confusing when we also add a newline between sections of a test. Appreciate some of these are existing rather than what you added but Any chance we could clean up the extra white spaces when touching that file? That's what I've been doing to slowly get rid of these.

tunetheweb (changes requested):

src/sqlfluff/testing/rules.py
juhoautio: I added them in https://github.com/sqlfluff/sqlfluff/pull/2681 as generic helpers for python tests, but there are no usages left after migrating the related test from python to yaml. I presume these functions are not needed in the future because the yaml framework covers the same needs.

src/sqlfluff/testing/rules.py
juhoautio: Could you elaborate what you mean?

tunetheweb: So the test coverage has dropped below 100% so we can't merge this. This line (and a similar exception below) are the reason.

So we either need to add a test case, or add a `  # pragma: no cover` to this exception (which isn't unusual for exceptions to be honest as they probably shouldn't really be reached except in, exceptional circumstances).

But curious why this area of code is not reached anyway since you have a test case for extra violations not being reached?

juhoautio, line 138: This is a hack though.

I wanted to simply write this:
```suggestion
    assert violation_info == prep_violations(test_case.rule, test_case.violations), \
        "Actual violations:\n" + yaml.dump(violation_info))
```

But it seems that the assertion message is hidden by the outer layers of the test framework 🤔 

I should figure out how to include the assertion message in output when the assertion fails.

But I suggest leaving that for later. This code works and can be merged, in order to migrate the problematic existing test cases ie. enforce explicit assertions for any test cases where violations remain after the fix.

tunetheweb: OK but then can we add a `  # pragma: no cover` to these lines so test coverage stays at 100%? I'm a bit nerdy about that, cause when less than 100% it's more difficult to see if new code has missing coverage, or rounding errors are causing a slight change. With 100% (even if some are `pragma: no cover` so not really 100%) makes it cleaner and easier to review code.

juhoautio: Ah, thanks, so it's about the coverage of "test framework".

I'll simplify the code as said here, that should fix the coverage problem: https://github.com/sqlfluff/sqlfluff/pull/2718#discussion_r815450507

juhoautio: > I should figure out how to include the assertion message in output when the assertion fails.

I tried to find out about this but I couldn't. pytest seems to include only the 1st line of the assert message.

Printing the actual violations in plain yaml format seems so useful, that I added the "hack" back.

juhoautio: Btw, to test coverage locally, is this the minimal scope that I need to run:

```
tox -e generate-fixture-yml,cov-init,py38,cov-report
```
?

Just wondering because for me the full py38 test suite takes quite long to run locally. I noticed that my IDE respects the `# pragma: no cover` comments so that is handy though, and let's me check individual cases quickly. But I'm looking for a one-liner to perform the "100% coverage" check locally. So that I can know that it will also pass on CI.

tunetheweb: That should be enough but I've noticed some lines aren't covered by that anymore (and haven't been for a while!) 😞 Been meaning to find some time to look at it but running locally is slow enough already so end up using CI for this check myself to be honest.

barrywhart: Do you see the full output if you run with `-vv`? I have often used that to see additional output (e.g. full text diff when expected and actual SQL do not match).

https://docs.pytest.org/en/latest/how-to/output.html

juhoautio: Ouch, ok. So even more tox targets would be needed to reach the 100%.

tunetheweb: Yeah think there is a windows specific check now for file paths, and also some dbt tests. Hency why I haven't priortised looking into this 🤨

juhoautio: @barrywhart `-vv` helps indeed – my bad 🤦‍♂️

Btw I'm running a single test with this command to test it:
```
pytest -s test/rules/yaml_test_cases_test.py -k L004_indented_comments_tab_config -vv
```

If my assert code is:
```
    assert violation_info == prep_violations(
        test_case.rule, test_case.violations_after_fix
    ), "Actual violations:\n" + yaml.dump(violation_info)
```

I can find this in the output:

> E       AssertionError: Actual violations:
E         - code: L004
E           description: Incorrect indentation type found in file. The indent occurs after other
E             text, so a manual fix is needed.
E           line_no: 2
E           line_pos: 4
E
E       assert [{'code': 'L004',\n  'description': 'Incorrect indentation type found in file. The indent occurs '\n                 'after other text, so a manual fix is needed.',\n  'line_no': 2,\n  'line_pos': 4}] == [{'code': 'L004',\n  'description': 'Incorrect indentation type found in file. The indent occurs '\n                 'after other text, so a manual fix is needed.',\n  'line_no': 2,\n  'line_pos': 2}]
E         At index 0 diff: {'line_no': 2, 'line_pos': 4, 'code': 'L004', 'description': 'Incorrect indentation type found in file. The indent occurs after other text, so a manual fix is needed.'} != {'code': 'L004', 'description': 'Incorrect indentation type found in file. The indent occurs after other text, so a manual fix is needed.', 'line_no': 2, 'line_pos': 2}
E         Full diff:
E           [
E            {'code': 'L004',
E             'description': 'Incorrect indentation type found in file. The indent occurs '
E                            'after other text, so a manual fix is needed.',
E             'line_no': 2,
E         -   'line_pos': 2},
E         ?               ^
E         +   'line_pos': 4},
E         ?               ^
E           ]

Having those lines prefixed with "E" is not friendly compared to the custom print (current test code) that is:

```
Actual violations_after_fix:
- code: L004
  description: Incorrect indentation type found in file. The indent occurs after other
    text, so a manual fix is needed.
  line_no: 2
  line_pos: 4
```

..which can be copied as is into yaml file. But that is even harder to find in the test output because it's printed some lines before the AssertionError :/ So for console output there is no perfect option.

Also, in the console the AssertionError printed quite early and my screen gets filled with lots of debug logging after it. Basically the last thing that is logged is this:

> FAILED test/rules/yaml_test_cases_test.py::test__rule_test_case[L004_indented_comments_tab_config] - AssertionError: assert [{'code': 'L004',\n  'description': 'Incorrect indentation type found in file. The indent occurs '\n                 'after o...

So it's hard to find the actual diff from the console output (Also need to realize to add `-vv`). I tried the same code with my IDE and it's basically nicer because it basically just shows the diff. But with that code my IDE entirely hides the "Actual violations_after_fix" message! It only shows me the string comparison.

The custom print (current code) works very nicely in IDE – there I basically only the important output:

```
Actual violations_after_fix:
- code: L004
  description: Incorrect indentation type found in file. The indent occurs after other
    text, so a manual fix is needed.
  line_no: 2
  line_pos: 4

FAILED
test/rules/yaml_test_cases_test.py:17 (test__rule_test_case[L004_indented_comments_tab_config])
[{'code': 'L004',
  'description': 'Incorrect indentation type found in file. The indent occurs '
                 'after other text, so a manual fix is needed.',
  'line_no': 2,
  'line_pos': 4}] != [{'code': 'L004',
  'description': 'Incorrect indentation type found in file. The indent occurs '
                 'after other text, so a manual fix is needed.',
  'line_no': 2,
  'line_pos': 2}]

<Click to see difference>
```

To conclude, at least for my workflow the custom print option works best.

juhoautio: Looks like the "pragma: no cover" worked. The coverage report shows 100%: https://app.codecov.io/gh/sqlfluff/sqlfluff/compare/2718?src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%20comments&utm_term=sqlfluff

tunetheweb (approved):

PR #2719, opened 2022-02-26
 Extend rule L051 to LEFT/RIGHT/OUTER JOINs

### LINKED ISSUE #2686: Extend rule L051 to LEFT and RIGHT JOINs
### Search before asking

### Description

Rule L051 could be extended to also include `LEFT` and `RIGHT JOIN`s to fully qualify them with the optional `OUTER` keyword.

Potentially offering a `join_type` config options: `[inner, outer]` with default being `inner` to keep backward compatibility.

Or maybe it makes more sense to make this into its own rule, I'm not sure which direction makes the most sense, but surely someone else does :smile: 

### Use case

The rule would flag the following anti-patterns:
```sql
SELECT
    baz.foo
FROM
    bar
    LEFT JOIN baz ON (baz.id = bar.id)
```

And suggest the following best practice:
```sql
SELECT
    baz.foo
FROM
    bar
    LEFT OUTER JOIN baz ON (baz.id = bar.id)
```

Same thing for RIGHT JOINs to be turned into RIGHT OUTER JOINs.

### Dialect

All dialects, as far as I could tell `LEFT JOIN` === `LEFT OUTER JOIN` in all SQL flavors.

### Code of Conduct

rpr-ableton:

fixes #2686

This is the first time I'm working on the rules and not the dialects, so I'm sure there are some things that I've missed/could be done in a neater way :slightly_smiling_face: Just let me know and I'll be happy to get it fixed.

I'm not ticking the PR checklist as I'm unsure what the 3rd item entails.

### Pull Request checklist

  - Other.


src/sqlfluff/core/rules/config_info.py
```diff
@@ -113,6 +113,10 @@
             "Should alias have an explict AS or is implicit aliasing required?"
         ),
     },
+    "fully_qualify_join_types": {
+        "validation": ["inner", "outer", "both"],
+        "definition": ("Which type of JOIN clause should be fully qualified?"),
```
tunetheweb, line 118:
```suggestion
        "definition": ("Which types of JOIN clauses should be fully qualified?"),
```

src/sqlfluff/rules/L051.py
```diff
@@ -36,23 +43,52 @@ class Rule_L051(BaseRule):
...
-            return None
+        join_clause_keywords = [
+            segment for segment in context.segment.segments if segment.type == "keyword"
+        ]
+
+        # We identify LEFT/RIGHT/OUTER JOIN and if the next keyword is JOIN.
+        if (
+            self.fully_qualify_join_types in ["outer", "both"]
+            and join_clause_keywords[0].name in ["right", "left", "full"]
+            and join_clause_keywords[1].name == "join"
```
tunetheweb, line 65: Is `name` always lowercase? This looks to be what previous version of the rule was using but curious if that was correct?

src/sqlfluff/rules/L051.py
```diff
@@ -36,23 +43,52 @@ class Rule_L051(BaseRule):
...
+        ):
+            # Insert OUTER after LEFT/RIGHT/OUTER
+            return LintResult(
+                context.segment.segments[0],
+                fixes=[
+                    LintFix.create_after(
+                        context.segment.segments[0],
+                        [WhitespaceSegment(), KeywordSegment("OUTER")],
```
tunetheweb, line 73: Should we set the case of the `OUTER` based on the case of the other keywords in case someone prefers lowercase?

src/sqlfluff/rules/L051.py
```diff
@@ -36,23 +43,52 @@ class Rule_L051(BaseRule):
...
+                    )
+                ],
+            )
+
+        # We identify lone JOIN by looking at first child segment.
+        if (
+            self.fully_qualify_join_types in ["inner", "both"]
+            and join_clause_keywords[0].name == "join"
```
tunetheweb, line 81: Same question about case sensitivity of `name`

src/sqlfluff/rules/L051.py
```diff
@@ -36,23 +43,52 @@ class Rule_L051(BaseRule):
...
+        ):
+            # Replace lone JOIN with INNER JOIN.
+            return LintResult(
+                context.segment.segments[0],
+                fixes=[
+                    LintFix.create_before(
+                        context.segment.segments[0],
+                        [KeywordSegment("INNER"), WhitespaceSegment()],
```
tunetheweb, line 89: Same question about whether we should set the case of the `INNER` based on the case of the other keywords in case someone prefers lowercase?

tunetheweb (changes requested):
Looks pretty good to be! I do have some questions, which look to be issues with the existing rules.

src/sqlfluff/rules/L051.py
rpr-ableton: Yes, `name` seem to always return lowercase (irrelevant of the code capitalisation). On the other hand `raw` returns the keyword with the capitalisation found in the SQL.

I must admit I didn't dig that far into the codebase yet and stuck to the original approach.

src/sqlfluff/rules/L051.py
rpr-ableton: That's a great question! I'm not sure how other rules that append code when fixing linting errors behave. I'll look into that and update the code if necessary. :+1:

tunetheweb: Most of them try to do that where possible (see L033 as an example). And think we have the information we need here. We've had a long standing issue with one of the rules not doing that but can't find it now...

rpr-ableton: L043.py doesn't seem to care about capitalization in its fix part.

I think that overall makes sense. We're probably better off leaving capitalization constraints and logic in the rules dedicated to that end?

Also while browsing through the code I noticed that L043.py was leveraging `raw_upper` instead of `name` which I will also adopt in this rule as it clearly answers your above question.

tunetheweb: > We're probably better off leaving capitalization constraints and logic in the rules dedicated to that end?

I think it depends on ho w easy it is to implement. For some of the spacing issues for example, there's a lot of logic to consider. L003 is one of our most complicated rules, so wouldn't want to duplicate that logic in other rules (it's difficult to maintain in one rule!).

However for capitalisation, when it's obvious I think we should do it in the rule fixing it. Otherwise we're just making it do extra work and another round of fixes to correct that. Also it's possible that those other rules, for `consistent` setting, it see's this rule's choice and changes the rest of the SQL to match this choice. Which would be very confusing to the end user!

So while I don't think we should implement the full capitalisation logic, I think in this case it's simple enough to copy the capitalisation of the existing keywords it's checking. Even if that's a very simplified logic (e.g. if `left` or `right` then use `outer`, otherwise use `OUTER`? Or perhaps if `join` use `outer` else use `OUTER`). WDYT?

rpr-ableton: Your argument totally makes sense as long as we keep to a simple capitalisation logic in this rule :slightly_smiling_face: At the end of the day, you have a better overall view of architecture of this project, so I went with your suggestion of capitalising INNER/OUTER based on JOIN.

tunetheweb: Very nice implementation! Beautifully simple.

tunetheweb (approved):

Issue #2708, opened 2022-02-23
TSQL: L049 breaks code using = alias


fdw:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Rule L049 breaks exisiting code that uses `=` syntax.

Example:
```sql
select
  name = null
from table
```

becomes
```sql
select
  name is null
from table
```

### Expected Behaviour

The code remains unchanges, as it is correct.

### Observed Behaviour

The assignment `name = null` is changed to the comparison `name is null` which makes no sense in that context.

### How to reproduce

Run `sqlfluff fix` on the above code sample.

### Dialect

T-SQL

### Version

Latest from git `main`

### Configuration

Nothing relevant apart from setting the dialect.

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

barrywhart:
In your example, you are aliasing the column as `null`. Does that work in T-SQL, given that `null` is a SQL keyword?

fdw:
No, in my example `null` is the value of a column called `name`. Maybe you need the column for a CTE or a `union`, but it's value is always `null`.

tunetheweb:
Seems related to #2552 

We don’t currently support this in SQLFluff so it’s not surprising the rules are getting confused and assuming you mean something else. I think if this support is added to the TSQL dialect and this is seen as an alias rather than an assignment then this rule may not longer trigger. But would be good to test that if/when that support is added to be sure.

T-SQL has many differences to most other SQL dialects 🤷

barrywhart:
@fdw: Can you clarify what the query is doing?

From your expanation:
>in my example null is the value of a column called name. Maybe you need the column for a CTE or a union, but it's value is always null.

it sounds like this is another way of writing:
```
SELECT
    null AS name
```

Can you confirm?

We need to **understand** this query in order to fix the bug. At this point, I'm not yet convinced this **is** a bug. Thanks!

fdw:
> it sounds like this is another way of writing:
> 
> ```
> SELECT
>     null AS name
> ```
> 
> Can you confirm?

Yes, exactly.

> Perhaps you haven't pulled down the latest code and don't have that fix?

[...]

Issue #2487, opened 2022-01-26
T-SQL rule: Make alias style consistent


fdw:
I'm not even sure if rules just for one dialect are possible, but if so, the following would be nice:

T-SQL allows an additional style of aliasing, namely
```sql
select
  alias1 = column1
```

Currently, it's possible to use both even in the same query. It would be nice if `sqlfluff` could make this consistent (always `as`, always `=`, consistent within the query).

tunetheweb:
To me this could be an extension of [L011](https://docs.sqlfluff.com/en/stable/rules.html#sqlfluff.core.rules.Rule_L011) or [L012](https://docs.sqlfluff.com/en/stable/rules.html#sqlfluff.core.rules.Rule_L012). Especially if it forces `as` - then it would just be a matter of recognising and replacing the `=` for t-sql.

An extras step would be to offer the option of preferring (`=`), though as that's T-SQL specific it might be more confusing for others reading the docs.

fdw:
Yes, sounds like L012.

However, I have to say I quite like the `=` syntax, so I would maybe want to enforce it with `sqlfluff`. But, as you say, that would make at least the docs more confusing (and likely the code as well).

infused-kim:
> Yes, sounds like L012.
> 
> However, I have to say I quite like the `=` syntax, so I would maybe want to enforce it with `sqlfluff`. But, as you say, that would make at least the docs more confusing (and likely the code as well).

Hell yeah... team-= uprising. I finally found another person who prefers the = syntax...

So I am adding another vote for the ability to enforce the = syntax.

peterbud:
Related to #6854

PR #2727, opened 2022-02-27
T-SQL: Allow aliases with =

### LINKED ISSUE #2552: TSQL: Add support for column alias with equals (`=`) operator to rule L013
In addition to the ansi alias syntax (`getdate() as ansi_alias`), TSQL also supports aliases using the equals (`=`) operator: `equal_alias = getdate()`.

I propose to add support for that alias syntax and perhaps add a rule that let's users select a preferred style.

I would be happy to try implementing this if someone could give me a pointer into the direction of how this could be accomplished. 

## Steps to Reproduce
- Set dialect to tsql
- Lint the code below

```sql
select
    equals_alias = getdate(),
    getdate() as ansi_alias
from my_table
```

## Expected Behaviour
To not trigger `L013`.

## Observed Behaviour
Rule `L013` is triggered:
```log
L:   2 | P:   5 | L013 | Column expression without alias. Use explicit `AS`
                       | clause.
```

## Dialect
TSQL

## Version

- sqlfluff, version 0.9.4
- Python 3.9.9
- dbt 1.0.0

## Configuration

```ini
[sqlfluff]
# Run `sqlfluff dialects` to get a list of all dialects
dialect = tsql

# Use dbt as the templater
templater = dbt

# Use jinja templater instead of dbt
#templater = jinja

[sqlfluff:templater:jinja]
# Make dbt default macros like available like `ref`
apply_dbt_builtins = true

# Load custom macros from the macro directory
load_macros_from_path = ./macros/

# Allow mocking dbt_utils and other packages through
# python files placed in this dir
library_path = ./sqlfluff_libs/

# Mock variables that are used in dbt
target = prod
```

fdw:

Now supports T-SQL's alias syntax with `=`.

Fixed #2552

I think the alias should still be recognized correctly as such, but I'm not 100% sure.

### Pull Request checklist

  - Other.


tunetheweb:
This will still flag for rule L012. But In guess we should just turn off that rule if using this style of SQL? However then no way of enforcing this style so someone could use mixed styles.

Ideally we'd update L012 to recognise this as an implicit alias but maybe that's beyond the scope of this PR? WDYT?

fdw:
For this PR, I just wanted it to parse 😉 (That might also fix #2708.)

However, there's also #2487. As I've written there, I would like to offer a third option next to `explicit` and `implicit` just for T-SQL that forces everything to use `=`, which would also mean that the parser can distinguish between the three styles.
If you're alright with this, I'd suggest using this PR just for parsing and to continue the discussion about the rules in the issue 🙂

tunetheweb:
SGTM!

tunetheweb (approved):

PR #2729, opened 2022-02-27
Forgive Shorthand_CAST only / bracket only expressions from L013

OTooleMichael:
TD;DR - Feedback Requirements
- Does this need a config setting? (what should it look like)
- Could/should we "fix" column defs whose output column name is implicit eg `bar_col::VARCHAR` --> `bar_col::VARCHAR AS bar_col`
- Is the bracketed forgiveness too far?
- Should we introduce a no_useless_brackets rule too (going a little farther/superset of Rule_L053) ?

# Explanation
This modifies L013 to not catch the following syntax
`SELECT bar_col::VARCHAR, foo_col FROM table_x`

AS Aliases are desired (or even partially required) when there are meaningful transformations on an origin column (or if a column is generated from static values (VARHCARs or INTs etc).

The need for this requirement is that the output column will have not obvious name, and the chosen name varies from engine to engine eg `?column?` or a stringified version of the col definition `SUM(x)`. These output column names are definitively bad practice.

However there are circumstances where a select expression is "safe" without an alias:
- (*covered*) If a column_reference is selected and just passed through (eg. the `foo_col` from the example at the top) - here the selected column name will be reused as the output column
- (*covered*) if there is only one output column [from table_y] (so called scalar output) as this is a common syntax for later using `SELECT * FROM table_y` to represent a single value: tbh this is less important than the example added by this PR
- (*added*) if there is only a shorthand recasting of a column ref's type - the output column name will just reuse the provided input column and no be confused by the casting
- (*added*) if a single column ref is redundantly bracketed - redundant brackets will not change the column output name. should /could be handled by a redundant bracket rule?. This is relavent  in a double cast `bar_col::INT::VARCHAR` (this happens in Redshift as you cannot cast BOOL direct to VARCHAR for reasons unknown. The preceding might well be expressed as `(bar_col::INT)::VARCHAR` - although this is pointless it is neither wrong nor confusing about the column output.

The added functionality could be added with a config setting if so required/advised. Maybe `allow_simple_recasting` or some such.

- All existing tests continue to pass as expected but perhaps didn't consider this senario.
- Mildly changed existing the code block to use guard statements for readability

### Pull Request checklist

  -  [x]`.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
- Added appropriate documentation for the change.=


OTooleMichael:
Change in test coverage is due to the removal of the use of functional call which is:
- not needed due to the code change
- not independently tested
- not used in any other part of the code (or at least test base)

What should be done about that?

tunetheweb:
We should remove the code if not needed. No point in keeping dead code around!

OTooleMichael:
It is part of functional set. 
```python
def get_type() -> Callable[[BaseSegment], str]:
    ...
    return _


def get_name() -> Callable[[BaseSegment], str]:
   ...
    return _
    ```
    
 etc. 
 100% I should remove, or just comment out perhaps?
 
 I guess the set of helper functions isn't complete or exposed as an API.
 Ill knock it out, but maybe independent testing would be worth it?

**Comment by tunetheweb** [comment]
> 100% I should remove, or just comment out perhaps?

Remove. We have git if we need it back.

> I guess the set of helper functions isn't complete or exposed as an API. Ill knock it out, but maybe independent testing would be worth it?

The CI test is the only test we have so anything in that should cover any usage. We do give a warning in our [API docs](https://docs.sqlfluff.com/en/stable/api.html#advanced-api-usage):

> As of version 0.4.0 this is considered as _experimental only_ as the internals may change without warning in any future release. If you come to rely on the internals of SQLFluff, please post an issue on GitHub to share what you’re up to. This will help shape a more reliable, tidy and well documented public API for use.

File: src/sqlfluff/core/rules/functional/segment_predicates.py [hunk 1]
```diff
@@ -97,15 +97,6 @@ def _(segment: BaseSegment) -> bool:
...
 
-def get_type() -> Callable[[BaseSegment], str]:
-    """Returns a function that gets segment type."""
-
-    def _(segment: BaseSegment) -> str:
-        return segment.get_type()
-
-    return _
-
-
```
OTooleMichael, line 108: Wasn't used elsewhere and thus caused coverage issues to have dead code around

src/sqlfluff/rules/L013.py
```diff
@@ -47,42 +48,67 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
         """
         segment = context.functional.segment
         children = segment.children()
-        if segment.all(sp.is_type("select_clause_element")) and not children.any(
-            sp.is_type("alias_expression")
+        # If we have an alias its all G
+        if children.any(sp.is_type("alias_expression")):
+            return None
+
+        # If this isnt a select_clause then this rule doesnt apply
```
tunetheweb, line 55:
```suggestion
        # If this isn't a select_clause then this rule doesn't apply
```

src/sqlfluff/rules/L013.py
```diff
@@ -47,42 +48,67 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+        "column_reference",
+        "wildcard_expression",
+        "cast_expression",
+        "bracketed",
+    ]
+    selector = sp.not_(sp.is_type(*forgiveable_types))
+    filtered = select_clause_or_exp_children.select(selector)
+    remaining_count = len(filtered)
+
+    # Once we have removed the above if nothing remains this statment/expression
```
tunetheweb, line 102:
```suggestion
    # Once we have removed the above if nothing remains this statement/expression
```

tunetheweb (changes requested):
Some small bits from me.

@barrywhart wouod be good to get your view on this change.

src/sqlfluff/rules/L013.py
```diff
@@ -47,42 +48,67 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+            # report an error.
+            immediate_parent = context.functional.parent_stack.last()
+            elements = immediate_parent.children(sp.is_type("select_clause_element"))
+            num_elements = len(elements)
+
+            if num_elements > 1:
+                return LintResult(anchor=context.segment)
+            return None
+
+        # Just error if we don't care.
```
barrywhart, line 85: Should this comment be removed or changed? Previously, the comment was present when the `return` was inside `if self.allow_scalar`. I'm not sure it makes sense anymore. (Did it ever?)

barrywhart (commented):
Looks good overall. There's one comment that needs a second look after the code reorg.

tunetheweb (approved):

Issue #2221, opened 2022-01-01
Evaluate use of awesome-pattern-matching package to simplify rule writing


barrywhart:
Related to/part of #1808

As discussed in #1808, some rules might be easier to write if we had a pattern-matching capability.

Python 3.10 adds a built-in pattern matching feature , and there is a backport (although the backport looks new and kind of unofficial). From my evaluation, the built-in pattern matching is okay but not as powerful as I'd hoped, e.g. it doesn't support patterns with both leading and trailing "don't care" sections. It's also unclear if it supports matching across multiple levels of a tree structure.

[`awesome-pattern-matching`](https://github.com/scravy/awesome-pattern-matching) is a promising package. Among the packages out there, it comes closest to my dream of:
* regex-like power applied to arbitrary objects
* support for patterns that span levels of a tree". (AFAICT, most of the other pattern-matching packages aren't even trying to do one or both of these.)

Another **major** appeal is that the format of patterns closely resembles SQLFluff dialect definitions, so using it to implement rules could be very approachable and intuitive for contributors.

It's currently immature, e.g.:
* Low downloads count
* Lack of support for nested multi-segment patterns (I asked about it [here], and the author has an enhancement now in progress.)

But I believe it's worth keeping an eye on, revisiting it from time to time by trialing it to write or rewrite some SQLFluff rules.

barrywhart:
In evaluating the package, I created a new extension class that's useful for patterns that span multiple levels (i.e. segments and their children). Pasting that code here for safekeeping:
```
class Attr(Pattern, Nested):
    """Extension class for apm. Allows processing of an object field.

    Adapted from the package's similar "At" class.
    """

    def __init__(self, path, pattern):
        if isinstance(path, str):
            self._path = path.split(".")
        else:
            self._path = list(path)
        self._pattern = pattern

    def match(self, value, *, ctx: MatchContext, strict: bool) -> MatchResult:
        """Override abstract parent function."""
        for k in self._path:
            try:
                value = getattr(value, k)
            except AttributeError:
                return ctx.no_match()
        return ctx.match(value, self._pattern)

    def descend(self, f):
        """Override abstract parent function."""
        return Attr(path=self._path, pattern=f(self._pattern))
```

barrywhart:
Prototype -- rewrite of rule L017 using the package:
```
"""Implementation of Rule L017."""

from apm import match, Check, Some

[...]

PR #2735, opened 2022-02-28
L007 autofix

### LINKED ISSUE #1432: Rule L007 should be autofixable
Rule L007 checks for operators before or after newlines.

That is:

```sql
SELECT
    a
    + b
FROM foo
```

is preferred over

```sql
SELECT
    a +
    b
FROM foo
```

(or vice versa) depending on config.

As we can detect this pretty reliably, it seems it should be one that is auto fixable reliably too.

Only concern would be clashing with rule L016 (line length).

OTooleMichael:
L007 hasn't been a fixable Rule. This PR fixes that! :)

Fixes #1432

No side-effects

### Pull Request checklist


tunetheweb:
Excellent! - added "Fixes #1432" to the description.

src/sqlfluff/rules/L007.py
```diff
@@ -60,49 +67,148 @@ def _eval(self, context: RuleContext) -> LintResult:
...
+            after_list = []
+
+        if symbol_hit:
+            after_list.append(el)
+            continue
+
+        if start_code:
+            before_list.append(el)
+    # This code is unreachable if a valid expr and symbol/operator segment are passed.
+    raise Exception("Ivalid Expr: must contain binary_operator or comparison_operator")  # pragma: no cover
```
OTooleMichael, line 214: Explicit no cover and Exception for this deliberately unreachable code seems reasonable right?

test/fixtures/rules/std_rule_cases/L007.yml
```diff
@@ -55,3 +96,21 @@ passes_on_after_override:
...
+        + b
+    from foo
+  fix_str: |
+    select
+        a +
+        -- comment1!
+        -- comment2!
+        b
+    from foo
+
```
tunetheweb, line 115: Could we remove all newlines from test cases (including the ones you have not touched)?

Since we use a single newline between test cases I always find the other single newlines distracting and have been on a mission to slowly clean them up (I really should just do a bulk clean up!).

tunetheweb:
@barrywhart could you give this one a review as our resident rules expert?

test/fixtures/rules/std_rule_cases/L007.yml
OTooleMichael: Which new lines, the ones between yaml keys or the ones within tested SQL?

The later are kind of required for the test and rules. I imagine you mean the former?

tunetheweb: Lines 56, 97 and 115. They look like they are part of a separate test rather than continuation of the previous test if you get what I mean?

src/sqlfluff/rules/L007.py
```diff
@@ -60,49 +67,149 @@ def _eval(self, context: RuleContext) -> LintResult:
...
+        memory["since_code"] = []
 
         # Anchor is our signal as to whether there's a problem
-        if anchor:
-            return LintResult(anchor=anchor, memory=memory, description=description)
-        else:
+        if not anchor:
             return LintResult(memory=memory)
+
+        expr = cast(Tuple[BaseSegment], parent_stack[-1].segments)
```
barrywhart, line 122: Why are we casting this here? Would it be reasonable to actually just create a tuple? Generally, using `cast` is more of a workaround.

src/sqlfluff/rules/L007.py
```diff
@@ -60,49 +67,149 @@ def _eval(self, context: RuleContext) -> LintResult:
...
+    anchor: BaseSegment,
+    operator_new_lines: str,
+) -> List[LintFix]:
+    """Generate Fixes.
+
+    Take operator_new_lines==after as the default case.
+    Consider "before" as a modification on this behaviour.
+
+    """
+    fixes = [LintFix.delete(anchor)]
```
barrywhart, line 143: Could you add a few high-level comments to help explain the fixes being generated? "Fix" code is often a bit hard to follow.

src/sqlfluff/rules/L007.py
```diff
@@ -60,49 +67,149 @@ def _eval(self, context: RuleContext) -> LintResult:
...
+    expr: Tuple[BaseSegment],
+    symbol: BaseSegment,
+) -> Tuple[List[BaseSegment], List[BaseSegment]]:
+    """Create two lists of symbols surrounding operator.
+
+    1 from last code to operator/symbol,
+    2 from operator to next code
+
+    """
+    before_list: List[BaseSegment] = []
```
barrywhart, line 189: Could you look at writing this code using the `Segments()` class? I think it would be much shorter. In other words, if `expr` were an instance of `Segments()`, I think you could use `select()` and the keyword arguments `start_seg` and `end_seg` to build `before_list` and `after_list`.

src/sqlfluff/rules/L007.py
```diff
@@ -60,49 +67,149 @@ def _eval(self, context: RuleContext) -> LintResult:
...
+
+def _copy_el(segment: BaseSegment):
+    """Safely duplicate segment without position markers."""
+    el = copy.deepcopy(segment)
+    el.pos_marker = None  # type: ignore
+    return el
+
+
+def _get_surrounding_segments(
+    expr: Tuple[BaseSegment],
```
barrywhart, line 180: Does `expr` contain multiple `BaseSegments`? If so, I think this should be:```suggestion
    expr: Tuple[BaseSegment, ...],
```

File: src/sqlfluff/rules/L007.py [hunk 7]
```diff
@@ -60,49 +67,149 @@ def _eval(self, context: RuleContext) -> LintResult:
...
+            after_list = []
+
+        if symbol_hit:
+            after_list.append(el)
+            continue
+
+        if start_code:
+            before_list.append(el)
+    # This code is unreachable if a valid expr and symbol/operator segment are passed.
+    err = "Ivalid Expr: no binary_operator/comparison_operator"  # pragma: no cover
```
barrywhart, line 214:
```suggestion
    err = "Invalid Expr: no binary_operator/comparison_operator"  # pragma: no cover
```

src/sqlfluff/rules/L007.py
OTooleMichael: Nice - I didn't know about that functionality. Will. use in the future. However not sure if it will work, as I need the distance from last start token (last code) to first symbol. and .select is first start token to first symbol.
Or am I wrong

barrywhart: What do you mean by:
>I need the distance from last start token (last code) to first symbol

I don't see anything using distance in the code. It looks like you're building two lists:  a list of segments before and after the symbol.

Don't worry about this; working code is 💯 . We do encourage use of `Segments()` and the similar classes in general, because they often result in shorter, easier to understand code. (Generally, they tend to reduce the need to write custom loop code.)

OTooleMichael: I rewrote the whole test in the functional manner. Fns for the win!

It turns out to be only a net add of 80 lines (not including test) - I think this is mostly due to not using memory etc. and basing our guard statement on is this segment an expr containing one of our operators.


I did my best to avoid loops. But like I said it cant really be handled by select and this adds quite a few cycles over a naive loop.


What I meant is that this sequence where the segs are numbered in brackets.



`"("[0] "col_ref_1"[1] ")"[2] " "[3] "+"[4] "\n"[5] " "[6] "col_ref2"[7]`

All segs are `is_code` with the exception of `3,5,6`.

We want the stretch from `2-->7` ie the nearest code elements either side of the operator[4].
With select we are doing a lot of dancing, we need some sort of memory or marker to understand the position relative to "+" which requires extra and to be honest just as confusing loops :)

src/sqlfluff/core/rules/functional/segments.py
```diff
@@ -128,6 +128,10 @@ def __getitem__(self, item: int) -> BaseSegment:  # pragma: no cover
     def __getitem__(self, item: slice) -> "Segments":  # pragma: no cover
         pass
 
+    def __iter__(self) -> Iterator[BaseSegment]:
+        # Loops understand we are looping BaseSegment
+        return super().__iter__()
+
```
OTooleMichael, line 134: This generally helps with typechecking which is complaining a lot when you actually pull or map elements.
this mapping is often a final pass towards and insert delete etc.

barrywhart: Nice!

src/sqlfluff/rules/L007.py
```diff
@@ -58,51 +68,103 @@ def _eval(self, context: RuleContext) -> LintResult:
...
+                res = [els.reversed() for els in reversed(res)]
+
+            change_list, anchor_list = res
+            # If the anchor side of the list has no newline
+            # then everything is ok already
+            if not anchor_list.any(sp.is_name("newline")):
+                continue
+
+            insert_anchor = anchor_list.last().get()
+            assert insert_anchor, "Insert Anchor must be present"
```
barrywhart, line 97: What could cause this to fail? I'm wondering if we should handle it without raising an exception. (I'm remembering past bugs that arose from "surprises" in someone's SQL we were processing.)

src/sqlfluff/rules/L007.py
```diff
@@ -58,51 +68,103 @@ def _eval(self, context: RuleContext) -> LintResult:
...
+        relevent_types = ["binary_operator", "comparison_operator"]
+        segment = context.functional.segment
+        # bring var to this scope so as to only have one type ignore
+        operator_new_lines: str = self.operator_new_lines  # type: ignore
+        expr = segment.children()
+        operator_segments = segment.children(sp.is_type(*relevent_types))
+        results: List[LintResult] = []
+        # If len(operator_segments) == 0 this will essentially not run
+        for operator in operator_segments:
+            start = expr.reversed().first(FirstCodeAfterOperator(operator))
```
barrywhart, line 80: Nitpick: Rather than `expr.reversed().first()`, could you use `expr.last()`?

src/sqlfluff/rules/L007.py
```diff
@@ -58,51 +68,103 @@ def _eval(self, context: RuleContext) -> LintResult:
...
+            end = expr.first(FirstCodeAfterOperator(operator))
```
barrywhart, line 81: Nitpick: Could you eliminate the need for `FirstCodeAfterOperator` by chaining two calls:
```
expr.select(start_seg=operator).first(sp.is_code())
```

src/sqlfluff/rules/L007.py
```diff
@@ -58,51 +68,103 @@ def _eval(self, context: RuleContext) -> LintResult:
...
+        LintFix(
+            edit_type=edit_type,
+            edit=list(map(lambda el: copy.deepcopy(el), reversed(inserts))),
+            anchor=insert_anchor,
+        ),
+        # remove the Op
+        LintFix.delete(operator),
+        # Delete the original elements (related to insert)
+        *[LintFix.delete(seg) for seg in change_list],
+    ]
```
barrywhart, line 164: I wonder if this would be easier to read?
```suggestion
    ] + [
        LintFix.delete(seg) for seg in change_list
    ]
```

barrywhart (commented):
I have a question about the `assert`, whether we think that could happen in real life.

A few other nitpicky suggestions, but nothing that "has" to change.

OTooleMichael:
This PR is getting a bit cumbersome - it starts hitting up against typing issues. But I think its resolved-ish now.

src/sqlfluff/rules/L007.py
OTooleMichael: Im using an assert cause it cant happen. This is just for the typechecking / gaurding.
Assert prevents making a dead branch and thus reducing the coverage

OTooleMichael: call signature of .get() has a None, but it will never happen here. If it did it would probably only be regression and not a runtime thing.

src/sqlfluff/rules/L007.py
OTooleMichael: Is that not also confusing the `.last(FirstCodeAfterOperator(operator))`
The last First is a bit hard to get right ?

src/sqlfluff/rules/L007.py
OTooleMichael: legit

src/sqlfluff/rules/L007.py
OTooleMichael: Black was giving me huge issues with complaining about various linting in and round these style things

src/sqlfluff/rules/L007.py
barrywhart: I added a comment with a _possible_ way to do this using `select()` and `first()`.

This is great work. Thanks for doing this!

When we added the functional API a few months ago, we also looked at using another package called `awesome-pattern-matching`. It can handle some tricky scenarios that are too complex for the functional API. There's a GitHub issue talking about it (#2221) which includes some rules I rewrote to use it (as a proof of concept). Take a look if you are interested. (Not for this PR!)

src/sqlfluff/rules/L007.py
OTooleMichael: Yea nah - I tried and black wants it on one line if you go for a + operator. and then the comment is floating in the wrong place

OTooleMichael: What about
`*change_list.apply(LintFix.delete),`

barrywhart: `apply()` is good. Don't worry too much about this; I was just curious -- don't want to obsess **too much** over style. How much is "just enough", ha ha? 🤔

src/sqlfluff/rules/L007.py
barrywhart: 👍

barrywhart (approved):
Nice work!

barrywhart:
@OTooleMichael: The coverage check is failing. I'm on my phone, so I can't see the error. Could you look into it?

src/sqlfluff/core/rules/functional/segments.py
```diff
@@ -64,6 +64,8 @@ def raw_slices(self) -> RawFileSlices:
             )
         raw_slices = set()
         for s in self:
+            if s.pos_marker is None:
+                raise ValueError("Segments include a positionless segement")
```
barrywhart, line 68: To make the coverage check happy.
```suggestion
                raise ValueError("Segments include a positionless segement")  # pragma: no cover
```

src/sqlfluff/core/rules/functional/segments.py
```diff
@@ -64,6 +64,8 @@ def raw_slices(self) -> RawFileSlices:
             )
         raw_slices = set()
         for s in self:
+            if s.pos_marker is None:
+                raise ValueError("Segments include a positionless segement")  # pragma: no cover
```
barrywhart, line 68:
```suggestion
            if s.pos_marker is None:
                raise ValueError(
                    "Segments include a positionless segment"
                )  # pragma: no cover
```

barrywhart:
@OTooleMichael: I think so. I added a `#pragma: no cover`. Still waiting for the build to finish. It seems very slow today.

PR #2737, opened 2022-02-28
L032 fixable in easy cases

OTooleMichael:
Make L032 fixable in cases when its possible to imply the tables involved

- This is the first case where we add / insert complex tree objects, There might be some effect related to this.

### Pull Request checklist

  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -84,11 +84,13 @@ class BaseSegment:
...
     # What other kwargs need to be copied when applying fixes.
     additional_kwargs: List[str] = []
+    # This typing isnt really true as we set the pos_marker to None in places
+    pos_marker: PositionMarker
 
     def __init__(
         self,
         segments,
-        pos_marker=None,
+        pos_marker: PositionMarker = None,
```
barrywhart, line 93: Having a default value of `None` means it must be `Optional`.
```suggestion
        pos_marker: Optional[PositionMarker] = None,
```

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -115,14 +117,17 @@ def __init__(
         if not pos_marker:
             # If no pos given, it's the pos of the first segment.
             if isinstance(segments, (tuple, list)):
-                pos_marker = PositionMarker.from_child_markers(
-                    *(seg.pos_marker for seg in segments)
-                )
+                if all(hasattr(seg.pos_marker, "source_slice") for seg in segments):
```
barrywhart, line 120: Is this to allow for `None` values? Could we instead do:
```suggestion
                if all(seg.pos_marker for seg in segments):
```
`hasattr()` is a bit of a confusing way to work around `None` values.

barrywhart (changes requested):

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -84,11 +84,13 @@ class BaseSegment:
     allow_empty = False
     # What other kwargs need to be copied when applying fixes.
     additional_kwargs: List[str] = []
+    # This typing isnt really true as we set the pos_marker to None in places
+    pos_marker: PositionMarker
```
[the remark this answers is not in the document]
barrywhart: I took a look at this on my machine. There's definitely a lot of places that assume `pos_marker` is set, even though it may not be. It's probably best to leave it as you have it.

src/sqlfluff/core/parser/segments/base.py
barrywhart, line 93: See PR #2741. I think we may want to undo some of your changes here and merge that one first.

OTooleMichael:
@barrywhart - I merged in your changes and resolved conflicts. Any other adjustments needed?

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -116,14 +116,17 @@ def __init__(
...
+                        *(seg.pos_marker for seg in segments)
+                    )
             else:  # pragma: no cover
                 raise TypeError(
                     f"Unexpected type passed to BaseSegment: {type(segments)}"
                 )
-        self.pos_marker: PositionMarker = pos_marker
+
+        # TODO static typing is off for this var throughout the code base
+        self.pos_marker = cast(PositionMarker, pos_marker)
```
barrywhart, line 129: Now that `BaseSegmentpos_marker` is [marked as `Optional`](https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/core/parser/segments/base.py#L87), could we remove the `cast`?

src/sqlfluff/rules/L032.py
```diff
@@ -44,18 +56,130 @@ class Rule_L032(BaseRule):
...
+        unfixable_result = LintResult(
+            anchor=anchor,
+            description=description,
+        )
+
+        tables_in_join = parent_stack.last().children(
+            sp.is_type("join_clause", "from_expression_element")
+        )
+
+        # If we have more than 2 tables we wont try to fix join.
```
barrywhart, line 85:
```suggestion
        # If we have more than 2 tables we won't try to fix join.
```

src/sqlfluff/rules/L032.py
```diff
@@ -44,18 +56,130 @@ class Rule_L032(BaseRule):
...
+        # TODO: if this is table 2 of 3 it is still fixable
+        if len(tables_in_join) > 2:
+            return unfixable_result
+
+        predicate = sp.is_type("select_statement")
+        parent_select = parent_stack.last(predicate).get()
```
barrywhart, line 91: Think we could eliminate the variable?
```suggestion
        parent_select = parent_stack.last(sp.is_type("select_statement")).get()
```

barrywhart (changes requested):
A few small suggestions. Looks great overall!

src/sqlfluff/rules/L032.py
OTooleMichael: Black got upset with line length here :(

barrywhart: 👍

barrywhart (approved):
Looks great. Thanks for contributing. 💯 Can we "hire" you for the project? 😉

OTooleMichael:
@barrywhart  Sure - Im in the slack channel you can PM me there (TimmyOToole) is the name I think

PR #2739, opened 2022-02-28
Support Pascal case for L030

### LINKED ISSUE #2736: Support Pascal Case for L030
### Search before asking

### Description

It often makes sense to have function names as PascalCase, for example `DateAdd` or `IsNull`. Unfortunately, L030 only offers `upper`, `lower` and `capitalize`, which are all less than ideal for these cases.

### Use case

I would like my function names to be easily readable. For names that consist of multiple words, PascalCase makes the most sense.

### Dialect

At least T-SQL, but probably relevant for all dialects.

### Code of Conduct

tunetheweb:

Fixes #2736

This is a breaking change as L030 goes from using `capitalisation_policy` to `extended_capitalisation_policy` so anyone using it will need to change their config. Should be a 0.11.0 release.

Unfortunately it's not possible to support both for backwards compatibility.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -114,7 +114,7 @@ ignore_words = None
 
 [sqlfluff:rules:L030]
 # Function names
-capitalisation_policy = consistent
+extended_capitalisation_policy = consistent
 # Comma separated list of words to ignore for this rule
 ignore_words = None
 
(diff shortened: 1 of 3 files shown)


tunetheweb:
Calling out this point form PR template:

> ### Are there any other side effects of this change that we should be aware of?
> This is a breaking change as L030 goes from using `capitalisation_policy` to `extended_capitalisation_policy` so anyone using it will need to change their config. Should be a 0.11.0 release.
>
> Unfortunately it's not possible to support both for backwards compatibility.

I still think it's a good change though as Function names are more likely to want PascalCase.

Only thing is it doesn't support lower pascal case (e.g. `functionName` instead of `FunctionName`) which may be wanted by some. Might want to consider adding that in future.

barrywhart (approved):

PR #2742, opened 2022-02-28
Prepare release 0.11.0

tunetheweb:
We have a number of changes piling up since the last release so let's do another release in next few days.

I've bumped the version number to 0.11.0 as #2739 is a breaking change.

@iajoiner @NiallRees @OTooleMichael @barrywhart you have open PRs so try to close in next couple of days if want to be included.

Anyone else working on a PR that might be ready let me know in a comment below if you want me to check in with you before releasing.


CHANGELOG.md
```diff
@@ -9,6 +9,63 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Note: Changes are now automatically tracked in [GitHub](https://github.com/sqlfluff/sqlfluff/releases) and will be copied in here on each release (please remember to update the issues and contributors to links!). There is no need to manually edit this file going forward.
 -->
 
+## [0.11.0] - 2022-03-03
+
+## Highlights
+
+Major changes include:
+* Changes rule L030 to use `extended_capitalisation_policy` to support Pascal case **BREAKING CHANGE**
+* Fixes are no longer applied when they corrupt the parse try causing potential other issues.
```
barrywhart, line 18: Actually, at present the fixes are still applied, but it does log a warning message such as "After fixes were applied, segment failed the parse() check." Future versions may increase the log level or discard the problematic fixes. As part of this change, we did find and fix several rule issues, including the infamous "white space in name" issue.

CHANGELOG.md
```diff
@@ -9,6 +9,63 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+*  Extend rule L051 to LEFT/RIGHT/OUTER JOINs [#2719](https://github.com/sqlfluff/sqlfluff/pull/2719) [@rpr-ableton](https://github.com/rpr-ableton)
+* T-SQL: Allow aliases with = [#2727](https://github.com/sqlfluff/sqlfluff/pull/2727) [@fdw](https://github.com/fdw)
+* T-SQL: Support table variables [#2728](https://github.com/sqlfluff/sqlfluff/pull/2728) [@fdw](https://github.com/fdw)
+* Support for checking violations in YAML rule tests [#2718](https://github.com/sqlfluff/sqlfluff/pull/2718) [@juhoautio](https://github.com/juhoautio)
+* Roll back PR #2610 [#2726](https://github.com/sqlfluff/sqlfluff/pull/2726) [@barrywhart](https://github.com/barrywhart)
+* Redshift: Allow Whitespace Around Cast Operators [#2721](https://github.com/sqlfluff/sqlfluff/pull/2721) [@PLBMR](https://github.com/PLBMR)
+* Support database links in Oracle [#2725](https://github.com/sqlfluff/sqlfluff/pull/2725) [@tunetheweb](https://github.com/tunetheweb)
+* Rule L019: Ignore comma placement violations if the adjacent code is templated [#2717](https://github.com/sqlfluff/sqlfluff/pull/2717) [@barrywhart](https://github.com/barrywhart)
```
barrywhart, line 39: This one may be worth including in the release notes, since two different users reported it (two different GitHub issues, same underlying problem)

CHANGELOG.md
tunetheweb: Updated to:

> Log warnings for fixes that seem to corrupt the parse SQL as may cause incorrect fixes in other rules.

CHANGELOG.md
tunetheweb: Done

barrywhart (approved):

jpy-git:
Would be good to get the postgres values fixes/spark3 values cleanup PR #2797 merged before this release 😄

tunetheweb:
> Would be good to get the postgres values fixes/spark3 values cleanup PR #2797 merged before this release 😄

Merged.

Will release 0.11.0 in one hour from now. So if anyone's about to submit a PR then nows the time to speak up, or it can wait for next release.

jpy-git:
Thanks! All good on my side in terms of PRs for this release 👍

jpy-git:
Before we release, I'm a bit concerned about that breaking change introduced in #2739.

This comment on the PR isn't camel case it's a word with random capitalisation `FoO`
https://github.com/sqlfluff/sqlfluff/pull/2739#discussion_r816127261
so the test was correctly failing and shouldn't have been changed to pass.

PascalCase requires that each word within the string is capitalised, which seems like a bit of a challenge (haven't looked at the logic myself yet).

This is introducing weird capitalisation behavior:
This is apparently good
image
whilst this is bad
image

tunetheweb:
Here's our definition of `PascalCase`:
https://github.com/sqlfluff/sqlfluff/blob/03a513411bc1a51ea02f98dfee1f0a6fcac9fb93/src/sqlfluff/rules/L010.py#L159-L168

PascalCase (as distinct from CamelCase) needs to start with a capital letter, as per Wikipedia:

> Camel case is often used as a [naming convention](https://en.wikipedia.org/wiki/Naming_convention_(programming)) in computer programming, but is an ambiguous definition due to the optional [capitalization](https://en.wikipedia.org/wiki/Capitalization) of the first letter. Some programming styles prefer camel case with the first letter capitalised, others not.[[1]](https://en.wikipedia.org/wiki/Camel_case#cite_note-1)[[2]](https://en.wikipedia.org/wiki/Camel_case#cite_note-nfrmwk-2)[[3]](https://en.wikipedia.org/wiki/Camel_case#cite_note-3) For clarity, this article calls the two alternatives upper camel case (initial uppercase letter, also known as Initial Capitals, Initial Caps, InitCaps or Pascal case) and lower camel case (initial lowercase letter, also known as [dromedary](https://en.wikipedia.org/wiki/Dromedary) case). Some people and organizations, notably [Microsoft](https://en.wikipedia.org/wiki/Microsoft), use the term camel case only for lower camel case, designating Pascal case for the upper camel case.[[2]](https://en.wikipedia.org/wiki/Camel_case#cite_note-nfrmwk-2)

So do think the first case is PascalCase, while the second is not.

You could argue it shouldn't end in a single capital letter, but in certain case it will: `TheFirstLetterOfAlphabetIsA` and currently our definition allows it. It also allows all caps as PascalCase as per above.

tunetheweb:
Oh and we don't have a whole dictionary of words so don't recognise that `FoO` is a word (which it isn't even btw!) and so should be `Foo`.

tunetheweb:
Let me know i that addresses your concern @jpy-git ? Will hold off releasing until tomorrow am, but I'm comfortable.

I'm more concerned with people that have this set:

```
[sqlfluff:rules:L030]
# Keywords
capitalisation_policy = lower
````

And won't realise they now need to change that to this, for the same behaviour:

```
[sqlfluff:rules:L030]
# Keywords
````

As otherwise it will default to `consistent` and so allow Uppercase, Capitalized, or Pascalcase as long as it's consistent.

**Comment by jpy-git** [comment]
> So do think the first case is PascalCase, while the second is not.

Neither of my examples are PascalCase, the key idea of  PascalCase is that each **word** is capitalised for readability (i.e. `PascalCase` not `PascaLcase`). That definition in L010 is just "as long as the first letter is capitalised anything goes".

We're now allowing this as `consistent` where previously we would have told the user to sensibly pick upper or lower case.
image

IMO pascal case shouldn't be an option to these rules since it clearly doesn't work. The option is unhelpful if they want PascalCase but have an UPPERCASE function or Capitalised function or MeMeCaSE function

At the very least it shouldn't be included in the default `consistent` option and could just be its own option.
e.g.
```
consistent: upper or lower
upper: upper
lower: lower
pascal: pascal (use at your peril)
```

jpy-git:
I think it would be best to revert that PR and make this a 0.10.2 release for and we can come up with a sensible approach for PascalCase in an issue afterwards.

tunetheweb:
> Neither of my examples are PascalCase, the key idea of PascalCase is that each **word** is capitalised for readability (i.e. `PascalCase` not `PascaLcase`). That definition in L010 is just "as long as the first letter is capitalised anything goes".

The discussion on what SQLFluff recognises as PascalCase has already been had in #820

By your stricter definition based on real words neither `Foo`, nor `Bar` are PascalCase as those are not "words". Only allowing real words also negates a function called `MyCoolFunc` since Func is not a "word". I don't think such a strict definition is either useful or practical. Our current definition is very loose but IMHO does allow a very common use (e.g. `DateAdd`), which currently requires turning off the rule completely (that's what I do currently as this rule is useless to me). This was the ask for #2736. 

> IMO pascal case shouldn't be an option to these rules since it clearly doesn't work. The option is unhelpful if they want PascalCase but have an UPPERCASE function or Capitalised function or MeMeCaSE function

I disagree that it's unhelpful, though accept the limitations of what we can reasonably detect and/or autofix here.

> At the very least it shouldn't be included in the default consistent option and could just be its own option.

Personally I don't think using `consistent` is that advisable. Users should make an explicit decision here otherwise you risk some SQL being accepted with `upper`, some with `lower`...etc. However, as a default, it does have its use and avoids SQLFluff being over opinionated on those rules which don't have a majority consensus. The current process of L030 by default capitalising function names is not helpful for that. Additionally having a different definition of what `consistent` and/or `pascalcase` does in rule L030 and L014 is confusing to the user. I'm not aware of any complaints on L014 (except the underscore issue discussed in #820 which has been resolved).

Perhaps, in hindsight, `pascalcase` is the wrong name that was picked in #820 and it should instead be `mixedcase` (or, more accurately `mixedcasewithinitialcapital`) but personally I find `pascalcase` reasonable and easier understood, even if whether we can support "true" PascalCase is questionable. Again, to me, this falls in line with SQLFluff's looser definition of ANSI and other grammars where it makes snse.

As I say, if this was an obscure, uncommon, use for function names then your objection would give me enough pause to pull it out. But as it is, IMHO, very common I think we should include it, as is, in 0.11.0.

The absolute worst that happens with this, is someone with the default `consistent` setting gets a few less linting errors or auto fixes, until they make that explicit choice. That is acceptable to me (more breakages would be a different case!).

@sqlfluff/sqlfluff-maintainers (or other contributors), any thought on this?

OTooleMichael:
Interesting read there - If its already been resolved I think it hardly matters, but it feels similar to your request when I made "consistent table aliases" fixable - I think the extended policy and tighter scoped config isn't the best or needed.

consistent is an escape hatch - Perhaps it should only consider all UPPPER or all LOWER as they are by far the most common - this could just be written in the documentation (mini bonus is that Pascal is by one definition not consistent :) so its even pedantically arguable to be true). Pascal can be added as an option but not be included in the definition of consistent. (same way in which the fixing of consistent isn't a choice)

jpy-git:
> By your stricter definition based on real words neither `Foo`, nor `Bar` are PascalCase as those are not "words". Only allowing real words also negates a function called `MyCoolFunc` since Func is not a "word".

I never said they aren't allowed. The nuance of Pascal case is that the capital separate what the user considers a word (not the dictionary definition). My point was `MyCoolFunc` is fine, `MycoOlfUNc` is bad. Obviously it's difficult (if not impossible) to reliably detect these word boundaries.

I can see similar discussions #820, so whilst i disagree with this broken feature I'll leave it be 🤷

tunetheweb:
> whilst i disagree with this broken feature I'll leave it be 🤷

That's @jpy-git . Also discussed in the maintainers slack and agreed to include this. So going to drive on with the release now.

Happy to hear further feedback or suggestions how we could improve it going forward. Not, by any means, suggesting it's 100% right now, but think it's a good improvement for those using this style, and a reasonable default for all (whether using this or not).

tunetheweb:
> Interesting read there - If its already been resolved I think it hardly matters, but it feels similar to your request when I made "consistent table aliases" fixable - I think the extended policy and tighter scoped config isn't the best or needed.

@OTooleMichael could you clarify what you mean by this? And, just to be clear the `extended_capitalisation_policy` replaces `capitalisation_policy` (you can't have both - which is why it is a breaking change) as users using L030 with the non-default setting will need to update their config with this release. Maybe we don't need two configs if that's what you mean? At the moment that is done to prevent use of `pascalcase` for those rules that don't support it.

PR #2750, opened 2022-03-01
L012 bug fix for TSQL alternative alias types

### LINKED ISSUE #2749: T-SQL: Explicit aliasing of columns broken when using =
### Search before asking

### What Happened

I fear #2727 broke something.
Take this example:
```sql
select key = value
from table
```

With L012 set to explicit aliases, the result is invalid SQL:
```sql
select as key = value
from table
```

Instead, I would expect
```sql
select value as key
from table
```
or (preferably)
```sql
select key = value
from table
```

### Expected Behaviour

See above: Not broken code, either using `as` or `=` but not both ;)

### Observed Behaviour

`sqlfluff` currently adds an `as`, but also keeps the aliasing with `=`.

### How to reproduce

See above: Use that snippet and call `sqlfluff fix` on it.

### Dialect

T-SQL

### Version

Current `master`

### Configuration

```
[sqlfluff]
dialect = tsql

[sqlfluff:rules:L011] # Aliasing
aliasing = implicit

[sqlfluff:rules:L012] # Aliasing
aliasing = explicit
```

### Code of Conduct

tunetheweb:

Fixes #2749

Change is in rule L011 rather than L012 as it inherits all the functionality from there

### Pull Request checklist

  - Other.


src/sqlfluff/rules/L011.py
```diff
@@ -61,6 +61,15 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
 
         if context.segment.is_type("alias_expression"):
+
+            # T-SQL supports alternative alias expressions for L012
+            # select alias = value
+            # instead of
+            # select value as alias
+            # Recognise this and exit early
+            if context.functional.segment.children()[-1].name == "raw_equals":
+                return None
```
tunetheweb, line 71: FYI here's what the parse tree looks like for the alternative alias syntax:

```
[L:  1, P:  8]      |                    select_clause_element:
[L:  1, P:  8]      |                        alias_expression:
[L:  1, P:  8]      |                            identifier:                       'key1'
[L:  1, P: 12]      |                            whitespace:                       ' '
[L:  1, P: 13]      |                            raw_comparison_operator:          '='
[L:  1, P: 14]      |                        whitespace:                           ' '
[L:  1, P: 15]      |                        column_reference:
[L:  1, P: 15]      |                            identifier:                       'value1'
```

Is this the best way of identifying the alias ending with an `=` operator or is there a better way @barrywhart ?

fdw:
If I read this right, it could be possible (in the future) to automatically switch between all three styles. Right?

tunetheweb:
Yes should be doable in future.

src/sqlfluff/rules/L011.py
barrywhart: It seems reasonable to me to check for `raw_equals`. I think maybe I'd structure it a bit differently:
```
if context.functional.segment.any(sp.is_name("raw_equals"))
```

This reads a bit simpler (checks anywhere rather than a specific child index).

barrywhart: Think we could put this code in `L012.py`? It seems a bit odd having L012-specific code in `L011.py`. It _would_ require adding a bit of code to the other rule file. What do you think?

barrywhart (approved):
Looks good! A possible suggestion or two for changes. Up to you if you want to bother.

src/sqlfluff/rules/L011.py
tunetheweb: It should only be the last segment as per parse tree above so don’t think we should do the first suggestion to be honest.

For the second can you think of an easy way to do this without duplicating lots of code? Could add a config and have it off in L011 and on in L012 but then the actual code would still be in L011 so not sure it’s worth it?

barrywhart: Let me think about it a bit and get back.

tunetheweb: Will do. Tempted to keep it as is for simplicity to be honest. Can’t see it being a problem for L011. And who knows TSQL might even add this aliasing for tables at some point! Or am I just looking for excuses for the lazy option 😁

barrywhart: You can add the following lines to L012.py. This is only about 2 more lines than the current code adds to L011.py. I think this is a worthwhile change, because rule inheritance can become very confusing very fast.

```
        # T-SQL supports alternative alias expressions for L012
        # select alias = value
        # instead of
        # select value as alias
        # Recognise this and exit early
        if (
            context.segment.is_type("alias_expression")
            and context.functional.segment.children()[-1].name == "raw_equals"
        ):
            return None
        return super()._eval(context)
```

barrywhart: It seems very unlikely they'd add this for tables. I think they do it for columns because `=` looks like variable assignment.

tunetheweb: I like your suggestion. Will do that.

tunetheweb: Done!

barrywhart (approved):
🎉 💯

PR #2756, opened 2022-03-02
BigQuery: allow identifiers starting with dash

### LINKED ISSUE #2755: Valid BigQuery column names with only underscores and digits fail to parse
### Search before asking

### What Happened

I attempted to run SQLFluff on a query that contained a column with only underscores and numbers (in _XXXX_XX format) and I got a parse error.  I then confirmed this was valid BigQuery SQL and nothing in the rules prevents you from having a column with only numbers and underscores as long as it starts with an underscore.

### Expected Behaviour

I would expect SQLFluff to be able to parse valid BigQuery SQL when using bigquery as the dialect, including column names that begin with an underscore and only contain numbers.

### Observed Behaviour

When I run SQLFluff on a file that uses a column that starts with an underscore and contains only integer, I get parse errors and the file is not properly linted.

### How to reproduce

(1). Set up SQLFluff with bigquery as the dialect.

(2). Parse the following file as test.sql:
```
SELECT _2010_01
FROM model
```
(3). Run SQLFluff on that file.

### Dialect

bigquery

### Version

sqlfluff, version 0.10.1
Python 3.8.9
Using dbt 1.0.1 although the issue does not relate to the DBT templater

### Configuration

[sqlfluff]
dialect = bigquery

[sqlfluff:rules]
max_line_length = 120
comma_style = leading

[sqlfluff:rules:L010]
capitalisation_policy = upper

[sqlfluff:rules:L030]
capitalisation_policy = upper

### Code of Conduct

tunetheweb:

Fixes #2755

Nope

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -10,28 +10,29 @@ import itertools
 
 from sqlfluff.core.dialects import load_raw_dialect
 from sqlfluff.core.parser import (
+    AnyNumberOf,
     Anything,
     BaseSegment,
-    OneOf,
-    Ref,
-    Sequence,
     Bracketed,
-    Delimited,
-    AnyNumberOf,
-    KeywordSegment,
-    SymbolSegment,
-    RegexLexer,
-    StringLexer,
     CodeSegment,
-    NamedParser,
-    StringParser,
-    RegexParser,
-    Nothing,
-    StartsWith,
-    OptionallyBracketed,
-    Indent,
     Dedent,
+    Delimited,
+    Indent,
+    KeywordSegment,
     Matchable,
+    NamedParser,
+    Nothing,
+    OneOf,
+    OptionallyBracketed,
+    Ref,
+    RegexLexer,
+    RegexParser,
+    SegmentGenerator,
+    Sequence,
+    StartsWith,
+    StringLexer,
+    StringParser,
+    SymbolSegment,
 )
 from sqlfluff.core.parser.segments.base import BracketedSegment
 from sqlfluff.dialects.dialect_bigquery_keywords import (
@@ -132,7 +133,7 @@ bigquery_dialect.add(
     ),
     # Add a Full equivalent which also allow keywords
     NakedIdentifierSegmentFull=RegexParser(
-        r"[A-Z0-9_]*[A-Z][A-Z0-9_]*",
+        r"[A-Z_][A-Z0-9_]*",
         CodeSegment,
         name="naked_identifier_all",
         type="identifier",
@@ -158,6 +159,19 @@ bigquery_dialect.add(
 
 
 bigquery_dialect.replace(
+    # Override to allow _01 type identifiers which are valid in BigQuery
+    # The strange regex here it to make sure we don't accidentally match numeric
+    # literals. We also use a regex to explicitly exclude disallowed keywords.
+    NakedIdentifierSegment=SegmentGenerator(
+        # Generate the anti template from the set of reserved keywords
+        lambda dialect: RegexParser(
+            r"[A-Z_][A-Z0-9_]*",
+            CodeSegment,
+            name="naked_identifier",
+            type="identifier",
+            anti_template=r"^(" + r"|".join(dialect.sets("reserved_keywords")) + r")$",
+        )
+    ),
     FunctionContentsExpressionGrammar=OneOf(
         Sequence(
             Ref("ExpressionSegment"),
--- a/test/fixtures/dialects/bigquery/select_identifiers.sql
+++ b/test/fixtures/dialects/bigquery/select_identifiers.sql
@@ -1,0 +1,19 @@
+--Identifiers can start with an underscore in BigQuery
+-- and can contin just _ and 0-9
+SELECT _01
+FROM _2010_01;
+
+--Identifiers can start with an underscore in BigQuery
+-- and can contin just _ and 0-9
+SELECT col_a AS _
+FROM table1;
+
+-- TODO: Currently we don't support this but should
+-- Table names can contain dashes from FROM and TABLE clauses
+-- But reluctant to add to general naked_identifier grammar and not
+-- sure worth adding specific syntax for this unless someone requests it
+-- https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical
+-- SELECT * FROM data-customers-287.mydatabase.mytable;
+
+-- Same as above but quoted
+SELECT * FROM `data-customers-287`.mydatabase.mytable;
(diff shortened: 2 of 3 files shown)


src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -10,28 +10,29 @@
...
-    Matchable,
+    Ref,
+    RegexLexer,
+    RegexParser,
+    SegmentGenerator,
+    Sequence,
+    StartsWith,
+    StringLexer,
+    StringParser,
+    SymbolSegment,
```
barrywhart, line 35: Side note: That's quite a list of imports! 😅

barrywhart (approved):
💯

test/fixtures/dialects/bigquery/select_identifiers.sql
```diff
@@ -0,0 +1,19 @@
...
+
+--Identifiers can start with an underscore in BigQuery
+-- and can contin just _ and 0-9
+SELECT col_a AS _
+FROM table1;
+
+-- TODO: Currently we don't support this but should
+-- Table names can contain dashes from FROM and TABLE clauses
+-- But reluctant to add to general naked_identifier grammar and not
+-- sure worth adding specific syntax for this unless someone requests it
```
kdw2126, line 14: Hey, original poster of the linked issue here.  

We've learned to escape dashes to deal with lack of SQLFluff support but I did want to note that we do use dashes in our project names and so native support for it would be fantastic.  Don't know how common this usage pattern is but just wanted to add a +1 here :)

src/sqlfluff/dialects/dialect_bigquery.py
tunetheweb: Ordered alphabetically so looks bigger than it is. Think I only added `SegmentGenerator` and `RegexParser` in this PR.

barrywhart: Oh yeah, I didn't mean anything was wrong. It's just rare in Python to see an import list that long. If I start "importing the world", sometimes I switch to something like:
```
from sqlfluff.core import parser
```

But that requires all the references in that module to be prefixed with `parser.`.

Just ignore me. I'm feeling talkative on a late winter morning...

test/fixtures/dialects/bigquery/select_identifiers.sql
tunetheweb: Well there's two options:
1. Add this to `NakedIdentifierSegment` and `NakedIdentifierSegmentFull`. But then dashes will also be allowed in unquoted column names and aliases. Which they are not allowed. So overly permissive. Not the worst in the world and if it's invalid identifier then BigQuery will pick it up when it's run, but ideally SQLFluff would have already picked this up.
2. Add a new segment (`NakedIdentifierSegmentWithDashes`?) and then slot that in in place of `NakedIdentifierSegment`, but that's a bigger change I wouldn't want to do in this PR. And could be tricky to identify all those cases.

Thoughts on preference?

Also BigQuery does allow dashes in quotes too, that SQLFluff does support that. But presume I can't persuade you to move to that style? 😁

src/sqlfluff/dialects/dialect_bigquery.py
tunetheweb: Shorter than ANSI (just!).

test/fixtures/dialects/bigquery/select_identifiers.sql
barrywhart: I'm not using BigQuery at the moment, but the first sounds reasonable for now. SQLFluff's dialects are generally supersets of the actual language (for various reasons). In general, my philosophy is that SQLFluff is a linter/fixer, **not** a 100% syntax checker.

tunetheweb: Normally would agree @barrywhart but these [dashes are very much the exception](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical):

> Table name identifiers have additional syntax to support dashes (-) when referenced in FROM and TABLE clauses.

So were being quite liberal with ALL naked identifiers for one very specific (and I'd argue, as a BigQuery user niche) use case.

Let me mull over it some more in my head before merging this...

@kwd2126 would love to hear your thoughts?

kdw2126: @tunetheweb, now that you've laid out the blockers here, I agree with your logic of not dealing with this case yet.  I definitely think solution #1 would be an net regression for our team -- as I said, we are used to quoting our project name to deal with SQLFluff issues but having proper unquoted column name parsing break would create way more problems than this would solve.  #2 would be ideal but as you said that's not an easy fix.

In the interest of contributing to this project outside of constantly raising issues, maybe it would be valuable to have documentation for differences between BigQuery and SQLFluff ideal syntax to capture these edge cases?  I'd be happy to go through the BigQuery documentation and start playing around with finding these edge cases if the maintainers think this would be valuable to the project's existing documentation.

Plus, I'd feel less worried about the bug (and any other differences between SQLFluff and BigQuery syntax in general) if there was some document that new employees could use to better understand these edge cases.

tunetheweb: I think we should document as GitHub Issues and work through them, rather than in a separate doc.

Could you raise one for this case? I'll add the BigQuery label (annoyingly only maintainers can add labels) and then it's easily searchable for known BigQuery issues: https://github.com/sqlfluff/sqlfluff/labels/bigquery

And feel free to raise whatever other ones you find. We like to know the issues! Just can't promise when/if we'll get round to fixing them!

tunetheweb: > but having proper unquoted column name parsing break would create way more problems than this would solve.

Just checking you understood the drawbacks of 1 here @kdw2126 ?

I'm saying this currently breaks:

```sql
select col-1 as alias-1
```

If we went with option 1 (which is the simpler change, and potentially could be included in this fix as same regex to change), this would no longer break SQLFluff and would be considered valid BigQuery SQL, even though it is not.

No extra _valid_ SQL would break as a result of that change. It would just allow some extra _invalid_ SQL to get through now, whereas before, it would have blocked it. As @barrywhart says SQLFluff is not a perfect match for each dialect and often is a bit more liberal than the actual dialect. I'm just not sure if this change would make it _too_ liberal.

kdw2126: Sorry for the lack of clarity, edited my comment to clarify.  

I still believe that option 1 would not be desirable.  I understand that using SQLFluff to figure out if a command is valid BigQuery syntax in every case is not something the tool is designed to do, but I do appreciate that it is close enough to BigQuery syntax that it flags the example you gave with a parse error.  Having it miss invalid syntax examples like this would be a regression from my point of view, and the extra time we'd save by not having to quote our project view is trivial by comparison.

Understood on the documentation point, will continue to raise issues and look for other documentation issues to try to contribute.

tunetheweb: OK sounds good. Will merge this as is then.

PR #2757, opened 2022-03-02
Rough autofix for L028

OTooleMichael:
This adds fixing L028

However L028 is maybe flawed from the bottom up? The changes are subject to opinion and maybe not all are wanted/needed

- remove aliases when we desire unqualified
- add aliases when qualified
- add fix directive `fix_inconsistent_to` setting for fix inconsistent to either unqualified or qualified when encountered
- check the "STUCT" case when possible. `a.struct_value FROM tbl_name` must be a struct as `a` is not the table reference. If a is intended to be the table this code will not run. If we assume its valid code and "fix" it, it will not change its runnableness (valid code will remain valid [but now linted], in valid code will remain invalid).
- `fix_inconsistent_to`  also works on struct dialect
- Added tests to cover alias case, and schema.table case and quoted case. Some of these cases are a bit weird in the past and also now. Namely `SELECT schema.table.col1, table.col2 FROM schema.table` is consistent and qualified.

Fixes catches and understands structs better but also differently.

### Pull Request checklist

  - [x]`.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.

Notes:
- Typing is off `references = list(sc.recursive_crawl("object_reference"))` have methods like `is_qualified` but are typed as `List[BaseSegment]`. We can't run any sensible guard statement like `assert isintanceof(ref, (A,B,C))` cause `ObjectRefs` are reimplmented in various dialects directly from `BaseSegment`
- The Rule Inheritance feels like a bad idea `Rule_L028(Rule_L020)` - in this particular case it serve very little other than to make things hard to read and to make method/call signatures weird
- Issue with creating more complex tree elements occurs again. Table Str_ref may be quoted or naked depending on the query and it is hard to construct the correct tree types. Without correct types how will future Rules fix things correctly
- Inheritance essentially just does this block 
```python
if context.segment.is_type("select_statement"):
            select_info = get_select_statement_info(context.segment, context.dialect)
            if not select_info:
                return None

            # Work out if we have a parent select function
            parent_select = None
            for seg in reversed(context.parent_stack):
                if seg.is_type("select_statement"):
                    parent_select = seg
                    break
```

probably setting up an fn is more reusable. Also on a side note `get_select_statement_info` might be a good candidate for memo'ing


src/sqlfluff/core/rules/config_info.py
```diff
@@ -42,6 +42,10 @@
         "validation": ["consistent", "qualified", "unqualified"],
         "definition": "The expectation for references in single-table select.",
     },
+    "fix_inconsistent_to": {
+        "validation": [None, "qualified", "unqualified"],
+        "definition": "Fixing behaviour for inconsistent single-table select.",
+    },
```
tunetheweb, line 48: To be honest I'm not sure we should add the extra config.

If people have a preference they should set it in `single_table_references` and if they don't, then we can just pick first one and go with that.

The only thing this gives you is option to save `None` but then that's confusing in itself. Does that mean don't fix? Or is it only used when `single_table_references` is not set to `consistent`?

src/sqlfluff/rules/L028.py
```diff
@@ -32,20 +37,24 @@ class Rule_L028(Rule_L020):
...
+            foo.b
         FROM foo
 
         -- Also good
 
         SELECT
-            foo.a,
-            foo.b
+            a,
+            b
```
tunetheweb, line 48: Any reason you changed this? Or just reflecting your preference?

test/fixtures/rules/std_rule_cases/L028.yml
```diff
@@ -4,6 +4,22 @@ rule: L028
...
+  fail_str: SELECT my_tbl.bar, baz FROM my_tbl
+  fix_str: SELECT bar, baz FROM my_tbl
+  configs:
+    rules:
+      L028:
+        fix_inconsistent_to: unqualified
+
+test_fail_mixed_qualification_fix_to_qualified:
+  fail_str: SELECT my_tbl.bar, baz FROM "my_tbl"
+  fix_str: SELECT my_tbl.bar, "my_tbl".baz FROM "my_tbl"
```
tunetheweb, line 17: This is a bit weird as different references for same table. I can see why it's done it (only needed to fix the second one) but should we do a simpler test without quote to avoid this confusing people? Or is it specifically the quotes you want to test her?

test/fixtures/rules/std_rule_cases/L028.yml
```diff
@@ -21,15 +37,15 @@ test_pass_single_table_consistent_references_2_subquery:
...
   configs:
     rules:
       L028:
         single_table_references: unqualified
 
 test_fail_single_table_reference_when_qualified_config:
-  fail_str: SELECT bar FROM my_tbl
-
+  fail_str: SELECT bar FROM my_tbl WHERE foo
+  fix_str: SELECT my_tbl.bar FROM my_tbl WHERE my_tbl.foo
```
tunetheweb, line 48: Ohh interesting that it affects `WHERE` clause too! Did it always do that? Do we need a simple "pass" test for that?

src/sqlfluff/rules/L028.py
```diff
@@ -123,3 +98,135 @@ def _eval(self, context: RuleContext) -> EvalResultType:
...
+        return None
+    # A buffer to keep any violations.
+    violation_buff: List[LintResult] = []
+    col_alias_names: List[str] = [c.alias_identifier_name for c in col_aliases]
+    table_ref_str: str = table_aliases[0].ref_str
+    # Check all the references that we have.
+    seen_ref_types: Set[str] = set()
+    for ref in references:
+        this_ref_type: str = ref.qualification()  # type: ignore
+        if this_ref_type == "qualified" and dialect in ["bigquery", "hive", "redshift"]:
```
tunetheweb, line 125: I'd like to move these to a `_stuct_dialects` variable as part of this if possible, rather than have them buried in the code. My fault for initially implementing this with a hard coded list but since you're in there...

tunetheweb (changes requested):
Not had too much of a look at the code, as I've a more fundamental question on the new attribute and whether it's even needed.

Love the fact you support `STRUCTS` now for this rule!

src/sqlfluff/core/rules/config_info.py
OTooleMichael: So I'm sort of with you. adding config shoudn't be done without thought.
My main things is this - **Every rule that is fixable should be fixable**

This is a really nice property of SQLFluff and linters in general (if you think of the workflows involving precommit etc).

The "consistent" setting I could argue against, but in so far as it exists it is still fixable - we just need to know to which state.

Seems a bit weird (as just picking "qualified" would also work), however if you have a huge code base (like I do) and you have a handful of in consistent files seems like we should fix them if we could. (Right now I have to either change everything or... hand change things).
The argument for having consistent at (which I guess is "well thats how my code base is now") is the same one for allowing us to fix those files that fail to be consistent

src/sqlfluff/rules/L028.py
OTooleMichael: I was having linting issues with black (which actually turned out not to be this line) my bad :)

test/fixtures/rules/std_rule_cases/L028.yml
OTooleMichael: As per the description it is a bit weird, but its already like this (at least in terms of passing). As I further mentioned How does one validate that they are building the right tree, or how does one leave this up to the next/another Rule to Fix (which seems logical).

I popped in the test, because it is like this and the test at least flag that this is the current bahaviour

test/fixtures/rules/std_rule_cases/L028.yml
OTooleMichael: There is some clever stuff about managing self references in certain dialects.
Redshift can ref a created column.

That said 
```sql
SELECT 
      a,
      b,
      a + b AS col_created_right_here,
      col_created_right_here + 1 AS sub_self_ref
  FROM tbl
  ```
  Is also valid in Redshift, Snowflake and BQ 
  So.....

File: src/sqlfluff/rules/L028.py (hunk 5 above)
Comment (OTooleMichael) [reply src/sqlfluff/rules/L028.py:125]: Not sure I get you ? Happy to make a change if you can elaborate

File: src/sqlfluff/core/rules/config_info.py (hunk 1 above)
Comment (tunetheweb) [reply src/sqlfluff/core/rules/config_info.py:48]: Yeah I'm not against fixing `consistent`. I just think we should pick one, rather than offer another option to the user. If they really don't care (and picking `consistent` seems to suggest they don't) then there's no "wrong" answer. So pick one!

I originally thought picking based on first option makes most sense. so this:

```sql
SELECT col1, a.col2 FROM table1 a
```

Should be fixed to:

```sql
SELECT col1, col2 FROM table1 a
```

Defaulting to `qualified` would also work though and if that's easier than I say go for it. As I say IMHO there is no wrong answer here.

test/fixtures/rules/std_rule_cases/L028.yml
OTooleMichael: Ahhh that actually could be quite a large problem.
```sql
SELECT 
      tbl.a,
      tbl.b,
      tbl.a + tbl.b AS col_created_right_here,
      col_created_right_here + 1 AS sub_self_ref  -- noqa: disable=L028
  FROM tbl
  ```
  This noqa: is very much intentional (I normally want qualified but this is actually a self ref and I know it), will it be correctly respected ?

File: src/sqlfluff/rules/L028.py (hunk 5 above)
Comment (tunetheweb) [reply src/sqlfluff/rules/L028.py:125]: If we need to add a new dialect to this list we have to hunt through the code to find it. It feels kind of buried. It would be nice if this was a list at the top of the rule like this on L054:

https://github.com/sqlfluff/sqlfluff/blob/44ec589f7c892b3edf571a916b65a75cac652fff/src/sqlfluff/rules/L054.py#L87

As I say, it's completely unrelated to this change (and my bad for originally burying this in the code!), but since you're in this space thought I'd ask!

Get me?

Raised similar issue here: https://github.com/sqlfluff/sqlfluff/pull/2744#discussion_r816254779

File: test/fixtures/rules/std_rule_cases/L028.yml (hunk 4 above)
Comment (tunetheweb) [reply test/fixtures/rules/std_rule_cases/L028.yml:48]: I think we recently discovered that `noqa` doesn't work in the YAML files.

My original comment was we have a `fail` and `fix` test case for `WHERE` reference, but no `pass` test case. Should we have one? In theory `fail`/`fix` should be sufficient but I always like to have a `pass` test case in these instances too.

**Comment by tunetheweb** [comment]
@OTooleMichael what's your thoughts on this change and v0.11.0? Think we're ready to go on that but happy to hold off if you think you'll get this finished out in next day or so?

**Comment by tunetheweb** [comment]
Cool. Will hold off for this then. And that other one!

File: src/sqlfluff/rules/L028.py (hunk 5 above)
Comment (OTooleMichael) [reply src/sqlfluff/rules/L028.py]: think I got it. I set up Class Props

Comment (tunetheweb) [reply src/sqlfluff/rules/L028.py]: That's the word! Sorry not a python developer by trade so dunno the right terms :-)

File: src/sqlfluff/rules/L028.py (hunk 2 above)
Comment (tunetheweb) [reply src/sqlfluff/rules/L028.py:54]: Thought you were going to revert this?

File: src/sqlfluff/rules/L028.py [hunk 6]
```diff
@@ -14,6 +23,8 @@ class Rule_L028(Rule_L020):
        structs which trigger false positives. It can be enabled with the
        ``force_enable = True`` flag.
```
tunetheweb, line 24: Since you now support structs in this rule do we still need this note?
And the `force_enable` flag?

Also nit but if keeping it, can we update text to "This rule is disabled by default for BigQuery, Hive and Redshift due to their use of"

src/sqlfluff/rules/L028.py
```diff
@@ -58,68 +78,171 @@ def _lint_references_and_aliases(
...
-                        description="{} reference {!r} found in single table "
-                        "select.".format(this_ref_type.capitalize(), ref.raw),
-                    )
-                )
-            seen_ref_types.add(this_ref_type)
-
-        return violation_buff or None
+        self.single_table_references: str
+        # This could be turned into an option
+        fix_inconsistent_to = "qualified"
```
tunetheweb, line 83: Should we make it a class prop? See I've learned the term now :-)

src/sqlfluff/rules/L028.py
```diff
@@ -58,68 +78,171 @@ def _lint_references_and_aliases(
...
-
+        self.dialect: str = context.dialect.name
         # Some dialects use structs (e.g. column.field) which look like
         # table references and so incorrectly trigger this rule.
         if (
-            context.dialect.name in ["bigquery", "hive", "redshift"]
+            context.dialect.name in self._dialects_with_structs
             and not self.force_enable
         ):
             return LintResult()
```
tunetheweb, line 106: As per above curious if we still need this?

test/fixtures/rules/std_rule_cases/L028.yml
tunetheweb: Still wonder if we should have a simple `pass_str` test for this?

tunetheweb (commented):
Looks pretty good to me now, but have some more comments for you.

@barrywhart could you give it a review too since you know the rules code better than me.

test/fixtures/rules/std_rule_cases/L028.yml
barrywhart: Re: `noqa` in YAML tests. This should now work correctly for the usual YAML tests (`test/rules/yaml_test_cases_test.py::test__rule_test_case`). Until recently, `pass` tests used a weird code path that skipped parts of the linter. As of a couple weeks ago, everything should work.

src/sqlfluff/rules/L028.py
```diff
@@ -58,68 +78,171 @@ def _lint_references_and_aliases(
...
+    fix_inconsistent_to: Optional[str],
+) -> Optional[List[LintResult]]:
+    """Iterate through references and check consistency."""
+    # How many aliases are there? If more than one then abort.
+    if len(table_aliases) > 1:
+        return None
+    # A buffer to keep any violations.
+    violation_buff: List[LintResult] = []
+    col_alias_names: List[str] = [c.alias_identifier_name for c in col_aliases]
+    table_ref_str: str = table_aliases[0].ref_str
```
barrywhart, line 135: Is it possible that `table_aliases` has length 0? If so, this would crash with an `IndexError`.

OTooleMichael: Not sure but will guard

test/fixtures/rules/std_rule_cases/L028.yml
OTooleMichael: Ill add a test for this.

src/sqlfluff/rules/L028.py
OTooleMichael: To what should we change the text?

I'd probably leave force enabled flag for now (issue for a bigger release) its a more serious change that could be breaking and probably outside the thought level I can give this before pushing out 0.11.1

tunetheweb: Sounds good.

At the moment the text just mentions BigQuery, but it's three dialects this affects so think we should mention the other two: "This rule is disabled by default for BigQuery, Hive and Redshift due to their use of..."

Or alternatively should say "This rule is disabled by default for dialects like BigQuery due to their use of..." to avoid having to keep it in sync in future, but do think listing all three is more friendly to users.

test/fixtures/rules/std_rule_cases/L028.yml
OTooleMichael: Test added. for fail and pass w/ noqa.

Later I'd like to revisit this / make another rule for lateral column references - they cause huge problems.

Rule like **lateral self references must begin with __self_**

then there would be some interaction with this and other rules.

tunetheweb (approved):

PR #2763, opened 2022-03-02
Snowflake: Add `CONNECT BY` clause

jpy-git:

Adds the CONNECT BY clause to Snowflake dialect to allow hierarchical queries.
https://docs.snowflake.com/en/sql-reference/constructs/connect-by.html

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_spark3.py
+++ b/src/sqlfluff/dialects/dialect_spark3.py
@@ -222,6 +222,26 @@ spark3_dialect.replace(
         "RLIKE",
         "REGEXP",
     ),
+    SelectClauseSegmentGrammar=Sequence(
+        "SELECT",
+        OneOf(
+            Ref("TransformClauseSegment"),
+            Sequence(
+                Ref(
+                    "SelectClauseModifierSegment",
+                    optional=True,
+                ),
+                Indent,
+                Delimited(
+                    Ref("SelectClauseElementSegment"),
+                    allow_trailing=True,
+                ),
+            ),
+        ),
+        # NB: The Dedent for the indent above lives in the
+        # SelectStatementSegment so that it sits in the right
+        # place corresponding to the whitespace.
+    ),
 )
 
 spark3_dialect.add(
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -292,6 +293,17 @@
         optional=True,
     ),
     TemporaryTransientGrammar=OneOf(Ref("TemporaryGrammar"), "TRANSIENT"),
+    BaseExpressionElementGrammar=OneOf(
+        # Allow use of CONNECT_BY_ROOT pseudo-columns.
+        # https://docs.snowflake.com/en/sql-reference/constructs/connect-by.html#:~:text=Snowflake%20supports%20the%20CONNECT_BY_ROOT,the%20Examples%20section%20below.
+        Sequence("CONNECT_BY_ROOT", Ref("ColumnReferenceSegment")),
```
jpy-git, line 299: N.B. the level_expression pseudo-columns are pretty poorly documented in the docs.
As far as I can see there is:
- `level` which just acts as a standard `ColumnReferenceSegment`
- `connect_by_root` which has an example in the docs and is handled here
- `connect_by_path` mentioned in [this](https://docs.snowflake.com/en/user-guide/queries-hierarchical.html#:~:text=The%20CONNECT%20BY%20syntax%20supports%20convenient%20pseudo%2Dcolumns%20such%20as%20LEVEL%2C%20CONNECT_BY_ROOT%2C%20and%20CONNECT_BY_PATH) doc section. However it does not seem to work the same way as either `level` or `connect_by_root` and there's zero explanation in the docs so I've omitted it from this PR. Users will likely use the alternative `SYS_CONNECT_BY_PATH` function instead (which we already parse) since that has an actual example.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -407,6 +419,33 @@
...
+    """A `CONNECT BY` clause.
+
+    https://docs.snowflake.com/en/sql-reference/constructs/connect-by.html
+    """
+
+    type = "connectby_clause"
+    match_grammar = StartsWith(Sequence("START", "WITH"))
+    parse_grammar = Sequence(
+        "START",
+        "WITH",
```
tunetheweb, line 433: This doesn't add anything and can be replaced with this:

```suggestion
    match_grammar = Sequence(
        "START",
        "WITH",
```

tunetheweb (changes requested):
Looks good. One small nit

PR #2770, opened 2022-03-02
Snowflake: Add dollar quoted string literal

jpy-git:

Snowflake allows dollar quoted string literals (similar to postgres but no tags). I've updated `QuotedLiteralSegment` to allow these and keep consistent with the postgres approach.

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -120,13 +120,6 @@
         name="warehouse_size",
         type="warehouse_size",
     ),
-    DoubleQuotedLiteralSegment=NamedParser(
-        "double_quote",
-        CodeSegment,
-        name="quoted_literal",
-        type="literal",
-        trim_chars=('"',),
-    ),
```
jpy-git, line 129: This isn't part of Snowflake and isn't used anywhere in the dialect so cleaning up.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -505,7 +513,7 @@ class FunctionDefinitionGrammar(BaseSegment):
     type = "function_definition"
     match_grammar = Sequence(
         "AS",
-        OneOf(Ref("QuotedLiteralSegment"), Ref("DollarQuotedLiteralSegment")),
```
jpy-git, line 508: `DollarQuotedLiteralSegment` isn't anywhere else in the code base. We used to have it in Postgres, but was since combined with `QuotedLiteralSegment` to simplify the dialect. This is probably just a holdover from when Snowflake inherited from Postgres.

tunetheweb (approved):
LGTM!

PR #2781, opened 2022-03-04
Snowflake: Add MATCH_RECOGNIZE clause

jpy-git:

Adds MATCH_RECOGNIZE clause as specified in the [docs](https://docs.snowflake.com/en/sql-reference/constructs/match_recognize.html).

N.B. The pattern grammar is pretty complicated so initially I've covered the basic use cases and added a TODO note on adding operators in a future PR.

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3840,3 +4003,46 @@ class CallStatementSegment(BaseSegment):
...
+        Sequence("ORDER", "BY"),
+        terminator=OneOf(
+            "LIMIT",
+            "HAVING",
+            "QUALIFY",
+            # For window functions
+            "WINDOW",
+            Ref("FrameClauseUnitGrammar"),
+            "SEPARATOR",
+            "MEASURES",
```
jpy-git, line 4023: This basically just adds `MEASURES` to the `ORDER BY` terminator list

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -221,6 +222,84 @@
...
+                "*",
+                SymbolSegment,
+                name="star",
+                type="star",
+            ),
+            StringParser(
+                "?",
+                SymbolSegment,
+                name="question_mark",
+                type="question_mark",
```
tunetheweb, line 243: TIL I learned you could use these Parsers directly in Segments! Makes sense but never thought about doing that before.

Saying that, we have segments for each of these so why not do this as easier to read?:

```suggestion
            Ref("PositiveSegment"),
            Ref("StarSegment"),
            Ref("QuestionMarkSegment"),
```

Note the last one will need to be added (you can copy from BigQuery). Also `PositiveSegment` has a different type so if that matters you could add a new one.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -221,6 +222,84 @@
...
+            ),
+        ),
+        # To put a quantifier into “reluctant mode”
+        StringParser(
+            "?",
+            SymbolSegment,
+            name="question_mark",
+            type="question_mark",
+            optional=True,
+        ),
```
tunetheweb, line 273: Again why not use ` `Ref("QuestionMarkSegment")`? Especially as using this twice now.

src/sqlfluff/dialects/dialect_snowflake.py
tunetheweb: You can insert it like this:

```
    match_grammar = ansi_dialect.get_segment(
        "OrderByClauseSegment"
    ).match_grammar.copy()
    match_grammar.terminator = (
        match_grammar.terminator.copy(
            insert=["MEASURES"],
        )
    )
```

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3840,3 +4015,46 @@ class CallStatementSegment(BaseSegment):
...
+            "HAVING",
+            "QUALIFY",
+            # For window functions
+            "WINDOW",
+            Ref("FrameClauseUnitGrammar"),
+            "SEPARATOR",
+            "MEASURES",
+        ),
+    )
+    parse_grammar = Sequence(
```
tunetheweb, line 4038: Similar to https://github.com/sqlfluff/sqlfluff/pull/2784#discussion_r819567281 I think we should replace this with:

```py
parse_grammar = ansi_dialect.get_segment("OrderByClauseSegment").parse_grammar.copy()
```

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_snowflake.py
jpy-git: Wasn't aware of this one from bigquery! will replace this thanks 😄

src/sqlfluff/dialects/dialect_snowflake.py
jpy-git: Think I was a bit concerned that the rules maybe moving things about since patterns don't follow the same formatting as expressions, let me have a play

src/sqlfluff/dialects/dialect_snowflake.py
jpy-git: Good shout, will do!

src/sqlfluff/dialects/dialect_snowflake.py
jpy-git: Nice will use this in #2784 as well 😄

jpy-git: FYI the insert expects a list of references so I had to use `Ref.keyword` to get this to work:
```python
    match_grammar = ansi_dialect.get_segment(
        "OrderByClauseSegment"
    ).match_grammar.copy()
    match_grammar.terminator = (
        match_grammar.terminator.copy(
            insert=[Ref.keyword("MEASURES")],
        )
    )
```

tunetheweb: Yeah did wonder if you'd have to do that.

src/sqlfluff/dialects/dialect_snowflake.py
tunetheweb: Sorry don't understand. The `Ref`s are exactly the same as the code that's currently in there?

tunetheweb (approved):
LGTM once tests pass

PR #2795, opened 2022-03-06
L042 Autofix - (CTE hoisting) Super Working PR

OTooleMichael:
Adding Fix for Subquery -> CTE by hoisting subqueries to CTEs.

This draft currently seems to fight against the recursion involved in _eval.
The test actually pass (minus one I put to show the fight against recursion). After each cycle we readdress the same tree (having been mutated by us last cycle) - however the new tree (although valid to parser matching) has not been correctly renested.

No side effects predicted.

Goals and cases:
- [x] hoists subquery
- [x] hoists multiple subqueries (from different joins)
- [x] puts subqueries in the correct order
- [x] cleans up aliases as needed
- [x] hoists when select already has CTEs
- [x] unwinds subquery that is a CTE
- [x] fixes / manages if CTEs might have clashing names

Non goals
- brings deeply nested subqueries all the way to the top
- ?identifies that two subqueries are actually the same.

### Pull Request checklist


src/sqlfluff/core/rules/functional/segments.py
```diff
@@ -143,6 +149,19 @@ def __getitem__(self, item):  # type: ignore
         else:
             return result
 
+    def dive_last(self, max_depth=1):
```
barrywhart, line 152:
```suggestion
    def dive_last(self, max_depth: int=1) -> BaseSegment:
```

test/fixtures/rules/std_rule_cases/L042.yml
```diff
@@ -30,17 +156,38 @@ with_fail:
...
+    with prep_1 as (
+        with d as (
+            select x, z from b
+        )
+        select * from d
+    )
+    select
+        a.x, a.y, b.z
+    from a
+    join prep_1 using (x)
```
OTooleMichael, line 169: @barrywhart How should this Fix handle this.
Nested Subqueries are silly. But that is not the scoped name of this Fix.

PS can/should I do a rule fo nested WITH?

test/fixtures/rules/std_rule_cases/L042.yml
```diff
@@ -7,17 +7,143 @@ select_fail:
...
+    ) as b on (a.x = b.x)
+  fix_str: |
+    with p_cte as (
+      select 1 as x, 2 as z
+      UNION ALL
+      select 1 as x, 2 as z
+    ),
+    c as (
+        select x, z from p_cte
+      ),
```
OTooleMichael, line 66: Indentation here comes back as it was (the indentation of the origin was correct for a subquery, but is now incorrect) will another pass/Rule pick this up?

test/fixtures/rules/std_rule_cases/L042.yml
barrywhart: I think nested `WITH` could be useful in some cases. Not that you technically need it, but for modularity. For example, outer CTE `a` is built using nested CTEs `a1`, `a2`, and `a3`, then outer CTE `b` is built using nested CTEs `b1` and `b2`.

test/fixtures/rules/std_rule_cases/L042.yml
barrywhart: Yes, if you run `sqlfluff lint`, the rules work "together" to provide reasonable output. We've discussed this for other rules, and for now, we generally agree it's better to let other rules handle it rather than have each rule doing "all the things". (We may change this someday, but there's no clear timeline for doing that. The current way works pretty well in practice.)

test/dialects/dialects_test.py
```diff
@@ -48,6 +48,15 @@ def test__dialect__base_file_parse(dialect, file):
...
 
+@pytest.mark.parametrize("dialect,file", parse_success_examples)
+def test__dialect__base_broad_fix(dialect, file, raise_critical_errors_after_fix):
+    """For given test examples, check successful parsing."""
+    raw = load_file(dialect, file)
+    # Load the right dialect
+    config = FluffConfig(overrides=dict(dialect=dialect))
+    Linter(config=config).lint_string_wrapped(raw, fix=True)
+
+
```
OTooleMichael, line 59: @barrywhart  I think this is an improvement over the bash script you added recently (also edited below).

It does basically the same thing but can actually be debugged in python. It could probably also be sped up with caching I guess.

As mentioned in the function above from which I derived it, there is way too much logging - can this be suppressed? (its not needed for the parsing for example)

test/fixtures/dialects/ansi/with_nested_in_with_statement.sql
```diff
@@ -1,5 +1,4 @@
-(
-    WITH mycte2 AS (
+(WITH mycte2 AS (
```
OTooleMichael, line 1: L053 fixes this by removing the brackets which I guess leads to `Whitespace(), Whitespace()` which is unparsable (I think) - Fixing this bug which is in L053 seemed out of scope so I just removed the padding to allow this case to pass for now.

test/test_rules_against_dialect_fixtures.sh
```diff
@@ -3,6 +3,11 @@
 echo "This script will run 'sqlfluff fix -f' on all our dialect fixtures"
 echo "to look for critical errors."
 echo "WARNING this will change the fixtures so do not commit these changes!"
+function rest_git_changes(){
+    echo "Attempting to reset Git Changes"
+    # If this doesnt work for some reason just ignore it
+    git checkout -- test/fixtures/dialects/ || true
+}
```
OTooleMichael, line 10: I think going with the python version I purposed above is prolly better, but this little bonus might save a little shock for someone like me who totally ignored the warnings before running this the first time.

src/sqlfluff/rules/L042.py
```diff
@@ -48,44 +70,281 @@ class Rule_L042(BaseRule):
...
+            )
+        )
+        for cte in root_select.children(is_type("common_table_expression")):
+            assert isinstance(cte, CTEDefinitionSegment), "TypeGaurd"
+            ctes.insert_cte(cte)
+
+    lint_results: List[LintResult] = []
+    CTESegment = Seg(CTEDefinitionSegment)
+    TableExpressionSeg = Seg(TableExpressionSegment)
+    TableReferenceSeg = Seg(TableReferenceSegment)
```
OTooleMichael, line 149: Here I felt the issue again, of not being able to just lex & pars the new desired input. `change[str] --> Segments`

Basically TableExpressionSeg and TableReferenceSegment have been subclassed based on Dialect. I was concerned about strange behaviour when adding a segment explicitly using the ansi class when the query being processed is using/expecting the BigQuery version of the class everywhere.

src/sqlfluff/rules/L042.py
```diff
@@ -48,44 +70,281 @@ class Rule_L042(BaseRule):
...
+                    TableReferenceSeg(
+                        segments=(
+                            CodeSegment(
+                                raw=alias_name,
+                                name="naked_identifier",
+                                type="identifier",
+                            ),
+                        )
+                    ),
                 )
```
OTooleMichael, line 181: Without all this nesting I was corrupting the parse tree - PS nice catch with the new tests @barrywhart

barrywhart: I have spent a **lot** of time staring at `sqlfluff parse` output the past few weeks!!

test/dialects/dialects_test.py
barrywhart: Can you tell me more about the logging? Python logging supports different levels. When you run the tests, logging and printing are suppressed/captured by default.

test/fixtures/dialects/ansi/with_nested_in_with_statement.sql
barrywhart: Can you capture enough detail and create an issue about L053? Sounds like something that should be addressed soon-ish.

barrywhart: BTW, the PR I merged today included some L053 fixes. Is it possible I already addressed the problem you were seeing?

test/dialects/dialects_test.py
barrywhart: BTW, @tunetheweb wrote the bash script, not me.

src/sqlfluff/rules/L042.py
barrywhart: It would be cool to do this. Interested in taking a look at that? I may have time, maybe not. (I'm starting a new job next week. We'll see how it affects my schedule.)

OTooleMichael: I will try if I get a chance :)

test/fixtures/dialects/ansi/with_nested_in_with_statement.sql
OTooleMichael: Yea loads of info, my edit to the cirtical errors test suite gives a detailed desc of "valid" ---> "invalid"

test/dialects/dialects_test.py
OTooleMichael: Fair enough, I guess I was tired by the time I was writing this and Id just seen your name under so many of the blame lines :)

Pytest suppresses logging but it captures it all and spits it out when a test fails - (only the output related to that test). So using this over the bash means we --fix all the files but actually reject and raise and therefore fail and get output about the SQL + rule causing the "Critical error"

src/sqlfluff/core/rules/base.py
```diff
@@ -623,6 +624,10 @@ def crawl(
         return vs, raw_stack, fixes, memory
 
     # HELPER METHODS --------
+    @staticmethod
+    def _log_critical_errors(error: Exception):  # pragma: no cover
+        """This method is monkey patched into a "raise" for certain tests."""
+        pass
```
OTooleMichael, line 631: Only here to all it to be patched during certain tests letting us raise during the test suite

test/dialects/dialects_test.py
tunetheweb: I’m totally cool with changing this from bash to Python. We talked about it at the time as would also allow Wndows to run it.

It was as a quick hack to make use of our many fixtures files (which were previously only used for parse testing) to also test rules - initially`lint` and then, as of #2830 merged last might, also `fix`. 

However, from a quick look on my phone, it looks like you’ve only implemented the `lint` version? As I say #2830 includes `fix`. Just need to be careful as that changes the file, so need to revert the change at the end. And also ensure no other test is running at the same time.

I also think this is a big enough change so it should be a separate PR.

src/sqlfluff/rules/L042.py
barrywhart: I think there are some existing parser or dialect tests that already do this in isolation, so it may be as easy as calling those same functions on context.dialect. 🤞

src/sqlfluff/core/rules/functional/segments.py
```diff
@@ -175,3 +181,15 @@ def select(
             if select_if is None or select_if(seg):
                 buff.append(seg)
         return Segments(*buff, templated_file=self.templated_file)
+
+    def iterate_segments(
+        self,
+        predicate: Optional[PredicateType] = None,
+    ) -> Iterable["Segments"]:
+        """Loop over eaach element as a fresh Segments."""
```
barrywhart, line 189:
```suggestion
        """Loop over each element as a fresh Segments."""
```

src/sqlfluff/rules/L042.py
```diff
@@ -48,44 +70,281 @@ class Rule_L042(BaseRule):
...
+            continue
+
+        return True
+
+    return False
+
+
+def _find_nested_subqueries(
+    select: Segments,
+) -> Generator[Tuple[str, Segments, Segments, BaseSegment], None, None]:
```
barrywhart, line 239: Think we could create a `NamedTuple` declaration for this? Could be a bit more readable.

src/sqlfluff/rules/L042.py
```diff
@@ -48,44 +70,281 @@ class Rule_L042(BaseRule):
...
+    output_select = root_select
+    if is_with:
+        output_select = root_select.children(
+            is_type(
+                "set_expression",
+                "select_statement",
+            )
+        )
+        for cte in root_select.children(is_type("common_table_expression")):
+            assert isinstance(cte, CTEDefinitionSegment), "TypeGaurd"
```
barrywhart, line 143:
```suggestion
            assert isinstance(cte, CTEDefinitionSegment), "TypeGuard"
```

src/sqlfluff/rules/L042.py
```diff
@@ -48,44 +70,281 @@ class Rule_L042(BaseRule):
...
+        res = LintResult(
+            anchor=subquery,
+            description=f"{parent_type} clauses should not contain "
+            "subqueries. Use CTEs instead",
+            fixes=[],
+        )
+        lint_results.append(res)
+
+    if ctes.has_duplicates() or is_recursive:
+        # If we have duplicate CTE names just don't fix anything
```
barrywhart, line 193: Should we return a `LintResult` here to warn about the issue (not try to fix it, though)?

src/sqlfluff/rules/L042.py
```diff
@@ -48,44 +70,281 @@ class Rule_L042(BaseRule):
...
+        )
+    )
+    # Add fixes to the last result only
+    res = lint_results.pop()
+    res.fixes = [LintFix.replace(root_select[0], edit_segments=[new_select])]
+    lint_results.append(res)
+    return lint_results
+
+
+def _is_child(maybe_parent: Segments, maybe_child: Segments) -> bool:
```
barrywhart, line 214: This function looks similar to `BaseSegment.path_to()`. Could you use that instead?

https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/core/parser/segments/base.py#L866L892

src/sqlfluff/rules/L042.py
```diff
@@ -48,44 +70,281 @@ class Rule_L042(BaseRule):
...
+            continue
+        # Type, parent_select, parent_sequence
+        yield (parent_type, select, this_seg, table_expression_el[0])
+
+
+def _get_casing_preference(root_select: Segments):
+    first_keyword = root_select.recursive_crawl("keyword", recurse_into=False).first()[
+        0
+    ]
+    if first_keyword.raw[0] == first_keyword.raw[0].lower():
```
barrywhart, line 279:
```suggestion
    if first_keyword.raw[0].islower():
```

src/sqlfluff/rules/L042.py
```diff
@@ -48,44 +70,281 @@ class Rule_L042(BaseRule):
...
+        self.name_idx = 0
+
+    def has_duplicates(self) -> bool:
+        return len(set(self.used_names)) != len(self.used_names)
+
+    def insert_cte(self, cte: CTEDefinitionSegment):
+        """Add a new CTE to the list as late as possible but before all its parents."""
+        output: List[BaseSegment] = []
+        id_seg = cte.get_identifier()
+        print(id_seg, id_seg.raw, id_seg.name)
```
barrywhart, line 300: Looks like debugging code -- please remove or change to logging.

src/sqlfluff/rules/L042.py
```diff
@@ -48,44 +70,281 @@ class Rule_L042(BaseRule):
...
+
+        self.used_names.append(cte_name)
+        # This should still have the position markers of its true position
+        inbound_subquery = Segments(cte).children().last()
+        for el in self.ctes:
+            if cte in output:  # pragma: no cover
+                output.append(el)
+                continue
+            if _is_child(Segments(el).children().last(), inbound_subquery):
+                output.append(cte)
```
barrywhart, line 313: The loop body is confusing, with the `continue` and two different `output.append()` calls. Could this be cleaned up a bit? Seems like the `continue` could be replaced using `else` or something like that. (I like `continue` better when the loop body is pretty long.)

src/sqlfluff/rules/L042.py
```diff
@@ -48,44 +70,281 @@ class Rule_L042(BaseRule):
...
+    def __init__(self) -> None:
+        self.used_names: List[str] = []
+        self.ctes: List[BaseSegment] = []
+        self.name_idx = 0
+
+    def has_duplicates(self) -> bool:
+        return len(set(self.used_names)) != len(self.used_names)
+
+    def insert_cte(self, cte: CTEDefinitionSegment):
+        """Add a new CTE to the list as late as possible but before all its parents."""
```
barrywhart, line 297: Could you add some high-level explanatory comments to this function, explaining the algorithm/the "why"? The code is pretty detailed, but I'm struggling to see the overall algorithm.

src/sqlfluff/rules/L042.py
```diff
@@ -48,44 +70,281 @@ class Rule_L042(BaseRule):
...
+            cte_segments = cte_segments + [
+                cte,
+                SymbolSegment(",", name="comma", type="comma"),
+                NewlineSegment(),
+            ]
+        return cte_segments[:-2]
+
+
+def _segmentify(input_el: str, casing: str) -> BaseSegment:
+    """Apply casing an convert strings to Keywords."""
```
barrywhart, line 345:
```suggestion
    """Apply casing and convert strings to Keywords."""
```

src/sqlfluff/rules/L042.py
OTooleMichael: We do line 194 will have a fixless Lint Result

src/sqlfluff/rules/L042.py
OTooleMichael: I think they are quite different that other function does a recursive crawl. This just quickly checks the position markers

barrywhart (approved):
Great work!

barrywhart:
@OTooleMichael: After we merged your PR, I noticed crashes in other areas of the linter. We decided to revert your PR until these issues are solved.

I haven't looked closely, but as a guess, perhaps it's related to the new/unusual technique you're using for creating some of the segments? Let us know what you find -- this could be a tricky issue.

To reproduce the issue, try running `sqlfluff lint` on this query:
```
SELECT *
FROM A_TABLE
INNER JOIN (
    SELECT margin
    FROM B_TABLE
) USING (SOME_COLUMN)
```

This is the error I saw:
```
(sqlfluff-3.10.1) ➜  sqlfluff git:(main) ✗ sqlfluff lint test.sql
CRITICAL   [L032] Applying rule L032 threw an Exception: not enough values to unpack (expected 2, got 1)                                                                                                
  File "/Users/bhart/dev/sqlfluff/src/sqlfluff/rules/L032.py", line 101, in _eval
    table_a, table_b = select_info.table_aliases
ValueError: not enough values to unpack (expected 2, got 1)
CRITICAL   [L046] Applying rule L046 threw an Exception:                                                                                                                                                
  File "/Users/bhart/dev/sqlfluff/src/sqlfluff/rules/L046.py", line 56, in _eval
AssertionError
WARNING    Unable to lint test.sql due to an internal error. Please report this as an issue with your query's contents and stacktrace below!                                                            
  File "/Users/bhart/dev/sqlfluff/src/sqlfluff/rules/L046.py", line 56, in _eval
AssertionError
```

OTooleMichael:
> @OTooleMichael: After we merged your PR, I noticed crashes in other areas of the linter. We decided to revert your PR until these issues are solved.
> 
> 
> 
> I haven't looked closely, but as a guess, perhaps it's related to the new/unusual technique you're using for creating some of the segments? Let us know what you find -- this could be a tricky issue.
> 
> 
> 
> To reproduce the issue, try running `sqlfluff lint` on this query:
> 
> ```
> 
> SELECT *
> 
> FROM A_TABLE
> 
> INNER JOIN (
> 
>     SELECT margin
> 
>     FROM B_TABLE
> 
> ) USING (SOME_COLUMN)
> 
> ```
> 
> 
> 
> This is the error I saw:
> 
> ```
> 
> (sqlfluff-3.10.1) ➜  sqlfluff git:(main) ✗ sqlfluff lint test.sql
> 
> CRITICAL   [L032] Applying rule L032 threw an Exception: not enough values to unpack (expected 2, got 1)                                                                                                
> 
> 
>   File "/Users/bhart/dev/sqlfluff/src/sqlfluff/core/rules/base.py", line 535, in crawl
> 
>     res = self._eval(context=context)
> 
>   File "/Users/bhart/dev/sqlfluff/src/sqlfluff/rules/L032.py", line 101, in _eval
> 
>     table_a, table_b = select_info.table_aliases
> 
> ValueError: not enough values to unpack (expected 2, got 1)
> 
> CRITICAL   [L046] Applying rule L046 threw an Exception:                                                                                                                                                
> 
> 
>   File "/Users/bhart/dev/sqlfluff/src/sqlfluff/core/rules/base.py", line 535, in crawl
> 
>     res = self._eval(context=context)
> 
>   File "/Users/bhart/dev/sqlfluff/src/sqlfluff/rules/L046.py", line 56, in _eval
> 
>     assert context.segment.pos_marker
> 
> AssertionError
> 
> WARNING    Unable to lint test.sql due to an internal error. Please report this as an issue with your query's contents and stacktrace below!                                                            
> 
> To hide this warning, add the failing file to .sqlfluffignore
> 
> 
>   File "/Users/bhart/dev/sqlfluff/src/sqlfluff/core/rules/base.py", line 535, in crawl
> 
>     res = self._eval(context=context)
> 
>   File "/Users/bhart/dev/sqlfluff/src/sqlfluff/rules/L046.py", line 56, in _eval
> 
>     assert context.segment.pos_marker
> 
> AssertionError
> 
> ```

But how did this also get past the tests?

barrywhart:
@OTooleMichael 

>But how did this also get past the tests?

The typical YAML tests run the rule in isolation. We do have other tests that run combinations of rules, e.g.
* The tests in `./test/fixtures/linter/autofix` run specified combinations of rules (controlled by the file `test-config.yml` in each test directory).
* The "Rules critical errors tests" tests run *all* the rules on a bunch of SQL files. I think we got unlucky in that the issue above did not arise in any of those files. In order for it to happen, they would've had to trigger L042, perhaps followed by some other rule.

We've recently been adding more checks to try and detect issues with rules that may affect downstream rules, e.g.:
* "parse after fix" check
* Detect when a rule returns conflicting fixes (basically, multiple fixes with the same anchor)

Whatever is happening here is something we aren't yet checking for. Once we figure that out, it would be good to add a check for that as well. Life on the 🩸 edge, argh. Please don't feel bad about this issue -- it seems like a tricky one, and if it arose because of the new fix code, that's because you were brave enough to develop a pretty intricate fix routine that nobody has tried yet!! 💪 We'll figure this out, it's just a 🪨  in the road.

PR #2801, opened 2022-03-06
Exasol: Add lua adapter scripts (7.1+)

sti0:

Allow lua adapter scripts which are introduced in 7.1. Furthermore I removed the `optional` flag for the script language definition since this is mandatory in 7.1+.

May break scripts which not defined a language (<7.1). This could easy fixed by adding a script language parameter.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -3657,15 +3657,15 @@ class CreateAdapterScriptStatementSegment(BaseSegment):
         Sequence(
             "CREATE",
             Ref("OrReplaceGrammar", optional=True),
-            OneOf("JAVA", "PYTHON", Ref("SingleIdentifierGrammar"), optional=True),
+            OneOf("JAVA", "PYTHON", "LUA", Ref("SingleIdentifierGrammar")),
             "ADAPTER",
             "SCRIPT",
         )
     )
     parse_grammar = Sequence(
```
tunetheweb, line 3665: Similar to my comment on other PR, should we merge these?

A separate `match_grammar` section should be added when there’s an easy start and terminator clause, to reduce the amount that needs to be parsed. But if it just repeats the first part of the `parse_grammar` then it doesn’t do anything and this change shows that it adds work, when you want to make changes.

tunetheweb (changes requested):
Could we add a LUA test case to make sure it doesn’t regress in future?

Also no concerns of < 7.1 users not using that optional clause still? Or is it arguably best practice to include to anyway so happy to make non-optional?

sti0:
> Also no concerns of < 7.1 users not using that optional clause still? Or is it arguably best practice to include to anyway so happy to make non-optional?

Like I said, it's easy to fix when adding a language keyword (this is still the prefered way). I think that adapter scripts are rarely used so this should not have great impact. Furthermore I think that we should add support for the latest version since 7.0.x will be EOL in mid 2022.

> Could we add a LUA test case to make sure it doesn’t regress in future?
I don't have a working test case since I don't used this functionallity before. There is no example in the docs. I will add a dummy one to make sure the parsing works as expected.

tunetheweb (approved):
LGTM assuming all tests pass!

PR #2803, opened 2022-03-06
L063: Remove redundant semi-colons and GO delimiters.

### LINKED ISSUE #2653: New rule proposal: avoid several semi-colons in a row.
### Search before asking

### Description

Multiple semi-colons are valid SQL and we even have a test case for this:

```sql
-- It's possible to have multiple semicolons between statements.
SELECT foo FROM bar;;
SELECT foo FROM bar;
;
;
SELECT foo FROM bar;
```

I think we should have a rule (or perhaps an enhancement to L052?) that removes these and replaces them with a single semi-colon.

### Use case

Above SQL should be changed to this:

```sql
SELECT foo FROM bar;
SELECT foo FROM bar;
SELECT foo FROM bar;
```

Where extraneous semi-colons (and whitespace surrounding them?) should be removed.

### Dialect

All

### Code of Conduct

jpy-git:

Closes #2653.

Couple things going on in this PR:
- Noticed that our current grammar wouldn't allow us to parse leading semi-colons/GO delimiters so I've updated the grammar accordingly and added the unit tests.
- Added a new rule that detects and fixes redundant semi-colons/GO delimiters: either occurring before the first statement or after another of the same delimiter.

The idea is that this rule removes redundant delimiters and L052 can focus on adding missing delimiters and repositioning them.

I also included GO delimiters in T-SQL for this rule as it's exactly the same problem we're solving. It raises a couple points:
- I'm happy to split the GO fix into it's own rule if you feel it makes the rule clearer?
- Potentially we want a GO version of L052 in the future to handle their positioning and add a trailing GO in? Probably a separate rule with similar logic to avoid making the code too long.

No

### Pull Request checklist

  - Other.


test/fixtures/rules/std_rule_cases/L063.yml
```diff
@@ -0,0 +1,111 @@
...
+  fix_str: |
+    select foo from bar;
+
+test_fail_newlines:
+  fail_str: "select foo from bar;\n;"
+  fix_str: "select foo from bar;\n"
+
+test_fail_block_comment:
+  fail_str: "select foo from bar;/*jpy-git is cool*/\n;"
+  fix_str: "select foo from bar;/*jpy-git is cool*/\n"
```
OTooleMichael, line 19: This one is interesting I don't think empty statements are valid SQL with most engines. So neither pre nor post fix would this run.

OTooleMichael:
Nice - This is a good rule to have. We often want to do a poor mans splitting of statements (where a full parse is too heavy) and this would really help.

Could we extend the Rule to be more general. In that it allows / has an opinion about "always" or "never" having terminating ";"

Feels like always terminal or never terminate are rules that we might want and would best fit under this rule (otherwise the rules might "compete" when being applied). This rule acts a bit like "consistent" which I think is a bit of mire

jpy-git:
> Could we extend the Rule to be more general. In that it allows / has an opinion about "always" or "never" having terminating ";"

L052

OTooleMichael:
Will you move this into L052 or why did you close?

Issue #1936, opened 2021-11-19
Add support for Spark3 merge into syntax


anzelpwj:
This is a bit farther afield, but if you are using Delta tables in Spark (e.g. if you're working in Databricks) there is an option to have an upsert command with `MERGE INTO`.
- https://docs.delta.io/latest/delta-update.html#upsert-into-a-table-using-merge
- https://docs.databricks.com/delta/delta-update.html#merge-examples

I'd like to have the ability to parse these commands. This may be a larger discussion since this appears to only be for delta tables and not base SparkSQL, so if it makes more sense to have a Spark3-Delta dialect instead to add this sort of functionality, that would be fine too.

## Expected Behaviour

From the Delta.io docs, we have the example

```sql
MERGE INTO events
USING updates
ON events.eventId = updates.eventId
WHEN MATCHED THEN
  UPDATE SET events.data = updates.data
WHEN NOT MATCHED
  THEN INSERT (date, eventId, data) VALUES (date, eventId, data)
```

I'd like this to be parseable.

## Observed Behaviour

```
echo "MERGE INTO events USING updates ON events.eventId = updates.eventId WHEN MATCHED THEN UPDATE SET events.data = updates.data WHEN NOT MATCHED THEN INSERT (date, eventId, data) VALUES (date, eventId, data);" | sqlfluff parse --dialect spark3 -
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    base:
[L:  1, P:  1]      |        unparsable:                                           !! Expected: 'StatementSegment'
[L:  1, P:  1]      |            raw:                                              'MERGE'
[L:  1, P:  6]      |            whitespace:                                       ' '
[L:  1, P:  7]      |            raw:                                              'INTO'
[L:  1, P: 11]      |            whitespace:                                       ' '
[L:  1, P: 12]      |            raw:                                              'events'
[L:  1, P: 18]      |            whitespace:                                       ' '
[L:  1, P: 19]      |            raw:                                              'USING'
[L:  1, P: 24]      |            whitespace:                                       ' '
[L:  1, P: 25]      |            raw:                                              'updates'
[L:  1, P: 32]      |            whitespace:                                       ' '
[L:  1, P: 33]      |            raw:                                              'ON'
[L:  1, P: 35]      |            whitespace:                                       ' '
[L:  1, P: 36]      |            raw:                                              'events'
[L:  1, P: 42]      |            raw:                                              '.'
[...]

PR #2797, opened 2022-03-06
Fix Postgres VALUES, make Spark3 VALUES consistent

jpy-git:
### Brief summary of the change made

Couple things fixed here:
- Fixes regression in Postgres VALUES clause caused by copying of Spark3 grammar in #2590. Expression segments are all that are required in a values clause. Spark grammar didn't account for them and tried to define the individual literal/function/... types.
- Spark3 VALUES grammar is different from every other grammar. This is because I fixed these in #2404 and Spark3 later got overwritten unnecessarily.

### Are there any other side effects of this change that we should be aware of?
No

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #2807, opened 2022-03-07
Add initial `MERGE` syntax to most dialects

### LINKED ISSUE #1125: Add Merge Statement to ANSI Dialect (was: BigQuery dialect)
Bigquery currently has the functionality to run [merge statments](https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#merge_statement)

Hoping to add the necessary segments to allow for the parsing of BigQuery Merge statements.

This will allow the parsing of this type of SQL, opening up rule checks for rules that require parsing (like L003).

PS: This is my first issue submission here. Let me know if I am missing anything!

PhilippLange:
Introduce MERGE Statement to ANSI dialect.
Modified TSQL, Snowflake, Exasol to reuse ANSI complaint sections.

Fixes #1125, makes progress on #1936.

-

### Pull Request checklist

  - Other.


PhilippLange:
What do I have to do to enable the changes I made in ANSI for Spark3?
Currently the parser results in:
```
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'MERGE INTO t USING
                       | a AS u ON ( a.a = a.b...'
```
I assume I have to add the keyword somewhere?

Spark3 follows mostly ANSI (some minore changes) - this would then most likely close #1936 (some dialect specific sections have to be added - e.g. AnyNumberOf() within the MergeMatchSegment() and optional MergeMatchedClauseSegment() / MergeNotMatchedClauseSegment()

tunetheweb:
I got this SQL to parse fine in Spark3 using this branch:

```sql
MERGE INTO logs1
USING newDedupedLogs
ON logs1.uniqueId = newDedupedLogs.uniqueId
WHEN MATCHED THEN
UPDATE SET a = b
WHEN NOT MATCHED THEN
INSERT (a) VALUES (b)
```

PhilippLange:
You're right, my parsing broke at the VALUES part (#2797 made some changes to ValuesClauseSegment which fixes it) - I hadn't rebased while asking above question...

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2304,6 +2304,110 @@ class InsertStatementSegment(BaseSegment):
...
 
+@ansi_dialect.segment()
+class MergeStatementSegment(BaseSegment):
+    """A `MERGE` statement."""
+
+    type = "merge_statement"
+    match_grammar = StartsWith(
+        Sequence("MERGE", "INTO"),
+    )
+    parse_grammar = Sequence(
```
tunetheweb, line 2315: Originally thought this adds no value as `match_grammar` is same as `parse_grammar`, but upon further digging I see it is needed to be separated. Let's make that clear.

```suggestion
    # Note separate `match_grammar` as overridden in other dialects.
    match_grammar = StartsWith(
        Sequence("MERGE", "INTO"),
    )
    parse_grammar = Sequence(
```

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2304,6 +2304,110 @@ class InsertStatementSegment(BaseSegment):
...
+        Ref("JoinOnConditionSegment"),
+        Ref("MergeMatchSegment"),
+    )
+
+
+@ansi_dialect.segment()
+class MergeMatchSegment(BaseSegment):
+    """Contains dialect specific merge operations.
+
+    hookpoint for dialect specific behavior
```
tunetheweb, line 2344: Nit:

```suggestion
    Hookpoint for dialect specific behavior
```

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2304,6 +2304,110 @@ class InsertStatementSegment(BaseSegment):
...
+class MergeInsertClauseSegment(BaseSegment):
+    """`INSERT` clause within the `MERGE` statement."""
+
+    type = "merge_insert_clause"
+    match_grammar = Sequence(
+        "INSERT",
+        Indent,
+        Ref("BracketedColumnReferenceListGrammar", optional=True),
+        Dedent,
+        Ref("ValuesClauseSegment", optional=True),
```
tunetheweb, line 2407: Should this be wrapped in another `Indent`/`Dedent`?

```suggestion
        Indent,
        Ref("ValuesClauseSegment", optional=True),
        Dedent,
```

src/sqlfluff/dialects/dialect_postgres_keywords.py
```diff
@@ -478,7 +478,7 @@ def get_keywords(keyword_list, keyword_type):
     ("MAXVALUE", "non-reserved"),
     ("MEASURES", "not-keyword"),
     ("MEMBER", "not-keyword"),
-    ("MERGE", "not-keyword"),
+    ("MERGE", "reserved"),
```
tunetheweb, line 481: I think this should be non-reserved:

```suggestion
    ("MERGE", "non-reserved"),
```

Also you need to update `MATCHED` to be a non-reserved keyword too.

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2304,6 +2304,110 @@ class InsertStatementSegment(BaseSegment):
     )
 
 
+@ansi_dialect.segment()
+class MergeStatementSegment(BaseSegment):
+    """A `MERGE` statement."""
+
+    type = "merge_statement"
+    match_grammar = StartsWith(
+        Sequence("MERGE", "INTO"),
```
tunetheweb, line 2313: For BigQuery (and T-SQL) the `INTO` is optional. Wonder if we should:
1. Override in those dialects
2. Have the `INTO` optional in ANSI
3. Define as a Grammar to make it easier to override in other dialects without having to copy and paste basically the whole grammar?

Think I'm veering towards 3. WDYT?

tunetheweb (changes requested):
Can we also add `MATCHED` as an unreserved keyword in BigQuery? With that and the optional `INTO` you get basic support in BigQuery (though there's more to add but that can be done in follow up PRs).

src/sqlfluff/dialects/dialect_ansi.py
PhilippLange: I agree with three.
One and two seem like workarounds.

src/sqlfluff/dialects/dialect_postgres_keywords.py
tunetheweb: Do you see the second point here @PhilippLange ?

> Also you need to update MATCHED to be a non-reserved keyword too.

src/sqlfluff/dialects/dialect_ansi.py
PhilippLange: done

src/sqlfluff/dialects/dialect_ansi.py
PhilippLange: agree, done

src/sqlfluff/dialects/dialect_ansi.py
PhilippLange: done

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -393,6 +393,9 @@
     DateTimeLiteralGrammar=Sequence(
         OneOf("DATE", "TIME", "TIMESTAMP", "INTERVAL"), Ref("QuotedLiteralSegment")
     ),
+    # hookpoint for other dialects
+    # e.g. BIGQUERY INTO is optional
```
tunetheweb, line 397: Nits:

```suggestion
    # Hookpoint for other dialects
    # e.g. INTO is optional in BIGQUERY
```

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2304,6 +2307,114 @@ class InsertStatementSegment(BaseSegment):
...
+@ansi_dialect.segment()
+class MergeMatchedClauseSegment(BaseSegment):
+    """The `WHEN MATCHED` clause within a `MERGE` statement."""
+
+    type = "merge_when_matched_clause"
+    match_grammar = Sequence(
+        "WHEN",
+        "MATCHED",
+        Indent,
+        "THEN",
```
tunetheweb, line 2369: For `NOT MATCHED` the `Indent` is after the `THEN` but here it's before. Can we settle on one?

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2304,6 +2307,114 @@ class InsertStatementSegment(BaseSegment):
...
+
+
+@ansi_dialect.segment()
+class MergeMatchedClauseSegment(BaseSegment):
+    """The `WHEN MATCHED` clause within a `MERGE` statement."""
+
+    type = "merge_when_matched_clause"
+    match_grammar = Sequence(
+        "WHEN",
+        "MATCHED",
```
tunetheweb, line 2367: Postgres, Snowflake, BigQuery all allow an Optional `AND` clause here (and similarly in NOT MATCHED). Could we add this (plus test cases)?

I think that would be also make this complete for BigQuery too for the examples here: https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#merge_examples so if we could add those examples as fixtures?

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2304,6 +2307,114 @@ class InsertStatementSegment(BaseSegment):
...
+
+
+@ansi_dialect.segment()
+class MergeUpdateClauseSegment(BaseSegment):
+    """`UPDATE` clause within the `MERGE` statement."""
+
+    type = "merge_update_clause"
+    match_grammar = Sequence(
+        "UPDATE",
+        Ref("SetClauseListSegment"),
```
tunetheweb, line 2398: Maybe some indents/dedents here?

```suggestion
        Indent,
        Ref("SetClauseListSegment"),
        Dedent,
```

tunetheweb (changes requested):
Almost there. Just a couple more pieces and this is good to merge and pretty complete for all languages!

src/sqlfluff/dialects/dialect_ansi.py
PhilippLange: I agree and have two futher suggestions, that are technically not Ansi but are supported by multiple dialects:
 - ```WHEN MATCHED DELETE``` (Postgres, Snowflake, Spark3, BigQuery)
 - multiple ```WHEN MATCHED / WHEN NOT MATCHED``` (Snowflake, Spark3, BigQuery)

WDYT?

Futhermore there are cases like Postgres ```WHEN NOT MATCHED DO NOTHING``` (and probably plenty more) that I will ignore since adding it to Ansi isn't right.

tunetheweb: Yeah, ANSI isn't intended to be ANSI:

https://docs.sqlfluff.com/en/stable/dialects.html#ansi
> This dialect doesn’t intend to be brutal in adhering to (and only to) the ANSI SQL spec (mostly because ANSI charges for access to that spec). It aims to be a representation of vanilla SQL before any other project adds their spin to it, and so may contain a slightly wider set of functions than actually available in true ANSI SQL.

So a bit of a judgement call as to what is common enough to put in ANSI (and then override in the **one or two** dialects that don't support that syntax) versus leaving out of ANSI (and then having to override it in **multiple** other dialects that do support this).

So we have flexibility here.

PhilippLange: I added the following three things to the ANSI Dialect:
 - ```WHEN MATCHED DELETE``` (Postgres, Snowflake, Spark3, BigQuery)
 - multiple ```WHEN MATCHED / WHEN NOT MATCHED``` (Snowflake, Spark3, BigQuery)
 - Optional ```AND``` Clause for ```WHEN MATCHED / WHEN NOT MATCHED```

tunetheweb: Sounds good. Thanks for that. Will give it one final review (hopefully) when tests are completed and then merge.

PhilippLange: No worries, thanks for guiding me!

tunetheweb (approved):
Awesome work! Long standing ask finally resolve so thanks for that.

PR #2809, opened 2022-03-07
Automated tests should fail if a lint fix introduces a parse error

barrywhart:
We recently modified SQLFluff to log a warning if a lint fix makes illegal changes to
the parse tree. This is the appropriate behavior. for _production_. However, in order
to catch new bugs in new or modified rules, we need to be more strict when running
automated tests. This PR changes the behavior during tests to raise an exception
rather than simply logging. This notifies rule authors of issues and forces them to
be addressed during PR review.

This PR also fixes some issues with L052 fixes. These had been fixed, but the fixes
were lost when we reverted a recent PR (to address other bugs).

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -1148,7 +1148,7 @@ class BaseSegment:
                     )
 
     @staticmethod
-    def _log_apply_fixes_check_issue(message, *args):
+    def _log_apply_fixes_check_issue(message, *args):  # pragma: no cover
         linter_logger.warning(message, *args)
 
     def iter_patches(self, templated_str: str) -> Iterator[FixPatch]:
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -1110,7 +1110,7 @@ class BaseSegment:
                     )
                 except ValueError:  # pragma: no cover
                     found_error = True
-                    linter_logger.warning(
+                    self._log_apply_fixes_check_issue(
                         "After fixes were applied, segment %r failed the "
                         "match() parser check. Fixes: %r",
                         segment,
@@ -1119,7 +1119,7 @@ class BaseSegment:
                 else:
                     if not match_result.is_complete():  # pragma: no cover
                         found_error = True
-                        linter_logger.warning(
+                        self._log_apply_fixes_check_issue(
                             "After fixes were applied, segment %r failed the "
                             "match() parser check. Result: %r Fixes: %r",
                             segment,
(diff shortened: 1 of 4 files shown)


tunetheweb (approved):
Looks good to me, but a couple of questions:

1. You've replaced a lot of `replace` with `create_after`. Should we limit `replace` to only work with one param?
2. If (when?) we change the default in production to fail, do we remove some of this?

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1084,13 +1084,13 @@ def apply_fixes(self, dialect, fixes):
...
             # Return the new segment with any unused fixes.
             return r, fixes
         else:
             return self, fixes
 
-    def _validate_segment_after_fixes(self, dialect, fixes_applied, segment):
+    def _validate_segment_after_fixes(self, rule_code, dialect, fixes_applied, segment):
         """Checks correctness of new segment against match or parse grammar."""
         found_error = False
         root_parse_context = RootParseContext(dialect=dialect)
```
OTooleMichael, line 1096: Are we checking the whole parse tree or how does RootParseContext work here? (just for general understanding)

barrywhart:
>You've replaced a lot of replace with create_after. Should we limit replace to only work with one param?
It's okay to use `replace()` with multiple segments. The reason I did this is that it was replacing a segment with itself plus others, which is (a) pointless and (b) interferes with the ability to move the anchors up the tree as we want.

Why pointless? Replacing a segment with itself is a "no-op".

barrywhart:
>If (when?) we change the default in production to fail, do we remove some of this?
Yes. We'd remove the test fixture. We'd probably replace the production function (logging) with the other version (raising error). I suggest an intermediate step first, though: Update the log level from warning to critical. There's no rush to do either of these. I'm thinking perhaps 2-3 months between each step.

tunetheweb:
> > You've replaced a lot of replace with create_after. Should we limit replace to only work with one param?
> 
> It's okay to use `replace()` with multiple segments. The reason I did this is that it was replacing a segment with itself plus others, which is (a) pointless and (b) interferes with the ability to move the anchors up the tree as we want.
> 
> Why pointless? Replacing a segment with itself is a "no-op".

That's what I'm asking. Should we programatically prevent this in the `replace()` function by throwing an exception if it includes itself as the first argument, instead of allowing it and then having a check for it after. For the reasons you give, I can't see any reason to do this, that `create_after()` wouldn't be the better option for.

On a similar point, should we allow replacing of one segment with multiple? Does that work? Or should it be a `replace()` (if needed) followed by `create_afters()` for each new segment?

barrywhart:
Re: Replacing one segment with multiple. This could be okay, depending on what the dialect allows.

Generally, I'd rather not restrict something until I have a clear idea that it's causing issues and what should be allowed versus not. Since we're checking changes against the dialect, that'll catch most things that are wrong. These things we're discussing here more like style, or trying to anticipate problems more aggressively.

My biggest concern with fixes is that it discards them if a rule returns multiple fixes with the same anchor. That's sloppy and can be really confusing to a rule author.

PR #2813, opened 2022-03-08
Post-fix parse check: Prefer parse_grammar over match_grammar

barrywhart:
When checking for parse errors after applying a fix, check for `parse_grammar` before `match_grammar`.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -1092,43 +1092,11 @@ class BaseSegment:
 
     def _validate_segment_after_fixes(self, rule_code, dialect, fixes_applied, segment):
         """Checks correctness of new segment against match or parse grammar."""
-        found_error = False
         root_parse_context = RootParseContext(dialect=dialect)
         with root_parse_context as parse_context:
-            if getattr(segment, "match_grammar", None):
-                try:
-                    for seg in segment.segments:
-                        seg.pos_marker = replace(
-                            seg.pos_marker,
-                            templated_file=self.pos_marker.templated_file,
-                        )
-                    match_result = segment.match(
-                        deepcopy(
-                            tuple([seg for seg in segment.segments if not seg.is_meta])
-                        ),
-                        parse_context,
-                    )
-                except ValueError:  # pragma: no cover
-                    found_error = True
-                    self._log_apply_fixes_check_issue(
-                        "After %s fixes were applied, segment %r failed the "
-                        "match() parser check. Fixes: %r",
-                        rule_code,
-                        segment,
-                        fixes_applied,
-                    )
-                else:
-                    if not match_result.is_complete():  # pragma: no cover
-                        found_error = True
-                        self._log_apply_fixes_check_issue(
-                            "After %s fixes were applied, segment %r failed the "
-                            "match() parser check. Result: %r Fixes: %r",
-                            rule_code,
-                            segment,
-                            match_result,
-                            fixes_applied,
-                        )
-            if not found_error and getattr(segment, "parse_grammar", None):
+            # Check for parse_grammar first. If both are present, it is more
+            # thorough and complete.
+            if getattr(segment, "parse_grammar", None):
                 try:
                     # :HACK: Calling parse() corrupts the segment 'r'
                     # in some cases, e.g. adding additional Dedent child
@@ -1149,6 +1117,37 @@ class BaseSegment:
                         r_copy,
                         fixes_applied,
                     )
+            elif getattr(segment, "match_grammar", None):
+                try:
+                    for seg in segment.segments:
+                        seg.pos_marker = replace(
+                            seg.pos_marker,
+                            templated_file=self.pos_marker.templated_file,
+                        )
+                    match_result = segment.match(
+                        deepcopy(
+                            tuple([seg for seg in segment.segments if not seg.is_meta])
+                        ),
+                        parse_context,
+                    )
+                except ValueError:  # pragma: no cover
+                    self._log_apply_fixes_check_issue(
+                        "After %s fixes were applied, segment %r failed the "
+                        "match() parser check. Fixes: %r",
+                        rule_code,
+                        segment,
+                        fixes_applied,
+                    )
+                else:
+                    if not match_result.is_complete():  # pragma: no cover
+                        self._log_apply_fixes_check_issue(
+                            "After %s fixes were applied, segment %r failed the "
+                            "match() parser check. Result: %r Fixes: %r",
+                            rule_code,
+                            segment,
+                            match_result,
+                            fixes_applied,
+                        )
 
     @staticmethod
     def _log_apply_fixes_check_issue(message, *args):  # pragma: no cover
--- a/test/fixtures/rules/std_rule_cases/L003.yml
+++ b/test/fixtures/rules/std_rule_cases/L003.yml
@@ -838,3 +838,42 @@ test_pass_tsql_statistics_indent:
   configs:
     core:
       dialect: tsql
+
+test_fail_snowflake_merge_statement:
+  fail_str: |
+    merge into foo.bar as tgt
+    using (
+    select
+      foo::date as bar
+    from foo.bar
+    where
+    split(foo, '|')[2] REGEXP '^\\d+\\-\\d+\\-\\d+ \\d+\\:\\d+$'
+    OR
+    foo IN ('BAR','FOO')
+    ) as src
+    on
+      src.foo = tgt.foo
+    when matched then
+    update set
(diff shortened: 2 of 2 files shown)


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1092,10 +1092,32 @@ def apply_fixes(self, dialect, rule_code, fixes):
 
     def _validate_segment_after_fixes(self, rule_code, dialect, fixes_applied, segment):
         """Checks correctness of new segment against match or parse grammar."""
-        found_error = False
         root_parse_context = RootParseContext(dialect=dialect)
         with root_parse_context as parse_context:
-            if getattr(segment, "match_grammar", None):
+            # Check for parse_grammar first. If both are present, it is more
+            # thorough and complete.
```
barrywhart, line 1098: This change looks bigger than it is. Basically, I swapped the two checks, preferring the `parse_grammar` if present, only falling back to `match_grammar` if there's no `parse_grammar`.

WittierDinosaur: This looks really good. The only question I have is why are you removing the found_error stuff?

src/sqlfluff/core/rules/base.py
```diff
@@ -798,20 +798,26 @@ def _choose_anchor_segment(context, edit_type, segment, filter_meta=False):
...
-            child = seg
+            # Which sets of children to check against.
+            childrenses = []
+            if filter_meta:
+                # Optionally check against filtered (non-meta only) children.
+                childrenses.append(
+                    [child for child in seg.segments if not child.is_meta]
+                )
+            # Always check against the full set of children.
+            childrenses.append(seg.segments)
```
WittierDinosaur, line 809: does this want an else??

src/sqlfluff/core/parser/segments/base.py
barrywhart: Basically, the `elif` removes the need for it.

Old code:
# Check `match_grammar` if exists.
# If we haven't found an error yet (i.e. using `match_grammar`), check `parse_grammar` if exists.

New code:
# Check `parse_grammar` if exists
# Else check `match_grammar` if exists

WittierDinosaur: Okay, makes sense - I suppose parse grammar is always more explicit

src/sqlfluff/core/rules/base.py
barrywhart: No, we want to check this regardless. I found a corner case where if the rule wants to anchor to a meta segment (such as a `Dedent`), we need to check against `seg.segments`, even if we were asked to ignore meta segments.

The details here can become a bit confusing. The high-level goal is to walk up the parse tree towards the top-level file segment as long as the higher-level segments end/start at the same file location as the one below. This is kind of a hack, but a useful one -- when adding some segments such as whitespace, comments, and semicolons, we want to put them higher in the stack because that's how the grammar rules seem to want them. Someday, this may need to be replaced with something smarter, but this has already been helpful for fixing post-fix parse issues with several rules, which is 😎 .

src/sqlfluff/core/parser/segments/base.py
tunetheweb: It’s not so much that it’s more explicit, it’s that that’s what the parser uses:

https://docs.sqlfluff.com/en/stable/architecture.html#stage-3-the-parser
> Each segment can specify a parse_grammar, and a match_grammar. If both a match_grammar and parse_grammar are defined in a segment, match_grammar is used to quickly prune the tree for branches which do not match segments being parsed, and the parse_grammar is then used to refine the branch identified as correct. If only a match_grammar is defined, then it serves the purpose of both pruning and refining.

src/sqlfluff/core/rules/base.py
WittierDinosaur: But what I don't understand is:
if filter_meta:
  append all non-meta
  then append everything?
  
won't that mean segments are there twice?

src/sqlfluff/core/parser/segments/base.py
tunetheweb, line 1098:
```suggestion
            # Check for parse_grammar first. If both are present, it is the
            # parse_grammar which is actually used for parsing.
```

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1092,10 +1092,32 @@ def apply_fixes(self, dialect, rule_code, fixes):
...
+                    # in some cases, e.g. adding additional Dedent child
+                    # segments. Here, we work around this by calling
+                    # parse() on a "backup copy" of the segment.
+                    r_copy = deepcopy(segment)
+                    for seg in r_copy.segments:
+                        seg.pos_marker = replace(
+                            seg.pos_marker,
+                            templated_file=self.pos_marker.templated_file,
+                        )
+                    r_copy.parse(parse_context)
```
tunetheweb, line 1111: This seems like the correct thing to do, even for `match_grammar`.

Why do we `parse` here, but `match` below? As per https://docs.sqlfluff.com/en/stable/architecture.html#stage-3-the-parser:

> Each segment can specify a parse_grammar, and a match_grammar. If both a match_grammar and parse_grammar are defined in a segment, match_grammar is used to quickly prune the tree for branches which do not match segments being parsed, and the parse_grammar is then used to refine the branch identified as correct. If only a match_grammar is defined, then it serves the purpose of both pruning and refining.

So even though it's ` match_grammar` shouldn't we still use it to `parse` in the case where there is no explicit `parse_grammar`? Surely a `match` check is only a partial check (and an easier one to pass than a full `parse`)?

src/sqlfluff/core/rules/base.py
```diff
@@ -798,20 +798,26 @@ def _choose_anchor_segment(context, edit_type, segment, filter_meta=False):
...
-            if edit_type == "create_before" and children[0] is child:
-                anchor = seg
-                assert anchor.raw.startswith(segment.raw)
-            elif edit_type == "create_after" and children[-1] is child:
-                anchor = seg
-                assert anchor.raw.endswith(segment.raw)
-            else:
-                break
-            child = seg
+            # Which sets of children to check against.
```
tunetheweb, line 801: Why did this all need to change? As per your description I would have thought the first set of changes should be enough?

tunetheweb (changes requested):

src/sqlfluff/core/rules/base.py
barrywhart: Look carefully -- `childrenses` is a list of lists. If it finds a match using _either_ of the lists, it proceeds up the tree. So if it finds a match using the first one, it ignores the second one.

src/sqlfluff/core/parser/segments/base.py
barrywhart: I'll try that. I wrote the original code here without much knowledge of the parser or input from those who do know it. 😬

src/sqlfluff/core/rules/base.py
barrywhart: The `_choose_anchor_segment()` changes fixed an issue with L003. Once I fixed that, I was getting a second parse issue with L052, where it was placing the trailing semicolon too low in the tree. Fixing this allowed the semicolon to be placed correctly. I'll see if I can add a test for that.

src/sqlfluff/core/parser/segments/base.py
barrywhart: Seems to work. Also much simpler code. 👍

src/sqlfluff/core/rules/base.py
WittierDinosaur: oh, gotcha - makes sense now 😅

src/sqlfluff/core/rules/base.py
tunetheweb: Can we also add a comment to the PR description then to note that's include in this PR too then?

barrywhart: I can't reproduce the L052 issue in isolation. If you think it's worthwhile, I can add a test that runs both L003 and L052 -- that would serve as a test for this change. Let me know!

tunetheweb: Well this code is covered. But still weird tests are good to have so maybe should add?

tunetheweb (approved):
LGTM.

As mentioned could you update the PR description (not the title, it's fine) to note other change included in this for L003/L052?

src/sqlfluff/core/rules/base.py
barrywhart: Test added!

PR #2830, opened 2022-03-09
Prevent rules incorrectly returning conflicting fixes to same position

### LINKED ISSUE #2827: Snowflake - Fix Error when semicolon is preceeded by space
### Search before asking

### What Happened

SQLFluff identifies fixable error - and fails to fix it

### Expected Behaviour

Semicolon snaps to end of statement

### Observed Behaviour

No semicolon identified at end of statement, unable to fix

### How to reproduce

sqlfluff fix test.sql --dialect=snowflake

```
SET foo = (SELECT foo FROM foo.foo) ;

```

### Dialect

Snowflake

### Version

Current HEAD of master

### Configuration

Default config - but snowflake

### Code of Conduct

barrywhart:
Fixes #2827 

Changes:
* Fixes L052 (as noted in the issue)
* Fixes L036, L050,L053 (which had other issues uncovered by CI or internal linter checks)
* Changes core linter behavior for rules returning multiple fixes with same anchor segment.
  * Old behavior: Arbitrarily applies one of the fixes, silently discards other fixes with the same anchor. Note this means fixes are not atomic -- bad!!
  * New behavior: Logs a warning and ignores the whole set of fixes. (During automated tests, raises an error rather than logging.)
 

### Pull Request checklist

  - Other.


src/sqlfluff/core/linter/linter.py
```diff
@@ -514,7 +515,16 @@ def lint_fix_parsed(
...
-                    if fixes == last_fixes:  # pragma: no cover
+                    anchor_usage = Counter(fix.anchor for fix in fixes)
+                    duplicate_anchors = [k for k, v in anchor_usage.items() if v > 1]
+                    if duplicate_anchors:
+                        linter_logger.warning(
+                            f"Rule {crawler.code} returned multiple fixes with "
+                            f"the same anchors. This is not supported, so the "
+                            f"fixes will not be applied. %r",
+                            fixes,
+                        )  # pragma: no cover
```
barrywhart, line 526: New error detection code.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1053,7 +1053,6 @@ def apply_fixes(self, dialect, rule_code, fixes):
                                 )
                             # We've applied a fix here. Move on, this also consumes the
                             # fix
-                            # TODO: Maybe deal with overlapping fixes later.
```
barrywhart, line 1056: Delete this comment now that we're dealing with them (by warning and discarding)

src/sqlfluff/rules/L052.py
```diff
@@ -170,20 +170,14 @@ def _handle_semicolon_same_line(
...
-        ]
-        fixes.extend(LintFix.delete(d) for d in info.whitespace_deletions)
+        fixes = self._create_semicolon_and_delete_whitespace(
+            context,
+            info.anchor_segment,
+            info.whitespace_deletions,
+            [
+                SymbolSegment(raw=";", type="symbol", name="semicolon"),
+            ],
+        )
```
barrywhart, line 180: The bug was here.

src/sqlfluff/rules/L052.py
```diff
@@ -224,27 +218,51 @@ def _handle_semicolon_newline(
...
-                ]
+                self._create_semicolon_and_delete_whitespace(
+                    context,
+                    anchor_segment,
+                    info.whitespace_deletions,
+                    [
+                        NewlineSegment(),
+                        SymbolSegment(raw=";", type="symbol", name="semicolon"),
+                    ],
+                )
```
barrywhart, line 229: This code was practically identical to the buggy code. I refactored it to use the same helper function, thus it gets the same fix.

WittierDinosaur:
New error seems to have blown up L036? Maybe more underlying bugs?

barrywhart:
Yes, L036 has a bug as well. Interesting that we hadn't noticed it before. I'll try and fix that in the same PR, as long as it's not a big nasty fix.

barrywhart:
Looking at one of the L036 failures, it is indeed returning multiple fixes with the same anchor. In this case, they are identical (both are `delete`s). That's why we haven't noticed previously. (If the fixes were different, we likely would've noticed, as with L052.) Should be a simple fix. 🤞

barrywhart:
Ok, I fixed the L036 bug. It appears L001 has a "duplicate anchors" bug as well!

tunetheweb:
> Ok, I fixed the L036 bug. It appears L001 has a "duplicate anchors" bug as well!

06C57055-C735-4B09-AA2F-6CDF6E392A37

barrywhart:
It turns out L001 was okay. In one test, it was returning two deletions with segments that had the same position info, but were different objects. I updated the PR to use a new class, `IdentitySet`, that checks for membership by object identity rather than equality.

barrywhart:
@tunetheweb, @WittierDinosaur: Ok, ready for review!!

src/sqlfluff/core/linter/linter.py
```diff
@@ -514,7 +518,17 @@ def lint_fix_parsed(
...
                     linter_logger.info(f"Applying Fixes [{crawler.code}]: {fixes}")
                     # Do some sanity checks on the fixes before applying.
-                    if fixes == last_fixes:  # pragma: no cover
+                    # :TRICKY: Use IdentitySet rather than set() since
+                    # different segments may compare as equal.
+                    unique_anchors = IdentitySet(fix.anchor for fix in fixes)
+                    if len(unique_anchors) < len(fixes):  # pragma: no cover
+                        message = (
+                            f"Rule {crawler.code} returned multiple fixes with "
+                            f"the same anchors. This is not supported, so the "
```
barrywhart, line 527:
```suggestion
                            f"the same anchor. This is not supported, so the "
```

src/sqlfluff/rules/L052.py
```diff
@@ -224,27 +218,54 @@ def _handle_semicolon_newline(
...
             fixes=fixes,
         )
 
+    def _create_semicolon_and_delete_whitespace(
+        self,
+        context: RuleContext,
+        anchor_segment: BaseSegment,
+        whitespace_deletions: Segments,
+        create_segments: List[BaseSegment],
+    ) -> List[LintFix]:
```
barrywhart, line 242: New function extracted from existing code. This is where we fixed the L052 bug.

src/sqlfluff/rules/L052.py
```diff
@@ -224,27 +218,54 @@ def _handle_semicolon_newline(
...
+        # :TRICKY: Use IdentitySet rather than set() since
+        # different segments may compare as equal.
+        whitespace_deletion_set = IdentitySet(whitespace_deletions)
+        if anchor_segment in whitespace_deletion_set:
+            # Can't delete() and create_after() the same segment. Use replace()
+            # instead.
+            lintfix_fn = LintFix.replace
+            whitespace_deletions = whitespace_deletions.select(
+                lambda seg: seg is not anchor_segment
+            )
```
barrywhart, line 256: This is the heart of the bug fix.

test/conftest.py
```diff
@@ -228,16 +229,24 @@ def generate_test_segments_func(elems):
...
+    @staticmethod
+    def raise_error_duplicate_anchors(message: str):  # pragma: no cover
+        raise ValueError(message)
+
     monkeypatch.setattr(
         BaseSegment, "_log_apply_fixes_check_issue", raise_error_apply_fixes_check_issue
     )
+    monkeypatch.setattr(
+        Linter, "_warn_duplicate_anchors", raise_error_duplicate_anchors
+    )
```
barrywhart, line 252: Extends the recently added fixture to make tests fail on fixes with duplicate anchors.

src/sqlfluff/core/linter/linter.py
```diff
@@ -329,6 +329,10 @@ def remove_templated_errors(
                 result.append(e)
         return result
 
+    @staticmethod
+    def _warn_duplicate_anchors(message: str):  # pragma: no cover
+        linter_logger.warning(message)
```
tunetheweb, line 334: Seems a bit of a pointless function? One line of code to log the param, to save you typing one line of code to log the param? The similar `_warn_unfixable` at least has the error message in the function with only the variable part passed but here the whole message is passed.

src/sqlfluff/core/linter/linter.py
```diff
@@ -514,7 +518,17 @@ def lint_fix_parsed(
...
+                    # :TRICKY: Use IdentitySet rather than set() since
+                    # different segments may compare as equal.
+                    unique_anchors = IdentitySet(fix.anchor for fix in fixes)
+                    if len(unique_anchors) < len(fixes):  # pragma: no cover
+                        message = (
+                            f"Rule {crawler.code} returned multiple fixes with "
+                            f"the same anchor. This is not supported, so the "
+                            f"fixes will not be applied. {fixes!r}"
+                        )
+                        cls._warn_duplicate_anchors(message)
```
tunetheweb, line 530: I presume this has to be unique per rule?

i.e. if you have this:

```sql
SELECT DATE_ADD(date,col1)
```

And you had to change that to:

```sql
SELECT DATE_ADD(date, "col1")
```

That is add a space AND change `col1` to be quoted, would that be same anchor point? But as it's different rules there shouldn't be a problem? Or do all rules fixes get queued up to one big set?

I know you say two fixes for same anchor points don't work, but just wondering if they do in some cases and we're now going to break that?

Should this throw a CRITICAL exception so my rules checker will see that?

test/fixtures/rules/std_rule_cases/L052.yml
```diff
@@ -464,3 +464,12 @@ test_fail_newline_create_table_inline_comment:
...
         multiline_newline: true
+
+test_fail_whitespace_after_snowflake_set:
+  fail_str: |
+    SET foo = (SELECT foo FROM foo.foo) ;
+  fix_str: |
+    SET foo = (SELECT foo FROM foo.foo);
+  configs:
+    core:
+      dialect: snowflake
```
tunetheweb, line 475: The simpler statement `SELECT 1  ;` worked before and autofixed.

So that clearly means it went down a different code path (cause it's actually `SELECT<indent>1<dedent> ;` and those `indents/dendents` seemed to be enough to allow it to auto fix with the old code.

Therefore, as it's going down a different code path (or at least used to!), I think we should add that as a test case too.

```suggestion
test_fail_whitespace_after_simple_select:
  fail_str: |
    SELECT 1 ;
  fix_str: |
    SELECT 1;
 
test_fail_whitespace_after_snowflake_set:
  fail_str: |
    SET foo = (SELECT foo FROM foo.foo) ;
  fix_str: |
    SET foo = (SELECT foo FROM foo.foo);
  configs:
    core:
      dialect: snowflake
```

tunetheweb (changes requested):
You're the expert here, but can't see any flags.

Do have a couple of comments/questions and one suggested additional test case.

src/sqlfluff/core/linter/linter.py
barrywhart: This function exists primarily in order to let us monkeypatch it at runtime (replacing it with a function that raises an exception).

src/sqlfluff/core/linter/linter.py
barrywhart: Unique per rule, yes. The linter runs one rule at a time, applies all the fixes, then runs the next rule, and this check occurs within each iteration of the rule loop.

I could change the log level if you like. This is similar to the recent "parse after fix", which currently log warnings. If we change this, should we consider changing that as well?

barrywhart:
@tunetheweb: I added the test case you suggested and replied to your other comments. Let me know what you think.

src/sqlfluff/core/linter/linter.py
tunetheweb: Should we add a comment to that effect?

src/sqlfluff/core/linter/linter.py
tunetheweb: I think I need to update my script to look for more than just CRITICAL log lines if we want to catch these things. Otherwise they’re silently hidden and we only get notice when users raise them.

Or maybe we should consider them critical?

We could also use that key “CRITICAL” word in the log, so my script catches, but then a bit weird that it’s just a warning log level if we do that. It mightn’t be a bad idea to do that before merging, let one run of CI go through, then revert it back assuming no issues. Longer term I should just fix my script though to look for more key phrases we want it to catch.

barrywhart: I'm fine with elevating it to CRITICAL. Since these are bugs in the rules, I don't mind drawing more attention to this.

tunetheweb: Cool let’s do that and see if my script picks up anything!

src/sqlfluff/core/linter/linter.py
barrywhart: Will do.

src/sqlfluff/core/linter/linter.py
barrywhart: Changing this now.

barrywhart:
@tunetheweb: Log level increased to `critical` for this message and the "parse after fix" message. Also added comment explaining why the one-line logging function exists.

test/fixtures/rules/std_rule_cases/L052.yml
```diff
@@ -464,3 +464,18 @@ test_fail_newline_create_table_inline_comment:
     rules:
       L052:
         multiline_newline: true
+
+test_fail_whitespace_after_simple_select:
+  fail_str: |
+    SELECT 1 ;
+  fix_str: |
+    SELECT 1;
+
```
tunetheweb, line 473:
```suggestion

```

tunetheweb:
Oh it found some effors! Good call to make critical.

barrywhart:
@tunetheweb: Whaaa, now more work to do? 😄 I might wait and look at this tomorrow. Stamping out bugs is 😎 , though! Teamwork!

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1008,13 +1048,22 @@ def apply_fixes(self, dialect, rule_code, fixes):
                 else:
                     seg = todo_buffer.pop(0)
 
-                    fix_buff = fixes.copy()
-                    unused_fixes = []
-                    while fix_buff:
-                        f = fix_buff.pop()
-                        # Look for identity not just equality.
-                        # This handles potential positioning ambiguity.
-                        if f.anchor is seg:
```
barrywhart, line 1017: Below, I reworked the core `apply_fixes()` logic:
* Consumes a dictionary of `AnchorEditInfo` rather than a list of fixes
* Add the ability to handle `create_before` and `create_after` the same anchor (new requirement for L053)

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1051,18 +1103,8 @@ def apply_fixes(self, dialect, rule_code, fixes):
...
-                            # TODO: Maybe deal with overlapping fixes later.
-                            break
-                        else:
-                            # We've not used the fix so we should keep it in the list
-                            # for later.
-                            unused_fixes.append(f)
                     else:
                         seg_buffer.append(seg)
-                # Switch over the the unused list
-                fixes = unused_fixes + fix_buff
```
barrywhart, line 1065: Note how this tricky "end of loop" logic all goes away now that we're using a dictionary instead.

src/sqlfluff/rules/L036.py
```diff
@@ -277,10 +279,25 @@ def _fixes_for_move_after_select_clause(
...
+                        )
+                        fixes_ = []
+                        for seg in delete_segments or []:
+                            if seg not in all_deletes:
+                                fixes.append(LintFix.delete(seg))
+                                all_deletes.add(seg)
+                        fixes_ += [
+                            LintFix.delete(seg)
+                            for seg in move_after_select_clause
+                            if seg not in all_deletes
```
barrywhart, line 298: @tunetheweb: I'd like your thoughts on a question. This PR adds a new feature that prohibits fixes with the same anchor with one exception: It's okay to have 2 fixes, one `create_before` and one `create_after`.

Should we also allow multiple **deletes** of the same segment? There's no ambiguity there, and it would avoid needing to "fix" this rule as well as L053.

It's kind of a philosophical question. By allowing this exception, we make it easier on rule writers. On the other hand, we're letting them be a bit sloppy. 🤷‍♂️ I could be convinced either way.

src/sqlfluff/rules/L050.py
```diff
@@ -69,6 +69,12 @@ class Rule_L050(BaseRule):
 
     def _eval(self, context: RuleContext) -> Optional[LintResult]:
         """Files must not begin with newlines or whitespace."""
+        # Only check raw segments. This ensures we don't try and delete the same
+        # whitespace multiple times (i.e. for non-raw segments higher in the
+        # tree).
+        if not context.segment.is_raw():
+            return None
```
barrywhart, line 76: If we allowed multiple deletions, we wouldn't need this change.

src/sqlfluff/rules/L039.py
```diff
@@ -119,4 +124,24 @@ def _eval(self, context: RuleContext) -> Optional[List[LintResult]]:
                     )
                 )
 
-        return violations or None
+        final_violations = []
+        if violations:
+            # Check each violation. If any of its fixes uses the same anchor
+            # as a previously returned fix, discard it. The linter can't handle
+            # applying fixes like this. Skipping this issue is okay because it
+            # will be detected and fixed during the next linter pass.
```
tunetheweb, line 132: Why is this code in L039? Feels like core code that should be in BaseRule.

src/sqlfluff/rules/L052.py
```diff
@@ -224,27 +218,54 @@ def _handle_semicolon_newline(
...
+        anchor_segment = self._choose_anchor_segment(
+            context, "create_after", anchor_segment, filter_meta=True
+        )
+        lintfix_fn = LintFix.create_after
+        # :TRICKY: Use IdentitySet rather than set() since
+        # different segments may compare as equal.
+        whitespace_deletion_set = IdentitySet(whitespace_deletions)
+        if anchor_segment in whitespace_deletion_set:
+            # Can't delete() and create_after() the same segment. Use replace()
+            # instead.
```
tunetheweb, line 252: This comment feels more like an explanation for this PR, rather than something that needs to be in the code.

```suggestion
```

test/fixtures/rules/std_rule_cases/L053.yml
```diff
@@ -65,3 +65,30 @@ test_pass_begin_end_statement_brackets_tsql:
...
+    (
+        SELECT
+            foo,
+            bar,
+            baz
+        FROM mycte2
+    );
+  # Yes, the formatting looks bad, but that's because we're only running L053
+  # here. In the real world, other rules will tidy up the formatting.
+  fix_str: "\n    SELECT\n        foo,\n        bar,\n        baz\n    FROM mycte2\n;\n"
```
tunetheweb, line 81: Why using `\n` when that's not what's used for the `fail_str`? Shouldn't it be consistent? Would also make the initial space look like "bad".

test/fixtures/rules/std_rule_cases/L053.yml
```diff
@@ -65,3 +65,30 @@ test_pass_begin_end_statement_brackets_tsql:
...
+    ( -- This
+        SELECT
+            foo,
+            bar,
+            baz
+        FROM mycte2
+    )
+  # Yes, the formatting looks bad, but that's because we're only running L053
+  # here. In the real world, other rules will tidy up the formatting.
+  fix_str: " -- This\n    SELECT\n        foo,\n        bar,\n        baz\n    FROM mycte2\n\n"
```
tunetheweb, line 94: Ditto.

tunetheweb (changes requested):
Still missing some coverage too.

src/sqlfluff/rules/L036.py
tunetheweb: Hmmm.... I think safer to not allow that. In theory it shouldn't be needed so the developer ease is not a strong enough argument in my mind. And curious why L036 currently does it? Looked at the code but wasn't immediately apparent to me.

src/sqlfluff/core/linter/linter.py
```diff
@@ -514,12 +520,22 @@ def lint_fix_parsed(
...
+                    ):  # pragma: no cover
+                        message = (
+                            f"Rule {crawler.code} returned multiple fixes with the "
+                            f"same anchor. This is only supported for create_before+"
+                            f"create_after, so the fixes will not be applied. {fixes!r}"
+                        )
+                        cls._report_duplicate_anchors_error(message)
+                    elif fixes == last_fixes:  # pragma: no cover
                         cls._warn_unfixable(crawler.code)
                     else:
```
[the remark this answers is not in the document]
barrywhart: Will do.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -50,6 +52,44 @@ class FixPatch(NamedTuple):
...
+        * 2 fixes: Valid if and only if types are create_before and create_after
+        """
+        if self.total <= 1:
+            # Definitely no duplicates if <= 1.
+            return True
+        if self.total != 2:  # pragma: no cover
+            # Definitely duplicates if > 2.
+            return False
+        # Special case: Ok to create before and after same segment.
+        return self.create_before == 1 and self.create_after == 1
```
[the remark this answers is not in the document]
barrywhart: I'll update it to something similar as you suggest.

As currently written, 0 will never occur because we only call `is_valid` if there are fixes. But I'd prefer to treat 0 as valid because it's harmless and gives us a bit of future proofing.

src/sqlfluff/rules/L036.py
barrywhart: There's not a "good" reason, and I am not that familiar with L036 details, but basically, there are a couple places that delete unnecessary whitespace, and there was no bookkeeping, so in some cases, the same whitespace gets deleted twice.

Happy to leave the fix checker "as is" for now (i.e. not allow multiple deletes).

src/sqlfluff/rules/L039.py
barrywhart: We don't want `BaseRule` to discard when multiple fixes have the same anchor, because we consider it an error (the `is_valid` check).

L039 had duplicate anchors on this new test case (extracted from one of the `.sql` fixtures):
```
test_excess_space_cast:
  fail_str: |
    select
        '1'    ::   INT as id1,
        '2'::int as id2
    from table_a
  fix_str: |
    select
        '1'::INT as id1,
        '2'::int as id2
    from table_a
```

L039 wants to make two fixes to the line `'1'    ::   INT as id1,`.
1. Replace excessively long whitespace with a single whitespace: "    " -> " " (2 places)
2. Entirely **remove** the whitespace around the `::`.

Thus, it's trying to both **replace** and **delete** the same whitespace. If we return both fixes, the core linter will (appropriately) complain and discard **both** fixes. This bookkeeping ensures that both get fixed, but it's split across two passes through the linter loop. There may be a smarter way to do this, but this approach seems reasonable. I'm trying really hard not to do big rewrites of existing rule code during these PRs -- the goal is to eliminate the critical errors but try and avoid going down a 🐰 hole.

src/sqlfluff/rules/L052.py
barrywhart: Really? I think it's helpful -- it explains in the line below why we are switching `lintfix_fn` from `create_after` to `replace`, which otherwise may not be obvious to a future maintainer. Do you think there's a better way to explain what's happening here?

test/fixtures/rules/std_rule_cases/L053.yml
barrywhart: It's because of a YAML limitation. The fix string has a blank line at the end, but the YAML parser doesn't pick it up; it assumes blank line means end of string. The `fail_str` doesn't have a blank line. Think I should change it? I prefer to use "normal" multi-line strings when possible, using quoted strings with `\n` or other escape sequences only when necessary or for readability. Happy to change the `fail_str` if you like, though!

barrywhart: I'll go ahead and change `fail_str`.

src/sqlfluff/rules/L052.py
tunetheweb: Ok see that now. Thought it was an explanation of why you rewrote it the way you did. Resolving

src/sqlfluff/rules/L039.py
tunetheweb: Thanks makes sense. Maybe add a comment saying, something like:

This rule works in two steps to remove unnecessary white space:
1. Replace duplicate whitespace to one single whitespace
2. Remove single white spaces if needed.

This can result in two delete being applied to same segment so area so check for that and replace with single delete.

src/sqlfluff/core/parser/segments/base.py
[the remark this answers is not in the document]
tunetheweb: SGTM

test/fixtures/rules/std_rule_cases/L053.yml
barrywhart: Done!

test/fixtures/rules/std_rule_cases/L053.yml
barrywhart: Done!

barrywhart:
@tunetheweb: Ready for another review.

barrywhart:
@alanmcruickshank: You may be interested in the changes to `BaseSegment.apply_fixes()`. We now handle one case of "overlapping fixes": two fixes with the same anchor, one is `create_before` and one is `create_after`.

@OTooleMichael: You may be interested because it adds a new "sanity check" to the fixes. Previously, rules could return conflicting fixes -- it would apply one fix and silently ignore the rest.

Issue #2771, opened 2022-03-03
SQLFluff improperly parses unquoted Bigquery project names with dashes


kdw2126:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Whenever I parse valid BigQuery code which contains my organization's primary project names (which contain dashes), SQLFluff treated the dash as an operator and attempts to apply operator fixes to this code.

### Expected Behaviour

I would expect that when an unquoted project name with a dash is clearly being cited as part of a FROM statement, SQLFluff properly understands that the dash is part of a project name and does not apply dash-related operator rules (related to subtraction) to it.

### Observed Behaviour

Currently, whenever I use a dashed project name without quoting it, the dash is treated as an operator and spaces are added before and after it (which breaks the SQL code I am working on).

### How to reproduce

(1). Generate a new BigQuery SQLFluff project.
(2)  Lint the following code with it:
```
SELECT col_foo
FROM foo-bar.foo.bar
```
(3). Note that the linter output takes the following form, indicating that the dash is being treated like an operator.  This can be reproduced from online.sqlfluff.com
Code | Line / Position | Description
-- | -- | --
2 / 9 | Missing whitespace before -
2 / 9 | Missing whitespace after -

### Dialect

bigquery

### Version

sqlfluff, version 0.10.1
Python 3.8.9
dbt version 1.0.1, although this issue does not involve the dbt templater

### Configuration

[sqlfluff]
dialect = bigquery
exclude_rules = L003,L008,L011,L014,L016,L029,L031,L034

[sqlfluff:rules]
max_line_length = 120
comma_style = leading

[sqlfluff:rules:L010]
capitalisation_policy = upper

[sqlfluff:rules:L030]
capitalisation_policy = upper

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

tunetheweb:
More discussion on this here: https://github.com/sqlfluff/sqlfluff/pull/2756#discussion_r817726019

These [dashes are very much the exception](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical) so should ensure we don't allow dashes in other unquoted identifiers.

[...]

PR #2839, opened 2022-03-10
Spark3: statements to `SET` and `RESET` spark runtime configurations

R7L208:

Add `ResetStatementSegment` and a DRAFT of `SetStatementSegment`

`NoscanKeywordSegment` was erroneously set as `SymbolSegment`; it is corrected to `KeywordSegment` in this PR.

### Pull Request checklist
- [] Please confirm you have completed any of the necessary steps below.

  - Other.


src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -286,11 +286,17 @@
...
     ),
     WhlKeywordSegment=StringParser(
         "WHL", KeywordSegment, name="whl", type="file_keyword"
     ),
+    SQLConfPropertiesSegment=StringParser(
+        "-v",
+        SymbolSegment,
+        name="sql_conf_properties_parameter",
+        type="sql_conf_properties",
+    ),
```
R7L208, line 299: I defined this segment to match "-v" thinking I would get similar behavior to `EqualsSegment_a` which matches on "=="; however, the parser is still splitting "-v" into "-" and "v". I've tried a test case where "-v" is isolated w/o any`SET` statement and added `SQLConfPropertiesSegment`to `StatementSegment` when doing this with no luck.

```bash
INFO       Found unparsable segment...
INFO       [L:  3, P:  4]      |unparsable:                                                   !! Expected: 'Nothing...'
[L:  3, P:  4]      |    whitespace:                                               ' '
[L:  3, P:  5]      |    raw:                                                      '-'
[L:  3, P:  6]      |    raw:                                                      'v'
```

Any suggestions on how to match/parse this? Using a combination of `MinusSegment` and `SingleIdentifierSegment` works but tries to enforce a whitespace after the "-" sign. Seemed like creating a new SymbolSegment would be a better solution than modifying a rule for this one edge-case for the spark dialect.

R7L208: Got it running 🙂  just needed to insert a new `StringLexer` before "minus".

Will clean up then remove DRAFT status.

R7L208: Got it running 🙂  just needed to insert a new `StringLexer` before "minus".

Will clean up then remove DRAFT status.

tunetheweb: Hmmm... I'm wondering if this si the right way of going about it?

Let's say you had an identifier called `-vmycolumn`. That would now be lexed and parsed as two identifiers.

That might be pretty unlikely but what about `SELECT col1-variable1`?

I think for the one exception it might be better to define `SQLConfPropertiesSegment` as being made up of two symbols:

```
    SQLConfPropertiesSegment=Sequence(
        Ref("MinusSegment"),
        StringParser("v", SymbolSegment, name="sql_conf_option_set", type="sql_conf_option"),
        allow_gaps=False,
    ),
```

R7L208: I tried a similar approach but kept getting the below:

```bash
INFO       LINTING (test/fixtures/dialects/spark3/set.sql)
== [test/fixtures/dialects/spark3/set.sql] LINTING
== [test/fixtures/dialects/spark3/set.sql] FAIL
L:   3 | P:   5 | L006 | Missing whitespace after -
==== summary ====
files:             1 violations:        1
clean files:       0 unclean files:     1
avg per file:   1.00 unclean rate:   100%
status:         FAIL
All Finished 📜 🎉!
```

Linter keeps suggesting a whitespace even though `allow_gaps` is False. Based on your feedback, seems like an exception to L006 is simpler than defining a new `StringLexer`

tunetheweb: Ah this works better then:

```
    SQLConfPropertiesSegment=Sequence(
        StringParser("-", SymbolSegment, name="dash", type="dash"),
        StringParser("v", SymbolSegment, name="sql_conf_option_set", type="sql_conf_option"),
    ),
```

tunetheweb: And this has just helped me fix #2771!

tunetheweb (approved):
LGTM!

PR #2840, opened 2022-03-11
Change rule test script from bash to pythong

OTooleMichael:
Refactor out the Bash in favour of PyTest.
- Readable
- Better outputs
- Tester can focus in on one test

Nope

### Pull Request checklist


OTooleMichael:
@tunetheweb This is the requested PR that moves these integration tests into pytest.

Would like to merge and then get #2795.

test/dialects/dialects_test.py
```diff
@@ -22,30 +25,94 @@
...
+    config = FluffConfig(overrides=json.loads(config_overrides))
     tokens, lex_vs = Lexer(config=config).lex(raw)
     # From just the initial parse, check we're all there
     assert "".join(token.raw for token in tokens) == raw
     # Check we don't have lexing issues
     assert not lex_vs
-
     # Do the parse WITHOUT lots of logging
     # The logs get too long here to be useful. We should use
     # specific segment tests if we want to debug logs.
```
OTooleMichael, line 39: `pytest test/dialects/dialects_test.py | grep -v "INFO\|DEBUG\|\[L\|#\|Initial\|^$"`
Really helps with the overly verbose logging that occurs during errors here. Should we put it somewhere for people connivence? eg in a comment here.
Excluding lines that:
- have INFO
- have DEBUG
- output the parse tree ([L ....
- mention Initial Structure (Initial)
- describe parse depth -  # Beginning Parse Depth 4: SelectStatementSegment
- Are blank `^$`

tunetheweb:
OK I spot a problem before I even look at the code, that would be good to hear your input on.

The tests normally take about 10 mins. And that's just for parsing. Windows test is usually last to finish and it takes a bit longer (maybe 12 or 14 mins).

Previously this `fix` test was run separately and in parallel and took about same length of time as Window test (12 or 14 mins) so that was acceptable as it didn't extend the length of CI. It only ran on one Python version and only on linux but should pretty much be the same, so I was comfortable with that.

Now that we've merged it into the main test run, it's running on all instances (cool for better coverage), but it's massively extended the run times:

image

And Windows in particular has doubled:

image

27 mins is an awful long time to wait for CI feedback!

Ideas:
- Drop this test from Windows run anyway (though some tox params?). That would still mean CI has gone from 12-14mins to 20 mins which is not great, but better than 12-14 mins to 27min.
- Move these `fix` tests back to being a parallel run and only run it once one one Python version. But likely to take just as long I presume? Or maybe would go down to half the time if only doing the `fix` and not the `parse` and `fix`? Actually surely `fix` does a `parse` so if we're doing that, then not sure of the benefit of `parse` now?

OTooleMichael:
> OK I spot a problem before I even look at the code, that would be good to hear your input on.
> 
> The tests normally take about 10 mins. And that's just for parsing. Windows test is usually last to finish and it takes a bit longer (maybe 12 or 14 mins).
> 
> Previously this `fix` test was run separately and in parallel and took about same length of time as Window test (12 or 14 mins) so that was acceptable as it didn't extend the length of CI. It only ran on one Python version and only on linux but should pretty much be the same, so I was comfortable with that.
> 
> Now that we've merged it into the main test run, it's running on all instances (cool for better coverage), but it's massively extended the run times:
> 
> image
> 
> And Windows in particular has doubled:
> 
> image
> 
> 27 mins is an awful long time to wait for CI feedback!
> 
> Ideas:
> 
> * Drop this test from Windows run anyway (through some tox params?). That would still mean CI has gone from 12-14mins to 20 mins which is not great, but better than 12-14 mins to 27min.
> * Move these `fix` tests back to being a parallel run and only run it once one one Python version. But likely to take just as long I presume? Or maybe would go down to half the time if only doing the `fix` and not the `parse` and `fix`? Actually surely `fix` does a `parse` so if we're doing that, then not sure of the benefit of `parse` now?

I think that 100% to be expected. We are adding Fixing to every single test in the dialect folder (in sequence).

Previously the bash was running on a separate process and not matrixed.

We could split out one more set of tests to run in parallel?

tunetheweb:
> We could split out one more set of tests to run in parallel?

Let's try that. Hoping it will be as fast as Bash (it's all Python under the hood anyway!)

test/dialects/dialects_test.py
```diff
@@ -22,57 +25,123 @@
...
+    print(f"Post-parse structure: {parsed.stringify()}")
+    # Check we're all there.
+    assert parsed.raw == raw
+    # Check that there's nothing unparsable
+    typs = parsed.type_set()
+    assert "unparsable" not in typs
+
+
+@pytest.mark.integration_test
+@pytest.mark.parametrize("dialect,file", parse_success_examples)
```
OTooleMichael, line 86: Add a marker to break up tests.

test/dialects/dialects_test.py
```diff
@@ -22,57 +25,123 @@
...
+def test__dialect__base_broad_fix(
+    dialect, file, raise_critical_errors_after_fix, cached_parser
+):
+    """Run a full fix with all rules, in search of critical errors."""
+    raw = load_file(dialect, file)
+    config_overides = dict(dialect=dialect)
+    # Lean on the cached result of the above test if possible
+    parsed: Optional[BaseSegment] = cached_parser(config_overides, raw)
```
tunetheweb, line 94: Will this every be cached now they are run separately? Wondering if it's costing us time to cache these (only to use them once), than it would be not to have the cache?

OTooleMichael: Yup I'm look if its worth having. I had a few thoughts
- isn't helping across runs (obviously)
- still help for local testing
- I noticed another place where we are parsing the same files again but on a 2sec fix it didnt work

OTooleMichael: I removed it for now

OTooleMichael:
Fixed the speed issue also - Just needed a `-n CORES` in the pytest

tunetheweb:
Nice. Will review now. Saw there was a network error on last run so just kicked them off again.

.github/workflows/ci-tests.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         pip install .
     - name: Run the tests
       run: |
-        tox -e ruleslinting
+        tox -e ruleslinting -- -n 4 test
```
tunetheweb, line 57: We use `-n 2` for the CI tests.

Wonder should we consolidate on `2` or `4` or have different ones?

@pwildenhain do you remember why you went for `2` initially?

tunetheweb (approved):
LGTM!

One small question about the number of processors to use?

.github/workflows/ci-tests.yml
OTooleMichael: If there are 4 it uses 4 if there are 2 it makes little difference. Most people locally will have at least 4

tunetheweb: Yeah just wonder if we should use `4` for the other sets rather than `2`? Would that be any faster?

tunetheweb: Let's see if this is any faster: https://github.com/sqlfluff/sqlfluff/pull/2841

It's not really to do with this PR, but curious and think would be better to be consistent, unless a good reason not to be.

tunetheweb: Ah looks like they are all two-core machines for GitHub Actions: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

And increasing to 4 is not looking any faster to me in #2841, presumably because of that.

Would you mind changing to two for consistency? That way if we want to change, we can find the all with one search of the code.

```suggestion
        tox -e ruleslinting -- -n 2 test
```

OTooleMichael: So I daw you tested and its not faster in Github, however should I switch them to all 4 or all 2. There is still an argument for 4 as its more common for the dev env. 

:)

tunetheweb: Well this is YML file is only for GitHub actions and not used locally?

Plus my pathetic machine slows to a crawl when using `sqlfluff lint -p 4`.

So my vote is 2,  but could be persuaded to 4 (automatic improvement if/when GitHub improve their compute).

barrywhart: If you want to be clever, you can do this in bash:
```
$(python -c 'import multiprocessing; print(multiprocessing.cpu_count())')
```

That returns the number of CPUs on the current machine.

test/dialects/dialects_test.py
barrywhart: One potential alternative: Could we use `caplog` to temporarily reduce the logging verbosity?

https://docs.pytest.org/en/6.2.x/logging.html#caplog-fixture

test/dialects/dialects_test.py
```diff
@@ -22,57 +24,111 @@
...
-        assert "unparsable" not in typs
+    if not raw:
+        return None
+
+    return Parser(config=config).parse(tokens)
+
+
+def lex_and_parse(config_overrides: Dict[str, Any], raw: str) -> Optional[BaseSegment]:
+    """Sets up cache parsable inputs from normal inpurts."""
+    cachable_str = json.dumps(config_overrides, sort_keys=True)
```
barrywhart, line 47: Why are we converting JSON to string, only to call a function that immediately converts it back to JSON?

test/dialects/dialects_test.py
```diff
@@ -22,57 +24,111 @@
...
+
+
+@pytest.mark.integration_test
+@pytest.mark.parametrize("dialect,file", parse_success_examples)
+def test__dialect__base_broad_fix(dialect, file, raise_critical_errors_after_fix):
+    """Run a full fix with all rules, in search of critical errors."""
+    raw = load_file(dialect, file)
+    config_overides = dict(dialect=dialect)
+    # Lean on the cached result of the above test if possible
+    parsed: Optional[BaseSegment] = lex_and_parse(config_overides, raw)
```
barrywhart, line 77: Do any of the test files use Jinja? I'm confused that both tests call `lex_and_parse()` with the raw file contents, but this test calls the templater, but only _after_ parsing. It seems like this couldn't possibly work if the files use templating.

Did you consider calling a higher level `Linter` functions directly, similar to the test functions in this file? https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/testing/rules.py

test/dialects/dialects_test.py
```diff
@@ -22,57 +24,111 @@
...
-            "in test/fixtures/dialects."
-        )
-    else:
+    if not parsed:
         assert parsed == res
+        return
+
+    # Verify the current parse tree matches the historic parse tree.
+    parsed_tree = parsed.to_tuple(code_only=code_only, show_raw=True)
+    # The prased tree consists of a tuple of "File:", followed by the
```
barrywhart, line 121:
```suggestion
    # The parsed tree consists of a tuple of "File:", followed by the
```

test/dialects/dialects_test.py
```diff
@@ -22,57 +24,111 @@
...
+    # statements. So only compare when there is at least one statement.
+    if parsed_tree[1] or res[1]:
+        assert parsed_tree == res
+    # Verify the current hash matches the historic hash. The main purpose of
+    # this check is to force contributors to use the generator script to
+    # to create these files. New contributors have sometimes been unaware of
```
barrywhart, line 127:
```suggestion
    # create these files. New contributors have sometimes been unaware of
```

test/dialects/dialects_test.py
tunetheweb: None of our dialect fixtures use Jinja.

test/dialects/dialects_test.py
OTooleMichael: Was cached in a middling step (cache_lru). I thought leaving it so designed with the doc strings wasn't a harm

test/dialects/dialects_test.py
OTooleMichael: Yup thats the way I started out, but then went about caching the Parse as it was expensive and slow.

OTooleMichael: I made this simpler

test/dialects/dialects_test.py
OTooleMichael: Got rid of all traces of caching

.github/workflows/ci-tests.yml
OTooleMichael: I changed it to 2. Its all the same then and thus at least not confusing.

barrywhart (approved):
This looks good to me.

Mostly, this PR touches areas that @tunetheweb created. @tunetheweb, would you like to take a final look and merge if you're happy with it?

PR #2847, opened 2022-03-12
Reduce unnecessary setting run ci

zhongjiajie:
* In job `python-version-tests`, we already told the version we
  setup in step `Set up Python`, so we could simply run command
  `tox -e py` to trigger the job, so we could remove the
  `Set Tox Python Param` step
* In job `other-tests` we just set an unused variable by one step
  and I think we should directly remove it.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -70,14 +70,9 @@ jobs:
     - name: Install dependencies
       run: |
         pip install tox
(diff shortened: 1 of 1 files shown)


zhongjiajie:
CI pass and it look good, PTAL @tunetheweb @barrywhart

tunetheweb (approved):
Good spots!

zhongjiajie:
Thanks

Issue #2844, opened 2022-03-11
Enhancements to generate_parse_fixture_yml.py command


tunetheweb:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

This takes a while to run and is only going to take longer as test coverage gets more and more comprehensive.

Could we add parallel ability?

Could we add ability to only generate for one file or folder so when making dialect changes you don't need to needlessly generate all the other dialects too (CI will protect us if someone incorrectly thinks they don't need to regenerate it all).

### Use case

Will make developing dialect improvements easier

### Dialect

All

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

tunetheweb:
@barrywhart / @OTooleMichael think one of you could look at this as more your area of expertise? You'd forever have my gratitude in return! 😉

OTooleMichael:
I can prolly have a go. How slow are they at the moment. I think I Ran this last week and it was fine

tunetheweb:
Need to persuade the wife to let me buy an M1 (or M2?) Max 😁

Takes a good 5-10 mins for me on my aging Mac. Which, when I only really need one generated is a bit much.

barrywhart:
The script is already parallel. See [these lines](https://github.com/sqlfluff/sqlfluff/blob/main/test/generate_parse_fixture_yml.py#L55L57).

barrywhart:
Letting you specify a directory  is a good idea. 👍

tunetheweb:
Just did some timing there (single runs, with terminal mostly in the foreground but lots of apps in the background):

It took 4 mins 19 seconds for me to run it in default mode (I have 4 cores btw, well 2 and 2 v)
```
% date; python test/generate_parse_fixture_yml.py; date
Fri 11 Mar 2022 21:29:33 GMT
Fri 11 Mar 2022 21:33:52 GMT
```

Hard coded some other values for number of cores and got following run times (again just a single run so not that scientific) shows much worse times - including reverting back to default of 4 cores!:

cores | run time
----|----
1| 8 mins 03 seconds
2| 7 mins 03 seconds
3| 8 mins 02 seconds
[...]

PR #2850, opened 2022-03-13
Add ability to regenerate subsets of fixture YAMLs (by dialect, or new only)

OTooleMichael:
#2844

Adds some flags for faster/ more minimal rebuilding
- `--new-only`
- `--dialect bigquery`
- `--filter *with_*`

Nope

### Pull Request checklist

### DIFF
--- a/test/generate_parse_fixture_yml.py
+++ b/test/generate_parse_fixture_yml.py
@@ -1,6 +1,11 @@
 """Utility to generate yml files for all the parsing examples."""
 import multiprocessing
 import os
+import fnmatch
+import re
+import click
+from typing import Callable, Dict, List, NamedTuple, Optional, TypeVar
+
 
 import yaml
 
@@ -8,7 +13,22 @@ from conftest import compute_parse_tree_hash, get_parse_fixtures, parse_example_
 from sqlfluff.core.errors import SQLParseError
 
 
-def generate_parse_fixture(example):
+S = TypeVar("S")
+
+
+def distribute_work(work_items: List[S], work_fn: Callable[[S], None]) -> None:
+    """Distribute work and ignore results."""
+    with multiprocessing.Pool(multiprocessing.cpu_count()) as pool:
+        for _ in pool.imap_unordered(work_fn, work_items):
+            pass
+
+
+class _ParseExample(NamedTuple):
+    dialect: str
+    sqlfile: str
+
+
+def generate_one_parse_fixture(example: _ParseExample) -> None:
     """Parse example SQL file, write parse tree to YAML file."""
     dialect, sqlfile = example
     tree = parse_example_file(dialect, sqlfile)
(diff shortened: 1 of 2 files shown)


OTooleMichael:
@tunetheweb - As requested. Does this do the trick?

tunetheweb (approved):
This is brilliant! Thank you sooo much!

One small nit on docs.

test/generate_parse_fixture_yml.py
```diff
@@ -1,60 +1,142 @@
...
+
+
+class _ParseExample(NamedTuple):
+    dialect: str
+    sqlfile: str
+
+
+def _create_yaml_path(example: _ParseExample) -> str:
+    dialect, sqlfile = example
+    root = sqlfile[:-4]
```
barrywhart, line 33: I realize this was pre-existing code that you moved, but could we use `os.path.splitext()` here instead?

test/generate_parse_fixture_yml.py
```diff
@@ -1,60 +1,142 @@
...
+    ]
+
+
+@click.command()
+@click.option(
+    "--filter", "-f", default=None, help="A glob filter to apply to file names."
+)
+@click.option("--dialect", "-d", default=None, help="Filter to a given dialect.")
+@click.option(
+    "--new-only",
```
barrywhart, line 117: It could also be cool to have a way to re-parse _modified_ files (where the `.sql` file is newer than the `.yml` file).

test/generate_parse_fixture_yml.py
```diff
@@ -1,60 +1,142 @@
...
+    default=False,
+    help="Only create missing fixtures.",
+)
+def generate_parse_fixtures(
+    filter: Optional[str], dialect: Optional[str], new_only: bool
+):
+    """Generate fixture or a subset based on dialect or filename glob match."""
+    filter_str = filter or "*"
+    dialect_str = dialect or "all"
+    print("Match Pattern Recieved:")
```
barrywhart, line 129:
```suggestion
    print("Match Pattern Received:")
```

barrywhart (approved):
A couple suggestions, but nothing to hold up merging.

test/generate_parse_fixture_yml.py
OTooleMichael: Good idea, but can we push this feat for another PR?

test/generate_parse_fixture_yml.py
OTooleMichael: Done

test/generate_parse_fixture_yml.py
OTooleMichael: I went ahead and did it

tunetheweb: Amazing! 🙌

barrywhart: Thanks, @OTooleMichael! About waiting, the answer was "yes", but it's too late now. 😄 Thanks again for the great work!!

PR #2852, opened 2022-03-13
Redshift: transaction statement

rpr-ableton:

Redshift transaction segment as specified in the documentation.

nope

### Pull Request checklist

  - Other.


tunetheweb:
These are all independent statements. That is simpler to implement and at least allows them to parse (which can unblock people), but does mean there is no linking between them (e.g. having a `BEGIN` without the corresponding `END` will be parsed as valid even though it is not).

T-SQL implemented proper block-level transaction statements if wanted to look at that?

However, I’m also happy to merge this as is if think it’s sufficient for now.

rpr-ableton:
Very good point, thanks! There's no reason not to do things as properly as possible. I'll have another pass at it tomorrow! 🙂

rpr-ableton:
Hi @tunetheweb, I'm afraid I won't have time to work on this until we go ahead with v0.11.1. I think it does the job well enough that we can go ahead and merge that for now. :sweat_smile:

tunetheweb (approved):

Issue #2677, opened 2022-02-17
L003: Jinja for-loop indentation not stable across runs (dbt, snowflake)


csimonsson:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

SQLFluff fix is not stable over the following query. Fixing query will remove indentation, running another fix will put the indentation back again.

### Expected Behaviour

I consider the query to be correctly formatted, no errors should be present.

### Observed Behaviour

SQLfluff is not stable, changing indentation of the query for every run.

### How to reproduce

Run SQLFluff on this query:

```
{% for t in ['table_a', 'table_b'] %}
    SELECT
        column_a,
        column_b
    FROM {{ t }}
    {% if not loop.last %} 
        UNION ALL
    {% endif %}
{% endfor %}
```

### Dialect

snowflake, but the error is also present in ansi.

### Version

sqlfluff 0.10.1
sqlfluff-templater-dbt 0.10.1
dbt-snowflake 1.0.0

### Configuration

```
[sqlfluff]
templater = dbt
tab_space_size = 4
exclude_rules = L031

[sqlfluff:templater:dbt]
project_dir = dwh/
profile = dbt_monorepo_snowflake

[sqlfluff:rules:L003]
indent_unit = space

[sqlfluff:rules:L004]
indent_unit = space

[sqlfluff:rules:L010]
capitalisation_policy = upper

[sqlfluff:rules:L016]
max_line_length = 120

[sqlfluff:rules:L057]
quoted_identifiers_policy = aliases
```

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

OTooleMichael:
With new L003 Draft pr this is stable across two passes. Perhaps due to fixes in the jinja templating.
However there are still oddities with the Lexer. The L003 cant indent it correctly as it is receiving a strange lex stream.

If line 6 has a trailing space during lex, we receive this stream.

```
[...]

PR #2857, opened 2022-03-14
V1 L003 refactor

OTooleMichael:
- L003 fails to understand dedents 
- L003 is a speed bottleneck (does tonnes of unnecessary loops) line M of N lines reprocesses M-1 [unchanged] lines. 
- L003 has a issue with inconsistent jinja - #2677
- Code is hard to read
- Code is ignoring types in many places


src/sqlfluff/rules/L003.py
```diff
@@ -416,23 +515,26 @@ def _process_current_line(
...
-        check_comment_line = this_line["line_buffer"]
-        if check_comment_line and all(
-            seg.is_type(
-                "whitespace", "comment", "indent"  # dedent is a subtype of indent
+        if this_line_segments.all(
+            sp.is_type(
+                "whitespace",
+                "comment",
+                "indent",
+                "newline",  # dedent is a subtype of indent
```
OTooleMichael, line 536: Our lines now also have the newline char. Makes little difference

src/sqlfluff/rules/L003.py
```diff
@@ -477,15 +579,14 @@ def _process_current_line(
...
-            if this_line["indent_size"] > 0:
-                self.logger.debug("    Indented First Line. #%s", this_line_no)
-                return LintResult(
-                    anchor=trigger_segment,
-                    memory=memory,
-                    description="First line has unexpected indent",
-                    fixes=[LintFix.delete(elem) for elem in this_line["indent_buffer"]],
-                )
+        if this_line_no == 1 and this_line["indent_size"] > 0:
+            self.logger.debug("    Indented First Line. #%s", this_line_no)
```
OTooleMichael, line 583: Un required nesting

src/sqlfluff/rules/L003.py
```diff
@@ -477,15 +579,14 @@ def _process_current_line(
...
+            return LintResult(
+                anchor=trigger_segment,
+                memory=memory,
+                description="First line has unexpected indent",
+                fixes=[LintFix.delete(elem) for elem in this_line["indent_buffer"]],
```
OTooleMichael, line 588: Consider not deleting INDENT/DEDENT

src/sqlfluff/rules/L003.py
```diff
@@ -552,7 +653,11 @@ def _process_current_line(
                 fixes = self._coerce_indent_to(
                     desired_indent=desired_indent,
                     current_indent_buffer=this_line["indent_buffer"],
-                    current_anchor=this_line["line_buffer"][0],
+                    current_anchor=[
+                        el
+                        for el in this_line["line_buffer"]
+                        if el not in this_line["indent_buffer"]
+                    ][0],
```
OTooleMichael, line 660: The Anchor here was the start of the indent buffer, which was inconsistent with other anchors/triggers

src/sqlfluff/rules/L003.py
```diff
@@ -708,27 +831,19 @@ def _process_current_line(
...
+                    # Calc the balance of this line.
+                    # Double the contribution for any indents at the start of the line.
+                    b_num = len(dedents) - len(positive_indents)
+                    # - len(start_line_indents)
+                    if len(dedents):
+                        b_num = b_num - len(indents.select(stop_seg=dedents.get(0)))
+
+                    self.logger.debug((b_num, this_line["line_buffer"]))
+
+                    if b_num < indent_diff:
```
OTooleMichael, line 846: Part of the dubious discussion from above

src/sqlfluff/rules/L003.py
```diff
@@ -806,30 +930,32 @@ def _process_current_line(
...
-                            desired_indent="".join(
-                                elem.raw for elem in this_line["indent_buffer"]
-                            ),
-                            current_indent_buffer=prev_line["indent_buffer"],
-                            current_anchor=anchor,
-                        )
-
-                        memory["problem_lines"].append(n)
-                else:
+                if n not in memory["comment_lines"]:
```
OTooleMichael, line 935: Flip to make guard statement

test/rules/std_L003_test.py
```diff
@@ -106,17 +107,17 @@ def test__rules__std_L003_process_raw_stack(generate_test_segments, test_elems,
...
     assert isinstance(res, dict)
     assert all(isinstance(k, int) for k in res.keys())
-    assert all(isinstance(v, dict) for v in res.values())
     # Check keys are all present
     assert all(
-        v.keys()
+        asdict(v).keys()
         == {
             "line_no",
             "templated_line",
```
OTooleMichael, line 123: Not sure if it holds value to test this kind of internal

src/sqlfluff/rules/L003.py
```diff
@@ -101,45 +153,75 @@ def _indent_size(segments: Sequence[BaseSegment], tab_space_size: int = 4) -> in
...
+                        "indent_buffer": indent_buffer,
+                        "indent_size": indent_size,
+                        # Indent balance is the indent at the start of the first content
+                        "indent_balance": this_indent_balance,
+                        "indent_balance_marker": indent_balance,
+                        "hanging_indent": hanger_pos if line_indent_stack else None,
+                        # Clean indent is true if the line *ends* with an indent
+                        # or has an indent in the initial whitespace.
+                        "clean_indent": clean_indent,
+                        "template_content": "".join(
```
barrywhart, line 218: Nice!

src/sqlfluff/rules/L003.py
```diff
@@ -315,40 +429,19 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
-                # comment_lines keeps track of lines which are all comment.
-                "comment_lines": [],
-                # segments we've seen the last child of
-                "finished": set(),
-                # First non-whitespace node on a line.
-                "trigger": None,
-            }
-        else:
-            memory = context.memory
+        memory: _MemoryDict = context.memory or _MemoryDict()
```
barrywhart, line 444: I like this. I wonder if we should add core support for "memory as a class". The current API encourages you to use dictionaries, which are awkward to build and work with.

OTooleMichael: Perhaps mainly this is for the typing. Which I think is more important than the class. In fact wouldn't be a fan of memory that did anything (ie not a class) just remembered (data)

barrywhart: I guess my main point is, I like having the default values for new memory in a canonical, easy to find place, rather than buried somewhere deep inside `_eval()`.

src/sqlfluff/rules/L003.py
```diff
@@ -567,207 +661,171 @@ def _process_current_line(
...
+                # to where we need to be? NB: This should only be applied if this is
+                # a CLOSING bracket.
+
+                # First work out if we have some closing brackets, and if so, how
+                # many.
+                # Due to strange inheritence all dedents are also indents
+
+                self.logger.debug((positive_indents, dedents, start_line_indents))
+                # Calc the balance of this line.
+                # Double the contribution for any indents at the start of the line.
```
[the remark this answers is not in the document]
OTooleMichael: Poorly expressed and WIP. Also I'm trying to reuse the existing logic to achieve something that it might not be able. 
Basically we are trying to see if during this line we balance out for example do we have a double indent somewhere and double decent here making this actually balanced. However sometimes we have lines essentially starting with indents which are requesting that we push this line deeper. I think an actually stack of pairing would solve this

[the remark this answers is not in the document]
barrywhart: +1 for leveraging good data structures where applicable. I think this code was prototyped into existence, and sometimes the patterns are only obvious in retrospect.

src/sqlfluff/rules/L003.py
OTooleMichael: Ahhh I've a nice idea for that, why not have a class property of the init value of memory.
then the _crawl could also implement it for you

src/sqlfluff/rules/L003.py
```diff
@@ -776,136 +825,263 @@ def _process_current_line(
...
+        """Yields the tuple of seg type and underlying type were applicable."""
+        for seg in self.current_line:
+            raw_seg = self.get_raw_slices(seg)
+            yield (seg.type, raw_seg[0].slice_type if raw_seg else None)
+
+    def iterate_type_pairs(self):
+        """Produce a list of pairs of each sequenctial combo of two."""
+        iterable = self.list_segement_and_raw_segement_types()
+        a, b = itertools.tee(iterable)
+        # consume the first item in b
```
OTooleMichael, line 963: This is a lot more verbose but much more understandable and seperated out than it was before

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -51,6 +55,45 @@ test_fail_tab_indentation:
     rules:
       indent_unit: tab
 
+# test_case_in_fn:
```
barrywhart, line 58: What's this for? If there's an issue that needs to be addressed before enabling this, could you mention that in a comment?

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -51,6 +55,45 @@ test_fail_tab_indentation:
...
+#             a
+#         ))) AS c,
+#         MAX(CASE
+#             WHEN b THEN 1
+#             WHEN c THEN 1
+#         END) as agg
+#     FROM my_tbl
+
+test_fix_progressive_indentation_errors:
+  # Using tabs as indents works
```
barrywhart, line 81: Are these actually tabs? If you're testing with tabs, I suggest using quoted strings with `\t` so someone's editor doesn't auto-convert them to spaces.

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -620,6 +663,44 @@ test_tsql_function:
...
+            column_b
+        FROM {{ t }}
+        {% if not loop.last %}
+            UNION ALL
+        {% endif %}
+    {% endfor %}
+    ;
+
+test_stable_jinja_loop:
+  # Matches the fix output from above
```
barrywhart, line 691: IIUC, the YAML test framework nowadays checks to confirm that the `fix_str` of a test yields no subsequent fixes. If so, this test case may not be needed. Can you confirm this, @tunetheweb?

test/rules/std_L003_test.py
```diff
@@ -1,145 +1,6 @@
...
-from unittest.mock import Mock
-
 import pytest
-
-from sqlfluff.core.config import FluffConfig
 from sqlfluff.rules.L003 import Rule_L003
-from sqlfluff.testing.rules import get_rule_from_set
-
-
-@pytest.mark.parametrize(
```
barrywhart, line 11: Do you think we are losing anything important by removing these tests?

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,156 @@
...
+        """Strip a line dict of buffers for logging."""
+        keys_to_strip = ("line_buffer", "indent_buffer", "ignoreable_reference")
+        print_dict: Dict = {
+            key: value for key, value in self.items() if key not in keys_to_strip
+        }
+        print_dict["raw"] = "".join(el.raw for el in self.line_buffer)
+        return print_dict
+
+    def trigger_el(self):
+        """Get the pivot element."""
```
barrywhart, line 68: Can you add some detail about what you mean by "pivot element"? A brief SQL example could be helpful for context.

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,156 @@
...
+        # Assume an unclean indent, but if the last line
+        # ended with an indent then we might be ok.
+        # Was there an indent after the last code element of the previous line?
+        for search_elem in reversed(line_buffer):
+            is_meta = search_elem.is_meta
+            if not search_elem.is_code and not is_meta:
+                continue
+            elif is_meta and search_elem.indent_val > 0:
+                self.clean_indent = True
+            break
```
barrywhart, line 104: Could it be worth creating a separate function for this? (More self documenting.)

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,156 @@
...
+
+        return self
+
+    def from_memo_line(self, line_no: int, templated_file: Optional[TemplatedFile]):
+        """Create a final summary from a memo line."""
+        memo_line = self
+        copied_line_buffer = memo_line.line_buffer[:]
+        template_info = _TemplateLineInterpreter(copied_line_buffer, templated_file)
+        # Generate our line summaary based on the current state of the MemoLine
```
barrywhart, line 113:
```suggestion
        # Generate our line summary based on the current state of the MemoLine
```

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,156 @@
...
+    indent_size: int = 1
+    indent_balance: int = 0
+    hanging_indent: Optional[int] = None
+    clean_indent: bool = True
+    template_content: str = ""
+    templated_line_type: Optional[str] = None
+    line_anchor: Optional[BaseSegment] = None
+
+    # These are only required for carrying state between lines
+    anchor_indent_balance: int = 0
```
OTooleMichael, line 53: @barrywhart  - Im a little confused by what `anchor_indent_balance` and `indent_balance` are at which time in which roles.

The code is really just the same as it was, just that the state of `anchor_indent_balance` is saved also to this object.

I feel as though one of them is the indent balance at the end of the line (once fully processed) and one is the balance at the line at the point where the anchor was / is

The line we are calculating fixes for will always have the same, as with the current algo we break when we reach the first anchor of the last/current line

There are some weird inversions going on (line 124) and Id like to correct them any idea,

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,156 @@
...
+                continue
+            if el.segments:
+                continue
+            if el.is_meta and el.indent_val != 0:
+                continue
+            return el
+
+        return None
+
+    def memo_line_reset(self):
```
barrywhart, line 82: What's a "memo line"? If it means "memory", maybe use the whole word in functions for clarity?

src/sqlfluff/rules/L003.py
```diff
@@ -100,197 +244,201 @@ def _indent_size(segments: Sequence[BaseSegment], tab_space_size: int = 4) -> in
...
-                    "indent_size": indent_size,
-                    # Indent balance is the indent at the start of the first content
-                    "indent_balance": this_indent_balance,
-                    "hanging_indent": hanger_pos if line_indent_stack else None,
-                    # Clean indent is true if the line *ends* with an indent
-                    # or has an indent in the initial whitespace.
-                    "clean_indent": clean_indent,
-                }
+                # convert our memo line into a true summary
+                # then rest the memo aspects that need reseting
```
barrywhart, line 285:
```suggestion
                # then reset the memo aspects that need reseting
```

src/sqlfluff/rules/L003.py
```diff
@@ -100,197 +244,201 @@ def _indent_size(segments: Sequence[BaseSegment], tab_space_size: int = 4) -> in
...
+
+        # Poor Man's correction of duplicate changes
+        seen = set()
+        for res in results:
+            if not res.fixes:
+                continue
+            fixes: List[LintFix] = []
+            for fix_el in res.fixes:
+                this_set = (fix_el.anchor, fix_el.edit_type)
+                if this_set in seen:
```
barrywhart, line 435: In some cases in segment world (I forget when), two segments will compare equal but are not the same object. This can cause "set-like" operations to misbehave. Is that a potential concern here? I've been working around that by using `id(my_object)` or using a recently added class, `sqlfluff.core.parser.segments.base.IdentitySet`.

src/sqlfluff/rules/L003.py
```diff
@@ -776,136 +830,263 @@ def _process_current_line(
...
-            ) and segment_info(idx) == ("indent", None):
-                return True
-        return False
+    def list_segement_and_raw_segement_types(self):
+        """Yields the tuple of seg type and underlying type were applicable."""
+        for seg in self.current_line:
+            raw_seg = self.get_raw_slices(seg)
+            yield (seg.type, raw_seg[0].slice_type if raw_seg else None)
+
+    def iterate_type_pairs(self):
```
barrywhart, line 964: Maybe rename for clarity?
```suggestion
    def iterate_adjacent_type_pairs(self):
```

src/sqlfluff/rules/L003.py
```diff
@@ -776,136 +830,263 @@ def _process_current_line(
...
+
+    def is_block_end(self):
+        # Check for this sequence
+        dedent_types = (("dedent", None),)
+        end_block = (
+            ("placeholder", "block_end"),
+            ("placeholder", "compound"),
+            ("placeholder", "block_mid"),
+        )
+        valid_combos = list(
```
barrywhart, line 1000: It looks like this is going to compute the cross product every time the function is executed -- probably not a huge cost, but seems unnecessary. Think we should cache the result of this computation?

barrywhart (commented):
Several small suggestions. Looks pretty good overall.

I noticed a few new functions that don't have docstrings. It would be cool to have at least a brief docstring for all the functions.

src/sqlfluff/rules/L003.py
barrywhart: I don't know. If you want to tag Alan, he may be the best able to answer questions about the existing code.

test/fixtures/rules/std_rule_cases/L003.yml
OTooleMichael: This is the main one Im working on. Without this the test "should have been passing". but some 3.7 issues are causing them to fail.

test/fixtures/rules/std_rule_cases/L003.yml
OTooleMichael: yea can do. This wasnt my test not sure why its touched

test/fixtures/rules/std_rule_cases/L003.yml
OTooleMichael: Not sure that could be true. One of the issues with L003 is that it was removing indent markers sometimes. causing the next pass to actually have different markers. It would be stable for a single run (as the parse tree would reach something.)

test/rules/std_L003_test.py
OTooleMichael: No 100% not. The ones I took away are just checking internals that are well covered by the tests at large. Also they weren't clearly covering complex contained internal logic that needed to stay stable. just the input and outout of a large function.

src/sqlfluff/rules/L003.py
OTooleMichael: Im still figuring out, its essentially the "tigger el". L003 was using a token to mark where to stop. This caused all previous line to be fully computed and the most recent line only to be computed up to the first non indentation token.

Id like to capture this logic such that each line knows about / holds enough info on both states. (allowing us to do a single pass) and also to consider senarios  where the whole of the current line is required for computation

eg my `\n))))` example where the bracket legitimately close off a bunch of indents

src/sqlfluff/rules/L003.py
```diff
@@ -561,213 +661,167 @@ def _process_current_line(
...
+                # a CLOSING bracket.
+
+                # First work out if we have some closing brackets, and if so, how
+                # many.
+                # Due to strange inheritence all dedents are also indents
+
+                # Calc the balance of this line.
+                # Double the contribution for any indents at the start of the line.
+                b_num = len(dedents) - len(positive_indents)
+                # - len(start_line_indents)
```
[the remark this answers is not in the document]
OTooleMichael: Yea still lots to do. Ill clean this up

src/sqlfluff/rules/L003.py
OTooleMichael: yup can do. Will moving this to a class property have that effect?

test/fixtures/rules/std_rule_cases/L003.yml
barrywhart: Maybe I was thinking of [this code](https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/testing/rules.py#L210L214): Tests _without_ a `fix_str` do not apply any fixes.

We should probably modify the test framework to detect and fail if a second round of fixing makes additional changes.

src/sqlfluff/rules/L003.py
barrywhart: Yes, assignments at class level are only evaluated when that module is first imported.

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -590,7 +632,7 @@ test_tsql_function:
             ) + '12' + CAST(24 + DATEPART(day, @DATE) AS char(2))) + 1;
     --Special case Dec 29-31 may belong to the next year
         IF ((DATEPART(mm, @DATE) = 12)
-        AND ((DATEPART(dd, @DATE) - DATEPART(dw, @DATE)) >= 28))
```
OTooleMichael, line 593: This single test is the only thing changed in terms of test - I noticed both now and before the final state of this query is based on the input state (which seems super weird to me). I found this test hard to reason about what is correct in terms of indent for this.

tunetheweb:
@OTooleMichael I presume this can be closed as superseded by #2884?

PR #2858, opened 2022-03-14
Fix/l032 using

OTooleMichael:
Fixes error noticed due to `L042`

```

CRITICAL   [L032] Applying rule L032 threw an Exception: not enough values to unpack (expected 2, got 1)                                                                                                

  File "/Users/bhart/dev/sqlfluff/src/sqlfluff/rules/L032.py", line 101, in _eval

    table_a, table_b = select_info.table_aliases

ValueError: not enough values to unpack (expected 2, got 1)
```

An unnamed subquery produced no table_alias when performing select analysis leading to a run time error.

Fixed and added an improvement to the algo

- Old Algo: only fixes joins with a max of one other table
- new Algo: Fixes regardless of the number of joins but only the first joined table (where the join condition is clear)

### Pull Request checklist

  - [x]`.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.

### DIFF
--- a/src/sqlfluff/core/rules/functional/segments.py
+++ b/src/sqlfluff/core/rules/functional/segments.py
@@ -1,9 +1,12 @@
 """Surrogate class for working with Segment collections."""
-from typing import Any, Callable, Iterator, List, Optional, overload
+from typing import Any, Callable, Iterable, Iterator, List, Optional, overload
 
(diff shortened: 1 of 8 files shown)


OTooleMichael:
Closing Due to being mixed up with L042

PR #2859, opened 2022-03-14
L032 bug fix and fix improvement

OTooleMichael:
- L032 was failing on unaliased subqueries due to runtime error
- L032 now fixes up to the first join (rather than only where there is exactly one join)

Nope

### Pull Request checklist

### DIFF
--- a/src/sqlfluff/rules/L032.py
+++ b/src/sqlfluff/rules/L032.py
@@ -82,9 +82,8 @@ class Rule_L032(BaseRule):
             sp.is_type("join_clause", "from_expression_element")
         )
 
-        # If we have more than 2 tables we won't try to fix join.
-        # TODO: if this is table 2 of 3 it is still fixable
-        if len(tables_in_join) > 2:
+        # We can only safely fix the first join clause
+        if segment.get(0) != tables_in_join.get(1):
             return unfixable_result
 
         parent_select = parent_stack.last(sp.is_type("select_statement")).get()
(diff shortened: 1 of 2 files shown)


barrywhart:
Looks good!

barrywhart:
Do you have any ideas what was happening with L046 in the stack trace I shared earlier?

```
WARNING    Unable to lint test.sql due to an internal error. Please report this as an issue with your query's contents and stacktrace below!                                                            
  File "/Users/bhart/dev/sqlfluff/src/sqlfluff/rules/L046.py", line 56, in _eval
AssertionError
```

When I saw that `pos_marker` was not set, I wondered if this could possibly be related to the recent changes you and I made, where (if I recall correctly) we made the `pos_marker` field `Optional`. Do you think the error above could've been a knock-on effect of that change? We don't have to address this now, but I'm making a mental note to watch for similar errors in case we need to take a closer look. (To be clear, I'm not saying the two things are related, just curious. In general, I try and watch for recurring patterns with bugs, so we can find and address root causes.)

OTooleMichael:
> Do you have any ideas what was happening with L046 in the stack trace I shared earlier?
> 
> ```
> WARNING    Unable to lint test.sql due to an internal error. Please report this as an issue with your query's contents and stacktrace below!                                                            
> To hide this warning, add the failing file to .sqlfluffignore
>   File "/Users/bhart/dev/sqlfluff/src/sqlfluff/core/rules/base.py", line 535, in crawl
>     res = self._eval(context=context)
>   File "/Users/bhart/dev/sqlfluff/src/sqlfluff/rules/L046.py", line 56, in _eval
>     assert context.segment.pos_marker
> AssertionError
> ```
> 
> When I saw that `pos_marker` was not set, I wondered if this could possibly be related to the recent changes you and I made, where (if I recall correctly) we made the `pos_marker` field `Optional`. Do you think the error above could've been a knock-on effect of that change? We don't have to address this now, but I'm making a mental note to watch for similar errors in case we need to take a closer look. (To be clear, I'm not saying the two things are related, just curious. In general, I try and watch for recurring patterns with bugs, so we can find and address root causes.)

Honestly no idea, Also I dont understand how things can keep going after a critical error. Perhaps the error break is in a weird place ?

Cause the L046 can only happen after the exception of L032 as far as I found

barrywhart:
"Critical error" is just a logging level. We log things as critical if we want:
* Users to report it to us as a bug
* CI builds to cause the build to fail (e.g. the "Rules critical errors" build step)

If you don't have ideas, no worries. SQLFluff is improving, but definitely not perfect. We have enough users that they find bugs on a regular basis, and we fix them. Easy peasy. 😄

PR #2864, opened 2022-03-15
Spark3: Support for `SHOW` statements

R7L208:

Implement `ShowStatement` in Spark3 to support various `SHOW` statements.

- Created `PropertiesNakedIdentifierSegment` and refactored a few Segments to use it since Spark allows the referencing of reserved keywords in table and runtime property names.
- Refactored `BracketedPropertyListGrammar` to use new segment and to split out grammar related to the property key into its own variable.

### Pull Request checklist

  - Other.


tunetheweb:
Did you check linting this? i.e. will rules L014, L029, L057, L059 still flag these correctly (guessing L014, L057 and L059 definitely should, and L029 probably should even though allowed).

R7L208:
> OK was more asking that the rules are still useful with this change than asking you to correct the test cases (many of our fixtures don't pass all the rules and that's fine).

> So I was asking to check it doesn't ignore PropertiesNakedIdentifierSegment when it shouldn't (or vice versa) and if the rules needed updated to include/exclude that additional identifier type? Hopefully not as it's still type="identifier".

Yeah just realized that and was re-writing my comment but you responded faster than I could.

Going to do some additional checking and work on this. Will convert PR to draft for now

R7L208:
- Fixed L014 so it applies; writing test cases then will push a new commit.
- L029 should not apply to `PropertiesNakedIdentifierSegment` since the use of keywords in properties is allowed by spark.
- For L057, I'm using the `QuotedLiteralSegment` instead of `QuotedIdentifierSegment` since it's a very common pattern for property keys to be like `'created.date'` where they contain one or several `.`. Open to suggestions on this but seemed the easiest implementation since they `.` is part of the string. This means that `PropertiesNakedIdentifierSegment` won't match special characters causing an unparsable error and `QuotedLiteralSegment` bypasses L057 (which only checks for quoted_identifiers and naked_identifiers.
- L059 is similarly only looking at quoted_identifers and naked_identifiers. `QuotedLiteralSegments` are not included in the rule

R7L208:
Once I get tests for L014 working, I'm going to refactor the `PropertyKeyGrammar` into a segment so that I can write a switch to `QuotedIdentifierSegment` then replace `.` for `` when checking for special characters.

R7L208:
Ran into an area where I could use some thoughts.

In the most recent commit, I've refactored `PropertiesKeyGrammar` to `PropertiesKeySegment`. I added this to the initial addition of segments and grammars to `spark3` and followed an existing pattern in rule L057. Are segments defined in `<dialect>.add()` not exposed in the same way? Trying to understand why this is working in L057 for `FileReferenceSegment` but not `PropertyKeySegment`?

My initial thought is that `type` must be defined for it to be surfaced as a unique segment but not seeing a way to define `type` in the current way `PropertyKeySegment` is implemented.

tunetheweb:
Yes my understanding is those Segments added through `add` are not full segments/dialect objects and are more simple segments for reference in the actual classes. Though some can be made into them through `SegmentGenerator`.

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -390,6 +400,20 @@
...
+    # not match keywords; however, SparkSQL allows keywords to be used in table
+    # and runtime properties.
+    PropertiesNakedIdentifierSegment=SegmentGenerator(
+        # Generate the anti template from the set of reserved keywords
+        lambda dialect: RegexParser(
+            r"[A-Z0-9]*[A-Z][A-Z0-9]*",
+            CodeSegment,
+            name="properties_naked_identifier",
+            type="identifier",
+        )
```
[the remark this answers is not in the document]
tunetheweb: Ooops. Got this wrong. Need to remove the SegmentGenerator wrapper from it too!

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -390,6 +400,18 @@
...
+    # and runtime properties.
+    PropertiesNakedIdentifierSegment=SegmentGenerator(
+        # Generate the anti template from the set of reserved keywords
+        RegexParser(
+            r"[A-Z0-9]*[A-Z][A-Z0-9]*",
+            CodeSegment,
+            name="properties_naked_identifier",
+            type="identifier",
+        )
+    ),
```
R7L208, line 414:
```suggestion
    PropertiesNakedIdentifierSegment=RegexParser(
        r"[A-Z0-9]*[A-Z][A-Z0-9]*",
        CodeSegment,
        name="properties_naked_identifier",
        type="identifier",
    ),
```

src/sqlfluff/rules/L057.py
```diff
@@ -118,12 +118,12 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
-                and context.parent_stack
-                and context.parent_stack[-1].name == "FileReferenceSegment"
-            ):
-                return None
+            if context.dialect.name in ["spark3"] and context.parent_stack:
+                if context.parent_stack[-1].name == "FileReferenceSegment":
+                    return None
+
+                if context.parent_stack[-1].name == "PropertyKeySegment":
+                    identifier = identifier.replace(".", "")
```
R7L208, line 126:
```suggestion
            if context.dialect.name in ["spark3"] and context.parent_stack:

                # Spark3 file references for direct file query
                # are quoted in back ticks to allow for identfiers common
                # in file paths and regex patterns for path globbing
                # https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-file.html
                #
                # Path Glob Filters (done inline for SQL direct file query)
                # https://spark.apache.org/docs/latest/sql-data-sources-generic-options.html#path-global-filter
                #
                if context.parent_stack[-1].name == "FileReferenceSegment":
                    return None

                # Spark3 properties keys used for setting table and runtime
                # configurations denote namespace using dots, so these are
                # removed before testing L057 to not trigger false positives
                # Runtime configurations:
                # https://spark.apache.org/docs/latest/configuration.html#application-properties
                # Example configurations for table:
                # https://spark.apache.org/docs/latest/sql-data-sources-parquet.html#configuration
                #
                if context.parent_stack[-1].name == "PropertyNameSegment":
                    identifier = identifier.replace(".", "")
```

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -308,17 +318,17 @@
     TerminatedByGrammar=hive_dialect.get_grammar("TerminatedByGrammar"),
     # Add Spark Grammar
     PropertyGrammar=Sequence(
-        OneOf(
-            Ref("LiteralGrammar"),
-            Ref("SingleIdentifierGrammar"),
-        ),
+        Ref("PropertyKeySegment"),
```
R7L208, line 321:
```suggestion
        Ref("PropertyNameSegment"),
```

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -2052,18 +2074,145 @@ class SetStatementSegment(BaseSegment):
...
         "SET",
         Ref("SQLConfPropertiesSegment", optional=True),
-        Sequence(
-            Delimited(
-                Ref("SingleIdentifierGrammar"),
-                delimiter=Ref("DotSegment"),
-                allow_gaps=False,
+        OneOf(
+            Ref("PropertyListGrammar"),
+            Ref("PropertyKeySegment"),
```
R7L208, line 2079:
```suggestion
            Ref("PropertyNameSegment"),
```

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -2368,3 +2520,21 @@ class FromExpressionElementSegment(BaseSegment):
     get_eventual_alias = ansi_dialect.get_segment(
         "FromExpressionElementSegment"
     ).get_eventual_alias
+
+
+@spark3_dialect.segment()
+class PropertyKeySegment(BaseSegment):
+    """A segment for key to set and retrieve table and runtime properties."""
+
+    type = "property_key_identifier"
```
R7L208, line 2529:
```suggestion
class PropertyNameSegment(BaseSegment):
    """A segment for a property name to set and retrieve table and runtime properties.
    
    https://spark.apache.org/docs/latest/configuration.html#application-properties
    """

    type = "property_name_identifier"
```

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -308,17 +318,17 @@
...
+        Ref("PropertyKeySegment"),
         Ref("EqualsSegment", optional=True),
         OneOf(
             Ref("LiteralGrammar"),
             Ref("SingleIdentifierGrammar"),
         ),
     ),
-    BracketedPropertyListGrammar=Bracketed(Delimited(Ref("PropertyGrammar"))),
+    PropertyKeyListGrammar=Delimited(Ref("PropertyKeySegment")),
+    BracketedPropertyKeyListGrammar=Bracketed(Ref("PropertyKeyListGrammar")),
```
R7L208, line 329:
```suggestion
    BracketedPropertyNameListGrammar=Bracketed(Ref("PropertyNameListGrammar")),
```

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -417,7 +439,7 @@
         "UNSET",
         "TBLPROPERTIES",
         Ref("IfExistsGrammar", optional=True),
-        Bracketed(Delimited(Ref("QuotedLiteralSegment"))),
+        Ref("BracketedPropertyKeyListGrammar"),
```
R7L208, line 442:
```suggestion
        Ref("BracketedPropertyNameListGrammar"),
```

src/sqlfluff/dialects/dialect_spark3.py
```diff
@@ -2052,18 +2074,145 @@ class SetStatementSegment(BaseSegment):
...
+                    "LIKE",
+                    Ref("QuotedLiteralSegment"),
+                    optional=True,
+                ),
+            ),
+            # SHOW TBLPROPERTIES
+            Sequence(
+                "TBLPROPERTIES",
+                Ref("TableReferenceSegment"),
+                Ref("BracketedPropertyKeyListGrammar", optional=True),
```
R7L208, line 2200:
```suggestion
                Ref("BracketedPropertyNameListGrammar", optional=True),
```

tunetheweb (approved):

R7L208:
@tunetheweb - Thanks so much for the help closing this out

PR #2870, opened 2022-03-15
Refactor JinjaTracer: Split into two classes, break up "_slice_template()" function

barrywhart:

Several recent fixes to `JinjaTracer` have demonstrated that it's a large class that's tricky to maintain. Two major concerns:
* The `JinjaTracer._slice_template()` function is 275 lines long with lots of intricate `if` conditions, it's hard to read and easy to break. This PR splits it up into a number of smaller functions, reducing it to about 100 lines.
* `JinjaTracer` performs two related but fairly distinct tasks. Combining these in one class makes it hard to understand because it's unclear at times which fields are used for one task and which for both.
  * Analyze a template to prepare for tracing.
  * Trace the template.

In creating this PR, I was pretty careful to avoid making behavior changes to the code. This code has pretty good automated testing now, as well as internal sanity checking of the `JinjaTracer` output. So, in reviewing this PR, I suggest focusing more on coding style and readability than correctness or "what changed from before" (because the text changes are massive due to moving things around).

### Pull Request checklist

  - Other.


requirements_dev.txt
```diff
@@ -22,6 +22,7 @@ types-chardet
 types-appdirs
 types-colorama
 types-pyyaml
+types-Jinja2
```
barrywhart, line 25: Unrelated to the rest of the PR, but I noticed we were missing this

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -186,7 +188,37 @@ def slice_info_for_literal(self, length, prefix="") -> RawSliceInfo:
         """
         unique_alternate_id = self.next_slice_id()
         alternate_code = f"\0{prefix}{unique_alternate_id}_{length}"
-        return RawSliceInfo(unique_alternate_id, alternate_code, [])
+        return self.make_raw_slice_info(unique_alternate_id, alternate_code, [])
+
+    def update_inside_set_or_macro(self, block_type, trimmed_parts):
```
barrywhart, line 193: Extracted this function from `update_inside_set_or_macro()`.

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -215,9 +247,9 @@ def _slice_template(self) -> List[RawFileSlice]:
         # https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.Environment.lex
         stack = []
         result = []
-        set_idx = None
```
barrywhart, line 218: This moved and was renamed as a new instance field on `JinjaTracer`: `inside_set_or_macro`.

barrywhart:
Tagging @OTooleMichael for review. I don't think I can explicitly add him as a reviewer yet (probably a GitHub group needs updating).

tunetheweb:
> Tagging @OTooleMichael for review. I don't think I can explicitly add him as a reviewer yet (probably a GitHub group needs updating).

@alanmcruickshank sent you an invite couple of days back. Check your email (and spam folder) @OTooleMichael

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -165,6 +170,29 @@ def record_trace(self, target_slice_length, slice_idx=None, slice_type=None):
         if slice_type in ("literal", "templated"):
             self.source_idx += target_slice_length
 
+
+class JinjaAnalyzer:
+    """Analyzes a Jinja template to prepare for tracing."""
+
+    re_open_tag = regex.compile(r"^\s*({[{%])[\+\-]?\s*")
+    re_close_tag = regex.compile(r"\s*[\+\-]?([}%]})\s*$")
```
OTooleMichael, line 178: Why are these not full Regexes - ie starting with ^ and ending with $?

Also wondering if gathering the whitespace before a tag has influence on indents (I note later on that search is used vs the lexed array) perhaps some of these matches are forced by what the lexer returns?

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -186,280 +214,319 @@ def slice_info_for_literal(self, length, prefix="") -> RawSliceInfo:
...
+                        )
+                    if block_type == "templated" and tag_contents:
+                        assert m_open and m_close
+                        raw_slice_info = self.track_templated(
+                            m_open, m_close, tag_contents
+                        )
+                self.update_inside_set_or_macro(block_type, tag_contents)
+                m_strip_right = regex.search(
+                    r"\s+$", raw, regex.MULTILINE | regex.DOTALL
+                )
```
OTooleMichael, line 313: What happening here ?

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -186,280 +214,319 @@ def slice_info_for_literal(self, length, prefix="") -> RawSliceInfo:
...
+            # Handle a tag received as individual parts.
+            trimmed_parts = str_parts[1:-1]
+            if trimmed_parts[0].isspace():
+                del trimmed_parts[0]
+            if trimmed_parts[-1].isspace():
+                del trimmed_parts[-1]
+        else:
+            # Handle a tag received in one go.
+            trimmed_content = str_buff[len(m_open.group(0)) : -len(m_close.group(0))]
+            trimmed_parts = trimmed_content.split()
```
OTooleMichael, line 441: Is there need for the If and the str_parts array at all, can the Else just be applied to both scenarios? (I guess its not expensive enough to matter about the splitting computation)

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -186,280 +214,319 @@ def slice_info_for_literal(self, length, prefix="") -> RawSliceInfo:
...
+
+        We detect skipped characters by looking ahead in in_str for the token
+        just returned from lex(). The token text will either be at the current
+        'idx_raw' position (if whitespace stripping did not occur) OR it'll be
+        farther along in 'raw_str', but we're GUARANTEED that lex() only skips
+        over WHITESPACE; nothing else.
+        """
+        # Find the token returned. Did lex() skip over any characters?
+        num_chars_skipped = self.raw_str.index(token, self.idx_raw) - self.idx_raw
+        if num_chars_skipped:
```
OTooleMichael, line 518: Personally Id prefer a guard `if not num_chars_skipped: \ return`

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -186,280 +214,319 @@ def slice_info_for_literal(self, length, prefix="") -> RawSliceInfo:
...
-                    # returns, it has simply grouped them differently than we
-                    # want.
-                    trailing_chars = len(m.group(0))
-                    result.append(
+                    # to behave similarly as the left stripping case. Note that
+                    # the stakes are a bit lower here, because lex() hasn't
+                    # *omitted* any characters from the strings it returns,
+                    # it has simply grouped them differently than we want.
+                    trailing_chars = len(m_strip_right.group(0))
+                    self.raw_sliced.append(
```
OTooleMichael, line 322: Perhaps just throw this into a fn for ease of reading its nice to balance 
`handle_left_whitespace_stripping` even if this one is easier

OTooleMichael (approved):
LGTM:
I read the code linearly, commented as I went and asked questions.

src/sqlfluff/core/templaters/slicers/tracer.py
tunetheweb: The templated block is a whole string and not the individual components.

So `{% if ... %}` is all one templated string.

So we want `^{%` for opening tag (simplified), and `%}$` for closing tag (simplified) of that string, as @barrywhart has.

src/sqlfluff/core/templaters/slicers/tracer.py
tunetheweb: Not sure you need the `DOTALL` here (there is no `.`)?

src/sqlfluff/core/templaters/slicers/tracer.py
barrywhart: In the past, we received one token at a time from Jinja, then combined them into a single string (`str_buff`) once they became a full block, e.g. `{% if a == 3 %}`. Now we still build `str_buff`, but we also keep the parts separate in an array `str_parts`. The regexes remain the same for legacy reasons. I _think_ it would work either way.

Mostly we want what the lexer returns, as the purpose of this class is to deduce exactly what Jinja is going to return for the "real" template, including tricky whitespace handling. That's been working pretty well for a while. If you look at `handle_left_whitespace_stripping()` and the code section (not function beginning with `# Right whitespace was stripped.`, you'll see how some of the trickier bits work.

There is _one_ case where Jinja returns the whole block at a time, and that's a `{% raw %} ... {% endraw %}` section.

src/sqlfluff/core/templaters/slicers/tracer.py
barrywhart: We're trying to determine if a block uses right whitespace stripping, e.g.
```
SELECT
  {{ 'col1,' -}}
  col2
```

Note the `-`. This tells Jinja to discard any whitespace between the `}}` and the next bit of text. We need to account for this because it affects the template output.

barrywhart: I can't remember why these flags were added. Maybe they were necessary back when we only had the whole block string, rather than the parts? Note that in dbt, it's common to do things like:
```
```

`DOTALL` actually means "allow the `.` pattern character to match newlines`, and `MULTILINE` means "allow multiline matches". https://www.codespeedy.com/re-dotall-in-python/

I'm hesitant to change this now. 😅  If the string were really long, I could see it slowing things down a bit, but this string is literally just one Jinja token -- should be very short.

tunetheweb: Fair point. It' snot changed in this PR and don't think it'll matter too much one way of another. So fine leaving it.

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -186,280 +214,319 @@ def slice_info_for_literal(self, length, prefix="") -> RawSliceInfo:
...
+                        raw_slice_info = self.track_templated(
+                            m_open, m_close, tag_contents
+                        )
+                self.update_inside_set_or_macro(block_type, tag_contents)
+                m_strip_right = regex.search(
+                    r"\s+$", raw, regex.MULTILINE | regex.DOTALL
+                )
+                if raw.startswith("-") and m_strip_right:
                     # Right whitespace was stripped. Split off the trailing
                     # whitespace into a separate slice. The desired behavior is
```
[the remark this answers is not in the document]
barrywhart: At this point, we're inside this `if` block:
```
if elem_type.endswith("_end") or elem_type == "raw_begin":
```

To confirm this, I set a breakpoint in the tests, and when I hit it, I saw the following:
```
> /Users/bhart/dev/sqlfluff/src/sqlfluff/core/templaters/slicers/tracer.py(316)analyze()
-> pass
(Pdb) print(raw)
-}}
  
(Pdb) print(self.raw_str)
SELECT
  {{ 'col1,' -}}
  col2

```

src/sqlfluff/core/templaters/slicers/tracer.py
barrywhart: I looked at doing that a few different ways. It's definitely doable, but the function ended up taking something like 5 parameters, and I couldn't find a reasonable way to get that down. (I tried to limit the new functions to 3 or fewer parameters.) The logic here is not complex, but it depends on a lot of state, so it seemed like _maybe_ it wasn't improving the code. Thoughts?

src/sqlfluff/core/templaters/slicers/tracer.py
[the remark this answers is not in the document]
tunetheweb: Yeah it's the `or elem_type == "raw_begin":` that concerned me.

Maybe add another `if elem_type.endswith("_end")` to this `if statement`?

```suggestion
                if elem_type.endswith("_end") and raw.startswith("-") and m_strip_right:
                    # Right whitespace was stripped after closing block. Split off the trailing
                    # whitespace into a separate slice. The desired behavior is
```

src/sqlfluff/core/templaters/slicers/tracer.py
barrywhart: Are you asking if we don't need to keep the individual `lex` results as separate strings?

The reason I did that is that Jinja doesn't just split on whitespace, so by keeping things as Jinja returns them, the code is more robust. Here's a test case for why:
```
{% set col= "col1" %}
SELECT {{ col }}
```

Note there's no space after `col`. Before switching to `str_parts`, we needed special handling since we might see any of:
```
{% set col= "col1" %}
{% set col = "col1" %}
{% set col ="col1" %}
```

This resulted in ugly, fragile code. Now it "just works"™️.

src/sqlfluff/core/templaters/slicers/tracer.py
barrywhart: I'll do that. Previously, I had just mechanically pasted this code after pulling it out of a larger function. 👍

src/sqlfluff/core/templaters/slicers/tracer.py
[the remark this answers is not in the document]
barrywhart: I'll make that change. In the future, maybe there's a way to break apart this section of code. It's a bit trickier to reason about since it mixes handling of two different cases (block end and "raw").

src/sqlfluff/core/templaters/slicers/tracer.py
barrywhart: I'll add the above as a code comment.

barrywhart: str_buff is required, but we prefer to use str_parts because it makes better use of Jinja lexing.

(Not sure I'm fully understanding, though.)

If you feel like tinkering with it, the tests in `test/core/templaters/jinja_test.py` are pretty thorough.

Issue #946, opened 2021-04-10
Resolve whitespace interdependencies between rules.


alanmcruickshank:
This issue tracks the conversation initially raised in #905 and #934 . Several rules move around sections of code, and either end up recreating sections of L003, or they disregard configuration and let L003 pick up the pieces.

My intent so far has been the latter, in the interest of keeping dependencies clean, *however* I think we need to revisit what the rules are here because it leads to a few odd cases.

As far as I see it we have a few options:
1. *The simple option*: Only L003 deals with indentation and only L039 deals with excess whitespace, other rules may also add additional functionality, but rules should assume that these rules will pick up and enforce any configuration. If we choose this option I think we should document this choice more clearly and potentially have some warnings around any configuration values which detail the implications of disabling some of these "catchall" rules.
2. *The DRY option*: Whitespace and indentation logic should be made generic and then can be called by other rules more specifically. While rules like L003 and L039 might apply that logic fairly broadly, other rules may apply exactly the same logic on specific lines or regions when reformatting code so that it remains consistent. This option results in the cleanest tests, and the least duplication. It's probably the most complex to achieve tidily, and also runs the risk of some quite subtle bugs being introduced by that underlying logic if it misbehaves in some cases.
3. *The Deluxe option*: We require that any rule which modifies positioning *also* manages it's own whitespace. This could lead to significant duplication, but does keep rules more separate and potentially keeps the whitespace logic simpler in each place that it exists. It does lead to probably some boilerplate code.

Based on conversations so far, I feel like people are leaning toward option 2? Any more viewpoints which we need to take into account on this?

@pwildenhain @barrywhart @NiallRees

alanmcruickshank:
While I was pondering this, I had a potentially neat idea on how to implement option 2 without adding a load of extra cruft to rules which want to edit code. We could add an additional kwarg to `LintFix` called `reflow`. It defaults to `False` but if set to `True` it would reflow all the whitespace _within the segment where the fix was applied_.

We might have to deal with a few edge cases, but I think this could mostly allow rules to do things like _"insert a thing here"_ or _"swap these two items"_ and not need to worry about whitespace in doing so.

This could even simplify L003 and L039, when then only have to **detect** issues, leaving `reflow` to selectively sort out whitespace if it's found not up to scratch. Thoughts?

pwildenhain:
`reflow` sounds neat, I'm up to see what that looks like 💪🏻 I agree that option 2 is most desirable and sustainable

[...]

PR #2884, opened 2022-03-18
L003  Refactor V2

OTooleMichael:
L003 Cleanup
- [ ] NEEDS RECONSIDERATION: Separating concerns (improved)
- [x] Improving typing
- [x] using functions & subfunctions
- [x] un-nesting code blocks (ish)
- [x] remove behaviours generating unparsable trees
- [x] fix incorrect tests
- [x] Reduces L003 breaking the parse tree
- [x] stop generating L002 errors
- [x] performance improvements - In long queries prevents exponential? [might just be log vs linear] time increase
- [ ] FOR Future PR: Fix Indentation for queries like:

```sql
SELECT
    MAX(NULLIF(FN_X(
        a
    ))) AS this_in_good_indenting
FROM table
```

### Pull Request checklist


src/sqlfluff/rules/L003.py
```diff
@@ -318,96 +407,57 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
             return LintResult(memory=memory)
 
-        if raw_stack and raw_stack[-1] is not context.segment:
-            raw_stack = raw_stack + (context.segment,)
-        res = self._process_raw_stack(
-            raw_stack,
-            memory,
+        line_summaries = self._process_raw_stack(
+            raw_stack=raw_stack,
+            memory=memory,
```
OTooleMichael, line 439: It should be possible to calculate all line on a single pass, (done in #2857 ) - I think this would improve performance further and simplify reasoning about the behaviour. (given a stream of tokens there is only one correct indent so we should be able to also avoid multiple passes by `self.crawl`

src/sqlfluff/rules/L003.py
```diff
@@ -318,96 +407,57 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
             tab_space_size=self.tab_space_size,
             templated_file=context.templated_file,
         )
+        memory.line_summaries = line_summaries
+        trigger_segment = memory.trigger
+        memory.trigger = None
+        if line_summaries and trigger_segment:
+            last_line_no = max(line_summaries.keys())
+            this_line = line_summaries[last_line_no]
```
OTooleMichael, line 448: @barrywhart What say you to a Small data class for the Dict of _LineSummaries that knows a few of these kinds of get methods

src/sqlfluff/rules/L003.py
```diff
@@ -416,113 +466,76 @@ def _process_current_line(
...
                 # Or they're if the indent balance is the same and the indent is the
                 # same AND the previous line was a hanger
                 or (
-                    this_line["indent_size"] == res[last_code_line]["indent_size"]
-                    and this_line["indent_balance"]
-                    == res[last_code_line]["indent_balance"]
-                    and last_code_line in memory["hanging_lines"]
+                    this_line.indent_size == last_line.indent_size
+                    and this_line.anchor_indent_balance
+                    == last_line.anchor_indent_balance
```
OTooleMichael, line 497: `anchor_indent_balance` replaces `indent_balance` in many contexts - the variable was holding two different concepts which were hard to reason about.

`anchor_indent_balance`  => Is the indent_balance state at the anchor element (first el after a lines starting whitespace) - we do most calculations with the current also using this.
`indent_balance` => The indent balance after the whole line has been calculated/considered

src/sqlfluff/rules/L003.py
barrywhart: Sounds great to me. I've been adding lots of data classes, named tuples, and type annotations the past year or two, especially after we stopped supporting Python 2.7.

OTooleMichael: Yea it's more a question of if the methods would be "too involved" I think it's safer to keep data inert (this code already suffers heavily from being mutative and being confused about what is "doing" and what is just data)

src/sqlfluff/rules/L003.py
```diff
@@ -416,358 +466,220 @@ def _process_current_line(
...
-                                    ],
-                                ),
-                            ],
-                        )
-                elif this_indent_num < comp_indent_num:
-                    memory["problem_lines"].append(this_line_no)
+                        break  # pragma: no cover
+
+                if b_num < indent_diff:
+                    # It doesn't. That means we *should* have an indent when
```
OTooleMichael, line 666: This Code is not correct at all. It takes Close brackets into account and not DEDENTs, the loop[ing mechanism is kind of crazy and if it "works" it is only covering exactly the tests related and the right outcome is basically by chance

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,180 @@
...
+
+    line_no: int = 0
+    templated_line: Optional[int] = None
+    line_buffer: List[BaseSegment] = dataclasses.field(default_factory=list)
+    indent_buffer: List[BaseSegment] = dataclasses.field(default_factory=list)
+    indent_size: int = 1
+
+    # The final balance of a line once fully considered
+    indent_balance: int = 0
+    # The indent as it was once we say this lines "Anchor"
```
barrywhart, line 34:
```suggestion
    # The indent as it was once we see this lines "Anchor"
```

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,180 @@
...
+    anchor_indent_balance: int = 0
+    hanging_indent: Optional[int] = None
```
barrywhart, line 36: It could be helpful to reorder and group these fields a bit differently, by category, e.g.:
* Move `line_anchor` adjacent to `anchor_indent_balance`
* These in a separate section: `templated_line_type`, `is_comment_line`, `is_empty_line`

Not sure if these are exactly right, just thinking to group the fields by functional area.

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,180 @@
...
+        self.clean_indent = _is_clean_indent(self.line_buffer)
+        self.line_buffer = []
+        self.line_indent_stack = []
+        self.indent_size = 0
+        self.hanger_pos = None
+        self.line_anchor = None
+        self.templated_line_type = None
+        return self
+
+    def from_memo_line(self, line_no: int, templated_file: Optional[TemplatedFile]):
```
barrywhart, line 81: Does "memo" here mean the same thing as "memory line" in the previous function?
```suggestion
    def from_memory_line(self, line_no: int, templated_file: Optional[TemplatedFile]):
```

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,180 @@
...
+            )
+            for seg in copied_line_buffer
+        )
+        is_empty_line = not any(
+            elem.is_code or elem.is_type("placeholder") for elem in copied_line_buffer
+        )
+        line_summary = self.__class__(
+            line_no=line_no,
+            templated_line=self.templated_line,
+            # Using slicing to copy line_buffer here to be py2 compliant
```
barrywhart, line 97: The comment seems misplaced and/or obsolete -- no copying happening here, and we no longer support Python 2.

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,180 @@
...
+            # or has an indent in the initial whitespace.
+            clean_indent=self.clean_indent,
+            # Solidify expensive immutable characteristics
+            templated_line_type=_get_template_block_type(
+                copied_line_buffer, templated_file
+            ),
+            is_comment_line=is_comment_line,
+            is_empty_line=is_empty_line,
+        )
+        # Rest this object so it can continue to be used to track current state
```
barrywhart, line 114:
```suggestion
        # Reset this object so it can continue to be used to track current state
```

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,180 @@
...
+    ):
+        """Create a Line state of this line upon reaching the anchor."""
+        self.anchor_indent_balance = self.indent_balance
+        self.indent_size = _indent_size(
+            self.indent_buffer,
+            tab_space_size=tab_space_size,
+        )
+        self.line_anchor = anchor
+
+        return self
```
barrywhart, line 131: Do we need this `return` statement? There's no type annotation about it returning anything, and the only place I see it being called, the return value is not used.

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,180 @@
...
 from sqlfluff.core.rules.base import BaseRule, LintResult, LintFix, RuleContext
 from sqlfluff.core.rules.doc_decorators import (
     document_fix_compatible,
     document_configuration,
 )
 from sqlfluff.core.templaters import TemplatedFile
+from sqlfluff.core.templaters.base import RawFileSlice
+
+
+@dataclasses.dataclass
```
barrywhart, line 18: Overall, big 👍 for adding these classes and moving some of the logic here.

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,180 @@
...
+        break
+
+    return False
+
+
+@dataclasses.dataclass
+class _Memory:
+    # problem_lines keeps track of lines with problems so that we
+    # don't compare to them.
+    problem_lines: List[int] = dataclasses.field(default_factory=list)
```
barrywhart, line 167: I wonder if some of these could be sets, rather than list. Sets are much more efficient for checking whether an item is in it, but I'm not sure they preserve the sequence of items, if that matters.

src/sqlfluff/rules/L003.py
```diff
@@ -56,241 +221,165 @@ def _make_indent(
...
-            raw = ""
-            for raw_slice in Segments(
-                elem, templated_file=templated_file
-            ).raw_slices.select(loop_while=rsp.is_slice_type("literal")):
-                # Compute and append raw_slice's contribution.
-                raw += sp.raw_slice(elem, raw_slice)
-
-        # convert to spaces for convenience (and hanging indents)
-        return raw.replace("\t", " " * tab_space_size)
+        return _segment_length(elem, tab_space_size)
```
barrywhart, line 235: This seems odd -- a one-lined function that forwards to another function. Should we consider eliminating this?

src/sqlfluff/rules/L003.py
```diff
@@ -56,241 +221,165 @@ def _make_indent(
...
 
     @staticmethod
     def _indent_size(segments: Sequence[BaseSegment], tab_space_size: int = 4) -> int:
-        indent_size = 0
-        for elem in segments:
-            raw = Rule_L003._segment_length(elem, tab_space_size)
-            indent_size += len(raw)
-        return indent_size
+        return _indent_size(segments, tab_space_size)
```
barrywhart, line 239: Same question as above.

src/sqlfluff/rules/L003.py
```diff
@@ -416,358 +466,220 @@ def _process_current_line(
...
+                # Does the comparison line have enough unused indent to get us back
+                # to where we need to be? NB: This should only be applied if this is
+                # a CLOSING bracket.
+
+                # First work out if we have some closing brackets, and if so, how
+                # many.
+                b_idx = 0
+                b_num = 0
+                while True:
+                    if len(this_line.line_buffer[b_idx:]) == 0:
```
barrywhart, line 651: As I think you pointed out before, this line looks odd. It's actually slicing the array, but not really "using" the result, only checking the length. It should be possible to replace this line with something that just checks the length. Something like:
```
if len(this_line.line_buffer) < b_idx
```

(I wrote the line above without much thought, definitely double check this before using.)

src/sqlfluff/rules/L003.py
```diff
@@ -776,136 +688,369 @@ def _process_current_line(
...
+            elem.pos_marker.source_slice
+        )
+        return slices or None
+
+
+def _get_template_block_type(
+    line_buffer: List[BaseSegment],
+    templated_file: Optional[TemplatedFile] = None,
+):
+    """Convience fn for getting 'start', 'end' etc of a placeholder line."""
```
barrywhart, line 1011:
```suggestion
    """Convenience fn for getting 'start', 'end' etc of a placeholder line."""
```

src/sqlfluff/testing/rules.py
```diff
@@ -53,7 +53,7 @@ def load_test_cases(
 def get_rule_from_set(code, config):
     """Fetch a rule from the rule set."""
     for r in get_ruleset().get_rulelist(config=config):
-        if r.code == code:
+        if r.code == code:  # pragma: no cover
```
barrywhart, line 56: Surprising that this line would not be covered!

src/sqlfluff/rules/L003.py
```diff
@@ -416,358 +466,220 @@ def _process_current_line(
...
-                "whitespace", "comment", "indent"  # dedent is a subtype of indent
-            )
-            for seg in check_comment_line
-        ):
+        if this_line.is_comment_line:
             # Comment line, deal with it later.
-            memory["comment_lines"].append(this_line_no)
-            self.logger.debug("    Comment Line. #%s", this_line_no)
+            memory.comment_lines.append(this_line_no)
+            self.logger.debug(f"    Comment Line. #{this_line_no:02}")
```
barrywhart, line 480: You don't want to use f-strings for logging:

The reason is, this will build the log string even if it's not going to log it (i.e. due to log level). The old-style formatting with `%s` delays building the string until it knows it's going to output that string. Definite performance impact.

src/sqlfluff/rules/L003.py
```diff
@@ -416,358 +466,220 @@ def _process_current_line(
...
+        )
 
+        if not prev_line:
+            return LintResult(memory=memory)
+        prev_line_no = prev_line.line_no
+        indent_diff = this_line.anchor_indent_balance - prev_line.anchor_indent_balance
+        # Is the indent balance the same?
+        if indent_diff == 0:
+            self.logger.debug(
+                f"    [same indent balance] Comparing to #{prev_line_no:02}"
```
barrywhart, line 545: f-string using for logging

src/sqlfluff/rules/L003.py
OTooleMichael: Just didn't want to change the test suite on this pass.
Will do a pass that removes this sort of thing - and makes a choice between functions, class methods, static methods and metods (we use a random mix at the moment)

src/sqlfluff/rules/L003.py
OTooleMichael: l'll refac this whole section, its basically junk :)

src/sqlfluff/rules/L003.py
alanmcruickshank: Agreed that it should be possible to calculate all on a single pass. I don't think it's true to say there is only one correct indent however:

```
SELECT
    MAX(NULLIF(FN_X(
        a
     ))) AS this_in_good_indenting
FROM table
```

and

```
SELECT
    MAX(NULLIF(
        FN_X(a)
    )) AS this_is_also_good_indenting
FROM table
```

are both valid and different.

src/sqlfluff/rules/L003.py
alanmcruickshank: This is a good change - very clearly explained 👍

src/sqlfluff/rules/L003.py
tunetheweb: Isn’t L003 only concerned with indents rather than adding/changing line breaks? So in that sense the above two examples are the same indents aren’t they?

src/sqlfluff/rules/L003.py
alanmcruickshank: I like the classes but I think I'm a bit confused with the usage of the `_LineSummary` class - as you've mentioned later on it's both representation of data and also mutable.

From my first reading through - I feel like we have a generic `_LineSummary` class, but that we should have a subclass for a `_MemoryLine` so that in our type hints we can be really clear on how the summary is being used.

IMHO, I also think the naming of `memo_line` or `memory_line` is potentially misleading because it sounds similar to the `memory` of the rule and the (quite well thought through) `_Memory` class that already exists. I think calling it something like `reference_line` or `comparison_line` might distinguish it better.

alanmcruickshank (commented):
I haven't reviewed everything here yet because it looks like several things are still in motion. The general move to splitting out things into smaller pieces is excellent though.

I've focussed on the new dataclasses, containers and naming for a first path. @OTooleMichael drop me a shout on slack if it's ever worth getting together on a call to talk through some of the bits here.

src/sqlfluff/rules/L003.py
OTooleMichael: @alanmcruickshank  Is that a typo? the indentation looks the same on both of those. Maybe a space difference.

Not really sure about the arguments that both are valid - I feel if they are both valid, they should be supported as types of indentation as it would far simplify L003.

However thats a mute point cause L003 doesn't accept (which I think is a result of trying to accept too much).
```
SELECT 
    MAX(NULLIF(FN_X(
        a
    ))) AS this_in_good_indenting
FROM table
```
And worse still it forces this kind of thing
```
SELECT 
    FN(MAX(CASE
        WHEN TRUE
            THEN res
            END)) AS this_hurts_my_soul
FROM table
```

OTooleMichael: @tunetheweb  L003 is the indentation one

tunetheweb: Oh I missed the fact all the closing brackets are indented slightly differently and thought @alanmcruickshank was more talking about the different line splits (which IMHO is not so much about indents and so nothing to do with L003).

I’m not sure I agree they are equally valid, and if we want to support the first (indenting closing brackets one space more) then that should be a configurable option so L003 works in either mode rather than trying to accept both.

src/sqlfluff/rules/L003.py
OTooleMichael: Fixedish ?

src/sqlfluff/rules/L003.py
```diff
@@ -416,358 +475,204 @@ def _process_current_line(
...
+            if this_line.indent_size != prev_line.indent_size:
                 # Indents don't match even though balance is the same...
-                memory["problem_lines"].append(this_line_no)
-
-                # The previous indent.
-                desired_indent = "".join(elem.raw for elem in res[k]["indent_buffer"])
-
-                # Make fixes
+                memory.problem_lines.add(this_line_no)
+                desired_indent = "".join(elem.raw for elem in prev_line.indent_buffer)
```
OTooleMichael, line 561: This line is very strange - Tests refused to pass when I try to generate it from the various numbers we have (as has been done with other such examples).
Seems that this would respect a mix of tabs and spaces (which I believe is the test that fails) - Why would that be?

barrywhart: Can you paste the "not working" code here? Not sure I understand.

src/sqlfluff/rules/L003.py
OTooleMichael: So... I went ahead and moved this one class to be a lot closer to immutable. We now generate fresh instances at each step. the concept of a Memory line is also most completely removed. Which is easier to reason about

src/sqlfluff/rules/L003.py
```diff
@@ -416,358 +452,211 @@ def _process_current_line(
...
+                # First work out if we have some closing brackets, and if so, how
+                # many.
+                b_num = 0
+                for elem in this_line.line_buffer:
+                    if not elem.is_code:
+                        continue
+                    if elem.is_type("end_bracket", "end_square_bracket"):
+                        b_num += 1
+                        continue
+                    break  # pragma: no cover
```
OTooleMichael, line 639: Still don't think this section is right.

src/sqlfluff/rules/L003.py
```diff
@@ -776,136 +665,381 @@ def _process_current_line(
...
+                        compared_to=prev_line.line_no,
+                    ),
+                    fixes=[
+                        LintFix.create_before(
+                            trigger_segment,
+                            [
+                                WhitespaceSegment(
+                                    # Make the minimum indent for it to be ok.
+                                    raw=self._make_indent(
+                                        num=comp_indent_num - this_indent_num,
```
OTooleMichael, line 685: why cant we coerce to comp_intent_num instead of add 1 worth of indent element. (tests dont pass if we do this)

test/fixtures/linter/autofix/snowflake/002_previously_parse_tree_damaging/after.sql
```diff
@@ -0,0 +1,14 @@
...
+  WHERE
+    split(foo, '|')[2] REGEXP '^\\d+\\-\\d+\\-\\d+ \\d+\\:\\d+$'
+    OR foo IN ('BAR', 'FOO')
+) AS src
+ON
+  src.foo = tgt.foo
+WHEN MATCHED THEN
+  UPDATE SET
+    tgt.foo = src.foo
+;
```
OTooleMichael, line 14: @barrywhart  This test almost killed me. I think my V1 would have worked but for this.

The test checks that the parse tree is damaged, and given this damage that we apply no fixes.

this PR fixes the damaged Parse tree and cause this set of fixes to work, meaning that the test fails.

Perhaps a test that uses a dummy bad rule to break the tree would be better.

OTooleMichael: I just switched it to its valid FIX output

barrywhart: I'm sorry about the wild goose chase. I don't remember my thought process in creating this test, but clearly I should've documented my intent.

src/sqlfluff/rules/L003.py
OTooleMichael: Sorry I actually figure this out.
```sql
SELECT
\t\s\s\s\s col_1,
col_2 --> [corrects to] ---> \t\s\s\s\s col_2,
```
Reusing the mixed spaces from above was hard coded into one of the tests. As Ive mentioned elsewhere I think this isnt a good idea (adding complexity and compute, for negative benefit)

The test was only flagging / forcing this behaviour on one level of misindent, which lead to the surprising behaviour of line 561 being the only one that couldn't be changed.

New algo always puts fresh spaces or tabs depending on the user preference.
1. Avoids expensive call to .raw
2. removes many branches from the code's algo

src/sqlfluff/rules/L003.py
alanmcruickshank: Ah yes - @tunetheweb is correct - I think I was making the point that given a stream of _code_ tokens there will be potentially multiple valid _whitespace_ configurations. That's because choosing where to split long lines is somewhat complicated.

If what you're actually saying is that given a stream of _code and newline_ tokens, that there is only one valid _indentation_ configuration (or at least only one _ideal_ setup), then **yes** I agree.

tunetheweb: L016 handles line splitting, and L003 only handle indents AFAIK.

Though not sure whether L016 also handles indents for any new splits, or leaves the next pass of L003 to deal with those.

alanmcruickshank: it's a bit of a mess right now. Hence #946.

alanmcruickshank: it's a bit of a mess right now. Hence https://github.com/sqlfluff/sqlfluff/issues/946.

OTooleMichael: This Refac to L003 improves it. But one would need rewrite (and rule reconsideration to fix things overall).

And yes I was saying that given new lines and a stream we can calc the indents

test/fixtures/linter/autofix/snowflake/002_previously_parse_tree_damaging/after.sql
OTooleMichael: No problem - its knowledge formed in the fire. I created an issue to try to help avoid this in the future

test/cli/commands_test.py
```diff
@@ -53,7 +53,7 @@ def invoke_assert_code(
 
 
 expected_output = """== [test/fixtures/linter/indentation_error_simple.sql] FAIL
-L:   2 | P:   4 | L003 | Indentation not hanging or a multiple of 4 spaces
+L:   2 | P:   4 | L003 | Expected 1 indentations, found 0 [compared to L#01]
```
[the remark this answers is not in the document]
OTooleMichael: I agree-ish but to be honest the rule naming has come back to shoot us in the foot. cause your version would read
`L003 | Expected 1 indentations, found 0 [compared to L01]`
L003 and L01 reference completely different concepts. And sadly I think L01 meaning line one is the "correct/common/obvious" one, but I guess we are stuck with L003 for Lint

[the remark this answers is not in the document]
tunetheweb: `L003 | Expected 1 indentations, found 0 [compared to line 01]`? Or too long?

test/cli/commands_test.py
```diff
@@ -53,7 +53,7 @@ def invoke_assert_code(
 
 
 expected_output = """== [test/fixtures/linter/indentation_error_simple.sql] FAIL
-L:   2 | P:   4 | L003 | Indentation not hanging or a multiple of 4 spaces
+L:   2 | P:   4 | L003 | Expected 1 indentations, found 0 [compared to line 01]
```
barrywhart, line 56: Have we settled on a final format for the line numbers? If we're going to use leading zeros, I would suggest allowing for 4 digits (0001). Many files will be more than 99 lines long.

test/fixtures/linter/autofix/snowflake/002_previously_parse_tree_damaging/before.sql
```diff
@@ -5,8 +5,7 @@ select
 from foo.bar
 where
 split(foo, '|')[2] REGEXP '^\\d+\\-\\d+\\-\\d+ \\d+\\:\\d+$'
-OR
-foo IN ('BAR','FOO')
+OR foo IN ('BAR','FOO')
```
barrywhart, line 8: Why'd we change the _input_ file? (Probably not important, just curious.)

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,157 @@
...
+            for key, value in self.__dict__.copy().items()
+            if key not in keys_to_strip
+        }
+        print_dict["raw_line"] = self.template_content
+        return print_dict.__repr__()
+
+    @property
+    def template_content(self):  # pragma: no cover
+        return "".join(
+            seg.raw or getattr(seg, "source_str", "") for seg in self.line_buffer
```
barrywhart, line 67: Which segment classes have a `source_str` property? I did a quick search of the code and couldn't find any (probably overlooking something).

src/sqlfluff/rules/L003.py
barrywhart: Were you still going to add the type annotation? TBH, I would prefer to omit the `return` and type hint in this PR, then maybe revisit these things together in a future PR. It's confusing and not useful for the function to return itself -- someone could easily think it's returning a copy.

src/sqlfluff/rules/L003.py
```diff
@@ -56,241 +198,169 @@ def _make_indent(
...
+            current_line.indent_balance += elem.indent_val  # type: ignore
+            if elem.indent_val > 0:  # type: ignore
+                # a "clean" indent is one where it contains
+                # an increase in indentation? Can't quite
+                # remember the logic here. Let's go with that.
+                current_line.clean_indent = True
+            return current_line
+
+        current_line.set_state_as_of_anchor(elem, tab_space_size)
+        return current_line
```
barrywhart, line 332: Same concern as noted elsewhere -- to avoid confusion, maybe don't return the same object we were passed, then consider adding back the `return` once it's returning a copy.

src/sqlfluff/rules/L003.py
```diff
@@ -776,136 +660,358 @@ def _process_current_line(
...
+            return None
+        # Handle Hanging Indents
+        is_anchor_indent_match = (
+            this_line.anchor_indent_balance == last_line.anchor_indent_balance
+        )
+        is_end_indent_match = this_line.indent_size == last_line.indent_size
+        is_known_hanging_line = last_line.line_no in memory.hanging_lines
+        # There MUST also be a non-zero indent. Otherwise we're just on the
+        # baseline.
+        if not this_line.indent_size > 0:
```
barrywhart, line 760: This comparison looks strange. Can we clean this up to avoid having `not` and `>` in the same expression?

src/sqlfluff/rules/L003.py
```diff
@@ -776,136 +660,358 @@ def _process_current_line(
...
+        desired_indent = "".join(elem.raw for elem in template_line.indent_buffer)
+        first_non_indent_i = len(this_line.indent_buffer)
+        current_anchor = this_line.line_buffer[first_non_indent_i]
+        fixes = self._coerce_indent_to(
+            desired_indent=desired_indent,
+            current_indent_buffer=this_line.indent_buffer,
+            current_anchor=current_anchor,
+        )
+        self.logger.debug(
+            "    !! Indentation does not match #%s." " Fixes: %s",
```
barrywhart, line 835: Can we combine the two strings on this line? I assume this was previously two lines that were combined.

src/sqlfluff/rules/L003.py
```diff
@@ -776,136 +660,358 @@ def _process_current_line(
...
-            return False
-        for idx in range(1, len(current_line)):
-            if segment_info(idx - 1) in (
-                ("placeholder", "block_start"),
-                ("placeholder", "compound"),
-                ("placeholder", "block_mid"),
-            ) and segment_info(idx) == ("indent", None):
-                return True
-        return False
+    def list_segement_and_raw_segement_types(self):
```
barrywhart, line 901:
```suggestion
    def list_segement_and_raw_segment_types(self):
```

src/sqlfluff/rules/L003.py
```diff
@@ -776,136 +660,358 @@ def _process_current_line(
...
+        """Yields the tuple of seg type and underlying type were applicable."""
+        for seg in self.current_line:
+            raw_seg = self.get_raw_slices(seg)
+            yield (seg.type, raw_seg[0].slice_type if raw_seg else None)
```
barrywhart, line 905: Do you think this might be  clearer?
```suggestion
            if raw_seg:
                yield (seg.type, raw_seg[0].slice_type
            else:
                yield None
```

test/cli/commands_test.py
alanmcruickshank: Line numbers in the spec for the start of the line default to three digits with a leading `L`. Line numbers longer than three digits automatically expand.

src/sqlfluff/rules/L003.py
alanmcruickshank: `TemplateSegment` elements have a `source_str` property. This is for templated elements.

src/sqlfluff/rules/L003.py
OTooleMichael: Funny I'd be totally on the other side. Returning and reassigning that return makes it clear what has changed. I'd actually be trying to indicate that this might as well be a new instance - if performance would allow I'd recreate here.

I could move this from being a method to a pure function.

src/sqlfluff/rules/L003.py
OTooleMichael: Funny black doesn't complain I was just quickly trying to invert things into guard statements

src/sqlfluff/rules/L003.py
OTooleMichael: Your suggestion isnt functionally the same. But maybe a middling var ?

src/sqlfluff/rules/L003.py
OTooleMichael: Which two string?

test/fixtures/linter/autofix/snowflake/002_previously_parse_tree_damaging/before.sql
OTooleMichael: I think the current way of indenting this I would consider wrong, and so didn't want to codify it

alanmcruickshank (changes requested):
Thanks for chatting earlier on the hangout - the context was useful for reviewing this.
- The core intent of breaking apart some monolithic code I think is well achieved 👍 .
- That places you've chosen to simplify or cut out functionality all make sense.

There are a few places where I think we can almost go a bit further in this step (see my note on `_find_previous_line`). There are also a few places where I think the naming is misleading - in particular the usage of `current_line` and `from_current_line`. I think we should clarify those names before merging, but the larger rethink of `_LineSummary` (and what the return value of `from_current_line` is) could also wait for your next pass on this, so I'm happy to merge with the structure as is for now (it would also make it easier to review that rethink as a separate pass).

On some relatively blunt testing I think it's faster too 👍 .

As a minor nit I also thing that the `profile.prof` is probably mistakenly part of this PR - can you add `*.prof` to the `.gitignore` file and remove it from the PR?

src/sqlfluff/rules/L003.py
barrywhart: Oh yes, I misread it. Variable might be good. Python ternary expressions are sometimes great and sometimes hard to read.

src/sqlfluff/rules/L003.py
barrywhart: Look carefully at the `"` on this line. I think there are 4, thus two different strings adjacent to one another.

src/sqlfluff/rules/L003.py
```diff
@@ -318,96 +388,57 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
-            # (since there may not be a newline on the very last line)..
+        if segment.is_type("newline"):
+            memory.in_indent = True
+        elif memory.in_indent:
+            is_non_raw_segment = bool(segment.segments)
+            is_placeholder = segment.is_meta and segment.indent_val != 0  # type: ignore
+            if not (segment.is_whitespace or is_non_raw_segment or is_placeholder):
+                memory.in_indent = False
+                # First non-whitespace element is our trigger
+                memory.trigger = segment
```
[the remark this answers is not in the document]
OTooleMichael: I think the intent is more that it has to be a raw segment. but I've switched it to has_childern

src/sqlfluff/rules/L003.py
```diff
@@ -1,15 +1,157 @@
...
+        print_dict["raw_line"] = self.template_content
+        return print_dict.__repr__()
+
+    @property
+    def template_content(self):  # pragma: no cover
+        return "".join(
+            seg.raw or getattr(seg, "source_str", "") for seg in self.line_buffer
+        )
+
+    def from_current_line(self, line_no: int, templated_file: Optional[TemplatedFile]):
```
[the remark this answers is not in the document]
OTooleMichael: I went with `finalise` and `working_state` for the variable. I feel that makes it clear as to your point.

I actually tried for quite a while to figure a pattern with two distinct classes where one is frozen, but anything I tried added significant overhead which just felt like it was making it more complicated.

cause _WorkingLineState needs to be able to create _LineSummary and vis-vera and also linting complains about inheriting and freezing (I think it would actually function at runtime)

alanmcruickshank (approved):
Top Work. 💯 

So glad that this rule is getting some love, and thanks for taking the time to handle what has been quite a hefty review process.

PR #2889, opened 2022-03-19
Row typecasting in hive `cast` function

KulykDmytro:
Extended version of ansi's `FunctionSegment` to add support of row typecasting
```
cast(row(val1, val2) as row(a integer, b integer))
```

### DIFF
--- a/src/sqlfluff/dialects/dialect_hive.py
+++ b/src/sqlfluff/dialects/dialect_hive.py
@@ -1,6 +1,7 @@
 """The Hive dialect."""
 
 from sqlfluff.core.parser import (
+    AnyNumberOf,
     BaseSegment,
     Sequence,
     Ref,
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_hive.py
```diff
@@ -620,3 +621,83 @@ class MsckTableStatementSegment(BaseSegment):
...
+            Sequence(
+                "ROW",
+                Bracketed(
+                    Delimited(
+                        Sequence(
+                            Ref("BaseExpressionElementGrammar"),
+                            Ref("DatatypeIdentifierSegment", optional=True),
+                        ),
+                    ),
+                ),
```
tunetheweb, line 668: This does not need to be wrapped in another `sequence(...)`.

Also this is causing some of the rules to fail as there is no `function_name` as they expect in a Function.

I tried using `ephemeral_name` but couldn't get that to work so best I could come up with it using `StringParser("ROW", KeywordSegment, type="function_name"),` which is a bit of a fudge, but seems to work:

```suggestion
            # This unusual syntax is used to cast the Keyword ROW to
            # to the function_name to avoid rule linting exceptions
            StringParser("ROW", KeywordSegment, type="function_name"),
            Bracketed(
                Delimited(
                    Sequence(
                        Ref("BaseExpressionElementGrammar"),
                        Ref("DatatypeIdentifierSegment", optional=True),
                    ),
                ),
            ),
```

tunetheweb (changes requested):

tunetheweb (approved):

PR #2893, opened 2022-03-21
Hive: fix to align with L027 rule (rowtype attribute name)

KulykDmytro:
In `Hive` dialect there is a false positive violation detected in `rowtype` definition when query/subquery operates with join
```sql
select cast(row(t1.attr, t2.attr) as row(fld1 double, fld2 double)) as flds
from sch.tab1 as t1
join sch.tab2 as t2 on t2.id = t1.id
```
`fld1`, `fld2` interpret as table column alias rather than rowtype attribute name

### DIFF
--- a/test/fixtures/rules/std_rule_cases/L027.yml
+++ b/test/fixtures/rules/std_rule_cases/L027.yml
@@ -208,3 +208,15 @@ test_pass_column_and_alias_same_name_2_tsql:
   configs:
     core:
       dialect: tsql
+
+test_pass_rowtype_with_join:
+  # Wrongly interpret rowtype attributes as field alias
+  # when more than one tables in join
+  pass_str: |
+    select
+        cast(row(t1.attr, t2.attr) as row(fld1 double, fld2 double)) as flds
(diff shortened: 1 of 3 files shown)


test/fixtures/rules/std_rule_cases/L027.yml
```diff
@@ -208,3 +208,15 @@ test_pass_column_and_alias_same_name_2_tsql:
   configs:
     core:
       dialect: tsql
+
+test_pass_rowtype_with_join:
+  # Wrongly interpret rowtype attributes as field alias
+  # when more than one tables in join
```
tunetheweb, line 214:
```suggestion
  # Check we don't wrongly interpret rowtype attributes
  # as field alias when more than one tables in join
```

tunetheweb (approved):
LGTM.

One nit on the comment.

test/fixtures/rules/std_rule_cases/L027.yml
KulykDmytro: 👍

PR #2896, opened 2022-03-22
Hive: support for complex types in `cast` `rowtype` definition

KulykDmytro:
Adding support for complex datatypes in `cast` `rowtype` definition:
```
select
    cast(row(col1, col2) as row(a bigint, b decimal(23, 2)))
from sch.tbl;
```

### DIFF
--- a/src/sqlfluff/dialects/dialect_hive.py
+++ b/src/sqlfluff/dialects/dialect_hive.py
@@ -687,7 +687,6 @@ class FunctionSegment(BaseSegment):
                 Delimited(
                     Sequence(
                         Ref("BaseExpressionElementGrammar"),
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_hive.py
```diff
@@ -687,7 +687,6 @@ class FunctionSegment(BaseSegment):
                 Delimited(
                     Sequence(
                         Ref("BaseExpressionElementGrammar"),
-                        Ref("DatatypeIdentifierSegment", optional=True),
```
KulykDmytro, line 690: this one not needed cause `row` as a function here accepting only sequence of expressions

tunetheweb (approved):

PR #2908, opened 2022-03-24
Fix bug where L043 corrupts SQL

### LINKED ISSUE #2903: L043 doesn't consider THEN statement in evaluation, alters selection logic
### Search before asking

### What Happened

When fixing in rule L043
```sql
SELECT
    COUNT(DISTINCT   
        CASE
            WHEN item.submitted_timestamp IS NOT NULL
            THEN item.sitting_id
        END
    ) AS num_submitted
FROM table
```
is modified to :
```sql
SELECT
    COUNT(DISTINCT
        response.submitted_timestamp
    ) AS num_submitted
FROM table
```

This is a different expression and means any submissions with the same timestamp will not be considered as distinct counts.

### Expected Behaviour

The `THEN` statement should probably be inspected to see if it's a different expression to the NOT NULL check in the WHEN.

### Observed Behaviour

If the THEN statement isn't inspected as part of the rule, this is going to regularly be changing the logic of CASE statements.

### How to reproduce

As above.

### Dialect

postgres

### Version

0.11.1

### Configuration

```
[sqlfluff]
dialect = postgres
templater = dbt

# Excluded rules (see https://docs.sqlfluff.com/en/stable/rules.html):
# L009: File end with newline - suggested in the docs when using the DBT templater
# L016: Line too long - bugged, doesn't honour line lengths in lines containing templating syntax
# L025: No unused table aliases - bugged, doesn't consider table aliases used in join criteria or subqueries, which can break queries
# L031: No table aliases in joins or froms - tables aliases are required in DBT if you're using ref or src
# L034: Column ordering - I think query authors should have explicit control over the order of columns
# L042: No subqueries in joins or froms - I think query authors should be able to use subqueries in JOIN clauses
exclude_rules = L009,L016,L025,L031,L034,L042

[sqlfluff:rules]
tab_space_size = 2

[sqlfluff:templater:dbt]
target = dev
```

### Code of Conduct

barrywhart:
Fixes #2903

### Pull Request checklist

  - Other.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,3 +1,4 @@
 [pytest]
 markers =
     dbt: marks tests needing the "dbt" plugin (deselect with '-m "not dbt"')
+    integration_test: marks integration tests
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -343,7 +343,7 @@ class BaseSegment:
         return self.get_raw_segments()
 
     @cached_property
-    def raw_segments_upper(self) -> Optional[str]:
+    def first_non_whitespace_raw_segments_upper(self) -> Optional[str]:
         """Returns the first non-whitespace subsegment of this segment."""
         for seg in self.raw_segments:
             if seg.raw_upper.strip():
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -663,7 +663,7 @@ class BaseSegment:
             "raw_upper",
             "matched_length",
             "raw_segments",
-            "raw_segments_upper",
+            "first_non_whitespace_segment_raw_upper",
         ]:
             self.__dict__.pop(key, None)
 
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -343,7 +343,7 @@ class BaseSegment:
         return self.get_raw_segments()
 
     @cached_property
-    def first_non_whitespace_raw_segments_upper(self) -> Optional[str]:
+    def first_non_whitespace_segment_raw_upper(self) -> Optional[str]:
         """Returns the first non-whitespace subsegment of this segment."""
         for seg in self.raw_segments:
             if seg.raw_upper.strip():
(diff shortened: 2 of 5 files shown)


pytest.ini
```diff
@@ -1,3 +1,4 @@
 [pytest]
 markers =
     dbt: marks tests needing the "dbt" plugin (deselect with '-m "not dbt"')
+    integration_test: marks integration tests
```
barrywhart, line 4: Unrelated minor fix: Avoids a pytest warning, "undeclared mark"

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -343,7 +343,7 @@ def raw_segments(self) -> List["BaseSegment"]:
         return self.get_raw_segments()
 
     @cached_property
-    def raw_segments_upper(self) -> Optional[str]:
+    def first_non_whitespace_raw_segments_upper(self) -> Optional[str]:
```
barrywhart, line 346: The name of this property was very deceptive. Updated it to reflect what it actually does.

src/sqlfluff/rules/L043.py
```diff
@@ -240,10 +240,7 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
                         description="Unnecessary CASE statement. "
                         "Use COALESCE function instead.",
                     )
-                elif (
-                    column_reference_segment.raw_segments_upper
-                    == then_expression.raw_segments_upper
-                ):
+                elif column_reference_segment.raw_upper == then_expression.raw_upper:
```
barrywhart, line 243: Not sure why the original version used the other, more obscure property; `raw_upper` is more commonly used and works just fine.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -343,7 +343,7 @@ def raw_segments(self) -> List["BaseSegment"]:
         return self.get_raw_segments()
 
     @cached_property
-    def raw_segments_upper(self) -> Optional[str]:
+    def first_non_whitespace_segment_raw_upper(self) -> Optional[str]:
```
barrywhart, line 346: This property was poorly named. Updated the name to better reflect what it does.

tunetheweb: Does this not also need changed on line 666 (the sign of the devil!) to ensure the cached version is recalculated?

barrywhart: Great catch! I used an automated tool to rename this, and it only updated the actual code.

If we added or changed cached fields very often, it'd probably be worthwhile to compute this list automatically. It's pretty rare, though.

barrywhart: Fixed.

PR #2930, opened 2022-03-26
To replace base dialect segment class, must subclass or provide same stuff

### LINKED ISSUE #2902: Add internal check: Segments in "child" dialects must inherit from the parent dialect
### Search before asking

### Description

@OTooleMichael suggests:
>One should not be able to add a segment class matching a base name if it's not a subclass. Ala the replace decorator should complain

If this proves to be too strict, we could attempt to determine if the segment has any additional fields or methods beyond just `match_grammar` and `parse_grammar`, because that's mainly what we're trying to preserve. In the PR linked below, a `get_identifier()` method on ANSI's `CTEDefinitionSegment` was being lost in the Postgres dialect.

Reference: PR #2899

### Use case

Warn at dialect load time if a dialect violates this policy. 

### Dialect

Postgres was the inspiration, but really all dialects.

### Code of Conduct

barrywhart:
Fixes #2902

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_exasol.py
+++ b/src/sqlfluff/dialects/dialect_exasol.py
@@ -3759,7 +3759,7 @@ class FunctionScriptStatementSegment(BaseSegment):
 
 
 @exasol_dialect.segment(replace=True)
-class StatementSegment(BaseSegment):
+class StatementSegment(ansi_dialect.get_segment("StatementSegment")):  # type: ignore
     """A generic segment, to any of its child subsegments."""
 
     type = "statement"
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -171,3 +171,9 @@ ignore_words = None
 [sqlfluff:rules:L062]
 # Comma separated list of blocked words that should not be used
 blocked_words = None
+
+[sqlfluff:rules:L063]
+# Data Types
+extended_capitalisation_policy = consistent
+# Comma separated list of words to ignore for this rule
+ignore_words = None
--- a/src/sqlfluff/core/dialects/base.py
+++ b/src/sqlfluff/core/dialects/base.py
@@ -148,7 +148,9 @@ class Dialect:
                 ):
                     base_dir = set(dir(self._library[n]))
                     cls_dir = set(dir(cls))
-                    missing = base_dir.difference(cls_dir)
+                    missing = set(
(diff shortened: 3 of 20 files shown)


src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -3759,7 +3759,7 @@ class FunctionScriptStatementSegment(BaseSegment):
 
 
 @exasol_dialect.segment(replace=True)
-class StatementSegment(BaseSegment):
+class StatementSegment(ansi_dialect.get_segment("StatementSegment")):  # type: ignore
```
tunetheweb, line 3762: Is it better to do this? Or just add the `get_table_references` function to this Segment instead? We normally do the latter.

barrywhart: If the code is the same, I'd say this is better. Code reuse is one of the main benefits of inheritance, and copy paste. Especially in this case, where we'd be copying a good bit of code (about 10 lines).

tunetheweb: Oh definitely wasn’t suggesting copying and pasting the whole function. It’s whether you inherit the class or point the function to the ansi one. We tend to do the latter. But maybe that’s not the best way and this is better.

barrywhart: Got it. I don't have a strong feeling. One reason I did it this way is that it always works, even if the base class had several functions -- inheriting gives you access to them all, while the other way might require you to copy individual things, one by one.

TBH, I think it'd be good if we had a cleaner way to do this. Currently, the code is ugly either way. We could probably support something like this pretty easily:
```
class StatementSegment(ansi_dialect.StatementSegment)
```

That might allow us to omit the `# type: ignore` comments too.

tunetheweb: Yeah was curious why we needed to ignore the type here?

barrywhart: The type checker sees the function call `get_segment()` being used to specify the base class for the class we're creating, and it thinks, "oh, this is tricky. That function could return anything, so I can't analyze this the usual way, where I assume the base class is a known, specific thing." In reality, `get_segment()` is predictable, too, but the type checker doesn't know about it. I'm thinking if we changed this so we were using `.`, the type checker would assume it's more predictable. (It's not really!)

tunetheweb (approved):

PR #2933, opened 2022-03-27
Simplify segment creation and inheritance in dialects

barrywhart:

This is a big PR, but most of the changes are very mechanical.
- [X] Removes the `@segment()` decorator from the `Dialect` class, replacing it with a more automated approach.
- [X] Segment inheritance now uses normal Python syntax. Benefit: `mypy` understands this syntax, eliminating the need for a number of `# type: ignore` comments.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1001,15 +1003,9 @@ class IntoClauseSegment(BaseSegment):
 
 
 @postgres_dialect.segment(replace=True)
-class UnorderedSelectStatementSegment(BaseSegment):
+class UnorderedSelectStatementSegment(ansi.UnorderedSelectStatementSegment):
```
barrywhart, line 1006: By using actual Python class inheritance, we eliminate a lot of duplicate code from the base class.

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1136,9 +1121,7 @@ class CreateRoleStatementSegment(BaseSegment):
 
 
 @postgres_dialect.segment(replace=True)
-class ExplainStatementSegment(
-    ansi_dialect.get_segment("ExplainStatementSegment")  # type: ignore
-):
+class ExplainStatementSegment(ansi.ExplainStatementSegment):
```
barrywhart, line 1124: Here, we were already using real inheritance, but using a function to look up the base class. This caused `mypy` to complain, requiring a `# type: ignore` comment. Now we can remove that.

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1001,15 +1003,9 @@ class IntoClauseSegment(BaseSegment):
...
 @postgres_dialect.segment(replace=True)
-class UnorderedSelectStatementSegment(BaseSegment):
+class UnorderedSelectStatementSegment(ansi.UnorderedSelectStatementSegment):
     """Overrides ANSI Statement, to allow for SELECT INTO statements."""
 
-    type = "select_statement"
-
-    match_grammar = ansi_dialect.get_segment(
-        "UnorderedSelectStatementSegment"
-    ).match_grammar.copy()
```
barrywhart, line 1011: We don't need to _copy_ it because we now _inherit_ it.

barrywhart:
@alanmcruickshank: Could you take a look at this approach and see what you think? It really cleans up the segment inheritance:
* Shorter code
* Makes the `mypy` type checker happy

IIRC, at some point we had concerns about circular dependencies with the dialect files, hence the creation of the `load_raw_dialect()` function. Do you recall the details? Does this PR potentially re-introduce the same issues?

barrywhart:
@tunetheweb: After our discussion on the last PR, I started this experiment. I asked Alan for his feedback, since it touches on something that had caused us issues a year or more ago.

tunetheweb:
I was thinking about this some more and wondering could we do more with the `replace=True` option? Could there be an `inherit=True` option as well? Or should replace=True do this by default?

barrywhart:
About `replace` and `inherit`, I wonder if we could possibly eliminate the need for the `@segment` decorator entirely. A dialect is basically the module contents. Could we just define the module and create the dialect automatically? Overall, if we can make more use of built-in Python language features like modules and classes, that can make it easier to work on dialects (for example, mypy and Python IDEs will help with writing and checking the code).

alanmcruickshank (approved):
Nice. Love it 👍 

Good work.

src/sqlfluff/core/dialects/base.py
```diff
@@ -196,6 +154,19 @@ def replace(self, **kwargs: DialectElementType):
                 raise ValueError(f"{n!r} is not already registered in {self!r}")
             self._library[n] = kwargs[n]
 
+    def add_update_segments(self, module_dct):
+        """Scans module dictionary, adding or replacing segment definitions.
+
+        Each dialect module (.py file) should call this function at the END OF
+        THE FILE, after all the segment classes have been defined.
```
tunetheweb, line 161: Anyway we could automate this so we don't have to add? e.g. by adding to `load_raw_dialect`?

tunetheweb (approved):
One question, but in general LGTM.

This is a fairly big (but good!) change to dialects (even if the code change itself is small). So flagging to @sqlfluff/sqlfluff-maintainers (especially @alanmcruickshank).

src/sqlfluff/core/dialects/base.py
barrywhart: Good call -- calling it from `load_raw_dialect()` seems to be working fine. I'll update.

PR #2936, opened 2022-03-28
Snowflake: Add `CREATE STREAM` support

### LINKED ISSUE #2934: Snowflake: create STREAM fails parser
### Search before asking

### What Happened

In our Snowflake implementation, we use [schemachange](https://github.com/Snowflake-Labs/schemachange) to maintain database migration scripts as our DCM. In our implementation, we use STREAMs as part of our ingest process.

Per the [CREATE STREAM documentation](https://docs.snowflake.com/en/sql-reference/sql/create-stream.html#syntax) on Snowflake's site, a stream must be created on an object.

However, when SQLFluff lints the CREATE STREAM statement using the `snowflake` dialect, we get a parsing error as follows.

Original query:
```
CREATE STREAM IF NOT EXISTS LOAD_DOMAIN.DOMAINTABLE_STREAM ON TABLE LOAD_DOMAIN.DOMAINTABLE APPEND_ONLY = TRUE;
```

Linter result:
```
L:   1 | P:  70 |  PRS | Line 1, Position 70: Found unparsable section: ' ON
                       | TABLE LOAD_CAMPAIGN.CAMPAIGNCONTROLL...'
```

Given how STREAM has a necessary dependency on another object, I wonder if (in this dialect) it should be split away from the generic CREATE function, in the same way that creating tables, stages, etc. are their own function.

### Expected Behaviour

The query as written is correct syntax, and should clear the checks.

### Observed Behaviour

Linter result:
```
L:   1 | P:  70 |  PRS | Line 1, Position 70: Found unparsable section: ' ON
                       | TABLE LOAD_CAMPAIGN.CAMPAIGNCONTROLL...'
```

### How to reproduce

Any CREATE STREAM ... ON <object> statement will fail.

### Dialect

snowflake

### Version

python 3.9
sqlfluff-0.11.2-py3-none-any.whl

### Configuration
```
# .sqlfluff
# Override certain default config, as found here:
# https://docs.sqlfluff.com/en/stable/configuration.html#default-configuration

[sqlfluff]
dialect = snowflake
exclude_rules = L009, L044

[sqlfluff:rules]
comma_style = leading
```

### Code of Conduct

HeyZiko:
snowflake dialect: Added and registered CREATE STREAM segment, and created test fixture

fixes #2934 

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3250,6 +3250,99 @@ class AlterStageSegment(BaseSegment):
...
+
+    match_grammar = Sequence(
+        "CREATE", "STREAM"
+    )
+
+    parse_grammar = Sequence(
+        "CREATE",
+        Ref("OrReplaceGrammar", optional=True),
+        "STREAM",
+        Ref("IfNotExistsGrammar", optional=True),
```
tunetheweb, line 3312: This is the problem.

This says match `CREATE STREAM` and then **just that match** with the parse grammar. Obviously the statement includes more than just that, so this isn't enough in `match_grammar`.

Typically we use `StartsWith` in `match_grammar` and then have a `terminator` argument to that.

However, having a separate `match_grammar` and `parse_grammar` is only useful if you can prune part of the SQL (using `match_grammar`) to make the parsing easier.

Here you're not doing that, so not need to have them separated and can just combine them into a joint `match_grammar`.

```suggestion
    match_grammar = Sequence(
        "CREATE",
        Ref("OrReplaceGrammar", optional=True),
        "STREAM",
        Ref("IfNotExistsGrammar", optional=True),
```

tunetheweb (changes requested):
I've a few other suggestions but think this is your main issue.

Could you also look at the other segments you've created and similarly see if you really need a separate `match_grammar` and `parse_grammar`? I don't think you do.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3250,6 +3250,99 @@ class AlterStageSegment(BaseSegment):
...
+        Sequence("COPY", "GRANTS", optional=True),
+        "ON",
+        OneOf(
+            Sequence(
+                OneOf("TABLE", "VIEW"),
+                Ref("ObjectReferenceSegment"),        
+                OneOf(
+                    Ref("FromAtExpressionSegment", optional=True),
+                    Ref("FromBeforeExpressionSegment", optional=True)
+                ),
```
tunetheweb, line 3316: Here's another thing you need to fix as well:

```suggestion
                OneOf(
                    Ref("FromAtExpressionSegment"),
                    Ref("FromBeforeExpressionSegment"),
                    optional=True
                ),
```

You'd think they would be equivalent, but they are not quite the same and with `OneOf` without the `optional=True` something has to match it (which it doesn't for your simple test case).

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3250,6 +3250,92 @@ class AlterStageSegment(BaseSegment):
...
+class InsertOnlyExpressionSegment(BaseSegment):
+    """A INSERT_ONLY = TRUE expression."""
+
+    type = "insert_only_expression"
+
+    match_grammar = Sequence(
+        "INSERT_ONLY",
+        Ref("EqualsSegment"),
+        Ref("TrueSegment")
+    )
```
tunetheweb, line 3289: There's a question as to whether these need to be separate statements or should just all be directly part of `CreateStreamStatementSegment`.

Having separate statements can be useful if:
1. Same code is used in multiple places
2. Want to be able to give it a `type` so can reference it in API or in rules.
3. Breaks up long code (though counter argument is you can't understand everything in one place).

I don't think 1 and 2 apply here and I could fall either way on 3 in this case. Is there a reason you think they are better as separate segments?

HeyZiko: I'm inclined to agree. I observed elsewhere that most segments were broken out to make it `DRY`, so I thought I'd do the same. However, I've only seen these statements written in this syntax as part of the STREAM object. 
I'll search the snowflake docs to double-check.

One after-thought is that the ALTER statement (which I haven't written into this PR but might choose to if I have the cycles) might reuse these statements.

src/sqlfluff/dialects/dialect_snowflake.py
HeyZiko: Bingo. This was the culprit. I'm uncommenting the rest of my test fixtures and testing locally.

src/sqlfluff/dialects/dialect_snowflake.py
tunetheweb: Well happy to keep them if you think they might be reused. As I say, I'm on the fence about them, rather than thinking they are wrong.

Alternatively they are relatively simple so, even if ALTER statement needs them, could just repeat them. Do want to DRY code for more complex statements, but for simple things a little repetition isn't the worst and has benefits.

Anyway, let's get rest of your tests fixed first and then can come back to this and decide one way or the other.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3250,6 +3250,92 @@ class AlterStageSegment(BaseSegment):
...
+
+
+@snowflake_dialect.segment()
+class ShowInitialRowsExpressionSegment(BaseSegment):
+    """A SHOW_INITIAL_ROWS = TRUE | FALSE expression."""
+
+    type = "show_initial_rows_expression"
+
+    match_grammar = Sequence(
+        "SHOW_INITIAL_ROWS",
```
tunetheweb, line 3273: This will need to be added to `dialect_snowflake_keywords.py`

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3250,6 +3250,92 @@ class AlterStageSegment(BaseSegment):
...
+
+
+@snowflake_dialect.segment()
+class InsertOnlyExpressionSegment(BaseSegment):
+    """A INSERT_ONLY = TRUE expression."""
+
+    type = "insert_only_expression"
+
+    match_grammar = Sequence(
+        "INSERT_ONLY",
```
tunetheweb, line 3286: This will need to be added to `dialect_snowflake_keywords.py`

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3250,6 +3250,92 @@ class AlterStageSegment(BaseSegment):
...
+                Ref("ShowInitialRowsExpressionSegment", optional=True)
+            ),
+            Sequence(
+                "EXTERNAL", 
+                "TABLE",
+                Ref("ObjectReferenceSegment"),
+                OneOf(
+                    Ref("FromAtExpressionSegment", optional=True),
+                    Ref("FromBeforeExpressionSegment", optional=True)
+                ),
```
tunetheweb, line 3327: Same fix as above:

```suggestion
                OneOf(
                    Ref("FromAtExpressionSegment"),
                    Ref("FromBeforeExpressionSegment"),
                    optional=True,
                ),
```

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3250,6 +3250,92 @@ class AlterStageSegment(BaseSegment):
...
+                Ref("InsertOnlyExpressionStatement", optional=True)
```
tunetheweb, line 3328:
```suggestion
                Ref("InsertOnlyExpressionSegment", optional=True)
```

tunetheweb (changes requested):
These changes seem to allow your test cases to parse

src/sqlfluff/dialects/dialect_snowflake.py
HeyZiko: I've opted to directly integrate them. At I was trying to resolve what turned out to be a misspelled reference (`InsertOnlyExpressionStatement` vs. `InsertOnlyExpressionSegment`), but then kept them out because of our dialog.

HeyZiko:
@tunetheweb I'm fixing the flake8 linting flaws, then I'll close this draft and generate a proper PR. 
Any additional thoughts or concerns before I do?
Would you prefer multiple PRs if I were to add an ALTER, or is it preferable wrangle all related grammars into one PR?

tunetheweb:
Can I take this out of draft now @HeyZiko ?

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3250,6 +3250,70 @@ class AlterStageSegment(BaseSegment):
...
+                Ref("ObjectReferenceSegment"),
+                OneOf(
+                    Ref("FromAtExpressionSegment"),
+                    Ref("FromBeforeExpressionSegment"),
+                    optional=True,
+                ),
+                Sequence(
+                    "APPEND_ONLY",
+                    Ref("EqualsSegment"),
+                    OneOf(Ref("TrueSegment"), Ref("FalseSegment")),
```
tunetheweb, line 3282: Nit, there's a segment for this

```suggestion
                   Ref("BooleanLiteralGrammar"),
```

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3250,6 +3250,70 @@ class AlterStageSegment(BaseSegment):
...
+                    optional=True,
+                ),
+                Sequence(
+                    "SHOW_INITIAL_ROWS",
+                    Ref("EqualsSegment"),
+                    OneOf(Ref("TrueSegment"), Ref("FalseSegment")),
```
tunetheweb, line 3288:
```suggestion
                   Ref("BooleanLiteralGrammar"),
```

tunetheweb (approved):
Two tiny nits

tunetheweb:
Great job @HeyZiko and now that first one's under your belt, hopefully you'll be keen to add more Snowflake functionality you need!

PR #2939, opened 2022-03-28
Snowflake: Add `ALTER STREAM` support

HeyZiko:

Added support for `ALTER STREAM` in the Snowflake dialect.

No.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3314,6 +3315,49 @@ class CreateStreamStatementSegment(BaseSegment):
...
+
+    https://docs.snowflake.com/en/sql-reference/sql/alter-stream.html
+    """
+
+    type = "alter_stream_statement"
+
+    match_grammar = Sequence(
+        "ALTER",
+        "STREAM",
+        Ref("IfExistsGrammar", optional=True),
```
tunetheweb, line 3330: I agree this is in the syntax but curious how you can alter a stream that doesn’t exist? Or will it just not do anything then?

Not a blocker for this PR as what you did is right according to the official snowflake docs but just curious for myself.

HeyZiko: IKR? I re-read it a few times to verify I was reading it right.
Here's an even stranger idiosyncrasy from the docs (haven't tried actually running it in any Snowflake environment): 
`UNSET TAG` doesn't even accept the optional `[ IF EXISTS ]` clause.

PR #2942, opened 2022-03-29
More dialect checking, fixes, inheritance cleanup

barrywhart:
* Add back some runtime checks for `replace()` that were deleted in prior PR
* Add a check that segment `type` matches base dialect when replacing a segment
* Add a check that if we're replacing a segment that defines both `match_grammar` and `parse_grammar`, the replacement must define **both** of them or **neither** of them
* Manual cleanup of dialects to use inheritance where appropriate

### Pull Request checklist

  - Other.


src/sqlfluff/core/dialects/base.py
```diff
@@ -152,7 +152,28 @@ def replace(self, **kwargs: DialectElementType):
         for n in kwargs:
             if n not in self._library:  # pragma: no cover
                 raise ValueError(f"{n!r} is not already registered in {self!r}")
-            self._library[n] = kwargs[n]
+
+            # To replace a segment, the replacement must either be a
```
barrywhart, line 156: This code was in the `segment()` decorator I removed in the previous PR. That code is still useful, so I moved it to the still-used `replace()` function.

src/sqlfluff/core/dialects/base.py
```diff
@@ -152,7 +152,33 @@ def replace(self, **kwargs: DialectElementType):
...
+            if (
+                isinstance(self._library[n], type)
+                and isinstance(cls, type)
+                and not issubclass(cls, self._library[n])
+            ):
+                if self._library[n].type != cls.type:
+                    raise ValueError(  # pragma: no cover
+                        f"Cannot replace {n!r} because 'type' property does not "
+                        f"match: {cls.type} != {self._library[n].type}"
+                    )
```
barrywhart, line 169: These lines are actually new. I had wondered if the `type` property was always consistent when replacing a parent segment, and I found several cases where it wasn't. This was presumably a mistake/accident. Several, but not all, of these mismatches were in the Exasol dialect.

Note that when using segment inheritance (now the recommended usual practice when replacing segments), the `type` property can be inherited and thus mismatches avoided.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -321,14 +321,9 @@ class SelectStatementSegment(ansi.SelectStatementSegment):
...
 
 
-class UnorderedSelectStatementSegment(BaseSegment):
+class UnorderedSelectStatementSegment(ansi.UnorderedSelectStatementSegment):
     """Enhance unordered `SELECT` statement to include QUALIFY."""
 
-    type = "select_statement"
-    match_grammar = ansi_dialect.get_segment(
-        "UnorderedSelectStatementSegment"
-    ).match_grammar.copy()
```
barrywhart, line 330: With inheritance, no need to copy if not modifying.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -536,13 +527,10 @@ class FunctionDefinitionGrammar(BaseSegment):
...
 
-class WildcardExpressionSegment(BaseSegment):
+class WildcardExpressionSegment(ansi.WildcardExpressionSegment):
     """An extension of the star expression for Bigquery."""
 
-    type = "wildcard_expression"
-    match_grammar = ansi_dialect.get_segment(
-        "WildcardExpressionSegment"
-    ).match_grammar.copy(
+    match_grammar = ansi.WildcardExpressionSegment.match_grammar.copy(
```
barrywhart, line 533: This PR eliminates dialects using `get_segment()` in favor of direct attribute access (cleaner).

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1038,13 +1031,10 @@ class SelectClauseSegment(BaseSegment):
         enforce_whitespace_preceding_terminator=True,
     )
 
-    parse_grammar = Ref("SelectClauseSegmentGrammar")
```
barrywhart, line 1041: This is now inherited from the base class.

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -2754,21 +2732,13 @@ class CreateIndexStatementSegment(BaseSegment):
...
-    type = "frame_clause"
-
-    _frame_extent = OneOf(
-        Sequence("CURRENT", "ROW"),
-        Sequence(
-            OneOf(Ref("NumericLiteralSegment"), "UNBOUNDED"),
-            OneOf("PRECEDING", "FOLLOWING"),
-        ),
-    )
+    _frame_extent = ansi.FrameClauseSegment._frame_extent
```
barrywhart, line 2741: I confirmed this is identical to ANSI, so I removed the duplication.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -1570,16 +1563,6 @@ class AlterWarehouseStatementSegment(BaseSegment):
     )
 
 
-class CommentClauseSegment(BaseSegment):
```
barrywhart, line 1573: This entire segment definition is identical to the base ANSI dialect, so I removed it.

src/sqlfluff/dialects/dialect_sparksql.py
```diff
@@ -981,7 +972,7 @@ class InsertStatementSegment(BaseSegment):
     https://spark.apache.org/docs/latest/sql-ref-syntax-dml-insert-overwrite-table.html
     """
 
-    type = "insert_table_statement"
+    type = "insert_statement"
```
barrywhart, line 975: Fixed a `type` mismatch with the ANSI segment type. Did not use inheritance because ANSI segment uses `parse_grammar` but this one doesn't.

test/fixtures/dialects/postgres/postgres_drop_trigger.yml
```diff
@@ -3,10 +3,10 @@
 # computed by SQLFluff when running the tests. Please run
 # `python test/generate_parse_fixture_yml.py`  to generate them after adding or
 # altering SQL files.
-_hash: 6670d8c469d2bc7b84b0e9ee89026706879b67416462305ae59412e51e70390f
+_hash: 7d795c868b9e279e23a114faf1e9e6fcdd63c15d70c543c0def0aeffa878953c
 file:
 - statement:
-    drop_trigger_statement:
+    drop_trigger:
```
tunetheweb, line 9: Why did this change?

test/fixtures/dialects/exasol/AlterTableColumn.yml
```diff
@@ -50,7 +50,7 @@ file:
                     start_bracket: (
                     literal: '10'
                     end_bracket: )
-              column_option:
+              column_constraint_segment:
```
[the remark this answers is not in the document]
barrywhart: Several dialects had segments that changed the base segment type name. I believe these are bugs, so I added a check for this and fixed all the places this happened.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3955,8 +3929,7 @@ class TransactionStatementSegment(BaseSegment):
     https://docs.snowflake.com/en/sql-reference/sql/rollback.html
     """
 
-    type = "transaction_statement"
-    match_grammar = OneOf(
+    match_grammar = OneOf(  # type: ignore
```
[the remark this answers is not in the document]
tunetheweb: Oh so this is because it moved form a `Sequence` to a `OneOf` is that it?

Hmm... that could be quite annoying then for people adding new grammar.

src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -1724,7 +1724,8 @@ class StatementSegment(postgres.StatementSegment):
 
     type = "statement"
 
-    parse_grammar = redshift_dialect.get_segment("StatementSegment").parse_grammar.copy(
+    match_grammar = postgres.StatementSegment.match_grammar
```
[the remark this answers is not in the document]
tunetheweb: On the one hand a bit of a waste. On the other hand quite like having them both together.

Think I like forcing to have both. Will resolve for now and reopen if I think of a change.

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -2657,48 +2657,39 @@ class RenameStatementSegment(BaseSegment):
...
 
-    type = "drop_table_statement"
-    match_grammar = ansi_dialect.get_segment(
-        "DropTableStatementSegment"
-    ).match_grammar.copy(
+    match_grammar = ansi.DropTableStatementSegment.match_grammar.copy(
         insert=[
             Ref("DelimiterSegment", optional=True),
         ],
     )
```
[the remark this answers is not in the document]
tunetheweb: OK let's leave as is.

test/fixtures/dialects/exasol/AlterTableColumn.yml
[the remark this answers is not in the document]
tunetheweb: OK see why now. Looks good.

src/sqlfluff/dialects/dialect_snowflake.py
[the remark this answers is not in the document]
barrywhart: I wonder if there's a solution. Mypy likes to guess types. It's annoying here because I think BaseSegment has the correct, general type, but mypy wants to enforce the more specific type it got from the ANSI definition. The type checker is great most of the time, but sometimes annoying.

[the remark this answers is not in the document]
barrywhart: According to this, the behavior is by design.

Note that you're not required to use inheritance, and in many cases, there's little or no advantage to doing so.

It's really useful for those segment types such as object names that add their own functions that are used various places, including the rules.

Any thoughts on this overall space? I started down this path because of a segment type that was missing a function defined on the ANSI "base", causing a runtime error in a rule. Since then, I've been exploring the possibilities on the core side (new/better error checks) and on the dialect files (shorter code, getting context-sensitive code help from editors that support type annotations such as PyCharm).

Happy to take or leave bits of this -- the goal is to simplify and reduce bugs, which can be a balancing act since the underlying language and tools have limits and quirks.

barrywhart:
@tunetheweb: Ok, I added type hints throughout the ANSI dialect, which allowed me to remove almost every use of `# type: ignore` in the other dialects. The only remaining instances are 4 cases like this, all involving cases where we override the base's `match_grammar.terminator`. I consider that code hacky, so I'm not bothered that we have to do this. At some point, we could look into providing a core function to do this, and probably eliminate the need for the `# type: ignore` at the same time.

```
match_grammar.terminator = match_grammar.terminator.copy(  # type: ignore
```

barrywhart:
@tunetheweb: I added the test we discussed. See `test/core/linter_test.py::test_require_match_parse_grammar`.

src/sqlfluff/core/dialects/base.py
```diff
@@ -152,7 +152,58 @@ def replace(self, **kwargs: DialectElementType):
...
+                # Overriding a subset of them is not necessarily wrong, but it's
+                # error-prone, hence this policy.
+                grammars = {"match_grammar", "parse_grammar"}
+                # TRICKY: The explicit use of __dict__ on the classes is
+                # deliberate. We are concerned with whether a class itself does
+                # or does not define a thing, IGNORING INHERITED VALUES.
+                if grammars.intersection(set(self._library[n].__dict__)) == grammars:
+                    overrides = grammars.intersection(set(cls.__dict__))
+                    if overrides and overrides != grammars:
+                        for grammar in grammars:  # pragma: no cover
```
tunetheweb, line 197: I presume this `no cover` can be removed now?

```suggestion
                        for grammar in grammars:
```

src/sqlfluff/core/dialects/base.py
```diff
@@ -152,7 +152,58 @@ def replace(self, **kwargs: DialectElementType):
...
+                            if (
+                                grammar in self._library[n].__dict__
+                                and grammar not in cls.__dict__
+                            ):
+                                raise ValueError(  # pragma: no cover
```
tunetheweb, line 202: And this one:

```suggestion
                                raise ValueError(
```

src/sqlfluff/dialects/dialect_postgres.py
tunetheweb: Did you answer this one @barrywhart ? Other than that good to go I think.

barrywhart: It's similar to a question you asked previously. Once the class has been created, `FrameClauseSegment._frame_extent` exists. But not _during_ the class definition. On line 2754, we need to use it 3 times, so it's shorter to assign it to a variable rather than repeating the full reference 3 times:
```
OneOf(ansi.FrameClauseSegment._frame_extent, Sequence("BETWEEN", ansi.FrameClauseSegment._frame_extent, "AND", ansi.FrameClauseSegment._frame_extent)),
```

I think I shared a Stack Overflow link earlier -- if you go back and read that, they explain how this works. It's definitely different than most other OO languages like C++ and Java. The behavior is arguably "simpler", but makes code inside a class more verbose sometimes.

tunetheweb: OK gotcha.

tunetheweb (approved):

PR #2949, opened 2022-03-30
TSQL: Support `OUTPUT` Params and `GOTO` Statements

### LINKED ISSUE #2948: TSQL: CREATE PROCEDURE does not support OUTPUT or GOTO
### Search before asking

### What Happened

While linting a stored procedure definition, the linter encountered a parser error. The error appears to be caused by the create statement. 

```
SET QUOTED_IDENTIFIER ON
GO
SET ANSI_NULLS ON
GO

CREATE PROCEDURE [dbo].[usp_process_tran_log]
	  @out_vchCode AS uddt_output_code OUTPUT
	, @out_vchMsg AS uddt_output_msg OUTPUT
	, @in_debug INT = 1
AS
--*******************************************************************************************
...
``` 

We encounter a parsing violation during linting. 
```
==== parsing violations ====
L:   4 | P:   3 |  PRS | Line 4, Position 3: Found unparsable section: '\n\nCREATE
                       | PROCEDURE [dbo].[usp_process_tr...'
```

### Expected Behaviour

Since `CREATE PROCEDURE` ... `AS` is valid `tsql`, I would expect to not encounter a parser error here.

### Observed Behaviour

As noted above, we get a parser violation. Checking the output of parse, yields this interesting output: 
```
...
[L:  4, P:  1]      |        keyword:                                              'GO'
[L:  4, P:  3]      |    unparsable:                                               !! Expected: 'Nothing...'
        Comments:
[L: 11, P:  1]      |            comment:                                          '--*******************************************************************************************'
        Code:
[L:  4, P:  3]      |            newline:                                          '\n'
[L:  5, P:  1]      |            newline:                                          '\n'
[L:  6, P:  1]      |            raw:                                              'CREATE'
...
```
Note the jump to Line 11 in the parser output followed by the return to line 4 before it continues. 

Removing the statements up to the `CREATE` results in no parser error.

## Basic Test 
Download `sample.txt` and rename to `sample.sql` 
Download `sqlfluff.txt` and rename to `.sqlfluff` or copy `.sqlfluff` config pasted below into a file called `.sqlfluff`. 
To view the raw parse output:
```
sqlfluff.exe parse ./sample.sql --dialect tsql --config ./.sqlfluff
``` 
 This should be the same as the `parse.txt` I've provided. 

Additionally, to view the parse exception in the lint output: 
```
sqlfluff.exe lint ./sample.sql --dialect tsql --config ./.sqlfluff
``` 

## Additional Test 
The removal of lines 6-10 in `sample.txt` will result in no parser error. Similarly, removal of lines 1-10 of `sample.txt` results in no parser error. This issue seems to be caused by the presence of the `tsql` definition keyword `CREATE PROCEDURE ... AS`

## Files
Example of `sqlfluff parse` output 
[parse.txt](https://github.com/sqlfluff/sqlfluff/files/8376745/parse.txt)
`sample.txt` will need to be renamed to `sample.sql` for the linter to work. Github rejects this file upload without renaming. 
[sample.txt](https://github.com/sqlfluff/sqlfluff/files/8376748/sample.txt)
`sqlfluff.txt` is the `.txt` version of my config file. Rename this to `.sqlfluff`. Github rejects this file upload without renaming. 
[sqlfluff.txt](https://github.com/sqlfluff/sqlfluff/files/8376750/sqlfluff.txt)

### Dialect

`tsql`

### Version

```
$ sqlfluff --version
sqlfluff, version 0.11.2
```

### Configuration

[sqlfluff]
dialect = tsql

[sqlfluff:indentation]
indented_joins = false
indented_using_on = true

[sqlfluff:rules]
indent_unit = tab
tab_space_size = 4
operator_new_lines = after
capitalisation_policy = upper
aliasing = explicit
allow_scalar = True
ignore_comment_clauses = True
ignore_comment_lines = True
max_line_length = 120
comma_style = leading
prefer_count_1 = True
multiline_newline = True
exclude_rules = L025, L031, L061

### Code of Conduct

tunetheweb:

Fixes #2948

Nope

### Pull Request checklist

  - Other.


tunetheweb:
FYI @jpers36

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -1626,9 +1627,11 @@ class ProcedureParameterListGrammar(BaseSegment):
     match_grammar = OptionallyBracketed(
         Sequence(
             Ref("FunctionParameterGrammar"),
+            OneOf("OUT", "OUTPUT", "READONLY", optional=True),
```
edmoren, line 1630: Noticed that we already had `OUTPUT` in `src/sqlfluff/dialects/dialect_tsql_keywords.py`. And that you only had to add `OUT` and `READONLY`

Is the reason the `OUTPUT` wasn't parsed due to that keyword was not checked for during procedure creation statements?

Right now I am guessing that `FunctionParameterGrammar` is invoked only a certain sections of already parsed `tsql` code. And deciding when to invoke it is the responsibility of a different piece of the architecture.

tunetheweb: I presume the keywords list was taken from here: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/reserved-keywords-transact-sql?view=sql-server-ver15. We generally want the RESERVED keywords listed (so we know they cannot be used as object names) but might not have all the other keywords that can be used if they are not reserved. `OUTPUT` is in the list of ODBC keywords, but `OUT` and `READONLY` weren't so, since we wanted to use them, I had to add them.

The grammar is what's more important though, but you can't use a keyword in the Grammar unless it's in our Keyword list (whether Reserved or Unreserved), hence why we had to add those two to keywords and then add all three to the Grammar.

So just cause we had `OUTPUT` listed in the Keywords, didn't mean that SQLFluff knew where that should be used. Hence why adding that to the grammar here allowed it to be parsed.

Basically what the parser does is try to match the provided SQL to one of the Grammars defined for that dialect. If it can match it, then it parses - if it can't then it tries another Grammar, and if it reaches the end then it fails. So "Right now I am guessing that FunctionParameterGrammar is invoked only a certain sections of already parsed tsql code" isn't exactly right - the grammar is basically what the parser is!

The linter then defines rules on those Grammars (e.g. we want Function Names to be in caps, or we want certain spacing between certain parts of grammar).

As you can see the grammar is fairly easy to understand if you understand the SQL dialect, even if you don't know python (I'm not really a Python developer myself). And the test cases added in this PR are just a good example of the SQL that should be parsed so we 1) know it works and 2) won't break that from working in future if we add more changes. The YAML equivalent of the SQL is auto generated so you don't need to worry about that.

So we definitely love it when people submit PRs to add simple grammar additions like this as you'll likely know better than us exactly how the grammar should be defined and can even test it on your TSQL instance, and hopefully this PR shows how the lack of SQLFluff code knowledge hopefully won't be a blocker. I don't use TSQL, but was able to figure out what needed to be added here. So if you spot anything else that's not working for your SQL then definitely do feel free to submit your own PRs and we're definitely happy to help with that. I've also a wiki here on how to do that: https://github.com/sqlfluff/sqlfluff/wiki/Contributing-Dialect-Changes

barrywhart (approved):

src/sqlfluff/dialects/dialect_tsql.py
edmoren: > Basically what the parser does is try to match the provided SQL to one of the Grammars defined for that dialect. If it can match it, then it parses - if it can't then it tries another Grammar, and if it reaches the end then it fails.

Gotcha, so it's actually simpler than I thought. Okay cool I was thinking through that as well. Like in this PR sometimes `READONLY` isn't optional on function parameters. But that is a very narrow case when the parameter being passed in is specifically a user defined table. But you couldn't know that without querying the sql-server instance in advance of running the linter. *Or* assume it based on the user defined variable name (which I would not recommend). 

I will read over the document for contributing dialect changes. I will likely have some more for `tsql`, specifically on sql-server instances. 

Thanks @tunetheweb 😄 👍

Issue #2944, opened 2022-03-29
Rule L045 doesn't recognise CTE usage in a subquery when rule L042 is enabled


Arshaku:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Create a test.sql file with the following content:
```
CREATE TABLE IF NOT EXISTS a_table AS
WITH
cte_1 AS (
    SELECT
        a,
        b
    FROM table_1
)

SELECT * FROM (
    SELECT table_2.* FROM table_2
    LEFT JOIN cte_1
        ON cte_1.c = table_2.c
)
```

run "sqlfluff lint ./test.sql" on it and you will get the following error:
```
L:   3 | P:   1 | L045 | Query defines CTE "cte_1" but does not use it.
L:  10 | P:  15 | L042 | from_expression_element clauses should not contain subqueries. Use CTEs instead
```
even though actually 'cte_1' is used in the query.

if you exclude the rule L042 either by adding it to the "exclude_rules" in the .sqlfluff file or by adding "-- noqa" on line 10, 
"sqlfluff lint" it doesn't compline about violation of rule L045 on line 3 anymore.

### Expected Behaviour

As you can see "cte_1" is actually used in the query, so rule L045 should recognise it no matter whether Rule L042 is enabled or not.

### Observed Behaviour

L045 doesn't recognise a CTE used in a subquery when L042 is enabled

### How to reproduce

as described above

### Dialect

tested on 'snowflake' dialect

### Version

sqlfluff, version 0.11.1
Python 3.8.12

### Configuration

[sqlfluff]
verbose = 1
dialect = snowflake
templater = jinja
exclude_rules = L027,L031,L032,L036,L044,L046,L034,L050
output_line_length = 121
sql_file_exts=.sql

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 250
indent_unit = space
comma_style = trailing
allow_scalar = True
single_table_references = consistent
unquoted_identifiers_policy = aliases

[sqlfluff:rules:L042]
forbid_subquery_in = both

[sqlfluff:rules:L010]  # Keywords
capitalisation_policy = upper

[sqlfluff:rules:L014]
extended_capitalisation_policy = upper

[sqlfluff:rules:L030]  # function names
extended_capitalisation_policy = upper

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

[...]

PR #2955, opened 2022-03-30
Bug: L042 modifies parse tree even during "lint"

barrywhart:
Makes progress on #2944.

It's a workaround, not a fix. The workaround is to let rules know if we're in "lint" or "fix" mode, so they can skip the "fix" logic when just linting. This is not a proper fix because L042 has a design flaw: It's modifying the parse tree directly, bypassing the use of `LintFix`. This will likely cause other issues in the future until it's fixed.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -513,6 +513,7 @@ class Linter:
                     ignore_mask=ignore_buff,
                     dialect=config.get("dialect_obj"),
                     fname=fname,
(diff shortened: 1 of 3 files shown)


barrywhart:
@OTooleMichael: Please see this PR and the discussion on #2944. We need to find a way for L042 to make its fixes using `LintFix`. Let's discuss it on the issue rather than here.

tunetheweb (approved):

PR #2959, opened 2022-03-31
TSQL: Redefine `DatatypeIdentifierSegment`

### LINKED ISSUE #2958: L029 flags datatype identifiers incorrectly in tsql
### Search before asking

### What Happened

This query:

```sql
select cast(a as date) as b from c
```

...presents this error (when running in `tsql` but not `ansi` or `snowflake`):
```
L:   1 | P:  18 | L029 | Keywords should not be used as identifiers.
```

...when the `unquoted_identifiers_policy` is set to `all`.

This is because in the parse tree `cast(a as date)` is interpreted as:

```
[L:  1, P:  8]      |                        function:
[L:  1, P:  8]      |                            function_name:
[L:  1, P:  8]      |                                keyword:                      'cast'
[L:  1, P: 12]      |                            bracketed:
[L:  1, P: 12]      |                                start_bracket:                '('
[L:  1, P: 13]      |                                [META] indent:
[L:  1, P: 13]      |                                expression:
[L:  1, P: 13]      |                                    column_reference:
[L:  1, P: 13]      |                                        identifier:           'a'
[L:  1, P: 14]      |                                whitespace:                   ' '
[L:  1, P: 15]      |                                keyword:                      'as'
[L:  1, P: 17]      |                                whitespace:                   ' '
[L:  1, P: 18]      |                                data_type:
[L:  1, P: 18]      |                                    identifier:               'date'
[L:  1, P: 22]      |                                [META] dedent:
[L:  1, P: 22]      |                                end_bracket:                  ')'
```

Where `date` is correctly identified as a `data_type`, but within that still contains an `identifier`. In `ansi` and `snowflake` (and I suspect many others), the parse tree is uses `data_type_identifier` rather than just `identifier` and so does not trigger L029.

### Expected Behaviour

The behaviour should be consistent between all dialects.

### Observed Behaviour

`tsql` dialect fails where others pass.

### How to reproduce

Query is above. Config file below. Run `lint`.

### Dialect

TSQL

### Version

sqlfluff 0.11.1

### Configuration

```
[sqlfluff]
dialect = tsql

[sqlfluff:rules:L029]
```

### Code of Conduct

alanmcruickshank:
Resolves #2958 .

For TSQL, the `DatatypeIdentifierSegment` was being redefined inconsistently to the other dialects. This removes that redefinition allowing it to fall back onto the base definition.


src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -258,7 +258,17 @@
...
     ),
-    DatatypeIdentifierSegment=Ref("SingleIdentifierGrammar"),
+    DatatypeIdentifierSegment=SegmentGenerator(
+        # Generate the anti template reserved keywords
+        lambda dialect: RegexParser(
+            r"[A-Z][A-Z0-9_]*|\[[A-Z][A-Z0-9_]*\]",
+            CodeSegment,
+            name="data_type_identifier",
+            type="data_type_identifier",
+            anti_template=r"^(" + r"|".join(dialect.sets("reserved_keywords")) + r")$",
```
tunetheweb, line 268: ANSI only excludes `NOT` rather than all keywords. That sounds familiar but can’t remember why. Should we do the same here for consistency?

tunetheweb (approved):
One question for you but in general LGTM!

src/sqlfluff/dialects/dialect_tsql.py
alanmcruickshank: Good question. I'll update the ansi one and see if it's better.

tunetheweb: I’m wondering if we should do it the other way around? Some data types might be keywords in some dialects. But tests should show that if it’s the case.

tunetheweb: Looks like it is causing issues. Is there any reason not to just allow keywords here @alanmcruickshank ? Worked fine for ANSI up until now.

alanmcruickshank: let's try that too :)

Issue #2961, opened 2022-04-01
Add `IS [NOT] DISTINCT FROM` to BigQuery dialect


dmohns:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

BigQuery supports the usage of NULL-aware comparison operators `IS [NOT] DISTINCT FROM`, see https://cloud.google.com/bigquery/docs/reference/standard-sql/operators#is_distinct

Currently, using `IS [NOT] DISTINCT FROM` fails `sqlfluff` with 

```
Found unparsable section: ' IS DISTINCT'
Found unparsable section: ' IS DISTINCT FROM...`
```

### Expected Behaviour

`sqlfluff` correctly parses queries using `IS [NOT] DISTINCT FROM` operators.

### Observed Behaviour

See the following examples

```sql
-- 1. Column distinctness in SELECT expression
SELECT a_column IS DISTINCT FROM b_column FROM t_table;
SELECT b_column IS NOT DISTINCT FROM c_column FROM t_table;

-- 2. Column distinctness in WHERE expression
SELECT a_column FROM t_table WHERE a_column IS DISTINCT FROM b_column;
SELECT a_column FROM t_table WHERE a_column IS NOT DISTINCT FROM b_column;

-- 3. Column distinctness in JOIN expression
SELECT t_table_1.a_column
FROM t_table_1
INNER JOIN t_table_2
    ON t_table_1.a_column IS DISTINCT FROM t_table_2.a_column;
SELECT t_table_1.a_column
FROM t_table_1
INNER JOIN t_table_2
    ON t_table_1.a_column IS NOT DISTINCT FROM t_table_2.a_column;

-- 4. Column distinctness in MERGE expression
MERGE INTO t_table_1
USING t_table_2
ON t_table_1.a_column IS DISTINCT FROM t_table_2.a_column
WHEN NOT MATCHED THEN INSERT (a) VALUES (b);
MERGE INTO t_table_1
USING t_table_2
ON t_table_1.a_column IS NOT DISTINCT FROM t_table_2.a_column
WHEN NOT MATCHED THEN INSERT (a) VALUES (b);
```

Running 
```sh
sqlfluff parse --dialect bigquery 1.sql
```

Yields

```
==== parsing violations ====
L:   2 | P:  16 |  PRS | Line 2, Position 16: Found unparsable section: ' IS DISTINCT'
[...]

PR #2962, opened 2022-04-01
Add basic `IS (NOT) DISTINCT FROM` support in most dialects

tunetheweb:

Makes progress on #2961 
Note we currently do not fully support this in the `SELECT` clause unless brackets are used.

### Pull Request checklist
- [z] Please confirm you have completed any of the necessary steps below.

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -387,6 +387,8 @@ ansi_dialect.add(
         Ref("LessThanOrEqualToSegment"),
         Ref("NotEqualToSegment"),
         Ref("LikeOperatorSegment"),
+        Sequence("IS", "DISTINCT", "FROM"),
+        Sequence("IS", "NOT", "DISTINCT", "FROM"),
     ),
     # hookpoint for other dialects
     # e.g. EXASOL str to date cast with DATE '2021-01-01'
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -387,6 +387,8 @@
         Ref("LessThanOrEqualToSegment"),
         Ref("NotEqualToSegment"),
         Ref("LikeOperatorSegment"),
+        Sequence("IS", "DISTINCT", "FROM"),
+        Sequence("IS", "NOT", "DISTINCT", "FROM"),
```
dmohns, line 391: Would it make sense to turn this into a Grammar or Segment of its own?

dmohns (commented):
Thanks for doing this so quickly 😉 

I added one question, which is probably more helping myself familiarising with sqlfluff code than anything 😂

src/sqlfluff/dialects/dialect_ansi.py
tunetheweb: Possibly. We tend to do that for a few reasons:
1) To avoid repeating lots of same code in lots of places (not the case here - though admittedly same code also in Postgres dialect)
2) To allow it to overridden in other dialects (not the case here)

As neither is really the case here, I'd leave it as is. Having another Grammar/Segment adds a level of redirection and don't think that's necessary here so clearer to have the actual grammar.

BTW if interested in contributing, then this is a good read: https://github.com/sqlfluff/sqlfluff/wiki/Contributing-Dialect-Changes

tunetheweb: BTW we could also consolidate this down to one statement:

```py
Sequence("IS", Ref.keyword("NOT", optional=True), "DISTINCT", "FROM"),
```

But again, think it's slightly clearer as two statements.

Always lots of ways of doing things in SQLFluff!

barrywhart (approved):

PR #2967, opened 2022-04-01
TSQL grammar deduplication

tunetheweb:

Adds a couple of Segments to DRY code.

Not sure if these should be Segments or Grammars?

Nope

### Pull Request checklist

  - Other.


test/fixtures/dialects/tsql/triggers.yml
```diff
@@ -33,7 +33,7 @@ file:
             - comma: ','
             - literal: '10'
             - end_bracket: )
-    statement_terminator: ;
+      - statement_terminator: ;
 - go_statement:
     keyword: GO
 - batch:
```
alanmcruickshank, line 39: This is inconsistent behaviour with the `ansi` dialect. I would expect the terminator to be outside the `statement` segment.

alanmcruickshank (changes requested):
I don't know why - but these changes are causing an odd shift in the classification of statement terminators. Any ideas why that's happening?

test/fixtures/dialects/tsql/triggers.yml
tunetheweb: I did look into these (should have commented).

It's due to this slight change included in here:

image

The original definition didn't allow terminators, but [looking at the grammar](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-trigger-transact-sql?view=sql-server-ver15) this was incorrect and it should have.

And because TSQL is odd and the terminator is optional (in both the substatement and the main statement) that leads to these sorts of this. 

It could be parsed as this:

```sql
CREATE TRIGGER example
AS
    SELECT * FROM table;
```

Or this:

```sql
CREATE TRIGGER example
AS
    SELECT * FROM table
;
```

Both are equally valid.

The other statements have this same issue (and have always had), but you see the change here because I've corrected the grammar.

So I don't think it's an issue. That optional semi-colon in tsql is just weird!

alanmcruickshank: Interesting 🤔 . I'm going to check out this branch and have a play with it.

alanmcruickshank: Ok I see why this a little more complicated. There are two nested statements, so it's easy to see why the terminator could be seen to terminate the inner one or the outer one. My reading of the query however is that the terminator, terminated the `CREATE` query, not just the `RaiseError` statement. Given that, I think the original placement is where it _should_ be.

tunetheweb: Don't disagree, but as @WittierDinosaur said, Delimted is currently greedy (is this fixed in your refactoring?) so currently it's put in the other place. However as it's optional in both it doesn't really matter one way of the other. And as I say, other TSQL statements already have the same issue.

So don't think this needs to block this PR. I can revert that change if you're concerned but technically the previous grammar was incorrect, even if it parsed these examples "better".

alanmcruickshank: Ok - I think I've got a solution which I think is a better representation of the underlying structure. I'm going to push to the same branch in the interests of ease, but @tunetheweb feel free to keep editing obv.

barrywhart:
About the placement of terminators, these kinds of things are more important to get consistent across dialects now that the linter re-parses affected areas of the parse tree after applying fixes. The more consistent we are across dialects, the simpler the rule "fix" logic can be.

barrywhart:
@tunetheweb: A grammar does not create its own segment in the parse tree, but a segment does. Segments are valuable for helping group and label things, e.g. expression, `FROM` clause, etc.

tunetheweb:
Thanks. Will add that to my wiki entry when next on computer. Or feel free to do it too if you want.

alanmcruickshank (approved):

alanmcruickshank:
> About the placement of terminators, these kinds of things are more important to get consistent across dialects now that the linter re-parses affected areas of the parse tree after applying fixes. The more consistent we are across dialects, the simpler the rule "fix" logic can be.

@barrywhart - agreed. Hence #2974

tunetheweb:
> @tunetheweb: A grammar does not create its own segment in the parse tree, but a segment does. Segments are valuable for helping group and label things, e.g. expression, `FROM` clause, etc.

> Thanks. Will add that to my wiki entry when next on computer. Or feel free to do it too if you want.

Added this: https://github.com/sqlfluff/sqlfluff/wiki/Contributing-Dialect-Changes#segments-and-grammars

tunetheweb:
> > About the placement of terminators, these kinds of things are more important to get consistent across dialects now that the linter re-parses affected areas of the parse tree after applying fixes. The more consistent we are across dialects, the simpler the rule "fix" logic can be.
> 
> @barrywhart - agreed. Hence #2974

To a point. But when it's optional we have to allow for the fact it won't be there anyway.

Still, I agree Alan's improvement is good here.

barrywhart (approved):

PR #2972, opened 2022-04-01
SparkSQL: Support for querying snapshots when reading data with Delta Lake

R7L208:

This added new segments and grammar, as well as updating `TableExpressionSegment` to support querying historical snapshots of Delta Tables

- I renamed an instance of `Spark3` to `SparkSQL` in the README that was missed.
- update `atsign_literal` to `at_sign_literal` in BigQuery, Exasol, and MySQL

### Pull Request checklist

  - Other.


R7L208:
I noticed that several other dialects implement a variation of `at_sign_literal` and a lexer for `@`. Is it worthwhile to move this into ANSI and then update dialects that inherit from it for specific syntax?

tunetheweb:
> Reference in new issue

Possibly, though it also means something else in other dialects (e.g. Oracle) so might not work. Willing to see.

However, you have a different definition here so could have a common one anyway:
- You allow starting with numbers, bigquery and exasol don't. Does sparksql allow numbers? Does it allow underscores like the others do? Or would you propose overriding it here to a slightly different definition? In which case not totally convinced it's worth it.
- You've called it `at_sign_literal`, while bigquery and exasol call it `atsign_literal`. I actually prefer yours but can we change the others for consistency? Or change yours if you prefer?

R7L208:
> You allow starting with numbers, bigquery and exasol don't. MySQL allows anything. Does sparksql allow starting with numbers? Or would you propose overriding it here to a slightly different definition? In which case not totally convinced it's worth it.

The only two uses for an `@` and inherently the `at_sign_literal` in SparkSQL is to specify a historical snapshot of a Delta Table which would be one of `@v###` to specify by version number or `@yyyymmddHHSSss` where the timestamp is referenced numerically (ie `19911002123000`) to reference a version by timestamp. I kept it broad since it would correctly parse those two segments and still be syntactically valid if used in other cases within Spark (but pretty sure it's otherwise not used).

I'm also not convinced it's worth it. Sometimes it seems much more cut and dry but pieces of it could be moved to ANSI, like having a lexer for `@`, that would allow for simplifying segments in other dialects. Definitely provides questionable improvements which would be marginal at best.

> You've called it at_sign_literal, while bigquer, exasol and mysql call it atsign_literal. I actually prefer yours but can we change the others for consistency? Or change yours if you prefer?

I also prefer at_sign_literal and will make the update in other dialects. 🙂

tunetheweb:
> like having a lexer for `@`, that would allow for simplifying segments in other dialects

I don't think we can do that even as we lex the whole literal and not just the at sign.

R7L208:
> > like having a lexer for `@`, that would allow for simplifying segments in other dialects
> 
> I don't think we can do that even as we lex the whole literal and not just the at sign.

That is an excellent point. Seems like there's no compelling argument to move this into ANSI unless multiple at_sign_literals are defined, ie at_sign_numeric_literal and at_sign_alpha_literal, but I think that is unnecessary given current usage

tunetheweb:
Agreed. Though good to look for things to have the discussion!

src/sqlfluff/dialects/dialect_sparksql.py
```diff
@@ -386,6 +393,21 @@
         # there since there are no significant syntax changes
         "JDBC",
     ),
+    AsOfGrammar=Sequence("AS", "OF"),
```
tunetheweb, line 396: Not sure this is really necessary. Only use it twice and the grammar shortcut is longer than than the actual grammar! That can make sense if want to override it in other dialects but not so sure here.

tunetheweb (changes requested):
In general looks good. Think we should change the AsOfGrammar as noted. What do you think?

R7L208:
> In general looks good. Think we should change the AsOfGrammar as noted. What do you think?

Agreed! I can make those changes early tomorrow

tunetheweb (approved):

PR #2975, opened 2022-04-01
Fix bug in L063 for BigQuery `STRUCT` params

tunetheweb:

Adds params to the exclude list

Nope

### Pull Request checklist

  - Other.


src/sqlfluff/rules/L063.py
```diff
@@ -62,6 +62,7 @@ class Rule_L063(Rule_L010):
     _exclude_elements: List[Tuple[str, str]] = [
         ("type", "identifier"),
         ("type", "literal"),
+        ("type", "parameter"),
     ]
```
tunetheweb, line 66: I'm a little worried that there will be more edge cases like this, but unfortunately there's no easy way to distinguish all the data type keywords as not all are marked as datatype and the parent selector is picking up more than it should :-(

barrywhart: Think it should be (could be) cleaned up?

barrywhart:
Is there an issue number for this PR?

test/fixtures/rules/std_rule_cases/L063.yml
```diff
@@ -134,3 +133,16 @@ test_pass_sparksql_complex_data_types:
     rules:
       L063:
         extended_capitalisation_policy: upper
+
+test_pass_bigquery_struct_params:
```
barrywhart, line 137: Think it'd be worth adding a similar failing test case?

tunetheweb:
Nah. Discovered it myself when linting my own code base with latest `main` as L063 was flagging for some of my queries when it shouldn’t have.

barrywhart (approved):

src/sqlfluff/rules/L063.py
tunetheweb: I had hoped after @alanmcruickshank fixed up TSQL data types recently, but still a lot of other ones that are just data-type->keywords or similar (e.g. the `WITH TIME ZONE` ones). So not holding out massive hope to be honest.

We could perhaps restrict the types that `parenttype` looks at, rather than allowing everything. Tried that on this branch but ended up chasing my own tail. Let me have a play with that some more before merging this.

test/fixtures/rules/std_rule_cases/L063.yml
tunetheweb: Will do.

tunetheweb: Actually, won’t do.

The point of this test case was to catch the false positive whereby the param name was being incorrectly checked by this rule. So there is no fail case as this rule shouldn’t be looking at params. This test case ensures that.

The things it SHOULD be looking at are covered by other test cases already.

src/sqlfluff/rules/L063.py
tunetheweb: Actually I found a better way. We only need `parent_type` for keywords. So by restricting to that, can avoid having to maintain an exclude list.

PR #2980, opened 2022-04-02
Rule L045 doesn't recognise CTE usage in a subquery when rule L042 is enabled

### LINKED ISSUE #2944: Rule L045 doesn't recognise CTE usage in a subquery when rule L042 is enabled

barrywhart:
Fixes #2944

### Pull Request checklist

  - Other.


src/sqlfluff/core/linter/linter.py
```diff
@@ -491,6 +491,7 @@ def lint_fix_parsed(
 
         save_tree = tree
         for loop in range(loop_limit):
+            linter_logger.info(f"Linter loop {loop+1}/{loop_limit}")
```
barrywhart, line 494: This was really handy for debugging this issue (which at times caused the linter to hit the loop limit). Seems like a good thing to keep around.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -322,7 +322,7 @@ def is_whitespace(self) -> bool:
         """Return True if this segment is entirely whitespace."""
         return all(seg.is_whitespace for seg in self.segments)
 
-    @cached_property
+    @property
```
barrywhart, line 325: Just for debugging -- should undo this before merge.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -885,25 +885,24 @@ def select_children(
...
-        """Recursively crawl for segments of a given type.
-
-        Args:
-            seg_type: :obj:`str`: one or more type of segment
-                to look for.
-            recurse_into: :obj:`bool`: When an element of type "seg_type" is
-                found, whether to recurse into it.
-        """
+    def recursive_crawl(
+        self, *seg_type: str, recurse_into: bool = True, all_descendants: bool = False
```
barrywhart, line 889: Added new parameter `all_descendants` to get _all_ the segments, rather than filtering by type.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -885,25 +885,24 @@ def select_children(
...
+    ):
+        """Recursively crawl for segments of a given type (or all descendants)."""
         # Check this segment
-        if self.is_type(*seg_type):
+        if all_descendants or self.is_type(*seg_type):
             match = True
             yield self
```
tunetheweb, line 895: This feels weird.

So If I say `recursive_crawl(seg_type="keywords", recurse_into=true, all_descendants=true)` it will return all descendents, even those that are not `keywords`?

Feels to me like this is a completely different (though admittedly similar!) function and should either provide a type, or `all_descendants` but not both.

Would a type of value `ALL` or similar negate the need for the new (confusing and sometimes contradictory) param?

tunetheweb: Or make the type optional, and if it is not given, then that implies you want all descendants.

barrywhart: Yes, it's weird. I was wondering about this myself, and I did this more out of laziness than feeling sure it was good design. I think providing a separate function will be safer/more consistent than treating "empty" as all or defining a special "ALL" value.

barrywhart: Ok, I created a new, separate `recursive_crawl_all()` function.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -885,15 +885,14 @@ def select_children(
...
-                found, whether to recurse into it.
-        """
+    def recursive_crawl_all(self):
+        """Recursively crawl all descendant segments."""
+        yield self
+        for seg in self.segments:
+            yield from seg.recursive_crawl_all()
+
+    def recursive_crawl(self, *seg_type: str, recurse_into: bool = True):
+        """Recursively crawl for segments of a given type (or all descendants)."""
```
tunetheweb, line 895: Restore comments:

```suggestion
        """Recursively crawl for segments of a given type.
        Args:
            seg_type: :obj:`str`: one or more type of segment
                to look for.
            recurse_into: :obj:`bool`: When an element of type "seg_type" is
                found, whether to recurse into it.
        """
```

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -903,7 +902,10 @@ def recursive_crawl(self, *seg_type, recurse_into=True):
         if recurse_into or not match:
             # Recurse
             for seg in self.segments:
-                yield from seg.recursive_crawl(*seg_type, recurse_into=recurse_into)
+                yield from seg.recursive_crawl(
+                    *seg_type,
+                    recurse_into=recurse_into,
+                )
```
tunetheweb, line 909: Do we need this change?

src/sqlfluff/core/rules/functional/segments.py
```diff
@@ -88,11 +88,19 @@ def raw_segments(self) -> "Segments":
             raw_segments_list.extend(s.raw_segments)
         return Segments(*raw_segments_list, templated_file=self.templated_file)
 
+    def recursive_crawl_all(self) -> "Segments":  # pragma: no cover
```
tunetheweb, line 91: Do none of the L042 test cases cover this so do we need the pragma no cover?

src/sqlfluff/core/rules/functional/segments.py
```diff
@@ -88,11 +88,19 @@ def raw_segments(self) -> "Segments":
...
+            for i in s.recursive_crawl_all():
+                segments.append(i)
+        return Segments(*segments, templated_file=self.templated_file)
+
     def recursive_crawl(self, *seg_type: str, recurse_into: bool = True) -> "Segments":
         """Recursively crawl for segments of a given type."""
         segments: List[BaseSegment] = []
         for s in self:
-            for i in s.recursive_crawl(*seg_type, recurse_into):
+            for i in s.recursive_crawl(*seg_type, recurse_into=recurse_into):
```
tunetheweb, line 103: Do we need this change?

tunetheweb (changes requested):

src/sqlfluff/core/parser/segments/base.py
barrywhart: Nope. Seems like Black is not behaving like I'd expect -- remove the original change, it should go back to one line. 🤔 I'll restore it manually.

src/sqlfluff/core/rules/functional/segments.py
barrywhart: They don't use it. I wouldn't have had to add it, but it seemed good to keep `Segments` API somewhat aligned with the underlying `BaseSegment`.

src/sqlfluff/core/rules/functional/segments.py
barrywhart: Nope. Will remove.

barrywhart: Hmm, mypy complains if I remove it. It thinks we're trying to pass `recurse_into` (a boolean value) as one of the type values. I think mypy is wrong, but I'm okay with this change -- passing it explicitly as a keyword argument _does_ make it a bit clearer, even if the old way works.

barrywhart: I think I know why `mypy` started complaining about this. In this PR, I added type annotations to the `recursive_crawl()` function. Previously, `mypy` didn't know that the `recurse_into` parameter of `BaseSegment.recursive_crawl()` was a boolean. Now that it has that information, it started to complain.

tunetheweb: Fine keeping it in.

barrywhart:
@tunetheweb: Ok, I think this is clean now. I'm heading out for the evening. If this looks good, feel free to go ahead and merge.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -885,15 +885,14 @@ def select_children(
...
-                found, whether to recurse into it.
-        """
+    def recursive_crawl_all(self):
+        """Recursively crawl all descendant segments."""
+        yield self
+        for seg in self.segments:
+            yield from seg.recursive_crawl_all()
+
+    def recursive_crawl(self, *seg_type: str, recurse_into: bool = True):
+        """Recursively crawl for segments of a given type."""
```
tunetheweb, line 895: Did you miss this one @barrywhart ?

```suggestion
        """Recursively crawl for segments of a given type.
        Args:
            seg_type: :obj:`str`: one or more type of segment
                to look for.
            recurse_into: :obj:`bool`: When an element of type "seg_type" is
                found, whether to recurse into it.
        """
```

tunetheweb (approved):

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -885,9 +885,14 @@ def select_children(
...
-    def recursive_crawl(self, *seg_type, recurse_into=True):
-        """Recursively crawl for segments of a given type.
+    def recursive_crawl_all(self):
+        """Recursively crawl all descendant segments."""
+        yield self
+        for seg in self.segments:
+            yield from seg.recursive_crawl_all()
 
+    def recursive_crawl(self, *seg_type: str, recurse_into: bool = True):
+        """Recursively crawl for segments of a given type.
```
tunetheweb, line 895:
```suggestion
        """Recursively crawl for segments of a given type.

```

PR #2981, opened 2022-04-03
L041: Fix bug when there is a newline after the select clause modifier

jmc-bbk:
### Issue

When there is a newline after the `SELECT CLAUSE MODIFIER`, rule L041 will require introduce unnecessary whitespace. This requires a second run of `sqlfluff fix <>` to resolve.

Example:

``` 
-- test.sql
CREATE OR REPLACE TABLE myschema.mytable AS (
  SELECT
  DISTINCT
    cola
    , colb
  FROM myschema.mytable
);
```

Using the CLI:

```
== [test.sql] FAIL                                                                                                                                                                                              
L:   2 | P:   3 | L041 | 'SELECT' modifiers (e.g. 'DISTINCT') must be on the same                                                                                                                               
                       | line as 'SELECT'.
==== fixing violations ====
Are you sure you wish to attempt to fix these? [Y/n] Y
```
Output:

```
CREATE OR REPLACE TABLE myschema.mytable AS (
  SELECT DISTINCT  
    cola
    , colb
  FROM myschema.mytable
);
```

Using the CLI again:

```
== [test.sql] FAIL                                                                                                                                                                                              
L:   2 | P:  18 | L001 | Unnecessary trailing whitespace.                                                                                                                                                       
==== fixing violations ====
Are you sure you wish to attempt to fix these? [Y/n] 
```

### Brief summary of the changes made

- Added a variable that has any leading whitespace between the `SELECT` and the `SELECT CLAUSE MODIFIER`.
- Added a conditional so that if there is a `NEWLINE` after the `SELECT CLAUSE MODIFIER` we delete this leading whitespace in addition to (the already deleted) newlines.

- None that I am aware of (first contribution)!

### Pull Request checklist

  - Other.


src/sqlfluff/rules/L041.py
```diff
@@ -93,9 +101,24 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+            fixes.extend((LintFix.delete(s) for s in leading_newline_segments))
+        # If there is a newline after the select clause modifier then delete both the
+        # newlines and whitespace between the select keyword and select clause modifier.
+        else:
+            fixes.extend(
+                (
+                    LintFix.delete(s)
+                    for s in leading_newline_segments + leading_whitespace_segments
+                )
+            )
```
tunetheweb, line 120: Does this need to be two different statements? Will `leading_whitespace_segments` not just be blank? So any harm in always using the section option?

jmc-bbk: Hey @tunetheweb, I initially tried that and found it broke test 1 & 2. It was incorrectly removing the white space from the trailing `SELECT ELEMENT`.

I’m not at my computer at the moment; but I can provide an example tomorrow!

tunetheweb: OK let's leave as is then.

src/sqlfluff/rules/L041.py
```diff
@@ -93,9 +101,24 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
-        # and delete the original select clause modifier.
-        fixes.extend((LintFix.delete(s) for s in leading_newline_segments))
+
+        # Delete original newlines and whitespace between select keyword
+        # and select clause modifier.
+
+        # If there is not a newline after the select clause modifier then delete
+        # newlines between the select keyword and select clause modifier.
+        if not trailing_newline_segments:
+            fixes.extend((LintFix.delete(s) for s in leading_newline_segments))
```
barrywhart, line 111: Nitpick: IIUC, the parentheses are not needed here. `LintFix.delete(s) for s in leading_newline_segments` is a generator expression, which can be passed directly to `extend()`.
```suggestion
            fixes.extend(LintFix.delete(s) for s in leading_newline_segments)
```

src/sqlfluff/rules/L041.py
```diff
@@ -93,9 +101,24 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+        # If there is a newline after the select clause modifier then delete both the
+        # newlines and whitespace between the select keyword and select clause modifier.
+        else:
+            fixes.extend(
+                (
```
barrywhart, line 116: Ditto the above comment.

barrywhart (approved):
Looks good. A couple of nitpicky suggestions for change (not 100% sure I'm correct about this, but I think so).

src/sqlfluff/rules/L041.py
tunetheweb: Test this and you are correct. Will change and merge.

PR #2985, opened 2022-04-03
dbt: Make `is_incremental()` defaults consistent

### LINKED ISSUE #2984: dbt builtin `is_incremental()` is very confusing in default configuration
### Search before asking

### What Happened

In current SQLFluff default configuration there exist "two" versions of the `is_incremental()` macro.

One defined via `_generate_dbt_builtins()` here https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/core/templaters/jinja.py#L185 which returns *false*

And a second one via `default_config.cfg` here https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/core/default_config.cfg which returns *true*

This can lead to extremely confusing behaviour when a user is not entirely familiar with SQLFluff internals.

### Expected Behaviour

`is_incremental()` should have a consistent default behaviour (probably `True` to enable coverage of incremental blocks in dbt).

However, I'm not a DBT expert so not 100% sure if there are any side-effects with this.

### Observed Behaviour

Inconsistent behaviour of `{{ is_incremental() }}` depending on how it's called.

### How to reproduce

`.sqlfluff`
```
weird_is_incremental = {% macro weird_is_incremental() %}{{ is_incremental() }}{%- endmacro -%}
```

```sql
SELECT
    {{ is_incremental() }},
    {{ weird_is_incremental() }}
```

Output
```
[L:  1, P:  1]      |file:
[L:  1, P:  7]      |                newline:                                      '\n'
[L:  2, P:  1]      |                whitespace:                                   '    '
[L:  2, P:  5]      |                select_clause_element:
[L:  2, P:  5]      |                    literal:                                  'True'
[L:  2, P: 27]      |                comma:                                        ','
[L:  2, P: 28]      |                newline:                                      '\n'
[L:  3, P:  1]      |                whitespace:                                   '    '
[L:  3, P:  5]      |                select_clause_element:
[L:  3, P:  5]      |                    literal:                                  'False'
[L:  3, P: 33]      |            [META] dedent:
```

### Dialect

ansi

### Version

==== sqlfluff ====
sqlfluff:               0.11.2 python:                 3.10.2
verbosity:                   2 templater:               jinja

### Configuration

```
weird_is_incremental = {% macro weird_is_incremental() %}{{ is_incremental() }}{%- endmacro -%}
```

### Code of Conduct

dmohns:
Between `_generate_dbt_builtins()` and `default_config.cfg`.

Not being an DBT expert I'm not sure if this will have other side effects then discussed below. Or if this behaviour was built like this on purpose.

Fixes #2984

Yes, if a user uses `is_incremental()` in a custom macro (consciously or not) their macro will have a different return value. Much like in `jinja_g_macro` test.

### Pull Request checklist

  - Other.


barrywhart (commented):
Makes sense that the defaults should be the same for both. I don't have an opinion on whether `True` or `False` is a better default. Any thoughts, @alanmcruickshank or @NiallRees?

tunetheweb:
> Makes sense that the defaults should be the same for both. I don't have an opinion on whether `True` or `False` is a better default. Any thoughts, @alanmcruickshank or @NiallRees?

Given the fact that the default config value is the default, and the code default will never actually be used. Changing the code config default to True (like this PR has done) seems sensible. Otherwise it's a breaking change for no good reason.

alanmcruickshank:
It's a dummy function just to make things work - I'm not aware of any meaningful thought on whether it should be one or the other - I'm pretty chill either way.

alanmcruickshank (approved):

PR #2986, opened 2022-04-03
Rename Grammars for consistency

tunetheweb:

Some of our Grammars are incorrectly named, either missing the Grammar at end of the name, or calling themselves Segments, when they are in fact Grammars.

Shouldn't be.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -134,16 +134,16 @@
         trim_chars=("@",),
     ),
     # Add a Full equivalent which also allow keywords
-    NakedIdentifierSegmentFull=RegexParser(
+    NakedIdentifierFullSegment=RegexParser(
```
barrywhart, line 137: Why no `Grammar` suffix here?

tunetheweb: Seems like all
The Identifiers are set up as Segments rather than Grammars. Which makes sense to me since they do get a name and a type. Do you disagree?

Btw there’s a few though that don’t do exactly this (e.g. `ParameterNameSegment` in BigQuery is a `OneOf` of two RegexParsers) but left that as was.

barrywhart: No strong feelings here, just kind of mechanically looking over the PR. Moving `Segment` to the end of the name seems like an improvement, regardless.

barrywhart (approved):

PR #2988, opened 2022-04-04
dbt: Read builtins from code

### LINKED ISSUE #2965: Using a Jinja variable called `var` breaks the templater
### Search before asking

### What Happened

When using the Jinja templater and configuring a variable called `var` the templating breaks with

```
Found unparsable section: "<Macro 'var'>"
```

Looks like it's trying to access a macro rather than a variable.

### Expected Behaviour

It should be possible to use `var` like every other variable name.

### Observed Behaviour

```
Found unparsable section: "<Macro 'var'>"
```

### How to reproduce

```sql
SELECT {{ varr }} FROM table_1;
SELECT {{ var }} FROM table_1;
```

With the following `.sqlfluff`

```
[sqlfluff]
templater = jinja

varr = 30
var = 30
```

yields

```
==== parsing violations ====
L:   2 | P:   8 |  PRS | Line 2, Position 8: Found unparsable section: "<Macro 'var'>"
```

### Dialect

ansi

### Version

```
==== sqlfluff ====
sqlfluff:               0.11.2 python:                 3.10.2
```

### Configuration

```
[sqlfluff]
templater = jinja

varr = 30
var = 30
```

### Code of Conduct

dmohns:

This PR changes the way dbt builtins are made available to the user. As part of this PR all dbt macros will be provided via `generate_dbt_builtins()` in the Jinja context. This removes the need to have macros hardcoded in `default_config`. Users can still fall back to overriding builtin macros via config if they so desire.

Using this PR as a discussion whether or not this has a place in SQLFluff.

**Why?**

- Currently, some (most) dbt builtins are made available via hardcoded macro presets in `default_config`. While others are made available via customised Jinja context. Trying to only have 1 place to handle it. From looking at some older commits it looks like this is just a legacy carry over, but not entirely sure here.
- Passing dbt builtins via macros has limitations. For example, `is_incremental()` is currently providing STRING `true` instead of BOOLEAN `true`. This accidentally works because most people call it as 
  ```sql
  {% if is_incremental() %}
  ```
- Users have no way of disabling the hardcoded macros rather than manually overriding them. See and fixes #2965

Yes. If a User has configured `apply_dbt_builtins = False` **but** is still using dbt macros in their SQL, this change will be breaking for them. However, I think this is a very unlikely scenario.

### Pull Request checklist

  - Other.


alanmcruickshank (approved):
I think this makes sense. The logic of being able to turn off the built-ins makes a lot of sense, and the default to keep them on I think is the right UX. The added benefit here of them returning the correct datatype feels like an added bonus 👍

PR #2993, opened 2022-04-04
TSQL: add support for parenthesized nested joins

pguyot:

T-SQL supports "nested joins" which sometimes can be parenthesized.
As a matter of fact, this seems to be in ANSI SQL 1999 standard, but it is not (?) supported in actual dialects such as PostgreSQL.

The attached patch allows for nested parenthesized joins by adding required parentheses. It does not exactly follow the ANSI standard grammar as the recursive definition will not work with the current parser.

### Pull Request checklist

  - Other.


test/fixtures/dialects/tsql/nested_joins.sql
```diff
@@ -18,3 +18,11 @@ RIGHT OUTER JOIN I
...
         ON I.SecondaryPID = CAST(P_2.IDEID AS varchar)
     ON CAST(BA.IDEID AS varchar) = I.BAID
+
+SELECT 1 AS RegionCode
+FROM BA
+LEFT OUTER JOIN (
+    I JOIN P
+        ON I.Pcd = P.Iid
+) ON BA.Iid = I.Bcd;
+GO
```
tunetheweb, line 28: When we lint this, it causes L026 to incorrectly fire:

```
L:   5 | P:  12 | L026 | Reference 'I.Pcd' refers to table/view not found in the
                       | FROM clause or found in ancestor statement.
L:   5 | P:  20 | L026 | Reference 'P.Iid' refers to table/view not found in the
                       | FROM clause or found in ancestor statement.
L:   6 | P:  15 | L026 | Reference 'I.Bcd' refers to table/view not found in the
                       | FROM clause or found in ancestor statement.
```

Think we need to see if we can tweak L026 to account for this before we can merge

tunetheweb (changes requested):

test/fixtures/dialects/tsql/nested_joins.sql
tunetheweb: Think I've found the fix for this so pushing to this branch if you don't mind.

pguyot: Excellent. I tried to do something similar but your code is definitely nicer.

tunetheweb (approved):

tunetheweb:
@barrywhart if you have a chance could you look over this since I had a hand in the code?

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1168,7 +1168,13 @@ def get_eventual_alias(self) -> Optional[AliasInfo]:
         tbl_expression = self.get_child("table_expression")
         if not tbl_expression:  # pragma: no cover
             tbl_expression = self.get_child("bracketed").get_child("table_expression")
-        ref = tbl_expression.get_child("object_reference")
+        # For TSQL nested, bracketed tables get the first table as reference
+        if tbl_expression and not tbl_expression.get_child("object_reference"):
+            if tbl_expression.get_child("bracketed"):
+                tbl_expression = tbl_expression.get_child("bracketed").get_child(
```
barrywhart, line 1174: Nitpick: I think this will work if there's _at most_ one level of parentheses. Should we worry about handling two or more levels?

barrywhart (approved):
Looks good. One question about whether we want to test for & handle multiple levels of parentheses.

src/sqlfluff/dialects/dialect_ansi.py
tunetheweb: I’m ok leaving this as is for now. It’s similar to the check above with similar restrictions. Also once you start allowing multiple levels you could be in expressions and sub clauses and all sorts. So I say leave for now.

barrywhart: I agree. Just wanted to make it a conscious decision. Might be worth adding a comment noting this, but really no big deal.

tunetheweb: Let’s leave for now. Not even sure the syntax is valid with multiple delimiters and comment might suggest there definitely is a valid syntax like this that is not supported.

PR #2995, opened 2022-04-04
Force user to specify a dialect

### LINKED ISSUE #2960: Improve usability for new users: Make them specify a dialect (don't default to ANSI)
### Search before asking

### Description

When new SQLFluff users often encounter parsing issues, it often turns out that they were using the default dialect (ANSI) rather than the dialect for the database they're using.

From a SQLFluff _developer_ perspective, having ANSI as a default dialect makes sense. But from an _end-user_ perspective, I think it's a bad idea because there's no good default for real users; there are so many quirks and differences that I think it's better to force them to specify.

Let's look into this and see if it's doable, is it a good idea, etc.

### Use case

When I run SQLFluff from the command line, if I haven't specified `--dialect` on the command line or in `.sqlfluff`, it should fail and tell me to pick one. There's a somewhat similar message that's printed on parse failure -- "have you configured a dialect"? This would be similarly, but more strongly worded. That other message could perhaps be removed.

Can we also change this in the API? That may be trickier than changing the command line behavior -- not sure.

### Dialect

n/a

### Code of Conduct

barrywhart:
Fixes #2960 

`FluffConfig` now raises an exception if no dialect was specified in `.sqlfluff` or on the command line. An override is available for those situations where a config is needed but won't be used for linting, e.g. the `sqlfluff dialects` command.

This PR is relatively simple, but it touches a lot of code. We may need to revisit this work and tweak things a bit if we accidentally broke a use case we didn't consider (or made something excessively annoying).

### Pull Request checklist

  - Other.


.sqlfluff
```diff
@@ -0,0 +1,2 @@
+[sqlfluff]
+dialect = ansi
```
tunetheweb, line 2: I’m not sure if this is going to make it into the final PR, or if this is temporary for now but I’m not loving this for a number of reasons:

1. it feels like a bit of a hack. You must specify a dialect, but we don’t really. It teaches people developing not to specify a dialect.
2. When running SQL from this the repo it will work, but running elsewhere will fail. It will be confusing.
3. I often add a specific `.sqlfluff` config file while developing and testing but definitively do not want that checked in but that might happen now. Though guessing testing would pick it up.

However I do recognise how many places you’re likely to need to specify a dialect without this :-( Luckily for dialect test fixtures you don’t need to as we have one for each of those, but for lots of other test files you might now need to.

barrywhart: Just trying to get the tests to pass now. Will look for better ways to do things once they do.

barrywhart: I'll keep SQLFLUFF_TESTENV in mind. Ideally, we can get all the tests working with changes on the test side, minimizing changes to the production code (i.e. stuff in `src/sqlfluff/`).

test/fixtures/linter/passing.sql
```diff
@@ -1 +1 @@
-select a from b
\ No newline at end of file
+select a from b
```
barrywhart, line 1: Given the filename, I assume this file was meant to pass linting, but it was missing a trailing newline, which caused it to fail. See `test/cli/commands_test.py::TestProgressBars` for some of the affected tests.

test/cli/commands_test.py
```diff
@@ -1220,35 +1263,36 @@ class TestProgressBars:
...
         assert "\r\rlint by rules:" not in raw_output
 
     def test_cli_lint_enabled_progress_bar(
-        self, mock_disable_progress_bar: MagicMock
+        self, mock_disable_progress_bar: MagicMock, monkeypatch
     ) -> None:
         """When progress bar is enabled, there should be some tracks in output."""
+        monkeypatch.chdir("test/fixtures/")
         result = invoke_assert_code(
-            ret_code=65,
```
barrywhart, line 1247: IIUC, checking for `exit_code=65` was because `passing.sql` was actually not passing the linter. I fixed `passing.sql`, so now the exit code is 0, indicating there were no linting issues.

test/cli/commands_test.py
```diff
@@ -1300,22 +1348,3 @@ def test_cli_lint_enabled_progress_bar_multiple_files(
         assert r"\rlint by rules:" in raw_output
         assert r"\rrule L001:" in raw_output
         assert r"\rrule L049:" in raw_output
-
-    def test_cli_lint_disabled_progress_bar_when_verbose_mode(
```
barrywhart, line 1304: This test has issues that I couldn't figure out, even in the `main` branch. More notes below.

test/cli/commands_test.py
```diff
@@ -1300,22 +1348,3 @@ def test_cli_lint_enabled_progress_bar_multiple_files(
...
-        self, mock_disable_progress_bar: MagicMock
-    ) -> None:
-        """Progressbar is disabled when verbose mode is set."""
-        result = invoke_assert_code(
-            ret_code=2,
-            args=[
-                lint,
-                [
-                    "-v" "test/fixtures/linter/passing.sql",
```
barrywhart, line 1313: Note that the test is checking for an error exit code above (`2`). Why? Because of this line. Notice there's no comma between `-v` and the path. This is invalid, so the automatic command-line error checking in `click` causes it to fail before it even gets into the `lint` command. After I fixed this, it seemed like the progress bar *was* enabled, but I don't see the same behavior when I manually test the same scenario. I think some of the class or test-level mocking may be involved.

src/sqlfluff/cli/commands.py
```diff
@@ -491,7 +495,7 @@ def lint(
     extra_config_path: Optional[str] = None,
     ignore_local_config: bool = False,
     **kwargs,
-) -> NoReturn:
+) -> None:
```
barrywhart, line 498: Unrelated change: The correct type hint is `None`. IIUC, `NoReturn` means the function never returns (i.e. always throws an exception). PyCharm was flagging this.

src/sqlfluff/core/config.py
```diff
@@ -455,6 +455,7 @@ def __init__(
         ignore_local_config: bool = False,
         overrides: Optional[dict] = None,
         plugin_manager: Optional[pluggy.PluginManager] = None,
+        require_dialect: bool = True,
```
barrywhart, line 458: By default, config requires a dialect to be set, but this can be overridden.

src/sqlfluff/core/default_config.cfg
```diff
@@ -5,7 +5,7 @@ verbose = 0
 nocolor = False
 # Supported dialects https://docs.sqlfluff.com/en/stable/dialects.html
 # Or run 'sqlfluff dialects'
-dialect = ansi
+dialect = None
```
barrywhart, line 8: This is a key part of the change. The built-in default config no longer specifies a dialect.

src/sqlfluff/core/dialects/__init__.py
```diff
@@ -81,7 +81,7 @@ def dialect_readout() -> Iterator[DialectTuple]:
 
 def dialect_selector(s: str) -> Dialect:
     """Return a dialect given its name."""
-    dialect = load_raw_dialect(s or "ansi")
+    dialect = load_raw_dialect(s)
```
barrywhart, line 84: Here, we no longer hardcode `"ansi"` as a fallback if no dialect was provided. Let it fail if it gets this far.

src/sqlfluff/rules/L009.py
```diff
@@ -106,12 +106,13 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
             select_if=sp.is_type("newline"),
             loop_while=sp.or_(sp.is_whitespace(), sp.is_type("dedent")),
         )
-        trailing_literal_newlines = trailing_newlines.select(
```
barrywhart, line 109: Unrelated fix to a bug that surfaced during testing. Some of the automated tests don't provide `context.templated_file`, so the rule was throwing a runtime error trying to use it.

test/cli/commands_test.py
```diff
@@ -52,22 +52,23 @@ def invoke_assert_code(
     return result
 
 
-expected_output = """== [test/fixtures/linter/indentation_error_simple.sql] FAIL
```
barrywhart, line 55: Many changes in this file. I generally updated these tests in one of two ways:
* `chdir` to the `test/fixtures/` directory, which now has a `.sqlfluff` file that specifies `dialect=ansi`
* Specify `--dialect=ansi` on the command line.

test/core/config_test.py
```diff
@@ -297,3 +304,10 @@ def test__config__from_kwargs():
     assert cfg.get("dialect") == "snowflake"
     assert cfg.get("rules") == "L001,L002"
     assert cfg.get("exclude_rules") == "L010,L011"
+
+
+def test__config_missing_dialect():
```
barrywhart, line 309: New test added so the "must configure a dialect" code gets coverage.

plugins/sqlfluff-templater-dbt/test/templater_test.py
```diff
@@ -421,7 +422,8 @@ def test__templater_dbt_handle_database_connection_failure(
 def test__project_dir_does_not_exist_error(dbt_templater, caplog):  # noqa: F811
     """Test an error is logged if the given dbt project directory doesn't exist."""
     dbt_templater.sqlfluff_config = FluffConfig(
-        configs={"templater": {"dbt": {"project_dir": "./non_existing_directory"}}}
+        configs={"templater": {"dbt": {"project_dir": "./non_existing_directory"}}},
+        overrides={"dialect": "ansi"},
```
tunetheweb, line 426: Why is one ` configs` and the other an `overrides` out of interest? Is this cause a default `templater` exists but a default `dialect` no longer does now?

src/sqlfluff/cli/commands.py
```diff
@@ -175,14 +175,14 @@ def core_options(f: Callable) -> Callable:
         f = click.option(
             "--dialect",
             default=None,
-            help="The dialect of SQL to lint (default=ansi)",
+            help="The dialect of SQL to lint",
```
tunetheweb, line 178: I wonder if we should make this mandatory?

```suggestion
            help="The dialect of SQL to lint (mandatory)",
```

src/sqlfluff/cli/commands.py
```diff
@@ -175,14 +175,14 @@ def core_options(f: Callable) -> Callable:
...
             shell_complete=dialect_shell_complete,
         )(f)
     else:  # pragma: no cover
         f = click.option(
             "--dialect",
             default=None,
-            help="The dialect of SQL to lint (default=ansi)",
+            help="The dialect of SQL to lint",
```
tunetheweb, line 185:
```suggestion
            help="The dialect of SQL to lint (mandatory)",
```

src/sqlfluff/cli/formatters.py
```diff
@@ -252,7 +252,10 @@ def _format_config(self, linter: Linter) -> str:
                 ("sqlfluff", get_package_version()),
                 ("python", get_python_version()),
                 ("implementation", get_python_implementation()),
-                ("dialect", linter.dialect.name),
+            ]
+            if linter.dialect:
+                config_content.append(("dialect", linter.dialect.name))
+            config_content += [
                 ("verbosity", self._verbosity),
             ] + linter.templater.config_pairs()
```
tunetheweb, line 260: Should we move this up to the previous section to make this a little more readable:

```py
            config_content = [
                ("sqlfluff", get_package_version()),
                ("python", get_python_version()),
                ("implementation", get_python_implementation()),
                ("dialect", linter.dialect.name),
                ("verbosity", self._verbosity),
            ]
            if linter.dialect:
                config_content.append(("dialect", linter.dialect.name))
            config_content += linter.templater.config_pairs()
```

src/sqlfluff/core/config.py
tunetheweb: Presumably this should only really be overridden for tests? Worth a comment?

Edit: Ah see further down it can also not be set at top level (my test case!). OK definitely think worthy of a comment.

```suggestion
        # Ideally a dialect should be set when config is read but sometimes
        # it might only be set in nested .sqlfluff config files, so allow it
        # to be not required.
        require_dialect: bool = True,
```

src/sqlfluff/core/config.py
```diff
@@ -503,13 +504,32 @@ def __init__(
...
-        self._configs["core"]["dialect_obj"] = dialect_selector(
-            self._configs["core"]["dialect"]
-        )
+        dialect: Optional[str] = self._configs["core"]["dialect"]
+        if dialect is not None:
+            self._configs["core"]["dialect_obj"] = dialect_selector(
+                self._configs["core"]["dialect"]
+            )
+        elif require_dialect:
+            self.verify_dialect_specified()
```
tunetheweb, line 513: This reads a little weird

```
if dialect is selected
   do something
else: # (so dialect is obviously NOT selected)
  verify if dialect is selected # even though we know it is not!!!
```

Would it be better to just drop the `elif`? Would also allow us to check if the above `dialect_selector` call failed.

```py
        if dialect is not None:
            self._configs["core"]["dialect_obj"] = dialect_selector(
                self._configs["core"]["dialect"]
            )
        self.verify_dialect_specified()
```

test/cli/commands_test.py
tunetheweb: Should we assert ret_code=0 then?

test/cli/commands_test.py
tunetheweb: Yeah confirmed on main that `-v` does not disable the progress bar. Found #1873 which fixed the fact it didn't work with progress bars, so agree this test can be removed as should have been removed then.

tunetheweb (approved):
Looks pretty good. A few comments, but nothing major.

plugins/sqlfluff-templater-dbt/test/templater_test.py
barrywhart: Based on this line in the `FluffConfig` constructor, I'm pretty sure either would work:
```
        self._configs = nested_combine(
            defaults, configs or {"core": {}}, {"core": overrides or {}}
        )
```

Once I found a way to get tests passing, I did a lot of copy paste. I'll look at changing this for simplicity.

barrywhart: Yes, that worked. Updating a couple of dbt tests.

src/sqlfluff/cli/commands.py
barrywhart: You can also specify the dialect in `.sqlfluff`, so I wonder if this would be confusing.

tunetheweb: Fair point. Let's leave it as is.

src/sqlfluff/cli/formatters.py
barrywhart: We could. I was trying to keep the same order, in case it broke some tests, but I'll give it a try.

src/sqlfluff/core/config.py
barrywhart: I'm worried this is one of those situations where I'm going to try and clean up the code but break it instead. Let's leave this as is, okay? Maybe another time...

barrywhart: We need to check the `require_dialect` flag, because if it's false, we don't want to call `verify_dialect_specified()` because it would definitely fail, which we don't want in this case.

There may be cleaner ways to do parts of this at the code level. Trying to retrofit smarter behavior into code that was built with certain assumptions is often messy on the first go. I think it works, and more worried about breaking it than leaving it a bit ugly.

tunetheweb: Grand. Not a big deal.

PR #3004, opened 2022-04-05
BigQuery: Support `AT TIME ZONE` in `EXTRACT`

tunetheweb:
### Brief summary of the change made

Fixes #3003 

### Are there any other side effects of this change that we should be aware of?
I moved the `AtTimeZone` Segment from Postgres to ANSI as seems like quite a few dialects support this.

I also changed it form a Quoted Literal to an Expression as per comment below. This changed some of the Postgres YAMLs.

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #3007, opened 2022-04-05
Support `TRIM` function parameters

### LINKED ISSUE #1935: Parsing issues with Spark trim command for both/leading/trailing
Spark SQL has a `TRIM` function for trimming strings: https://spark.apache.org/docs/3.0.0/api/sql/index.html#trim. I'm running into parsing issues around these (initially prompted by a `BOTH FROM` command).

## Expected Behaviour

From the link above, we have the following example strings (which should all parse):

```sql
> SELECT trim('    SparkSQL   ');
 SparkSQL
> SELECT trim(BOTH FROM '    SparkSQL   ');
 SparkSQL
> SELECT trim(LEADING FROM '    SparkSQL   ');
 SparkSQL
> SELECT trim(TRAILING FROM '    SparkSQL   ');
     SparkSQL
> SELECT trim('SL' FROM 'SSparkSQLS');
 parkSQ
> SELECT trim(BOTH 'SL' FROM 'SSparkSQLS');
 parkSQ
> SELECT trim(LEADING 'SL' FROM 'SSparkSQLS');
 parkSQLS
> SELECT trim(TRAILING 'SL' FROM 'SSparkSQLS');
 SSparkSQ
```

## Observed Behaviour

Running each of these into `echo "X" | sqlfluff parse --dialect spark3 -` and looking for parsing errors:
- `SELECT trim('    SparkSQL   ');` succeeds
- `SELECT trim(BOTH FROM '    SparkSQL   ');` fails
- `SELECT trim(LEADING FROM '    SparkSQL   ');` fails
- `SELECT trim(TRAILING FROM '    SparkSQL   ');` fails
- `SELECT trim('SL' FROM 'SSparkSQLS');` succeeds
- `SELECT trim(BOTH 'SL' FROM 'SSparkSQLS');` fails
- `SELECT trim(LEADING 'SL' FROM 'SSparkSQLS');` fails
- `SELECT trim(TRAILING 'SL' FROM 'SSparkSQLS');` fails

## Steps to Reproduce

See above.

## Dialect

Spark3

## Version

```
python==3.9.7
sqlfluff==0.8.2
```

## Configuration

Default

tunetheweb:

Fixes #1935 

These seem to be supported in many dialects. For those that don't support this I've overridden the Grammar to `Nothing()` to prevent it trying to find Keywords that don't exist.

### Pull Request checklist

  - Other.


test/fixtures/dialects/ansi/trim_functions.yml
```diff
@@ -0,0 +1,146 @@
...
+      select_clause:
+        keyword: SELECT
+        select_clause_element:
+          function:
+            function_name:
+              function_name_identifier: trim
+            bracketed:
+            - start_bracket: (
+            - expression:
+                column_reference:
```
WittierDinosaur, line 51: Is it an issue that these are appearing as column references?

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -943,6 +944,13 @@ class QualifiedNumericLiteralSegment(BaseSegment):
         Ref("ExpressionSegment"),
         # A Cast-like function
         Sequence(Ref("ExpressionSegment"), "AS", Ref("DatatypeSegment")),
+        # Trim function
+        Sequence(
```
barrywhart, line 948: Are we checking the function name, or can this be used with any function? Seems like the latter. Any concerns with allowing that?

barrywhart (approved):
One question about whether to allow these new parameters with function names other than `TRIM`. Overall, looks good.

src/sqlfluff/dialects/dialect_ansi.py
tunetheweb: Yeah I thought about that. You're right we do allow with any function. But very specific set of syntax and grammar so think will only match that.

This who `FunctionContentsGrammar` is made up of lots of syntax types that could be specific to certain function names but isn't, and this PR continues that theme.

To limit it to certain function names would require a change similar to #3004 where we get around it being a keyword (so it's still a function name type). But that gets messy and potentially unmaintainable pretty quickly. Hence I did initially implement #3004 in a similar manner to this. However, as that was specific to BigQuery, and similar to another pattern in Function I changed my mind on that.

For this TRIM one, I think we follow the previous pattern, as more generic and unlikely to match incorrectly.

PR #3012, opened 2022-04-05
Separate slow CI job to it's own workflow

tunetheweb:

When GitHub Actions or Coverage experiences issues (as they are want to do!) we sometimes have to rerun the CI jobs. The Critical Rules Test is slow to run, but doesn't impact Coverage, so separate that out to it's own job so it doesn't need to be rerun if it passes.

Nope.

### Pull Request checklist

  - Other.

### DIFF
--- a/.github/workflows/critical-rules-test.yml
+++ b/.github/workflows/critical-rules-test.yml
@@ -12,6 +12,9 @@
 name: Critical CI Tests
 on:
   workflow_dispatch:
+  schedule:
+    # 2am each night
+    - cron: '00 2 * * *'
   pull_request:
   push:
     branches:
--- a/.github/workflows/critical-rules-test.yml
+++ b/.github/workflows/critical-rules-test.yml
@@ -1,0 +1,36 @@
+#############################
+## GitHub Actions CI Tests ##
+#############################
+#
+# This lints all our dialect fixtures to check rules can handle a variety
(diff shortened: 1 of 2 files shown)


pwildenhain (approved):
Super sensible 👍🏻

.github/workflows/critical-rules-test.yml
```diff
@@ -0,0 +1,39 @@
...
+# It's run as a separate job as takes longer than the CI jobs and allows
+# them to be rerun separately if GitHub Actions or Coverage is experiencing
+# issues.
+#
+name: Critical CI Tests
+on:
+  workflow_dispatch:
+  schedule:
+    # 2am each night
+    - cron: '00 2 * * *'
```
barrywhart, line 17: Since it's running on a schedule, how will we be alerted if it fails? Email?

tunetheweb: The GitHub altering on scheduled runs is a bit lacking to be honest. Long been complained about: https://github.community/t/no-email-notification-on-schedule-cron-job-failure/119638/5

I think only the last person to alter it (me!) gets a notification. Which sucks to be honest.

Now to be honest this cron job isn’t really needed. I like it in case there is no activity on repo for a while (not something we have a problem with!!) so can see when something broke if, for example, dependencies update.

PR #3013, opened 2022-04-05
SparkSQL: Adding support for Delta Lake table schema updates

R7L208:

This proposes additions to support Delta Lake-specific `ALTER TABLE` syntax.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_sparksql.py
```diff
@@ -691,25 +691,59 @@ class AlterTableStatementSegment(ansi.AlterTableStatementSegment):
...
                 OneOf("ALTER", "CHANGE"),
-                "COLUMN",
-                Ref("ColumnReferenceSegment"),
-                Sequence("TYPE", Ref("DatatypeSegment"), optional=True),
+                Ref.keyword("COLUMN", optional=True),
+                Indent,
+                AnyNumberOf(
+                    Ref("ColumnReferenceSegment"),  # matches WHATEVER IS NEXT
+                    max_times=2,
+                ),
```
R7L208, line 720: This is causing problems, and I haven't found a great solution. Hoping to get some feedback.

Delta supports renaming columns in place so the syntax would be `old_col_name new_col_name`. This isn't supported in native SparkSQL. Adding this snippet causes `test/fixtures/dialects/sparksql/alter_table.sql` to fail mainly because whatever comes after the one and only column specified also gets matched as a column.

As coded in the test cases, its `COMMENT` but I've played around with adjusting the test case to verify this is what's happening. Using `SingleIdentifierGrammar` causes the same issue but matches both as `keyword` instead of `column_reference`.

I'm guessing expanding the reserved keywords will be the easiest way to solve this, but my preference is to leave keywords as-is and trying to identify an alternative solution.

test/fixtures/dialects/sparksql/delta_update_table_schema.sql
```diff
@@ -0,0 +1,60 @@
...
+ALTER TABLE table_name CHANGE COLUMN col_name_old col_name_new STRING;
+
+ALTER TABLE table_name CHANGE COLUMN
+    col_name_old col_name_new STRING COMMENT "new_col_comment";
+
+ALTER TABLE table_name CHANGE COLUMN
+    col_name_old col_name_new STRING COMMENT "new_col_comment" FIRST;
+
+ALTER TABLE table_name CHANGE COLUMN
+    col_name_old col_name_new STRING COMMENT "new_col_comment" AFTER col_a_name;
```
R7L208, line 26: Relevant Delta test cases

src/sqlfluff/dialects/dialect_sparksql.py
R7L208:
```sql
-- relevant sparksql test cases
-- ALTER OR CHANGE COLUMNS
ALTER TABLE StudentInfo ALTER COLUMN Name COMMENT "new comment";

ALTER TABLE StudentInfo CHANGE COLUMN Name COMMENT "new comment";
```

```sql
-- relevant delta test cases
-- change column comment/ordering
ALTER TABLE table_name CHANGE col_name_old col_name_new STRING;

ALTER TABLE table_name CHANGE COLUMN col_name_old col_name_new STRING;

ALTER TABLE table_name CHANGE COLUMN
    col_name_old col_name_new STRING COMMENT "new_col_comment";

ALTER TABLE table_name CHANGE COLUMN
    col_name_old col_name_new STRING COMMENT "new_col_comment" FIRST;

ALTER TABLE table_name CHANGE COLUMN
    col_name_old col_name_new STRING COMMENT "new_col_comment" AFTER col_a_name;
```

tunetheweb: This is what you want:

```py
                AnyNumberOf(
                    OneOf(
                        exclude=Ref.keyword("COMMENT")
                    ),
                    max_times=2,
                ),
```

Or if there's more you want to exclude:

```py
                AnyNumberOf(
                    OneOf(
                        exclude=OneOf(
                            "COMMENT",
                            "TYPE",
                            Ref("DatatypeSegment"),
                            "FIRST",
                            "AFTER",
                            "SET",
                        ),
                    ),
                    max_times=2,
                ),
```

tunetheweb: Note you have to wrap the `Ref("ColumnReferenceSegment"),` in a `OneOf` as `Ref` doesn't have an `exclude` option (maybe we should add it?) but `OneOf` does.

R7L208: thank you! I forgot about `exclude`.

For exclude in `Ref`, I haven't seen a need for it in SparkSQL. The only use I can think of would be breaking up large `exclude` arguments like in `AliasExpressionSegment` which could increase readability and make it easier to use such segments in new places.

For `AliasExpressionSegment` in particular though, given the sequence of segments almost that entire `OneOf` in the exclude argument would have to be defined each place `AliasExpressionSegment` is used and would create a lot of redundant code.

I don't really see a benefit for it inside of SparkSQL itself right now but good to keep in mind.

R7L208: @tunetheweb - Going back to this. I'm now realizing that `exclude` doesn't work inside of `AnyNumberOf` without `OneOf`. 😁 

So going back "maybe we should add it?", would probably be helpful

R7L208: > @tunetheweb - Going back to this. I'm now realizing that `exclude` doesn't work inside of `AnyNumberOf` without `OneOf`. 😁
> 
> So going back to "maybe we should add it?", would probably be helpful

I started looking at what would be required for this. I'm happy to do it either as part of this PR or another but think it would be beneficial. To be honest, would need some guidance on what changes need to be made in `base.py` to include the `exclude` argument for `Ref` but happy to do the work.

tunetheweb (approved):

src/sqlfluff/dialects/dialect_sparksql.py
tunetheweb: Actually let's merge this, and then I'll clean up in #3028 as waiting on one of the other maintainers reviewing that.

tunetheweb: Merged and updated in my PR now.

PR #3016, opened 2022-04-06
Postgres better `CREATE USER`/`CREATE ROLE` support

### LINKED ISSUE #3015: Parse error with postgres CREATE USER statements
### Search before asking

### What Happened

When linting postgres sql with `CREATE USER` statements parsing errors are thrown

### Expected Behaviour

As `CREATE USER` is an alias of `CREATE ROLE` in postgres it should be parsed the same.

### Observed Behaviour

```
❯ sqlfluff lint --dialect=postgres test.sql
== [test.sql] FAIL
                       | USER steve'
L:   2 | P:   1 |  PRS | Line 2, Position 1: Found unparsable section: 'CREATE
                       | USER frank WITH CONNECTION LIMIT ...'
L:   3 | P:   1 |  PRS | Line 3, Position 1: Found unparsable section: 'CREATE
                       | USER frank WITH IN ROLE frank'
L:   4 | P:   1 |  PRS | Line 4, Position 1: Found unparsable section: 'CREATE
                       | USER frank WITH SUPERUSER CREATED...'
L:   5 | P:   1 |  PRS | Line 5, Position 1: Found unparsable section: 'CREATE
                       | USER frank WITH INHERIT LOGIN REP...'
L:   6 | P:   1 |  PRS | Line 6, Position 1: Found unparsable section: "CREATE
                       | USER davide WITH PASSWORD 'jw8s0F..."
L:   7 | P:   1 |  PRS | Line 7, Position 1: Found unparsable section: "CREATE
                       | USER miriam WITH LOGIN PASSWORD '..."
L:   8 | P:   1 |  PRS | Line 8, Position 1: Found unparsable section: 'CREATE
                       | USER admin WITH CREATEDB CREATERO...'
```

### How to reproduce

Following examples from postgres docs as well as created all throw error
```
CREATE USER steve;
CREATE USER frank WITH CONNECTION LIMIT 1;
CREATE USER frank WITH IN ROLE frank;
CREATE USER frank WITH SUPERUSER CREATEDB CREATEROLE;
CREATE USER frank WITH INHERIT LOGIN REPLICATION BYPASSRLS;
CREATE USER davide WITH PASSWORD 'jw8s0F4';
CREATE USER miriam WITH LOGIN PASSWORD 'jw8s0F4' VALID UNTIL '2005-01-01';
CREATE USER admin WITH CREATEDB CREATEROLE;
```

### Dialect

postgres

### Version

0.11.2

### Configuration

n/a

### Code of Conduct

mrf:
Created new Segment for create user. Applies same parameter parsing for CREATE ROLE to make the two consistent.
fixes #3015 

Existing CREATE ROLE behavior may be affected as the original config was more permissive.

### Pull Request checklist
~I was unable to generate new test fixtures due to `RuntimeError: Grammar refers to 'NoreplicationKeywordSegment' which was not found in the postgres dialect` error but do have additional test cases ready to go documented in linked issue~
Updated test cases and harness now included


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1,5 +1,5 @@
 """The PostgreSQL dialect."""
-
+from abc import ABC
```
tunetheweb, line 2: Presume this isn't needed?

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1081,12 +1081,62 @@ class CreateRoleStatementSegment(ansi.CreateRoleStatementSegment):
...
+                    OneOf("BYPASSRLS", "NOBYPASSRLS"),
+                    Sequence("CONNECTION", "LIMIT", Ref("NumericLiteralSegment")),
+                    Sequence("PASSWORD", OneOf(Ref("QuotedLiteralSegment"), "NULL")),
+                    Sequence("VALID", "UNTIL", Ref("QuotedLiteralSegment")),
+                    Sequence("IN", "ROLE", Ref("RoleReferenceSegment")),
+                    Sequence("IN", "GROUP", Ref("RoleReferenceSegment")),
+                    Sequence("ROLE", Ref("RoleReferenceSegment")),
+                    Sequence("ADMIN", Ref("RoleReferenceSegment")),
+                    Sequence("USER", Ref("RoleReferenceSegment")),
+                    Sequence("SYSID", Ref("NumericLiteralSegment")),
```
tunetheweb, line 1100: Any keywords added to the grammar need to be added to `src/sqlfluff/dialects/dialect_postgres_keywords.py`. That will also solve your YAML generation error.

tunetheweb (changes requested):

tunetheweb:
Also can you merge with `main` and resolve the conflicts?

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1081,12 +1081,62 @@ class CreateRoleStatementSegment(ansi.CreateRoleStatementSegment):
...
+                ),
             )
         ],
     )
 
+class CreateUserStatementSegment(BaseSegment):
```
tunetheweb, line 1106: As this is a new class you will need to add it to StatementSegment.

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1081,12 +1081,62 @@ class CreateRoleStatementSegment(ansi.CreateRoleStatementSegment):
...
+                Sequence("VALID", "UNTIL", Ref("QuotedLiteralSegment")),
+                Sequence("IN", "ROLE", Ref("RoleReferenceSegment")),
+                Sequence("IN", "GROUP", Ref("RoleReferenceSegment")),
+                Sequence("ROLE", Ref("RoleReferenceSegment")),
+                Sequence("ADMIN", Ref("RoleReferenceSegment")),
+                Sequence("USER", Ref("RoleReferenceSegment")),
+                Sequence("SYSID", Ref("NumericLiteralSegment")),
+            ),
+        ),
+    )
```
tunetheweb, line 1139: If this is identical to above you should be able to do this, to reduce code duplicate.

```suggestion
    match_grammar = postgres.CreateRoleStatementSegment
```

mrf: It's unfortunately not 100% identical as the ANSI definition is "CREATE" "ROLE" for the first two lines and this is "CREATE" "USER". Is there a way for me to merge them or do a partial override?

tunetheweb: We could just have `CREATE USER` be part of `CREATE ROLE`?

Or could try to do an `insert` and a `remove`, though not 100% sure it will work:
```py
match_grammar = postgres.CreateRoleStatementSegment.match_grammar.copy(
    insert=[Ref.keyword("USER")],
    before=Ref.keyword("ROLE"),
    remove=[Ref.keyword("ROLE")],
)
```

Or we could create a `UserRoleGrammar` that both can refer to?

WDYT?

mrf: I like the idea of combining the two, postgres docs make it clear that USER is just an alias of ROLE in how they handle it so would make sense to do the same treatment here. Let me give that a pass.

test/fixtures/dialects/postgres/postgres_create_role.yml
```diff
@@ -3,26 +3,189 @@
 # computed by SQLFluff when running the tests. Please run
 # `python test/generate_parse_fixture_yml.py`  to generate them after adding or
 # altering SQL files.
-_hash: 8e48269b8cd80d99bf7ffdeb0458e75ffdd8c46adfafb5848697400671b310c5
+_hash: 3c065b7b23271c44a013142a8e90b5338e344df2496bafe7752ec4b47a1a7eb0
 file:
+- statement:
+    create_user:
```
tunetheweb, line 9: Looks like you need to regenerate your YAML files since `create_user` no longer exists as a type now as these will be `create_role`.

mrf:
Not sure why the redshift keyword additions were necessary, but kept getting `RuntimeError: Grammar refers to 'SuperuserKeywordSegment' which was not found in the redshift dialect` type errors until I added those after my latest change. I could use another set of eyes from someone that understands that inheritance between those two dialects to make sure that change is ok.

tunetheweb:
> Not sure why the redshift keyword additions were necessary, but kept getting `RuntimeError: Grammar refers to 'SuperuserKeywordSegment' which was not found in the redshift dialect` type errors until I added those after my latest change. I could use another set of eyes from someone that understands that inheritance between those two dialects to make sure that change is ok.

Redshift inherits from Postgres (think they started from the same, but have branched apart now - but it's still a better base than ANSI).

However, it defines it's own keywords. So when you add syntax to Postgres, you also add it to Redshift (which is what we want most of the time!) unless you override it to redefine it. Which also means you need to add the keywords to Redshift, since it defines it's own keyword list.

That make sense?

It's the same when adding syntax to our ANSI dialect - need to check if ALL the dialects need those keywords, or if only some do (e.g. if that bit of syntax can never be reached from some dialects).

I do wonder if we should (or if we can?) define Unreserved keywords automatically if not explicitly listed? From the error message we obviously know it's a keyword, so might be able to default to add it to the syntax. Wouldn't want to do that for Reserved keywords obviously, but for Unreserved might be OK - with the downside that an error is more obvious to developer to highlight the issue and let them decide if it's a typo, or a reserved keyword or yes an unreserved keyword (probably the most likely).

mrf:
Thanks for review and explanation @tunetheweb much appreciated!

PR #2831, opened 2022-03-09
Delimited Refactor

WittierDinosaur:
This PR refactors Delimited(), to remove recursion.

It instead implements a method whereby it oscillates between matching for the matchers, and the delimiters. A few changes that went with it:
- adds the concept of a terminator to longest trimmed match, such that it can exit gracefully when the longest possible match has been found
- changes the logic so that longer matches are only preferred if they are still longer when trimmed

This PR also fixes the greedy problem (delimiters didn't respect their surrounding grammars, causing parsing errors and/or suboptimal parsing).

Numerous dialect changes have gone in to remove grammar syntax that relied on greediness as a crutch.

This massively improves the performance on long queries. Running sqlfluff parse --dialect=postgres on:
```
SET SESSION datestyle TO postgres, 'dmy'; * 1000
```
Parses in 66 seconds on this branch. On the main branch, it takes 5065 seconds. Also, the parsing of the test suite (mostly small scripts) take about 60% as long as it does on main.

Fixes #2427
Fixes #2152
Makes progress on #2269

PR #3024, opened 2022-04-06
Remove Delimited workarounds

tunetheweb:

Thanks to the excellent work by @WittierDinosaur in #2831 we no longer need to manually construct comma delimited grammars. This PR reverts all the previous use of this pattern.

No. there was one change to YAML file, but this looks to be a good change to me, and previous was incorrect.

### Pull Request checklist

  - Other.


test/fixtures/dialects/tsql/stored_procedure_begin_end.yml
```diff
@@ -76,8 +76,7 @@ file:
       - parameter: '@out_vchMsg'
       - data_type:
           data_type_identifier: uddt_output_msg
-      - data_type:
-          data_type_identifier: OUTPUT
+      - keyword: OUTPUT
```
tunetheweb, line 79: These are good changes. Should never have been flagged as a data_type before.

PhilippLange:
There is one more instance of this in sparksql (line 849):
```python
Bracketed(
                    # Manually rebuild Delimited.
                    # Delimited breaks complex (MAP, STRUCT) datatypes
                    # (Comma splits angle bracket blocks)
                    Sequence(
                        OneOf(
                            Ref("GeneratedColumnDefinitionSegment"),
                        ),
                        Ref("CommentGrammar", optional=True),
                    ),
```

src/sqlfluff/dialects/dialect_teradata.py
```diff
@@ -450,9 +450,9 @@ class TdCreateTableOptions(BaseSegment):
     """
 
     type = "create_table_options_statement"
-    match_grammar = AnyNumberOf(
-        Sequence(
-            Ref("CommaSegment"),
+    match_grammar = Sequence(
+        Ref("CommaSegment"),
+        Delimited(
             OneOf(
```
WittierDinosaur, line 456: Is this OneOf() now unnecessary?

tunetheweb: Technically yes it is not needed as `Delimited` inherits from `OneOf`.

However, I always forget whether it inherits from `Sequence` or `OneOf` (I had to look it up again) so usually explicitly add the type I want underneath the `Delimited`. A bit of a waste, but more clear IMHO.

WDYT? Think I should just learn this properly and remove this unnecessary `OneOf`?

tunetheweb: Technically yes it is not needed as `Delimited` inherits from `OneOf`.

However, I always forget whether it inherits from `Sequence` or `OneOf` (I had to look it up again) so usually explicitly add the type I want underneath the `Delimited`. A bit of a waste, but more clear IMHO.

WDYT? Think I should just learn this properly and remove this unnecessary `OneOf`?

tunetheweb:
> There is one more instance of this in sparksql (line 849):

I thought that one was being taken care of in your #3023 PR and was trying to avoid a merge conflict with that one but see that one was slightly different. Have added it now.

src/sqlfluff/dialects/dialect_teradata.py
tunetheweb: Looks like I'm not the only one! There's 45 instances of `Delimited\(\n\s*OneOf` in our dialects.

As I say, I think it's clearer so would prefer to leave it. But can remove them all if you think it's not a good idea?

barrywhart (approved):

src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -1402,31 +1402,18 @@ class ProcedureParameterListSegment(BaseSegment):
...
-                        OneOf(
-                            Ref("ParameterNameSegment"),
-                            exclude=OneOf(_param_type, Ref("ArgModeGrammar")),
-                            optional=True,
-                        ),
-                        Ref("ArgModeGrammar", optional=True),
-                        max_times_per_element=1,
+        Delimited(
+            Sequence(
+                AnyNumberOf(
```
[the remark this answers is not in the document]
tunetheweb: In this case there is a `_param_type` grammar shortcut as well as the `AnyNumberOf` so it really is a sequence.

PR #3030, opened 2022-04-06
Fix bug in patch generation for segments made of templated + literal fixes

### LINKED ISSUE #3022: L030: Auto-fix can produce broken SQL
### Search before asking

### What Happened

This is a very weird case, but running auto fix for L030 in this scenario produces broken SQL.

```
SELECT
    ROW_NUMBER(),
    `{{ params.project_id }}`.some_dataset._f_some_function(TRUE),
    `{{ params.project_id }}.some_dataset._f_some_function`(TRUE)
FROM `some_table`
;
```

### Expected Behaviour

I'm not even sure what I did expect as the output of L030. Probably this

```sql
SELECT
    ROW_NUMBER(),
    `{{ params.project_id }}`.some_dataset._F_SOME_FUNCTION(TRUE),
    `{{ params.project_id }}.some_dataset._f_some_function`(TRUE)
FROM `some_table`
;
```

### Observed Behaviour

```sql
SELECT
    ROW_NUMBER(),
    `{{ params.project_id }}`.some_dataset._F_SOME_FUNCTION_f_some_function(TRUE),
    `{{ params.project_id }}.some_dataset._f_some_function`(TRUE)
FROM `some_table`
;
```

Note the `_F_SOME_FUNCTION_f_some_function` part.

### How to reproduce

`sqlfluff fix --rules L030` on the example above

### Dialect

bigquery (not sure if it's related to bigquery)

### Version

Latest `main`

### Configuration

[sqlfluff]
templater = jinja
dialect = bigquery

params = {
    "project_id": "test_project",
    }

### Code of Conduct

barrywhart:
Fixes #3022 

I think I introduced this bug when fixing an earlier issue adding missing newlines to the end of a file (PR #2862). In that PR, I updated `BaseSegment.iter_patches()` to return an `EnrichedFixPatch`, which specifies the location of the change in both source and templated space. The bug was, if the modified segment is a mix of templated and literal code, the source slice was incorrect. which caused code **replacement** (delete + insert) to be misapplied as just an insertion, i.e. leaving the old, bad code in place, followed by the new, corrected code.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -1268,6 +1268,7 @@ class BaseSegment:
             # This segment isn't a literal, but has changed, we need to go deeper.
 
             # Iterate through the child segments
+            source_idx = self.pos_marker.source_slice.start
             templated_idx = self.pos_marker.templated_slice.start
             insert_buff = ""
             for seg_idx, segment in enumerate(self.segments):
@@ -1311,15 +1312,19 @@ class BaseSegment:
 
                 # Once we've dealt with any patches from the segment, update
                 # our position markers.
+                source_idx = segment.pos_marker.source_slice.stop
                 templated_idx = segment.pos_marker.templated_slice.stop
 
             # After the loop, we check whether there's a trailing deletion
             # or insert. Also valid if we still have an insertion buffer here.
             end_diff = self.pos_marker.templated_slice.stop - templated_idx
             if end_diff or insert_buff:
-                source_slice = segment.pos_marker.source_slice
+                source_slice = slice(
+                    source_idx,
+                    self.pos_marker.source_slice.stop,
+                )
                 templated_slice = slice(
-                    self.pos_marker.templated_slice.stop - end_diff,
+                    templated_idx,
                     self.pos_marker.templated_slice.stop,
                 )
                 # By returning an EnrichedFixPatch (rather than FixPatch), which
--- a/test/fixtures/rules/std_rule_cases/L030.yml
+++ b/test/fixtures/rules/std_rule_cases/L030.yml
@@ -60,3 +60,8 @@ test_pass_ignore_templated_code:
     SELECT
         {{ "greatest(a, b)" }},
         GREATEST(i, j)
+
+test_fail_func_name_templated_literal_mix:
+  # Issue 3022. This was actually a bug in BaseSegment.iter_patches().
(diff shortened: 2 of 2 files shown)


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1268,6 +1268,7 @@ def iter_patches(
             # This segment isn't a literal, but has changed, we need to go deeper.
 
             # Iterate through the child segments
+            source_idx = self.pos_marker.source_slice.start
```
barrywhart, line 1271: Note that `source_idx` is computed the same way as `templated_idx`, but using `source_slice` rather than `templated_slice`.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1311,15 +1312,19 @@ def iter_patches(
 
                 # Once we've dealt with any patches from the segment, update
                 # our position markers.
+                source_idx = segment.pos_marker.source_slice.stop
```
barrywhart, line 1315: Ditto

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1311,15 +1312,19 @@ def iter_patches(
...
             end_diff = self.pos_marker.templated_slice.stop - templated_idx
             if end_diff or insert_buff:
-                source_slice = segment.pos_marker.source_slice
+                source_slice = slice(
+                    source_idx,
+                    self.pos_marker.source_slice.stop,
+                )
                 templated_slice = slice(
-                    self.pos_marker.templated_slice.stop - end_diff,
+                    templated_idx,
```
barrywhart, line 1327: These are equivalent -- let's go back to high-school algebra. 🙂 

Above, we compute:
```
end_diff = self.pos_marker.templated_slice.stop - templated_idx
```

 So this:
```
self.pos_marker.templated_slice.stop - end_diff
```

is the same as this (replace `end_diff` with the formula):
```
self.pos_marker.templated_slice.stop - (self.pos_marker.templated_slice.stop - templated_idx)
```

Which is just:
```
templated_idx
```

Why did I bother with this?

1. It reads a little simpler.
2. It matches up with how we compute `source_slice` above without having to add a new `diff` variable, which would've made us want to rename `end_diff` as `templated_end_diff`, etc. etc.

tunetheweb (approved):
LGTM

PR #2982, opened 2022-04-03
Added support for MySQL Update Statements

WittierDinosaur:
### Brief summary of the change made
Adds support for MySQL's Update Statements.

Fixes #2832

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

Issue #2832, opened 2022-03-10
Parse MySQL multi-table UPDATE


jasonaowen:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

SQLFluff is unable to parse [MySQL multiple table `UPDATE` statements](https://dev.mysql.com/doc/refman/5.7/en/update.html).

### Expected Behaviour

SQLFluff to parse the valid, executable SQL and report on any rule violations.

### Observed Behaviour

> Line 1, Position 1: Found unparsable section: 'UPDATE a, b SET a.name = b.name WHERE a....

### How to reproduce

With `multi-update.sql` as
```sql
UPDATE a, b SET a.name = b.name WHERE a.id = b.id;
```

running SQLFluff results in a parse error:

```sh
$ sqlfluff parse --verbose --ignore-local-config --dialect mysql multi-update.sql 
==== sqlfluff ====
sqlfluff:               0.11.0 python:                 3.8.10
implementation:        cpython dialect:                 mysql
verbosity:                   1 templater:               jinja

=== [ path: multi-update.sql ] ===

[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    statement:
[L:  1, P:  1]      |        update_statement:
[L:  1, P:  1]      |            unparsable:                                       !! Expected: 'UpdateStatementSegment'
[L:  1, P:  1]      |                keyword:                                      'UPDATE'
[L:  1, P:  7]      |                whitespace:                                   ' '
[L:  1, P:  8]      |                raw:                                          'a'
[L:  1, P:  9]      |                comma:                                        ','
[L:  1, P: 10]      |                whitespace:                                   ' '
[L:  1, P: 11]      |                raw:                                          'b'
[L:  1, P: 12]      |                whitespace:                                   ' '
[L:  1, P: 13]      |                raw:                                          'SET'
[L:  1, P: 16]      |                whitespace:                                   ' '
[L:  1, P: 17]      |                raw:                                          'a'
[L:  1, P: 18]      |                raw:                                          '.'
[L:  1, P: 19]      |                raw:                                          'name'
[L:  1, P: 23]      |                whitespace:                                   ' '
[L:  1, P: 24]      |                raw:                                          '='
[L:  1, P: 25]      |                whitespace:                                   ' '
[L:  1, P: 26]      |                raw:                                          'b'
[L:  1, P: 27]      |                raw:                                          '.'
[...]

PR #3031, opened 2022-04-07
MySQL: Allow `JOIN`s in `UPDATE` expressions

zapion:
in UpdateStatementSegment
followup to #2982 and #2832, add test case in update.sql as well
try to tackle with query like
```update a join b on a.id = b.id set a.type = b.type where a.type is null```


tunetheweb (approved):
LGTM!

Issue #3011, opened 2022-04-05
BigQuery: Trailing comma in columns of `CREATE TABLE` statement is unparsable


dmohns:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

A BigQuery `CREATE TABLE` statement that has a trailing comma after the last column yields unparsable error. However, it perfectly valid BigQuery SQL.

### Expected Behaviour

SQLFluff can parse this BigQuery specific.

### Observed Behaviour

```
Line 1, Position 1: Found unparsable section: 'CREATE TABLE (
    col1 STRING,
)'
```

### How to reproduce

```sql
CREATE TABLE (
    col1 STRING,
)
```

https://online.sqlfluff.com/fluffed?sql=H4sIAMxkTGIC_3MOcnUMcVUIcXTycVXQ4OVSAILk_BxDheCQIE8_dx1eLk0A3ISaliMAAAA%3D&dialect=bigquery

### Dialect

BigQuery

### Version

0.11.2 -> webversion

### Configuration

default

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

tunetheweb:
This is easily fixed by adding `allow_trailing=True` in the CreateTableStatementSegment:

```py
        # Column list
        Sequence(
            Bracketed(
                Delimited(
                    OneOf(
                        Ref("TableConstraintSegment"),
                        Ref("ColumnDefinitionSegment"),
                    ),
                    allow_trailing=True,
                )
            ),
```

I agree we should enhance L038 to also cover this. Maybe with a separate config option (`create_table_colons_trailing_comma`?).

PR #3039, opened 2022-04-07
L038: Add config to enable linting for trailing commas in `CREATE TABLE`

dmohns:

Follow up to #3011

Trying to make L038 also cover `CREATE TABLE` statements. Achieved by adding an abstraction level to `CREATE TABLE` statements. Not sure, how much I like it, as it seems to be very specific to L038. Open for comments.

If we want to go that route, we probably should check `CreateTableStatementSegment` of other dialects. For demonstration purposes I only added to `bigquery` dialect.

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -118,6 +118,7 @@ force_enable = False
 [sqlfluff:rules:L038]
 # Trailing commas
 select_clause_trailing_comma = forbid
+create_table_trailing_comma = forbid
 
 [sqlfluff:rules:L040]
 # Null & Boolean Literals
--- a/src/sqlfluff/core/rules/config_info.py
+++ b/src/sqlfluff/core/rules/config_info.py
@@ -74,6 +74,13 @@ STANDARD_CONFIG_INFO_DICT = {
             "Should trailing commas within select clauses be required or forbidden?"
         ),
     },
+    "create_table_trailing_comma": {
+        "validation": ["forbid", "require"],
+        "definition": (
+            "Should trailing commas within create table clauses be required or "
+            "forbidden?"
+        ),
+    },
     "ignore_comment_lines": {
         "validation": [True, False],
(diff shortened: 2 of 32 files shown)


tunetheweb:
Hmmm... it seems the cleanest way of adding this, but is a LOT of change, for basically only one dialect that supports this.

But the alternative or tracing down the tree in L038.

We could just change the rule to look for any comma followed by a closing bracket? Maybe within a `CreateTableStatementSegment`? I can't think of any reason that would allow this except trailing commas.

@sqlfluff/sqlfluff-maintainers any other ideas here?

barrywhart:
Can we try and think about other places this rule could also apply and how we could design this in a more extensible way? If there are two places it could apply, there are probably more that we haven't thought of yet. E.g. `ALTER TABLE`, `MERGE TABLE`, `CREATE VIEW`, `CREATE INDEX`...

dmohns:
I was testing around with this a bit. A trailing comma does not seem to be valid anywhere in BigQuery other than `SELECT` and `CREATE TABLE` COLUMNS. However, the fact that the latter is valid doesn't seem to be documented anywhere. 

Maybe it's not a good idea to add this.

If you agree, feel free to close this PR.

tunetheweb:
It's documented here for SELECT only: https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#trailing_commas

Interestingly the `CREATE TABLE` definition doesn't seem to support it, and makes no mention of it: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_table_statement

So I think close this for now. If we do reopen it later, then might want to revisit the option to look for comma + closing bracket as a simpler fix?

I'm OK with leaving it in the SQLFluff grammar for now, so it doesn't cause a parsing error in `CREATE TABLE` statements since you say it works. But let's not bother adding a rule to catch this unless BigQuery confirms this one way or another.

Issue #3037, opened 2022-04-07
Linter - Some rules deal with file structure, and have no need to step through the entire tree


WittierDinosaur:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

Any highly specific rules should take trivial amounts of time to run, such as ones that address the end of the file

### Use case

-

### Dialect

-

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

barrywhart:
L009 and L050 look like good initial candidates for this. Are you aware of any others? (I could figure this out from the code, but there are lots of rules to study!)

@WittierDinosaur @tunetheweb @alanmcruickshank

Issue #3035, opened 2022-04-07
Linter - Turn off looping behaviour for rules that don't need it


WittierDinosaur:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

Things like capitalisation policy, which make use of the parse tree, should never need to loop, and the parsed element types shouldn't be changed by the linter. Turning off looping behaviour would be massively benefitial for performance

### Use case

-

### Dialect

-

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

PR #3041, opened 2022-04-07
Update individual rules to take advantage of core rule processing changes

### LINKED ISSUE #3034: Speed up Rules critical errors tests
### Search before asking

### Description

With the recent performance improvement, the CI jobs are taking a lot less time. Except for the "Rules critical errors tests" one which takes 20mins or so.

Could we speed it up? This runs a lint against ALL our dialect fixtures, so is going to take some time, and is a great test but still, it's slowness is annoying and will only grow as we increase out dialect coverage and add new dialects.

I don't think parallel processing will help as it already uses 2 processors, and GitHub Action machines are currently limited to two-core machines anyway.

Could we split it into multiple jobs that can run in parallel? I wouldn't want to hard code the dialects, but maybe have dialects beginning a-m in one job, and n-z in another? Or have ansi (our biggest job) in one job, and the rest in another?

### Use case

PRs not to be held up on a long running job

### Dialect

N/A

### Code of Conduct

barrywhart:
- [x] Update L009 and L050 to only be called once (parse tree root)
- [x] Update rules 9, 10, 14, 30, 40, 50, 63 to run in the `post` linter phase.
- [x] Update docs for rule authors to cover the new iteration options

Fixes #3034, #3035, #3037

### Pull Request checklist

  - Other.


src/sqlfluff/core/linter/linter.py
```diff
@@ -506,6 +507,10 @@ def lint_fix_parsed(
             )
 
             for crawler in progress_bar_crawler:
+                # Performance: After first loop pass, skip rules that don't do fixes.
+                if fix and loop > 0 and not is_fix_compatible(crawler):
+                    continue
```
barrywhart, line 512: About 1/4 of the rules aren't fix compatible. But we were executing them on every linter pass. Wasteful!

src/sqlfluff/core/rules/base.py
```diff
@@ -416,6 +416,48 @@ def raw_segments(self):
         )
 
 
+class CrawlBehavior:
```
barrywhart, line 419: Extracted segment traversal code from `BaseRule.crawl()` into a separate class. This makes it easier to use different iteration strategies.

tunetheweb:
Nice improvement on "Rules critical errors tests" already! Maybe this fixes #3034 sufficiently that we can close that?

barrywhart:
@tunetheweb:
>Nice improvement on "Rules critical errors tests" already! Maybe this fixes https://github.com/sqlfluff/sqlfluff/issues/3034 sufficiently that we can close that?

Oh, nice! I hadn't noticed. I updated the PR description to say "fixes" that issue. 👍

src/sqlfluff/core/linter/linter.py
```diff
@@ -494,112 +495,129 @@ def lint_fix_parsed(
...
-                # "anchor", the segment to look for either to edit or to insert BEFORE.
-                # The second is the element to insert or create.
-                linting_errors, _, fixes, _ = crawler.crawl(
-                    tree,
-                    ignore_mask=ignore_buff,
-                    dialect=config.get("dialect_obj"),
-                    fname=fname,
-                    fix=fix,
-                    templated_file=templated_file,
+        for phase in ["main", "post"]:
```
barrywhart, line 498: Use "Hide whitespace" to review the changes in this section, since it adds a new level of indentation.

src/sqlfluff/core/rules/base.py
```diff
@@ -372,7 +391,7 @@ class FunctionalRuleContext:
     def __init__(self, context: RuleContext):
         self.context = context
 
-    @cached_property
```
barrywhart, line 375: Removed caching of properties now that we are reusing `RuleContext` objects

src/sqlfluff/rules/L003.py
```diff
@@ -228,15 +232,17 @@ def _process_raw_stack(
             starting_indent_balance = result_buffer[cached_line_count].indent_balance
 
         working_state = _LineSummary(indent_balance=starting_indent_balance)
-        line_no = 1
-        started = line_no == (cached_line_count + 1)
-        for elem in raw_stack:
+
+        line_no = memory.line_no
+        target_line_no = cached_line_count + 1
+        for idx, elem in enumerate(raw_stack[memory.start_process_raw_idx :]):
```
barrywhart, line 238: Some unrelated, specific tuning of L003 here. Previously, it was starting from the beginning of `raw_stack` every time, which caused O(n^2) performance. Now, it remembers the last line number and position in `raw_stack`, so it can resume scanning on the "current" line. This shows up as a lot fewer executions of these lines in the Python line profiler and a big reduction in "percent of overall function time".

barrywhart:
This is ready for review, other than me needing to add docs.

src/sqlfluff/core/rules/base.py
```diff
@@ -349,18 +349,56 @@ def _raw_slices_from_templated_slices(
...
     dialect: Dialect
-    path: Optional[pathlib.Path]
     fix: bool
     templated_file: Optional[TemplatedFile]
+    path: Optional[pathlib.Path]
+
+    # These change within a file.
+    segment: BaseSegment
+    parent_stack: Tuple[BaseSegment, ...] = field(default=tuple())
+    raw_segment_pre: Optional[RawSegment] = field(default=None)
```
barrywhart, line 361: New field, often can be used as a more efficient alternative to `raw_stack`

src/sqlfluff/core/rules/base.py
```diff
@@ -349,18 +349,56 @@ def _raw_slices_from_templated_slices(
...
+    raw_stack: Tuple[RawSegment, ...] = field(default=tuple())
+    memory: Any = field(default_factory=dict)
+    segment_idx: int = field(default=0)
+
+    @property
```
barrywhart, line 366: Changed `siblings_pre` and `siblings_post` to compute on demand because they are very rarely used.

src/sqlfluff/core/rules/base.py
```diff
@@ -349,18 +349,56 @@ def _raw_slices_from_templated_slices(
...
+    def siblings_post(self) -> Tuple[BaseSegment, ...]:
+        """Return sibling segments after self.segment."""
+        if self.parent_stack:
+            return self.parent_stack[-1].segments[self.segment_idx + 1 :]
+        else:
+            return tuple()
 
     @cached_property
+    def final_segment(self) -> BaseSegment:
```
barrywhart, line 383: New property

src/sqlfluff/core/rules/base.py
```diff
@@ -484,149 +586,107 @@ def _eval(self, context: RuleContext) -> EvalResultType:
 
     def crawl(
```
barrywhart, line 587: This function is no longer recursive -- the recursion has moved to the `CrawlBehavior` helper class.

src/sqlfluff/core/rules/base.py
```diff
@@ -484,149 +586,107 @@ def _eval(self, context: RuleContext) -> EvalResultType:
...
         self,
-        segment,
-        ignore_mask,
-        dialect,
-        parent_stack=None,
```
barrywhart, line 490: Several of these parameters were already stored on `RuleContext`. Eliminated these as individual parameters; now they are stored and passed using `RuleContext`. This was basically old tech debt from before `RuleContext` was added.

src/sqlfluff/core/rules/base.py
```diff
@@ -484,149 +586,107 @@ def _eval(self, context: RuleContext) -> EvalResultType:
...
-                fix=fix,
-                templated_file=templated_file,
-            )
-            vs += dvs
-            fixes += child_fixes
-        return vs, raw_stack, fixes, memory
+            for lerr in new_lerrs:
+                self.logger.debug("!! Violation Found: %r", lerr.description)
+            for lfix in new_fixes:
+                self.logger.debug("!! Fix Proposed: %r", lfix)
```
barrywhart, line 684: Renamed `fix` to `lfix` to avoid type error with the differently typed `fix` parameter.

src/sqlfluff/rules/L003.py
```diff
@@ -408,7 +414,7 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
                 # First non-whitespace element is our trigger
                 memory.trigger = segment
 
-        is_last = self.is_final_segment(context)
+        is_last = context.segment is context.final_segment
```
barrywhart, line 417: `is_final_segment()` is very complex and slow. Switching to the similar, much faster `RuleContext.final_segment`.

src/sqlfluff/rules/L004.py
```diff
@@ -77,8 +77,8 @@ def _eval(self, context: RuleContext) -> LintResult:
                 )
                 # Only attempt a fix at the start of a newline for now
                 and (
-                    len(context.raw_stack) == 0
-                    or context.raw_stack[-1].is_type("newline")
+                    context.raw_segment_pre is None
+                    or context.raw_segment_pre.is_type("newline")
```
barrywhart, line 81: Switching from `raw_stack` to `raw_segment_pre` so this rule doesn't require  the heavyweight`raw_stack` anymore.

src/sqlfluff/rules/L005.py
```diff
@@ -43,8 +44,8 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
         We need at least one segment behind us for this to work.
 
         """
-        if len(context.raw_stack) >= 1:
-            cm1 = context.raw_stack[-1]
+        if context.raw_segment_pre is not None:
```
barrywhart, line 47: Switching from raw_stack to `raw_segment_pre` so this rule doesn't require the heavyweight `raw_stack` anymore.

src/sqlfluff/rules/L009.py
```diff
@@ -1,13 +1,36 @@
...
-from sqlfluff.core.parser import NewlineSegment
+from typing import List, Optional, Tuple
 
+from sqlfluff.core.parser import BaseSegment, NewlineSegment
 from sqlfluff.core.rules.base import BaseRule, LintResult, LintFix, RuleContext
 from sqlfluff.core.rules.doc_decorators import document_fix_compatible
 from sqlfluff.core.rules.functional import Segments, sp, tsp
 
 
+def get_trailing_newlines(segment: BaseSegment) -> List[BaseSegment]:
```
barrywhart, line 10: Reworked this rule so it is only called once, with the root segment.

src/sqlfluff/rules/L011.py
```diff
@@ -68,14 +68,14 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
 
                             # Remove the AS as we're using implict aliasing
                             fixes.append(LintFix.delete(context.segment.segments[0]))
-                            anchor = context.raw_stack[-1]
```
barrywhart, line 71: Switching from `raw_stack` to `raw_segment_pre` so this rule doesn't require the heavyweight `raw_stack` anymore.

src/sqlfluff/rules/L050.py
```diff
@@ -66,46 +67,45 @@ class Rule_L050(BaseRule):
     """
 
     targets_templated = True
+    needs_raw_stack = False
+    # TRICKY: Tells linter to only call _eval() ONCE, with the root segment
+    recurse_into = False
+    lint_phase = "post"
```
barrywhart, line 73: Reworked this rule so it is only called once per parse tree, with the root segment.

test/cli/commands_test.py
```diff
@@ -678,33 +678,24 @@ def test__cli__fix_error_handling_behavior(sql, fix_args, fixed, exit_code, tmpd
             assert not fixed_path.is_file()
 
 
-_old_crawl = BaseRule.crawl
```
barrywhart, line 681: Had to update this test now that `crawl()` is no longer called recursively. It now mocks the `_eval()` function instead.

test/core/rules/config_test.py
```diff
@@ -21,8 +24,14 @@ def _eval(self, context):
         pass
 
 
+@document_fix_compatible
 class Rule_T001(BaseRule):
-    """A deliberately malicious rule."""
+    """A deliberately malicious rule.
+
+    **Anti-pattern**
```
barrywhart, line 31: Had to add this verbiage to the docstring so `@ document_fix_compatible` would work.

test/core/rules/config_test.py
```diff
@@ -21,8 +24,14 @@ def _eval(self, context):
         pass
 
 
+@document_fix_compatible
```
barrywhart, line 27: This rule is intended to cause an endless loop of "fixes". Had to add this decorator so the new, smarter lint loop would actually call it every loop. (Now on pass 2 and beyond, it skips calling rules that don't declare they return fixes.)

src/sqlfluff/core/rules/base.py
```diff
@@ -349,18 +349,54 @@ def _raw_slices_from_templated_slices(
...
+    # These change within a file.
+    segment: BaseSegment
+    parent_stack: Tuple[BaseSegment, ...] = field(default=tuple())
+    raw_segment_pre: Optional[RawSegment] = field(default=None)
+    raw_stack: Tuple[RawSegment, ...] = field(default=tuple())
+    memory: Any = field(default_factory=dict)
+    segment_idx: int = field(default=0)
+
+    @property
+    def siblings_pre(self) -> Tuple[BaseSegment, ...]:  # pragma: no cover
```
tunetheweb, line 367: Why no cover?

tunetheweb (commented):
There is a LOT going on here. It looks good overall, but it's almost impossible to review properly.

Would you consider pulling them into separate, smaller PRs to make it easier to review and merge?

src/sqlfluff/core/rules/base.py
barrywhart: No rules were actually using this property. When it was just a regular field, it didn't matter, but now that it's a function that's called "on demand", it's not called unless something uses it.

barrywhart:
@tunetheweb: I could easily split the L003 changes into a separate PR. Would that be enough splitting, or what would you like to see?

The core changes (`linter.py` and `base.py`) would be too hard to split, so I think they need to stay together.

The rule 9 and 50 changes could be split out. They would need to come _after_ the core one, though.

Another possibility than splitting the PR -- could we do a 30-minute or 1-hour video meeting to walk through the PR? If we can do this, that might achieve the same goal more efficiently.

src/sqlfluff/core/rules/base.py
barrywhart: I could've simply removed it, but since it's symmetrical with `siblings_post`, which _is_ used (although by only one rule itself), I thought it better to keep it.

tunetheweb:
Let’s pull out ones that are easy to pull out. Can leave this PR open, and hopefully changes will get smaller and smaller as the other PRs are merged, and then hopefully can merge this.

I did look through it all, and could approve as didn’t see anything majorly wrong, but think it could be easier if we did above if some changes (like L003 as you suggest) are reasonably independent if that’s not to much additional effort.

Will also be easy for later if there is a bug and can see them as separate commits to `main`.

src/sqlfluff/rules/L003.py
OTooleMichael: Legend

src/sqlfluff/rules/L005.py
OTooleMichael: Switch to a guard here right ?
```suggestion
        if context.raw_segment_pre is None:
            return
```

OTooleMichael (approved):
LGTM

OTooleMichael:
> 

Im happy enough with everything going on here (especially having read your first split out first). It seems pretty "straightforward". (I could see some places to go further with the code splitting / Rule Flags) :) but all the change look good

src/sqlfluff/rules/L005.py
barrywhart: I would still need a second `return None` because of the nested `if`, so I wonder if this would really be an improvement. I think I'll combine the two `if`s instead, maybe?

src/sqlfluff/rules/L005.py
```diff
@@ -38,19 +39,15 @@ class Rule_L005(BaseRule):
...
-            cm1 = context.raw_stack[-1]
-            if (
-                context.segment.is_type("comma")
-                and cm1.is_type("whitespace")
-                and cm1.pos_marker.line_pos > 1
-            ):
-                anchor = cm1
-                return LintResult(anchor=anchor, fixes=[LintFix.delete(cm1)])
-        # Otherwise fine
+        """Commas should not have whitespace directly before them."""
```
barrywhart, line 42: The primary change to this file is switching from `raw_stack` to `raw_segment_pre`, which is faster. Also did some tidying, but no other meaningful changes. @OTooleMichael

src/sqlfluff/core/rules/base.py
```diff
@@ -503,12 +520,12 @@ class BaseRule:
     # Lint loop / crawl behavior. When appropriate, rules can (and should)
     # override these values to make linting faster.
     recurse_into = True
-    needs_raw_stack = True
+    needs_raw_stack = False  # False is faster & most rules don't need it
```
tunetheweb, line 523: Can we give more commentary as towhy a rule would need or not need `raw_stack`? I presume it's if you are looking at a segment's parents to find if it's nested in a particular statement? Remind me again what the difference is. between `raw_stack` and `raw_segment_pre`

src/sqlfluff/rules/L009.py
```diff
@@ -1,13 +1,35 @@
...
+def get_last_segment(segment: Segments) -> Tuple[List[BaseSegment], Segments]:
+    """Returns rightmost & lowest descendant and its "parent stack"."""
+    parent_stack: List[BaseSegment] = []
+    while True:
+        children = segment.children()
+        if children:
+            parent_stack.append(segment[0])
+            segment = children.last()
+        else:
+            return parent_stack, segment
```
tunetheweb, line 30: Why doesn't it use the new `final_segment` function?

src/sqlfluff/core/rules/base.py
barrywhart: I'll add more comments`.  Briefly,`raw_segment_pre` is the **one** raw segment prior to the current one, while `raw_stack` is a list of **every** raw segment prior. The former is pretty lightweight to compute, while the latter is pretty heavyweight, as we're creating a new (and increasingly lengthy) tuple as we proceed through the file.

tunetheweb: Oh I didn’t get that.

Is that just the parent Segment? Would that be a better name for it rather than `raw_stack_prev`?

src/sqlfluff/core/rules/base.py
barrywhart: Updated to use ternary operator, no more "no cover", plus it's used now in this PR.

src/sqlfluff/core/rules/base.py
barrywhart: No, it's not the parent. Raw segments are (IIRC) those at the leaf of the tree, i.e. no children. Generally, these are a single keyword, operator, literal, or a meta segment like indent or dedent, like the "skin" of the tree as opposed to the "bones", I guess?

src/sqlfluff/rules/L009.py
barrywhart: Because it also needs the parent stack, which is not provided by `RuleContext.final_segment`. I could've had `RuleContext` return both, but that'd be harder to document clearly, and AFAIK, this is the only rule that needs it.

barrywhart:
@tunetheweb: Ok, ready for final review. That is, I added the new docs to `developingrules.rst`. There's probably some duplication between that file and some of the code comments. Might be worth moving things around a bit? It always feels tricky to get this right.

src/sqlfluff/core/rules/base.py
OTooleMichael: `previous_raw_segement` seems better as a name or `raw_segment_previous`.
at least `prev`

docs/source/developingrules.rst
```diff
@@ -3,11 +3,70 @@
...
+but is only run 2 times (once to fix, and once again to confirm no remaining
+issues).
+
+The two phases add complexity, but they also improve performance by allowing
+SQLFluff to run fewer rules during the ``main`` phase, which often runs several
+times.
+
+NOTE: ``post`` rules also run on the *first* pass of the ``main`` phase so that
+any issues they find will be presented in the list of issues output by
+``sqlfluff fix`` and ``sqlfluff lint``.
```
tunetheweb, line 69:
```suggestion
2. The ``post`` phase is for post-processing rules, not expected to trigger
any downstream rules, e.g. capitalization fixes. They are run in a
post-processing loop at the end. This loop is identical to the ``main`` loop,
but is only run 2 times at the end (once to fix, and once again to confirm no
remaining issues).

The two phases add complexity, but they also improve performance by allowing
SQLFluff to run fewer rules during the ``main`` phase, which often runs several
times.

NOTE: ``post`` rules also run on the *first* pass of the ``main`` phase so that
any issues they find will be presented in the list of issues output by
``sqlfluff fix`` and ``sqlfluff lint``.
```

docs/source/developingrules.rst
```diff
@@ -3,11 +3,70 @@
...
+large files. For example, see rules ``L050`` and ``L009`` , which only look at
+the beginning or end of the file, respectively.
+
+``_works_on_unparsable``
+^^^^^^^^^^^^^^^^^^^^^^^^
+By default, `SQLFluff` calls ``_eval()`` for all segments, even "unparsable"
+segments, i.e. segments that didn't match the parsing rules in the dialect.
+This causes issues for some rules. If so, setting ``_works_on_unparsable``
+to ``False`` tells SQLFluff not to call ``_eval()`` for unparsable segments and
+their descendants.
```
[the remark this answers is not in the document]
barrywhart: That's what I thought, too, but I checked. I think maybe I (we) were confusing this with the `--fix-even-unparsable` thing added recently. We've got so many settings. 🤯 

IIUC, the setting documented here controls whether the rule is called for unparsable segments, and `--fix-even-unparsable` controls whether any fixes would be applied. (Note that you can lint without fixing!)

src/sqlfluff/core/rules/base.py
```diff
@@ -503,12 +520,27 @@ class BaseRule:
     # Lint loop / crawl behavior. When appropriate, rules can (and should)
     # override these values to make linting faster.
     recurse_into = True
-    needs_raw_stack = True
+    # False is faster & most rules don't need it. Rules that use it are usually
```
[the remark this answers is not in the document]
barrywhart: Will update to mention the property name. 👍

src/sqlfluff/rules/L003.py
barrywhart: Maybe rename the function? The property (`RuleContext.final_segment`) is the actual, final segment. The function supports optional filtering of `meta` segments. Currently, only one rule (L052) uses the function, so we could potentially move it into L052.py. Feel free to create an issue or PR for this. I'll probably go ahead and merge, since this doesn't seem like a blocker.

PR #3043, opened 2022-04-07
Postgres add `ALTER ROLE`/`ALTER USER` support

mrf:
### Brief summary of the change made
Creates new class that provides support for both ALTER ROLE and ALTER USER statements following spec in postgres docs.

### Are there any other side effects of this change that we should be aware of?
New functionality so shouldn't affect existing behavior.

### Pull Request checklist
New supporting test cases and fixtures added

fixes #3042

PR #3045, opened 2022-04-08
Exasol: `CREATE/DROP/ALTER USER/ROLE` clean up for consistency

tunetheweb:

Noticed while reviewing #3043 that these statements were inconsistent with other dialects.

Shouldn't be

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_exasol.py
+++ b/src/sqlfluff/dialects/dialect_exasol.py
@@ -2332,13 +2332,13 @@ class FileOptionSegment(BaseSegment):
 ############################
 # USER
 ############################
-class CreateUserSegment(BaseSegment):
+class CreateUserStatementSegment(BaseSegment):
     """`CREATE USER` statement.
 
     https://docs.exasol.com/sql/create_user.htm
     """
 
-    type = "create_user"
+    type = "create_user_statement"
 
     is_ddl = False
     is_dml = False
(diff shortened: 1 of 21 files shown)


src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -2332,13 +2332,13 @@ class FileOptionSegment(BaseSegment):
 ############################
 # USER
 ############################
-class CreateUserSegment(BaseSegment):
+class CreateUserStatementSegment(BaseSegment):
```
barrywhart, line 2335: Is it worth using inheritance for some of these segment types? Then you'd be inheriting the `type` property, as well as all the `is_...` properties.

barrywhart (approved):

src/sqlfluff/dialects/dialect_exasol.py
tunetheweb: `CreateUser` doesn't exist in ANSI (maybe it should) but `CreateRole` does.

BTW this was one advantage if the `replace=True` decorator - was more obviously which ones were overriding and which weren't.

tunetheweb: We have a `Drop User` in ANSI so I added a `CREATE USER` too. Which allows this to be inherited now.

barrywhart: Re: `replace`, what if we provided a `@new_segment` decorator that you **have** to use on new segments, and anything else is assumed to be (and automatically verified to be) a replacement?

tunetheweb: Not sure it's worth the effort. It's not a big deal. Was just a little handy to know that.

PR #3058, opened 2022-04-09
Add editorconfig and precommit for SQL and YML files

tunetheweb:

When we first put in editorconfg and precommit we excldued .sql and .yml files as some of those have trailing whitespace. This PR moves those few cases to exception lists, so the rest are checked as they very much are the minority. And I'm bored of asking people to add trailing newlines :-)

Future, similar exceptions will need to be managed similarly.

### Pull Request checklist

  - Other.


.editorconfig
```diff
@@ -14,9 +14,16 @@ indent_size = 2
 [*.py]
 indent_size = 4
 
-# Don't correct spacing for sql and yaml files as sometimes want them wrong for tests
-[*.sql]
```
tunetheweb, line 18: Review this entire PR with whitespace changes hidden:

image

.editorconfig
```diff
@@ -14,10 +14,34 @@ indent_size = 2
...
+# Don't correct indentation for sql and yaml files as sometimes want them wrong for tests
+[*.{yml,yaml,sql}]
+indent_style = unset
+
+# Some specific tests with trailing newlines
+# If adding any exceptions here, make sure to add them to .pre-commit-config.yaml as well
+[test/fixtures/linter/sqlfluffignore/*/*.sql]
+indent_style = unset
+trim_trailing_whitespace = unset
```
WittierDinosaur, line 25: I wonder if this should be unset more broadly? With some tests adding trailing whitespace is useful

tunetheweb: My personal preference would be to have trailing white space tests in Python strings (or quote YAML like in https://github.com/sqlfluff/sqlfluff/blob/main/test/fixtures/rules/std_rule_cases/L001.yml). Having them as actual trailing whitespace is not great as they are invisible to the human eye.

tunetheweb: But willing to let these few exceptions pass for now btw. And maybe add some more in future if we really have to. But would prefer them to very much be the exception.

WDYT?

WittierDinosaur: Happy with that - it was just a thought 👍

WittierDinosaur (approved):
LGTM

PR #3061, opened 2022-04-09
Refactor core rule processing for flexibility and speed

barrywhart:
Makes progress on #3035, #3037

These are just the core changes. In a separate PR, we will change some of the new settings (e.g. `needs_raw_stack` will default to `False` rather than `True`) and update existing rules to take advantage of the changes.

### Pull Request checklist

  - Other.


src/sqlfluff/core/linter/linter.py
```diff
@@ -494,112 +495,134 @@ def lint_fix_parsed(
             ignore_buff = []
 
         save_tree = tree
-        for loop in range(loop_limit):
```
barrywhart, line 497: Review with "Hide whitespace".

src/sqlfluff/core/linter/linter.py
```diff
@@ -494,112 +495,134 @@ def lint_fix_parsed(
...
-                    ignore_mask=ignore_buff,
-                    dialect=config.get("dialect_obj"),
-                    fname=fname,
-                    fix=fix,
-                    templated_file=templated_file,
+        # There are two phases of rule running. The main loop is for most rules.
+        # These rules are assumed to interact and cause a cascade of fixes
+        # requiring multiple passes. The post loop is for post-processing rules,
+        # not expected to trigger any downstream rules, e.g. capitalization fixes.
+        for phase in ["main", "post"]:
```
barrywhart, line 502: Adds phases to rule processing (#3035)

src/sqlfluff/core/rules/base.py
```diff
@@ -349,18 +349,37 @@ def _raw_slices_from_templated_slices(
 class RuleContext:
     """Class for holding the context passed to rule eval functions."""
 
-    segment: BaseSegment
```
barrywhart, line 352: Reordered the fields so the ones that have a "natural" default value are at the bottom. Python "dataclasses" require them to be last. Providing defaults here allows us to reduce the long list of `crawl()` parameters and simplify the code (for human readability).

src/sqlfluff/core/rules/base.py
```diff
@@ -349,18 +349,37 @@ def _raw_slices_from_templated_slices(
...
+    # These change within a file.
+    segment: BaseSegment
+    parent_stack: Tuple[BaseSegment, ...] = field(default=tuple())
+    raw_segment_pre: Optional[RawSegment] = field(default=None)
+    raw_stack: Tuple[RawSegment, ...] = field(default=tuple())
+    memory: Any = field(default_factory=dict)
+    segment_idx: int = field(default=0)
+
+    @property
+    def siblings_pre(self) -> Tuple[BaseSegment, ...]:  # pragma: no cover
```
barrywhart, line 367: Changed from a normal field to a property computed on the fly. Very few rules. use this or `siblings_post`, so don't waste time building these tuples unless needed.

src/sqlfluff/core/rules/base.py
```diff
@@ -372,7 +391,7 @@ class FunctionalRuleContext:
     def __init__(self, context: RuleContext):
         self.context = context
 
-    @cached_property
+    @property
```
barrywhart, line 394: Eliminated caching for most fields now that the same `RuleContext` instance is used throughout processing of a rule. Caching would be wrong in this case (or we'd have to keep clearing the cache between segments).

src/sqlfluff/core/rules/base.py
```diff
@@ -416,6 +435,59 @@ def raw_segments(self):
         )
 
 
+class CrawlBehavior:
```
barrywhart, line 438: I created this class by moving the parse tree traversal from `crawl()` to here. Making it a separate class makes it easier to vary the behavior, make it configurable, etc. with less risk of breaking other things (or having future changes to the other parts of the linter loop break _this_ code).

src/sqlfluff/core/rules/base.py
```diff
@@ -432,6 +504,17 @@ class BaseRule:
     _adjust_anchors = False
     targets_templated = False
 
+    # Lint loop / crawl behavior. When appropriate, rules can (and should)
```
barrywhart, line 507: These are new rule behaviors that rules can leverage to make linting faster. The rules will be updated in a separate PR.

src/sqlfluff/core/rules/base.py
```diff
@@ -484,149 +567,107 @@ def _eval(self, context: RuleContext) -> EvalResultType:
 
     def crawl(
```
barrywhart, line 568: Simplified the parameter list of this function. Also it is no longer recursive, now that `CrawlBehavior` does the actual iterating over the parse tree.

src/sqlfluff/core/linter/linter.py
```diff
@@ -494,112 +495,142 @@ def lint_fix_parsed(
...
-                        else:
-                            # Applying these fixes took us back to a state which we've
-                            # seen before. We're in a loop, so we want to stop.
-                            cls._warn_unfixable(crawler.code)
 
-            if loop == 0:
-                # Keep track of initial errors for reporting.
-                initial_linting_errors = all_linting_errors.copy()
+                for crawler in progress_bar_crawler:
+                    # Performance: After first loop pass, skip rules that don't
```
barrywhart, line 528: @tunetheweb asked in Slack: I’m still confused why we don’t need these in the final phase (which is a post) for any new code added by other rules

barrywhart: This is a good point. I think you could call this a design decision, and the linter has **always** behaved this way. When the user runs `fix`, they only see `initial_linting_errors`, not `all_linting_errors`.

For more context, (`crawl()`) computes two sets of errors:
* `initial_linting_errors`: This is what the function actually returns. It is assigned exactly once. Old behavior: Assigned on loop 1. New behavior: Assigned on loop 1 of the main phase (basically same behavior). Note that loop 1 of the main phase runs all the rules, including the post rules. I'll add PR comments to these places with additional comments.
* `all_linting_errors`:

src/sqlfluff/core/linter/linter.py
```diff
@@ -494,112 +495,142 @@ def lint_fix_parsed(
...
+            else:
+                rules_this_phase = rule_set
+            for loop in range(loop_limit if phase == "main" else 2):
+                linter_logger.info(f"Linter phase {phase}, loop {loop+1}/{loop_limit}")
+                changed = False
+
+                if phase == "main" and loop == 0:
+                    # In order to compute initial_linting_errors correctly, need
+                    # to run all rules on the first loop of the main phase.
+                    rules_this_phase = rule_set
```
barrywhart, line 519: Main phase, loop 1 **always runs all the rules**. This guarantees that `initial_linting_errors` includes output from all the rules. Should be identical behavior as before.

src/sqlfluff/core/linter/linter.py
```diff
@@ -494,112 +495,142 @@ def lint_fix_parsed(
...
+                                continue
+                            else:
+                                # Applying these fixes took us back to a state
+                                # which we've seen before. We're in a loop, so
+                                # we want to stop.
+                                cls._warn_unfixable(crawler.code)
+
+                if phase == "main" and loop == 0:
+                    # Keep track of initial errors for reporting.
+                    initial_linting_errors = all_linting_errors.copy()
```
barrywhart, line 596: This is where we capture `initial_linting_errors`, which is what the function returns (and what the user sees in the list of violations). See above -- this pass runs all rules (whether main or post, whether they are "fix compatible" or not. So the user sees a list of initial errors, but they don't see any linting errors introduced by fixes. This is the past behavior and remains true with these changes.

I'll see about some code comments and/or small code changes to make this clearer.

src/sqlfluff/core/linter/linter.py
```diff
@@ -472,6 +473,14 @@ def lint_fix_parsed(
...
         # Keep track of the linting errors
         all_linting_errors = []
+        # Keep track of the linting errors on the very first linter pass. The
+        # list of issues output by "lint" and "fix" only includes issues present
+        # in the initial SQL code, EXCLUDING ANY ISSUES THAT MAY BE CREATED BY
+        # THE FIXES THEMSELVES.
+        # The value is initially None, and is set to a list value on the first
+        # pass. The value None is used as a sentinel to indicate we're on the
+        # first pass.
+        initial_linting_errors = None
```
barrywhart, line 483: @tunetheweb: I added this code and comment to make the code clearer. This allowed me to update two places that were doing `if phase == "main" and loop == 0:` with `if initial_linting_errors is None:`. I think this clarifies the behavior. The type checker in PyCharm (not mypy itself, though) has always complained that `initial_linting_errors` was not guaranteed to be assigned a value, and this makes it happier too.

src/sqlfluff/core/linter/linter.py
barrywhart: I went back and refactored how this works. The old code was unnecessarily tricky -- the `all_linting_errors` was really not used after the first pass, so why have it? 🤯 

Note that this PR doesn't change any tests except for those two oddball "runaway" tests, so if the tests pass, I think we can be fairly confident that changing the bookkeeping didn't break anything. (In fact, several of the little changes inside the linter loop that we've been discussing were added because some tests failed, i.e. caught some initial bugs. 🎉 🥳 )

src/sqlfluff/core/rules/base.py
```diff
@@ -416,6 +435,59 @@ def raw_segments(self):
...
+    def __init__(
+        self, works_on_unparsable: bool, recurse_into: bool, needs_raw_stack: bool
+    ):
+        self.works_on_unparsable = works_on_unparsable
+        self.recurse_into = recurse_into
+        self.raw_stack: Optional[List[RawSegment]]
+        if needs_raw_stack:
+            self.raw_stack = []
+        else:
+            self.raw_stack = None  # pragma: no cover
```
barrywhart, line 450: "no cover" for now. Will be used in a subsequent PR, when we update rules to take advantage of this capability.

src/sqlfluff/core/linter/linter.py
```diff
@@ -489,117 +493,152 @@ def lint_fix_parsed(
...
+            if len(phases) > 1:
+                rules_this_phase = [
+                    rule for rule in rule_set if rule.lint_phase == phase
+                ]
+            else:
+                rules_this_phase = rule_set
+            for loop in range(loop_limit if phase == "main" else 2):
+
+                def is_first_linter_pass():
+                    return phase == phases[0] and loop == 0
```
OTooleMichael, line 518: Why defining this function inline, It cant be used anywhere else because of scoping issues right?

src/sqlfluff/core/rules/base.py
OTooleMichael, line 450: Can you not drop the else and assign directly (or with a ternary) - seems cleaner and avoids the no cover
```suggestion
        self.raw_stack: Optional[List[RawSegment]] = [] if needs_raw_stack else None
```

src/sqlfluff/core/rules/base.py
```diff
@@ -416,6 +435,59 @@ def raw_segments(self):
...
+            for idx, child in enumerate(context.segment.segments):
+                # For performance reasons, don't create a new RuleContext for
+                # each segment; just modify the existing one in place. This
+                # requires some careful bookkeeping, but it avoids creating a
+                # HUGE number of short-lived RuleContext objects
+                # (#linter loops x #rules x #segments).
+                context.segment = child
+                context.parent_stack = new_parent_stack
+                context.segment_idx = idx
+                yield from self.crawl(context)
```
OTooleMichael, line 488: Is there a benefit to mutation. I assume its faster (cause of python), but it always gives me a bad feeling.
Is there anything preventing the crawl fn from further mutating this state and blowing things up in a completely undebuggable way?

src/sqlfluff/core/linter/linter.py
```diff
@@ -470,8 +471,11 @@ def lint_fix_parsed(
         formatter: Any = None,
     ) -> Tuple[BaseSegment, List[SQLBaseError], List[NoQaDirective]]:
         """Lint and optionally fix a tree object."""
-        # Keep track of the linting errors
-        all_linting_errors = []
+        # Keep track of the linting errors on the very first linter pass. The
+        # list of issues output by "lint" and "fix" only includes issues present
+        # in the initial SQL code, EXCLUDING ANY ISSUES THAT MAY BE CREATED BY
+        # THE FIXES THEMSELVES.
```
tunetheweb, line 477: Bit OTT

```suggestion
        # in the initial SQL code, EXCLUDING any issues that may be created by
        # the fixes themselves.
```

src/sqlfluff/core/rules/base.py
```diff
@@ -432,6 +504,17 @@ class BaseRule:
...
+    # override these values to make linting faster.
+    recurse_into = True
+    needs_raw_stack = True
+    # Rules can override this to specify "post". "Post" rules are those that are
+    # not expected to trigger any downstream rules, e.g. capitalization fixes.
+    # They run on two occasions:
+    # - On the first loop of the main phase
+    # - In a second linter loop after the main rules run
+    lint_phase = "main"
+
```
OTooleMichael, line 517: Anything to be said for a memory_factory prop?
then we can instantiate the memory at the right times in the crawl. and save that first bit of boilerplate? And encourage typing?

src/sqlfluff/core/rules/base.py
```diff
@@ -484,149 +567,107 @@ def _eval(self, context: RuleContext) -> EvalResultType:
...
+                        fixes=[],
+                        description=(
+                            f"Unexpected exception: {str(e)};\n"
+                            "Could you open an issue at "
+                            "https://github.com/sqlfluff/sqlfluff/issues ?\n"
+                            "You can ignore this exception for now, by adding "
+                            f"'-- noqa: {self.code}' at the end\n"
+                            f"of line {exception_line}\n"
+                        ),
+                    )
```
OTooleMichael, line 627: Might be nice to move this sort of _safe_exec logic into a wrapper of some sort.
such that one could focus on the loop logic.

OTooleMichael (approved):
LGTM. Really cool improvements, comments mostly relate to bits that you didn't rewrite

src/sqlfluff/core/linter/linter.py
```diff
@@ -489,117 +493,152 @@ def lint_fix_parsed(
...
-                            # seen before. We're in a loop, so we want to stop.
-                            cls._warn_unfixable(crawler.code)
 
-            if loop == 0:
-                # Keep track of initial errors for reporting.
-                initial_linting_errors = all_linting_errors.copy()
+                for crawler in progress_bar_crawler:
+                    # Performance: After first loop pass, skip rules that don't
+                    # do fixes. Any results returned won't be seen by the user
+                    # anyway, so there's absolutely no reason to run them.
```
tunetheweb, line 537:
```suggestion
                    # do fixes. Any results returned won't be seen by the user
                    # anyway (linting errors ADDED by rules changing SQL, are
                    # not reported back to the user - only initial linting errors),
                    # so there's absolutely no reason to run them.
```

tunetheweb (approved):
A few comments on the comments.

Other than that can't spot anything.

@OTooleMichael I see you've thumbs up a few things. Are you planning on reviewing this later?

src/sqlfluff/core/linter/linter.py
barrywhart: I made it inline so it'd have access to the variables it needs. Making a separate function with 3 parameters that's only used from one place would've made things harder to understand, I think.

There are a few different ways to do this. The main goal was to make it pretty clear which bits of code inside the loop only run the very first time. (Having two phases was making the old logic -- which was already a little messy -- too difficult to follow.)

src/sqlfluff/core/rules/base.py
barrywhart: Will do! I was debating with myself earlier the best way to do this.

src/sqlfluff/core/rules/base.py
barrywhart: Performance, yes. Creating new objects is pretty expensive, and as noted in the comment above, this was probably **the** most created object in SQLFluff prior to this change (#linter loops x #rules x #segments).

Note we do create a new rule context per rule, so if a rule mutates it, it only affects that one rule, which I think helps a lot when it comes to debugging.

barrywhart: We could look into having the rules access the properties through property getter functions, while the core linter modifies them by changing private fields (e.g. prefixed with an underscore) behind the scenes. Probably a later PR...

src/sqlfluff/core/rules/base.py
barrywhart: I'd definitely like to streamline use of memory some time. Probably a separate issue/PR (performance vs tech debt/design improvement).

src/sqlfluff/core/rules/base.py
barrywhart: I'll do this in the other PR, to avoid having to wait for another CI cycle.

barrywhart: Welp, I'll go ahead and do it now. Got to "Update branch" from `main` anyway.

PR #3075, opened 2022-04-11
Snowflake: Add Create Storage Integration grammar.

jmc-bbk:

We don't currently support `CREATE ... STORAGE INTEGRATION` grammar in Snowflake and as a result this is unparsable.

https://docs.snowflake.com/en/sql-reference/sql/create-storage-integration.html

Added CREATE Storage Integration grammar to Snowflake.

Not that I'm aware of. 

### Pull Request checklist

  - Other.


jmc-bbk:
I have left a couple of `TODO:` statements that it would be great to receive some help with.

- We should limit one argument to be `'bucket-owner-full-control'` which is the only valid string instead of any string.
- One argument should be a quoted wildcard character (`'*'`). I'm unsure how to handle this - similar to above.
- Finally, I left lots of comments (which can be removed) about the valid PATHs for `STORAGE_ALLOWED_LOCATIONS`. I split out S3, GCS, and Azure into their own grammars, as they have their own unique parameters. However, the current set up would (e.g.) allow you to have S3 cloud parameters with GCS paths in `STORAGE_ALLOWED_LOCATIONS`.

On the last point above - I could move `STORAGE_ALLOWED_LOCATIONS` and `STORAGE_BLOCKED_LOCATIONS` into the <Cloud-Provider>StorageIntegrationParameters segments. But I feel like this is less readable/easy to understand when comparing against the Snowflake docs.

Open to opinions and I will either delete my comments in the PR or move the logic 👍 .

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -2342,6 +2342,65 @@ class CreateStatementSegment(BaseSegment):
...
+                    Ref("EqualsSegment"),
+                    Bracketed(
+                        Delimited(
+                            OneOf(
+                                Ref("S3Path"),
+                                Ref("GCSPath"),
+                                Ref("AzureBlobStoragePath"),
+                                Ref(
+                                    "QuotedLiteralSegment"
+                                ),  # TODO: Need help. Should only be '*'.
```
tunetheweb, line 2372: As this one is used elsewhere, you should add a `QuotedStarSegment` in the section starting at line 89.

```py
    QuotedStarSegment=StringParser(
        "'*'",
        CodeSegment,
        name="quoted_star",
        type="identifier",
        trim_chars=("'",),
    ),
```

Then can change to this:

```suggestion
                                Ref("QuotedStarSegment"),
```

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -2342,6 +2342,65 @@ class CreateStatementSegment(BaseSegment):
...
```
tunetheweb, line 2389:
```suggestion
                                Ref("QuotedStarSegment"),
```

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -2764,6 +2823,58 @@ class StorageLocation(BaseSegment):
...
+        Sequence("STORAGE_PROVIDER", Ref("EqualsSegment"), "S3"),
+        Sequence(
+            "STORAGE_AWS_ROLE_ARN", Ref("EqualsSegment"), Ref("QuotedLiteralSegment")
+        ),
+        Sequence(
+            "STORAGE_AWS_OBJECT_ACL",
+            Ref("EqualsSegment"),
+            Ref(
+                "QuotedLiteralSegment"
+            ),  # TODO: Can only be 'bucket-owner-full-control'.
```
tunetheweb, line 2845: For one ofs, rather than defining a segment, you can embed a StringParser directly in the dialect:

```suggestion
            StringParser("'bucket-owner-full-control'", CodeSegment, type="literal"),
```

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -2342,6 +2342,65 @@ class CreateStatementSegment(BaseSegment):
...
+                OneOf(
+                    Ref("S3StorageIntegrationParameters"),
+                    Ref("GCSStorageIntegrationParameters"),
+                    Ref("AzureStorageIntegrationParameters"),
+                ),
+                # Note Could result in mismatched IntegrationParams and AllowedLocations
+                # E.G. S3 IntegrationParameters would parse with GCS compliant paths
+                # Alternative would be to add these to IntegrationParameter Segments but
+                # I feel like that is harder to understand the match_grammar
+                # when comparing against Snowflake docs.
```
tunetheweb, line 2360: It's not necessary to write 100% compliant parsing code. While it would be nice to only parse exactly what Snowflake would accept., you need to weigh up how likely the above is, whether it would be expected for a SQL linter to catch those, versus the complexity of the code you're adding to SQLFluff. We've lots of places where we are overly permissive as to what we allow.

Also how likely is it someone will add S3 in one place and GCS on another? And does Snowflake itself throw a parsing error in this case? Or a runtime error?

Saying that I don't think adding to the IntegrationParameters is that bad. A comment would explain it.

Or alternatively do something like this:

```py
                OneOf(
                    Sequence(
                        Ref("S3StorageIntegrationParameters"),
                        Ref("S3StorageAllowdLocations"),
                    ),
                    Sequence(
                        Ref("GCSStorageIntegrationParameters"),
                        Ref("GCSStorageAllowdLocations"),
                    ),
                    Sequence(
                        Ref("AzureStorageIntegrationParameters"),
                        Ref("AzureStorageAllowdLocations"),
                    ),
                ),
```

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_snowflake.py
jmc-bbk: In this case - it would only be a runtime error, but the parsing would be fine.

Given the above, do you think I should leave it as it is (with my comments removed) or I could also change to as you suggest (with slight edit):

```
 OneOf(
                    AnySetOf(
                        Ref("S3StorageIntegrationParameters"),
                        Ref("S3StorageAllowdLocations"),
                    ),
```

I think I'm happy to leave the code as it is - like you mention, we can parse it and so can Snowflake. It's only the equivalent to someone adding a bad S3 path that doesn't actually exist!

Let me know and I can update tomorrow :) Saw your other comments too, thanks super helpful - will implement them tomorrow morn!

tunetheweb: Leaving as is SGTM!

jmc-bbk:
Note - I made another minor change that wasn't requested. For both `STORAGE_ALLOWED` and `STORAGE_BLOCKED` I updated this snippet 

```
                        Bracketed(
                            Delimited(
                                OneOf(
                                    Ref("S3Path"),
                                    Ref("GCSPath"),
                                    Ref("AzureBlobStoragePath"),
                                   Ref("QuotedStarSegment"),
                                )
                            )
                        ),
```

to this snippet

```
                    OneOf(
                        Bracketed(
                            Delimited(
                                OneOf(
                                    Ref("S3Path"),
                                    Ref("GCSPath"),
                                    Ref("AzureBlobStoragePath"),
                                )
                            )
                        ),
                        Bracketed(
                            Ref("QuotedStarSegment"),
                        ),
```

Before, I was getting any number of QuotedStarSegments in the `Delimited()` - but you can only have 1 Bracketed, QuotedStarSegment - so I have updated, and tests still pass.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -2523,6 +2530,63 @@ class CreateStatementSegment(BaseSegment):
...
+                                    Ref("AzureBlobStoragePath"),
+                                )
+                            )
+                        ),
+                        Bracketed(
+                            Ref("QuotedStarSegment"),
+                        ),
+                    ),
+                    optional=True,
+                ),
```
tunetheweb, line 2580: As these are very similar I wonder if they can be combined? As they are part of an `AnySetOf` it should still be allowed to have both.

Also the second one ahas an `optional=True` on line 2579 but don't think that's needed as it's part of an `AnySetOf` so optionality is implied.

```suggestion
                Sequence(
                    OneOf("STORAGE_ALLOWED_LOCATIONS", "STORAGE_BLOCKED_LOCATIONS"),
                    Ref("EqualsSegment"),
                    OneOf(
                        Bracketed(
                            Delimited(
                                OneOf(
                                    Ref("S3Path"),
                                    Ref("GCSPath"),
                                    Ref("AzureBlobStoragePath"),
                                )
                            )
                        ),
                        Bracketed(
                            Ref("QuotedStarSegment"),
                        ),
                    ),
                ),
```

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -2523,6 +2530,63 @@ class CreateStatementSegment(BaseSegment):
...
+                Sequence(
+                    "COMMENT",
+                    Ref("EqualsSegment"),
+                    Ref("QuotedLiteralSegment"),
+                    optional=True,
```
tunetheweb, line 2585: Don't think this is needed as implied as part of `AnySetOf`

```suggestion
```

tunetheweb (changes requested):
Looks pretty good, but have a few more suggestions and then can merge this.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -2523,6 +2530,63 @@ class CreateStatementSegment(BaseSegment):
...
+                ),
+            ),
+            optional=True,
```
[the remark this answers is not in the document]
jmc-bbk: Ah, I was going from the docs where it has this syntax:

```
...
  ENABLED = { TRUE | FALSE }
  STORAGE_ALLOWED_LOCATIONS = ('<cloud>://<bucket>/<path>/', '<cloud>://<bucket>/<path>/')
  [ STORAGE_BLOCKED_LOCATIONS = ('<cloud>://<bucket>/<path>/', '<cloud>://<bucket>/<path>/') ]
```

so I think it comes down to the parse versus. runtime error issue.

With the above, `STORAGE_ALLOWED_LOCATIONS` is required - but it can come before the other arguments (e.g.) `ENABLED` - whereas `STORAGE_BLOCKED_LOCATIONS` is optional.

I suppose what I'm really looking for is an `UnorderedSequence()` rather than `AnySetOf()` - I didn't actually realise that `AnySetOf()` is implicitly optional.

Is there anything like `UnorderedSequence` available?

If not - I'm happy to do your suggestions of removing the `optional=True` from these. Let me know!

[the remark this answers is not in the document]
tunetheweb: Don't think we can really handle that to be honest. So think we need to basically treat both as optional.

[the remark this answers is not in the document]
jmc-bbk: Sure thing, I'll start to make the above changes :)

jmc-bbk:
Hi @tunetheweb, sorry for the delay here!

I actually noticed from your comments that I had a problem in my original code. `STORAGE_BLOCKED_LOCATIONS` cannot accept `('*')` as a valid argument.

I have now removed that from that section.

Given that makes the `Sequence()` different for `STORAGE_BLOCKED_LOCATIONS` and `STORAGE_ALLOWED_LOCATIONS`, I haven't merged those as in your above suggestion. But I'm still open to, if you think that's the right move?

---

I also looked at the `optional = True` statements you mentioned in the other sections (e.g. `PIPE` / `WAREHOUSE`). I've noticed that all those `Sequence()`s are `optional=True`.

It looks like this `CreateStatementSegment(...)` could be too broad at the moment. It might need to be decomposed into

```
CreatePipeSegment
CreateIntegrationSegment
...
CreateXSegment
```

What do you think?

If the above sounds right, I could work on it in another PR - but maybe it's not in scope of the current work?

Anyway, let me know - and happy to make changes :)

tunetheweb:
I actually typed that as a suggestion previously! 😁 Then delighted it as thought it wasn’t fair to you to have to take that on in this PR. Do agree we should do it as a separate PR once this is merged.

jmc-bbk:
Great! Let me know if there's anything outstanding :)

src/sqlfluff/dialects/dialect_snowflake.py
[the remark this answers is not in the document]
tunetheweb: I still don't think we need this `optional=True`

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -2944,6 +3001,56 @@ class StorageLocation(BaseSegment):
...
+    match_grammar = AnySetOf(
+        Sequence("STORAGE_PROVIDER", Ref("EqualsSegment"), "S3"),
+        Sequence(
+            "STORAGE_AWS_ROLE_ARN", Ref("EqualsSegment"), Ref("QuotedLiteralSegment")
+        ),
+        Sequence(
+            "STORAGE_AWS_OBJECT_ACL",
+            Ref("EqualsSegment"),
+            StringParser("'bucket-owner-full-control'", CodeSegment, type="literal"),
+            optional=True,
```
tunetheweb, line 3022: Nor this.

```suggestion
```

tunetheweb (changes requested):
Looks pretty much there too me. But still think we can drop two of the `optional=True`s

jmc-bbk:
https://github.com/sqlfluff/sqlfluff/pull/3075#discussion_r850582411

For the optional on `L:2580`, I think we do require it. As I'm setting all the arguments for the `CREATE STORAGE INTEGRATION` syntax to be optional.

If I were to remove that optional, I think we would be unable to parse the next section for `CREATE PIPE` - as it would be expecting the `CREATE STORAGE INTEGRATION` arguments beforehand?

Or I might still be misunderstanding `AnySetOf` 🥴 - I have pushed the change to `L:3022` will try the test suite on my local if I remove the one on `L:2580`.

jmc-bbk:
Just tested `pytest -s test/dialects/dialects_test.py -k snowflake-snowflake_create_pipe.sql` without the `optional=True` on `L:2580` and it would fail!

tunetheweb:
Oh I guess I’m misunderstanding it! I thought `AnySetOf` was 0 or 1 of each of the options. Therefore the 0 was the same as optional.

Maybe you need an optional on the overall `AnySetOf` if none of them apply? And maybe that’s currently being effectively picked up by the fact one of your options is optional?

jmc-bbk:
I'm actually having trouble viewing this in the Github UI - maybe this is causing some confusion - but I think I already have the above suggestion. Hopefully easier to view if I paste the relevant snippet:

```
        Sequence(
            AnySetOf(
                Sequence("TYPE", Ref("EqualsSegment"), "EXTERNAL_STAGE"),
                Sequence("ENABLED", Ref("EqualsSegment"), Ref("BooleanLiteralGrammar")),
                OneOf(
                    Ref("S3StorageIntegrationParameters"),
                    Ref("GCSStorageIntegrationParameters"),
                    Ref("AzureStorageIntegrationParameters"),
                ),
                Sequence(
                    "STORAGE_ALLOWED_LOCATIONS",
                    Ref("EqualsSegment"),
                    OneOf(
                        Bracketed(
                            Delimited(
                                OneOf(
                                    Ref("S3Path"),
                                    Ref("GCSPath"),
                                    Ref("AzureBlobStoragePath"),
                                )
                            )
                        ),
                        Bracketed(
                            Ref("QuotedStarSegment"),
                        ),
                    ),
                ),
                Sequence(
                    "STORAGE_BLOCKED_LOCATIONS",
                    Ref("EqualsSegment"),
                    Bracketed(
                        Delimited(
                            OneOf(
                                Ref("S3Path"),
                                Ref("GCSPath"),
                                Ref("AzureBlobStoragePath"),
                            )
                        )
                    ),
                ),
                Sequence(
                    "COMMENT",
                    Ref("EqualsSegment"),
                    Ref("QuotedLiteralSegment"),
                ),
            ),
            optional=True,
        ),
```

So I'm expecting that to be:

```
AnySetOf(
  <x>,
  <y>,
  <z>,
  optional=True,
)

```

Which I think is what you're suggesting?

jmc-bbk:
Hold on - maybe I'm really misunderstanding - let me try remove the first `Sequence()`...

src/sqlfluff/dialects/dialect_snowflake.py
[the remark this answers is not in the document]
tunetheweb: Ah yes. This was me IS needed!

tunetheweb:
I think you’re good now actually.

tunetheweb (approved):
I’m happy to merge this now if you are?

jmc-bbk:
I'm going to try and remove the `Sequence()` - will be 2 minutes! Want to test on my local before pushing :)

jmc-bbk:
I was speaking to @WittierDinosaur and he pointed out that `AnySetOf()` is 0 or 1. In this case, I had it wrapped in `Sequence()` which expects at least one thing to be returned, and that's why the `optional=True` here was previously necessary.

Now I've unwrapped it, hopefully the tests pass - and if you agree, it should be ready for merge!

I'm definitely keen to take a crack at decomposing this class though. As with the above knowledge, the `CREATE PIPE` part below is definitely suspect (an optional sequence full of optional sequences!).

tunetheweb:
We good to merge this now @jmc-bbk ?

jmc-bbk:
Yep! LGTM.

PR #3098, opened 2022-04-14
L010: Add `ignore_words_regex` configuration option

### LINKED ISSUE #3089: Provide more flexible way of defining `ignore_words` for rules
### Search before asking

### Description

My suggestion here is to add an additional configuration option `ignore_words_regex` along side `ignore_words` to handle more complex exclusion cases.

My original use-case is L030, but I think all rules using `ignore_words` are based off of L010, so we might just add this to all children rules as well.

### Use case

The example where this use-case came up is L030. In my use-case we want to capitalize all functions, however UDF we want to be excluded from that policy.

Currently, we have to manage a very long `ignore_words` list, which is impracticable. However, if we could provide a regex this would be simple as all our UDF follow some naming convention (`_f_<function_name>` in our case).

### Dialect

all

### Code of Conduct

dmohns:

Fixes #3089 

~I added this to all rules that are based on `L010`, but this does not cover all rules that have the `ignore_words` option.~

### Pull Request checklist

  - Other.


tunetheweb:
> I added this to all rules that are based on L010, but this does not cover all rules that have the ignore_words option.

I think this is confusing, and think we should ensure it's available everywhere `ignore_words` is. It's a little extra work, but not too much.

src/sqlfluff/rules/L010.py
```diff
@@ -87,19 +87,25 @@ def _eval(self, context: RuleContext) -> LintResult:
             cap_policy = self.cap_policy
             cap_policy_opts = self.cap_policy_opts
             ignore_words_list = self.ignore_words_list
+            ignore_words_regex = self.ignore_words_regex
```
tunetheweb, line 90: This is not exactly how this work.

You may not have noticed but we have the `ignore_words` config and ` ignore_words_list` attribute. Look similar but not quite the same, and the `ignore_words_list` are a split version of `ignore_words` config. So we do that once to cache it.

There is no processing needed for `ignore_words_regex`, but still think I'd call the attribute differently.

Or, since it needs no processing, you could just not have an attribute and delete this and call `getattr(self, "ignore_words_regex"))` directly on line 106 below? Though do need it twice so probably better to stick it into an attribute.

tunetheweb (changes requested):

src/sqlfluff/rules/L010.py
dmohns: I see. I'm not not 100% sure I understood correctly, but can you check latest commit to see if this is what you're saying?

tunetheweb: Yes that looks good now. Forgot you could use `self.ignore_words_regex` directly for configs/

test/fixtures/rules/std_rule_cases/L030.yml
```diff
@@ -65,3 +65,26 @@ test_fail_func_name_templated_literal_mix:
   # Issue 3022. This was actually a bug in BaseSegment.iter_patches().
   fail_str: SELECT RO(), {{ "t" }}.func()
   fix_str: SELECT RO(), {{ "t" }}.FUNC()
+
+test_pass_ignore_words_regex_simple:
+  pass_str: SELECT MAX(id), f_test_udf(id) FROM TABLE1
+  configs:
+    rules:
+      L030:
+        ignore_words_regex: ^f_
```
tunetheweb, line 74: Since L030 inherits from L010 should the tests be on L010 instead of (or as well as?) L030?

Happy to leave on here as well (since this is the one your really want!), and don't think it's necessary to add tests to ALL the rules adding this config, but would add it to L010 as the base rule for most of these.

And similarly to the other base rules that need this when you do add the, (L029, L057, L059).

dmohns: Yeah, makes a lot of sense. I added new commits addressing this.

dmohns:
While working on writing the tests I stumbled upon one small thing.

I am using `regex.match()` to apply the regex. However, maybe the way `match` works is unintuitive to users not too familiar with Python internals.

For example

```
>>> print(regex.match("hash$", "aliashash"))
None
>>> print(regex.match("^alias", "aliashash"))
<regex.Match object; span=(0, 5), match='alias'>
```

Would it be more suitable to use `regex.search` here?

tunetheweb:
`regex.search` seems more sensible to me, but not a python expert here.

@barrywhart @WittierDinosaur any thoughts or concerns here?

Obviously there is a risk of someone slowing down SQLFluff with a really complex regex, but I think given it will only be applied to small identifiers, and that this is configured by the user, I'm OK with that. I presume python does exit early if it's not set, in code like this and won't actually execute the regex in this instance?:

```py
        if self.ignore_words_regex and regex.match(
            self.ignore_words_regex, context.segment.raw
        ):
```

tunetheweb (approved):
LGTM!

Would still be good to get @barrywhart and @WittierDinosaur 's view if they get a chance so won't merge yet.

src/sqlfluff/rules/L010.py
```diff
@@ -100,6 +103,12 @@ def _eval(self, context: RuleContext) -> LintResult:
         if ignore_words_list and context.segment.raw.lower() in ignore_words_list:
             return LintResult(memory=context.memory)
 
+        # Skip if matches ignore regex
+        if self.ignore_words_regex and regex.search(
```
barrywhart, line 107: I notice we're using `search()` rather than `match()`.
* Should we consider using `match()`? (Less risk of accidental matches causing confusing behavior.)
* If not, should we document that the patterns should probably either start with `^`, end with `$`, or both?

tunetheweb: Look in comments above @barrywhart. Search seemed easier on users than match.

barrywhart: I'm okay with ‘search()`, but I do think we should document it. It could be simple, just an example in the rule docs page or a commented-out example (starting with `^`) in the `.py` file that documents what the option means.

dmohns: I like the idea of just clarifying it in the docs. I added small description.

tunetheweb:
This good to merge now?

barrywhart (approved):
Looks good!

Issue #3099, opened 2022-04-14
SparkSQL: Interval Literal inside of MERGE causes parsing violations


R7L208:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

```bash
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    statement:
[L:  1, P:  1]      |        unparsable:                                           !! Expected: 'StatementSegment'
[L:  1, P:  1]      |            raw:                                              'MERGE'
[L:  1, P:  6]      |            whitespace:                                       ' '
[L:  1, P:  7]      |            raw:                                              'INTO'
[L:  1, P: 11]      |            newline:                                          '\n'
[L:  2, P:  1]      |            whitespace:                                       '    '
[L:  2, P:  5]      |            raw:                                              'logs'
[L:  2, P:  9]      |            newline:                                          '\n'
[L:  3, P:  1]      |            raw:                                              'USING'
[L:  3, P:  6]      |            newline:                                          '\n'
[L:  4, P:  1]      |            whitespace:                                       '    '
[L:  4, P:  5]      |            raw:                                              'new_deduped_logs'
[L:  4, P: 21]      |            newline:                                          '\n'
[L:  5, P:  1]      |            raw:                                              'ON'
[L:  5, P:  3]      |            whitespace:                                       ' '
[L:  5, P:  4]      |            raw:                                              'logs'
[L:  5, P:  8]      |            raw:                                              '.'
[L:  5, P:  9]      |            raw:                                              'unique_id'
[L:  5, P: 18]      |            whitespace:                                       ' '
[L:  5, P: 19]      |            raw:                                              '='
[L:  5, P: 20]      |            whitespace:                                       ' '
[L:  5, P: 21]      |            raw:                                              'new_deduped_logs'
[L:  5, P: 37]      |            raw:                                              '.'
[L:  5, P: 38]      |            raw:                                              'unique_id'
[L:  5, P: 47]      |            whitespace:                                       ' '
[L:  5, P: 48]      |            raw:                                              'AND'
[L:  5, P: 51]      |            whitespace:                                       ' '
[L:  5, P: 52]      |            raw:                                              'logs'
[L:  5, P: 56]      |            raw:                                              '.'
[L:  5, P: 57]      |            raw:                                              'date'
[L:  5, P: 61]      |            whitespace:                                       ' '
[...]

PR #3102, opened 2022-04-14
SparkSQL: Update to support all valid Literal Types

R7L208:

- Redefine `IntervalExpressionSegment` for Spark specific syntax
- adjust _eval in L006 to not flag `binary_operator` inside of `INTERVAL`
- - adjust _eval in L048 to not flag `binary_operator` inside of `INTERVAL`

Makes progress on #3099 

### Pull Request checklist

  - Other.


src/sqlfluff/rules/L006.py
```diff
@@ -154,6 +155,17 @@ def _eval(self, context: RuleContext) -> EvalResultType:
...
+                    # NB: Should not check for whitespaces inside of Interval Literals for SparkSQL dialect
+                    # https://spark.apache.org/docs/latest/sql-ref-literals.html#interval-literal
+                    possible_interval_segment = self._find_segment(
+                        idx, context.segment.segments, before=True, step=-2
+                    )
+                    if (
+                        possible_interval_segment is not None
+                        and possible_interval_segment.raw.upper() == "INTERVAL"
+                    ):
+                        continue
```
R7L208, line 168: I think this is brittle but the best could come up with exploring the rule for the first time.

Would be very open to suggestions on how to improve this.

src/sqlfluff/rules/L048.py
```diff
@@ -61,3 +68,140 @@ def _missing_whitespace(seg: BaseSegment, before=True) -> bool:
...
+                # https://spark.apache.org/docs/latest/sql-ref-literals.html#interval-literal
+                if context.dialect.name in ["sparksql"]:
+                    possible_interval_segment = self._find_segment(
+                        idx, context.segment.segments, before=True, step=-3
+                    )
+                    if (
+                        possible_interval_segment is not None
+                        and possible_interval_segment.raw.upper() == "INTERVAL"
+                    ):
+                        continue
```
R7L208, line 150: Similarly, I think this is brittle and would welcome feedback on improving it.

src/sqlfluff/rules/L006.py
```diff
@@ -145,15 +146,27 @@ def _eval(self, context: RuleContext) -> EvalResultType:
                         )
                         check_before = True
                     if self.matches_target_tuples(trailing, self._target_elems):
-                        after_anchor = trailing
-                        self.logger.debug(
+                        after_anchor = trailing  # pragma: no cover
+                        self.logger.debug(  # pragma: no cover
```
R7L208, line 150: Are there examples of tests cases that would provide coverage here? TBH, not sure why this is no longer covered with changes made

src/sqlfluff/rules/L006.py
```diff
@@ -145,15 +146,27 @@ def _eval(self, context: RuleContext) -> EvalResultType:
...
                             "Found Target [trailing] @%s: %r",
                             after_anchor.pos_marker,
                             after_anchor.raw,
                         )
-                        check_after = True
+                        check_after = True  # pragma: no cover
```
R7L208, line 155: Same as above on test cases

src/sqlfluff/rules/L048.py
```diff
@@ -89,7 +89,7 @@ def _eval(self, context: RuleContext) -> EvalResultType:
         # to have at least three children.
 
         if self._require_three_children and len(context.segment.segments) <= 2:
-            return LintResult()
+            return LintResult()  # pragma: no cover
```
R7L208, line 92: same as test cases comment above

src/sqlfluff/rules/L006.py
barrywhart: I don't know which test cases, but you could try [this feature](https://coverage.readthedocs.io/en/6.3.2/contexts.html#dynamic-contexts) of the coverage package we use for measuring coverage. It should be able to tell you which tests use which code.

I haven't used the technique above, but you can also do something similar, manually:
Switch back to the main branch and set a breakpoint on these lines, or log a stack trace. That'll tell you which test cases are executing these lines.

barrywhart (commented):
I did a quick, initial review. I'm wondering, could this be split into different PRs?

The dialect changes seem related to the linked issue. I have several questions/concerns about the rule changes:
* What issue are they addressing?
* Are they necessary? Could the same issues be addressed another, narrower way? I don't see use of `binary_operator` in the test cases, only leading (unary) minus.
* The L048 changes are large, and making multiple changes to the rule behavior.
* As you noted, the loss of coverage is concerning. Perhaps the existing tests were not very thorough?

R7L208:
`binary_operator` wouldn't actually be needed. I double-checked the docs and a few snippets to test. Only minus and plus are valid here. I'll include test cases with `+` too.

Let me narrow the scope of the valid symbol to just plus and minus inside of the Interval Segment in SparkSQL dialect. I will leave the rules for now so that the whitespace gets added since the change is largely based on my perception of formatting SparkSQL. If it gets raised as a formatting issue in the future, we can revisit it.

Thoughts on this?

barrywhart:
That sounds like an improvement. I also wonder if there's a possibility to handle most or all of this with dialect changes. (I don't know if it's possible, but I think there are similar cases where we've named similar segments differently, based on context.)

Happy to continue iterating and thinking on this.

@tunetheweb: Any thoughts on this? You are more familiar with the dialects and how they interact with the rules than I am.

barrywhart:
Small changes are generally better, and we can always revisit them later. In my own work, I often see surprises once something is released, unintended consequences (bugs). 😛 Now that we have lots of users, it's good because we get more testing, but it also is a motivation for smaller, lower impact PRs.

tunetheweb:
> That sounds like an improvement. I also wonder if there's a possibility to handle most or all of this with dialect changes. (I don't know if it's possible, but I think there are similar cases where we've named similar segments differently, based on context.)
> 
> Happy to continue iterating and thinking on this.
> 
> @tunetheweb: Any thoughts on this? You are more familiar with the dialects and how they interact with the rules than I am.

Yes, there is a much easier way to do this, without needing to change the rules at all.

Symbols can only be lexed once. However the same lexed symbol can be parsed in any number of ways. In much the same way, that an unreserved keyword can be parsed as a keyword, or an identifier for example. Don't want something to be parsed as a binary operator? Then just set up another parser with the same contents but a different type! We actually have two parsers for `+` and `-` already:
- `+` - `PlusSegment` (which is a binary operator and meant for `a + b`) and `PositiveSegment` (which is NOT a binary operator and meant for `+a`).
- `-` - `MinusSegment` (which is a binary operator and meant for `a - b`) and `NegativeSegment` (which is NOT a binary operator and meant for `-a`).

tunetheweb (changes requested):

tunetheweb:
Actually it appears L048 does flag this, so do need changes there.

However I think this would be easier to set up as a new class with a type (`signed_quoted_literal`?) to stop it being caught by the rule, rather than make complicated rule changes.

src/sqlfluff/dialects/dialect_sparksql.py
```diff
@@ -2572,3 +2582,48 @@ class UpdateStatementSegment(ansi.UpdateStatementSegment):
...
+        OneOf(
+            # multi-units syntax
+            OneOf(
+                AnyNumberOf(
+                    Sequence(
+                        OneOf(  # TODO: disable L47
+                            Ref("PlusSegment"),
+                            Ref("MinusSegment"),
+                            optional=True,
+                        ),
```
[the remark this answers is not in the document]
tunetheweb: I've made this update in #3108 and also added the `SignedSegmentGrammar`. Merge main into this branch and you can use it here too.

tunetheweb:
The `name` is `quoted_literal` and that's what the rule is using, but the `type` (which is what is shown in the parse tree) is `literal`:

```
    QuotedLiteralSegment=OneOf(
        NamedParser("single_quote", CodeSegment, name="quoted_literal", type="literal"),
        NamedParser("double_quote", CodeSegment, name="quoted_literal", type="literal"),
    ),
```

What I'm suggesting is the following:

Add this around line 447 to set up a new Segment, with a new `name` (that won't be picked up by rule L048 since it only looks for `quoted_literal`):

```py
    # This is the same as QuotedLiteralSegment but is given a different type to stop L048 flagging
    SignedQuotedLiteralSegment=OneOf(
        NamedParser("single_quote", CodeSegment, name="signed_quoted_literal", type="literal"),
        NamedParser("double_quote", CodeSegment, name="signed_quoted_literal", type="literal"),
    ),
```

Then change your class to use that instead of SignedQuotedLiteralSegment:

```py
class IntervalLiteralSegment(BaseSegment):
    """

    """

    type = "interval_literal"

        Ref("SignedSegmentGrammar", optional=True),
        OneOf(
            Ref("NumericLiteralSegment"),
            Ref("SignedQuotedLiteralSegment"),
        ),
        Ref("DatetimeUnitSegment"),
        Ref.keyword("TO", optional=True),
        Ref("DatetimeUnitSegment", optional=True),
    )
```

R7L208:
That makes sense. Thank you for the clarification @tunetheweb.

Will push final commits shortly

tunetheweb:
> Will push final commits shortly

Let's also revert the rule changes fully. Hoping you won't need the `no cover`s now you're not changing the rules anymore.

src/sqlfluff/rules/L006.py
```diff
@@ -64,7 +64,7 @@ def _missing_whitespace(seg, before=True):
         if seg.name.endswith("_bracket"):
             if seg.name.startswith("start_" if before else "end_"):
                 return False
-        if seg.is_meta:  # pragma: no cover
+        if seg.is_meta:
```
tunetheweb, line 67: This wasn’t one you added so let’s revert that.

```suggestion
        if seg.is_meta:   # pragma: no cover
```

src/sqlfluff/rules/L048.py
```diff
@@ -3,6 +3,7 @@
 from typing import Tuple, List
 
 from sqlfluff.core.parser import BaseSegment
+
```
tunetheweb, line 6: Let’s revert

```suggestion
```

test/fixtures/dialects/sparksql/literals.sql
```diff
@@ -0,0 +1,63 @@
...
+SELECT TRUE AS col;
+SELECT FALSE AS col;
+
+-- numeric literal
+SELECT -2147483648 AS col;
+SELECT 9223372036854775807l AS col;
+SELECT -32Y AS col;
+SELECT 482S AS col;
+
+
```
tunetheweb, line 22: Nit, only double new line

```suggestion
```

tunetheweb (changes requested):

src/sqlfluff/rules/L006.py
```diff
@@ -64,7 +64,7 @@ def _missing_whitespace(seg, before=True):
         if seg.name.endswith("_bracket"):
             if seg.name.startswith("start_" if before else "end_"):
                 return False
-        if seg.is_meta:  # pragma: no cover
+        if seg.is_meta:   # pragma: no cover
```
tunetheweb, line 67:
```suggestion
        if seg.is_meta:  # pragma: no cover
```

tunetheweb (approved):

Issue #2426, opened 2022-01-22
Add parameter to make subset of AnyNumberOf() or AnySetOf() elements required


chwiese:
There can be situations where statement parameters can be in arbitrary order, but they are still required. One example would be Snowflakes [CREATE EXTERNAL TABLE](https://docs.snowflake.com/en/sql-reference/sql/create-external-table.html) statement, where `LOCATION` and `FILE_FORMAT` are required, but can be in arbitrary order with the other parameters, which are optional.

A parameter to make an element in a `AnyNumberOf()` or `AnySetOf()` required would address these situations. See #2421 for reference.

PR #3104, opened 2022-04-15
Snowflake: Add `CREATE FILE FORMAT` Support

jmc-bbk:
`CREATE FILE FORMAT ...` is currently unparsable as it is defined in `CreateStatementSegment()` which does not include parameters for file formats (it currently handles Warehouses, Pipes, and soon Integrations #3075).

An example of unparsable code:

```
CREATE OR REPLACE FILE FORMAT my_file_format
    TYPE = CSV
    SKIP_HEADER = 1
    COMPRESSION = NONE
    FIELD_DELIMITER = '|'
    DATE_FORMAT = 'DD/MM/YYYY'
;
```

In this PR, I want to introduce `CreateFileFormatSegment()` so that we can:

- Begin to breakdown `CreateStatementSegment()` into it's component parts.
- Provide better support for optional parameters for file formats (e.g. `CSV, JSON, ARVO`).

I have made some good progress, and the above would now be parsable. But I want to get advice from the maintainers before I continue building out this pattern. I will also post on the Slack channel for some help.

One big issue that I'm unsure on how to resolve is that both comma separation and no comma separation are valid syntax:

```
-- No comma separation.
CREATE OR REPLACE FILE FORMAT my_file_format
    TYPE = CSV
    SKIP_HEADER = 1

-- Comma separation.
CREATE OR REPLACE FILE FORMAT my_file_format
    TYPE = CSV,
    SKIP_HEADER = 1,
```

This PR currently only handles for no comma separation.

#### To do

- [ ] Handle for both comma and non-comma separated sets.
- [ ] Update other similar classes (e.g. `FileFormatSegment`) to reduce duplication / redundant code.
- [ ] Add test cases and a specific `create_file_format.sql` file.
- [ ] Determine whether the above pattern is correct (or too specific) - and make changes.

### Pull Request checklist

  - Other.


tunetheweb:
> One big issue that I'm unsure on how to resolve is that both comma separation and no comma separation are valid syntax:

We don't currently have an `OptionallyDelimted` type, and don't think it's worth introducing for this.

Instead what I'd suggest something like this:

```py
        OneOf(
            Delimited(
                Ref("CsvFileFormatTypeOptions"),  # TODO: Add Other File Formats.
             ),
            AnySetOf(
                Ref("CsvFileFormatTypeOptions"),  # TODO: Add Other File Formats.
            ),
        ),
```

jmc-bbk:
Starting to go a bit crazy with this one. Here's some examples of "valid" syntax on Snowflake:

```
-- Repeated arguments, some delimited
CREATE OR REPLACE FILE FORMAT my_file_format
    TYPE = CSV
    TYPE = CSV,
    TYPE = CSV,
    TYPE = CSV
;
```

```
-- No type specified (valid in CREATE OR REPLACE but not elsewhere)
CREATE OR REPLACE FILE FORMAT my_file_format
    SKIP_HEADER = 1,
    RECORD_DELIMITER = '|'
;
```

```
-- Duplicated type, mixed delimited, TYPE succeeds optional parameters
CREATE or replace EXTERNAL TABLE dev_test
LOCATION = @database.schema.my_stage
FILE_FORMAT = (SKIP_HEADER = 1 TYPE = CSV, TYPE = CSV);
```

And then an example of "not valid"

```
-- No type specified (note this is acceptable in create or replace)
CREATE or replace EXTERNAL TABLE dev_test
LOCATION = @database.schema.my_stage
FILE_FORMAT = (SKIP_HEADER = 1);
```

🤷‍♂️

tunetheweb:
I suggest you get the basic working. Can always enhance it later.

jmc-bbk:
I've added in the remaining file types, so made a bit more progress. But think it would be helpful for your opinions at this stage @tunetheweb !

The code can now handle:

- Delimited arguments `a = 1, b = 2, c = 3`
- Non-delimited arguments `a = 1, b = 2, c = 3`

But I have deliberately ignored mixed arguments `a = 1 b = 2, c = 3`, and these would cause a parse error.

Another thing I can't currently handle is:

```
-- TYPE coming after optional file parameters (which is valid syntax in CREATE OR REPLACE).

CREATE OR REPLACE my_file_format
  SKIP_HEADER = 1
  TYPE = CSV 
```

The above 👆 could be solved by moving type into the file optional parameters but this feels wrong, as it isn't really an optional parameter.

Moving into into the optional parameters, would mean that we would parse this "invalid" syntax where `TYPE` is required:

```
Example (1)

-- No type specified (note this is acceptable in CREATE OR REPLACE)
CREATE or replace EXTERNAL TABLE dev_test
LOCATION = @database.schema.my_stage
FILE_FORMAT = (SKIP_HEADER = 1);
```

Last thing!

We can't currently handle `TYPE` being anywhere in the parameters, as shown by this snippet in `class FileFormatSegment(BaseSegment):` that expects TYPE to come first.

```
Example (2)

                Sequence(
                    "TYPE",
                    Ref("EqualsSegment"),
                    Ref("FileType"),
                    # formatTypeOptions - To Do to make this more specific
                    Ref("FormatTypeOptionsSegment", optional=True),
                ),
```

If we did move type into the optional parameters I've created here, we would resolve the issue of TYPE appearing anywhere - but introduce the issue that we parse code without `TYPE` (see `Example 1`).

Sorry that's really long to explain - Snowflake's parser makes really weird decisions in this area.

tunetheweb:
I would move it in to the params. SQLFluff syntax is not expected to 100% match the actual dialect - in part because we can't always match.

#2426 is a suggestion to add an ability to make an option mandatory, but that's not be implemented yet.

jmc-bbk:
Going to wait to see if the above tests pass, then I will edit this comment to explain some key things following latest changes.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -144,6 +144,14 @@
         name="warehouse_size",
         type="warehouse_size",
     ),
+    # We use a RegexParser instead of keywords as the arguments are optionally quoted.
+    CompressionType=RegexParser(
+        r"'?AUTO'?|'?GZIP'?|'?BZ2'?|'?BROTLI'?|'?ZSTD'?|'?DEFLATE'?|'?RAW_DEFLATE'?|"
+        r"'?LZO'?|'?NONE'?|'?SNAPPY'?",
```
tunetheweb, line 150: Slightly more efficient and easier to read regex:

```suggestion
        r"'?(AUTO|GZIP|BZ2|BROTLI|ZSTD|DEFLATE|RAW_DEFLATE|LZO|NONE|SNAPPY)'?",
```

The down side to this is it will accept unbalanced quotes (e.g. `'ANSI`).

To avoid this, you could have this as a `OneOf`:

```py
    CompressionType=OneOf(
         RegexParser( r"'(AUTO|GZIP|...)'",....
         RegexParser( r"(AUTO|GZIP|...)",....
    ),
```

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -2813,6 +2822,372 @@ class AlterMaterializedViewStatementSegment(BaseSegment):
...
+            Ref("JsonFileFormatTypeParameters"),
+            Ref("AvroFileFormatTypeParameters"),
+            Ref("OrcFileFormatTypeParameters"),
+            Ref("ParquetFileFormatTypeParameters"),
+            Ref("XmlFileFormatTypeParameters"),
+        ),
+        Sequence(
+            # Use a Sequence and include an optional CommaSegment here.
+            # This allows a preceding comma when above parameters are delimited.
+            Ref("CommaSegment", optional=True),
```
tunetheweb, line 2854: Wouldn't you be better to add `allow_trailing=True` to the `Delimited` sections of each `XXXTypeParameters`?

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -144,6 +144,14 @@
         name="warehouse_size",
         type="warehouse_size",
     ),
+    # We use a RegexParser instead of keywords as the arguments are optionally quoted.
+    CompressionType=RegexParser(
+        r"'?AUTO'?|'?GZIP'?|'?BZ2'?|'?BROTLI'?|'?ZSTD'?|'?DEFLATE'?|'?RAW_DEFLATE'?|"
+        r"'?LZO'?|'?NONE'?|'?SNAPPY'?",
+        CodeSegment,
+        name="compression_type",
+        type="compression_type",
```
tunetheweb, line 153: Giving these separate `types` and `names` has positives and negatives.

On one hand, it's good to classify them separately when you can. You could then potentially treat them differently.

However it also means it won't be caught for Keyword capitalisation rules (e.g. L010).

If you want them included in that you could do it like this:

```suggestion
        name="compression_type",
        type="keyword",
```

This also allows you to target it separately (using `name`) but also treat them the same (using `type`) sometimes.

I'm in two minds whether to leave as is, or go with this. As I say pluses and minuses.

WDYT? Are they keywords? Should they be?

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -2813,6 +2822,372 @@ class AlterMaterializedViewStatementSegment(BaseSegment):
...
+    _file_format_type_parameter = OneOf(
+        Sequence(
+            "TYPE",
+            Ref("EqualsSegment"),
+            RegexParser(
+                r"'?CSV'?",
+                CodeSegment,
+                name="csv_file_type",
+                type="file_type",
+            ),
```
tunetheweb, line 2878: Again a `OneOf` would prevent unbalanced quotes. And would allow you to use the quick `StringParser`:

```suggestion
            OneOf(
                StringParser(
                    r"'CSV'",
                    CodeSegment,
                    name="csv_file_type",
                    type="file_type",
                )
                StringParser(
                    r"CSV",
                    CodeSegment,
                    name="csv_file_type",
                    type="file_type",
                )
            ),
```

Same for the rest of the TypeParameters.

Also not convinced we need `name="csv_file_type",`, versus the more generic `name="file_type",`. So we really need a special type for each type? Isn't that obvious from the contents?

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_snowflake.py
WittierDinosaur: I'd probably suggest exploding this out to all of the options in StringParser. I dislike RegexParser, as it's really not performant - and I'd make the case for not using it anywhere

tunetheweb: In thought you might say this, cause I know you've raised this before. 😁

But is it really that badly performant? Especially for simple regexes like this? Realistically we're not going to get away from the RegexParser for lots of things and do think the StringParser leads to a lot more verbose code for things like this.

Btw, think I've spotted a small optimisation there so will open a separate PR and send to you to review.

tunetheweb: Actually ignore that last comment. I don't have an optimisation after all.

jmc-bbk:
Hopefully quick question @tunetheweb and @WittierDinosaur - and I will try make suggested changes today (or ASAP).

Should we:

(a) leave the `file_type` in the individual classes (e.g. `CsvFileFormatTypeParameters`) but with the `StringParser()`?

(b) change this pattern to align with `CompressionType` and have something like `FileType =...Parser('(CSV|JSON|AVRO...` and then have a single `FileFormatParametersSegment` class?

Advantages of (a):

- Can make changes to individual file format classes without affecting the others
- Parse tree will recognise `csv_file_format_parameters`

Advantages of (b):

- Generalisable class handles all parameters
- Changes affect all file formats / non-specific parse tree

---

The reason why I want to make sure, is there is the exact same argument for `CompressionType` - but at the moment I'm handling that in the style of (b) rather than (a).

The reason I went with the generalisable compression type is that:

- CSV/JSON/AVRO/XML all handle the same compression types
- PARQUET has a different set of compression types

I thought it starts to get messy if I have a different `CompressionType` class for each file format, or have two that handle the different formats above - but unsure on this.

---

Anyway! Let me know whether to go for method (a) or (b) for the `FileFormat` and I will try wrap this up ASAP.

tunetheweb:
I think if you're having different sections for each (which you have currently done), then it makes more sense to further restrict to the actual type, rather than the whole group. I think `StringParser` is better to prevent `'CSV` or `CSV'` being allowed as per my comment above.

For the over Compression Type I think two `RegexParser`s are the way to go one with quotes, one without, wrapped in a `OneOf()`. But know @WittierDinosaur has some concerns on `RegexParser`s but did ask some questions above as to how valid those concerns are for this specific case.

The other alternative is to generalise this into one big statement. I.e. rather than having separate `CsvFileFormatTypeParameters`, `JsonFileFormatTypeParameters`...etc. grammars could just have a more generic `FileFormatTypeParameters` that accepts any of those formats for any of the types. Technically less accurate but could be a lot less code. But then you've done the work now, so might as well keep the specific ones.

src/sqlfluff/dialects/dialect_snowflake.py
jmc-bbk: Should be done!

src/sqlfluff/dialects/dialect_snowflake.py
jmc-bbk: I spoke to @WittierDinosaur and he also thought `keyword`, done!

src/sqlfluff/dialects/dialect_snowflake.py
jmc-bbk: Gone with your suggestion 👍

src/sqlfluff/dialects/dialect_snowflake.py
jmc-bbk: I've left this as it is -

the reason being is that when these segments are used elsewhere (e.g. `ExternalTableSegment`) a trailing comma isn't valid syntax.

AFAIK it is only the `CREATE FILE FORMAT` segment where the trailing comma could be valid - and then, it's only valid when there is a `COMMENT`. I think that makes sense to group it with the `COMMENT` sequence in this case!

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -235,6 +250,13 @@
         name="file_type",
         type="file_type",
     ),
+    IntegerSegment=RegexParser(
+        # An unquoted integer that can be passed as an argument to Snowflake functions.
+        r"[0-9]+",
+        CodeSegment,
+        name="integer_segment",
+        type="integer_segment",
+    ),
```
tunetheweb, line 259:
```suggestion
    IntegerSegment=RegexParser(
        # An unquoted integer that can be passed as an argument to Snowflake functions.
        r"[0-9]+",
        CodeSegment,
        name="integer_literal",
        type="literal",
    ),
```

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -2813,59 +2829,435 @@ class AlterMaterializedViewStatementSegment(BaseSegment):
...
+
+    type = "csv_file_format_type_parameters"
+
+    _file_format_type_parameter = OneOf(
+        Sequence(
+            "TYPE",
+            Ref("EqualsSegment"),
+            OneOf(
+                StringParser(
+                    r"'CSV'",
```
tunetheweb, line 2882: Personally I'd remove a the `r""` syntax from StringParser instances as slightly easier to read, and we're not using any backquotes:

```suggestion
                    "'CSV'",
```

Can you apply this to the other StringParser instances you added?

tunetheweb (changes requested):

tunetheweb (approved):

PR #3105, opened 2022-04-15
Don't allow fixes to span template blocks

### LINKED ISSUE #3079: L052 + Jinja: Autofix breaks templated SQL if `;` is outside the template
### Search before asking

### What Happened

I ran into a weird case where the autofix of `L052` produces an invalid SQL. Arguably the SQL template in this example is not very well written, but I don't think broken SQL as result of autofix should be acceptable.

### Expected Behaviour

The result of autofix should be a valid SQL.

### Observed Behaviour

```sql
{% set some_variable = (1 == 1) %}

{% if some_variable %}
SELECT 1;{% else %}
SELECT 2
{% endif %}
;
```

Which is broken as `SELECT 1;;` fails when running.

### How to reproduce

```sql
{% set some_variable = (1 == 1) %}

{% if some_variable %}
SELECT 1
{% else %}
SELECT 2
{% endif %}
;
```

https://online.sqlfluff.com/fluffed?sql=H4sIAO2EVmIC_6tWVShOLVEozs9NjS9LLMpMTMpJVbBV0DBUsLVVMNRUUK3l5eLlqlZVyExDUwSSCXb1cXUOUTAEq0jNKUYRNYKI5qUAtYKErQHzPtgYbAAAAA%3D%3D&dialect=ansi

### Dialect

ansi

### Version

0.12 -> webversion

### Configuration

default

### Code of Conduct

barrywhart:
Fixes #3079

A `LintResult` whose fixes span multiple template blocks (where a "block" is where a Jinja `if` or `for` starts or ends) should be discarded. That's because one slice may be executed separately from the other at template rendering time, so a fix touching both may corrupt the code.

This new check is similar in spirit to an older one I removed a few months ago. The old one prevented a fix from touching 3 or more slices. It was more of a "rule of thumb", and there were no tests for it. This check, while similar, is more formal, and I've added a test for it.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -258,7 +258,9 @@ class LintFix:
         """Create edit segments after the supplied anchor segment."""
         return cls("create_after", anchor_segment, edit_segments, source)
 
-    def get_fix_slices(self, templated_file: TemplatedFile) -> Set[RawFileSlice]:
+    def get_fix_slices(
+        self, templated_file: TemplatedFile, within_only: bool
+    ) -> Set[RawFileSlice]:
         """Returns slices touched by the fix."""
         # Goal: Find the raw slices touched by the fix. Two cases, based on
         # edit type:
@@ -271,19 +273,21 @@ class LintFix:
         anchor_slice = self.anchor.pos_marker.templated_slice
         templated_slices = [anchor_slice]
 
+        # If "within_only" is set for a "create_*" fix, the slice should only
+        # include the area of code "within" the area of insertion, not the other
+        # side.
+        adjust_boundary = 1 if not within_only else 0
         if self.edit_type == "create_before":
             # Consider the first position of the anchor segment and the
             # position just before it.
             templated_slices = [
-                slice(anchor_slice.start, anchor_slice.start + 1),
-                slice(anchor_slice.start - 1, anchor_slice.start),
+                slice(anchor_slice.start - 1, anchor_slice.start + adjust_boundary),
             ]
         elif self.edit_type == "create_after":
             # Consider the last position of the anchor segment and the
             # character just after it.
             templated_slices = [
-                slice(anchor_slice.stop - 1, anchor_slice.stop),
-                slice(anchor_slice.stop, anchor_slice.stop + 1),
+                slice(anchor_slice.stop - adjust_boundary, anchor_slice.stop + 1),
             ]
         # TRICKY: For creations at the end of the file, there won't be an
         # existing slice. In this case, the function adds file_end_slice to the
@@ -803,7 +807,7 @@ class BaseRule:
         target_tuples: List[Tuple[str, str]],
         parent: Optional[BaseSegment] = None,
     ):
-        """Does the given segment match any of the given type tuples."""
+        """Does the given segment match any of the given type tuples?"""
         if seg.name in [elem[1] for elem in target_tuples if elem[0] == "name"]:
             return True
         elif seg.is_type(*[elem[1] for elem in target_tuples if elem[0] == "type"]):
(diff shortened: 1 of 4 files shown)


src/sqlfluff/core/rules/base.py
```diff
@@ -258,51 +258,55 @@ def create_after(
         """Create edit segments after the supplied anchor segment."""
         return cls("create_after", anchor_segment, edit_segments, source)
 
-    def has_template_conflicts(self, templated_file: TemplatedFile) -> bool:
```
barrywhart, line 261: The diffs below may be easier to read if you keep in mind that I split `has_template_conflicts()` into two functions:
* `get_fix_slices()`
* `has_template_conflicts()`

src/sqlfluff/core/rules/base.py
```diff
@@ -258,51 +258,55 @@ def create_after(
...
+        # include the area of code "within" the area of insertion, not the other
+        # side.
+        adjust_boundary = 1 if not within_only else 0
         if self.edit_type == "create_before":
             # Consider the first position of the anchor segment and the
             # position just before it.
             templated_slices = [
-                slice(anchor_slice.start, anchor_slice.start + 1),
-                slice(anchor_slice.start - 1, anchor_slice.start),
+                slice(anchor_slice.start - 1, anchor_slice.start + adjust_boundary),
```
barrywhart, line 284: I made a small refactoring here that doesn't affect correctness, but is a bit more efficient. The old code created an array of two adjacent slices. The new code creates a single slice covering the same range.

src/sqlfluff/core/rules/base.py
```diff
@@ -258,51 +258,55 @@ def create_after(
...
             ]
-            check_fn = all
         elif self.edit_type == "create_after":
             # Consider the last position of the anchor segment and the
             # character just after it.
             templated_slices = [
-                slice(anchor_slice.stop - 1, anchor_slice.stop),
-                slice(anchor_slice.stop, anchor_slice.stop + 1),
+                slice(anchor_slice.stop - adjust_boundary, anchor_slice.stop + 1),
```
barrywhart, line 290: Ditto the above comment.

src/sqlfluff/core/rules/base.py
```diff
@@ -258,51 +258,55 @@ def create_after(
...
         # a fix to *templated* code and incorrectly discarded.
-        fix_slices = self._raw_slices_from_templated_slices(
+        return self._raw_slices_from_templated_slices(
             templated_file,
             templated_slices,
             file_end_slice=RawFileSlice("", "literal", -1),
         )
 
-        # We have the fix slices. Now check for conflicts.
+    def has_template_conflicts(
```
barrywhart, line 304: No changes in behavior for `has_template_conflicts()`, just split out `get_fix_slices()` so it could be used elsewhere.

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -361,16 +367,19 @@ def analyze(self, make_template: Callable[[str], Template]) -> JinjaTracer:
                             block_type,
                             self.idx_raw,
                             block_subtype,
+                            block_idx,
                         )
                     )
                     self.raw_slice_info[self.raw_sliced[-1]] = raw_slice_info
-                    block_idx = len(self.raw_sliced) - 1
+                    slice_idx = len(self.raw_sliced) - 1
```
barrywhart, line 374: Renamed the existing `block_idx` variable to `slice_idx` because that's the more correct name. This allowed me to create a new `block_idx` variable in this PR, with different meaning.

tunetheweb (approved):

Issue #3106, opened 2022-04-16
Add native Github annotations output


dmohns:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

Currently SQLFluff's `github-annotation` output is a custom JSON output tailored for this https://github.com/yuzutech/annotations-action Github Action.

Reading through the related issue and PR it's not really clear to my what is the advantage of using this additional Github Action over just writing native Github Action annotation output.

A good example is yamllint's `-f github` option. 

```sh
❯ yamllint 1.yaml -f github
::error file=1.yaml,line=3,col=3::syntax error: expected <block end>, but found '<block mapping start>' (syntax)
```

This can be integrated into Github actions extremely easy.

### Use case

Simplify Github Actions workflows be requiring one less action.

Additionally: I haven't tested it a lot, but currently it looks like it's not possible to have the CI report failure _and_ post annotations. For the annotations action to work, we need to run SQLFluff with `--nofail`.

### Dialect

all

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

dmohns:
Initial support for this was added in https://github.com/sqlfluff/sqlfluff/pull/3107 via `github-annotation-native` output format.

One question remained open, so I'm taking it here for visibility:

Do wo we want to continue to maintain both output formats `github-annotation` and `github-annotation-native`?

barrywhart:
@emilyhawkins-drizly: If I recall correctly, you werethe first user of the earlier annotation format. Any thoughts on whether to keep it?

Personally, I'd like to depreciate and eventually remove it, but I'm not adamant about it.

tunetheweb:
Let's close this and track as separate issue as the issue title has been addressed.

PR #3107, opened 2022-04-16
Add native Github-actions output

dmohns:

Makes progress on #3106

This is (for now) a PoC for adding native Github Actions annotation capabilities. The code is essentially copy/pasted from `yamllint` github action implementation and added to SQLFluff.

Here is a small demo how this would look like when using in Github Actions: https://github.com/dmohns/sqlfluff-test-github-action-annotation/pull/1/files

Before continuing to work on that I would like to understand if this feature has a place in SQLFluff.

### Open questions 
If we continue to work on this, there is a few questions to be considered:

- Do we continue to maintain both "github-annotation" output formats?
  - Option 1: Replace "old" `github-annotation` with github-annotation native. This will probably break existing pipelines
  - Option 2: Add a new `github-annotation-native` format for now, mark `github-annotation` as deprecated. As part of the next major release flip.
  - Option 3: Maintain both

### Todo

- [x] Add tests
- [x] Update Documentation and examples

### Pull Request checklist

  - Other.

### DIFF
--- a/test/cli/commands_test.py
+++ b/test/cli/commands_test.py
@@ -967,7 +967,9 @@ def test__cli__command_fail_nice_not_found(command):
     assert "could not be accessed" in result.output
 
 
-@pytest.mark.parametrize("serialize", ["yaml", "json", "github-annotation"])
+@pytest.mark.parametrize(
+    "serialize", ["yaml", "json", "github-annotation", "github-annotation-native"]
+)
 @pytest.mark.parametrize("write_file", [None, "outfile"])
 def test__cli__command_lint_serialize_multiple_files(serialize, write_file, tmp_path):
     """Check the general format of JSON output for multiple files.
@@ -1012,6 +1014,12 @@ def test__cli__command_lint_serialize_multiple_files(serialize, write_file, tmp_
         result = json.loads(result_payload)
         filepaths = {r["file"] for r in result}
         assert len(filepaths) == 1
+    elif serialize == "github-annotation-native":
+        result = result_payload.split("\n")
+        # SQLFluff produces trailing newline
+        if result[-1] == "":
+            del result[-1]
+        assert len(result) == 24
     else:
         raise Exception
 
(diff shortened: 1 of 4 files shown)


tunetheweb:
This is great! Just need to add a test for it.

test/cli/commands_test.py
```diff
@@ -1012,6 +1014,12 @@ def test__cli__command_lint_serialize_multiple_files(serialize, write_file, tmp_
         result = json.loads(result_payload)
         filepaths = {r["file"] for r in result}
         assert len(filepaths) == 1
+    elif serialize == "github-annotation-native":
+        result = result_payload.split("\n")
+        # SQLFluff produces trailing newline
+        if result[-1] == "":
+            del result[-1]
+        assert len(result) == 24
```
tunetheweb, line 1022: What is this here for? Shouldn't it be variable based on the number of results?

tunetheweb (commented):
This is great!

One question on the code.

Also would it be possible to set this in config file? Might allow it to be set on by default for things like the GitHub Super Linter and MegaLinter.

Would also be good if @barrywhart could review from a code perspective. But it looks good to me.

test/cli/commands_test.py
dmohns: I wasn't entirely sure either. But the trailing new line behaves differently between "write to file" and reading the raw output. The raw output has trailing newline while the file has not.

However, looking at the code again, I might have misunderstood what this test is trying to ... test 😂

tunetheweb: I'm OK with removing the trailing line. I'm more questioning why it should be exactly 24 lines long?

tunetheweb:
> Also would it be possible to set this in config file? Might allow it to be set on by default for things like the GitHub Super Linter and MegaLinter.

Actually I'm less convinced about this, the more I think about it. Should just be a command line option used by those. Otherwise it encourages different .sqlfluff config files for CI and for normal used, which is probably not a good idea.

test/cli/commands_test.py
dmohns: The input file throws 12 linting errors. So, with 24 we are checking that the file actually gets linted twice. But I will try to refactor this a bit to make it more intuitive

barrywhart: This test lints the same input file into various output formats. So this line is only executed once, when the file is linted into `"github-annotation-native"` format. (It's probably confusing because here we're using `parametrize` to vary the _format_ rather than the _input file_.

barrywhart (approved):
Looks good to me!

test/cli/commands_test.py
tunetheweb: Ah I missed this was a test. Ok hard coding the expected results makes total sense then. But yeah maybe a brief investigation to see why it’s doing it twice and if that can be avoided and, if not, a comment to explain.

barrywhart: Oh, just noticed the part about twice. That's really weird, and worth investigating. Does the same thing happen if you run the same command manually (from the terminal)?

barrywhart: I was working on this same test for another PR (#3109) and noticed that it is specifically designed to process the same file twice. Kinda weird test, but nothing wrong with this PR!

See these lines in the test:
```
    # note the file is in here twice. two files = two payloads.
    result = invoke_assert_code(
        args=[lint, cmd_args],
        ret_code=65,
    )
```

PR #3112, opened 2022-04-18
TSQL: Exclude non-function-name keywords from function names

jpers36:
The current definition for function names is a parser function and allows for anything that matches a simple regex definition.  This does not apply any sort of check to the name with regard to keywords.  As a result, I've come across a situation where a FROM clause is being interpreted as a function and unable to be parsed.

This PR would change the function name identifier definition to include an anti_template which excludes all keywords other than those used for built-in functions.

### Pull Request checklist


tunetheweb:
>  As a result, I've come across a situation where a FROM clause is being interpreted as a function and unable to be parsed.

Could you provide this example? Just so we fully understand it, and whether there's any other way of solving this.

jpers36:
@tunetheweb sure. The example code has been added to the test YML in this branch. I'm thinking about alternative approaches, but maybe won't be able to test anything until Wednesday.

test/fixtures/dialects/tsql/functions_a.sql
```diff
@@ -24,3 +24,24 @@ CREATE FUNCTION dbo.no_paramters() RETURNS INT AS
...
     RETURN 2;
 END
+
+GO
+
+/*
+SQLFluff should parse this FROM as a keyword and not as a function name
+
+*/
+SELECT a.*
```
jpers36, line 34: @tunetheweb

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -256,12 +256,6 @@
         Ref("DatatypeSegment"),
         Sequence(Ref("EqualsSegment"), Ref("ExpressionSegment"), optional=True),
     ),
-    FunctionNameIdentifierSegment=RegexParser(
-        r"[A-Z][A-Z0-9_]*|\[[A-Z][A-Z0-9_]*\]",
-        CodeSegment,
-        name="function_name_identifier",
-        type="function_name_identifier",
-    ),
```
tunetheweb, line 264: Would ljst adding an exclude list here work instead, similar to how we define`NakedIdentifierSegment` and `DatatypeIdentifierSegment`?

```py
    FunctionNameIdentifierSegment=SegmentGenerator(
        # Generate the anti template from the set of reserved keywords
        # minus the function names that are reserved words.
        lambda dialect: RegexParser(
            r"[A-Z][A-Z0-9_]*|\[[A-Z][A-Z0-9_]*\]",
            CodeSegment,
            name="function_name_identifier",
            type="function_name_identifier",
            anti_template=r"^("
            + r"|".join(
                dialect.sets("reserved_keywords")
                - {
                    "COALESCE",
                    "CONVERT",
                    "CURRENT_TIMESTAMP",
                    "CURRENT_USER",
                    "LEFT",
                    "NULLIF",
                    "RIGHT",
                    "SESSION_USER",
                    "SYSTEM_USER",
                }
            )
            + r")$",
        )
    ),
```

This would allow us to keep the parse tree the same as other dialects rather than introducing another level of Identifier or Keyword. That would also result in a lot less differences.

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_tsql.py
jpers36: @tunetheweb that's the direction I was thinking once I hit the critical error, but I haven't worked with the anti_template pattern yet.  I'll try it out!

tunetheweb: Looks much better and cleaner now! Hopefully tests will all pass too (they should IMHO)

tunetheweb:
BTW see you've included a load of `.vs` files? Could you remove them? Not sure what they are but happy to add to our `.gitignore` to avoid you accidentally excluding them in future if that's useful?

jpers36:
@tunetheweb looks good!

PR #3118, opened 2022-04-19
New Rule L064: Consistent usage of preferred quotes for quoted literals

### LINKED ISSUE #3090: Rule suggestion: Add rule for consistent STRING quotes for dialect that allow multiple quote types
### Search before asking

### Description

Some dialects allow STRINGs to be quoted using different kind of quotes. Most of the times teams seem to have a tendency towards one style of quotes and mights wish to specify this as a part of their SQLFluff rule set.

The rule could have a configuration option `string_quotes` that takes values `consistent`, `single_quotes` or `double_quotes`.

### Use case

From what I can see this affects `mysql` and `bigquery`. In both of which the following is a valid query

```sql
SELECT
  'abc' AS col1,
  "abc" AS col2,
  'abc' = "abc" AS col3
```

and could be linted to 

```sql
SELECT
  'abc' AS col1,
  'abc' AS col2,
  'abc' = 'abc' AS col3
```

in the case of `consistent` or `single_quotes`.

And linted to 

```sql
SELECT
  "abc" AS col1,
  "abc" AS col2,
  "abc" = "abc" AS col3
```

in the case of `double_quotes`.

### Dialect

bigquery, mysql

### Code of Conduct

dmohns:

Fixes #3090

This PR is an idea to implement a rule for consistent usage of preferred quotes for quoted literals (note NOT quoted identifiers as they have different constraints).

The implementation is essentially copy/pasted from Black https://github.com/psf/black/blob/main/src/black/strings.py and adapted to our use-case

- Added more logging output
- Allow passing of `preferred_quote_char` and `alternate_quote_char` (Black hardcodes to `"`)
- Allow exiting the rule when neither single nor double quotes are found. (e.g. when Dollar quotes are used to define literals)

New rule might trigger on existing SQL code. But this should be always the case for new rules.

### Pull Request checklist

  - Other.


tunetheweb:
> While working on this I realised that an option consistent (which has the original idea in the ticket) is more constraining than helpful for the end user. We can copy all of Black's nice work (reducing quotes on STRINGs etc..) when using notion of "preferred" quotes and "alternate" quotes. This is what is implemented here.

The main benefit IMHO of having a `consistent` option, and making it the default, is that it allows rolling out this rule with relatively little risk of over alerting or enforcing a style that a user may not want.

If a user generally preferred single quotes, then `consistent` would ensure there is not mixed quotes within the same SQL query (but admittedly not across different SQL queries). Same if the user generally used double quotes.

Without that, the user must accept the default of double-quotes, or must change that default. Which potentially makes this a breaking change for existing SQLFluff users.

So I'd still vote to add `consistent` for that, and also to make it consistent (pun intended!) with the other rules that enforce one style over another.

WDYT @barrywhart ?

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,234 @@
...
+                                name="quoted_literal",
+                                type="literal",
+                            )
+                        ],
+                    )
+                ],
+                description=(
+                    "Inconsistent use of preferred quote style '"
+                    f"{QUOTES_MAPPING[self.preferred_string_quotes]['common_name']}'. "
+                    f"Use {fixed_string} instead of {context.segment.raw}."
```
barrywhart, line 128: Can we omit including the actual string in the lint message? Reasons:
* The string could be quite long.
* In some cases, the string may contain sensitive information that shouldn't appear in a build log.

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,234 @@
...
+        unescaped_new_quote = regex.compile(rf"(([^\\]|^)(\\\\)*){new_quote}")
+        escaped_new_quote = regex.compile(rf"([^\\]|^)\\((?:\\\\)*){new_quote}")
+        escaped_orig_quote = regex.compile(rf"([^\\]|^)\\((?:\\\\)*){orig_quote}")
+        body = s[first_quote_pos + len(orig_quote) : -len(orig_quote)]
+
+        if "r" in prefix.casefold():
+            if unescaped_new_quote.search(body):
+                # There's at least one unescaped new_quote in this raw string
+                # so converting is impossible
+                # self.logger.debug()
```
barrywhart, line 191: Can this line be removed?

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,234 @@
...
+        if first_quote_pos == -1:
+            return s  # There's an internal error
+
+        prefix = s[:first_quote_pos]
+        unescaped_new_quote = regex.compile(rf"(([^\\]|^)(\\\\)*){new_quote}")
+        escaped_new_quote = regex.compile(rf"([^\\]|^)\\((?:\\\\)*){new_quote}")
+        escaped_orig_quote = regex.compile(rf"([^\\]|^)\\((?:\\\\)*){orig_quote}")
+        body = s[first_quote_pos + len(orig_quote) : -len(orig_quote)]
+
+        if "r" in prefix.casefold():
```
barrywhart, line 187: Is there an advantage to using `casefold` here, given that `r` is a simple alphabetical character, not a fancy Unicode character?

barrywhart:
@tunetheweb: Re: `consistent`, it's a good point. We could also disable the rule by default for all dialects.

barrywhart (changes requested):

src/sqlfluff/rules/L064.py
dmohns: It's really just used for case sensitivity here. I.e. the raw prefix could be `r` or `R`. Let me just use `lower` 👍

tunetheweb:
> @tunetheweb: Re: consistent, it's a good point. We could also disable the rule by default for all dialects.

Well if you have `consistent` then you CAN enable it by default IMHO. Without `consistent`, it's a bit more opinionated so that makes enabling it by default more difficult.

But I don't think we have a good way of disabling by default? If you include in the default `exclude_rules`, that will simply be override by any user who uses that.

src/sqlfluff/rules/L064.py
tunetheweb: Quite a few other rules use this, and it can be handy.

Not sure I totally agree with the security concerns since this is raw text from SQL file.

dmohns:
Let me try to re-phrase 😂 I think `consistent` is incompatible with what the rule is currently doing. 

This rule works very well one way, i.e. given a quote style it can produce a well-defined linting result. However, the other way around is not that easy. For example `'"'` this _is_ actually both quotes styles. For single quotes it's just straight forward. However, it's also the preferred way of representing a single `"` in double quote style to reduce escapes.

Hence, to implement a `consistent` mode we would have to traverse the parse tree and check both quote style all the time until we have found (or guessed) the preferred quote style.

I think this would make the rule implementation unnecessarily complex for very little gain.

I see the point of trying to not be opinionated by default. I think disabling the rule by default is a better solution here then.

From a users perspective there are two cases

1. Users that **don't** care about quote style. Either because their dialect doesn't support double quotes, or because they simply don't care.
2. Users that **do** care about quote style and are willing to enforce it as part of their linting process. The users will have a quote style in mind and will need to configure the rule anyways.

I think for either case users will be fine if the rule is disabled by default.

src/sqlfluff/rules/L064.py
dmohns: In scenarios where the string contains a lot of quotes it might be very hard for the user to tell what is the "correct" way of quoting. So, when users only use the linter (for example as a plugin to their editor) I imagined it to be rather frustrating if we only tell them that there is a problem but not what the solution is.

I already removed this, let me know if I can revert this.

barrywhart:
When I mentioned the idea of disabling by default, I was thinking that it would "run" but not do anything unless you set `force_enable`. I don't think any current rules do this, but we could!

src/sqlfluff/rules/L064.py
barrywhart: I don't feel strongly about this, so it's good to add back. 👍

tunetheweb:
> When I mentioned the idea of disabling by default, I was thinking that it would "run" but not do anything unless you set `force_enable`. I don't think any current rules do this, but we could!

Interesting idea.

I wonder if we should choose another word other than `force_enable` so we don't have two different definitions of that. E.g. `fix = True`?

tunetheweb:
> Let me try to re-phrase 😂 I think `consistent` is incompatible with what the rule is currently doing.
> 
> This rule works very well one way, i.e. given a quote style it can produce a well-defined linting result. However, the other way around is not that easy. For example `'"'` this _is_ actually both quotes styles. For single quotes it's just straight forward. However, it's also the preferred way of representing a single `"` in double quote style to reduce escapes.
> 
> Hence, to implement a `consistent` mode we would have to traverse the parse tree and check both quote style all the time until we have found (or guessed) the preferred quote style.

We could just take the first quoted_literal and use the first character of that as the preferred style?

i.e. if `preferred_string_quotes ` is set to `consistent` (which would be the default) then, when we see the first `quoted_literal`, we set it to that, and then use `memory` to persist it for the rest of that Query's parsing.

> I think this would make the rule implementation unnecessarily complex for very little gain.

As I say the main "gain" is the ability to have this on by default, but limiting the impact for those that haven't made an explicitly default.

> I see the point of trying to not be opinionated by default. I think disabling the rule by default is a better solution here then.

As per above, we don't have an easy way of disabling by default. Should think about best way of doing this. @barrywhart 's suggestion to enable linting by default, but not fixing is an interesting idea that at least won't do mass updates to people's code, yet still surface the rule forcing people to make a choice of setting here. But still think that will lead to lots of false positives becuase we need an opinionated default - and we wouldn't see that really with `consistent`. Which would allow us to avoid the whole issue of having to disable by default.

I've also gone back and forth several times in my mind about whether disabling some by default is even a good idea - see #1812. Currently I'm thinking better to have them all enabled by default, hence why I'm arguing the case here a bit. Not sure this rule is big enough to force the decision here on something we've talked about, but never solved, for a bit now...
 
> From a users perspective there are two cases
> 
> 1. Users that **don't** care about quote style. Either because their dialect doesn't support double quotes, or because they simply don't care.
> 2. Users that **do** care about quote style and are willing to enforce it as part of their linting process. The users will have a quote style in mind and will need to configure the rule anyways.
> 
> I think for either case users will be fine if the rule is disabled by default.

As per above, so far our other rules are enabled by default, but (mostly) in the least disruptive way. So this would be a change from that past. Maybe we do need to finally solve that, instead of putting it off so much, but not 100% sure this is the rule that finally merits that...

Still mulling it over, but interested to hear both your opinions. And the rest of the @sqlfluff/sqlfluff-maintainers

dmohns:
It turns out that getting a `consistent` option is not as hard as I thought. So, let's go with this and **enable** this rule by default?

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,262 @@
...
+
+        if (
+            context.dialect.name not in self._dialects_with_double_quoted_strings
+            and not self.force_enable
+        ):
+            return LintResult(memory=context.memory)
+
+        # Only care about quoted literal segments.
+        if not context.segment.name == "quoted_literal":
+            return None
```
tunetheweb, line 107: Why only `quoted_literal` and not also `quoted_identifiers`?

Is it because in some dialects quoted_identifiers are no in single/double quotes (e.g. square brackets in TSQL, back ticks in BigQuery)? If so, I get that, but maybe want to make it more clear in the documentation as I (incorrectly!) thought it was literals and identifiers.

Also think BigQuery can be in-scope if it's just literals: https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical
And maybe also MySQL unless the default of ANSI_QUOTE is used?: https://dev.mysql.com/doc/refman/8.0/en/string-literals.html

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,262 @@
...
+
+        # If quoting style is set to consistent we use the quoting style of the first
+        # quoted_literal that we encounter.
+        if self.preferred_string_quotes == "consistent":
+            memory = context.memory
+            preferred_string_quotes = memory.get("preferred_string_quotes")
+
+            if not preferred_string_quotes:
+                # Getting the quote from LAST character to be able to handle STRING
+                # prefixes
```
tunetheweb, line 117: Nice!

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,262 @@
...
+        "alternate_quote_char": '"',
+    },
+    "double_quotes": {
+        "common_name": "double quotes",
+        "preferred_quote_char": '"',
+        "alternate_quote_char": "'",
+    },
+}
+# BigQuery string prefix characters.
+STRING_PREFIX_CHARS: Final = "rbRB"
```
tunetheweb, line 33: Should these be moved into the Rule_L064 class?

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,262 @@
...
+
+        fixed_string = self._normalize_preferred_string_quotes(
+            context.segment.raw,
+            preferred_quote_char=QUOTES_MAPPING[preferred_string_quotes][
+                "preferred_quote_char"
+            ],
+            alternate_quote_char=QUOTES_MAPPING[preferred_string_quotes][
+                "alternate_quote_char"
+            ],
+        )
```
tunetheweb, line 140: Slight performance nit, that you're calling this even for the first iteration (when preferred_quote is not set so will be whatever it is currently, so no need to fix).

Should you exit early above?

Or do you still need this for the case when the first quote contains an unnecessary escaped quote?

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,262 @@
...
+            )
+
+        return None
+
+    def _sub_twice(self, regex: regex.Pattern, replacement: str, original: str) -> str:
+        """Replace `regex` with `replacement` twice on `original`.
+
+        This is used by string normalization to perform replaces on overlapping matches.
+        """
+        return regex.sub(replacement, regex.sub(replacement, original))
```
tunetheweb, line 172: I'm not getting what this is for? And why you need to do this twice (and only twice)?

My Python is poor btw, so maybe this is obvious to a real Python developer.

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,262 @@
...
+        self, s: str, preferred_quote_char: str, alternate_quote_char: str
+    ) -> str:
+        """Prefer `preferred_quote_char` but only if it doesn't cause more escaping.
+
+        Adds or removes backslashes as appropriate.
+        """
+        value = s.lstrip(STRING_PREFIX_CHARS)
+
+        if value[:3] == preferred_quote_char * 3:
+            # TODO: Are we not replacing unnecessary quotes here?
```
tunetheweb, line 184: Do you mean unnecessary escaped quotes?

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,262 @@
...
+            self.logger.debug(
+                "Changing quote style would introduce more escapes in the body. "
+                "Before: %s After: %s . Skipping.",
+                body,
+                new_body,
+            )
+            return s  # Do not introduce more escaping
+
+        if new_escape_count == orig_escape_count and orig_quote == preferred_quote_char:
+            # Use preferred_quote_char
```
tunetheweb, line 259:
```suggestion
            # Already using preferred_quote_char, and no escape benefit to changing
```

tunetheweb (commented):
@barrywhart 's the rules expert but gave some comments and questions below.

src/sqlfluff/rules/L064.py
dmohns: Yeah, exactly. Just because the quotes match already, doesn't mean we are not going to touch the string. We will go into this case https://github.com/sqlfluff/sqlfluff/pull/3118/files#diff-c74eb3c9d616e01ddba16d42924a3a06bee0828c413408b642f6255e8c3a53d7R186-R190

src/sqlfluff/rules/L064.py
dmohns: My knowledge in other SQL dialects other than BigQuery and a bit of Postgres is actually not that large. From what I've seen rules for identifiers are pretty wild. However, for this rule to work we need a notion of preferred and alternate quotes. I don't even know if there is any dialect that has similar rule set for identifiers?

Even if there is I would prefer to avoid this for now and potentially add this when the use-case arises?

I agree that the documentation should be clear on that. Which part made you beliebe that identifiers were also covered by this rule? I tried to use the terminology of "STRING datatype" in rule documentation. But I'd be open for another (more clear) wording.

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,262 @@
...
+            "abc",
+            'abc',
+            "\"",
+            "abc" = 'abc'
+        from foo
+
+    **Best practice**
+
+    Ensure all STRING literals use preferred quotes, unless escaping can be reduced by
+    using alternate quotes.
```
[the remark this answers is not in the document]
dmohns: If we wanted that functionality we would have to implement if very carefully (basically from scratch). There are a lot of cases to consider, i.e. when converting `'"'` to `"\""` and interaction with tripple quotes.

I would absolutely not be against it, if we had such an option. However, the beauty of the current rule set is that we can just use the implementation by `black`. Hence I would prefer to leave as it is.

src/sqlfluff/rules/L064.py
dmohns: Yeah, I meant unnecessary escaped quotes. After double checking more carefully I realised that the idea is to never replace escaped quotes in tripple quoted strings. I will update the comment.

src/sqlfluff/rules/L064.py
dmohns: I have to admit I only understand 80% of the regex magic that is done in the implementation. I just copied this. It's talking about overlapping matches, so I'm prepared to just believe that this is actually needed 😂

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,262 @@
...
+    literal UDF Body definitions.
+
+    .. note::
+       This rule can be dangerous. If accidentally enabled for dialects that do not
+       support single *and* double quotes automated fixes can potentially break working
+       SQL code.
+
+       This rule is disabled by default for dialects that don't allow double quotes for
+       STRING datatype (e.g. Postgres). It can be enabled with the
+       ``force_enable = True`` flag.
```
tunetheweb, line 34:
```suggestion
    r"""Consistent usage of preferred quotes for quoted literals.

    Some databases allow single quotes as well as double quotes to express a quoted
    literal. Prefer one type of quotes as specified in rule setting. Falling back to
    alternate quotes to reduce escapes.

    Dollar quoted raw strings are excluded from this rule, as they are mostly used for
    literal UDF Body definitions.

    .. note::
       This rule only checks quoted literals and not quoted identifiers as they often
       cannot interchange single and double quotes

       This rule can be dangerous. If accidentally enabled for dialects that do not
       support single *and* double quotes automated fixes can potentially break working
       SQL code.

       This rule is disabled by default for dialects that don't allow double quotes for
       quoted literals (e.g. Postgres). It can be enabled with the
       ``force_enable = True`` flag.
```

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,262 @@
...
+        select
+            "abc",
+            'abc',
+            "\"",
+            "abc" = 'abc'
+        from foo
+
+    **Best practice**
+
+    Ensure all STRING literals use preferred quotes, unless escaping can be reduced by
```
tunetheweb, line 50:
```suggestion
    Ensure all quoted literals use preferred quotes, unless escaping can be reduced by
```

src/sqlfluff/rules/L064.py
tunetheweb: > Which part made you beliebe that identifiers were also covered by this rule? 

I just assumed. ass-you-and-me applies :-)

I thin we should use "quoted literals" over "STRING datatype". That's more consistent with what we use in other rules, for example: https://docs.sqlfluff.com/en/stable/rules.html#sqlfluff.core.rules.Rule_L048

Also suggest we explicitly call out that quoted identifiers are out of scope.

Made some suggested wording changes in latest review.

src/sqlfluff/rules/L064.py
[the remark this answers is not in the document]
tunetheweb: OK to leave as is.

src/sqlfluff/rules/L064.py
tunetheweb: See you accepted those suggestions!

Only outstanding question on this comment thread then is whether we bring BigQuery and MySQL into scope then?

dmohns: What do you mean by into scope here?

tunetheweb: You’re excluding those dialects but for quoted literals I don’t think you need to exclude them? I agree for quotes identifiers they have special rules but since that’s not in scope we don’t need to worry about them.

dmohns: I'm including them 😂 Currently, this rule is enabled for all dialects in this list https://github.com/sqlfluff/sqlfluff/pull/3118/files#diff-c74eb3c9d616e01ddba16d42924a3a06bee0828c413408b642f6255e8c3a53d7R69-R74
so `bigquery`, `hive `, `mysql` and `sparksql`.

tunetheweb: Ah no you’re NOT excluding - you’re including! Ignore me.

However, since you’re using an inclusion pattern I wonder if we can tone down the warning slightly?

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,260 @@
...
+    alternate quotes to reduce escapes.
+
+    Dollar quoted raw strings are excluded from this rule, as they are mostly used for
+    literal UDF Body definitions.
+
+    .. note::
+       This rule only checks quoted literals and not quoted identifiers as they often
+       cannot interchange single and double quotes
+
+       This rule can be dangerous. If accidentally enabled for dialects that do not
```
tunetheweb, line 31: Suggest we don’t need the scary first sentence since we’re working on inclusion list:

```suggestion
       If accidentally enabled for dialects that do not
```

Btw might want to change line breaks rather than just accept this.

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,260 @@
...
+       support single *and* double quotes automated fixes can potentially break working
+       SQL code.
+
+       This rule is disabled by default for dialects that don't allow double quotes for
```
tunetheweb, line 35: I’d change this paragraph to say what dialects are included.

src/sqlfluff/rules/L064.py
dmohns: Yeah, maybe even remove the warning entirely? It's a left over from the approach where it was disabled by default. I think it should be clear to users that if they put `force_enable = True` that they are leaving territory the rule was intended for. WDYT?

tunetheweb: I say leave it in, but removing first sentence. Made some more suggestions.

src/sqlfluff/core/default_config.cfg
```diff
@@ -180,3 +180,9 @@ extended_capitalisation_policy = consistent
 # Comma separated list of words to ignore for this rule
 ignore_words = None
 ignore_words_regex = None
+
+[sqlfluff:rules:L064]
+# Consistent usage of preferred quotes for STRING datatype
+preferred_string_quotes = consistent
+# Disabled for dialects that don't allow double quotes for STRING datatype (e.g. Postgres)
```
dmohns, line 187:
```suggestion
# Disabled for dialects that do not support single and double quotes for quoted literals (e.g. Postgres)
```

tunetheweb (approved):
I’m happy with this now. Cheers for working through all my comments!

@barrywhart think it would still be good to get your eyes on this.

src/sqlfluff/core/default_config.cfg
```diff
@@ -180,3 +180,9 @@ extended_capitalisation_policy = consistent
 # Comma separated list of words to ignore for this rule
 ignore_words = None
 ignore_words_regex = None
+
+[sqlfluff:rules:L064]
+# Consistent usage of preferred quotes for STRING datatype
```
dmohns, line 185:
```suggestion
# Consistent usage of preferred quotes for quoted literals
```

tunetheweb:
Just ran this on my code base and lots of flags for this rule. It's good and it's doing what it's supposed to, but think we should go to 0.13.0 next instead of 0.12.1 because of that.

src/sqlfluff/core/rules/config_info.py
```diff
@@ -177,6 +177,14 @@
             "in statements."
         ),
     },
+    "preferred_string_quotes": {
+        "validation": ["consistent", "single_quotes", "double_quotes"],
+        "definition": (
+            "Preferred quoting style to use for the quoted literals. If set to "
+            "``consistent`` quoting style is derived from the first quoted literal"
+            "encountered by the parser."
```
barrywhart, line 185:
```suggestion
            "``consistent`` quoting style is derived from the first quoted literal"
            "in the file."
```

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+@document_configuration
+@document_fix_compatible
+class Rule_L064(BaseRule):
+    r"""Consistent usage of preferred quotes for quoted literals.
+
+    Some databases allow single quotes as well as double quotes to express a quoted
+    literal. Prefer one type of quotes as specified in rule setting. Falling back to
+    alternate quotes to reduce escapes.
```
barrywhart, line 22:
```suggestion
    Some databases allow quoted literals to use either single or double quotes.
    Prefer one type of quotes as specified in rule setting, falling back to
    alternate quotes to reduce the need for escapes.
```

barrywhart (changes requested):

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,259 @@
...
+    _string_prefix_chars = "rbRB"
+
+    def _eval(self, context: RuleContext) -> Optional[LintResult]:
+        # Config type hints
+        self.preferred_string_quotes: str
+        self.force_enable: bool
+
+        if (
+            context.dialect.name not in self._dialects_with_double_quoted_strings
+            and not self.force_enable
```
barrywhart, line 97: Puts the fast flag check first, also tweaks the logic to eliminate use of `not` (easier to read IMHO). Also, I suggest we move this `if` below the one that checks the segment name, because the segment name check is faster. (We like `_eval()` functions that exit quickly when there's nothing to do.)
```suggestion
            self.force_enable or
            context.dialect.name in self._dialects_with_double_quoted_strings
```

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,259 @@
...
+
+    Dollar quoted raw strings are excluded from this rule, as they are mostly used for
+    literal UDF Body definitions.
+
+    .. note::
+       This rule only checks quoted literals and not quoted identifiers as they often
+       cannot interchange single and double quotes
+
+       For dialects that do not support single *and* double quotes for quoted literals
+       automated fixes can potentially break working SQL code.
```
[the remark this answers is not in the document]
tunetheweb: Yeah discussed above. It was worse. It's more a warning for `force-enabled=True`.

I'm OK removing it as you suggest (and @dmohns suggested above)

src/sqlfluff/rules/L064.py
barrywhart: In that case, can we include a comment/link to the code it was copied from?

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,259 @@
...
+        self, s: str, preferred_quote_char: str, alternate_quote_char: str
+    ) -> str:
+        """Prefer `preferred_quote_char` but only if it doesn't cause more escaping.
+
+        Adds or removes backslashes as appropriate.
+        """
+        value = s.lstrip(self._string_prefix_chars)
+
+        if value[:3] == preferred_quote_char * 3:
+            # In tripple quoted strings we are not replacing escaped quotes.
```
barrywhart, line 180:
```suggestion
            # In triple-quoted strings we are not replacing escaped quotes.
```

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,259 @@
...
+            orig_quote = preferred_quote_char
+            new_quote = alternate_quote_char
+        elif value[:3] == alternate_quote_char * 3:
+            orig_quote = alternate_quote_char * 3
+            new_quote = preferred_quote_char * 3
+        elif value[0] == alternate_quote_char:
+            orig_quote = alternate_quote_char
+            new_quote = preferred_quote_char
+        else:
+            self.logger.debug(
```
barrywhart, line 195: How could this happen?

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,259 @@
...
+                alternate_quote_char,
+            )
+            return s
+
+        first_quote_pos = s.find(orig_quote)
+        prefix = s[:first_quote_pos]
+        unescaped_new_quote = regex.compile(rf"(([^\\]|^)(\\\\)*){new_quote}")
+        escaped_new_quote = regex.compile(rf"([^\\]|^)\\((?:\\\\)*){new_quote}")
+        escaped_orig_quote = regex.compile(rf"([^\\]|^)\\((?:\\\\)*){orig_quote}")
+        body = s[first_quote_pos + len(orig_quote) : -len(orig_quote)]
```
barrywhart, line 209: Just a thought -- I wonder if it'd be worthwhile to move this logic to a new property or function on the segment class. (For reuse by other rules, etc.)

src/sqlfluff/core/rules/config_info.py
```diff
@@ -177,6 +177,14 @@
             "in statements."
         ),
     },
+    "preferred_string_quotes": {
```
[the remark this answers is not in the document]
dmohns: Yeah, good idea. The "string" is a bit of a leftover from earlier wording.

Any suggestions for a good name for that setting? `preferred_quoted_literal_quotes` seems a bit odd 😂

src/sqlfluff/rules/L064.py
```diff
@@ -0,0 +1,259 @@
...
+        if "r" in prefix.lower():
+            if unescaped_new_quote.search(body):
+                self.logger.debug(
+                    "There's at least one unescaped new_quote in this raw string "
+                    "so converting is impossible."
+                )
+                return s
+
+            # Do not introduce or remove backslashes in raw strings
+            new_body = body
```
barrywhart, line 220: This comment confuses me -- do we **know** that we have introduced or removed backslashes here? Would it be correct (and clearer) to say "Do not modify the body of raw strings"?

src/sqlfluff/core/rules/config_info.py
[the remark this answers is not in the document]
barrywhart: I checked, and we have other settings using the word "style". 🎉
* `comma_style`
* `group_by_and_order_by_style`

src/sqlfluff/rules/L064.py
dmohns: It's a bit of a safety net. Right now, it only really does something if the rule is force enabled for dialects that allow quoted literal styles other than single or double quotes. This test is covering it: https://github.com/sqlfluff/sqlfluff/pull/3118/files#diff-ad130ac57b12a68f9bda9d3474dcb126f54009c5b3e766641d314bdd5876dbc8R231-R243

What I had in mind is: Maybe there exists (right now, or in the future) a dialect that allows single quotes, double quotes _and_ dollar quotes.

src/sqlfluff/rules/L064.py
dmohns: Hmm, I guess it's a bit of personal taste question. To me the from the code itself it's quite clear that the body of raw strings is not getting modified, or at least `new_body = body` would suggest that to me. The question that needs to be answered is: Why are bodies of raw strings not modified? Because we don't want to introduce or remove escaped quotes, as this changes the actual string value.

I will try to make this more clear.

barrywhart (approved):
This is a really great contribution. Thanks for your hard work and patience! ❤️💯

Issue #3116, opened 2022-04-19
Aliasing forced on type in cast


fdw:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Similar to #2710, the `max` in `nvarchar(max)` is understood as a column reference and prefixed with an alias.
So, given
```sql
select cast(blah, nvarchar(max))
from table t
```
`sqlfluff` changes it to
```sql
select cast(t.blah, nvarchar(t.max))
from table t
```

### Expected Behaviour

`nvarchar(max)` is understood as a type and not changed.

### Observed Behaviour

sqlfluff produces broken SQL because the alias is prepended.

### How to reproduce

See above

### Dialect

T-SQL

### Version

Latest git `main`

### Configuration

Nothing relevant, I think

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

tunetheweb:
To me this is not "Similar to #2710" but is a direct duplicate. And the answer would be the same. Is there a reason you see it as different?

fdw:
#2710 results in an error, but in my case, it is fixed (but broken). I wasn't sure if that is the same problem or something only similar, so I opened this issue.

Sorry for the additional work. I'll see if I can implement your suggestion 🙂

tunetheweb:
Yeah think #270 is for `lint` and complaining about rule L028.
You ran fix and L028 "fixed" this incorrectly.
So same thing I think.

The solution should be simple enough, as suggested in #270 with a test case.

Gimme a shout if that's not the case and you need any help.

tunetheweb:
Closed with #3130

PR #3130, opened 2022-04-20
T-SQL: Support `NVARCHAR(MAX)`

### LINKED ISSUE #2710: TSQL nvarchar(max) incorrectly flagged as L027
### Search before asking

### What Happened

This nvarchar(max) is flagged L027 (References should be qualified if select has more than one referenced table/view.)
but the max in this case refers to the size of the nvarchar. 

```
CAST(columnname AS NVARCHAR(MAX)) AS columname
```

### Expected Behaviour

I expect the linter to accept this 

### Observed Behaviour

`Unqualified reference 'max' found in select with more than one referenced table/view`  

### How to reproduce

write a query like

```
select
cast(columname as nvarchar(max)) as columname,
anothercolumn
from table1
some join table2
```
You would get this error

### Dialect

tsql

### Version

0.10.1 

### Configuration

sqfluff lint . --dialect tsql --exclude-rules L031

Also a .sqlfluff file present with

```
exclude_rules = L031

variable=thingy
```
but that file is not relevant for the sql file that contains the L027 error

### Code of Conduct

fdw:

Support `nvarchar(max)` and similar phrasings.

fixes #2710, #3116
I don't think so

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_tsql.py
+++ b/src/sqlfluff/dialects/dialect_tsql.py
@@ -1235,8 +1235,8 @@ class DatatypeSegment(BaseSegment):
         ),
         Bracketed(
             OneOf(
-                Delimited(Ref("ExpressionSegment")),
                 "MAX",
+                Delimited(Ref("ExpressionSegment")),
                 # The brackets might be empty for some cases...
                 optional=True,
             ),
--- a/src/sqlfluff/dialects/dialect_tsql.py
+++ b/src/sqlfluff/dialects/dialect_tsql.py
@@ -1236,6 +1236,7 @@ class DatatypeSegment(BaseSegment):
         Bracketed(
             OneOf(
                 Delimited(Ref("ExpressionSegment")),
+                "MAX",
                 # The brackets might be empty for some cases...
                 optional=True,
             ),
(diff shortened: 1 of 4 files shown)


src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -1236,6 +1236,7 @@ class DatatypeSegment(BaseSegment):
         Bracketed(
             OneOf(
                 Delimited(Ref("ExpressionSegment")),
+                "MAX",
```
tunetheweb, line 1239: In the YAML below you can see the `MAX` is being parsed as a column identifier. Switch these around to give the `MAX` the precendence.

```suggestion
                "MAX",
                Delimited(Ref("ExpressionSegment")),
```

This will also allow other datatypes to use this (e.g. `VARCHAR(MAX)`). Not sure how much of a problem that would really be and if this simple option is fine? A more complex, but potentially more accurate solution is given here: https://github.com/sqlfluff/sqlfluff/issues/2710#issuecomment-1053556317

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_tsql.py
fdw: I decided on this solution on purpose, as `varchar(max)` is also valid SQL 🙂

tunetheweb: OK. Did you see my comment about switching the order? Need to do that and regenerate the YAMLs as currently they are wrong.

tunetheweb (approved):

PR #3142, opened 2022-04-21
Add rule groups and a core rules group

### LINKED ISSUE #811: Group or tag the core rules to make bulk enabling/disabling of rules a possibility
Overall, the growing number of core rules in SQLFluff is a great thing, because it catches more issues. However, many users don't want to simply enable all the rules. For example:
* Some of the rules may not be universally accepted as the one best way.
* Some rules make bigger changes than others. Even when the changes are good, bigger code changes are harder to review, and users may want to make the changes in stages.

For these reasons, rules should be able to be grouped or tagged and enabled/disabled as a group. Below is a proposed set of groups:

- Formatting: Whitespace, comma placement, indentation, line length: L001-L009, L016-L019, L022-L024, L036, L038, L039, L041
- Things that won't run: L026 
- Capitalization: e.g. L010, L014, L030, L040
- Style, readability: L011, L012, L032, L042
- Clarify code, make it more readable, address common misinterpretations: L015, L021, L027-L029, L033, L034, L037
- Remove unused/unnecessary code: L025, L035, #796
- Avoiding corner cases, unpredictable/surprising behavior: #380

pwildenhain:

Fixes #811 

As I studied the problem more, it made more sense for rule groupings to be something controlled by the config, rather than by the rule itself (like rule tags or something like that). What this change now essentially amounts to is an opinionated default list of core rules, that users can modify if they wish to include more core rules. It also makes it easier to roll out SQLFluff to a team by only running core rules at first.

User can now define their own rule groups, though I'm not sure there's much utility in that, since they still need to write out all the rule names. I think a good addition to this feature would be "common sense" names for each of the rules (i.e `L001 = "trailing-whitespace"`) which makes it easier for users to define things like that.

It also opens up the potential for different types of groups that the package maintainers might want to offer in the future (`dbt-best-practice`, `safe-sql`, etc.) for user convenience

If we have agreement on this, I'll document this new feature as well as the core rule list in the docs.

### Pull Request checklist

  - Other.
- **Added appropriate documentation for the change**. (Not done yet)


tunetheweb:
> As I studied the problem more, it made more sense for rule groupings to be something controlled by the config, rather than by the rule itself (like rule tags or something like that). 

I still think it might be better in the rule, for a few reasons:
1. Less likely to be missed when a new rule is added
2. Can be in rule docs with a rule decorator

pwildenhain:
Totally open to changing this approach. How should we identify when a rule group has been added to the 'rules' or 'excludes_rules' config?

If it's a part of the rule, then we need to check for every rule if it's been included/excluded vs. just checking one time once the .sqlfluff config is loaded.

Am I seeing that right? Or am I missing something?

tunetheweb:
Yeah that would be a downside. It would presumably have to go through all rules and then check whether that rule is in the currently configured group, and skip if not. Would still be pretty quick though, but yeah maybe there's a better way to only do that once (maybe on the first query run?) and not have to do that for every SQL file.

pwildenhain:
> (maybe on the first query run?) and not have to do that for every SQL file.

Yeah I agree there likely is a better way. I don't think the proposed way would work because different sql files can have different config values, if the user nests multiple `.sqlfluff` files across their directories

src/sqlfluff/core/config.py
```diff
@@ -518,6 +527,21 @@ def __init__(
             self._configs["core"]["templater"]
         )
 
+    def _lookup_rules_from_group(
+        self, rules: List[str], rule_groups: List[str]
+    ) -> List[str]:
+        """Look up group rules and add them to the a rules list."""
```
barrywhart, line 533: Looks like a typo: to the a rules list

src/sqlfluff/core/default_config.cfg
```diff
@@ -58,6 +58,47 @@ allow_scalar = True
 single_table_references = consistent
 unquoted_identifiers_policy = all
 
+# Rules can be grouped together for easy enabling/disabling
+[sqlfluff:rules:groups]
+core =
```
barrywhart, line 63: Are any tests using this group of rules? I see the group_exclude and group_include tests. Seems like we should have a test for this group, since it's the one users are most likely to use.

barrywhart:
Re: If the group membership was on the rules, we are already doing something similar with rule phases here: https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/core/linter/linter.py#L514L516, i.e. looping over the rules, checking a property, deciding what to run.

barrywhart:
I don't see any doc changes in the PR -- were you planning to do that once there's agreement on the general approach?

pwildenhain:
> I don't see any doc changes in the PR -- were you planning to do that once there's agreement on the general approach?

Correct -- it seems like you agree with @tunetheweb that this should be something that lives with the rules?

src/sqlfluff/core/default_config.cfg
pwildenhain: Right now, no but I could add a test for that

barrywhart:
I don't have a strong opinion where it should live, but wanted to share a note about how it could work. 

I may have some thoughts to share later about how we could enforce assigning a new rule somewhere, even if it's in config.

barrywhart:
The thought I mentioned earlier: If we want to be sure that we consider each new rule's membership in some group, we could add a startup check that every rule belongs to at least (or exactly one) group. Then define groups "core", "noncore", and a few others if we want.

pwildenhain:
Having an all group makes sense to me. 

Do we all agree that we want users to use the feature like this:

```
$ sqlfluff my_file.sql --rules core
```

Or is there a different vision we have for how rule groupings will be used by the user?

barrywhart:
For a moment, I was worried that using `--rules` for both groups and individual rules could cause problems or confusion, but TBH I like it. People shouldn't have groups whose names start with `L`.

We should spend a little time thinking, documenting, and writing tests for what happens if someone mixes rules and groups in various ways, e.g.:
```
--exclude-rules L003 --rules core
```
Which wins? One possible answer: Always process `--rules` first, then `--exclude`. (Note that we won't know the order they appear on the command line, at least using `click`.)

pwildenhain:
> Always process `--rules` first, then `--exclude`.

Agree that this makes the most sense, and we should make weird test cases that will mimic what users will _most certainly_ do in the real world

barrywhart:
>Agree that this makes the most sense, and we should make weird test cases that will mimic what users will most certainly do in the real world

Exactly. I don't necessarily care that they do it, but then they'll ask us to explain why it's not working, and that's hard work. 🤣

pwildenhain:
@barrywhart and @tunetheweb I've got a draft pushed. Tested manually and all the following should work

```
$ sqlfluff lint my_file.sql --rules core --exclude-rules L001
$ sqlfluff lint my_file.sql --exclude-rules core
```

Holding off on adding tests/docs until we come to an consensus on the implementation. I think this is pretty good. Only looks once, while it's also looking for the regular allow/deny list rules. I added `groups` as a top level attribute to `self._register` but this seemed important enough to warrant that. Also can confirm that trying to a new rule without any groups will trigger an `AttributeError`

src/sqlfluff/core/rules/base.py
```diff
@@ -1067,11 +1078,42 @@ def eval(self, **kwargs):
...
             )
-        self._register[code] = dict(code=code, description=description, cls=cls)
+
+        try:
+            groups = cls.groups
+        except AttributeError:
+
+            raise AttributeError(
+                (
+                    "Rule {!r} doesn't belong to any rule groups. "
```
barrywhart, line 1088: This error message implies that every rule must belong to at least one group, but it appears to allow `groups` to be an empty collection. Which of these is our true intention?

src/sqlfluff/core/rules/base.py
```diff
@@ -1103,12 +1145,19 @@ def get_rulelist(self, config) -> List[BaseRule]:
         """
         # Validate all generic rule configs
         self._validate_config_options(config)
+        # Find all valid groups for ruleset
+        valid_groups = set(
+            [group for _, attrs in self._register.items() for group in attrs["groups"]]
```
barrywhart, line 1150:
```suggestion
            [group for attrs in self._register.values() for group in attrs["groups"]]
```

src/sqlfluff/core/rules/base.py
```diff
@@ -1103,12 +1145,19 @@ def get_rulelist(self, config) -> List[BaseRule]:
...
         # default the allowlist to all the rules if not set
         allowlist = config.get("rule_allowlist") or list(self._register.keys())
         denylist = config.get("rule_denylist") or []
 
         allowlisted_unknown_rule_codes = [
-            r for r in allowlist if not fnmatch.filter(self._register, r)
+            r
+            for r in allowlist
+            # Add valid groups to the register when searching for invalid rules _only_
+            if not fnmatch.filter({**self._register, **dict.fromkeys(valid_groups)}, r)
```
barrywhart, line 1160: I'm having a pretty hard time understanding this line of code. Is there a way to write it that's more clear?

src/sqlfluff/rules/L001.py
```diff
@@ -30,6 +30,7 @@ class Rule_L001(BaseRule):
         FROM foo
     """
 
+    groups = ("all", "core")
```
barrywhart, line 33: What's the benefit of having an "all" group?

plugins/sqlfluff-plugin-example/src/example/rules.py
```diff
@@ -68,15 +68,16 @@ class Rule_Example_L001(BaseRule):
         ORDER BY bar
     """
 
-    config_keywords = ["forbidden_columns"]
```
barrywhart, line 71: Nitpick: Why was this moved below the `__init__` function?

src/sqlfluff/core/rules/base.py
pwildenhain: I think the former. What do you think about explicitly checking that the rule is a part of the 'all' group, and making the error message pertain to that?

src/sqlfluff/rules/L001.py
pwildenhain: I think if in the future we only enable core rules by default, this will give users the option to turn on all the rules with `--rules all`

plugins/sqlfluff-plugin-example/src/example/rules.py
pwildenhain: No good reason -- I'll move it up

src/sqlfluff/core/rules/base.py
pwildenhain: Previously `fnmatch.filter` was being used on the register dict, meaning it had all the rules as the keys of the dict to search through valid rule codes. I wanted to add all the valid groups as keys to that dict

Perhaps a better way would be to just create a separate list of rule codes from the register dict keys and then add these valid groups to that list. How about that?

src/sqlfluff/core/rules/base.py
barrywhart: Sounds good!

src/sqlfluff/core/rules/base.py
barrywhart: That sounds good. To be clear, the existing code is legit, just a little too terse for my brain to process, at least on a Friday evening. 😅

src/sqlfluff/rules/L001.py
barrywhart: Ah, good point!

src/sqlfluff/core/rules/base.py
tunetheweb: WDYT about not needing to explicitly specify `all`? Should it be implicit? That way the `groups` config is more of an "additional groups".

Can see positives and negatives:

positives:
- a rule can't be forgotten to be added to `all`
- less needless boilerplate

negatives:
- test would catch the case where it's forgotten
- potentially clearer to list all the groups - including `all`
- documentation might be easier

I think I've veering on keeping it as you have, but definitely think we need a test to check someone doesn't add a new rule and forget to add to `all`.

src/sqlfluff/rules/L001.py
tunetheweb: ON that, should we update `default_config` to be `all` rather than the current `None`:

https://github.com/sqlfluff/sqlfluff/blob/b31402765168b65841165b47a3d76b509c52a64d/src/sqlfluff/core/default_config.cfg#L11-L12

We can leave `none` to work as it currently does (i.e. being same as `all`) but putting `all` in our default config would be clearer IMHO.

src/sqlfluff/core/rules/base.py
barrywhart: We could update the existing check in this PR, the one that catches `AttributeError`, and also verify the value is not empty.

tunetheweb: Think it needs to be more than just empty. Need to ensure `all` is in there. Otherwise that could be missed by a rule author.

barrywhart: True.

pwildenhain: I'll keep the attribute error check in case someone forgets to add the groups variable at all, and also add a check to make sure the 'all' group is present

docs/source/rules.rst
```diff
@@ -9,6 +9,26 @@ a particular rule or set of rules. The intent is that the definition of
...
+Core Rules
+__________
+
+Certain rules are belong to the :code:`core` rule group. In order for
+a rule to be designated as :code:`core`, it must meet the following
+criteria:
+
+* Stable
+* Applies to most dialects
+* Could detect a parsing issue
```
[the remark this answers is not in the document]
pwildenhain: Ah I meant 'syntax error'

Something like this

```sql
select
    x,
    y
    z
from my_table
```

This is 'valid' SQL, but very likely the user forgot a comma

This would warn about implicit aliasing, which would point to the  larger syntax oversight 

I can change the language I used in the docs

barrywhart:
@pwildenhain: Where are we with this PR? Would be good to get it merged!

Above, you mentioned you were going to make a docs change. IIRC, overall the PR was looking pretty good to me.

@tunetheweb

pwildenhain:
Sure thing -- I can push the necessary changes tomorrow if we have agreement

src/sqlfluff/core/rules/base.py
```diff
@@ -1067,11 +1077,45 @@ def eval(self, **kwargs):
...
-        self._register[code] = dict(code=code, description=description, cls=cls)
+
+        try:
+            assert (
+                "all" in cls.groups
+            ), "Rule {!r} must belong to the 'all' group".format(code)
+            groups = cls.groups
+        except AttributeError:
+
+            raise AttributeError(
```
barrywhart, line 1088: Should we use `raise from` here?

https://www.pythontutorial.net/python-oop/python-raise-from/

src/sqlfluff/core/rules/base.py
```diff
@@ -1067,11 +1077,45 @@ def eval(self, **kwargs):
...
+        self, rule_list: List[str], valid_groups: Set[str]
+    ) -> List[str]:
+        expanded_rule_list: List[str] = []
+        for r in rule_list:
+            if r in valid_groups:
+                rules_in_group = [
+                    rule
+                    for rule, rule_dict in self._register.items()
+                    if r in rule_dict["groups"]
+                ]
```
barrywhart, line 1112: Maybe a question for a future PR: Does this support nested rule groups, e.g., maybe later we want to define the core group as the combination of several groups, e.g. whitespace, "making things explicit", "foobar". In that case, expansion of groups would need to be recursive.

barrywhart (approved):
A few small suggestions and questions. Overall LGTM.

docs/source/rules.rst
```diff
@@ -9,6 +9,26 @@ a particular rule or set of rules. The intent is that the definition of
 each specific rule should be really streamlined and only contain the logic
 for the rule itself, with all the other mechanics abstracted away.
 
+Core Rules
+__________
+
+Certain rules are belong to the :code:`core` rule group. In order for
```
tunetheweb, line 15:
```suggestion
Certain rules belong to the :code:`core` rule group. In order for
```

docs/source/rules.rst
```diff
@@ -9,6 +9,26 @@ a particular rule or set of rules. The intent is that the definition of
...
+criteria:
+
+* Stable
+* Applies to most dialects
+* Could detect a syntax issue
+* Isn’t too opinionated toward one style (e.g. the :code:`dbt` style guide)
+
+Core rules can also make it easier to rollout SQLFluff to a team by
+only needing to follow a 'common sense' subset of rules, rather
+than wasting time debating the finer points of other more
```
tunetheweb, line 26: Dislike the "wasting time" wording. Can we think of better phrasing?

src/sqlfluff/rules/L002.py
```diff
@@ -41,6 +41,7 @@ class Rule_L002(BaseRule):
 
     """
 
+    groups = ("all", "core")
     config_keywords = ["tab_space_size"]
```
tunetheweb, line 45: Why is group a Tuple, and `config_keywords` a List?

Should `groups` be a List?

tunetheweb (changes requested):
Small number of questions/suggestions.

src/sqlfluff/core/rules/base.py
pwildenhain: I could see that as a nice DRY alternative, rather than manually specifying the same rule in different groups over and over again. I agree though that we should leave it as a consideration for a future PR -- who knows, maybe no one will care about rule groups 😂

tunetheweb (approved):
Approved, but made one further suggestion to the wording.

PR #3145, opened 2022-04-21
L025 bug fix: stop incorrectly flagging on nested inner joins

### LINKED ISSUE #3144: incorrect linter errors in version 0.12
### Search before asking

### What Happened

Upgraded to sqlfluff 0.12 and immediately got errors that (a) we didn't get before and (b) I suspect to be invalid.

### Expected Behaviour

SQL file gets linted without error

### Observed Behaviour

sqlfluff throws error

> L:  26 | P:  95 | L025 | Alias 'ceb' is never used in SELECT statement

You'll see form looking at the SQL that I provided below that alias `ceb` is indeed used

```sql
                        select
                                ceb.emailaddresskey,
                                dac.accountkey,
                                row_number() over (partition by dac.currentemailaddresskey order by coalesce(dac.lastprofileupdatedatetime, dac.accountcreationdate) desc) as rnum
                            from `${datalake_project_id}.edw_red.REDDimAccount` as dac
                            inner join `${datalake_project_id}.edw_red.REDDimEmailAddress` as ceb
                                on ceb.emailaddresskey = dac.currentemailaddresskey   --noqa
                            where dac.currentemailaddresskey != -1
                                and coalesce(dac.AccountCreationDate, '1900-01-01') < timestamp_trunc(current_timestamp(), day)
                                and coalesce(dac.LastProfileUpdateDateTime, '1900-01-01') < timestamp_trunc(current_timestamp(), day)
                                and trim(ceb.EmailMarketableStatus) not in ( 'Disposable', 'invalid_account', 'invalid_address', 'invalid_domain', 'possible_spamtrap')
                                and ceb.EmailAddressKey != (-1)
                        
```

therefore I don't think this is valid.

### How to reproduce

write this to a file somewhere:
```sql
with max_key as (
    select max(ExtractAuditKey) as ExtractAuditKey
    from `${project_containing_tables_required_for_braze}.edw.AccountSavedAddressBridge`
),

CRMEmailMyProfileLstAddressKey as (
    select
        abg.emailaddresskey,
        abg.addresskey
    from
        (select
                dac.emailaddresskey,
                abg.addresskey,
                row_number() over (partition by abg.accountkey order by abg.LastUpdatedDateTime desc, abg.addresskey desc) as rnum
            from `${project_containing_tables_required_for_braze}.edw.AccountSavedAddressBridge` as abg
            inner join
                (select     --noqa
                        emailaddresskey,
                        accountkey
                    from
                        (select
                                ceb.emailaddresskey,
                                dac.accountkey,
                                row_number() over (partition by dac.currentemailaddresskey order by coalesce(dac.lastprofileupdatedatetime, dac.accountcreationdate) desc) as rnum
                            from `${datalake_project_id}.edw_red.REDDimAccount` as dac
                            inner join `${datalake_project_id}.edw_red.REDDimEmailAddress` as ceb
                                on ceb.emailaddresskey = dac.currentemailaddresskey   --noqa
                            where dac.currentemailaddresskey != -1
                                and coalesce(dac.AccountCreationDate, '1900-01-01') < timestamp_trunc(current_timestamp(), day)
                                and coalesce(dac.LastProfileUpdateDateTime, '1900-01-01') < timestamp_trunc(current_timestamp(), day)
                                and trim(ceb.EmailMarketableStatus) not in ( 'Disposable', 'invalid_account', 'invalid_address', 'invalid_domain', 'possible_spamtrap')
                                and ceb.EmailAddressKey != (-1)
                        )
                    where rnum = 1) as dac
                on abg.accountkey = dac.accountkey   --noqa
                    and abg.isactive = 'Yes'
                    and coalesce(abg.CreatedDateTime, '1900-01-01') < timestamp_trunc(current_timestamp(), day)
                    and coalesce(abg.LastUpdatedDateTime, '1900-01-01') < timestamp_trunc(current_timestamp(), day)
            where abg.ExtractAuditKey > coalesce((
                        select HWMTimeStamp
                        from `${audit_dataset_id}.${HighWaterMark_table_id}`
                        where Process = 'MSM_IN_MODEL.UserProfile_Events_AttitudinalSegment'
                        order by CompletedTime desc
                        limit 1), 0)
        ) as abg
    where rnum = 1
),

CRMAttitudinalSegment as (
    select
        cma.EmailAddressKey,
        pal.AttitudinalSegment
    from CRMEmailMyProfileLstAddressKey as cma
    inner join `${datalake_project_id}.edw_red.REDDimAddress` as daddr
        on cma.addresskey = daddr.addresskey
    left join `${project_containing_tables_required_for_braze}.edw.PostcodeAttLookup` as pal
              on upper(
                replace(
                    daddr.PostCode,
                    ' ',
                    ''
                )
            ) = upper(
                replace(
                    pal.PostCodeFormat3,
                    ' ',
                    ''
                )
            )
    group by cma.EmailAddressKey, pal.AttitudinalSegment
),

ExtAttitudinalSegment as (
    select
        uc.UserConsentId as ExternalId,
        cas.AttitudinalSegment,
        lower(uc.EmailAddress) as EmailAddress
    from CRMAttitudinalSegment as cas
    inner join `${datalake_project_id}.edw_red.REDFctUserConsent` as uc
        on cas.EmailAddresskey = uc.EmailAddresskey
            and uc.UserConsentUpdatedDateTime >= cast(date_sub(current_date(), interval 3 year) as timestamp)
            and upper(uc.UserConsentBrandName) = 'MONEYSUPERMARKET'
    group by uc.UserConsentId, uc.EmailAddress, cas.AttitudinalSegment
)

select
    a.ExternalId,
    a.EmailAddress,
    a.AttitudinalSegment,
    max(b.ExtractAuditKey) as ExtractAuditKey
from ExtAttitudinalSegment as a
cross join max_key as b
group by a.ExternalId, a.EmailAddress, a.AttitudinalSegment
```

then run 
`sqlfluff lint --dialect bigquery --exclude-rules L016,L031,L014,L057 path/to/file.sql`

### Dialect

bigquery

### Version

0.12

### Configuration

n/a we configure via command line options

### Code of Conduct

tunetheweb:

Fixes #3144

Adds a new, optional, parameter to `recursive_crawl` to allow you to provide a type that should not be recursed into.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1460,7 +1460,10 @@ def get_eventual_aliases(self) -> List[Tuple[BaseSegment, AliasInfo]]:
             buff.append((from_expression, alias))
 
         # In some dialects, like TSQL, join clauses can have nested join clauses
-        for join_clause in self.recursive_crawl("join_clause"):
+        # recurse into them - but not if part of a sub-select statement (see #3144)
+        for join_clause in self.recursive_crawl(
```
barrywhart, line 1464: I don't think we're fixing the real issue here. It seems to me that we don't need to ignore `select_statement`. Instead, we need to detect that the `on` clause is using the alias, therefore it's not unused.

Do we know why the behavior changed from previous versions?

tunetheweb: This was discussed more in [this comment in #3144](https://github.com/sqlfluff/sqlfluff/issues/3144#issuecomment-1105721052)

Basically, previously it was this:

```py
for join_clause in self.get_children("join_clause"):
```

That is, we only got the direct join children, rather than using `recursive_crawl`.

However #2993 was added to make it recursive to allow SQL like this:

```sql
    SELECT
        1 AS RegionCode
    FROM BA
    LEFT OUTER JOIN (
        I LEFT OUTER JOIN P
            ON I.Pcd = P.Iid
    ) ON BA.Iid = I.Bcd
```

Note that the outermost `ON` clause is referencing the inner table `I`. This is a specific syntax allowed in T-SQL (but no where else I don't think?), so the above code needed to change to be `recursive_crawl` instead of `get_children`.

Unfortunately there is no easier way of identifying just this case as there will be brackets, and there could be nested, nested calls. But `recursive_crawl` will find them so all good.

That solve that problem, but then broke it for this SQL - hence this PR to partially revert that:

```sql
select col1
from table1 as abg
inner join (
    select
        ceb.emailaddresskey,
        dac.accountkey
    from table2 as dac
    inner join table3 as ceb
        on ceb.col2 = dac.col2
) as abh
on abg.col1 = abh.col1
```

Here the subquery should **not** be accessible at all to the outer query, and only the results of that (as `abh`) as it's not that specific T-SQL instance above, which is the only one that should allow nesting. So effectively it should be a black box, and should look like this (at least while while the rule process the outer select - it'll get to the nested one later):

```sql
select col1
from table1 as abg
inner join (
    ---something we don't need to worry about now
) as abh
on abg.col1 = abh.col1
```

Then we process that inner join later in the rule.

That is what used to happen before #2993 , and then, while processing the outer select statement,  it would see table references `abg` and `abh` and then uses of both those in the `ON` clause - great, no rule to fire!

Instead what happens (and why we need this PR) is we now pick up recursive "joins" so we ALSO pick up table reference `ceb` when we shouldn't. And there is no reference to `ceb` in the outer select, hence the rule fires complaining about that.

Note that we don't pick up the inner table reference `dac` btw cause that is an ``FROM` clause and not a `JOIN` clause and the bug is specific to `JOIN` clauses. Again, showing that this is wrong to traverse into the inner select for just `JOIN`s.

As the nested `SELECT` should be a black box (except in very specific use cases for T-SQL, when it's an nested `JOIN`), I've added functionality to **not** recurse into it anymore, if it is it's own select statement (which is exactly what happened before), but still recurse for anything else (so the original T-SQL bug is still resolved). There may be other instances we might need to do that, but that's the only one I can think of now.

The alternative is to gather ALL the table references for the whole query (even the inner, nested ones that should not be referenceable, and including `dac` in this example) and ALL the column references, and then decide which are used and which are not. But that is not the correct thing to do here, since the inner, nested table references are not be accessible outside of it.

Here's a BigQuery example showing the inner table canot be referenced in the outer select:

image

Only via the outer alias:

image

Does that explain it all?

WittierDinosaur: This makes sense to me, but I'll leave it to @barrywhart to ✅ if he is happy with the explanation

barrywhart: Makes sense!

In general, visibility goes upwards, but not downwards.

barrywhart (approved):

PR #3128, opened 2022-04-20
Snowflake: `StartsWith()` in `FromExpressionElementSegment` caused performance issues for large queries

WittierDinosaur:
### Brief summary of the change made
StartsWith() is not performant (to be addressed). For now, I've removed this from FromExpressionElementSegment(). This has parity with ANSI, and caused no errors or changes in our suite of Snowflake dialect tests

Fixes #3127

### Are there any other side effects of this change that we should be aware of?
No

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #3165, opened 2022-04-23
Remove unnecessary `StartsWith` and make `terminator` mandatory when using it

tunetheweb:

A `StartsWith` in `match_grammar` that just repeats the start of `parse_grammar` without a `terminator` clause is pointless. We've proven previously there is no performance benefit of this.

It needlessly repeats grammar.

Worse is that it can cause a performance problem - see #3128 

This PR cleans it up.

This makes terminator mandatory when using `StartsWith`. Don't think there are any case for using it without that?

### Pull Request checklist

  - Other.


barrywhart (approved):
Nice! How did you identify these cases? Seems like we could add an automated check to detect and prevent future segment definitions that violate this. Think that's worth doing? (Could be a separate issue/PR.)

tunetheweb:
> Nice! How did you identify these cases?

Simple search in VS Code and going through them one by one.

> Seems like we could add an automated check to detect and prevent future segment definitions that violate this. Think that's worth doing? (Could be a separate issue/PR.)

Was thinking about this. think I could do a simple assert in `StartsWith` definition to confirm a `terminator` exists.

Trying to think of any case where we might want this, but can't really.

tunetheweb:
Added that in https://github.com/sqlfluff/sqlfluff/pull/3165/commits/26bea26d7d4b6477eca48e9d37a3fa5916c43eb1 @barrywhart

tunetheweb:
Btw would like to get @WittierDinosaur ’s opinion on this before merging so hold off merging this for now @barrywhart

src/sqlfluff/core/parser/grammar/greedy.py
```diff
@@ -161,6 +163,10 @@ def __init__(self, target, *args, **kwargs):
         self.target = self._resolve_ref(target)
         self.terminator = self._resolve_ref(kwargs.pop("terminator", None))
         self.include_terminator = kwargs.pop("include_terminator", False)
+
+        # StartsWith should only be used with a terminator
+        assert self.terminator
```
barrywhart, line 168: Could we look at using this technique? "Required keyword argument". If it works, it seems perfect for this, because Python itself would enforce it, avoiding the need for an assertion or other runtime check.

WittierDinosaur (approved):
I think this is great. I do worry a bit with making the grammar more restrictive that we may catch the odd edge case that isn't in our test suite. But, StartsWith() is not performant, so I'm very on board with stripping it out where possible

tunetheweb:
Well `StartsWith` Is only used in `match_grammar` and even with that it would need to match `parse_grammar` after that anyway so don’t _think_ it’ll cause an issue.

Well check out @barrywhart ‘s suggestion as an alternative way to enforce the terminator and then merge this with either that or as is.

src/sqlfluff/core/parser/grammar/greedy.py
tunetheweb: Had a look at this @barrywhart but not as easy if you want to also keep `*args` and `**kwargs`. Removing all of them on both `StartsWith` and the underlying `GreedyUntil` would be a bit of refactoring, that I think is beyond the scope of this PR.

You would also need to assign the `terminator` anyway so it would still be a line of code.

Finally I'm not convinced it's any clearer. In fact I think current code is clearer! And more in keeping with the rest of the style we use elsewhere.

So going to merge as is. Can revisit in a separate PR or in future if we want.

PR #3168, opened 2022-04-24
Highlight `PRS` errors in red

OTooleMichael:
In large outputs PRS errors are almost impossible to see with the human eye. In addition they tend to be the only thing you care about if they exist. This PR just changes the color printed to be RED if the violation is a PRS error

No Side Effects

### DIFF
--- a/src/sqlfluff/cli/formatters.py
+++ b/src/sqlfluff/cli/formatters.py
@@ -56,7 +56,7 @@ def split_string_on_spaces(s: str, line_length: int = 100) -> List[str]:
 
 def format_violation(violation: SQLBaseError, max_line_length: int = 90) -> str:
     """Format a violation."""
-    if not isinstance(violation, SQLBaseError):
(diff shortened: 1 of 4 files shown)


OTooleMichael:
@tunetheweb  or @tunetheweb  Do either of you understand why mypy is failing. I can't recreate it locally

tunetheweb:
> @tunetheweb or @tunetheweb Do either of you understand why mypy is failing. I can't recreate it locally

Pinging @barrywhart or @WittierDinosaur since I’m away today and not back till later.

And presumed you wanted to ping one of them anyway rather than me twice 😁

tunetheweb (approved):
Change LGTM.

Quite a bit of clean-up of code in here, and typing. It's good, but would be nice to mention this is included in PR comment. And/or highlight the actual change in a comment, and mention the rest is just clean up.

We also don't have a test for this change. But then don't think we have for the other colours either. Adding that is probably overkill.

PR #722, opened 2021-01-30
Issue 356: Rules that fail when using BigQuery UNNEST() with alias

barrywhart:
Addresses #356. This PR adds a new ANSI-level set called `value_table_functions`. It's empty at the ANSI level, but BigQuery updates it to add `UNNEST`. This set is not used in the grammars, but when there is a `table_expression` with an alias, affected rules use this set to determine the nature of that alias: Is it a table alias or a column alias? This is a weird concept, but that's essentially what BigQuery does.

Context: `UNNEST()` is usually, but not always, used to unpack an array of scalar values. In this case, it returns a "table" with only one column. BigQuery has the concept of ["value tables"](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#value_tables):
>In BigQuery, a value table is a table where the row type is a single value. In a regular table, each row is made up of columns, each of which has a name and a type. In a value table, the row type is just a single value, and there are no column names.

In this situation, BugQuery seems to sort of treat the "table" and its single column as the same thing, allowing the alias for the value table to be used directly as a column. It's fairly natural for users of `UNNEST()`, but it's pretty weird from a SQLFluff perspective, since the nature of the alias has to be deduced with more complex logic. And, as you'll see in some of my comments on the issue (#356), it's probably not perfect for two reasons, both covered in the BigQuery docs [here](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#value_tables):
* While `UNNEST()` is usually used with scalar arrays, BigQuery also supports arrays of `STRUCT`, which return regular tables, not value tables, when unnested.
* BigQuery supports **implicit** unnesting of array columns, in which case, the `UNNEST()` function does not appear in the query.

This PR makes no attempt to address either of these last two challenges, but it does address all the issues initially raised in #356, which I think will cover the majority of array unnesting use cases.

PR #3194, opened 2022-04-27
Snowflake: Handle FLATTEN() table function aliases correctly in L025, L027, L028

### LINKED ISSUE #3178: Snowflake: Interaction between L025 and L028 corrupts SQL
### Search before asking

### What Happened

sqfluff fix 
```
SELECT
  r.rec:foo::string AS foo
  , value:bar::string AS bar
FROM foo.bar r, LATERAL FLATTEN(input => rec:result) x
;
```
Produces:
```
L:   3 | P:   5 | L028 | Unqualified reference 'value' found in single table select.
L:   4 | P:  54 | L025 | Alias 'x' is never used in SELECT statement.

SELECT
  r.rec:foo::string AS foo
  , r.value:bar::string AS bar
FROM foo.bar r, LATERAL FLATTEN(input => rec:result)
;
```
The issue here is it doesn't recognise the `LATERAL FLATTEN` as a table entity. Replacing the `LATERAL FLATTEN` with `bar` gives

```
SELECT
  r.rec:foo::string AS foo
  , value:bar::string AS bar
FROM foo.bar r, bar x
;

L:   4 | P:  21 | L025 | Alias 'x' is never used in SELECT statement.
```

### Expected Behaviour

LATERAL FLATTEN should be treated like a table

### Observed Behaviour

-

### How to reproduce

-

### Dialect

Snowflake

### Version

0.13.0

### Configuration

Unaffected by config

### Code of Conduct

barrywhart:
Fixes #3178

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -86,6 +86,10 @@
     ("exclude", "StartExcludeBracketSegment", "EndExcludeBracketSegment", True)
 )
 
+# In Snowflake, FLATTEN() returns a "value table".
+# https://docs.snowflake.com/en/user-guide/json-basics-tutorial-flatten.html
+snowflake_dialect.sets("value_table_functions").update(["flatten"])
```
barrywhart, line 91: Because this was missing, [this code](https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/core/rules/analysis/select.py#L137L138) was not executing, preventing the `FLATTEN()` call from being recognized as returning a table.

WittierDinosaur:
So there are actually a whole bunch of table functions in snowflake: 
https://docs.snowflake.com/en/sql-reference/functions-table.html
I'll leave it up to you whether you think adding them all is in-scope of this ticket. Happy to approve if you want to do this bugfix independently

barrywhart:
@WittierDinosaur: Let's create a separate issue for the other functions. Also, even with this fix, the query is still being modified by other rules, although it's now leaving the `x` alias in place. I'll look at avoiding these other "fixes" as well. I'm guessing other rules are still ignoring the `FLATTEN` as a "table", even though the core function works now.
```
SELECT
    foo.bar.rec:foo::string AS foo,
    foo.bar.value:bar::string AS bar
FROM foo.bar, LATERAL FLATTEN(input => rec:result) AS x;
```

WittierDinosaur:
Cool, scope-wise then. I think stopping the query being modified is in scope, I'll make another issue for further table function support

tunetheweb (approved):

barrywhart:
I'm putting this PR back in "draft" status. The Snowflake `FLATTEN()` function is a bit different than the similar `UNNEST()` function in BigQuery, which we added support for in PR #722.

`UNNEST()` returns an ambiguous thing that is both a table and a column. On the other hand, I think `FLATTEN()` returns a true table.

src/sqlfluff/rules/L025.py
```diff
@@ -78,8 +81,16 @@ def _eval(self, context: RuleContext) -> EvalResultType:
...
+                    # Don't warn or remove aliases for table functions, e.g.
+                    # Snowflake FLATTEN(). They are special because unlike real
+                    # "tables", they don't have a natural name to fall back to
+                    # if the alias is removed. Also, removing them can in turn
+                    # trigger false positives with L028.
+                    if not alias.table_expression or not has_table_function(
+                        alias.table_expression, context.dialect
+                    ):
+                        # Unused alias. Report and fix.
+                        violations.append(self._report_unused_alias(alias))
```
barrywhart, line 93: Important decision here -- we're not going to remove unused aliases for table functions, even if they're unused. I'm not sure this is the perfect solution, but it seems good enough for now.

tunetheweb:
> `UNNEST()` returns an ambiguous thing that is both a table and a column. On the other hand, I think `FLATTEN()` returns a true table.

If it is a true table, and the alias isn’t used, then isn’t it right to remove it?

As per discussion in #3178 @WittierDinosaur was happy L025 was flagging - just not happy L028 was flagging in `fix` even though not in `lint`.

src/sqlfluff/core/rules/analysis/select.py
```diff
@@ -114,31 +114,40 @@ def get_aliases_from_select(segment, dialect=None):
     table_aliases = []
     standalone_aliases = _get_pivot_table_columns(segment, dialect)
     for table_expr, alias_info in aliases:
-        if _has_value_table_function(table_expr, dialect):
+        if has_value_table_function(table_expr, dialect):
             if alias_info[0] not in standalone_aliases:
                 standalone_aliases.append(alias_info[0])
-        elif alias_info not in standalone_aliases:
+        elif alias_info not in table_aliases:
```
barrywhart, line 120: This is unrelated to the overall fix. It looks like a typo -- the intent here was to avoid adding duplicate entries to `table_aliases`.

barrywhart:
@WittierDinosaur: This issue required some tricky, fairly deep changes. Would you like to review and/or test this on your actual SQL?

barrywhart:
@tunetheweb: Can you review this again? It has grown and changed a lot since your initial review.

barrywhart:
@tunetheweb

Re:
>If it is a true table, and the alias isn’t used, then isn’t it right to remove it?

Please see the comments I added to L025. It acts like a table in the sense that it has multiple, named columns. It is **not** like a table in that if you remove the alias, there's no underlying table name you can use instead of the alias to access that data.

Once the alias is removed, L028 gets confused. It thinks there's only one possible source for columns, so it makes a bad fix. Do you have an alternate proposal that will make L025, L027, and L028 do something reasonable and not interact or interfere with one another in strange ways? Look back at the discussion between you and @WittierDinosaur on the related issue -- very confusing interaction. I think this is a workable solution, and I'm reluctant to change "just one thing" which re-introduces some of those other problems. What's "right" in isolation (one rule) isn't necessarily going to work well in combination with the other rules. 

If we don't have other ideas how to make all these things work well, I propose let's go with the approach here (which I believe is safe, even if it allows unused aliases to linger). Better to leave the SQL a little unclean than to break it! We can always revisit this later.

barrywhart:
I do have one other idea -- I'll branch off this branch, then create a second PR for consideration (if it works!).

barrywhart:
@tunetheweb: The other idea worked out so well that I just went ahead and updated this PR instead. 🙌 😎

test/fixtures/rules/std_rule_cases/L027.yml
```diff
@@ -257,3 +257,26 @@ test_pass_column_name_from_unnest_bigquery:
...
+  # FLATTEN() returns a table, thus there are two tables, thus lint failure,
+  # even though there's no alias provided for FLATTEN()
+  fail_str: |
+    SELECT
+        r.rec:foo::string AS foo,
+        value:bar::string AS bar
+    FROM foo.bar AS r, LATERAL FLATTEN(input => rec:result)
+  configs:
+    core:
+      dialect: snowflake
```
tunetheweb, line 282: Should we have a third test here?

```suggestion
test_fail_table_plus_flatten_snowflake_1:
  # FLATTEN() returns a table, thus there are two tables, thus lint failure
  fail_str: |
    SELECT
        r.rec:foo::string AS foo,
        value:bar::string AS bar
    FROM foo.bar AS r, LATERAL FLATTEN(input => rec:result) AS x
  configs:
    core:
      dialect: snowflake

test_fail_table_plus_flatten_snowflake_2:
  # FLATTEN() returns a table, thus there are two tables, thus lint failure,
  # even though there's no alias provided for FLATTEN()
  fail_str: |
    SELECT
        r.rec:foo::string AS foo,
        value:bar::string AS bar
    FROM foo.bar AS r, LATERAL FLATTEN(input => rec:result)
  configs:
    core:
      dialect: snowflake

test_fail_table_plus_flatten_snowflake_3:
  # FLATTEN() returns a table, thus there are two tables, thus lint failure,
  # even though there's no alias provided for FLATTEN()
  pass_str: |
    SELECT
        r.rec:foo::string AS foo,
        x.value:bar::string AS bar
    FROM foo.bar AS r, LATERAL FLATTEN(input => rec:result) AS x
  configs:
    core:
      dialect: snowflake
```

tunetheweb (approved):
LGTM - one small suggestion for an additional test case

test/fixtures/rules/std_rule_cases/L027.yml
barrywhart: Sure! I'll add that. I assume you want it to be a passing test case, in spite of "fail" in the suggested name and comment?

barrywhart: Added!

tunetheweb: Lazy copy and paste :-)

PR #3205, opened 2022-04-27
Alternate fix for issue 3178, *does* remove the alias from the FLATTEN() "table"

### LINKED ISSUE #3178: Snowflake: Interaction between L025 and L028 corrupts SQL

barrywhart:
Fixes #3178

Alternative to #3194. This version allows L025 to remove the unused alias, and it prevents L028 from making a bad fix by changing `dialect_ansi.FromExpressionElementSegment.get_eventual_alias()` to return an `AliasInfo` (rather than `None`). This in turn allows L028 to see that the query has two tables, thus it can't "fix" the column reference because of ambiguity (two tables, thus don't know which is being used).

### Pull Request checklist

  - Other.


barrywhart:
@tunetheweb: This is the other PR I mentioned. It implements the behavior you requested. If we like this approach, I'll need to do some cleanup before merging (basically removing some changes introduced in the original PR which would now be unnecessary).

barrywhart:
I like this PR better. I'm going to merge this branch into the original PR branch and close this one.

PR #3221, opened 2022-04-28
Fix bug where L003 and L036 fixes caused a parse error

barrywhart:
### Brief summary of the change made
Fixes #3179

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #3232, opened 2022-04-29
Enhance `apply_fixes()` to automatically fix violations of `can_start_end_non_code`

barrywhart:

About a month ago, I reworked the core linter to re-parse areas of the tree affected by fixes. This detects misbehaving rules so they can be fixed. One downside is that there's one kind of parse error that can be difficult to avoid: Generally, a segment's first and last children must be code or meta (for example, whitespace is not allowed).

This PR changes `BaseSegment.apply_fixes()` to automatically detect and fix these issues. It's a pretty simple and safe change, and I think it'll save future debugging and frustration for rule developers.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -1000,7 +1000,7 @@ class BaseSegment:
             else:
                 pre_nc = ()
                 post_nc = ()
-                idx_non_code = self._find_start_or_end_non_code()
+                idx_non_code = self._find_start_or_end_non_code(segments)
                 if idx_non_code is not None:  # pragma: no cover
                     raise ValueError(
                         f"Segment {self} {'starts' if idx_non_code == 0 else 'ends'} "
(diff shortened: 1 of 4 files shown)


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1000,21 +1000,12 @@ def parse(
             else:
                 pre_nc = ()
                 post_nc = ()
-                if (not segments[0].is_code) and (
```
barrywhart, line 1003: No changes in functionality in this area. Just refactored this code into separate functions for reuse elsewhere.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1090,7 +1081,22 @@ def parse(
 
         return self
 
-    def apply_fixes(self, dialect, rule_code, fixes):
+    @staticmethod
```
barrywhart, line 1084: Created these two functions from the code in the `parse()` function above.

test/fixtures/linter/autofix/snowflake/003_previously_parse_tree_damaging/after.sql
```diff
@@ -1,2 +1,3 @@
 set cutoff = (select foo
+
```
barrywhart, line 2: Previously, L036 was deleting this blank line not because we wanted to do this, but simply to avoid a parse error after the fix. Now that `apply_fixes()` is handling this, we can remove the hacky fix from L036, and the newline is automatically preserved, and there's no parse error anymore. 🎉

src/sqlfluff/rules/L036.py
```diff
@@ -391,30 +391,6 @@ def _fixes_for_move_after_select_clause(
                 ),
             ]
 
-            between_newline_and_select_target = select_children.select(
```
barrywhart, line 394: This code was added yesterday in #3221. Now it's not needed anymore.

tunetheweb (approved):

Issue #3242, opened 2022-05-02
Enhancement: Add option to provide Airflow builtins


dmohns:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

Airflow is commonly used to run and schedule SQL queries on various databases. It's even referenced in SQLFluff documentation https://docs.sqlfluff.com/en/stable/realworld.html#modularity

Currently, all users that want to integrate SQLFLuff into their Airflow code base have to go through the same process of understanding, how to

- Add custom Variables
- Add custom Macros
- Add custom Filters

into SQLFluff before they can start using SQLFluff. With the latter (custom Filters) currently not even possible.

It would be great if SQLFluff had first class support for Airflow. One idea would be to provide an option `apply_airflow_builtins` that acts similar as `apply_dbt_builtins`, but provides variables, macros and filters as per https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html

### Use case

Greatly increase the user experience of Airflow users, by reducing the number of customisation needed to rollout SQLFluff to Airflow code.

### Dialect

all

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

barrywhart:
@dmohns: Have you already defined these in your configuration? Care to share that? That would make it easier to add this.

dmohns:
I don't use `configuration` but a library to implement the Airflow variables and macros. This is because Airflow provided macros need to be in `macro` name space. And also to be able to more complex templating like `{{ task_instance.logical_date }}`.

What I have is pretty similar to the code in https://github.com/sqlfluff/sqlfluff/pull/3245 if you want to have a first glance.

dmohns:
Some preliminary work has been done on https://github.com/sqlfluff/sqlfluff/pull/3245

If someone wants to take this, they can take the PR as a starting point or source of ideas how to tackle this.

tdstark:
@dmohns I'm thinking about taking this. Any chance you could get me up to speed on what's left over?

dmohns:
Hey @tdstark yes, sure! I would love to see this getting picked up.

[...]

PR #3245, opened 2022-05-02
Jinja: Provide Airflow builtins

dmohns:

Makes progress on #3242 

This is a PoC and not yet ready to be released. However, I would like to gather some feedback if we should continue in this direction or not.

tbd.

### Pull Request checklist

  - Other.


src/sqlfluff/core/templaters/jinja.py
```diff
@@ -200,6 +204,152 @@ def __str__(self):  # pragma: no cover TODO?
...
+    @staticmethod
+    def _generate_airflow_builtins():
+        """Generate the airflow builtins which are injected in the context.
+
+        https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html
+        """
+        # Copied from _generate_dbt_builtins
+        # This feels a bit wrong defining these here, they should probably
+        # be configurable somewhere sensible. But for now they're not.
+        # TODO: Come up with a better solution.
```
barrywhart, line 216: Because this is a large amount of code, I would suggest creating a new module file, e.g. `src/sqlfluff/core/templaters/builtins/airflow.py` or similar.

The question about making it configurable is also a good one. Just thinking (not necessarily recommending), I wonder if we could structure this more like external code, i.e. the same directory/file structure used by `load_macros_from_path`. The more it resembles that structure, the easier it would be to make it configurable, either now or later.

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -200,6 +204,152 @@ def __str__(self):  # pragma: no cover TODO?
...
+
+            def __str__(self):
+                return self.dag_id
+
+        class DagRunEmulator:
+            """A class which emulates the `task` class from Airflow."""
+
+            dag_id = "sample_dag"
+            run_id = "sample_dag__run_id"
+            execution_date = datetime.datetime.now(datetime.timezone.utc)
```
barrywhart, line 232: Will SQLFluff users want or need to configure the value of individual object fields, both these dynamic ones as well as ones that you've hardcoded?

src/sqlfluff/core/templaters/jinja.py
dmohns: Yeah, putting it into a new module makes sense. I'll have a look at it.

dmohns: I tested around with loading the Airflow builtins exactly like an external library. There is a few problems with it still, i.e. I don't know yet, how I get this to work with Jinja filters, but in general I like this approach. WDYT?

barrywhart: Sorry for the slow reply -- I'm not receiving GitHub email notifications recently for some reason.

I like it! One suggested change -- since the Airflow functionality is specific to the Jinja templater, I'm thinking we may want to reorganize things a little into this structure:
src/sqlfluff/core/templaters/jinja_builtins/
src/sqlfluff/core/templaters/jinja_builtins/airflow/

This emphasizes that it's for Jinja, and creates a place for future builtins, e.g. dbt if we wanted to organize them the same way.

dmohns: No worries. I'm in no particular rush with this one. I moved it around and fixed few more bugs related to loading it as a library.

dmohns:
I have a few questions here, were I can use some help:

1. `mypy` is complaining at a line which I think is unrelated to my PR 🤔 Not really sure, what I need to do here.
2. Coverage is not quite happy about the modules. The lines it is complaining about are covered in the SQL. 

And a more general question: I'm looking for a way to access the SQLFluffConfig from my module. Anyone an idea how I could achieve this?

src/sqlfluff/core/templaters/jinja.py
dmohns: Of course it's thinkable that users have some very specific DAG setup, and do some filtering within Jinja on the `dag_id`s. However, I have rarely seen these being used in SQL in the first place. Most of the time they are used in INSERT statements or such.

My vision here is: I want to provide some builtins that covers "normal" use of the Airflow builtins. When user do something super crazy with the Airflow variables on template level, they can (and should) always fallback to providing them themselves via `.sqlfluff` config.

barrywhart:
Re `mypy`: I don't see anything obvious that would explain the mypy errors, and I'm not getting these errors on the latest `main` with the latest versions of `mypy` and `jinja2`. Sometimes these issues are triggered by a new release of mypy and/or a new release of some other package (e.g. `jinja2` in this case, since it's complaining about code that calls `jinja2` functions).

I updated your branch with the latest code from `main`. Let's see if the error still happens.

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -266,6 +270,21 @@ def get_context(self, fname=None, config=None, **kw) -> Dict:
                     if name not in live_context:
                         live_context[name] = dbt_builtins[name]
 
+            apply_airflow_builtins = config.get_section(
+                (self.templater_selector, self.name, "apply_airflow_builtins")
+            )
+            if apply_airflow_builtins:
+                import sqlfluff.core.templaters.jinja_builtins.airflow as airflow_builtins  # noqa
+
+                live_context.update(
```
barrywhart, line 279: You asked about how to provide your module with access to `FluffConfig`. Could you add a line here, something like:
```
live_context["fluff_config"] = config
```

(In terms of design quality, it would be better to pass only the required portion of the config, rather than passing everything, so the above is just a starting point.)

tunetheweb:
Is this still being worked on @dmohns ?

PR #3259, opened 2022-05-04
Bug fix: L036 handle single-column `SELECT` with comment on same line as `SELECT` keyword

### LINKED ISSUE #3252: L036 - sqlfluff fix will remove a single SELECT target on a new line if there's a comment after the SELECT
### Search before asking

### What Happened

If a violation of the rule **L036** (_Select targets should be on a new line unless there is only one select target_) is identified, with a single SELECT target on a new line but there's a double hyphen comment after the SELECT statement, `sqlfluff fix` will remove the new line resulting in the select target being appended to the end of the comment. 

### Expected Behaviour

For the following SQL query:

```SQL
SELECT --some comment
CASE WHEN a THEN 1 END;
```

I would expect the violation of L036 to be ignored, perhaps with a warning about the comment after the SELECT statement

```SQL
SELECT --some comment
  CASE WHEN a THEN 1 END;
```

### Observed Behaviour

However the following is observed:

```SQL
SELECT --some comment
CASE WHEN a THEN 1 END;
```

```SQL
SELECT --some comment CASE WHEN a THEN 1 END;
```

### How to reproduce

```bash
$ echo "SELECT --some comment\nCASE WHEN a THEN 1 END;" > test.sql
$ sqlfluff fix --dialect snowflake test.sql
```

### Dialect

Tested with:
**snowflake**
**sqlite**
**bigquery**

### Version

```bash
$ sqlfluff --version
sqlfluff, version 0.13.0
```

```bash
$ python --version
Python 3.9.6
```

### Configuration

Tested with default configurations but also with the following **.sqlfluff** config:

```
# https://docs.sqlfluff.com/en/stable/configuration.html#default-configuration
verbose = 0
nocolor = False
dialect = snowflake
templater = jinja
rules = None
exclude_rules = L011,L016,L031,L034,L035
recurse = 0
output_line_length = 120
runaway_limit = 10
ignore = None
ignore_templated_areas = True
encoding = autodetect
disable_noqa = False
fix_even_unparsable = False

[sqlfluff:indentation]
indented_joins = False
indented_ctes = False
indented_using_on = True
template_blocks_indent = True

[sqlfluff:templater]
unwrap_wrapped_queries = True

[sqlfluff:templater:jinja]
apply_dbt_builtins = True

[sqlfluff:rules]
tab_space_size = 2
max_line_length = 120
indent_unit = space
comma_style = leading
allow_scalar = True

# https://docs.sqlfluff.com/en/stable/rules.html
[sqlfluff:rules:L003]
hanging_indents = False

operator_new_lines = after

capitalisation_policy = upper

[sqlfluff:rules:L011]  # Table aliasing
aliasing = implicit

[sqlfluff:rules:L012]  # Column aliasing
aliasing = explicit

[sqlfluff:rules:L013]  # Column expressions
allow_scalar = False

[sqlfluff:rules:L016]
ignore_comment_lines = False

[sqlfluff:rules:L038]

[sqlfluff:rules:L042]
# By default, allow subqueries in FROM clauses, but not JOIN clauses.
forbid_subquery_in = join

prefer_count_1 = False
prefer_count_0 = False

[sqlfluff:rules:L052]  # Statements must end with a semi-colon
multiline_newline = True

[sqlfluff:rules:L059] # Unnecessary quoted identifier
ignore_words = date%
```

### Code of Conduct

barrywhart:
Fixes #3252

### Pull Request checklist

  - Other.


src/sqlfluff/rules/L036.py
```diff
@@ -301,12 +315,7 @@ def _fixes_for_move_after_select_clause(
                         ]
                         fixes_.append(
                             LintFix.create_after(
-                                self._choose_anchor_segment(
-                                    context,
-                                    "create_after",
-                                    select_clause[0],
-                                    filter_meta=True,
-                                ),
+                                select_clause[0],
```
barrywhart, line 318: Unrelated to the issue, but no longer necessary to call `_choose_anchor_segment()` here due to recent improvements in the core linter.

tunetheweb:
Instead of moving the select target _up_, before the comment, could we move the comment _down_ after the select target (or temporary delete it)? Then fix the select target as it would if not comment was there?

src/sqlfluff/rules/L036.py
```diff
@@ -384,15 +392,25 @@ def _fixes_for_move_after_select_clause(
                             ],
                         )
 
-            fixes += [
```
barrywhart, line 387: Review this section with "Hide whitespace" enabled.

barrywhart:
@tunetheweb: We could. I don't think I'm up for tackling it now. Wna tocreate a separate issue for this? Often when I look at one of these rules that moves stuff around, I am surprised how complicated it is and how many special cases arise, and how easy it is to make a mistake and corrupt the code. Here are a couple of other test cases that occur to me if we try to handle this:
```
SELECT -- comment 1
a -- comment 2;

SELECT --comment 1
/* comment 2 */ a -- comment 3;
```

🤯

tunetheweb:
Fair enough!

tunetheweb (approved):

PR #3274, opened 2022-05-06
Support ignoring complete schemas

### LINKED ISSUE #3226: ignore_words_regex results in error
### Search before asking

### What Happened

I'm trying to use the new `ignore_words_regex` in rule L014. The regex is similar to `^(SchemaA.*)$` (so I want it to ignore all tables in a certain schema.

### Expected Behaviour

Everything is follows the naming policy, but identifiers starting with `SchemaA` are not changed.

### Observed Behaviour

```
CRITICAL   [L014] Applying rule L014 threw an Exception: string index out of range
  File "/usr/local/lib/python3.8/dist-packages/sqlfluff/core/rules/base.py", line 634, in crawl
  File "/usr/local/lib/python3.8/dist-packages/sqlfluff/rules/L014.py", line 87, in _eval
    return super()._eval(context=context)
  File "/usr/local/lib/python3.8/dist-packages/sqlfluff/rules/L010.py", line 120, in _eval
    if context.segment.raw[0] != context.segment.raw[0].upper():
IndexError: string index out of range
```

Additionally, identifiers starting with `SchemaA` are changed.

### How to reproduce

see above

### Dialect

T-SQL

### Version

Latest `main`

### Configuration

[sqlfluff]
dialect = tsql

[sqlfluff:rules:L014] # Unquoted identifiers
extended_capitalisation_policy = pascal
ignore_words_regex = "^(Landing.*)$"

### Code of Conduct

fdw:

Now you should be able to exclude certain schemas from re-formatting

Fixes #3226

I don't think so, but this is my first change in the rules...

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -71,6 +71,7 @@ capitalisation_policy = consistent
 # Comma separated list of words to ignore for this rule
 ignore_words = None
 ignore_words_regex = None
+ignore_schemas = None
 
 [sqlfluff:rules:L011]
 # Aliasing preference for tables
@@ -86,6 +87,7 @@ extended_capitalisation_policy = consistent
 # Comma separated list of words to ignore for this rule
 ignore_words = None
 ignore_words_regex = None
+ignore_schemas = None
 
 [sqlfluff:rules:L016]
 # Line length
@@ -109,6 +111,7 @@ quoted_identifiers_policy = none
 # Comma separated list of words to ignore for this rule
 ignore_words = None
 ignore_words_regex = None
+ignore_schemas = None
 
 [sqlfluff:rules:L030]
 # Function names
@@ -116,6 +119,7 @@ extended_capitalisation_policy = consistent
 # Comma separated list of words to ignore for this rule
 ignore_words = None
 ignore_words_regex = None
+ignore_schemas = None
 
 [sqlfluff:rules:L031]
 # Avoid table aliases in from clauses and join conditions.
@@ -132,6 +136,7 @@ capitalisation_policy = consistent
 # Comma separated list of words to ignore for this rule
 ignore_words = None
 ignore_words_regex = None
+ignore_schemas = None
 
 [sqlfluff:rules:L042]
 # By default, allow subqueries in from clauses, but not join clauses
@@ -163,12 +168,14 @@ allow_space_in_identifier = False
 additional_allowed_characters = ""
 ignore_words = None
 ignore_words_regex = None
+ignore_schemas = None
 
 [sqlfluff:rules:L059]
 # Policy on quoted and unquoted identifiers
 prefer_quoted_identifiers = False
 ignore_words = None
 ignore_words_regex = None
+ignore_schemas = None
 force_enable = False
 
 [sqlfluff:rules:L062]
@@ -181,6 +188,7 @@ extended_capitalisation_policy = consistent
 # Comma separated list of words to ignore for this rule
 ignore_words = None
 ignore_words_regex = None
+ignore_schemas = None
 
 [sqlfluff:rules:L064]
 # Consistent usage of preferred quotes for quoted literals
--- a/src/sqlfluff/core/rules/config_info.py
+++ b/src/sqlfluff/core/rules/config_info.py
@@ -95,6 +95,9 @@ STANDARD_CONFIG_INFO_DICT = {
             " when linting line lengths?"
         ),
     },
+    "ignore_schemas": {
+        "definition": ("Comme separated list of schemas to ignore from rule"),
+    },
     "ignore_words": {
         "definition": ("Comma separated list of words to ignore from rule"),
     },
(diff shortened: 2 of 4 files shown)


src/sqlfluff/core/rules/config_info.py
```diff
@@ -95,6 +95,9 @@
             " when linting line lengths?"
         ),
     },
+    "ignore_schemas": {
+        "definition": ("Comme separated list of schemas to ignore from rule"),
```
tunetheweb, line 99:
```suggestion
        "definition": ("Comma separated list of schemas to ignore from rule"),
```

src/sqlfluff/rules/L010.py
```diff
@@ -9,6 +9,7 @@
     document_fix_compatible,
     document_configuration,
 )
+from sqlfluff.dialects import dialect_ansi as ansi
```
tunetheweb, line 12: Why do we need this import?

```suggestion
```

test/fixtures/rules/std_rule_cases/L010.yml
```diff
@@ -177,3 +177,11 @@ test_fail_bigquery_week:
...
       L010:
         capitalisation_policy: upper
+
+test_pass_ignore_schema:
+  pass_str: |
+    SELECT ignoreme.AhAhAhAh FROM dbo
+  configs:
+    rules:
+      L010:
+        ignore_schemas: ignoreme
```
tunetheweb, line 187: I don't see how this test is a valid test. L010 is about capitalisation of keywords but `ignoreme` is not a keyword.

src/sqlfluff/core/default_config.cfg
```diff
@@ -71,6 +71,7 @@ capitalisation_policy = consistent
 # Comma separated list of words to ignore for this rule
 ignore_words = None
 ignore_words_regex = None
+ignore_schemas = None
```
tunetheweb, line 74: Is this valid for L010? Or do you want to ignore all keywords capitalisation rules for all statements with a `FROM` clause using this schema? That's a MUCH bigger ask than I was understanding...

I thought you only wanted it for L014?

tunetheweb (changes requested):
I'm not sure I understand the intent of this PR...

fdw:
> I'm not sure I understand the intent of this PR...

Oh, I'm sorry!
This is my attempt to implement the changes suggested by you in #3226. The intent is to ignore words by their qualifying schema (f.e. I want to not correct the capitalisation for tables in the schema `weird`). Does that make sense?

src/sqlfluff/rules/L010.py
fdw: I'm later checking whether the parent is of type ObjectReferenceStatement which is defined in dialect_ansi. Would you prefer a different import?

test/fixtures/rules/std_rule_cases/L010.yml
fdw: The idea was - and maybe I misunderstood something - that AhAhAhAh is not changed as it is in the schema that I want to ignore.

src/sqlfluff/core/default_config.cfg
fdw: I originally only wanted this for L014, but looking at the code, it seemed easier (and more consistent) to do it in the parent rule. It should work for all that descend from L010, as far as I can see.

fdw:
I don't want to annoy you, so if you don't like the PR or the intent behind it, feel free to close it. Maybe we can find another solution :)

src/sqlfluff/rules/L010.py
tunetheweb: OK but it might not be the ANSI version we want if any dialect overrides it.

Have a look at `_get_table_refs` in L026 which also seems to want the same info as you want, but doesn't do this ANSI import.

test/fixtures/rules/std_rule_cases/L010.yml
tunetheweb: This test should be in L014.yml as that's the rule that deals with capitalisation of unquoted identifiers, not in L010.yml (which deals with capitalisation of keywords).

src/sqlfluff/core/default_config.cfg
tunetheweb: I don't think the wording `ignore_schema` fits for L010 (or any of its other descendent rules). That is only relevant for L014.

So either we should make this L014-specific or choose a better name (`enhance_ignores`?)

Personally I'd prefer this to be L014-specific.

tunetheweb:
Don't worry about annoying me! 😁

I was just confused with your decision to do this in L010 and don't think your test case is valid for L010.

I really think this is L014-specific so should be coded as such, and PR title should make that clear ("Support ignoring complete schemas in L014 using `ignore_regex`")

fdw:
Looking through this, I see why you'd prefer it in L014. However, L063 (capitalisation of datatypes) could use this as well (and I would actually like it in there). What do you suggest?

tunetheweb:
I'm happy to have in whatever rules it makes sense.

Just the current test in L010 is not a good test as will not fail no matter whether this is configured or not.

tunetheweb:
FYI @fdw I'm aiming for a release this Friday. So would be good to get this finished out to allow it to be included in that.

fdw:
Thanks for the heads-up! I'll try to work on it, but I can't make any promises, sorry.

PR #3279, opened 2022-05-06
L046 (spaces around Jinja tags) should check all slices in a segment

### LINKED ISSUE #3275: L046: Only triggers for first Jinja tag in a compound statement with multiple tags
### Search before asking

### What Happened

When linting a SQL with a given statement consists of multiple Jinja tags, only the first Jinja tags is subject to L046.

See an example for BigQuery

```sql
CREATE TABLE `{{ project }}.{{ dataset }}.{{table}}`;
```

--> No Lint error

However, for 

```sql
CREATE TABLE `{{project}}.{{ dataset }}.{{table}}`;
```

```
L:   1 | P:  14 | L046 | Jinja tags should have a single whitespace on either
                       | side.
```

### Expected Behaviour

I would expect an L046 to be raised because `{{table}}` is clearly violating L046.

### Observed Behaviour

No lint error reported.

### How to reproduce

Run `sqlfluff lint --dialect bigquery 1.sql` on the example above

### Dialect

bigquery (but probably not related to bigquery)

### Version

latest `main` from Github

### Configuration

project=project_id
dataset=dataset_id
table=table_name

### Code of Conduct

barrywhart:
Fixes #3275

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L046.py
+++ b/src/sqlfluff/rules/L046.py
@@ -1,7 +1,7 @@
 """Implementation of Rule L046."""
 from typing import Tuple
 
-from sqlfluff.core.rules.base import BaseRule, LintResult, RuleContext
+from sqlfluff.core.rules.base import BaseRule, EvalResultType, LintResult, RuleContext
 from sqlfluff.core.rules.functional import rsp
 from sqlfluff.core.rules.doc_decorators import document_groups
 
@@ -54,18 +54,24 @@ class Rule_L046(BaseRule):
         pos = main.find(inner)
         return main[:pos], inner, main[pos + len(inner) :]
 
-    def _eval(self, context: RuleContext) -> LintResult:
+    def _eval(self, context: RuleContext) -> EvalResultType:
         """Look for non-literal segments."""
         assert context.segment.pos_marker
         if context.segment.is_raw() and not context.segment.pos_marker.is_literal():
+            if not context.memory:
+                memory = set()
+            else:
+                memory = context.memory
+
             # Does it actually look like a tag?
             templated_raw_slices = context.functional.segment.raw_slices.select(
                 rsp.is_slice_type("templated")
             )
+            result = []
             for raw_slice in templated_raw_slices:
                 src_raw = raw_slice.raw
                 if not src_raw or src_raw[0] != "{" or src_raw[-1] != "}":
-                    return LintResult(memory=context.memory)  # pragma: no cover
+                    continue  # pragma: no cover
 
                 # Dedupe using a memory of source indexes.
                 # This is important because several positions in the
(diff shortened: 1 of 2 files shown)


src/sqlfluff/rules/L046.py
```diff
@@ -54,30 +54,32 @@ def _get_whitespace_ends(s: str) -> Tuple[str, str, str]:
...
+            if not context.memory:
+                memory = set()
+            else:
+                memory = context.memory
+
             # Does it actually look like a tag?
             templated_raw_slices = context.functional.segment.raw_slices.select(
                 rsp.is_slice_type("templated")
             )
+            result = []
```
barrywhart, line 70: Basically, this fix changes the rule so it loops over **all** the slices of a segment, rather than `return`-ing after the first.

src/sqlfluff/rules/L046.py
```diff
@@ -88,10 +90,19 @@ def _eval(self, context: RuleContext) -> LintResult:
 
                 # Check the initial whitespace.
                 if not ws_pre or (ws_pre != " " and "\n" not in ws_pre):
-                    return LintResult(memory=memory, anchor=context.segment)
+                    result.append(
+                        LintResult(
+                            memory=memory,
+                            anchor=context.segment,
+                            description=f"Jinja tags should have a single "
```
barrywhart, line 97: Updated the result to include the problem code, because there can be multiple Jinja tags in the same segment, but the position (line number and character position) will always point to the **start** of the segment, which could be confusing. (See the YAML test case -- it'll be clearer then.)

src/sqlfluff/rules/L046.py
```diff
@@ -54,30 +54,32 @@ def _get_whitespace_ends(s: str) -> Tuple[str, str, str]:
...
         return main[:pos], inner, main[pos + len(inner) :]
 
-    def _eval(self, context: RuleContext) -> LintResult:
+    def _eval(self, context: RuleContext) -> EvalResultType:
         """Look for non-literal segments."""
         assert context.segment.pos_marker
         if context.segment.is_raw() and not context.segment.pos_marker.is_literal():
+            if not context.memory:
+                memory = set()
+            else:
```
tunetheweb, line 63: I know you didn't change this, and just copied it from below, but I'm curious why the other rules using `memory` do not have to initialise it like this?

tunetheweb (approved):
LGTM. One question though, not directly related to this change.

src/sqlfluff/rules/L046.py
barrywhart: IIRC, memory defaults to an empty **dictionary**, so I'm guessing it's because this rule uses a **set** instead.

Issue #3188, opened 2022-04-26
Support MySQL GRANT / CREATE USER statements


acdha:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

We have some SQL files which define service users and their privileges. Currently sqlfluff will report unparsable sections for each of these lines as well as common variations (e.g. using `IDENTIFIED BY 'password'` instead of the AWS RDS IAM extension, granting only a single permission, etc.:

```sql
CREATE USER 'prj_svc' IDENTIFIED WITH AWSAuthenticationPlugin AS 'RDS';
GRANT INSERT, UPDATE, DELETE, SELECT, REFERENCES ON prj_table TO 'prj_svc';
FLUSH PRIVILEGES;
```

### Use case

We're following the process here: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.DBAccounts.html#UsingWithRDS.IAMDBAuth.DBAccounts.MySQL

In a larger context, this is similar to #901 where I'd like to be able to have sqlfluff enabled for *.sql without exclude rules.

### Dialect

MySQL

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

PR #3286, opened 2022-05-08
MySQL: Support account names

greg-finley:
Makes more progress on https://github.com/sqlfluff/sqlfluff/issues/3188. Support all the myriad variations in `SingleIdentifierGrammar`.

These really do work in MySQL!
```
mysql> GRANT INSERT, UPDATE, DELETE, SELECT, REFERENCES ON prj_table TO prj_svc;
Query OK, 0 rows affected (0.00 sec)

mysql> GRANT INSERT, UPDATE, DELETE, SELECT, REFERENCES ON prj_table TO 'prj_svc';
Query OK, 0 rows affected (0.00 sec)

mysql> GRANT INSERT, UPDATE, DELETE, SELECT, REFERENCES ON prj_table TO "prj_svc";
Query OK, 0 rows affected (0.00 sec)

mysql> GRANT INSERT, UPDATE, DELETE, SELECT, REFERENCES ON prj_table TO `prj_svc`;
Query OK, 0 rows affected (0.00 sec)

mysql> GRANT INSERT, UPDATE, DELETE, SELECT, REFERENCES ON prj_table TO 'prj_svc'@'%';
Query OK, 0 rows affected (0.00 sec)

mysql> GRANT INSERT, UPDATE, DELETE, SELECT, REFERENCES ON prj_table TO "prj_svc"@"%";
Query OK, 0 rows affected (0.00 sec)

mysql> GRANT INSERT, UPDATE, DELETE, SELECT, REFERENCES ON prj_table TO `prj_svc`@`%`;
Query OK, 0 rows affected (0.00 sec)
```

It could make other stuff mistakenly parse to `SingleIdentifierGrammar` in MySQL.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -577,6 +568,19 @@ class IntervalExpressionSegment(BaseSegment):
...
+    https://dev.mysql.com/doc/refman/8.0/en/account-names.html
+    https://dev.mysql.com/doc/refman/8.0/en/role-names.html
+"""
+single_identifier_segment = ansi_dialect.get_grammar("SingleIdentifierGrammar").copy(
+    insert=[
+        Ref("SessionVariableNameSegment"),
+        Ref("SingleQuotedIdentifierSegment"),
+        Ref("DoubleQuotedLiteralSegment"),
+    ]
+)
```
greg-finley, line 582: I feel I should have ended up with a `SingleIdentifierSegment` that I can `Ref` below, but I'm not sure how to do this with `get_grammar`.

greg-finley: This works:

```python
class SingleIdentifierSegment(BaseSegment):
    """A SingleIdentifierSegment. Override the default SingleIdentifierGrammar to allow
    two of these segments separated by an at sign.

    https://dev.mysql.com/doc/refman/8.0/en/account-names.html
    https://dev.mysql.com/doc/refman/8.0/en/role-names.html
    """

    type = "single_identifier"
    match_grammar: Matchable = ansi_dialect.get_grammar("SingleIdentifierGrammar").copy(
        insert=[
            Ref("SessionVariableNameSegment"),
            Ref("DoubleQuotedLiteralSegment"),
        ]
    )
```

but it diffs all the MySQL YML test files.

tunetheweb: That's fine. The point of the YAMLs being in git, is so we can ensure there are no unexpected changes. This is expected.

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -524,15 +524,6 @@ class IntervalExpressionSegment(BaseSegment):
...
 mysql_dialect.add(
-    DoubleForwardSlashSegment=StringParser(
-        "//", SymbolSegment, name="doubleforwardslash", type="statement_terminator"
-    ),
-    DoubleDollarSignSegment=StringParser(
-        "$$", SymbolSegment, name="doubledollarsign", type="statement_terminator"
-    ),
-    AtSignSignSegment=StringParser(
-        "@", SymbolSegment, name="at_sign", type="user_designator"
-    ),
```
tunetheweb, line 535: Why are these no longer needed?

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -577,6 +568,27 @@ class IntervalExpressionSegment(BaseSegment):
...
+    type = "single_identifier"
+    match_grammar: Matchable = ansi_dialect.get_grammar("SingleIdentifierGrammar").copy(
+        insert=[
+            Ref("SessionVariableNameSegment"),
+            Ref("SingleQuotedIdentifierSegment"),
+            Ref("DoubleQuotedLiteralSegment"),
+        ]
+    )
+
+
```
tunetheweb, line 591: Oh I see what you meant earlier now. There's a double `identifier`. No I don't think we should do that.

Unless you explicitly want the label, this should be a grammar, rather than a class with a `type`.

But I don't think it's worth having either here. So I'd delete this.

```suggestion
```

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -585,8 +597,13 @@ class IntervalExpressionSegment(BaseSegment):
...
     ),
-    SingleIdentifierGrammar=ansi_dialect.get_grammar("SingleIdentifierGrammar").copy(
-        insert=[Ref("SessionVariableNameSegment")]
+    SingleIdentifierGrammar=Sequence(
+        Ref("SingleIdentifierSegment"),
+        Sequence(
+            Ref("AtSignLiteralSegment"),
+            Ref("SingleIdentifierSegment"),
+            optional=True,
+        ),
```
tunetheweb, line 606: I'd just define the full thing here, even though it's a bit of duplication:

```suggestion
    SingleIdentifierGrammar=Sequence(
        OneOf(
            Ref("NakedIdentifierSegment"),
            Ref("QuotedIdentifierSegment"),
            Ref("SessionVariableNameSegment"),
            Ref("SingleQuotedIdentifierSegment"),
            Ref("DoubleQuotedLiteralSegment"),
        ),
        Sequence(
            Ref("AtSignLiteralSegment"),
            OneOf(
                Ref("NakedIdentifierSegment"),
                Ref("QuotedIdentifierSegment"),
                Ref("SessionVariableNameSegment"),
                Ref("SingleQuotedIdentifierSegment"),
                Ref("DoubleQuotedLiteralSegment"),
            ),
            optional=True,
        ),
```

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_mysql.py
greg-finley: They weren't referenced by anything, and, in the case of at sign, was redundant.

src/sqlfluff/dialects/dialect_mysql.py
greg-finley: Great, thanks for the suggestion!

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -585,8 +576,28 @@ class IntervalExpressionSegment(BaseSegment):
...
+            Sequence(
+                Ref("AtSignLiteralSegment"),
+                OneOf(
+                    Ref("NakedIdentifierSegment"),
+                    Ref("QuotedIdentifierSegment"),
+                    Ref("SingleQuotedIdentifierSegment"),
+                    Ref("DoubleQuotedLiteralSegment"),
+                ),
+                optional=True,
+            ),
```
tunetheweb, line 599: If this sequence is only valid for account names and roles is it really necessary to put them everywhere a Single Identifier is valid?

Would, it not be better to create a new Grammar (or Grammars) for these and just reference them in any `ObjectReferenceSegment` instead? Or, better yet, update `AccessStatementSegment` to reference them?

tunetheweb (commented):
The changes look good and it works.

Just one question on approach?

src/sqlfluff/dialects/dialect_mysql.py
greg-finley: Agree that it would be better to limit the scope. In addition to granting access, this same thing exists for `CREATE USER` (and maybe more): https://dev.mysql.com/doc/refman/8.0/en/create-user.html . That makes me think `ObjectReferenceSegment` might be best?

I wasn't sure how to override `AccessStatementSegment`. Ideally I'd want to keep all the complexity already in ANSI and just change this one part:
https://github.com/sqlfluff/sqlfluff/blob/3de129150dc758fac709fde51f926bb34ce7e9ea/src/sqlfluff/dialects/dialect_ansi.py#L3032
Does that one part become a grammar in ANSI that gets overwritten in MySQL? Or is there a different way to change this one part for MySQL?

tunetheweb: Yeah pulling a piece of to its own Grammar and overriding just that is the best way to have a small change to a more complex statement.

The alternative would be to build this in `ObjectReferenceSegment` instead of `SingleIdentifierGrammar`? That would at least limit the scope a little (and prevent the duplication currently in this PR).

Or we could just merge as is. It works and isn’t completely without precedent as a few of the dialects have more complex structures in `SingleIdentifierGrammar`.

WDYT? I think I’m veering towards the middle section (defining this in `ObjectReferenceSegment`

greg-finley: I'll give `ObjectReferenceSegment` a shot 👍

greg-finley: I think the resulting changes to `x_definer.yml` parsings is actually better after https://github.com/sqlfluff/sqlfluff/pull/3286/commits/da6143e9fb8628c86b253d993234332bcd9d7350

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -602,6 +594,35 @@ class IntervalExpressionSegment(BaseSegment):
...
 
 
+class ObjectReferenceSegment(ansi.ObjectReferenceSegment):
+    """A reference to an object.
+
+    https://dev.mysql.com/doc/refman/8.0/en/account-names.html
+    https://dev.mysql.com/doc/refman/8.0/en/role-names.html
+    """
+
+    match_grammar: Matchable = Sequence(
```
tunetheweb, line 604: This is a lot more complicated in ANSI with `DotSegments` that you've lost now. Do you need them too?

greg-finley: Ah, my bad ... is this the proper thing to do? https://github.com/sqlfluff/sqlfluff/pull/3286/commits/19bafe55e9bb124216577215bb3a5a0dd8ae4f3e

tunetheweb: I would have redefined the whole thing. But this is much cleverer!

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -602,6 +594,36 @@ class IntervalExpressionSegment(BaseSegment):
...
+            Sequence(
+                Ref("AtSignLiteralSegment"),
+                OneOf(
+                    Ref("NakedIdentifierSegment"),
+                    Ref("QuotedIdentifierSegment"),
+                    Ref("SingleQuotedIdentifierSegment"),
+                    Ref("DoubleQuotedLiteralSegment"),
+                ),
+                optional=True,
+            ),
```
tunetheweb, line 622: I presume should be no spaces? So might wanna make that obvious:

```suggestion
                optional=True,
                allow_gaps=False,
            ),
            allow_gaps=False,
```

tunetheweb (commented):
One more small change and I think we're there!

src/sqlfluff/dialects/dialect_mysql.py
greg-finley: Ummmm .... it's ok with the first but not the second 🙃  However, probably good to just enforce no gaps anyways.


```mysql
mysql> GRANT INSERT, UPDATE, DELETE, SELECT, REFERENCES ON prj_table TO `prj_svc` @`%`;
Query OK, 0 rows affected (0.00 sec)

mysql> GRANT INSERT, UPDATE, DELETE, SELECT, REFERENCES ON prj_table TO `prj_svc` @ `%`;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '`%`' at line 1
```

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -602,6 +594,36 @@ class IntervalExpressionSegment(BaseSegment):
 )
 
 
+class ObjectReferenceSegment(ansi.ObjectReferenceSegment):
+    """A reference to an object.
+
+    https://dev.mysql.com/doc/refman/8.0/en/account-names.html
+    https://dev.mysql.com/doc/refman/8.0/en/role-names.html
+    """
```
tunetheweb, line 602:
```suggestion
    Override ANSI to allow account names and role names as objects as well:
    https://dev.mysql.com/doc/refman/8.0/en/account-names.html
    https://dev.mysql.com/doc/refman/8.0/en/role-names.html
    """
```

src/sqlfluff/dialects/dialect_mysql.py
tunetheweb: OK than can just enforce the first one and not the second (note they are in reverse order to your example):

```suggestion
                optional=True,
                allow_gaps=False,
            ),
```

tunetheweb (changes requested):
As discussed in Slack, let’s take a different approach, to limit this further to just where it’s suppose to be used rather than every Object:

Let’s create a `RoleReferenceSegment` in ANSI and update `AccessStatementSegment`. `CreateUserStatementSegment`…etc. to use that instead of `ObjectReferenceSegment`. Might also need to check other dialects to see if they require similar changes.

Can then add this syntax in MySQL’s version of `RoleReferenceSegment` or, maybe it’s actually good to define this in ansi’s `RoleReferenceSegment` anyway since it’s fairly common to lots of dialects?

greg-finley:
I extended `RoleReferenceSegment` in MySQL and it seems good. I tried putting the logic into ANSI itself, but it broke a lot of stuff. I think this is ok for now?

tunetheweb (approved):
LGTM!

There's a few other cleanups for the other dialects but this'll do for this PR.

src/sqlfluff/dialects/dialect_mysql.py
tunetheweb: Actually I don't think you did need these. DotSegments make sense for other references (tables, views...etc.) but not for users I don't think. Cleaning up in #3287 along with a few other changes.

PR #3289, opened 2022-05-09
MySQL: Support `Create User`

### LINKED ISSUE #3188: Support MySQL GRANT / CREATE USER statements

greg-finley:
Fleshes out `Create User` per the MySQL doc. Tests various cases, including all samples from [the MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/create-user.html).

Closes #3188

I locally solved for literal strings can be either single or double quoted in MySQL (and stuff like users can also be backticked, as we saw [here](https://github.com/sqlfluff/sqlfluff/pull/3286)), but we should file a separate issue for a more holistic solution.
```python
    _string_literal = OneOf(
        Ref("QuotedLiteralSegment"), Ref("DoubleQuotedLiteralSegment")
    )
```

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_mysql.py
+++ b/src/sqlfluff/dialects/dialect_mysql.py
@@ -321,20 +321,46 @@ class CreateUserStatementSegment(ansi.CreateUserStatementSegment):
     https://dev.mysql.com/doc/refman/8.0/en/create-user.html
     """
 
-    auth_option = Sequence(
+    _random_password = Sequence("RANDOM", "PASSWORD")
+    _auth_plugin = Ref("ObjectReferenceSegment")
+    _auth_string = Ref("SingleQuotedIdentifierSegment")
+
+    _initial_auth_option = Sequence(
+        "INITIAL",
+        "AUTHENTICATION",
+        "IDENTIFIED",
+        OneOf(Sequence("BY", OneOf(_random_password, _auth_string))),
+        Sequence(
+            "WITH",
+            _auth_plugin,
+            "AS",
+            _auth_string,
+        ),
+    )
+
+    _auth_option = Sequence(
         "IDENTIFIED",
         OneOf(
             Sequence(
                 "BY",
-                OneOf(
-                    Sequence("RANDOM", "PASSWORD"),
-                    Ref("SingleQuotedIdentifierSegment"),
-                ),
+                OneOf(_random_password, _auth_string),
             ),
             Sequence(
                 "WITH",
-                Ref("ObjectReferenceSegment"),
-                Sequence("AS", Ref("SingleQuotedIdentifierSegment"), optional=True),
+                _auth_plugin,
+                Sequence(
+                    OneOf(
+                        Sequence(
+                            "BY",
+                            OneOf(
+                                _random_password, Ref("SingleQuotedIdentifierSegment")
+                            ),
+                        ),
+                        Sequence("AS", Ref("SingleQuotedIdentifierSegment")),
+                        _initial_auth_option,
+                    ),
+                    optional=True,
+                ),
             ),
         ),
     )
@@ -345,7 +371,7 @@ class CreateUserStatementSegment(ansi.CreateUserStatementSegment):
         Ref("IfNotExistsGrammar", optional=True),
         Delimited(
             Sequence(
-                Ref("RoleReferenceSegment"), Delimited(auth_option, delimiter="AND")
+                Ref("RoleReferenceSegment"), Delimited(_auth_option, delimiter="AND")
             ),
             delimiter=Ref("CommaSegment"),
         ),
--- a/src/sqlfluff/dialects/dialect_mysql.py
+++ b/src/sqlfluff/dialects/dialect_mysql.py
@@ -321,9 +321,13 @@ class CreateUserStatementSegment(ansi.CreateUserStatementSegment):
     https://dev.mysql.com/doc/refman/8.0/en/create-user.html
     """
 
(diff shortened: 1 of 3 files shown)


greg-finley:
Hmm, I'm not sure why the CI failure. Locally the generation yields no diff:

```shell
(.venv) gregoryfinley@Gregorys-MacBook-Pro sqlfluff % time python3  test/generate_parse_fixture_yml.py                
Match Pattern Received:
        filter=* dialect=all new-only=False
Found 975 file(s) to generate
Fixture built: 975
python3 test/generate_parse_fixture_yml.py  158.24s user 3.23s system 701% cpu 23.030 total
(.venv) gregoryfinley@Gregorys-MacBook-Pro sqlfluff % git status
On branch mysql-create-user
Your branch is up to date with 'greg/mysql-create-user'.

nothing to commit, working tree clean
(.venv) gregoryfinley@Gregorys-MacBook-Pro sqlfluff % 
```

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -314,6 +330,142 @@ class CreateTableStatementSegment(ansi.CreateTableStatementSegment):
...
 
+class CreateUserStatementSegment(ansi.CreateUserStatementSegment):
+    """`CREATE USER` statement.
+
+    https://dev.mysql.com/doc/refman/8.0/en/create-user.html
+    """
+
+    _string_literal = OneOf(
+        Ref("QuotedLiteralSegment"), Ref("DoubleQuotedLiteralSegment")
+    )
```
tunetheweb, line 341: Should this be globally available to all of MySQL?

In fact should it be added to `QuotedLiteralSegment`?

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -314,6 +330,142 @@ class CreateTableStatementSegment(ansi.CreateTableStatementSegment):
...
+
+    _random_password = Sequence("RANDOM", "PASSWORD")
+    _auth_plugin = Ref("ObjectReferenceSegment")
+    _auth_string = _string_literal
```
tunetheweb, line 345: I'm not convinced it's worth the extra layer of abstraction to have these for one, or two references.

Any abstraction risks making the code more difficult to understand and I think doe these it would be clearer to just remove these and inline the values of them into the grammar.

For some of the other ones below, a level of abstraction and deduplication perhaps makes more sense (though even then I'm not convinced!) but think we should avoid overusing this IMHO as it does make reading the code more confusing in some senses as have to hope back and forth all over the place.

Inline comments are better for these IMHO if we want to make it clear that an `ObjectReferenceSegment` is an `_auth_plugin`?

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -314,6 +330,142 @@ class CreateTableStatementSegment(ansi.CreateTableStatementSegment):
...
+        OneOf(
+            Sequence("BY", OneOf(_random_password, _auth_string)),
+            Sequence(
+                "WITH",
+                _auth_plugin,
+                "AS",
+                _auth_string,
+            ),
+        ),
+    )
```
tunetheweb, line 360: This is only used in one place. So is it worth pulling out to it's own grammar?

Actually it's the same with ALL of the below.

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_mysql.py
greg-finley: Did this: https://github.com/sqlfluff/sqlfluff/pull/3289/commits/dbbd96802905cd401d2bf011dcbf0dcfcbc8d761

It's possible it could be iterated on even more.

src/sqlfluff/dialects/dialect_mysql.py
greg-finley: Yeah, you're right. I did it that way to organize as I was writing it, but it does read cleaner this way. https://github.com/sqlfluff/sqlfluff/pull/3289/commits/9e475bdb1e1683e9cf3bf150e103a1323df7a381

src/sqlfluff/dialects/dialect_mysql.py
greg-finley: ibid https://github.com/sqlfluff/sqlfluff/pull/3289/commits/9e475bdb1e1683e9cf3bf150e103a1323df7a381

tunetheweb (approved):
Thanks for doing those changes - much cleaner!

PR #3300, opened 2022-05-10
L064: Apply preferred quote-style for partially templated quoted literals

### LINKED ISSUE #3294: L064: Set preferred quotes for templated variables
### Search before asking

### Description

Currently, L064 ignores all STRING literals that involve quotes. For example a common Airflow templated STRING

```sql
SELECT '{{ ds }}'
```

is excluded from the the rule. (Assuming preferred quoting style is set to double quotes in this example). 

Now, as a user I could disable `ignore_templated_areas`. 
However, this is undesirable as it will pull in a bunch of false positives where the quotes are _inside_ the template. For example:

```sql
SELECT
    "a_string" AS col1,
    "{{ 'a_templated_string' }}" AS col2,
    "strings" IN {{ ('a', 'set', 'of', 'strings') }} AS col3
```

where the entire STRING is coming from a template.

```
L:   4 | P:  18 | L064 | Inconsistent use of preferred quote style 'double
                       | quotes'. Use "a" instead of 'a'.
L:   4 | P:  18 | L064 | Inconsistent use of preferred quote style 'double
                       | quotes'. Use "set" instead of 'set'.
L:   4 | P:  18 | L064 | Inconsistent use of preferred quote style 'double
                       | quotes'. Use "of" instead of 'of'.
L:   4 | P:  18 | L064 | Inconsistent use of preferred quote style 'double
                       | quotes'. Use "strings" instead of 'strings'.
```

It would be great, if we could expand L064 to cover cases where the quotes are outside the template. 

I would be willing to work on this if someone could point me into a loose direction on how this would be achievable.

### Use case

Increase consistency in SQL files w.r.t. to quoting styles.

### Dialect

Dialects with double quotes

### Code of Conduct

dmohns:

Fixes #3294

More lint errors will raise for Users having L064 enabled.

### Pull Request checklist

  - Other.


src/sqlfluff/rules/L064.py
```diff
@@ -102,6 +105,20 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+        templated_raw_slices = context.functional.segment.raw_slices.select(
+            rsp.is_slice_type("templated")
+        )
+        for raw_slice in templated_raw_slices:
+            pos_marker = context.segment.pos_marker
+            # This is to make mypy happy. Probably there is a better way of doing it?
+            assert isinstance(pos_marker, PositionMarker)
+            # quotes are part of a template, nothing we can do
+            if pos_marker.source_str() == raw_slice.raw:
+                return None
```
dmohns, line 120: Should this return `None` or `LintResult(memory=context.memory)`. I don't I understand the difference 😄

src/sqlfluff/rules/L064.py
```diff
@@ -102,6 +105,20 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+        # This rule can also cover quoted literals that are partially templated.
+        # I.e. when the quotes characters are _not_ part of the template we can
+        # meanigfully apply this rule.
+        templated_raw_slices = context.functional.segment.raw_slices.select(
+            rsp.is_slice_type("templated")
+        )
+        for raw_slice in templated_raw_slices:
+            pos_marker = context.segment.pos_marker
+            # This is to make mypy happy. Probably there is a better way of doing it?
+            assert isinstance(pos_marker, PositionMarker)
```
barrywhart, line 117: What you have is okay with me. Pretty sure I've done similar things. mypy is usually nice, but sometimes the workarounds are ugly.

src/sqlfluff/rules/L064.py
barrywhart: Probably `LintResult(memory=context.memory)`, since it's more explicit. IIRC, they behave the same, but I'd have to look at the source to confirm.

test/fixtures/rules/std_rule_cases/L064.yml
```diff
@@ -259,3 +259,66 @@ test_pass_date_constructor_strings_are_ignored_2:
...
+        '{{ "another_templated_string" }}'
+        {% endif %}
+  configs:
+    core:
+      dialect: bigquery
+    rules:
+      L064:
+        preferred_quoted_literal_style: double_quotes
+
+test_pass_fully_templated_quoted_literals_are_ignored:
```
barrywhart, line 316: I suggest some additional test cases. These are not the precise test cases, just to give an idea of the areas to test.

Multiple templated slices in one string:
```
SELECT '{{ "project" }}.{{ "dataset" }}.{{"table"}}'
```

Mix of templated and non-templated quotes, template slice spans segments:
```
SELECT '{{ "project' FROM table1" }}
```

Similarly:
```
{{ "SELECT 'proje" -}}ct' FROM table1
```

barrywhart:
I mentioned earlier that once we have it detecting issues, we'll have to figure out how to fix issues with code inside the curly braces. I still don't know how that will work, but to figure it out, I think we'll spend some time studying [this code](https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/core/parser/segments/base.py#L1293), which runs after fixes have been applied to the tree. This code is involved in converting the fixed parse tree back into source code that will be written to the `.sql` file.

test/fixtures/rules/std_rule_cases/L064.yml
dmohns: Yeah, good catch. If onside of the quotes are inside the template. This wont work. Adding some logic for this.

barrywhart:
With the example you gave, I suggest this rule should not touch the code inside `{{ }}`, so the second fix would be preferred (and is totally valid syntax).

And yes, there are rules which warn about things, and support auto-fix, but they don't auto-fix **everything**. This is typically because:
* It will take more effort to auto-fix. (I.e. we could do it, but we haven't yet.)
* Attempting to auto-fix is risky, either because of possible bugs in the rule, or ambiguity in how the rule interprets the SQL, so we feel safer flagging it and letting the user decide.

I think the work you've done to detect these issues is a nice improvement, whether or not it also autofixes them. Let us know what you plan to do, so we can review accordingly.

dmohns:
> I think the work you've done to detect these issues is a nice improvement, whether or not it also autofixes them. Let us know what you plan to do, so we can review accordingly.

Got it. I had a brief look at the the code and I think I will only do the linting part for now. We can leave the corresponding issue open, if someone wants to have a go at it.

tunetheweb:
Where is this at? @barrywhart you happy to merge this as is?

src/sqlfluff/rules/L064.py
```diff
@@ -102,6 +105,31 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
         ):
             return LintResult(memory=context.memory)
 
+        # This rule can also cover quoted literals that are partially templated.
+        # I.e. when the quotes characters are _not_ part of the template we can
+        # meanigfully apply this rule.
```
tunetheweb, line 110:
```suggestion
        # meaningfully apply this rule.
```

tunetheweb (approved):
I think this looks good. Just one minor typo to fix.

Issue #3322, opened 2022-05-11
SOQL: Recognize nested objects


greg-finley:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

In SOQL, many objects, like `Account` and `Contact`, have a relationship under the hood. Instead of writing a `JOIN`, you reference columns from the other object like so:

```soql
SELECT Account.Name FROM Contact
```

### Use case

It should be parseable

### Dialect

soql

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

barrywhart:
The example SQL you provided parses successfully already in both `soql` and `ansi` dialects. Can you clarify what you mean by "It should be parseable"?

greg-finley:
I get this, with no sqlfluff config.

```
gregoryfinley@Gregorys-MacBook-Pro git-repos % echo "SELECT Account.Name FROM Contact" > test.sql
gregoryfinley@Gregorys-MacBook-Pro git-repos % sqlfluff lint test.sql --dialect ansi             
== [test.sql] FAIL                                                                                                                                       
L:   1 | P:   8 | L026 | Reference 'Account.Name' refers to table/view not found
                       | in the FROM clause or found in ancestor statement.
All Finished 📜 🎉!
gregoryfinley@Gregorys-MacBook-Pro git-repos % 
```

It should not raise this warning, because it's ok that `Account` is not in this query.

greg-finley:
Though it will generate a parse tree. I'm confused ...
```
# YML test files are auto-generated from SQL files and should not be edited by
# hand. To help enforce this, the "hash" field in the file must match a hash
# computed by SQLFluff when running the tests. Please run
# `python test/generate_parse_fixture_yml.py`  to generate them after adding or
# altering SQL files.
_hash: a4667f2c8fa6b42fe78bcabf1b39505d18fc6a7a89d261aba8476d7441f7cd81
file:
  statement:
    select_statement:
      select_clause:
        keyword: SELECT
        select_clause_element:
          column_reference:
          - identifier: Account
          - dot: .
          - identifier: Name
      from_clause:
        keyword: FROM
        from_expression:
          from_expression_element:
            table_expression:
              table_reference:
[...]

PR #3338, opened 2022-05-13
SOQL: Disable L026 rule

greg-finley:

Closes https://github.com/sqlfluff/sqlfluff/issues/3322

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L026.py
+++ b/src/sqlfluff/rules/L026.py
@@ -60,13 +60,14 @@ class Rule_L026(BaseRule):
 
     groups = ("all", "core")
     config_keywords = ["force_enable"]
+    _dialects_disabled_by_default = ["bigquery", "hive", "redshift", "soql", "sparksql"]
 
     def _eval(self, context: RuleContext) -> EvalResultType:
         # Config type hints
         self.force_enable: bool
 
         if (
-            context.dialect.name in ["bigquery", "hive", "redshift", "soql", "sparksql"]
+            context.dialect.name in self._dialects_disabled_by_default
             and not self.force_enable
         ):
             return LintResult()
--- a/src/sqlfluff/rules/L026.py
+++ b/src/sqlfluff/rules/L026.py
@@ -66,7 +66,7 @@ class Rule_L026(BaseRule):
         self.force_enable: bool
 
         if (
-            context.dialect.name in ["bigquery", "hive", "redshift", "sparksql"]
+            context.dialect.name in ["bigquery", "hive", "redshift", "soql", "sparksql"]
             and not self.force_enable
         ):
             return LintResult()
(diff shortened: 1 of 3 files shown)


test/dialects/soql_test.py
```diff
@@ -2,22 +2,41 @@
...
     """Test that non-SELECT commands are not parseable."""
     cfg = FluffConfig(configs={"core": {"dialect": "soql"}})
     lnt = Linter(config=cfg)
     result = lnt.lint_string(raw)
-    assert result.num_violations() > 0
+    assert len(result.violations) == 1
+    assert isinstance(result.violations[0], SQLParseError)
+
+
+def test_ignore_unreferenced_object() -> None:
```
tunetheweb, line 34: This test should just be added to the L026.yml file rather than as a Python test.

The rule YAML files can’t check as much as pure Python (e.g. they can’t check for unparsable errors) but for simple parsing, rule passes or rule fixes we try to use them as simpler.

tunetheweb (changes requested):

test/dialects/soql_test.py
greg-finley: Yeah, I just discovered this file from looking at other PRs. Will do. 👍

src/sqlfluff/rules/L026.py
```diff
@@ -66,7 +66,7 @@ def _eval(self, context: RuleContext) -> EvalResultType:
         self.force_enable: bool
 
         if (
-            context.dialect.name in ["bigquery", "hive", "redshift", "sparksql"]
+            context.dialect.name in ["bigquery", "hive", "redshift", "soql", "sparksql"]
```
tunetheweb, line 69: So you think we should move these to a class variable (or whatever it’s called), similar to how we’ve done for dialect exceptions in other rules? Because it’s almost like a bit of config like those are.

tunetheweb:
Last thing and then I swear I’ll approve. The list of excluded dialects is also in the note in the doc string (would be nice if we could dynamically generate this instead of hardcoding - but a problem for another day).

greg-finley:
> Last thing and then I swear I’ll approve. The list of excluded dialects is also in the note in the doc string (would be nice if we could dynamically generate this instead of hardcoding - but a problem for another day).

Whoops, didn't even see it. Thanks!

greg-finley:
> Last thing and then I swear I’ll approve. The list of excluded dialects is also in the note in the doc string (would be nice if we could dynamically generate this instead of hardcoding - but a problem for another day).

We could even put the different reason why it's disabled for SOQL, but maybe that's too much detail.

tunetheweb:
> We could even put the different reason why it's disabled for SOQL, but maybe that's too much detail.

Up to you if you wanna tweak it slightly. Or say “due to support of things like structs” rather than list every reason for every dialect?

tunetheweb (approved):

PR #3371, opened 2022-05-19
Proof of concept: Release via Python util script

greg-finley:

As discussed with @tunetheweb , maybe we can do the release tasks via a Python script instead of shell commands. It should be more interoperable than shell commands, allowing the functions to be developed/run locally much more easily. After this change, [bot PRs still look ok](https://github.com/greg-finley/sqlfluff/pull/31/files).

Next step would be make an (idempotent!) `def start_changelog_entry` to replace the shell version, but I wanted to ensure this was headed the right direction first.

No.

### Pull Request checklist

  - Other.

### DIFF
--- a/.github/workflows/create-release-pull-request.yaml
+++ b/.github/workflows/create-release-pull-request.yaml
@@ -12,12 +12,18 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v3
+        with:
+          python-version: '3.10'
+
+      - name: Install dependencies
+        run: |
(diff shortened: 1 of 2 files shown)


tunetheweb:
It seems a lot of code to replace the one-line `sed` command.

While it may make sense for the more complicated CHANGELOG change, not sure it's necessary for the version number upgrades.

greg-finley-bond:
> It seems a lot of code to replace the one-line `sed` command.
> 
> While it may make sense for the more complicated CHANGELOG change, not sure it's necessary for the version number upgrades.

Fair. Let me come back with the CHANGELOG part in Python and see if it makes sense to combine all of it or not.

Issue #2991, opened 2022-04-04
BigQuery: Usage of `DECLARE`ed variable triggers L027


dmohns:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

When declaring a BigQuery variable using `DECLARE` and then later referencing it in a `SELECT` statement will trigger a L027 linting error, when more than one table is involved in the `FROM` clause.

### Expected Behaviour

As a user I would expect L027 to understand that I'm trying to use a variable declared earlier rather than a column alias. Similar to how L027 not fires in the USING case.

### Observed Behaviour

```
Unqualified reference 'test' found in select with more than one referenced table/view.
```

### How to reproduce

```sql
DECLARE test INT64 DEFAULT 42;

SELECT test
FROM t_table1
LEFT JOIN t_table_2
    ON TRUE
```

yields 

```
Unqualified reference 'test' found in select with more than one referenced table/view.
```

### Dialect

BigQuery

### Version

```
==== sqlfluff ====
sqlfluff:               0.11.2 python:                 3.10.3
implementation:        cpython dialect:                  bigquery
verbosity:                   2 templater:               jinja
```

### Configuration

N/A

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

tunetheweb:
Is it possible `test` is both a variable and a column reference in above example? If so, then the error could be valid, and SQLFluff has no way of knowing this.

Though guess we could make the assumption that it's the variable in these cases.

barrywhart:
Seems like the potential impact of this goes far beyond any single rule. It basically suggests that for BigQuery (and potentially other dialects that support similar behavior), we'd need to add a step that processes any declarations like this and makes them available later in the file. Does BigQuery allow conditionals? If so, then it seems possible a variable declaration may or may not be executed, and potentially it's impossible to know until the SQL is actually executed.

TL;DR: Potentially a significant amount of effort. Changing from "bug" to "enhancement".

barrywhart:
Google has removed the documentation page for this feature. What does that mean?

[...]

PR #3398, opened 2022-05-25
L027: Add support for `ignore_words`

dmohns:

Makes progress on #2991

Adding `ignore_words` is meant as a workaround for the linked issue. For users that follow a convention when declaring (SQL) variables (like a common prefix), they can be excluded from `L027` and reduce the need for many `-- noqa: L027` in the SQL.

No

### Pull Request checklist

  - Other.


tunetheweb (approved):
LGTM!

src/sqlfluff/rules/L027.py
```diff
@@ -45,13 +47,33 @@ def _lint_references_and_aliases(
...
         if len(table_aliases) <= 1:
             return None
+
+        # Get the ignore_words_list configuration.
+        try:
+            ignore_words_list = self.ignore_words_list
+        except AttributeError:
+            # First-time only, read the settings from configuration. This is
+            # very slow.
+            ignore_words_list = self._init_ignore_words_list()
```
tunetheweb, line 63: @barrywhart question for you.

Should this use `memory` in any returns, like L010 to save the cached config?

Never really understood that, and I notice that L059 doesn't use this (but maybe should?)

I notice

barrywhart: I prefer this technique for cached configuration, and `memory` for cached rule state (i.e. what happened earlier in the SQL file).

This caching technique is appropriate when executing complex code to get the rule configuration, rather than a value provided directly by the config as is typical. Here's the L010 code. Note how it involves loops, etc. Executing this code every time `_eval()` is called is **very** expensive.

```
        cap_policy_name = next(
            k for k in self.config_keywords if k.endswith("capitalisation_policy")
        )
        self.cap_policy = getattr(self, cap_policy_name)
        self.cap_policy_opts = [
            opt
            for opt in get_config_info()[cap_policy_name]["validation"]
            if opt != "consistent"
        ]
        # Use str() as L040 uses bools which might otherwise be read as bool
        ignore_words_config = str(getattr(self, "ignore_words"))
        if ignore_words_config and ignore_words_config != "None":
            self.ignore_words_list = self.split_comma_separated_string(
                ignore_words_config.lower()
            )
```

barrywhart (approved):

PR #3406, opened 2022-05-28
Make all Postgres identifiers quotable

benji-york:
Postgres allows any identifier to be quoted.

This PR modifies the Postgres dialect to allow any identifier (called `parameter` in the code) to be optionally double-quoted.

None that I can see.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -240,7 +240,7 @@
         )
     ),
     ParameterNameSegment=RegexParser(
-        r"[A-Z_][A-Z0-9_$]*", CodeSegment, name="parameter", type="parameter"
+        r'[A-Z_][A-Z0-9_$]*|"[^"]*"', CodeSegment, name="parameter", type="parameter"
```
tunetheweb, line 243: I think we should differentiate these, like we do with other identifiers. This allows rules to handle them differently (e.g. remove unnecessary quotes).

What do you think about creating a new `QuotedParameterNameSegment` and then also a `ParameterNameGrammar=OneOf(ParameterNameSegment, QuotedParameterNameSegment)` and then use `ParameterNameGrammar` everywhere we currently use `ParameterNameSegment`?

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_postgres.py
benji-york: Thanks for looking at this!  That approach sounds good to me.

It may be a couple of days before I have a chance to work on it.

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -207,6 +207,12 @@
     MultilineConcatenateDelimiterGrammar=AnyNumberOf(
         Ref("MultilineConcatenateNewline"), min_times=1, allow_gaps=False
     ),
+    QuotedParameterNameSegment=RegexParser(
+        r'"[^"]+"', CodeSegment, name="quoted_parameter", type="parameter"
+    ),
+    ParameterNameGrammar=OneOf(
+        Ref("ParameterNameSegment"), Ref("QuotedParameterNameSegment")
+    ),
```
benji-york, line 215: This is the core of the change.  The other changes in this file are mechanical replacements of `ParameterNameSegment` with `ParameterNameGrammar`.

tunetheweb: Hmmm.... I'm starting to have second thoughts about my suggestion here.

I thought when I was first reviewing this, that `ParameterNameSegment` was a postgres dialect specific thing, but it's used in ansi dialect - and lots of others. And Postgres still inherits some of the syntax from ansi version.

I also noticed BigQuery already has this set as a `OneOf` of normal and quoted.

So think we should either reverse your last change. Or alternatively change it in all the other dialects - but that's a much bigger change.

Sorry about that!

benji-york: No worries.  I'll revert the changes and see where we land.

tunetheweb: Get a chance to look at this @benji-york ? Can we just revert last few changes or is it not as simple as that?

And apologies again for sending you on this wild goose chase!

benji-york: My git-fu has failed me and I've made a royal mess of this branch.  I won't have time to work on this for at least the next few days.  If you want some movement on this before I'm able to get to it, feel free to lift whatever you want from this work and merge it.

One consideration: just about everything is quotable in Postgres, so I'm sure there is more that should be done along this route.  On the other hand, we shouldn't let the perfect be the enemy of the good.

tunetheweb:
Recreated this in #3442 

Best never to force push, as that makes reverting things a lot more difficult.

PR #3283, opened 2022-05-07
L028: In tsql dialect, table variables cannot be used to qualify references

barrywhart:
### Brief summary of the change made
Fixes #3243 

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #3407, opened 2022-05-28
Fix bug in `apply_fixes()` with leading/trailing whitespace

### LINKED ISSUE #3399: sqlfluff fix incorrectly move `FROM` to the end of its previous line
### Search before asking

### What Happened

For the below sql 
image

If I ran `sqlfluff fix`

image

### Expected Behaviour

The keyword `FROM` should not be moved

### Observed Behaviour

Then it became 

image

Please note that the keyword `FROM` is moved to an incorrect place

### How to reproduce

```sql
SELECT
    a.*,
    b as col
FROM
    {{source("ids","shop")}} a;
```

just run sqlfluff fix with the above SQL

### Dialect

tsql

### Version

sqlfluff==0.13.2
sqlfluff-templater-dbt==0.9.4
dbt-core==1.0.0
dbt-extractor==0.4.0

### Configuration

```
[sqlfluff]
verbose = 0
nocolor = False
dialect = tsql
templater = jinja
# rules = L001,L002,L003,L004,L005,L006,L007,L008,L009,L010,L011,L012,L014,L029,L030,L039,L040,L050,L059,L062,L063,L065
exclude_rules = L031,L044
recurse = 0
output_line_length = 80
runaway_limit = 10
ignore_templated_areas = True
encoding = autodetect
disable_noqa = False

[sqlfluff:indentation]
indented_joins = True
indented_using_on = True
template_blocks_indent = True

[sqlfluff:templater]
unwrap_wrapped_queries = False

[sqlfluff:templater:jinja]
apply_dbt_builtins = True
load_macros_from_path=./macros

dbt_source = {% macro source(source_name, table) %}{{source_name}}.{{table}}{% endmacro %}

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 80
indent_unit = space
comma_style = trailing
allow_scalar = True

operator_new_lines = before

capitalisation_policy = upper

[sqlfluff:rules:L011]  # Aliasing
aliasing = explicit

[sqlfluff:rules:L012]  # Aliasing
aliasing = explicit

[sqlfluff:rules:L016]
ignore_comment_lines = False

[sqlfluff:rules:L026]
force_enable = False

[sqlfluff:rules:L028]
force_enable = False

capitalisation_policy = upper

[sqlfluff:rules:L038]

capitalisation_policy = upper

[sqlfluff:rules:L042]
forbid_subquery_in = join

prefer_count_1 = False
prefer_count_0 = False

[sqlfluff:rules:L052]  # Semi-colon formatting approach.
multiline_newline = False

[sqlfluff:rules:L054]  # GROUP BY/ORDER BY column references.

[sqlfluff:rules:L057]  # Special characters in identifiers
```

### Code of Conduct

barrywhart:
Fixes #3399. This bug was introduced in PR #3283.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -1199,7 +1199,12 @@ class BaseSegment:
 
             before = []
             after = []
-            if not r.can_start_end_non_code and r.parse_grammar:
+            # If there's a parse grammar and this segment is not allowed to
+            # start or end with non-code, check for (and fix) misplaced
+            # segments. The reason for the parse grammar check is autofix if and
+            # only if parse() would've complained, and it has the same parse
+            # grammar check prior to checking can_start_end_non_code.
+            if r.parse_grammar and not r.can_start_end_non_code:
                 idx_non_code = self._find_start_or_end_non_code(seg_buffer)
                 # Are there misplaced segments from a fix?
                 if idx_non_code is not None:
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -1199,7 +1199,7 @@ class BaseSegment:
 
             before = []
             after = []
-            if not r.can_start_end_non_code:
+            if not r.can_start_end_non_code and r.parse_grammar:
                 idx_non_code = self._find_start_or_end_non_code(seg_buffer)
                 # Are there misplaced segments from a fix?
                 if idx_non_code is not None:
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -1009,6 +1009,7 @@ ansi_dialect.add(
             ),
         ),
         Sequence(OneOf("IGNORE", "RESPECT"), "NULLS"),
+        Ref("IndexColumnDefinitionSegment"),
     ),
     PostFunctionGrammar=OneOf(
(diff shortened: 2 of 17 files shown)


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1199,7 +1199,7 @@ def apply_fixes(
 
             before = []
             after = []
-            if not r.can_start_end_non_code:
+            if not r.can_start_end_non_code and r.parse_grammar:
```
tunetheweb, line 1202: Can you explain why this is necessary and how the lack of it caused the bug? Bit confused.

barrywhart: Agree. The intent of this code is to avoid parse errors. If you look at the other place where we check this flag (in the parse() function about 200 lines above), you'll see that it only does this check if there's a parse grammar. So I can't answer the deeper "why", but the simple answer is to match that behavior, only changing things if not doing so would raise an error later there.

barrywhart: I'll see about adding a comment and/or changing the two areas of code to be more clearly "aligned".

barrywhart: Comment added, code tweaked _slightly_ to align a bit more with `parse()`.

tunetheweb (approved):
Still don’t fully get it, but accept your comment about it being similar to the other instance of this. And it seems to work!

PR #3409, opened 2022-05-30
Snowflake: IDENTIFIER pseudo-function

### LINKED ISSUE #3404: Snowflake `identifier` - sqlfluff fix unable to parse
### Search before asking

### What Happened

When running `sqlfluff fix` on files that have a line with `identifier()` in it, the file cannot be parsed.  We have files that that use the `identifier` function in snowflake and we do a workaround of commenting out that line to then run `sqlfluff fix`.

This is an example of a line making use of the snowflake `identifier` function.

```sql
USE SCHEMA identifier('{{ params.schema_name }}');
```

### Expected Behaviour

When running `sqlfluff fix`, the command is able to change the file rather than returning a "parsing error".

### Observed Behaviour

when running `sqlfluff fix` on a file with the `identifier` function, the result is "parsing error"

```sql
USE SCHEMA identifier('{{ params.schema_name }}');
```

I can confirm that without the jinja it returns a parsing error

```sql
USE SCHEMA identifier('tt');
```

### How to reproduce

A file with a line like this, with `identifier`
```sql
USE SCHEMA identifier('{{ params.schema_name }}');
```
run sqlfluff fix on that file

### Dialect

snowflake

### Version

0.13.0

### Configuration

# Usage : README_SQLFULFF.md

[sqlfluff:core]
templater = jinja
dialect = snowflake
ignore = parsing
# L031 - weird table aliasing rule
# L032 - allow 'using', rather than "sqlfluff fix" changing to 'on' clause
# L034 - the order of elements in the select clause
# The reason is that it will change the order and effect the insert statement
# https://docs.sqlfluff.com/en/stable/rules.html#sqlfluff.core.rules.Rule_L034
exclude_rules = L009, L031, L032, L034

[sqlfluff:rules]
comma_style = trailing

[sqlfluff:rules:L016]
max_line_length = 300

[sqlfluff:rules:L010]
# Keywords
capitalisation_policy = upper

[sqlfluff:rules:L014]
# Keywords

[sqlfluff:rules:L030]
# functions

[sqlfluff:rules:L040]
# booleans and null
capitalisation_policy = upper

[sqlfluff:rules:L063]
# datatypes

dag_run.conf = abc
params = {'schema_name': 'test', 'days_back': 1}
dag_run = {'conf': {'start_date': 'test', 'end_date': 'w', 'window_start_date': '2021-01-01'}}
ds = 'tt'

### Code of Conduct

mdahlman:

fixes #3404 

Nope.

### Pull Request checklist

  - added snowflake_identifier_pseudo_function.sql
  - existing snowflake_grant_revoke.yml parses differently now. It's better.

### DIFF
--- a/src/sqlfluff/dialects/dialect_snowflake.py
+++ b/src/sqlfluff/dialects/dialect_snowflake.py
@@ -17,6 +17,7 @@ from sqlfluff.core.parser import (
     Dedent,
     Delimited,
     Indent,
+    Matchable,
     NamedParser,
     Nothing,
     OneOf,
@@ -31,6 +32,7 @@ from sqlfluff.core.parser import (
     StringParser,
     SymbolSegment,
 )
+from sqlfluff.core.parser.segments.base import BracketedSegment
 from sqlfluff.dialects.dialect_snowflake_keywords import (
     snowflake_reserved_keywords,
     snowflake_unreserved_keywords,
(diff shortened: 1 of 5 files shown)


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -682,6 +684,171 @@
...
+            Sequence(
+                "IDENTIFIER",
+                Bracketed(
+                    OneOf(
+                        Ref("SingleQuotedIdentifierSegment"),
+                        Ref("ReferencedVariableNameSegment"),
+                    ),
+                ),
+            ),
+        ),
```
tunetheweb, line 707: I don't think you need the `Delimited` here. That's only used in the next section to allow "dot" delimiters (e.g. `schema.table.column`). As you are not overriding the delimiter here from the default comma, it's not needed.

```suggestion
        Sequence(
            "IDENTIFIER",
            Bracketed(
                OneOf(
                    Ref("SingleQuotedIdentifierSegment"),
                    Ref("ReferencedVariableNameSegment"),
                ),
            ),
        ),
```

Though I wonder if `schema.table.IDENTIFIER('column')` is valid? I presume not?

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -682,6 +684,171 @@
...
+
+class SequenceReferenceSegment(ObjectReferenceSegment):
+    """A reference to a sequence.
+
+    Overriding to capture Snowflakes's override of ObjectReferenceSegment
+    """
+
+    type = "sequence_reference"
+
+
```
tunetheweb, line 796: Do you really need these? They inherit from `ObjectReferenceSegment` so would have though they would automatically have picked up the changes to that above?

```suggestion
```

Or did you try that, and it didn't work?

tunetheweb (changes requested):
Left some suggestions below for your implementation.

I'm also wondering if, instead of this, we could just override `SingleIdentifierGrammar` since they all ultimately use that? It would avoid a lot of these changes.

src/sqlfluff/dialects/dialect_snowflake.py
mdahlman: Holy cow. `SingleIdentifierGrammar` seems much cleaner. Good idea.

src/sqlfluff/dialects/dialect_snowflake.py
mdahlman: From the docs, it doesn't seem likely. But they aren't explicit. It might be a valid edge case, but I'm guessing that it's not.

As it happens... this variation currently parses fine:
`create or replace table my_db.my_schema.identifier('my_table') (c1 number);`
I have not added a test case, since I don't know if it's truly valid. But it's more likely than your column example.

src/sqlfluff/dialects/dialect_snowflake.py
mdahlman: As you'll see in the updated version, I removed nearly all of those extra classes once the `IDENTIFIER` syntax moved to `SingleIdentifierGrammar`. But `FunctionNameSegment` seems to be distinct. I couldn't find a way around keeping that one.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -682,6 +692,43 @@
...
+                ),
+            ),
+            # Base function name
+            OneOf(
+                Ref("FunctionNameIdentifierSegment"),
+                Ref("QuotedIdentifierSegment"),
+            ),
+            allow_gaps=False,
+        ),
+    )
```
tunetheweb, line 729: I would replace this, with this:

```suggestion
    match_grammar: Matchable = Sequence(
        # Project name, schema identifier, etc.
        AnyNumberOf(
            Sequence(
                Ref("SingleIdentifierGrammar"),
                Ref("DotSegment"),
            ),
        ),
        # Base function name
        OneOf(
            Ref("FunctionNameIdentifierSegment"),
            Ref("QuotedIdentifierSegment"),
            # Snowflake's IDENTIFIER pseudo-function
            # https://docs.snowflake.com/en/sql-reference/identifier-literal.html
            Sequence(
                "IDENTIFIER",
                Bracketed(
                    OneOf(
                        Ref("SingleQuotedIdentifierSegment"),
                        Ref("ReferencedVariableNameSegment"),
                    ),
                ),
            ),
        ),
        allow_gaps=False,
    )
```

src/sqlfluff/dialects/dialect_snowflake_keywords.py
```diff
@@ -524,4 +527,9 @@
 WRITE
 XML
 ZONE
-ZSTD"""
+ZSTD
+"""
+
+# These are not official Snowflake keywords, but SQLFluff needs them to parse well.
+snowflake_unreserved_keywords += """IDENTIFIER
+"""
```
tunetheweb, line 535: We often add unreserved words (e.g. `ABORT`) so no need to pull them into a separate section. Too much to track that way. So can you move this above?

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_snowflake_keywords.py
mdahlman: No problem.

tunetheweb (approved):

PR #3431, opened 2022-06-07
Add better handling for SQLFluffUserError when running core cli commands

alanmcruickshank:
When testing the project against Airflow - I ran into a few errors which weren't presented in a very user friendly way - this addresses that. The only major change here is exactly how we wrap errors for the `parse` cli command - I'm effectively covering less with the potential `OSError` catch, but that should only have been raised by the bits I'm still covering so I think this should still be appropriate.


barrywhart (approved):
Looks good to me. I see that Black is complaining about line length.

If you want to take the time, I think the code would read a bit cleaner if the error handling were implemented as a context manager rather than a wrapper function. This article has an example: https://vaclavkosar.com/software/Python-Context-Manager-With-Statement-Exception-Handling

alanmcruickshank:
> I think the code would read a bit cleaner if the error handling were implemented as a context manager rather than a wrapper function

Thanks @barrywhart - that's what I wanted to do initially but I wasn't aware of the `__exit__` method mentioned in that blog post which was the key. Thanks for the tipoff!

PR #3433, opened 2022-06-07
Handle templater validation errors more gracefully

alanmcruickshank:
I'm getting a template error (will post issue separately), but the error totally kills the parsing of a whole folder, when it only needs to kill a particular file. The user also gets a full error trace rather than a nicely formatted error. This solves the formatting and UX. Separate issue coming for the error.


barrywhart (approved):
Looks good to me. I see that the Python lint checks are failing now due to an unused import.

PR #3440, opened 2022-06-09
Add `databricks` dialect as an alias for `sparksql` dialect

R7L208:

This adds a new dialect, which is essentially an alias for Spark SQL. 

n/a

### Pull Request checklist

  - Other.


src/sqlfluff/core/dialects/__init__.py
```diff
@@ -36,6 +36,7 @@
     "teradata": ("dialect_teradata", "teradata_dialect"),
     "tsql": ("dialect_tsql", "tsql_dialect"),
     "sparksql": ("dialect_sparksql", "sparksql_dialect"),
+    "databricks": ("dialect_databricks", "databricks_dialect"),
```
tunetheweb, line 39: I wonder would it work to have this, so you don’t actually need the dialect files?

```suggestion
    "databricks": ("dialect_sparksql", "sparksql_dialect"),
```

Or do you see this not being a complete alias and having some differences in future?

R7L208: so this is how I initially implemented the alias but was getting failures with `/Users/lorin.dawson/repos/sqlfluff/test/cli/autocomplete_test.py`.

I think what you are suggesting is the better way to do this but couldn't figure out how to pass tests w/o changing the test case.

R7L208: Adding the dialect files was an easy way to get everything passing and start the review process :) but I agree that they're fundamentally not needed

test/cli/autocomplete_test.py
```diff
@@ -8,7 +8,7 @@
     "incomplete,expected",
     [
         ["an", ["ansi"]],
-        ["s", ["snowflake", "soql", "sparksql", "sqlite"]],
+        ["s", ["sparksql", "snowflake", "soql", "sparksql", "sqlite"]],
```
R7L208, line 11: Additional instance of sparksql is generated by the "databricks" dialect

tunetheweb: I think you can fix this by changing this:

https://github.com/sqlfluff/sqlfluff/blob/20f065c846245307bad853b91b557f5619ac492a/src/sqlfluff/cli/autocomplete.py#L21

to this:

```py
    dialect_names = [e.label for e in list_dialects()]
```

Which also allows you to add back your test to `test/cli/autocomplete_test.py`:

```py
        ["d", ["databricks", "db2"]],
```

tunetheweb (approved):

PR #3442, opened 2022-06-10
Make all Postgres identifiers quotable 

tunetheweb:
**RECREATING PR #3406 after that branch got messed up in git.**

Postgres allows any identifier (called "parameter" here) to be quoted.

This PR modifies the Postgres dialect to allow any identifier (called parameter in the code) to be optionally double-quoted.

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -240,7 +240,7 @@
         )
     ),
     ParameterNameSegment=RegexParser(
-        r"[A-Z_][A-Z0-9_$]*", CodeSegment, name="parameter", type="parameter"
+        r'[A-Z_][A-Z0-9_$]*|"[^"]*"', CodeSegment, name="parameter", type="parameter"
```
barrywhart, line 243: This regex doesn't appear to allow "escaping" quotes. Should it?

barrywhart (approved):
Looks good! One question about whether identifiers with escaped quotes should be supported.

src/sqlfluff/dialects/dialect_postgres.py
tunetheweb: I don’t think it should. The original implementation didn’t and this just allows quotes around it.

And really, what manic would use that in a parameter name! 😄

PR #3466, opened 2022-06-17
L026: Fix false positive with table function aliases

### LINKED ISSUE #3341: Postgres: L026 rule violation for tables with function expression inside
### Search before asking

### What Happened

Code
```
select
	sc.col1 as colx
	, pn.col1 as coly
from sch1.tbl1 as sc
cross join unnest(array[111, 222]) as pn(col1)
```
violates rule L026 in postgres dialect.

### Expected Behaviour

No L026 rule violations

### Observed Behaviour

== [D:\py_code\sqlfluff\real_tests\v_v.sql] FAIL
L:   3 | P:   4 | L026 | Reference 'pn.col1' refers to table/view not found in

### How to reproduce

run sqlfluff lint v_v.sql --rules L026 --config config\sqlfluff\.sqlfluff

### Dialect

Postgres

### Version

SQLFluff 0.13.1

### Configuration

[sqlfluff]
exclude_rules = L007,L015,L017,L020,L021,L022,L023,L024,L029,L031,L032,L033,L034,L035,L037,L038,L041,L042,L043,L044,L045,L046,L047
rules = L050
sql_file_exts = .sql
dialect = postgres

[sqlfluff:indentation]
indented_joins = False
template_blocks_indent = True

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 250
indent_unit = tab
comma_style = leading
allow_scalar = False

[sqlfluff:rules:L010] # Keywords
capitalisation_policy = lower

[sqlfluff:rules:L016]
ignore_comment_lines = False

[sqlfluff:rules:L030] # Function names
capitalisation_policy = lower
[sqlfluff:rules:L040] # Null & Boolean Literals
capitalisation_policy = lower

[sqlfluff:rules:L019]
comma_style = leading

[sqlfluff:rules:L028]

### Code of Conduct

barrywhart:
Fixes #3341 

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/rules/analysis/select.py
+++ b/src/sqlfluff/core/rules/analysis/select.py
@@ -16,6 +16,7 @@ class SelectStatementColumnsAndTables(NamedTuple):
     select_targets: List[BaseSegment]
     col_aliases: List[ColumnAliasInfo]
     using_cols: List[str]
+    from_expression: Optional[BaseSegment]
 
 
 def get_select_statement_info(
@@ -57,7 +58,9 @@ def get_select_statement_info(
     # from ON clauses.
     using_cols = []
     fc = segment.get_child("from_clause")
+    from_expression: Optional[BaseSegment] = None
     if fc:
+        from_expression = fc.get_child("from_expression")
         for join_clause in fc.recursive_crawl("join_clause"):
             seen_using = False
             for seg in join_clause.iter_segments():
@@ -93,6 +96,7 @@ def get_select_statement_info(
         select_targets=select_targets,
         col_aliases=col_aliases,
         using_cols=using_cols,
+        from_expression=from_expression,
     )
 
 
--- a/src/sqlfluff/core/rules/analysis/select.py
+++ b/src/sqlfluff/core/rules/analysis/select.py
@@ -12,7 +12,7 @@ class SelectStatementColumnsAndTables(NamedTuple):
     select_statement: BaseSegment
     table_aliases: List[AliasInfo]
     standalone_aliases: List[str]
-    value_table_aliases: List[AliasInfo]
+    value_table_function_aliases: List[AliasInfo]
     reference_buffer: List[BaseSegment]
     select_targets: List[BaseSegment]
     col_aliases: List[ColumnAliasInfo]
@@ -24,14 +24,16 @@ def get_select_statement_info(
 ) -> Optional[SelectStatementColumnsAndTables]:
     """Analyze a select statement: targets, aliases, etc. Return info."""
     assert segment.is_type("select_statement")
-    table_aliases, standalone_aliases, value_table_aliases = get_aliases_from_select(
-        segment, dialect
-    )
+    (
+        table_aliases,
+        standalone_aliases,
+        value_table_function_aliases,
+    ) = get_aliases_from_select(segment, dialect)
     if (
         early_exit
         and not table_aliases
         and not standalone_aliases
-        and not value_table_aliases
+        and not value_table_function_aliases
     ):
         return None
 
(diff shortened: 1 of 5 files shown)


tunetheweb:
@barrywhart did you see this code?:

https://github.com/sqlfluff/sqlfluff/blob/0f5d8db9dce96200f31e06f20f1e4a32d6f0c20a/src/sqlfluff/core/rules/analysis/select.py#L112-L121

I wonder if, instead of your fix we should just make that filtering optional and not filter it for L026?

barrywhart:
@tunetheweb: Are you suggesting adding a new parameter to `get_aliases_from_select()` to return all the aliases in `table_aliases`? That could definitely work. I was wondering yesterday which rules (if any) actually care about the kind of alias or not. This function is pretty old, and we may have simply been conservative when we added standalone aliases (i.e. kept them separate from fear of breaking things).

tunetheweb:
Yes. Seems more sensible to filtering them out, then adding them back in again with an alternative method. Which is what this PR is currently doing if I understand it right.

Alternative is to just include them always and see what breaks - to see if the parameter is even needed. I presume there was a reason to filter it out when that code was added? But maybe no longer needed?

barrywhart:
@tunetheweb: I refactored things based on your suggestion. When we first added support for value table function aliases, we lumped them in with standalone aliases, which worked for some rules but not others (as we see here with rule L026). I changed things now so value table aliases are stored separately, thus we now have three types:
* Regular aliases
* Value table aliases
* Standalone aliases

Now, each rule can choose for itself which ones it cares about and what to do with them. This affects a few other rules, but in simple ways, and it likely avoids future tricky issues that come from treating different alias types as if they were the same.

tunetheweb:
Seems cleaner. Even if more code.

What’s the difference between a table alias and a stand-alone alias? I’d the stand-alone alias the usage intbhe select clause?

barrywhart:
A table alias is your basic alias, e.g.:
```
SELECT * FROM foo as bah
```

From a quick review of the code and test cases, standalone aliases (our term, not an official SQL term) are from pivot table queries, e.g.:
```
    select
      t1._id
      , [1] as lvl_1
      , [2] as lvl_2
      , [3] as lvl_3
    from
      table1 t1
    pivot
      (max(value) for rn in([1], [2], [3]) ) as pvt
```

(The `pvt` alias above.) IIRC, pivot table syntax **requires** providing an alias, so it confounds some of our rules that think you shouldn't provide an alias if you don't use it.

Perhaps we should rename that field as `pivot_table_aliases`. 🤔 But I think this is good enough for now. What do you think?

tunetheweb:
We’ll just wondering should a standalone alias then? What rule didn’t like them in there do you remember?

barrywhart:
It was probably L025: Tables should not be aliased if that alias is not used.

barrywhart:
@tunetheweb: Any other questions or concerns with this one?

tunetheweb:
I think this is fine as it’s written, and possibly I’m being overly picky, but wanted to have another look at whether we really needed the extra alias category as still struggling to understand the difference between that and standalone alias to be honest. Will have a look later and either approve as is or suggest a change.

tunetheweb:
So I had another look and the value_table_functions ARE already being included in standalone aliases (as well as pivot table columns), as shown in the code above:

https://github.com/sqlfluff/sqlfluff/blob/0f5d8db9dce96200f31e06f20f1e4a32d6f0c20a/src/sqlfluff/core/rules/analysis/select.py#L112-L121

Therefore I don't think there's any need to include them in a third alias category. And we can just get L026 to look these up there. This will also include pivot table columns but don't think that's an issue to be honest.

I've opened a simpler PR at #3470 that does this and all tests seem to pass.

WDYT think? Can we close this, and merge that instead? Or have I missed something?

PR #3488, opened 2022-06-23
Fix `load_macros_from_path` to actually support multiple paths

emancu:
The definition of `macro_ctx` inside the `for` loop basically kills the feature of "comma-separated list of `.sql` or folders" (see [docs](https://docs.sqlfluff.com/en/stable/configuration.html))

If a folder has multiple entries, only the last one will prevail because the `macro_ctx = {}` is being executed on each iteration.

Also, the `return` value of the function is a variable defined in an inner scope!

Since the _extraction_ of macros is accumulative (updating the same dictionary) I suggest updating the docs to explicitly say the last path takes precedence.

⚠️ Not sure how to test this. I couldn't find any test related to the code that I'm changing.
Should I create a bunch of directories with files defining dummy macros and prove it works?

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/templaters/jinja.py
+++ b/src/sqlfluff/core/templaters/jinja.py
@@ -67,12 +67,12 @@ class JinjaTemplater(PythonTemplater):
     @classmethod
     def _extract_macros_from_path(cls, path: List[str], env: Environment, ctx: Dict):
         """Take a path and extract macros from it."""
+        macro_ctx = {}
         for path_entry in path:
             # Does it exist? It should as this check was done on config load.
             if not os.path.exists(path_entry):
                 raise ValueError(f"Path does not exist: {path_entry}")
 
-            macro_ctx = {}
             if os.path.isfile(path_entry):
                 # It's a file. Extract macros from it.
                 with open(path_entry) as opened_file:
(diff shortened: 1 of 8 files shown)


barrywhart (commented):
The fix looks good. Could you add a test case to help make sure this doesn't get broken again in the future?

barrywhart:
The test `test/core/templaters/jinja_test.py::test__templater_full` hits this code, but only passes a 1-item list. Perhaps create a new test using similar code. I wonder if you could create a test that directly calls `_extract_macros_from_path()`. That seems like a nicer approach since it avoids exercising a bunch of other, somewhat unrelated code.

barrywhart:
Another possibility: create a test similar to [this one](https://github.com/sqlfluff/sqlfluff/blob/main/test/cli/commands_test.py#L402L403) . This would require little if no Python test code. You'd just create a new directory similar to the one used by that test (`test/fixtures/cli/unknown_jinja_tag/`):
* `.sqlfluff` file
* "Main" SQL file
* Macros

emancu:
@barrywhart Thank you so much for pointing out how to test it. Took me a while but I got it 😅 

I decided to go with Python implementation because there is no option for the cli.
I hope it is ok.

I verified that the test fails if I run the original code and that it passes with the new changes.

Any ETA on the release including this? I could use it at carwow.

barrywhart:
The test looks great -- thanks!

Releases are usually every 2-4 weeks, depending on the number of PRs merged. I would guess we'll release within the next two weeks.

barrywhart (approved):

PR #3499, opened 2022-06-27
Add a new configuration for L018 to ignore indentations

### LINKED ISSUE #3498: Add configuration to ignore indentation in L018
### Search before asking

### Description

In case CTE is inside template block, the L018 and L003 contradict each other.

### Use case

```
with
{% if true %}
  cte as (
      select 1
  )
{% else %}
  cte as (
      select 2
  )
{% endif %}
select * from cte
```

I won't expect any lint errors here, but L018 suggests to de-indent closing brackets, and L003 suggests to indent them back.

### Dialect

No specific dialect

### Code of Conduct

zidder:

fixes #3498 

Add a new configuration variable for rule L018 to ignore indentations. This will be helpful in case the CTE is in a template block. In that case, currently L018 fails suggesting to move closing bracket, which contradicts to L003. The new solution basically suggests to move closing bracket to a new line, but to not look at the indent, L003 will take care of indentations.

No

### Pull Request checklist

  - Other.


tunetheweb:
I\m a bit confused with this change. What's the difference between ignoring indents and just turning off the rule?

If `ignore_indents` is set to True, won't this basically mean the rule never runs?

zidder:
> I\m a bit confused with this change. What's the difference between ignoring indents and just turning off the rule?
> 
> If `ignore_indents` is set to True, won't this basically mean the rule never runs?

Actually no, see the `test_pass_with_clause_closing_misaligned_indentation_ignoring_indent` test in yaml tests. It fails with both `ignore_indent` and without. The idea is to move the closing bracket to the new line, but not check indents and leave it to L003. If you check the L018.py, there are two cases where the rule fails, and one of them is turned off in case `ignore_indent=True`.

zidder:
I see the issue now. I was thinking about linting only, and lints will pass only for the nice SQL:
```
    with
    {% if true %}
      cte as (
          select 2
      )
    {% else %}
      cte as (
          select 2
      )
    {% endif %}
    select * from cte
```

But the fix doesn't work properly. I'll make this work if I can.
Thanks for reviewing.

tunetheweb:
So I guess you're going to try to indent that properly? In that case maybe the config name should be changed from `ignore_indents` to `allow_indents`? In which case the first case should check the correct number of whitespaces have been given as well, rather than just any whitespaces?

I'm still not convinced it's worth it, rather than just disabling the rule. But if you're happy to work on it, and think you can get something clean and reasonably easy to code and maintain then happy to review once you do.

zidder:
@tunetheweb I found the reason the fix didn't work properly, and I'm not happy with it :( It tried to apply fix to the initial file during the 3rd iteration of the loop.

zidder:
@barrywhart 
Do you mean changing logic of the rule? Are you sure this won't break somebody else's code (SQL) ?
I'll be satisfied with that solution if there won't be any negative effects.
I'm still not very familiar with the code, but I'll try to change it properly.

barrywhart:
That's what I mean. I haven't thought about this option carefully, but as @tunetheweb says, if we can find a solution we're happy with, that sounds good.

zidder:
@barrywhart , Changed L018 so that it won't take responsibility to fix indentations.

Tested with the following SQLs when L003 is enabled.

First SQL:
```
WITH
{% if true %}
    cte AS (
        SELECT 1)
{% else %}
    cte AS (
        SELECT 2)
{% endif %}
SELECT * FROM cte
```

First Fixed SQL:
```
WITH
{% if true %}
    cte AS (
        SELECT 1
    )
{% else %}
    cte AS (
        SELECT 2)
{% endif %}
SELECT * FROM cte
```

Second SQL:
```
WITH
{% if true %}
    cte AS (
        SELECT 2
)
{% endif %}

SELECT * FROM cte
```

Second Fixed SQL:
```
WITH
{% if true %}
    cte AS (
        SELECT 2
    )
{% endif %}

SELECT * FROM cte
```

Third SQL:
```
WITH cte AS (
    SELECT 1
    ) SELECT * FROM cte
```

Third Fixed SQL:
```
WITH cte AS (
    SELECT 1
) SELECT * FROM cte
```

Forth SQL:
```
    WITH cte AS (
        SELECT 1
)

SELECT * FROM cte
```

Forth Fixed SQL:
```
WITH cte AS (
    SELECT 1
)

SELECT * FROM cte
```

As you can see, those SQLs were fixed with L003 and L018 combined, and L018 doesn't contradict L003 anymore. Test cases don't show fixes by L003, but you can see, that L018 adds new lines only.

tunetheweb:
Note you can do combination tests (e.g. L018 and L003) with a python test.
Have a look here at this example: https://github.com/sqlfluff/sqlfluff/blob/main/test/rules/std_L003_L065_combo_test.py

barrywhart:
You can also do combination tests without writing Python. See `test/rules/std_fix_auto_test.py`. It picks up test cases from the `test/fixtures/linter/autofix` directory.

barrywhart:
Actually, one of those `autofix` tests is failing: `test/rules/std_fix_auto_test.py::test__std_fix_auto[ansi-007_with_clause]`

test/fixtures/linter/autofix/ansi/007_with_clause/after.sql
```diff
@@ -1,5 +1,5 @@
 -- Dealing with complicated indents before with clauses.
     	WITH cte as (
             select a from tbla
-        )
+)
         select a from cte
```
tunetheweb, line 5:
```suggestion
        select a from cte

```

test/fixtures/linter/autofix/ansi/027_l003_l018_templated_block_without_newline_cte/after.sql
```diff
@@ -0,0 +1,10 @@
+WITH
+{% if true %}
+    cte AS (
+        SELECT 1
+    )
+{% else %}
+    cte AS (
+        SELECT 2)
```
tunetheweb, line 8: Should this not be fixed too?

test/fixtures/linter/autofix/ansi/027_l003_l018_wrong_indent_cte/after.sql
```diff
@@ -0,0 +1,3 @@
+WITH cte AS (
+    SELECT 1
+) SELECT * FROM cte
```
tunetheweb, line 3: Should the rule be changed to be on a new line of it's own?

To fix this to:

```sql
WITH cte AS (
    SELECT 1
)

SELECT * FROM cte
```

test/fixtures/rules/std_rule_cases/L018.yml
```diff
@@ -11,28 +11,19 @@ test_pass_with_clause_closing_oneline:
...
   pass_str: with cte as (select 1) select * from cte
 
-test_fail_with_clause_closing_misaligned_indentation:
+test_pass_with_clause_closing_misaligned_indentation:
   # Fix with statement indentation
-  fail_str: |
+  pass_str: |
     with cte as (
         select 1
         ) select * from cte
```
tunetheweb, line 19: To me this is a regression. This should fail IMHO and be fixed.

Instead of aligning with the `WITH` should we align with wherever the CTE opening bracket line starts? (which in this case is at the start of the `WITH`, but in your templated example might not be.).

test/fixtures/rules/std_rule_cases/L018.yml
```diff
@@ -11,28 +11,19 @@ test_pass_with_clause_closing_oneline:
...
-    ) select * from cte
-
-test_fail_with_clause_closing_misaligned_negative_indentation:
+test_pass_with_clause_closing_misaligned_negative_indentation:
   # Fix with statement that has negative indentation
-  fail_str: |2
+  pass_str: |2
         with cte as (
         select 1
     ) select * from cte
```
tunetheweb, line 26: Again I see this as a regression due to this change - this should be fixed.

test/fixtures/rules/std_rule_cases/L018.yml
```diff
@@ -59,3 +50,42 @@ test_pass_cte_with_column_list:
...
+
+  fix_str: |
+    with
+    {% if true %}
+      cte as (
+          select 1
+    )
+    {% else %}
+      cte as (
+          select 2)
```
tunetheweb, line 89: Again I feel this line should be fixed.

src/sqlfluff/rules/L018.py
```diff
@@ -22,25 +21,23 @@
 @document_fix_compatible
 @document_configuration
 class Rule_L018(BaseRule):
-    """``WITH`` clause closing bracket should be aligned with ``WITH`` keyword.
+    """``WITH`` clause closing bracket should be on a new line.
```
tunetheweb, line 24: As mentioned elsewhere I think the rule description should be changed to "Common Table Expressions should have the closing bracket aligned to the opening bracket line". Or something like that.

tunetheweb (changes requested):

src/sqlfluff/rules/L018.py
tunetheweb: Ah I missed that. In that case a lot of my review may be rubbish :-)

test/fixtures/linter/autofix/ansi/027_l003_l018_templated_block_without_newline_cte/after.sql
barrywhart: No, because this part of the template is not rendered (it's in the `else` block of an `if true`.) If this `if` statement were inside a loop that at times evaluated to both `true` and `false`, then SQLFluff would "see" this code and fix it. It's an odd side effect of the two phase linting (templating, then parsing).

test/fixtures/linter/autofix/ansi/027_l003_l018_wrong_indent_cte/after.sql
barrywhart: Interesting. Is there a rule which does that? It seems out of scope for this PR. 🤔

test/fixtures/rules/std_rule_cases/L018.yml
barrywhart: IIUC, this PR is updating L018 so it doesn't concern itself with indentation -- it leaves that job to L003. Now L018 is **only** responsible for making sure the closing bracket of a CTE definition is on a separate line.

test/fixtures/rules/std_rule_cases/L018.yml
barrywhart: Seems like it's okay because the `)` is on a line by itself, and that's all L018 will be responsible for after this PR.

test/fixtures/linter/autofix/ansi/007_with_clause/after.sql
```diff
@@ -1,5 +1,5 @@
 -- Dealing with complicated indents before with clauses.
     	WITH cte as (
             select a from tbla
-        )
+)
```
tunetheweb, line 4: I think this test case is a bit pointless. It's only one rule. It shouldn't really be in this folder and should be in the rule YAML file.

However, now we are making L003 do the indentation I think we could leave it here, if we added L003 to the test config and then make sure that the closing bracket is aligned. In which case we can revert this change.

test/fixtures/linter/autofix/ansi/027_l003_l018_templated_block_without_newline_cte/after.sql
tunetheweb: OK makes sense, but not obvious. Can we add a comment to that effect?

test/fixtures/rules/std_rule_cases/L018.yml
tunetheweb: Ignore. L003 will handle this. Resolving.

test/fixtures/linter/autofix/ansi/027_l003_l018_templated_block_without_newline_cte/after.sql
barrywhart: Could we change this test to include a loop and an if? Something like:
```
WITH
{% for i in [0, 1] %}
    {% if i == 0 %}
        cte0 AS (
            SELECT 1
        )
    {% else %}
        cte1 AS (
            SELECT 2)
    {% endif %}
{% endfor %}
SELECT * FROM cte1
UNION ALL
SELECT * FROM cte2
```

test/fixtures/linter/autofix/ansi/027_l003_l018_wrong_indent_cte/after.sql
tunetheweb: Good point. L022 handles that. Maybe should be added to this test config?

barrywhart: Sounds good to me. I see these tests as narrowly defined real-world tests, i.e. choose the minimum set of rules to do a nice job on the query.

test/fixtures/linter/autofix/ansi/007_with_clause/after.sql
zidder: Added the `newline` in both `before.yml` and `after.yml`

test/fixtures/linter/autofix/ansi/007_with_clause/after.sql
zidder: Added the L003 and changed the `after.yml`

tunetheweb:
Good work @zidder !

zidder:
Thanks for being patient and reviewing my pull request :)

Issue #3390, opened 2022-05-22
Dialects: Deprecate as many RegexParsers as we can


WittierDinosaur:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

RegexParser doesn't support a hash matching - making it far less performant than its StringParser equivalent. In an ideal world everything would be direct from the Lexer, but if that can't be, StringParser is better. I doubt that we can remove every instance of this, but the more we remove, the better.

### Dialect

All

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

barrywhart:
Can you explain what you mean by hash matching? Is this something in the `StringParser` code itself, or something about how Python string `==` works?

WittierDinosaur:
i just mean the simple() matching route in the BaseSegment

alanmcruickshank:
I might be able to have a go at this while I try and unpick the remnants of #3692. I think they're related.

alanmcruickshank:
To make progress on this my suggestion builds on #3819.

The introduction of `TypedMatcher` doesn't directly solve this problem, but I think it provides a potential path forward. The way I see it, the constraints we have are:
- `.simple()` matching is currently tied to string matching.
- Some segments don't have a single string form, or even one which can be addressed using `MultiStringParser`.
- There isn't an obvious way to match very custom strings in the simple route.

Given those constraints I think we need to break one of them. That boils down to making `.simple()` matching more powerful, to do more than just strings, but in a way that is likewise very performant. In my mind there are two options for doing that:
1. Making regexes a first class citizen and being able to effectively look ahead using them.
2. Leveraging the `type` of future segments as a way of doing effective lookaheads.

I think the latter is much more viable. I don't think I can see a way to performantly do the first one.

More specifically I'm suggesting:
- All regex operations are pushed up into the lexing step. That produces a set of raw segments of different types which can be later picked up. We then deprecate the `RegexParser` and replace all use cases with the `TypedParser`.
- We extend the `.simple()` matching route to have two paths:
  - The existing `string` based route where we use `in` and `index` on the raw strings to identify matches ahead in the sequence of segments.
[...]

PR #3510, opened 2022-06-29
Core: Add `MultiStringParser` to match a collection of strings

judahrand:

This parser takes a collection of strings and returns a match if any of
them are found. This is more performant than implementing this through
the `RegexParser` as a hash matching route can be taken.

This looks to at least partially address #3390 and does see a modest 2% speed up in parsing when running `tox -e bench`

Before:
```
==== overall timings ====
Clock time:    10.77
=== templating ===
avg:            0.01
=== lexing ===
cnt:               1 sum:            0.02
min:            0.02 max:            0.02
avg:            0.02
=== parsing ===
cnt:               1 sum:            2.02
min:            2.02 max:            2.02
avg:            2.02
=== linting ===
cnt:               1 sum:            8.68
min:            8.68 max:            8.68
avg:            8.68
===END PROCESS OUTPUT===
Fix command failed with return code: 1
===== Done =====
Run    #0: {'004_L003_indentation_3': 0.5691906670108438, 'B_001_package': 0.7864880409906618}
Run    #1: {'004_L003_indentation_3': 0.40543387504294515, 'B_001_package': 0.675880833005067}
Run    #2: {'004_L003_indentation_3': 0.3828648329945281, 'B_001_package': 0.6732317919959314}
```

After:
```
==== overall timings ====
Clock time:    10.70
=== templating ===
avg:            0.01
=== lexing ===
cnt:               1 sum:            0.03
min:            0.03 max:            0.03
avg:            0.03
=== parsing ===
cnt:               1 sum:            1.98
min:            1.98 max:            1.98
avg:            1.98
=== linting ===
cnt:               1 sum:            8.66
min:            8.66 max:            8.66
avg:            8.66
===END PROCESS OUTPUT===
Fix command failed with return code: 1
===== Done =====
Run    #0: {'004_L003_indentation_3': 1.0236480419989675, 'B_001_package': 0.7732945830211975}
Run    #1: {'004_L003_indentation_3': 0.40719666599761695, 'B_001_package': 0.6686641669948585}
Run    #2: {'004_L003_indentation_3': 0.39612441597273573, 'B_001_package': 0.6604930419707671}
```
### Pull Request checklist

  - Other.


src/sqlfluff/core/parser/parsers.py
```diff
@@ -100,6 +100,43 @@ def match(
...
+        type: Optional[str] = None,
+        optional: bool = False,
+        **segment_kwargs,
+    ):
+        self.templates = templates
+        self.raw_class = raw_class
+        self.name = name
+        self.type = type
+        self.optional = optional
+        self.segment_kwargs = segment_kwargs or {}
```
tunetheweb, line 120: Should we uppercase during `_init_` to prevent having to do it for each match?

Also should this just call super like RegexParse does for the other values? In case the base StringParser ever changes in future?

src/sqlfluff/core/parser/parsers.py
```diff
@@ -100,6 +100,43 @@ def match(
...
+
+    def simple(self, parse_context: "ParseContext") -> Optional[List[str]]:
+        """Return simple options for this matcher.
+
+        Because string matchers are not case sensitive we can
+        just return the template here.
+        """
+        return [template.upper() for template in self.templates]
```
tunetheweb, line 128: Can simplify this if uppering at initialisation

src/sqlfluff/core/parser/parsers.py
```diff
@@ -100,6 +100,43 @@ def match(
...
+
+    def _is_first_match(self, segment: BaseSegment):
+        """Does the segment provided match according to the current rules."""
+        # Is the target a match and IS IT CODE.
+        # The latter stops us accidentally matching comments.
+        upper_templates = {template.upper() for template in self.templates}
+        if segment.raw.upper() in upper_templates and segment.is_code:
+            return True
```
tunetheweb, line 136: Can remove need for upper_templates if uppering at init.

Also would it be slightly quicker to do the `is_code` check first? Especially for larger lists?

```py
if segment.is_code and segment.raw.upper() in upper_templates:
```

tunetheweb (changes requested):
This is great! 🚀🚀🚀

Not only is it (slightly) faster, it's also more readable.

Could we add some tests to `test/core/parse/grammar_test.py`? I realise code coverage has returned 100% but still think it would be good to have some explicit tests (positive and negative).

Also have some other feedback below.

judahrand:
>Could we add some tests to test/core/parse/grammar_test.py? I realise code coverage has returned 100% but still think it would be good to have some explicit tests (positive and negative).

@tunetheweb there doesn't seem to be a precedent for how to add tests for a new `Parser`. For example, the `RegexParser` appears just twice in `grammar_test.py` and then only to test other elements of the matching. Did you have an idea of how you'd like the new `Parser` included?

tunetheweb:
Oh that's a good point. I just presumed they were in there when I did a quick search but didn't look at the test themselves!

That's a bit poor of us to be honest. Though, as I say, we still have 100% code coverage as the dialect tests do test.

Stil think it would be good to add. Could we add a simple test case to `test__parser__grammar_oneof`?

judahrand:
@tunetheweb I hope that was what you were after?

src/sqlfluff/core/parser/parsers.py
judahrand: > including setting template to an empty string?
This is what feels odd to me. I'm not sure we should set an attribute which isn't relevant to the class at all.  

> I wonder should StringParser also be changed to upper at __init__ time?
Quite possibly, though I suspect the overhead is very small!

tunetheweb:
Ideally we'd leave the current test along, and add a couple of new ones to be something like:

`Ref.keyword("foo")` matches `MultiStringParser(["foo","bar"], KeywordSegment))`
`Ref.keyword("fo")` does not match `MultiStringParser(["foo","bar"], KeywordSegment))`

judahrand:
@tunetheweb I'm still not convinced `grammar_test.py` is the right place to be adding a test for the `Parser` at all... After all it inherits directly from `Matchable` not `BaseGrammar` and so isn't a 'grammar' at all.

There should probably be a `parser_test.py`?

src/sqlfluff/core/parser/parsers.py
tunetheweb: > This is what feels odd to me. I'm not sure we should set an attribute which isn't relevant to the class at all.

Yeah I get your hesitancy. Let's leave this comment open and let the others weigh in.

tunetheweb:
There is a `parse_test.py`. Maybe belongs in there instead?

judahrand:
@tunetheweb That should be better?

tunetheweb (approved):
Changes LGTM!

I'll leave it open to see if any of the others spot anything.

src/sqlfluff/core/parser/parsers.py
```diff
@@ -100,6 +100,47 @@ def match(
         return MatchResult.from_unmatched(segments)
 
 
+class MultiStringParser(StringParser):
```
barrywhart, line 103: Is there some benefit to inheriting from `StringParser`? I notice that:
* In the constructor, we delete one of the parent class' fields (`template`)
* None of the methods use the superclass

Basically, I'm wondering if this class doesn't need to inherit from `StringParser`.

src/sqlfluff/core/parser/parsers.py
```diff
@@ -100,6 +100,47 @@ def match(
...
+        # Delete attribute which is replaced by `self.templates` for this `Parser``
+        del self.template
+
+    def simple(self, parse_context: "ParseContext") -> Optional[List[str]]:
+        """Return simple options for this matcher.
+
+        Because string matchers are not case sensitive we can
+        just return the templates here.
+        """
+        return list(self.templates)
```
barrywhart, line 133: Is this function called frequently? I notice that we're creating and returning a new list each time, which could be slow.

src/sqlfluff/core/parser/parsers.py
```diff
@@ -100,6 +100,47 @@ def match(
...
+
+    def _is_first_match(self, segment: BaseSegment):
+        """Does the segment provided match according to the current rules."""
+        # Is the target a match and IS IT CODE.
+        # The latter stops us accidentally matching comments.
+        if segment.is_code and segment.raw.upper() in self.templates:
```
barrywhart, line 139: Could we use the `raw_upper` property rather than `raw.upper()`? It's a cached property on `BaseSegment`, and since it's cached, could be faster to access.

barrywhart (commented):
Nice work! I have a couple questions which could potentially improve the speed a bit more.

barrywhart:
@WittierDinosaur: Would you like to review this since it relates to one of the performance issues you created?

src/sqlfluff/core/parser/parsers.py
judahrand: There are other methods which are used from StringParser. Though, you are right that the inheritance is odd. I reckon there should be a `BaseParser` perhaps which inherits from `Matchable` and implements the reusable bits of `StringParser` given that it looks to me like all other Parsers inherit from `StringParser`?

src/sqlfluff/core/parser/parsers.py
judahrand: The compromise here was whether to cache the `set` which is used for matching or to cache the `list` which is expected to be returned by `simple`. We could cache both as class attributes?

src/sqlfluff/core/parser/parsers.py
judahrand: Sounds good - should probably be done `StringParser` too

src/sqlfluff/core/parser/parsers.py
judahrand: Had a go at this and keen for feedback.

src/sqlfluff/core/parser/parsers.py
judahrand: I've implemented a `self._simple` which is set in `__init__` which should avoid creating the `list` multiple times (this goes for `StringParser` too).

alanmcruickshank (approved):
Neat - I like this. The regex construction for the set matching was a bit hacky and this provides a really neat solution to that.

Good work 👍 .

WittierDinosaur:
@barrywhart I would, but unfornately I'm on holiday now and for a couple of weeks. I'll evaluate it post-hoc when I'm back, but conceptually I like the idea. I'll leave it to you guys on the implementation!

tunetheweb:
OK in that case I vote to merge rather than waiting for you.

@barrywhart could you review the changes for your feedback?

barrywhart (approved):

PR #3511, opened 2022-06-29
Core: Compile regexes at init time to avoid overhead

judahrand:

This means that the regexes are only compiled once rather than everytime
a `RegexParser` is evaluated.

Before:
```
==== overall timings ====
Clock time:    10.86
=== templating ===
avg:            0.01
=== lexing ===
cnt:               1 sum:            0.03
min:            0.03 max:            0.03
avg:            0.03
=== parsing ===
cnt:               1 sum:            2.02
min:            2.02 max:            2.02
avg:            2.02
=== linting ===
cnt:               1 sum:            8.78
min:            8.78 max:            8.78
avg:            8.78
===END PROCESS OUTPUT===
Fix command failed with return code: 1
===== Done =====
Run    #0: {'004_L003_indentation_3': 0.6412674590246752, 'B_001_package': 0.7322651249705814}
Run    #1: {'004_L003_indentation_3': 0.4330607500160113, 'B_001_package': 0.67768708296353}
Run    #2: {'004_L003_indentation_3': 0.4425301660085097, 'B_001_package': 0.6980577500071377}
```

After:
```
==== overall timings ====
Clock time:    10.76
=== templating ===
avg:            0.01
=== lexing ===
cnt:               1 sum:            0.03
min:            0.03 max:            0.03
avg:            0.03
=== parsing ===
cnt:               1 sum:            1.97
min:            1.97 max:            1.97
avg:            1.97
=== linting ===
cnt:               1 sum:            8.72
min:            8.72 max:            8.72
avg:            8.72
===END PROCESS OUTPUT===
Fix command failed with return code: 1
===== Done =====
Run    #0: {'004_L003_indentation_3': 0.5703778340248391, 'B_001_package': 0.7369135419721715}
Run    #1: {'004_L003_indentation_3': 0.3819767920067534, 'B_001_package': 0.6609561250079423}
Run    #2: {'004_L003_indentation_3': 0.4420522919972427, 'B_001_package': 0.6928695419919677}
```

### Pull Request checklist

  - Other.


tunetheweb:
I had a look at doing this before but then gave up because apparently it doesn't make a difference since https://stackoverflow.com/a/452143

WDYT?

judahrand:
> I had a look at doing this before but then gave up because apparently it doesn't make a difference since https://stackoverflow.com/a/452143
> 
> WDYT?



> I had a look at doing this before but then gave up because apparently it doesn't make a difference since https://stackoverflow.com/a/452143
> 
> WDYT?

Interesting, however, `sqlfluff` doesn't appear to be using the standard library (`re`) and so the `_compile` function is actually this one: https://github.com/mrabarnett/mrab-regex/blob/8a514bc62fbfe0006e966124a993562fd4969ca5/regex_3/regex.py#L453-L495 which appears to do quite a lot more than the one in `re` so there is likely more benefit than if we were using the standard library where the cache lookup is trivial?

tunetheweb:
Interesting. This was introduced in #2027. I that case there might be a benefit.

tunetheweb:
Actually that seems to suggest there is a cache in there, and it's cached first time it runs. Any thoughts on why you're still seeing a performance improvement?

judahrand:
> Actually that seems to suggest there is a cache in there, and it's cached first time it runs. Any thoughts on why you're still seeing a performance improvement?

There's absolutely a cache, wasn't saying there wasn't. But there's also quite a lot of object creation, comparisons, etc. going on in order to create the key to look up in `regex` vs `re`. It's still probably a small gain but larger than if we were using `re`.

If you don't think it is worth the change for what is probably a small performance improvement (but there definitely would be one - few function calls / lookups / object creations isn't going to be slower!) we can close this but I also don't really see the downside to compline the regexes up front?

barrywhart:
Also note that the `regex` cache appears to be [limited to 500 entries](https://github.com/mrabarnett/mrab-regex/blob/8a514bc62fbfe0006e966124a993562fd4969ca5/regex_3/regex.py#L450). I don't know if we're hitting that limit, but if we are, some regexes will get recompiled.

IIRC, @WittierDinosaur switched SQLFluff to using `regex` because it supported some feature(s) he needed that aren't available in standard `re`.

barrywhart (approved):
This change looks good to me. Regardless of whether it's faster, it's a simple change that's very unlikely to be **slower**.

Also note that performance improvements often have a much bigger impact on huge real-world SQL files. By comparison, I think our existing benchmark files are at most "medium sized".

PR #3528, opened 2022-06-30
Disable L059 by default for Postgres

### LINKED ISSUE #3485: L059 Unnecessary quoted identifier on PostgreSQL causing case sensitivity problems
### Search before asking

### What Happened

Running sqlfluff-lint and sqlfluff-fix through pre-commit, many previously quoted column names were unquoted which caused the SQL to error because the column names were now incorrect.

(These columns are pre-existing and I cannot rename them.)

### Expected Behaviour

It should not be considered unnecessary quoting if there are any uppercase characters within the quotes, as it changes the meaning of the SQL.

### Observed Behaviour

In PostgreSQL, the column name `"Column"` is not the same as `column`, but sqlfluff will consider the former a violation of L059 and rewrite it as the latter, which then causes an error when trying to execute the query.

> Quoting an identifier also makes it case-sensitive, whereas unquoted names are always folded to lower case. For example, the identifiers FOO, foo, and "foo" are considered the same by PostgreSQL, but "Foo" and "FOO" are different from these three and each other. (The folding of unquoted names to lower case in PostgreSQL is incompatible with the SQL standard, which says that unquoted names should be folded to upper case. Thus, foo should be equivalent to "FOO" not "foo" according to the standard. If you want to write portable applications you are advised to always quote a particular name or never quote it.)

### How to reproduce

Here are a couple of simplified examples that illustrate the erroneous behaviour. It seems that it does quite inconsistent things depending on the surrounding context.

```sql
SELECT T."ColumnName"
FROM schema_name.table_name AS T;
```
becomes
```sql
SELECT SCHEMA_NAME.TABLE_NAME.COLUMNNAME
FROM SCHEMA_NAME.TABLE_NAME;
```
And a slightly different one:
```sql
SELECT "ColumnName"
FROM schema_name.table_name;
```
becomes
```sql
SELECT ColumnName
FROM Schema_Name.Table_Name;
```
Another:
```sql
INSERT INTO public.other_table (test_id, request_year)
SELECT "Test",
       "ReqYear"
FROM my_schema.my_table;
```
becomes
```sql
INSERT INTO public.other_table (test_id, request_year)
SELECT
    test,
    reqyear
FROM my_schema.my_table;
```

### Dialect

`postgres`

### Version

sqlfluff 1.0.0
python 3.10.2

### Configuration

`.sqlfluff`:
```
[sqlfluff]
dialect = postgres

[sqlfluff:rules]
max_line_length = 100
```

### Code of Conduct

tunetheweb:

Fixes #3485 

Nope

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L059.py
+++ b/src/sqlfluff/rules/L059.py
@@ -27,7 +27,7 @@ class Rule_L059(BaseRule):
     unnecessary, except for reserved keywords and special characters in identifiers.
 
     .. note::
-       This rule is disabled by default for Postgres abd Snowflake because the allow
+       This rule is disabled by default for Postgres and Snowflake because they allow
        quotes as part of the column name. In other words, ``date`` and ``"date"`` are
        two different columns.
 
--- a/src/sqlfluff/rules/L059.py
+++ b/src/sqlfluff/rules/L059.py
@@ -27,9 +27,9 @@ class Rule_L059(BaseRule):
     unnecessary, except for reserved keywords and special characters in identifiers.
 
     .. note::
-       This rule is disabled by default for Snowflake because it allows quotes as
-       part of the column name. In other words, ``date`` and ``"date"`` are two
-       different columns.
+       This rule is disabled by default for Postgres abd Snowflake because the allow
+       quotes as part of the column name. In other words, ``date`` and ``"date"`` are
+       two different columns.
 
        It can be enabled with the ``force_enable = True`` flag.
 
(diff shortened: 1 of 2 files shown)


src/sqlfluff/rules/L059.py
```diff
@@ -27,9 +27,9 @@ class Rule_L059(BaseRule):
     unnecessary, except for reserved keywords and special characters in identifiers.
 
     .. note::
-       This rule is disabled by default for Snowflake because it allows quotes as
-       part of the column name. In other words, ``date`` and ``"date"`` are two
-       different columns.
+       This rule is disabled by default for Postgres abd Snowflake because the allow
```
greg-finley, line 30:
```suggestion
       This rule is disabled by default for Postgres and Snowflake because the allow
```

barrywhart (approved):

src/sqlfluff/rules/L059.py
tunetheweb: Good spot - thanks!

PR #3552, opened 2022-07-04
Groundwork for a fix routine for L046

alanmcruickshank:
I'm working on a fix routine for L046, which is a bit different because we're making changes in the _source_ file which aren't in the _templated_ file.

To support that work, it needed a little bit of a refactor of L046, which I'm putting here as an initial PR so that the rest of the work isn't too big (it's already getting quite big and deep).

The existing tests should pass as usual, and this approach should also handle some more complicated cases a little better (see the new test). I've also added a commented out line to show how the new fixed version will be calculated when the mechanisms to apply it actually exist.

I've also added some more helpful logging to the rule, which explains some of the additional lines of code.


alanmcruickshank:
Ok, I've scaled back my initial changes because I now understand better why they were done that way in the first place. The new edits address the new test case, but make fewer changes which mean less additional code for the next PR. I think this is now ready for review @barrywhart

src/sqlfluff/rules/L046.py
```diff
@@ -38,21 +43,25 @@ class Rule_L046(BaseRule):
     targets_templated = True
 
     @staticmethod
-    def _get_whitespace_ends(s: str) -> Tuple[str, str, str]:
+    def _get_whitespace_ends(s: str) -> Tuple[str, str, str, str, str]:
```
barrywhart, line 46: Could you add a doctest example for this routine? I.e. input string, then the values it returns?

https://docs.python.org/3/library/doctest.html

This will be useful mainly for documentation, but I _think_ pytest will automatically detect and execute it as well (much like a "normal" Python test).

barrywhart (approved):
Looks good!

Happy to see you're looking into this area. Interesting possibilities here! IIRC, there's an enhancement issue about autoformatting the code **inside** the Jinja curly braces (e.g. with dbt, when you call a macro that has lots of functions and is spread across multiple lines, the indentation may need fixing).

src/sqlfluff/rules/L046.py
alanmcruickshank: Great idea.

alanmcruickshank: Good idea. I've added doctests - but they weren't being picked up by tox, so I've added an explicit tox environment to run them. Not many in there yet but it's a useful environment for them.

Issue #3496, opened 2022-06-27
Lint templated file before evaluation


anna-azizian:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

Lint the actual templated files instead of evaluating `if`s and `for`s.

### Use case

Currently, not all errors are found in the following SQL
```sql
SELECT *
FROM T
{% if some_jinja_condition %}
    where some_condition
{% else %}
    where some_other_condition
{%endif%}
```
If keywords are enforced to be uppercase, the linter finds only one of `where`s depending on how will the `some_jinja_condition` be evaluated. Instead, it could scan the whole document without looking at those templates (for some rules). Thus, the expected result would be 2 errors suggesting to make `where` uppercase.

### Dialect

N/A

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

barrywhart:
The only way to do this would be to lint the file multiple times with different settings.

barrywhart:
@anna-azizian: Do you use the Jinja templater or the dbt templater? Eventually, it would be good to support this for both templaters, but Jinja would probably be easier to support first.

barrywhart:
Also, if you can provide an example that looks a little more like real code, that would be helpful. Not absolutely necessary, but it may help us understand the expected usage better.

Do you expect to need this for both Jinja variables and Jinja macros? (Your example uses a variable.)

anna-azizian:
@barrywhart Currently I'm using dbt templater, but having jinja support first would be fine.
Here is a code example with macro usage (which is similar to real-world example). Also, L003 suggests a wrong indentation for GROUP BY line, which doesn't happen without the template section.

```sql
SELECT
    a,
    COUNT(*) AS cnt
FROM tbl
WHERE first_condition
    {% if is_incremental() %}
        AND second_condition
    {% endif %}
GROUP BY a
```

alanmcruickshank:
I'm going to attack L003 as part of #3673 which might address solving some of this.

alanmcruickshank:
I'm going to close this issue. I think the templated indents are solved as part of #3673. The part about linting unrendered sections of code isn't yet, but I think that's covered as part of #4061 .

Issue #3133, opened 2022-04-21
L003: Jinja directive outside of SELECT interfers with Jinja conditionals inside 


dmohns:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Linting the following SQL produces unexpected results:

```sql
{% set test = (1 == 1) %}

SELECT *
FROM table_1
WHERE TRUE
{% if test %}
AND TRUE
{% endif %}
```

### Expected Behaviour

```sql
{% set test = (1 == 1) %}

SELECT *
FROM table_1
WHERE TRUE
    {% if test %}
        AND TRUE
    {% endif %}
```

This is based on the result of: https://online.sqlfluff.com/fluffed?sql=H4sIACcmYWIC_wt29XF1DlHQ4uVyC_L3VShJTMpJjTfk5Qr3cA1yVQgJCnXl5apWVchMU9AwVLC1VTDUVFCt5eVy9HNBSKbmpQDlQcIAJZvHVk4AAAA%3D&dialect=ansi

```sql
SELECT *
FROM table_1
WHERE TRUE
{% if (1 == 1) %}
AND TRUE
{% endif %}
```

produces

```sql
SELECT *
FROM table_1
WHERE TRUE
    {% if (1 == 1) %}
        AND TRUE
    {% endif %}
```

### Observed Behaviour

```sql
{% set test = (1 == 1) %}

SELECT *
FROM table_1
WHERE TRUE
    {% if test %}
    AND TRUE
{% endif %}
```

Note the odd placement of `{% endif %}`

### How to reproduce

https://online.sqlfluff.com/fluffed?sql=H4sIAHYmYWIC_6tWVShOLVEoSS0uUbBV0DBUsLVVMNRUUK3l5eLlCnb1cXUOUdDi5XIL8vdVKElMykmNN-TlCvdwDXJVCAkKdeXlqlZVyEyD6AdpcvRzQUik5qUA5VRrASwShStlAAAA&dialect=ansi

### Dialect

ansi

### Version

0.12.0 -> webversion

### Configuration

default

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

alanmcruickshank:
[...]

PR #3562, opened 2022-07-06
Change From Clause indentations to fix L003 rule on template blocks

zidder:

makes progess on #3496

Indent after `FROM` Key word, not before `FromExpression`.
This makes `autofix` examples nicer, and in case of `indented_joins=true`, it's consistent (and in my opinion, ideal). 
Two yaml tests show, that template from expression must be indented, and group by clause after that should not (didn't work on main branch).

The `FromExpression` has Dedent before Indent, the corresponding Indent and Dedent are in the `FromClause`. In case there are several `FromExpression`s, the indent of the first `FromExpression` is dedented in the second `FromExpression`.

### Pull Request checklist

  - Other.
  - The issue is mentioned in #3496 , and several (#2677, #3133) can serve as a good followups. Can be solved in case we can dedent after empty template segment, or take this into consideration when linting those empty sections.


zidder:
Can you help me fix the `mypy` issue in `dialect_mysql.py`? :)

zidder:
Also, this introduces possibility of subsequent (indent, dedent) or (dedent, indent) segments. I think it would be better to remove those, even if they're in different segments (but are subsequent in raw segments).

barrywhart:
This looks good to me. Tagging @tunetheweb for review, as he is more familiar with reviewing dialect changes.

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -531,6 +534,18 @@ class InsertStatementSegment(BaseSegment):
...
+
+    type = "indented_from_expression"
+
+    match_grammar = Sequence(
+        Indent,
+        Ref("FromExpressionSegment"),
+        Dedent,
+    )
+
+
```
tunetheweb, line 548: I don't see the need for this class, and the extra level it creates in the parse tree. I agree it is used twice, but think cleaner to just `Indent`/`Dedent` directly in both those occasions - like the other dialects do.

This would also reduce some of the changes in these PR (including most of the mysql YAML changes?).

```suggestion
```

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -543,7 +558,7 @@ class DeleteUsingClauseSegment(BaseSegment):
     parse_grammar = Sequence(
         "USING",
         Delimited(
-            Ref("FromExpressionSegment"),
+            Ref("IndentedFromExpressionSegment"),
```
tunetheweb, line 561:
```suggestion
            Indent,
            Ref("FromExpressionSegment"),
            Dedent,
```

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -2008,7 +2023,7 @@ class UpdateStatementSegment(BaseSegment):
         "UPDATE",
         Ref.keyword("LOW_PRIORITY", optional=True),
         Ref.keyword("IGNORE", optional=True),
-        Delimited(Ref("TableReferenceSegment"), Ref("FromExpressionSegment")),
+        Delimited(Ref("TableReferenceSegment"), Ref("IndentedFromExpressionSegment")),
```
tunetheweb, line 2026:
```suggestion
        Delimited(Ref("TableReferenceSegment"), Indent, Ref("FromExpressionSegment"), Dedent),
```

src/sqlfluff/rules/L003.py
```diff
@@ -79,10 +79,16 @@ def finalise(self, line_no: int, templated_file: Optional[TemplatedFile]):
...
+            copied_line_buffer, templated_file
+        )
         has_placeholder = any(
             elem.is_type("placeholder") for elem in copied_line_buffer
         )
-        is_empty_line = not has_code_segment and not has_placeholder
+
+        is_empty_line = not has_code_segment and (
+            not has_placeholder or templated_line_type == "end"
+        )
```
tunetheweb, line 91: Why is this change needed? Should we have a comment here to explain?

test/fixtures/linter/autofix/bigquery/003_templating/after.sql
```diff
@@ -14,15 +14,15 @@ SELECT
...
-        -- This next section gets very strange with forced template_blocks_indent
-        {{action}}_raw_effect_sizes
-    {% else %}
-    JOIN
+    {% for action in considered_actions %}
+        {% if loop.first %}
+            -- This next section gets very strange with forced template_blocks_indent
+            {{action}}_raw_effect_sizes
+        {% else %}
+        JOIN
```
tunetheweb, line 22: Is this not indented too much now?

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -1100,3 +1100,51 @@ test_tsql_bubble_up_newline_after_fix:
...
+            1 AS col
+        {% else %}
+            2 AS col
+        {% endif %}
+    FROM
+        {% if true %}
+            a
+        {% else %}
+        b -- won't be fixed as it's in else block
+    {% endif %} -- won't indent, as it is outside from block
```
tunetheweb, line 1133: Not sure I like this. While I agree it's difficult to make the `else` clause indent, surely the `{% endif %}` should be recognised as part of the `FROM` block?

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1280,6 +1279,7 @@ class FromExpressionSegment(BaseSegment):
             Ref("JoinClauseSegment"), Ref("JoinLikeClauseGrammar"), optional=True
         ),
         Conditional(Dedent, indented_joins=True),
+        Indent,
```
tunetheweb, line 1282: Not sure about this change - see comment later.

test/fixtures/templater/jinja_l_metas/001.yml
```diff
@@ -28,16 +28,18 @@ file:
...
               - from_expression:
-                  indent: ''
                   from_expression_element:
                     table_expression:
                       table_reference:
                       - identifier: DATAHUB
                       - dot: .
                       - identifier: SNAPSHOT_DAILY
                   dedent: ''
+                  indent: ''
```
tunetheweb, line 41: I don't understand why we have these two? A `dedent` immediately followed by an `indent`? Is this because of the first change above on 1271/1282?

tunetheweb (changes requested):
Think we should revert the mysql changes, and have questions on the others.

zidder:
Sorry, can't find the right way to implement this :(

PR #3567, opened 2022-07-07
Simplify lint fixing (prep for source fixes)

alanmcruickshank:
This is in preparation for allowing source fixes. Lots of deletions 🚀 

Two main changes I'm making here:
1. We used to have `FixPatch` (which only had templated position) and `EnrichedFixPatch` (which had templated and source position). I've combined these into one class because I think we can always go straight to creating the latter. Perhaps confusingly, I've renamed the latter to the former, so now we just have `FixPatch`, but it also has templated position. This involved pushing a little logic upstream but not much.
2. In `.fix_string()` there were a few `if` statements for conditions that haven't been covered with tests and I've tried quite hard to trigger them. I'm relatively sure that the new template tracing logic means that our positioning is now good enough that we don't need them anymore. They were useful when tracing was less precise, in that they handled potential ambiguity in the positioning. Without a test case I think it's going to be challenging to support those code paths, so I've added warning asking people to submit queries if they would ever have been used.

I'm getting closer to actually adding functionality! I promise. It's just involving picking through a load of very old code which I'm going to need to rethink a bit.

### DIFF
--- a/src/sqlfluff/core/linter/linted_file.py
+++ b/src/sqlfluff/core/linter/linted_file.py
@@ -27,7 +27,6 @@ from sqlfluff.core.errors import (
     SQLLintError,
     CheckTuple,
 )
-from sqlfluff.core.string_helpers import findall
 from sqlfluff.core.templaters import TemplatedFile
 
 # Classes needed only for type checking
@@ -326,7 +325,7 @@ class LintedFile(NamedTuple):
                 filtered_source_patches.append(patch)
                 dedupe_buffer.append(patch.dedupe_tuple())
             else:
-                NotImplementedError( # pragma: no cover
+                NotImplementedError(  # pragma: no cover
                     "Template tracing logic means that this situation should "
                     "never occur. Please report this as an issue on GitHub "
                     "with a version of your query which triggers this "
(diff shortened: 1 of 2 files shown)


src/sqlfluff/core/linter/linted_file.py
```diff
@@ -315,88 +302,66 @@ def fix_string(self) -> Tuple[Any, bool]:
...
-                    findall(enriched_patch.templated_str, enriched_patch.source_str)
-                )
+                positions = list(findall(patch.templated_str, patch.source_str))
                 if len(positions) != 1:
-                    linter_logger.debug(
+                    # NOTE: This section is not covered in tests. While we
+                    # don't have an example of it's use (we should), the
+                    # code after this relies on there being only one
+                    # instance found - so the safety check remains.
+                    linter_logger.debug(  # pragma: no cover
```
barrywhart, line 341: Possible future work: Could we avoid the possibility of this occurring by allowing (requiring) rules to provide additional position info when making changes in templated areas?

barrywhart (approved):
Looks good! Happy to see more simplification here. I've removed several bits of other "safety" code over the past year or so now that slicing is more accurate. (100% accurate, I think -- now it logs an error if the positioning seems inaccurate, i.e. adjacent slice ends don't match up.)

Issue #3494, opened 2022-06-27
Postgres PRS issue on COLLATION/COLLATE


jonyscathe:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

I have a custom COLLATION type: https://www.postgresql.org/docs/current/sql-createcollation.html

That line is not parsable.
And then when I use that custom type in a table that is also not parsable.

### Expected Behaviour

Expect it to be parsable

### Observed Behaviour

```
== [test.sql] FAIL                                                                                                                    
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'CREATE
                       | COLLATION numeric (provider = icu...'
L:   3 | P:   1 |  PRS | Line 3, Position 1: Found unparsable section: 'CREATE
                       | TABLE orders\n(\n id bigint NOT ...'
WARNING: Parsing errors found and dialect is set to 'postgres'. Have you configured your dialect correctly?
```

### How to reproduce

CREATE COLLATION numeric (provider = icu, locale = 'en@colNumeric=yes');

CREATE TABLE orders
(
    id bigint NOT NULL DEFAULT NEXTVAL('orders_id_seq'::regclass),
    order_number text NOT NULL COLLATE numeric
)

### Dialect

postgres

### Version

1.0.0

### Configuration

```
[tool.sqlfluff.core]
dialect = 'postgres'
exclude_rules = 'L008, L044'
sql_file_exts = '.sql'

[tool.sqlfluff.indentation]
indented_ctes = true
indented_joins = false
indented_using_on = true
template_blocks_indent = true

[tool.sqlfluff.rules]
allow_scalar = true
comma_style = 'trailing'
indent_unit = 'space'
max_line_length = 120
tab_space_size = 4
unquoted_identifiers_policy = 'all'

[tool.sqlfluff.rules.L010]
capitalisation_policy = 'upper'

[tool.sqlfluff.rules.L014]
extended_capitalisation_policy = 'lower'

[tool.sqlfluff.rules.L028]
single_table_references = 'qualified'

[tool.sqlfluff.rules.L029]
ignore_words = 'main'

[tool.sqlfluff.rules.L030]
extended_capitalisation_policy = 'upper'

[tool.sqlfluff.rules.L040]
capitalisation_policy = 'lower'
ignore = 'NULL'

[tool.sqlfluff.rules.L059]
[...]

PR #3571, opened 2022-07-07
Postgres: CREATE COLLATION

greg-finley:

Closes https://github.com/sqlfluff/sqlfluff/issues/3494

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_postgres_keywords.py
```diff
@@ -925,18 +925,21 @@ def get_keywords(keyword_list, keyword_type):
     ("DATE", "non-reserved"),
     ("DEPTH", "non-reserved"),
     ("DESCRIBE", "non-reserved"),
+    ("DETERMINISTIC", "non-reserved"),
```
greg-finley, line 928: If someone can explain what goes into `postgres_docs_keywords` vs `postgres_nondocs_keywords`, it would be great to document that in the code. I just guessed.

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1420,13 +1420,15 @@ class CreateTableStatementSegment(ansi.CreateTableStatementSegment):
...
-                                Sequence(
-                                    "COLLATE",
-                                    Ref("QuotedLiteralSegment"),
-                                    optional=True,
-                                ),
                                 AnyNumberOf(
-                                    Ref("ColumnConstraintSegment", optional=True)
+                                    OneOf(
+                                        Ref("ColumnConstraintSegment", optional=True),
+                                        Sequence(
```
barrywhart, line 1426: If I read this correctly, this allows any number of `COLLATE`s, e.g.:
```
COLLATE a COLLATE b COLLATE c
```

Is this the intent? I don't see any tests doing this. Should we add one?

barrywhart: From reading the docs, I think only **one** `COLLATE` clause is permitted.

https://www.postgresql.org/docs/current/collation.html

greg-finley: Related to the test I put, just ran this in Postgres and it worked:

```
gregoryfinley=# CREATE TABLE orders3
(
order_number text NOT NULL COLLATE numeric,
order_number2 text COLLATE numeric NOT NULL
);
CREATE TABLE
gregoryfinley=# 
```

barrywhart: That makes sense, and it should be okay. The dialect is often more permissive/flexible than what is actually allowed (or makes sense to do).

I'd like @tunetheweb to review this PR as well, since he is more the expert on dialect changes. He's on vacation right now, so I'm doing a preliminary review.

barrywhart: Thanks for contributing!

It's funny, 1-2 years ago, we naively thought we'd actually finish with the dialects someday. There is a never-ending supply of missing things, even with the popular dialects like Postgres, Snowflake, and BigQuery. Incredible. 🤯

greg-finley: Yep! I have considered crazy ideas like, could we build a parser to auto-convert Postgres (etc.) docs into sqlfluff match grammar. 🤔

barrywhart: I've also wondered, at least for the commercial data stores, could we get vendors to do the work.

greg-finley: Or sponsor the project 🤞 .

tunetheweb: `AnyNumberOf` supports a `max_times_per_element`, which if set to 1 is the same as `AnySetOf` so can use that instead

tunetheweb: This works:

```py
                                AnySetOf(
                                    Sequence(
                                        "COLLATE",
                                        optional=True,
                                    ),
                                    AnyNumberOf(
                                        Ref("ColumnConstraintSegment"),
                                        optional=True,
                                    ),
                                ),
```

Note it won't allow `ColumnConstraintSegment` + `COLLATE clause` + `ColumnConstraintSegment` if that's valid?

src/sqlfluff/dialects/dialect_postgres_keywords.py
tunetheweb: @WittierDinosaur ?

I'm guessing from the name that `postgres_nondocs_keywords` refers to keywords that appear to be keywords but are not listed in https://www.postgresql.org/docs/current/sql-keywords-appendix.html

greg-finley: Does the distinction even matter? Does it matter if I've put in the "wrong" place for now?

tunetheweb: No it doesn’t presently. Type is important but not which list it appears in.

src/sqlfluff/dialects/dialect_postgres.py
greg-finley: You can do a `ColumnConstraintSegment` before and after:

```
gregoryfinley=# CREATE TABLE orders4
(
order_number2 text PRIMARY KEY COLLATE numeric NOT NULL
);
CREATE TABLE
gregoryfinley=# 
```

Maybe what we have currently is fine?

tunetheweb: Yeah I’m happy to merge as is if you are. Won’t be the first place the syntax is slightly overly permissive and usually that’s better than overly restrictive.

greg-finley: Yeah I'm ok with it. I pushed another commit with more tests on multiple column constraints: https://github.com/sqlfluff/sqlfluff/pull/3571/commits/e5ad52363566221e50d7c701fc80dd05810245fb

tunetheweb: You actually removed two
of the tests and replaced it with one all-inclusive test. I wonder if we should keep those two tests in though as well, in case anyone removes the all-inclusive test as an unlikely, unnecessary complication but doesn’t then realise order is unimportant too?

greg-finley: I think I did the opposite, moving one CREATE to two? I did just to use PRIMARY KEY in different places in the syntax.

greg-finley: I think this works better: https://github.com/sqlfluff/sqlfluff/pull/3571/commits/010cb8b298781d9158bb7d125ecdf1f88956d8ac

tunetheweb: Ok but now we don’t have the `CONSTRAINT` + `COLLATE` + `CONSTRAINT` example so someone could try to clean this up in future using the `AnySetOf` syntax I proposed above and not realise we specifically choose NOT to use that. So I think you need all three tests for complete coverage?

Though, to be honest, think it’s pretty unlikely anyone would actually use that third type, but there’s all sorts out there!!

greg-finley: I think my test _is_ an example of `CONSTRAINT` + `COLLATE` + `CONSTRAINT`, isn't it?

`order_number text UNIQUE COLLATE numeric NOT NULL`

`UNIQUE` is a constraint, then `COLLATE`, then `NOT NULL` is a second constraint.

Or am I missing something?

tunetheweb: Ah you’re right. What you’re missing then is `COLLATE` + `CONSTRAINT` example.

Though really getting picky now so happy enough to just merge if you prefer?

greg-finley: This should be everything? Let me know if not: https://github.com/sqlfluff/sqlfluff/pull/3571/commits/0f95ac5bb961755638ff2f9dfcdffeedbb804818

tunetheweb (approved):

PR #3576, opened 2022-07-08
Support MySQL system variables

### LINKED ISSUE #3574: Add missing support for MySQL system variables
### Search before asking

### What Happened

MySQL allows the use of system variables in queries, using @@global or @@session as a prefix. This is especially useful to for timezone support, see the [official docs](https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html).

```sql
SELECT @@GLOBAL.time_zone, @@SESSION.time_zone;
```

When trying to convert a timestamp from the server timezone to UTC, I found that SQLFluff doesn't support system variables.

```shell
$ echo 'SELECT @@global.time_zone' | sqlfluff parse - --dialect mysql

L:  1, P:  1]      |file:
[L:  1, P:  8]      |                    column_reference:
[L:  1, P:  8]      |                        variable:                             '@'
[L:  1, P:  9]      |                    alias_expression:
[L:  1, P:  9]      |                        variable:                             '@global'
[L:  1, P: 16]      |                unparsable:                                   !! Expected: 'Nothing...'
[L:  1, P: 16]      |                    raw:                                      '.'
[L:  1, P: 17]      |                    raw:                                      'time_zone'
[L:  1, P: 26]      |            [META] dedent:
[L:  1, P: 26]      |    newline:                                                  '\n'

==== parsing violations ====
L:   1 | P:  16 |  PRS | Line 1, Position 16: Found unparsable section: '.time_zone'
```

However I didn't always get that parsing error. While trying different queries, I realized there's also a bug in the lexer, because an "@" alone shouldn't be considered a valid variable (like above). This makes the following query valid but the system variable is incorrectly parsed as a column reference.

```shell
$ echo 'SELECT CONVERT_TZ(NOW(), @@global.time_zone, "+00:00")' | sqlfluff parse - --dialect mysql
[L:  1, P:  1]      |file:
[L:  1, P:  8]      |                    function:
[L:  1, P:  8]      |                        function_name:
[L:  1, P:  8]      |                            function_name_identifier:         'CONVERT_TZ'
[L:  1, P: 18]      |                        bracketed:
[L:  1, P: 18]      |                            start_bracket:                    '('
[L:  1, P: 19]      |                            [META] indent:
[L:  1, P: 19]      |                            expression:
[L:  1, P: 19]      |                                function:
[L:  1, P: 19]      |                                    function_name:
[L:  1, P: 19]      |                                        function_name_identifier:  'NOW'
[L:  1, P: 22]      |                                    bracketed:
[L:  1, P: 22]      |                                        start_bracket:        '('
[L:  1, P: 23]      |                                        end_bracket:          ')'
[L:  1, P: 24]      |                            comma:                            ','
[L:  1, P: 25]      |                            whitespace:                       ' '
[L:  1, P: 26]      |                            expression:
[L:  1, P: 26]      |                                column_reference:
[L:  1, P: 26]      |                                    variable:                 '@'
[L:  1, P: 27]      |                            expression:
[L:  1, P: 27]      |                                column_reference:
[L:  1, P: 27]      |                                    variable:                 '@global'
[L:  1, P: 34]      |                                    dot:                      '.'
[L:  1, P: 35]      |                                    identifier:               'time_zone'
[L:  1, P: 44]      |                            comma:                            ','
[L:  1, P: 45]      |                            whitespace:                       ' '
[L:  1, P: 46]      |                            expression:
[L:  1, P: 46]      |                                literal:                      '"+00:00"'
[L:  1, P: 54]      |                            [META] dedent:
[L:  1, P: 54]      |                            end_bracket:                      ')'
[L:  1, P: 55]      |            [META] dedent:
[L:  1, P: 55]      |    newline:                                                  '\n'
```

### Expected Behaviour

No parsing error for system variables.

### Observed Behaviour

Depending on the query, either a parsing error or an incorrectly parsed query that can generate false positives on linting.

### How to reproduce

`echo 'SELECT @@global.time_zone' | sqlfluff parse - --dialect mysql`

### Dialect

MySQL

### Version

1.1.0

### Configuration

No configuration needed

### Code of Conduct

qgallet:
Fixes #3574 
MySQL system variables `@@global.xxx` and `@@session.xxx` are now recognized by the parser.

None that I can think of.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -710,7 +717,7 @@ class IntervalExpressionSegment(BaseSegment):
     [
         RegexLexer(
             "at_sign",
-            r"[@][a-zA-Z0-9_]*",
+            r"@@?[a-zA-Z0-9_.]*",
```
barrywhart, line 720: I think `.` here means "match any character. Should this be `\.`?

As written, I think it will also allow the name to begin and/or end with a `.`. That worries me a little. Can we prevent this?

barrywhart (changes requested):

src/sqlfluff/dialects/dialect_mysql.py
qgallet: I double checked, metacharacters (except `\`) are matched literally inside a character class in Python regex. This fails as expected on a variable with an invalid char in the name : `echo 'SELECT @var&broken | sqlfluff parse - --dialect mysql`.

For your second point, this indeed looks wrong, but I've found MySQL is so bizarre this is actually valid. The [documentation](https://dev.mysql.com/doc/refman/8.0/en/user-variables.html) states:

> User variables are written as @var_name, where the variable name var_name consists of alphanumeric characters, ., _, and $. A user variable name can contain other characters if you quote it as a string or identifier (for example, @'my-var', @"my-var", or @\`my-var\`).

I hardly believe anyone in their right mind would use a variable name with a `.` at the start or end, so I'm happy to change it to something like `r"@@?[a-zA-Z0-9_]+\.?[a-zA-Z0-9_]*"`.  The `$` sign should be added as well, just let me know what's preferable for the tool and I'll prepare the commit

barrywhart: The above looks close. I think you want the dot and the remainder to look like ‘(\.[a-zA-Z0-9_]+)?`, to prevent it ending with a dot.

qgallet: Alright, I've pushed the change with the $ sign also supported, and added more fixtures for session variables. Thanks for the feedback!

barrywhart (approved):
Looks great, thanks!

PR #3578, opened 2022-07-09
Add source fixing capability and fix routines for L046

alanmcruickshank:
This follows on from the refactor PRs which I've already merged (#3552, #3567).

Quite a big one, but here's the summary:
- Adds much better test coverage for the elements of `.fix_string()`. All part of the new `linted_file_test.py`.
- Adds a new `SourceFix` dataclass and methods on segments to both add them (via `.edit()`) and also to retrieve them.
- Adds capability in the linter and linted file to handle source fixes.
- Updates L046 to use the new capabilty.
- Adds test cases for fixing in L046.

Big PR, and a few places that could use some review.


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1299,6 +1361,25 @@ def _validate_segment_after_fixes(self, rule_code, dialect, fixes_applied, segme
     def _log_apply_fixes_check_issue(message, *args):  # pragma: no cover
         linter_logger.critical(message, exc_info=True, *args)
 
+    def _iter_source_fix_patches(
+        self, templated_file: TemplatedFile
+    ) -> Iterator[FixPatch]:
+        """Yield any source patches as fixes now.
+
+        NOTE: This yields source fixes for the segment and any of it's
```
barrywhart, line 1369:
```suggestion
        NOTE: This yields source fixes for the segment and any of its
```

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -113,6 +116,11 @@ def segments(self):
         """
         return []
 
+    @property
+    def source_fixes(self) -> List[SourceFix]:
+        """Return an source fixes as list."""
```
barrywhart, line 121:
```suggestion
        """Return any source fixes as list."""
```

test/core/linted_file_test.py
```diff
@@ -0,0 +1,350 @@
...
+        )
+    assert result == expected_result
+
+
+@pytest.mark.parametrize(
+    "source_patches,source_only_slices,raw_source_string,expected_result",
+    # NOTE: For all of these examples we're not setting the patch_category
+    # of the fix patches. They're not used at this step so irrelevant for
+    # testing.
+    [
```
barrywhart, line 77: I'm having trouble grasping the test cases for this test `test__linted_file__slice_source_file_using_patches` -- could you add more detailed comments explaining one or two of them in more detail?

alanmcruickshank: I've added some more detailed comments on all of them explaining what's supposed to happen. I think there's definitely some room for streamlining here however. Especially now that the tracing is pretty accurate I wonder whether some of the safety measures in here are a little unnecessary now.

I'm not planning on addressing that here, but I've left the TODOs in the code as a reminder for later than there are some things here that we might want to work on in future. I think I've been looking at this code for too long to see any more obvious quick wins 😬 .

test/fixtures/rules/std_rule_cases/L046.yml
```diff
@@ -31,12 +36,21 @@ test_fail_templated_segment_contains_leading_literal:
...
 
 test_fail_segment_contains_multiple_templated_slices_last_one_bad:
   fail_str: CREATE TABLE `{{ "project" }}.{{ "dataset" }}.{{"table"}}`
+  fix_str: CREATE TABLE `{{ "project" }}.{{ "dataset" }}.{{ "table" }}`
   configs:
     core:
       dialect: bigquery
+
+test_fail_jinja_tags_no_space_no_content:
+  fail_str: SELECT {{-""-}}1
```
barrywhart, line 55: Worth adding more tricky tests where the template output crosses segment boundaries? E.g.

```
SELECT a{{-"1 + b"}}
```

```
SELECT a{{-"1 + b"-}}2
```

alanmcruickshank: ooooh good idea

barrywhart:
I'll try and review tonight. We're on vacation this week.

barrywhart (approved):
Great work! I've thought about this for a while, but you made it happen. 💪 🎉

PR #3580, opened 2022-07-09
Allow no alias for selects in CTEs with a column list

### LINKED ISSUE #3558: [postgres] Complaining about missing alias in CTE
### Search before asking

### What Happened

Linting errors raised where no errors are expected

### Expected Behaviour

No errors.

### Observed Behaviour

```
$ sqlfluff lint --nocolor --verbose --dialect postgres 2022-07-05.cte-column-alias.sql 
==== sqlfluff ====
sqlfluff:                1.1.0 python:                 3.8.10
implementation:        cpython verbosity:                   1
dialect:              postgres templater:               jinja
==== readout ====

=== [ path: 2022-07-05.cte-column-alias.sql ] ===

== [2022-07-05.cte-column-alias.sql] FAIL                                                                                                                                         
L:   2 | P:  15 | L022 | Blank line expected but not found after CTE closing
                       | bracket.
L:   5 | P:   9 | L013 | Column expression without alias. Use explicit `AS`
                       | clause.
L:   6 | P:   9 | L013 | Column expression without alias. Use explicit `AS`
                       | clause.
==== summary ====
violations:        3 status:         FAIL
All Finished!
```

While also (still) raising issue #3474, it also complains about the missing alias as soon as one uses a non-column/non-constant expression.

### How to reproduce

Use the following `2022-07-05.cte-column-alias.sql` file:
```sql
WITH
cte (a, b, c) AS (
    SELECT
        column_a,
        min(column_b),
        sum(column_c)
    FROM my_table
    GROUP BY 1
)

SELECT
    a,
    b,
    c
FROM cte;
```

### Dialect

postgres

### Version

```
sqlfluff:                1.1.0 python:                 3.8.10
implementation:        cpython verbosity:                   1
dialect:              postgres templater:               jinja
```

### Configuration

defaults

### Code of Conduct

pdebelak:

In cases where the CTE has a column list, we don't need an alias for columns. This checks if the select has a `common_table_expression` parent and if that parent has a `cte_column_list` child. In that case, it doesn't complain about a missing alias.

Fixes #3558

I'm not 100% positive that this won't lead to false negatives in some weirdly nested situation. Open to feedback there.

Also, I'm not sure if this rule should be checking if the number of items in the `cte_column_list` matches the number of selects. It seems like that is out of scope for this rule (since it probably means that you have a syntax error), but I could be wrong.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L013.py
+++ b/src/sqlfluff/rules/L013.py
@@ -48,7 +48,8 @@ class Rule_L013(BaseRule):
         elements there are.
 
         """
-        segment = context.functional.segment
+        functional_context = context.functional
+        segment = functional_context.segment
         children = segment.children()
         # If we have an alias its all good
         if children.any(sp.is_type("alias_expression")):
@@ -66,6 +67,16 @@ class Rule_L013(BaseRule):
         ):
             return None
 
+        parent_stack = functional_context.parent_stack
+
+        # Ignore if it is part of a CTE with column names
+        if (
+            parent_stack.last(sp.is_type("common_table_expression"))
+            .children()
+            .any(sp.is_type("cte_column_list"))
+        ):
+            return None
+
         select_clause_children = children.select(sp.not_(sp.is_name("star")))
         is_complex_clause = _recursively_check_is_complex(select_clause_children)
         if not is_complex_clause:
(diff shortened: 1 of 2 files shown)


barrywhart (approved):
Nice work!

I agree that we don't need to check the number of column names in the list.

Issue #3577, opened 2022-07-08
tsql: signs and operators are not distinguished and thus cannot parse simple statements


fmms:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Originally, I was wondering why `- 1` is not formatted as expected (i.e. without the space).

Trying to see numbers in sqlfluff i found a bug.

Consider the following statement:
```
SELECT 1 * - 1
```

I believe #2845 was very related...

### Expected Behaviour

statement can be parsed

### Observed Behaviour

statement cannot be parsed

### How to reproduce

```
PS > cat .\sqlfluff-10.sql
SELECT 1 * - 1

PS > cat .\sqlfluff-10.config
[sqlfluff]
dialect = tsql
templater = raw

PS > sqlfluff lint .\sqlfluff-10.sql --config .\sqlfluff-10.config --verbose
==== sqlfluff ====
sqlfluff:                1.0.0 python:                 3.10.4
implementation:        cpython verbosity:                   1
dialect:                  tsql templater:                 raw
rules:                                all
==== readout ====

=== [ path: .\sqlfluff-10.sql ] ===

== [sqlfluff-10.sql] FAIL
L:   1 | P:  10 |  PRS | Line 1, Position 10: Found unparsable section: '* - 1\n'
WARNING: Parsing errors found and dialect is set to 'tsql'. Have you configured your dialect correctly?
==== summary ====
violations:        1 status:         FAIL
All Finished 📜 🎉!

### Dialect

tsql

### Version

1.1.0

### Configuration

see above

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

tunetheweb:
I believe this is a duplicate of #3544 rather than #2845

fmms:
@tunetheweb unfortunately I am not at all a Python developer so not sure if one fix acutally solves both.
Just allowing `- 1` seems to be one thing. However, all database I tried accept even stuff such as `SELECT 1 * -+- 1` which might require the formula parser to be adjusted a bit more.

tunetheweb:
So there's three issues here:
1. Spaces before the signed are not currently accepted by SQLFluff
[...]

PR #3581, opened 2022-07-10
Allow spaces and multiple signs for numeric literals

### LINKED ISSUE #3544: tsql: RAISEERROR breaks parsing
### Search before asking

### What Happened

I have the following code:

```
CREATE PROCEDURE [UTIL].[createClusteredColumnstoreIndex] AS
BEGIN
    SET NOCOUNT ON;

    RAISERROR (
                'The specified table does not exist. Please enter @tableName in the following format: Schemaname.Tablename OR [Schemaname].[Tablename]'
                ,11
                ,- 1
                );

END
```

I tried to use sqlfluff to format it.

### Expected Behaviour

Code is formatted

### Observed Behaviour

sqlfluff cannot parse the file.

### How to reproduce

```
PS > cat .\sqlfluff-5.config
[sqlfluff]
dialect = tsql
templater = raw
PS > cat .\sqlfluff-5.sql
CREATE FUNCTION [dbo].[getItemList] (
     @list ItemList READONLY
)
RETURNS nvarchar(max)
AS

BEGIN
      DECLARE @str nvarchar(max) = ''

      SELECT @str = @str + [item] FROM (
        SELECT TOP (9999) [item]
        FROM @list
        ORDER BY [order]
      ) i

      RETURN @str
END
PS > sqlfluff parse .\sqlfluff-5.sql --config .\sqlfluff-5.config
```

### Dialect

tsql

### Version

```==== sqlfluff ====
sqlfluff:                1.0.0 python:                 3.10.4
implementation:        cpython verbosity:                   3
dialect:                  tsql templater:                 raw
```

### Configuration

see above

### Code of Conduct

tunetheweb:

Fixes #3544 
Makes progress on #3577 

Don't think so

### Pull Request checklist

  - Other.


test/fixtures/dialects/sparksql/numeric_literal.sql
```diff
@@ -28,3 +28,5 @@ WHERE
     AND baz > +3.E+3
     AND baz > -3.E-3D
     AND baz > -3.e-3d
+    AND baz > -+-1
+    AND baz > -+- 1
```
barrywhart, line 32: I found a closely related issue involving multiple signs for column expressions. Worth fixing this too?

This fails to parse:
```
SELECT - - c1
FROM foo
```

Also this:
```
SELECT --c1
FROM foo
```

barrywhart (approved):
The changes look good. I found a very similar issue involving column expressions rather than literals. If not too tricky, might be worth fixing now as well.

test/fixtures/dialects/sparksql/numeric_literal.sql
tunetheweb: I updated `SignedSegmentGrammar` instead to capture both (plus any future ones).

Note your last example is not valid as that (double negative with no whitespace) becomes a comment.

Issue #3277, opened 2022-05-06
Parser incorrectly swaps BigQuery struct grammars when `parse_grammar` is not supplied.


dmohns:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

When a `STRUCT` is used in conjunction with an `AS` alias, it's columns are recognised as datatypes, which can trigger L063.

### Expected Behaviour

`STRUCT` column names should not trigger L063.

### Observed Behaviour

```
L:   3 | P:   9 | L063 | Datatypes must be upper case.
L:   4 | P:   9 | L063 | Datatypes must be upper case.
```

### How to reproduce

Run `sqlfluff lint` on

```sql
SELECT
    STRUCT(
        some_field,
        some_other_field
    ) AS col,
    STRUCT(
        some_field,
        some_other_field
    )
FROM table
```

### Dialect

bigquery

### Version

0.13.1

### Configuration

```cfg
[sqlfluff]
templater = jinja
dialect = bigquery

[sqlfluff:rules:L063]
# Inconsistent capitalisation of datatypes
extended_capitalisation_policy = upper
```

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

jmc-bbk:
Just doing a bit of investigation on this - although not really familiar with BigQuery or Ansi.

It's not to do with the `alias` segment. You will find the same result with:

```
SELECT
    STRUCT(
        some_field,
        some_other_field
    ),
    STRUCT(
        some_field,
        some_other_field
    )
FROM table
```

Going to try figure it out - but anyone else welcome to jump on the ticket!

barrywhart:
Yes, the parser seems to be behaving inconsistently. For this query, the first 3 `STRUCT` parameters parse as `data_type_identifier`, but the 4th parses as `column_reference`:
```
SELECT
    STRUCT(field1, field2) AS col1,
    STRUCT(field1, field2),
    STRUCT(field1, field2) AS col2,
    STRUCT(field1, field2)
FROM table
```

Here, all 6 parse as `data_type_identifier`:
```
SELECT
    STRUCT(field1, field2),
    STRUCT(field1, field2),
    STRUCT(field1, field2),
[...]

PR #3590, opened 2022-07-11
Fix Bracketed Struct issue

alanmcruickshank:
This addresses #3277.

As @tunetheweb pointed out in the other PR (#3428 - that while immutability helps with the issue, it wasn't the root cause). This PR addresses the root cause.

There are a few routines which check whether a segment has already matched a given set of segments. I.e. if a `SelectStatement` encounters a `SelectStatement` it doesn't need to reparse it - it knows it's already a `SelectStatement`. This allows shortcuts and better performance from not reparsing sections.

The root cause for this is that when checking bracketed sections, all we check is that it's a bracketed section (and not the _type_ of brackets). I.e. a pre-existing bracketed section using `()` would recognise a separate bracketed section using (`<>`). Given the issue we're facing is two segment definitions which only differ in their bracket use (and the content is equivalent) - this solves that problem.

### DIFF
--- a/src/sqlfluff/core/parser/grammar/sequence.py
+++ b/src/sqlfluff/core/parser/grammar/sequence.py
@@ -268,6 +268,11 @@ class Bracketed(Sequence):
 
         # Are we dealing with a pre-existing BracketSegment?
         if seg_buff[0].is_type("bracketed"):
+            # Check it's of the right kind of bracket
+            if not start_bracket.match(seg_buff[0].start_bracket, parse_context):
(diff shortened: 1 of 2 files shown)


alanmcruickshank:
> Or if this intended to replace the other one?

This was intended to more directly fix the other one - but I do still want to make some of the other changes too. Let's get this one done first and then I'll revisit the other PR properly and address the second issue too. I suspect there is a similar issue affecting other segments other than Bracketed too.

PR #3595, opened 2022-07-11
Inherit rather than redefine segments from postgres

alanmcruickshank:
`GroupingSetsClauseSegment`, `CubeRollupClauseSegment` and GroupingExpressionList` are defined in parallel in `postgres`, `sparksql` and `exasol` in an almost identical ways. I think it's safe to assume that the postgres implementation inspired the others.

This sets up exasol and spark to inherit their definition from postgres and remove the duplicated code.

I did a brief scan for other low lying fruit on this front but nothing obvious.

### DIFF
--- a/src/sqlfluff/dialects/dialect_exasol.py
+++ b/src/sqlfluff/dialects/dialect_exasol.py
@@ -43,6 +43,7 @@ from sqlfluff.dialects.dialect_exasol_keywords import (
 from sqlfluff.dialects import dialect_ansi as ansi
 
 ansi_dialect = load_raw_dialect("ansi")
+postgres_dialect = load_raw_dialect("postgres")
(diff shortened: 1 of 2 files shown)


tunetheweb:
Hmmm.... we don't tend to do this, cross dialect imports, that much (with the exception of hive/sparksql and redshift inheriting from postgres).

Is this the way we should go? Especially for only a few statements?

Wouldn't we be better to just define common stuff in ansi dialect and either not reference them in ansi (so they aren't actually inherited), or `Nothing()` them out in the other dialects that don't need them?

barrywhart:
If we can put things in ANSI, that generally seems better to me. I do see some potential for confusion if mostly-unrelated dialects reuse bits from one another. 

If ANSI is not a fit, is it possible (and would it be nicer) to put the reusable bits into one or separate modules that each of the dialects imports from?

Issue #3537, opened 2022-07-01
Automatically skip linting files over a certain size


barrywhart:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

SQLFluff `lint` and `fix` commands should do an up-front file size check, and if it's longer than X, just print a message saying, "we're skipping this file". Code editors sometimes do this, and SQLFluff should, too. This avoids confusing users, who frequently try to lint large files and then report an issue saying SQLFluff is hanging.

We should make the limit variable, perhaps defaulting it to 500 or 1,000 lines.

### Use case

n/a

### Dialect

n/a

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

fmms:
@barrywhart maybe it would make sense to add a timeout instead?

To my understanding it is not really the lines of a file that make sqlfluff slow.

I have CREATE TABLE statements with 1000 rows. Those are linted quickly.
It seems to be the query complexity.

I would be happy to have a --maximum-seconds-per-file parameter that skips a file if the fixing process took longer than e.g. 60 seconds ...

barrywhart:
We could look into it, but timeouts are not gracefully supported for in-process code, only for network calls. I have in one case used some ugly code that simulated pressing Control-C after a certain amount of time. 🤢😳

fmms:
As there is the —processes option I assumed there is some control thread kicking of the parallelism that could abort the process on one file. 
i just believe most/all people just are afraid of too long execution when either formatting lots of files or in ci/cd scenarios ..

alanmcruickshank:
> @barrywhart maybe it would make sense to add a timeout instead?

I agree timeout would be better, but only viable in the case that someone is running with the parallel (`-p`) option. We could *also* add that option too for users who wanted more control?

alanmcruickshank:
I'm happy to pick this one up - I might actually aim for a _character_ limit rather than a _line_ limit because that's a) slightly easier and b) more representative of complexity.

PR #3600, opened 2022-07-12
Add large file check

alanmcruickshank:
This addresses #3537. I've gone for a character limit rather than a line limit because it's a) easier and b) probably more indicative of parsing complexity.

Setting the detail at 20k chars puts it around the size I'd expect from a 1k line file, and comfortably above most of the manually written sql in the tails.com repo.

I've implemented as a decorator to make it as reusable as possible, but if we define new templaters it will need to be imported and added directly rather than bundled "by default" in a shim. Doing the latter might be tidier but we don't otherwise currently have a shim around `.process()` with each templater defining their own and that being called directly - so it felt like too much additional change to add it for just this.

### DIFF
--- a/src/sqlfluff/core/templaters/base.py
+++ b/src/sqlfluff/core/templaters/base.py
@@ -3,6 +3,7 @@
 import logging
 from bisect import bisect_left
 from typing import Dict, Iterator, List, Tuple, Optional, NamedTuple, Iterable
+from sqlfluff.core.config import FluffConfig
 
 from sqlfluff.core.errors import SQLTemplaterSkipFile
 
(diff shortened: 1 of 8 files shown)


src/sqlfluff/core/templaters/base.py
```diff
@@ -22,6 +23,32 @@ def iter_indices_of_newlines(raw_str: str) -> Iterator[int]:
...
+    """Raise an exception if the file is over a defined size.
+
+    Designed to be implemented as a decorator on `.process()` methods.
+
+    If no config is provided or the relevant config value is set
+    to zero then the check is skipped.
+    """
+
+    def _wrapped(
+        self, *, in_str: str, fname: str, config: FluffConfig = None, **kwargs
```
barrywhart, line 36: Since the parameter list is closely tied to the templaters' `process()` function, what would you think of moving the check to the templater base class function, `RawTemplater.process()`? We could have each templater's `process()` function call the base class `process()` function before doing any work.

That seems a bit cleaner to me (because decorators are usually somewhat generic in terms of the functions they can wrap).

alanmcruickshank: That approach is definitely cleaner. It feels like a big change - but maybe you're right and I should just bit the bullet on that. It probably triggers a bigger version change.

barrywhart: I don't think it'll be bad -- just adding a call to `super().process()` in the subclasses. Dbt could be a bit different, if it inherits from Jinja templater.

Let me know if I can help with tactical advice or whatever!

barrywhart: Another option: add a `large_file_check()` function to the base class, then each process() function calls it. A bit simpler than calling the base class process() function.

alanmcruickshank: Actually on second thoughts I'm not sure. Here are the options I see:

1. Decorator - as per what I've done.
2. Simple function call (probably a method of the base class like `.do_large_file_check()`) which each `.process()` method calls before doing anything else. I did this first and it felt messy. Defined once yes, but feels a little fragile. Simpler than option 1 though.
3. Put it in the `RawTemplater.process()` method and then call `.super()` in other process methods. This gets messy because the dbt templater inherits from then jinja templater, which inherits from the python templater, which inherits from the RawTemplater. We wouldn't just want to call `super()` there - we'd be referencing `super()` in some cases and `RawTemplater` in others. This just feels like it just ends up as a more complicated version of option 2.
4. Have two methods on templaters one which calls the other. The outer is only defined on the `RawTemplater`, and does the large file checks, then calls the inner. The inner is the one defined on all the other templaters.
   i. Keeping the inner method as `.process()` and calling the outer something else changes the API which the templaters expose to the linter and other internal tooling.
   ii. Keeping the outer method as `.process()` which then calls a renamed inner method means none of the outer APIs change but all the templaters need updating.

I'm not sure any of these options is meaningfully cleaner than another. @barrywhart - views?

alanmcruickshank: The simplest option actually feels like option 1 🤷 . Dumb, but simple and easy.

barrywhart: Yep, let's go with it. I pride myself on being an "it depends" kind of person. Thanks for thinking it through!

barrywhart:
The failing tests were caused by some sort of network error. I triggered a rerun...

barrywhart (approved):

Issue #3587, opened 2022-07-11
dbt.exceptions.InternalException: Asked to compile <class 'dbt.contracts.graph.compiled.CompiledModelNode'> node, but it has no compiled form


reubster:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

While linting any dbt model the error is raised part-way through linting (apparently after all the fail lines have been produced and displayed). 

I've tried sqlfluff versions 

0.10.1 - no error 
0.11.0, 0.13.2, 1.0.0, 1.1.0 - error thrown

### Expected Behaviour

sqlfluff reports linting failures. 

### Observed Behaviour

SQL model materialising as a view. This is really the WEB_USERS model from the fivetran segment dbt package, just renamed to use our naming standard. 

```
with

SOURCE as (
  select
    *
  from
    {{ ref('SG__WEB_USERS__UNION') }}
),

RENAMED as (
  select
    _DBT_SOURCE_RELATION,
    EMAIL,

    {{ sv_utils.get_email_domain('EMAIL') }} as EMAIL_DOMAIN,

    {{ sv_utils.is_valid_email('EMAIL') }} as IS_EMAIL_VALID,

    ID as USER_ID,
    PERSON_ID,
    USERNAME,
    EMAIL_ADDRESS,
    {{ sv_utils.get_email_domain('EMAIL_ADDRESS') }} as EMAIL_ADDRESS_DOMAIN,
    {{ sv_utils.is_valid_email('EMAIL_ADDRESS') }} as IS_EMAIL_ADDRESS_VALID,
    "NAME" as FULL_NAME,
    SHORT_NAME,
    FIRST_NAME,
    INITIALS,
    SURNAME,
    LAST_NAME,
    USER_TYPE,

    CONTEXT_IP as IP,
    -- CONTEXT_LIBRARY_NAME, -- not useful
    SUBSCRIPTIONS,
    CONTEXT_PAGE_SEARCH as PAGE_URL_QUERY,
    CONTEXT_PAGE_URL as PAGE_URL,
    {{ dbt_utils.get_url_host('CONTEXT_PAGE_URL') }} as PAGE_URL_HOST,
    -- CONTEXT_LIBRARY_VERSION, -- not useful
    CONTEXT_PAGE_REFERRER as PAGE_REFERRER,

    replace(
[...]

PR #3603, opened 2022-07-12
Add a skip and better logging for fatal dbt issues

alanmcruickshank:
This is a draft PR to assist with debugging #3587.

We *may* actually want to consider this a permanent fixture (or a variation of it), to allow dbt projects to a) better understand failing files and b) still lint the rest of the project.

In short, if a `dbt_compiler.compile_node()` raises any exception - we skip the file and output to the user a stringified version of the dbt error _and a reference to the file which caused it_ for later debugging.

### DIFF
--- a/plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
+++ b/plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
@@ -471,7 +471,7 @@ class DbtTemplater(JinjaTemplater):
                     "Please report this error on github at "
                     "https://github.com/sqlfluff/sqlfluff/issues, including "
                     "both the raw and compiled sql for the model affected.",
-                    fname
+                    fname,
                 )
                 # Additional error logging in case we get a fatal dbt error.
                 raise SQLTemplaterSkipFile(  # pragma: no cover
--- a/plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
+++ b/plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
@@ -464,6 +464,12 @@ class DbtTemplater(JinjaTemplater):
                     node=node,
                     manifest=self.dbt_manifest,
                 )
+            except Exception as err:
+                # Additional error logging in case we get a fatal dbt error.
+                raise SQLTemplaterSkipFile(  # pragma: no cover
+                    f"Skipped file {fname} because dbt raised a fatal "
+                    f"exception during compilation: {err!s}"
+                )
             finally:
                 # Undo the monkeypatch.
                 Environment.from_string = old_from_string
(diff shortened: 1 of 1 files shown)


plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -464,6 +464,12 @@ def make_template(in_str):
                     node=node,
                     manifest=self.dbt_manifest,
                 )
+            except Exception as err:
+                # Additional error logging in case we get a fatal dbt error.
+                raise SQLTemplaterSkipFile(  # pragma: no cover
```
barrywhart, line 469: What does the logging look like from this? Curious if we should do one or both of:

Logging where the exception was raised from:
```
logger.exception("<<MESSAGE>>")
```

Exception chaining:
```
raise SQLTemplaterSkipFile(...) from err
```

alanmcruickshank: Both are things I've never used before 🤔 .

On reading - I think we should definitely do the latter. I think it will still result in a skipped file. Doing the former seems like a good idea too. Perhaps with some more information attached.

Great ideas!

barrywhart: I'm a bit of a Python 🤓, I suppose. 😝

alanmcruickshank: Take a look at what I've done now. Used both 🤓

barrywhart (approved):

alanmcruickshank:
@barrywhart - do you reckon this is a good idea to merge to main, or should I keep this is a debugging branch for now?

On one side, this makes the fatal compilation errors much more manageable for dbt users, and allows some files to be linted (hopefully) - but it also reduces the visibility of those bugs and unless people *see* the errors, they might no know they're there.

I think I'm leaning toward merge, but could be persuaded otherways.

Views?

@NiallRees - do you have views as a regular dbt user?

NiallRees:
> @barrywhart - do you reckon this is a good idea to merge to main, or should I keep this is a debugging branch for now?
> 
> On one side, this makes the fatal compilation errors much more manageable for dbt users, and allows some files to be linted (hopefully) - but it also reduces the visibility of those bugs and unless people _see_ the errors, they might no know they're there.
> 
> I think I'm leaning toward merge, but could be persuaded otherways.
> 
> Views?
> 
> @NiallRees - do you have views as a regular dbt user?

I think it makes sense to merge given that they were failing anyway, I've not encountered scenarios where the first time I find a model won't compile is from SQLFluff though!

alanmcruickshank:
Given what @NiallRees says, I'm going to merge this later today unless anyone else has strong views.

alanmcruickshank:
> One more possibility -- change the logging output to display in red and/or add a similar `print()` that displays in red. This makes it less likely to be overlooked.

The text from the templater skip error will already get printed in red 😄 👍

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
noel: Would this resolve a problem I have seen where sqlfluff appears to have worked when in fact dbt didnt compile?

image

noel:
would this resolve the issue where it appears everything is good when dbt compile actually failed?

image

PR #3619, opened 2022-07-15
BUGFIX: dbt templater now needs 1.2.0 or later

alanmcruickshank:
newest dbt templater also needs newest sqlfluff. This should probably go out as a hotfix.


tunetheweb:
I'd rather to #3613 and keep it in sync (which includes this change).

WDYT?

alanmcruickshank:
Agreed

PR #3631, opened 2022-07-19
Rationalise Rule Imports

alanmcruickshank:
This moves the generation of rule docs to the module that those rules live in. Having done that it make `sqlfluff.core.rules` behave more like other modules in exposing some of the basics (more specifically the objects from `rules.base`) directly from the module itself.

This fits into the roadmap for reflow because I'm going to be leveraging some of the rule mechanics to do reflow within our existing frameworks.

I think for now, keeping `rules.doc_decorators` and `rules.functional` as separate imports still make sense. They're quite discrete units.


barrywhart (commented):
Seems like a good plan!

I'm noticing several of the CI checks are failing. Can you get those working, and I'll take another look?

barrywhart (approved):
Looks good! Did you manually build the docs to verify `rules.html` is being built correctly?

PR #3641, opened 2022-07-20
Snowflake - `ALTER TABLE IF EXISTS` and `WHEN SYSTEM$STREAM_HAS_DATA()`

chrisalexeev:

Added snowflake dialect support for the optional `ALTER TABLE IF EXISTS...` syntax as well as tasks that use the WHEN SYSTEM$STREAM_HAS_DATA() conditional.

No.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -347,6 +347,12 @@
         name="integer_literal",
         type="literal",
     ),
+    SystemFunctionName=RegexParser(
+        r"SYSTEM\$([A-z0-9]*)",
```
tunetheweb, line 351: This looks like a typo, but I don't think it is because it does match `_` as well.

Should this be more explicit:

```suggestion
        r"SYSTEM\$([A-Za-z0-9_]*)",
```

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_snowflake.py
chrisalexeev: Good note, implemented.

tunetheweb (approved):

PR #3638, opened 2022-07-20
Improve handling of BigQuery hyphenated table names

barrywhart:
### Brief summary of the change made
Makes progress on #3609 

@greg-finley: This is based on the PR you started and closed (#3614). It was a good plan, but it required some improvements to the `iter_raw_references()` function in the BigQuery dialect. 🤞 that this is working now.

### Are there any other side effects of this change that we should be aware of?
The `iter_raw_references()` function is a vital part of the `get_select_statement_info()` function, used by almost every rule that cares about aliases (e.g. L026). I was pretty careful in writing and testing the new implementation, but the code is inherently tricky, now that it needs to handle both simple (non-hyphenated) as well as hyphenated references.

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #3645, opened 2022-07-20
L057: In BigQuery, allow hyphens by default

barrywhart:
Fixes #3638

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -174,7 +174,7 @@ group_by_and_order_by_style = consistent
 unquoted_identifiers_policy = all
 quoted_identifiers_policy = all
 allow_space_in_identifier = False
-additional_allowed_characters = ""
+additional_allowed_characters = None
 ignore_words = None
 ignore_words_regex = None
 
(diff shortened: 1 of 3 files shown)


src/sqlfluff/core/default_config.cfg
```diff
@@ -174,7 +174,7 @@ group_by_and_order_by_style = consistent
 unquoted_identifiers_policy = all
 quoted_identifiers_policy = all
 allow_space_in_identifier = False
-additional_allowed_characters = ""
+additional_allowed_characters = None
```
barrywhart, line 177: Empty quotes don't return an empty string. They literally return the string `""`, which I believe is not what we want -- i.e., we don't want to allow `"` in names.

tunetheweb (approved):
Wonder if we should have a test without quotes?

barrywhart:
I added another test case, `test_pass_hyphen_bigquery_no_quotes`, as you suggested. This was already "working" by accident, as we discussed on the related issue, so it's not related to the changes in this PR, but I agree it seems like a good test to prevent future breakage.

Issue #3651, opened 2022-07-22
Reshift dialect doesn't understand CONVERT function


pdebelak:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

The redshift dialect thinks the data type identifier in the `CONVERT` function (see [CAST and CONVERT docs](https://docs.aws.amazon.com/redshift/latest/dg/r_CAST_function.html)) is a column identifier which can cause spurious lint errors.

### Expected Behaviour

The `CONVERT` function is correctly parsed with a data type identifier and a column identifier.

### Observed Behaviour

The `CONVERT` function is parsed as having two column identifier arguments.

### How to reproduce

You can see the `L027` rule failure is due to `integer` being parsed as a column identifier, not a data type identifier.

```
$ echo 'SELECT
    sellers.name,
    CONVERT(integer, sales.pricepaid) AS price
FROM sales
LEFT JOIN sellers ON sales.sellerid = sellers.sellerid
WHERE sales.salesid = 100;' | sqlfluff lint --dialect redshift -
== [stdin] FAIL                                                                                                                            
L:   3 | P:  13 | L027 | Unqualified reference 'integer' found in select with
                       | more than one referenced table/view.
All Finished 📜 🎉!
```

### Dialect

Redshift

### Version

```
sqlfluff, version 1.1.0
Python 3.9.13
```

### Configuration

None

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

pdebelak:
Fixed in #3653

PR #3653, opened 2022-07-22
Add handling for redshift convert function data type argument

pdebelak:

Fixes issue #3651.

Previously, the first argument to the `convert` redshift function was interpreted as a column reference rather than a data type. This could cause rule L027 to fail since the data type wasn't qualified with a table name even though that doesn't make sense. This adds explicit handling for `convert` to eliminate that issue.

Not that I'm aware of.

### Pull Request checklist

  - Other.


tunetheweb (approved):
LGTM!

Btw ideally don’t force push if you can. Was fine for this one, but if a PR review takes a few iterations the force pushing can make it difficult to see what’s changed in each one.

PR #3658, opened 2022-07-24
Reposition before recursion in fixes to avoid internal error

alanmcruickshank:
This resolves #3656 .

The issue was caused by child segments trying to apply fixes to segments inserted or edited by a parent which hadn't yet assigned position markers using `_position_segments`. This simply adds an intermediate step _after_ applying fixes to the parent where we optionally reposition `self` before applying fixes to children.

This will add a very slight performance overhead, but I don't think the `fix` routine is currently the bottleneck. Potentially in future we might be able to more selectively run this routine, but for now this feels ok. Added a test case too.

PR #3661, opened 2022-07-24
Use UUIDs for matching

alanmcruickshank:
This partly sets some of the groundwork for more immutable segments in future, but really I think this solves the test fails I'm getting on #3658.

Replace `id()` matching with explicit `uuid` so that we can control where to match and where not.

### DIFF
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -25,6 +25,7 @@ from typing import (
     TYPE_CHECKING,
 )
 import logging
+from uuid import UUID, uuid4
 
 from tqdm import tqdm
 
@@ -237,6 +238,7 @@ class BaseSegment:
         segments,
         pos_marker: Optional[PositionMarker] = None,
         name: Optional[str] = None,
+        uuid: Optional[UUID] = None,
     ):
         # A cache variable for expandable
         self._is_expandable: Optional[bool] = None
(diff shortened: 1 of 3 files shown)


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -237,6 +238,7 @@ def __init__(
         segments,
         pos_marker: Optional[PositionMarker] = None,
         name: Optional[str] = None,
+        uuid: Optional[UUID] = None,
```
barrywhart, line 241: I'm curious: Under what circumstances would a caller provide this parameter?

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -56,6 +58,8 @@ def __init__(
         self._is_expandable = None
         # Keep track of any source fixes
         self._source_fixes = source_fixes
+        # UUID for matching
```
barrywhart, line 61: Nitpick: Instead of this, should it call the class (`BaseSegment`) constructor instead?

barrywhart (approved):
Looks good. A couple of questions/suggestions.

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: It's effectively called during a `.copy()` operation - in theory you'd call it if you were ever making a new segment which you wanted to represent (or at least act as an anchor for) an existing segment. Going deeper into immutability might require that a little more.

src/sqlfluff/core/parser/segments/raw.py
alanmcruickshank: Yeah I thought about that. On balance I thought so much is different about the two constructors that inheriting seemed the more complex route.

PR #3679, opened 2022-07-26
Scan both issue title and body for dialect labels

greg-finley:

Sorry for the barrage of small PRs, but sounds like it's better to scan the issue body for dialect mentions also.

We would occasionally add labels if a dialect that isn't truly relevant is mentioned in passing.

### Pull Request checklist

  - Other.


greg-finley:
We can do this if we want, but the latest conversation makes it sound like this would hit a lot of false positives.

PR #3686, opened 2022-07-28
T-SQL: Properly parse collation names

borchero:

The main branch of SQLFluff currently parses collation names appearing in `COLLATE` expressions as column identifiers. For example, the following SQL ...

```
SELECT table1.col COLLATE Latin1_General_CI_AS
FROM table1;
```

... generates the following errors with dialect `tsql`:

```
L:   1 | P:  27 | L014 | Unquoted identifiers must be consistently lower case.                                                                      
L:   1 | P:  27 | L028 | Unqualified reference 'Latin1_General_CI_AS' found in
                       | single table select.
L:   1 | P:  27 | L028 | Unqualified reference 'Latin1_General_CI_AS' found in
                       | single table select which is inconsistent with previous
                       | references.
```

Instead, we want to parse the collation as dedicated identifier `"collation"`. This PR causes the above SQL query to be linted without errors.

Probably not, but I did modify `Expression_A_Grammar`. Should be checked carefully.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -472,6 +472,7 @@ ansi_dialect.add(
     # non-standard keywords)
     IsNullGrammar=Nothing(),
     NotNullGrammar=Nothing(),
+    CollateGrammar=Nothing(),
     FromClauseTerminatorGrammar=OneOf(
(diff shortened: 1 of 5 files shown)


greg-finley (approved):
Nice solution.

PR #3697, opened 2022-07-30
L042: Autofix sometimes results in "fix looping", hitting the linter "loop limit"

### LINKED ISSUE #3572: L042 should not remove queries to CTEs when the queries reference other tables in outer query.
### Search before asking

### Description

The rule L042 is a good idea I think but it does not support to config more advanced options where you - on purpose - might want to use a sub-query in a left join by using the language specific functions LATERAL (in PostgreSQL) or OUTER APPLY (in TSQL).

The Anti-pattern
``` sql
select
    a.x, a.y, b.z
from a
join (
    select x, z from b
) using(x)
```
is good, but the following pattern might be wanted, as the subquery is directly joining to the outer query:

### PostgreSQL
``` sql
select
    a.x, a.y, b.z
from a
join lateral (
    select x, z
    from b
    where b.x = a.x
    order by z
    limit 1
) on true
```

### TSQL
``` sql
select
    a.x, a.y, b.z
from a
outer apply (
    select top 1 x, y
    from b
    where b.x = a.x
    order by z
) as b
```

~I suggest to add a new config option `join-lateral` to the existing options `join` and `from`.~

Moving these queries to CTEs without refactoring the SQL will break the SQL.
I would suggest automatically excluding these subqueries from L042

### Use case

_No response_

### Dialect

see above

### Code of Conduct

barrywhart:
Fixes #3572
Makes progress on #3598

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/L042.py
+++ b/src/sqlfluff/rules/L042.py
@@ -3,7 +3,6 @@ import copy
 from functools import partial
 from typing import Generator, List, NamedTuple, Optional, Set, Type, TypeVar, cast
 from sqlfluff.core.dialects.base import Dialect
-from sqlfluff.core.parser.markers import PositionMarker
 from sqlfluff.core.parser.segments.base import BaseSegment
 from sqlfluff.core.parser.segments.raw import (
     CodeSegment,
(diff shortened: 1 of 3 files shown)


src/sqlfluff/rules/L042.py
```diff
@@ -218,8 +217,67 @@ def _calculate_fixes(
...
+    if not nested_select:
+        return False
+    select_statement = _get_first_select_statement_descendant(nested_select[0])
+    if not select_statement:
+        return False
+    nested_select_info = get_select_statement_info(select_statement, dialect)
+    if nested_select_info:
+        for r in nested_select_info.reference_buffer:
+            for tr in r.extract_possible_references(  # type: ignore
+                level=r.ObjectReferenceLevel.TABLE  # type: ignore
```
barrywhart, line 269: Used `type: ignore` because Python doesn't know `r` is a `ObjectReferenceSegment` or related type.

barrywhart:
This PR is good for #3572, but I'm going to try and address issue #3598 in the same PR, if it's not too difficult.

tunetheweb (approved):

PR #3701, opened 2022-08-01
Name Deprecation (Part 1)

### LINKED ISSUE #3692: Enhancement: Deprecate BaseSegment.name in favour of extending types.
### Search before asking

### Description

This is potentially a controversial one, but I think it makes sense.

The `name` keyword on segments I think is confusing. Sometimes it's used like a type, sometimes it means something else. We have rules that sometimes search by `name` and sometimes by `type`. I think however, that **name is always used as a more specific type**.

The type interface for segments is already much more flexible that when originally designed. `BaseSegment.is_type()` already checks against both the declared type and the types of any parent classes, so we already have the concept of more than one valid `type` for a segment. This presents a dilemma in delivering #3668, because it's not clear how to crawl easily for types **and** for names.

The place where `name` becomes most obvious is in the parse tree where the `name` defines what is shown there.

My proposal is:
- Remove `name` as an attribute.
- Allow (optionally) the `type` of a segment to be given as a tuple of strings rather than just a string.
- When type checking we check against all types declared.
- In all the cases where we need a single `type` for a segment, we get the **last** in the tuple of types.

As an example, we currently have:
```python
PlusSegment=StringParser("+", SymbolSegment, name="plus", type="binary_operator")
```
This would become:
```python
PlusSegment=StringParser("+", SymbolSegment, type=("binary_operator", "plus"))
```

The _functionality_ would remain roughly the same, but rather than checking the `name` or checking `type`, `PlusSegment` would return true for `.is_type("plus")` **and** `istype("binary_operator")`.

In several cases I think this would also simplify code where we're setting the `type` and the `name` to be the same thing at the moment.

### Use case

This simplifies the code and removes some bloat.

### Dialect

all

### Code of Conduct

alanmcruickshank:
Working on #3692 has a lot of knock-on effects. This is the first meaningful milestone of that. It's already a _massive_ PR, so in the interests of incremental change I feel a little uncomfortable going much further without merging some things.

There will be further development on #3692 to strip it out more, but this PR makes progress on that by:
- Adding more extensive types and inheritance of some existing types, so that names 
- Removing all use of `.name` in rules (in refactoring the rules and dialects so that `name` is never read _or set_ after the initial parsing).
- The above changes trigger a lot of the yaml files to change because we had been using `name` as a more specific `type` in some cases (e.g. `naked_identifier` as a name just had the type `identifier`, and so that appeared in the yaml tree. Now this just has the main type `naked_identifier` but inherits from the `identifier` type and so `.is_type('identifier')` would still return `True`).

The last part is why it's such a big PR.

This PR _does not_ address some of the remaining issues:
- `NamedParser` still exists which I'd like to remove.
- The link between `...Parser` classes, lexing and segments is really convoluted and relies somewhat on names.
- `BaseSegment.name` still exists as an attribute and is still used in a few places.


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -199,85 +200,80 @@
...
     StartSquareBracketSegment=StringParser(
-        "[", SymbolSegment, name="start_square_bracket", type="start_square_bracket"
-    ),
-    EndSquareBracketSegment=StringParser(
-        "]", SymbolSegment, name="end_square_bracket", type="end_square_bracket"
+        "[", SymbolSegment, type="start_square_bracket"
     ),
+    EndSquareBracketSegment=StringParser("]", SymbolSegment, type="end_square_bracket"),
     StartCurlyBracketSegment=StringParser(
-        "{", SymbolSegment, name="start_curly_bracket", type="start_curly_bracket"
```
barrywhart, line 223: For the changes in this section, it looks like we're losing whatever `name` we were setting previously. Is that a concern?

src/sqlfluff/rules/L006.py
```diff
@@ -63,8 +63,9 @@ def _missing_whitespace(seg, before=True):
         if seg.is_whitespace:
             return False
         # And it's not an opening/closing bracket
-        if seg.name.endswith("_bracket"):
-            if seg.name.startswith("start_" if before else "end_"):
+        seg_type = seg.get_type()
```
barrywhart, line 66: Is there a reason for requiring a call to `.get_type()` rather than exposing the type directly as a property? (Note that Python "properties" can actually be set up to call a function.)

https://www.programiz.com/python-programming/property

If the intent is to remind people that types are a bit complex (e.g. that sometimes you need to use `.is_type(blah)` rather than `.type == blah`), that totally makes sense. 👍

src/sqlfluff/rules/L011.py
```diff
@@ -95,17 +95,17 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
-                    if context.raw_segment_pre.name not in ["whitespace", "newline"]:
+                    if not context.raw_segment_pre.is_type("whitespace", "newline"):
                         insert_buff.append(WhitespaceSegment())
 
                     # Add an AS (Uppercase for now, but could be corrected later)
                     insert_buff.append(KeywordSegment("AS"))
 
                     # Add a trailing whitespace if we need to
-                    if context.segment.segments[0].name not in [
+                    if not context.segment.segments[0].is_type(
```
barrywhart, line 105: Now that we're using types, I'm curious is `.is_type("whitespace")` would also work here. I.e., does `newline` type inherit from `whitespace` type?

src/sqlfluff/rules/L052.py
```diff
@@ -179,7 +179,7 @@ def _handle_semicolon_same_line(
             info.anchor_segment,
             info.whitespace_deletions,
             [
-                SymbolSegment(raw=";", type="symbol", name="semicolon"),
+                SymbolSegment(raw=";", type="statement_terminator"),
```
barrywhart, line 182: Not in scope for this PR, but it seems like there should be a way to create this segment using the type alone (not requiring the hardcoded `";"`).

src/sqlfluff/rules/L053.py
```diff
@@ -71,7 +71,7 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
             *[
                 segment
                 for segment in context.segment.segments
-                if segment.name not in bracket_set and not segment.is_meta
+                if segment.get_type() not in bracket_set and not segment.is_meta
```
barrywhart, line 74: Have we (or could we) document somewhere the difference between `.is_type(a, b, c)` versus `.get_type() in {a, b, c}`. I assume it's pretty subtle but also pretty important. (And I'm not sure I know the difference myself.)

src/sqlfluff/rules/L006.py
alanmcruickshank: Yeah I'm thinking the latter. We're moving to a world where a segment doesn't just have _one_ type - it has a specific one _and_ and all of the types of it's parent segments.

If we just check `.type` then we only get the last one (and in the case of `KeywordSegment`, not even that) - so by consistently called `.get_type()` we always get the effect we want.

src/sqlfluff/rules/L011.py
alanmcruickshank: It's an excellent point - however for some reason I found this note in the `NewlineSegment` class:

```
    NOTE: NewlineSegment does not inherit from WhitespaceSegment.
    Therefore NewlineSegment.is_type('whitespace') returns False.

    This is intentional and convenient for rules. If users want
    to match on both, call .is_type('whitespace', 'newline')
```

Maybe I should revisit that - but for now this seems intentional.

src/sqlfluff/rules/L052.py
alanmcruickshank: yeah I agree - it's a bit messy. For now I think it's also out of scope but I'll keen an eye on it.

src/sqlfluff/rules/L053.py
alanmcruickshank: - `.get_type() in {a, b, c}` only checks the last type in the stack.
- `.is_type(a, b, c)` checks against all parent types.

src/sqlfluff/dialects/dialect_ansi.py
alanmcruickshank: I was thinking about this. Where the answer is _yes_ then I'll create another parent segment (like what I've done with identifiers) - where the answer is _no_ then I'll just leave it.

src/sqlfluff/rules/L049.py
```diff
@@ -65,29 +65,34 @@ def _eval(self, context: RuleContext) -> Optional[List[LintResult]]:
...
+                continue
             null_literal = after_op_list.first(sp.is_code())
             # if the next bit of code isnt a NULL then we are good
-            if not null_literal.all(sp.is_name("null_literal")):
+            if not null_literal.all(sp.is_type("null_literal")):
                 continue
 
             sub_seg = null_literal.get()
-            assert sub_seg, "TypeGaurd: Segement Must exist Must exist"
+            assert sub_seg, "TypeGaurd: Segment must exist"
```
barrywhart, line 87:
```suggestion
            assert sub_seg, "TypeGuard: Segment must exist"
```

barrywhart (approved):

barrywhart:
YAML parser checks are failing. PR looks good to me, though!

PR #3703, opened 2022-08-01
Attempt to handle logging issues at teardown

### LINKED ISSUE #3702: CI Bug: Logging is very very verbose due to pytest issue
### Search before asking

### What Happened

Logging is full of logging errors in test suite. See example: https://github.com/sqlfluff/sqlfluff/runs/7611283306?check_suite_focus=true

Appears to be due to pytest issue: https://github.com/pytest-dev/pytest/issues/5502

[Suggested fix](https://github.com/pytest-dev/pytest/issues/5502#issuecomment-1190557648) is to add a teardown handler:
```
def clear_loggers():
    """Remove handlers from all loggers"""
    import logging
    loggers = [logging.getLogger()] + list(logging.Logger.manager.loggerDict.values())
    for logger in loggers:
        if not hasattr(logger, "handlers"):
            continue
        for handler in logger.handlers[:]:
            logger.removeHandler(handler)
```

### Expected Behaviour

More sensible logs :)

### Observed Behaviour

Too many logs

### How to reproduce

fails in python 3.7 test suite at least

### Dialect

N/A

### Version

master

### Configuration

default CI setup.

### Code of Conduct

alanmcruickshank:
Resolves #3702.

I'm testing the effects of this on the draft name deprecation PR.

### DIFF
--- a/test/cli/commands_test.py
+++ b/test/cli/commands_test.py
@@ -11,6 +11,7 @@ import subprocess
 import sys
 import tempfile
 import textwrap
+import logging
 from unittest.mock import MagicMock, patch
 
 import chardet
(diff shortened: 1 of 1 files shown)


alanmcruickshank:
Confirm effective on #3701 . Logging much more sensible.

Issue was due to adding logging handlers in the CLI module which aren't otherwise properly cleaned up.

test/cli/commands_test.py
```diff
@@ -29,6 +30,31 @@
 re_ansi_escape = re.compile(r"\x1b[^m]*m")
 
 
+@pytest.fixture()
```
barrywhart, line 33:
```suggestion
@pytest.fixture(autouse=True)
```

I think if you make this change, it'll automatically apply this fixture to every test in this `.py` file without you having to explicitly add it to each test. (This can be handy because it means future tests added to this file will automatically get the same behavior.)

https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request

barrywhart (approved):
Looks good! One possible suggestion for how to simplify/futureproof this a bit.

test/cli/commands_test.py
alanmcruickshank: I tried that and it didn't work 🤔 . I'll have another go - but if not then I'll just merge as is.

barrywhart: No worries. We can always do it later. I've successfully used `autouse` fixtures a few times, so there may be something subtle going wrong. They are definitely a bit magical, which is both good and bad. 😬

PR #3712, opened 2022-08-05
Align raw_comparison_operator and comparison_operator

alanmcruickshank:
Found this in the process of working on #3701 and #3692.

Touches lots of files but relatively simple change.

I don't think we need separate `comparison_operator` and `raw_comparison_operator` types - this gets rid of the latter and makes them all the same.

@tunetheweb do you know if there was a reason for this initially?


alanmcruickshank:
Hold up - I think I understand more now. I don't think this is ready to merge - will think again.

alanmcruickshank:
I get it - I'm going to close this. I don't think it's a good idea anymore.

tunetheweb:
Sorry was away last week. Know you've closed it, but if you still remember can you explain why we need both? Don't know myself and curious now...

Issue #3673, opened 2022-07-26
Enhancement: Whitespace reflow utils master issue


alanmcruickshank:
### Description

There are too many other issues related to this and I think it makes sense to bring them all together, and track progress in one place.

This encompasses:
- `black`-like deterministic cli utility: #2436 (where a lot of discussion has gone into how we might achieve this).
- Smarter whitespace and line break treatment: #1839 #2466 #365
- Centralising and rationalising reflow functions: #1230 #946 
- Better indentation: #2867 #3705 #3865 #3691 #3978
- Indentation bugs in templated loops where solutions oscillate due to an if statement which behaves differently on each loop: #2677 #3185

The vision for success is:
- All reflow code works within the existing frameworks for safely editing templated files by utilising the existing `LintFix` functionality (we may isolate this more in the process but that's a bonus).
- All rules have central `reflow` utilities they can call on. These return `LintFix` objects (as above), which can then be returned by the rule as it's own suggestions.
- Reflow routines are recursive such that they can be applied to whole files or sections of files.
- There is a cli command `reflow` which wraps a version of this code and applies it to the whole file.

There are a few stages of delivering this:

- [x] Build out a vision of how to do this (see #3824)
- [x] Implement _spacing_ framework (see #3824)
- [x] Implement depth map to support indentation (see #3824)
- [x] Implement indentation within `ReflowSequence` (i.e. `L003`, `L002` & `L004`). #3942 
- [x] Implement line breaks (i.e. `L016`, potentially also `L036`). #4067
- [ ] Handle edge case: `L052` as a special case (probably using the `rebreak` logic) 
- [x] Handle edge case: `L007` & `L019` as special cases: #3847
- [x] Handle edge case: `L065` as special case (using `alone` probably). #3847 
- [x] Handle missed case: `L008` (seems very possible with current capability). #3908
- [ ] Handle missed case: `L009` (unless handled with `L052`).
- [ ] Handle keyword based (i.e. not _type_ based) line breaks: `L041`.
- [ ] Handle compulsory empty line rules: `L022`
- [ ] Bring `L050` into the `respace` logic (start of file).
- [ ] Build test harness (and potentially API method concurrently) to apply reflow code in a _rule-like_ harness.
[...]

PR #3717, opened 2022-08-07
Build out rule crawling mechanisms

alanmcruickshank:
This builds on the work started with `CrawlBehavior` in the base rules file. I've extended that into a new module with a few variants on the idea and some more flexibility for specific rules. More specifically the newest logic is in `SegmentSeekerCrawler` which more aggressively prunes the tree if segments of particular types do not exist anywhere within a segment or it's children. On the less clever side, I've also introduced `RootOnlyCrawler` which will be relevant for rules which want to handle their own crawling entirely (which is what I'm thinking with reflow application, see #3673). To facilitate this, `RuleContext` has now also been moved (mostly untouched) to it's own file.

To achieve this I've done a few other things:
- `SegmentMetaclass` to allow caching of `_class_types`.
- Building out some more mechanics on `BaseSegment` for accessing types.
- A `repr()` method for `LintResult` and several places with additional logging to help debugging.
- Strip out the old `CrawlBehavior` class.
- Removed `.is_final_segment()` which is now unused (I thought about this one where I could just put a `pragma: no cover` on it - but we could always recreate this method if needed later from the git history).
- `context.raw_segment_pre` has been removed, partly because it's hard to reimplement in this model but also because the new, more efficient, crawling is already a lot more efficient.
- Refactoring of _all_ rules to work better with the new structure. Significant refactorings on:
  - `L016`
  - `L039`
  - `L052`
  - `L053`
  - `L063`

This is a **BIG** pull request, but I'm not sure there's a simple way of doing it in less than one big push.


barrywhart:
I'll do a proper review later. One quick thought about `raw_segment_pre`: it would be pretty easy to implement if each segment knew its parent segment (happy to expand on this idea if you like).

alanmcruickshank:
> I'll do a proper review later. One quick thought about `raw_segment_pre`: it would be pretty easy to implement if each segment knew its parent segment (happy to expand on this idea if you like).

Go on?

The problem I see is that at the moment a segment isn't _aware_ of it's parent. All the references go down the tree not up. While crawling we do have good access to the parent, but the previous raw segment might not be in the parent, it might be several layers up. Combine that with more aggressive skipping, where we don't process every raw segment anyway - then the easiest way to get the previous raw would be to use `.raw_segments` and get the last one (by doing something like `.raw_segments[-1]` every time we skip). If we're calling `.raw_segments` anyway then I feel like we're doing almost as much processing as just tracking the `raw_stack` - unless the only issue is the continuous tuple arithmetic, in which case maybe we just use a list rather than a tuple and just make a raw stack a more efficient thing to provide.

barrywhart:
Definitely not a necessity. It was a good optimization for the current design. We can get rid of it for now. 👍

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -632,19 +665,24 @@ def is_optional(cls):
...
+        # Work backward in the class heirachy to see if there's a match
+        for class_type in cls._class_types():
+            if class_type in seg_type:
                 return True
-            elif base_class.type == "base":
-                break
         return False
 
+    @classmethod
+    def _class_types(cls) -> Iterator[str]:
```
barrywhart, line 675: Think it'd be worth precomputing this at class definition time? (Potential performance improvement.)

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -391,6 +395,35 @@ def raw(self) -> str:
         """Make a string from the segments of this segment."""
         return "".join(seg.raw for seg in self.segments)
 
+    @cached_property
```
barrywhart, line 398: See my comment on the `_class_types()` function. It seems like it may be possible to fully precompute this at class definition time.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -632,19 +665,24 @@ def is_optional(cls):
...
-        # Do we match on the type of _this_ class.
-        if cls.type in seg_type:
-            return True
-        # If not, check types of parents.
-        for base_class in cls.__bases__:
-            if base_class is object:
-                break
-            elif base_class.type in seg_type:
+        # Work backward in the class heirachy to see if there's a match
+        for class_type in cls._class_types():
```
barrywhart, line 669:
```suggestion
        # Work backward in the class hierarchy to see if there's a match
        for class_type in cls._class_types():
```

Note if we precompute `_class_types()`, this loop can be replaced with a simple set-based check.

src/sqlfluff/core/rules/crawlers.py
```diff
@@ -0,0 +1,125 @@
+"""Definitions of crawlers."""
+
+from typing import Iterator, Set, cast
+from sqlfluff.core.parser.segments.base import BaseSegment
+from sqlfluff.core.parser.segments.raw import RawSegment
+
+from sqlfluff.core.rules.context import RuleContext
+
+
+class BaseCrawler:
```
barrywhart, line 10: We could declare this class as an abstract base class to prevent creating it. I think that'd remove the need for the special `crawl()` implementation that just raises an error.

https://docs.python.org/3/library/abc.html

src/sqlfluff/core/rules/crawlers.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def passes_filter(self, segment: BaseSegment):
+        """Returns true if this segment considered at all."""
+        return self.works_on_unparsable or not segment.is_type("unparsable")
+
+    def crawl(self, context: RuleContext) -> Iterator[RuleContext]:
+        """Yields a RuleContext for each segment the rule should process."""
+        raise NotImplementedError(  # pragma: no cover
+            "Use a sepecific crawler, not the BaseCrawler. If you're seeing "
+            "this error, a rule has likely not set it's crawl_behaviour attribute."
```
barrywhart, line 24:
```suggestion
            "Use a specific crawler, not the BaseCrawler. If you're seeing "
            "this error, a rule has likely not set its crawl_behaviour attribute."
```

src/sqlfluff/core/rules/crawlers.py
```diff
@@ -0,0 +1,125 @@
...
+    """A crawler that doesn't crawl.
+
+    This just yields one context on the parent segment of the file.
+    """
+
+    def crawl(self, context: RuleContext) -> Iterator[RuleContext]:
+        """Yields a RuleContext for each segment the rule should process."""
+        if not self.passes_filter(context.segment):
+            return  # pragma: no cover
+        yield context
```
barrywhart, line 38: Simpler & avoids `no cover`:
```suggestion
        if self.passes_filter(context.segment):
            yield context
```

src/sqlfluff/rules/L060.py
```diff
@@ -41,12 +42,13 @@ class Rule_L060(BaseRule):
...
     groups = ("all",)
+    crawl_behaviour = SegmentSeekerCrawler({"function_name_identifier"})
 
     def _eval(self, context: RuleContext) -> Optional[LintResult]:
         """Use ``COALESCE`` instead of ``IFNULL`` or ``NVL``."""
-        # We only care about function names.
-        if not context.segment.is_type("function_name_identifier"):
-            return None
+        # We only care about function names, and they should be the
+        # onlythings we get
```
barrywhart, line 50:
```suggestion
        # only things we get
```

barrywhart (commented):
Looks great so far! A few small tactical suggestions, but overall direction is 💯.

@WittierDinosaur, are there any specific rules you've noticed use a lot of runtime that could benefit from using the new crawl behavior? Interested in testing this at some point on some of your large SQL files?

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: That's a great idea.... but how? I'm not aware of how to compute and store class attribute on definition. @barrywhart are you?

Best I can think of would be another cached property - but that's not computed on definition, just on first call.

barrywhart: Metaclasses are great for this. They basically provide a way to call a custom hook on `BaseSegment` that is called at class definition time (for `BaseSegment` or any subclass of it).

I threw this together quickly, but I think it will do what we need. It's adapted from the example in this [tutorial](https://www.pythontutorial.net/python-oop/python-metaclass-example/).

```
class SegmentMetaclass(type):
    def __new__(mcs, name, bases, class_dict):
        class_obj = super().__new__(mcs, name, bases, class_dict)
        class_types = {class_dict["type"]}
        for base in bases:
            class_types.update(base.class_types)
        class_obj.class_types = class_types
        return class_obj


class BaseSegment(metaclass=SegmentMetaclass):
    type = "base"


class RawSegment(BaseSegment):
    type = "raw"


print(BaseSegment.class_types)
print(RawSegment.class_types)
```

alanmcruickshank: 🤯 learning new things every day! I've never used metaclasses before but I'll have a go.

alanmcruickshank: @barrywhart - ok I think I got the metaclasses working. Check out what I've done and see what you think.

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: We can precompute it for most, but for segments like `KeywordSegment` we need to be able to add a shim for some instance-specific logic. See my comment in reply to your suggestion of metaclasses.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -391,6 +423,38 @@ def raw(self) -> str:
         """Make a string from the segments of this segment."""
         return "".join(seg.raw for seg in self.segments)
 
+    @cached_property
```
barrywhart, line 426: There's probably no reason to cache this, since it's simply returning an already-computed set.

I think could even eliminate this function completely by changing the metaclass to populate `class_types` rather than `_class_types`.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -391,6 +423,38 @@ def raw(self) -> str:
...
+    def class_types(self) -> Set[str]:
+        """The set of types for this segment."""
+        # NOTE: This version is simple, but some dependent classes
+        # (notably RawSegment) override this with something more
+        # custom.
+        return self._class_types
+
+    @cached_property
+    def child_type_set(self) -> Set[str]:
```
barrywhart, line 435: Would it be clearer to call this `descendant_type_set` (child, grandchild, etc.)?

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -120,6 +120,16 @@ def segments(self):
         """
         return []
 
+    @property
```
barrywhart, line 123: Is this worth caching?

src/sqlfluff/core/rules/base.py
```diff
@@ -99,6 +99,14 @@ def __init__(
         # store a description_override for later
         self.description = description
 
+    def __repr__(self):
+        if not self.anchor:
+            return "LintResult(<empty)"
```
barrywhart, line 104:
```suggestion
            return "LintResult(<empty>)"
```

src/sqlfluff/core/rules/base.py
```diff
@@ -99,6 +99,14 @@ def __init__(
...
+        fix_coda = f"+{len(self.fixes)}F" if self.fixes else ""
```
barrywhart, line 105: Is the capital `F` at the end of the string intentional?

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: `RawSegment` overrides `class_types` to allow it to include the instance variable `_surrogate_type`. Is still needs the class method, but this shim is there to be overridden. You're probably right that it doesn't need to be cached though.

src/sqlfluff/core/parser/segments/raw.py
alanmcruickshank: If the things which call it are cached - probably not.

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: Good idea.

src/sqlfluff/core/rules/base.py
alanmcruickshank: Yeah it's supposed to be short for "fixes". I'll add a note.

barrywhart:
> If we're calling `.raw_segments` anyway then I feel like we're doing almost as much processing as just tracking the raw_stack - unless the only issue is the continuous tuple arithmetic, in which case maybe we just use a list rather than a tuple and just make a raw stack a more efficient thing to provide.

I think the performance issue comes from continuously building and appending to long sequences (every raw segment in the file). That's pretty expensive whether a list or a tuple is used. We don't need to address it in this PR, which significantly improves performance in other ways that probably drown out the impact of this. Just sharing the thought as a possibility for future performance work.

src/sqlfluff/rules/L016.py
```diff
@@ -481,28 +486,37 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+            )
+            if quoted_literals:
+                self.logger.debug(
+                    "Comment Checking: Found quoted literals: %s", quoted_literals
+                )
+                root = context.parent_stack[0]
+
+                def is_in_comment(seg):
+                    path = root.path_to(seg)
+                    parent = path[-2]
```
barrywhart, line 508: Are we guaranteed `path[-2]` won't cause an `IndexError`? A comment here about the structure we expect (i.e. why `-2`) would be helpful.

src/sqlfluff/rules/L039.py
```diff
@@ -34,139 +33,90 @@ class Rule_L039(BaseRule):
     """
 
     groups = ("all", "core")
-    needs_raw_stack = True
+    # We're looking for whitespace.
+    crawl_behaviour = ParentOfSegmentCrawler({"whitespace"})
+
+    @staticmethod
+    def _check_diff_idx_for_casting_operator(segments, idx: int, diff: int) -> bool:
```
barrywhart, line 40: Could you add a docstring to this function? Also perhaps for each case, a small example or explanation of what that case matches?

src/sqlfluff/rules/L039.py
```diff
@@ -34,139 +33,90 @@ class Rule_L039(BaseRule):
...
                         )
-                prev_newline = False
-                prev_whitespace = None
-
-            if seg.is_type("object_reference"):
-                for child_seg in seg.get_raw_segments():
-                    if child_seg.is_whitespace:
+                    # If we find whitespace at the start of a segment it's probably
+                    # from a fix, so leave it be. It otherwise shouldn't be there.
+                    elif idx == 0:
```
barrywhart, line 96: I'm surprised this still happens. A few months back, I modified the `_apply_fixes()` function so any "misplaced" whitespace would be "bubbled up" to an ancestor segment, attempting to force fixes to leave the updated parse tree in a "clean" state. (This was knock-on work from the earlier work to re-parse modified areas of the tree to ensure fixes didn't introduce parse errors.)

src/sqlfluff/rules/L052.py
```diff
@@ -79,7 +80,7 @@ def _handle_preceding_inline_comments(before_segment, anchor_segment):
                 if s.is_comment
                 and s.name != "block_comment"
                 and s.pos_marker.working_line_no
-                == anchor_segment.pos_marker.working_line_no
+                == anchor_segment.raw_segments[-1].pos_marker.working_line_no
```
barrywhart, line 83: Is it possible `raw_segments` is empty and will fail with `IndexError`?

src/sqlfluff/rules/L053.py
```diff
@@ -48,56 +49,77 @@ class Rule_L053(BaseRule):
...
-        # We only care about bracketed segments that are direct
-        # descendants of a top-level statement segment.
-        filtered_parent_stack = [
-            segment for segment in context.parent_stack if segment.type != "batch"
-        ]
-        if not (
-            context.segment.is_type("bracketed")
-            and [segment.type for segment in filtered_parent_stack]
-            == ["file", "statement"]
+        # Because of the root_only_crawler, this can control it's own
```
barrywhart, line 77:
```suggestion
        # Because of the root_only_crawler, this can control its own
```

test/core/parser/segments_base_test.py
```diff
@@ -46,6 +46,23 @@ def test__parser__base_segments_type():
     assert DummySegment.class_is_type("base", "foo", "bar")
 
 
+def test__parser__base_segments_class_types():
+    """Test the metclass ._class_types attribute."""
```
barrywhart, line 50:
```suggestion
    """Test the metaclass ._class_types attribute."""
```

barrywhart (changes requested):
Looks good overall! Lots of code to review, but I tried to identify any red flags. The existing test cases plus the "rules critical error" checks tend to be very helpful, but I did find a few potential issues where an IndexError may be triggered.

src/sqlfluff/rules/L005.py
```diff
@@ -40,10 +41,11 @@ class Rule_L005(BaseRule):
     """
 
     groups = ("all", "core")
+    crawl_behaviour = SegmentSeekerCrawler({"comma"}, provide_raw_stack=True)
 
     def _eval(self, context: RuleContext) -> Optional[LintResult]:
         """Commas should not have whitespace directly before them."""
-        anchor: Optional[RawSegment] = context.raw_segment_pre
+        anchor: Optional[RawSegment] = context.raw_stack[-1]
```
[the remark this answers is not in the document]
alanmcruickshank: It's very unlikely that a file begins with a comma - but I think strictly you're probably right. I'll add a clause.

src/sqlfluff/rules/L016.py
alanmcruickshank: Will add a docstring to explain and a clause for if `path` is too short.

src/sqlfluff/rules/L005.py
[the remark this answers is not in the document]
barrywhart: Thanks! I'm not generally paranoid, but for a while, we'd have a bug or two every release where a user would hit one of these unexpected conditions. We added the rules critical errors" test to help catch these things, and it's gotten better since then.

src/sqlfluff/rules/L039.py
alanmcruickshank: Yeah - I agree. I'm not going to address this straight away. I'll bring that into the work on reflow. I think that extends the aims of this PR too far.

src/sqlfluff/rules/L052.py
alanmcruickshank: Good question 🤔 . I'm going to have to look into that one.

src/sqlfluff/rules/L039.py
barrywhart: Agreed. I have no boundaries, ha ha. And it's not urgent, more of a 🤔 thing.

src/sqlfluff/rules/L052.py
alanmcruickshank: I looked into this one - we can actually be sure that `anchor_segment` is actually a `RawSegment` itself. Calling `.raw_segments` on a `RawSegment` just returns `[self]`, so this can just be reduced to `anchor_segment.pos_marker.working_line_no`. I've updated the mypy type hints accordingly and simplified the function.

alanmcruickshank: Nope - I'm wrong. `anchor_segment` isn't always a `RawSegment` (found that out in the now failing tests). However, `raw_segments` will always have length - either a segment will have children, or it will return `[self]` on `.raw_segments` if it's a `Rawsegment`.

src/sqlfluff/rules/L039.py
```diff
@@ -34,139 +33,112 @@ class Rule_L039(BaseRule):
     """
 
     groups = ("all", "core")
-    needs_raw_stack = True
+    # We're looking for whitespace.
+    crawl_behaviour = ParentOfSegmentCrawler({"whitespace"})
+
+    @staticmethod
+    def _check_diff_idx_for_casting_operator(segments, idx: int, diff: int) -> bool:
+        """Check whether the this segment adjoins a casting operator.
```
barrywhart, line 41:
```suggestion
        """Check whether this segment adjoins a casting operator.
```

barrywhart (approved):
Looks good! I fixed one small typo in a docstring. Good to merge once the build passes.

Excellent work!! 🎉🥳

PR #3718, opened 2022-08-07
Build out type hints in Grammars

alanmcruickshank:
This PR builds out the type interfaces on base grammar functions.

This is linked to #3692, but somewhat incidental in the process.

I don't think there's anything controversial in here, just more type hints and removing a couple of code paths which are clear that they're not used now the type hints are tighter.


src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -304,8 +309,6 @@ def _look_ahead_match(cls, segments, matchers, parse_context):
 
         # Do some type munging
         matchers = list(matchers)
```
barrywhart, line 311: A bit off topic for the PR, but a possible performance improvement: I looked through the function, and I see no reason to make a new list here. Could we remove this copying? It's likely to have some impact on performance.

src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -413,7 +417,7 @@ def _trim_elem(seg):
 
         # Make some buffers
         seg_buff = segments  # pragma: no cover
```
barrywhart, line 419: Could the lines from here to the end of the function be removed? (They are all marked `no cover`.)

src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -492,10 +499,8 @@ def _bracket_sensitive_look_ahead_match(
         """
         # Type munging
         matchers = list(matchers)
-        if isinstance(segments, BaseSegment):  # pragma: no cover TODO?
-            segments = [segments]
 
-        # Have we been passed an empty list?
+        # Have we been passed an empty tuple?
         if len(segments) == 0:
```
barrywhart, line 504: Performance nitpick -- this will likely be faster (also PEP-8 compliant, not that I am a stickler for that):
```suggestion
        if not segments:
```

barrywhart:
Looks good, but a few questions & suggestions. Happy to approve, but wanted to have a quick discussion first.

src/sqlfluff/core/parser/grammar/base.py
alanmcruickshank: 👍

src/sqlfluff/core/parser/grammar/base.py
alanmcruickshank: I wondered that myself but I think was too chicken to try it. If we're both thinking it I'll try!

alanmcruickshank: Hmmmm - actually I think it's more complicated than that - I'll make a separate issue for it though, to talk through things. Great idea though.

alanmcruickshank: new issue is live: #3721 Let's handle that there.

barrywhart (approved):
Looks good!

Issue #3721, opened 2022-08-08
Remove non-simple matching code in _look_ahead_match


alanmcruickshank:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

As raised in #3718, there's a section of code in the back half of `_look_ahead_match` in [src/sqlfluff/core/parser/grammar/base.py](https://github.com/sqlfluff/sqlfluff/pull/3718/files/59028f1033bfe85860f111848c7a8e4adad90b7f#diff-98d814491a9ac9986937dfdfaabdc1fd54956e961d23bce7811460fe9ac4120f). The latter half of that function is covered with `# pragma: no cover` which implies it's not covered in our current test suite.

This feels confusing, because the section it refers to the _non-simple_ matching process. It not being covered in tests implies that everything has a `simple` match method (which I don't believe is true). So either there's a bug where it's not being used (which we should fix) - or my understanding of the code is wrong (and so we should delete this code - or at least remove it and put an error in that place to explain what's happened).

### Dialect

all

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

alanmcruickshank:
@barrywhart @WittierDinosaur do either of you have views on this? I feel like you might be the others which have been down in this code.

tunetheweb:
My view is we have pretty good, comprehensive cover so it likely isn't needed. And if it is I'd rather know about it.

So I'm in favour of removing it. Incidentally there's 21 `# pragma: no cover`'s in that file, of which about half of them are in exceptions (fair enough - ideally shouldn't be reached but safety so don't mind them having no cover), and half of them not (so unused code and can be removed?).

This is why I like 100% cover, even though it meant adding a few `# pragma: no cover`'s 😁 - makes it easier to spot things like this.

barrywhart:
I haven't worked much in this area of the parser, but I tend to believe our tests are thorough enough to trust in this case. Not 100% sure, but reasonably confident.

alanmcruickshank:
I'm going to close this now I've merged #3723

PR #3723, opened 2022-08-08
Revise no cover direction and remove unused code.

alanmcruickshank:
Draft PR to start resolving #3721.

I've removed some definitely unrequired code and removed the `no cover` tags from the other section. I want to see what coverage says before going further - when I initially tried just removing the other section I got some test fails so that implies that it _is_ being used.

### DIFF
--- a/src/sqlfluff/core/parser/grammar/base.py
+++ b/src/sqlfluff/core/parser/grammar/base.py
@@ -2,7 +2,7 @@
 
 import copy
 from dataclasses import dataclass
-from typing import List, Optional, Union, Type, Tuple, Any
+from typing import TYPE_CHECKING, List, Optional, Union, Type, Tuple, Any
 
 from sqlfluff.core.errors import SQLParseError
 from sqlfluff.core.string_helpers import curtail_string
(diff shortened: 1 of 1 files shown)


src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -882,9 +886,6 @@ def match(self, segments, parse_context):
                 if self.exclude.match(segments, parse_context=ctx):
                     return MatchResult.from_unmatched(segments)
 
-        if not elem:  # pragma: no cover
-            raise ValueError(f"Null Element returned! _elements: {self._elements!r}")
-
```
tunetheweb, line 887: Definitely want to remove this one? I'm usually OK leaving in `no cover`s for exceptions.

alanmcruickshank: I've removed it because the value it refers to should never be `None` - see the type hints I've added to `_get_elem` above. It calls `Dialect.ref()` which will raise an exception rather than return `None`.

alanmcruickshank:
Looks like from the coverage reports that those lines which had `no cover` on them were actually being used. There is a clause within there which isn't tested - but I think that might be a separate job to work out how to test it.

tunetheweb (approved):

PR #3726, opened 2022-08-09
New rule L067 for adjusting the subquery inside parentheses

XiChu2333:
### Brief summary of the change made
fixes:https://github.com/sqlfluff/sqlfluff/issues/1862

### Pull Request checklist
- [ ] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

Issue #1862, opened 2021-11-09
Formatting rules for subqueries


fdw:
Similar to #1861, I would expect sqlfluff to have formatting rules regarding subqueries, for example
```sql
if exists(select 1 from...)
begin
  drop ...
end
```

I personally would prefer the subquery to be in a new, indented line, but it's also okay if it's consistently in the same line. Currently, however, there is no alignment whatsoever.

barrywhart:
Can you provide a full example query? Also, what dialect are you using above?

fdw:
Sure! Here's a fuller example:
```sql
if exists (select * from sys.objects where object_id = object_id(N'[SomeSchema].[ProcedureName]') and type in (N'P'))
begin
  drop procedure SomeSchema.ProcedureName
end
```

The SQL dialect is T-SQL.

jpy-git:
I'd be happy to implement a rule for subqueries that formats them on a new indented line. This approach would work nicely with rule L003 as it would ensure that the subquery segments have correct indentation relative to the outer statement.

So
```SQL
SELECT *
FROM (SELECT * FROM foo)
```
becomes
```SQL
SELECT *
FROM (
    SELECT * FROM foo
)
```

barrywhart:
It sounds good to me. I could see this as a new rule out a new, optional behavior for L003. (Either seems okay from a user perspective. I'm not sure which would be easier to implement.)

jpy-git:
I think a new rule would be simplest, let L003 handle indentation only rather than also moving segments vertically. I'll pick this up tomorrow 🚀

barrywhart:
>I think a new rule would be simplest
That's my guess, too. If necessary, you can copy or otherwise reuse some code from L003. We have a few other rules which do that.

fdw:
What is the plan for long subqueries? Can they be broken down according to L003, but with a larger indent?

jpy-git:
There's already a rule for line too long so that should handle that

barrywhart:
We'll want some tests for that scenario. Idea: Copy several of the existing L003 tests and make the existing query a subquery.

jpy-git:
Update:
I've got this rule working fine for the subquery contents but it seems that L003 (or some other rule) is also trying to mess with subquery bracket position which I don't think we want. L003 is a bit of a beast so I'll need to find some time to figure out what's going on there.

barrywhart:
[...]

PR #3736, opened 2022-08-12
New rule L068 for adjusting bracket with preceding/following codes

XiChu2333:

Like https://github.com/sqlfluff/sqlfluff/pull/3726, this rule is also a proposal for totally fixing https://github.com/sqlfluff/sqlfluff/issues/1862. I tried to write them into one rule, but found it's very hard to mix these logically.
So this rule intends to work with L067 together, to make parentheses more beautiful.
The L067 focuses on the inside of brackets, while this one focus on the outside of brackets.
I will continue test on this rule. Tell me if you have any suggestions or ideas on this.

### Pull Request checklist

  - Other.


src/sqlfluff/rules/L068.py
```diff
@@ -0,0 +1,258 @@
...
+    .. code-block:: sql
+        :force:
+
+        SELECT
+            A.SOMETHING
+        FROM (SELECT B FROM D GROUP BY 1) AS A
+
+    """
+    groups = ("all",)
+
```
alanmcruickshank, line 55: (same as my comment on #3726 ).

Now that I've merged #3717 you'll need to add a `crawl_behaviour` attribute to govern how this rule crawls.

Looking at what you've done, it might be a bit more complicated in this case.

To pick up your conditions on `if context.segment.is_type('from_clause')` and `if context.segment.is_type('join_clause')`, you could very easily just set `crawl_behaviour = SegmentSeekerCrawler({"from_clause", "join_clause"})`.

The tricky one will be `if children.any(sp.is_type('alias_expression'))`. In this case I think you have a few options:
1. Given you're looking for bracketed expressions followed by an alias, you could also add `bracketed` to your search query and then use `parent_stack` to go up one level and see if the `bracketed` is followed by an `alias_expression`.
2. You could find the instances where aliases might follow a table expression, which I think involves only adding `select_clause_element` to your query (to catch examples like `SELECT (SELECT 1) as bar`).

My gut says the latter would be less effort. Let me know if you need a had reworking this - I realise this is new functionality, but it should drastically increase performance for rules exactly like this one.

src/sqlfluff/rules/L068.py
```diff
@@ -0,0 +1,258 @@
...
+    document_groups,
+)
+import sqlfluff.core.rules.functional.segment_predicates as sp
+
+
+@document_groups
+@document_configuration
+@document_fix_compatible
+class Rule_L068(BaseRule):
+    """The brackets should stand at the same line with from/as/join keywords.
```
barrywhart, line 21: Have you tested how this rule interacts with other rules such as L003 (indentation) and L016 (line length), perhaps others? I think there's potential for unexpected rule interactions.

test/fixtures/rules/std_rule_cases/L068.yml
```diff
@@ -0,0 +1,82 @@
+rule: L068
+
+test_pass_one_line_subquery:
+  pass_str: |
+      SELECT A.SOMETHING
+      FROM (SELECT B FROM D GROUP BY 1) AS A
+
+# The fix of indentation rely on L003
+# If there is a subquery inside the bracket, the fix also rely on L067
+test_fail_example_in_description:
```
barrywhart, line 10: Can you add tests where there are comments around the area the rule considers? Handling comments is often tricky. The rule may either:
* Handle comments correctly (harder)
* Detect comments and not try to auto-fix (easier)

barrywhart (commented):
Please add more test cases as mentioned. Also, please update with the latest changes from `main`. As Alan noted, you'll need to make code changes for this to work, and he can help if needed.

greg-finley:
Hello! I am closing out some stale PRs. Let's chat if you are interested in working on this more, as we have other work in flight that may affect this.

PR #3743, opened 2022-08-15
Migrate analysis, functional and testing to utils

alanmcruickshank:
This addresses and issue I raised on slack:

> Triggered by a few things:
> - I was talking to @WittierDinosaur about the RuleContext which is passed around during linting. We do quite a lot to populate it with data pre-emptively and it also complicates rules significantly when passed around. It would be better if we could lighten the context.
> - I've been thinking about how to streamline the core part of sqlfluff to make it easier to make changes within it, potentially also to rationalise how testing happens - in that we could have a slimmer test suite for core which runs first before triggering the others.
> - I've been trying to work out where I put the reflow routines and how we might interact with them. So far I've been thinking that they should sit outside core but importable by rules which need them. I'd also love to take a leaf from some of the functional work that's been done so far.
> - Looking at the sqlfluff.core.rules.analysis and sqlfluff.core.rules.functional modules, they're both only really used by rules and nothing else in the core library depends on them (except the latter as a way of making it easy to import).

> How do people feel like me making a sqlfluff.util or sqlfluff.analysis module which would house analysis and functional  (and eventually reflow) and bringing them out of core. The only downside I can see for this is that rules that want to use the functional approach would have to import it directly rather than accessing it directly from the context. In the process I would propose slimming down the context significantly and allowing more options for rules to pull what they need rather than being provided it by default.

This is a draft of what I'm talking about. This would eventually be where the reflow utils will live.


src/sqlfluff/utils/functional/context.py
```diff
@@ -0,0 +1,54 @@
+"""Define FunctionalContext class."""
+
+from sqlfluff.utils.functional import Segments
+from sqlfluff.core.rules import RuleContext
+
+
+class FunctionalContext:
```
barrywhart, line 7: I have a possible alternative to consider for how we expose this:

For crawl behavior, there are multiple implementations, and a rule "declares" what it wants by setting the class-level `crawl_behaviour` field.

Should we consider doing something similar for rule context? In some sense, this would make the two variants of `RuleContext` "equal" in the sense of being equally easy for rules to use.

To be clear, the PR as written seems reasonable, but from a more philosophical point of view, it's interesting to consider. If we wanted to do this, it occurs to me that it may complicate the type annotations for the `_eval()` function (because the type of the "context" parameter is no longer a given).

test/core/plugin_test.py
```diff
@@ -6,13 +6,15 @@
...
-    # The plugin import order is non-deterministic
-    assert sorted(
-        [plugin_module.__name__ for plugin_module in plugin_manager.get_plugins()]
-    ) == [
+    # The plugin import order is non-deterministic.
+    # Use sets in case the dbt plugin (or other plugins) are
+    # already installed too.
+    assert set(
+        plugin_module.__name__ for plugin_module in plugin_manager.get_plugins()
+    ) >= {
```
barrywhart, line 14: Nitpick: I tend to prefer the named functions on `set`, because it's clearer to me what operation is being used. What function does `>=` correspond to?

barrywhart (approved):
Seems good overall! I had one philosophical question which could be deferred, and one nitpicky suggestion.

test/core/plugin_test.py
alanmcruickshank: _is a subset of_.

I'm happy to update it to the function name - you're right that it's clearer.

src/sqlfluff/utils/functional/context.py
```diff
@@ -0,0 +1,54 @@
...
+            *self.context.raw_stack, templated_file=self.context.templated_file
+        )
+
+    @property
+    def raw_segments(self):
+        """Returns a Segments object for all the raw segments in the file."""
+        file_segment = self.context.parent_stack[0]
+        return Segments(
+            *file_segment.get_raw_segments(), templated_file=self.context.templated_file
+        )
```
OTooleMichael, line 54:
```suggestion
        )
 
def functional_eval(func: Callable[[rule_self: Any, context: FunctionalContext]]):
    def wrapper(self, context: RuleContext):
        fn_context = FunctionalContext(context)
        return func(self, fn_context)
    return wrapper
```

src/sqlfluff/utils/functional/context.py
alanmcruickshank: Based on the slack convo, I think the based approach for this is probably to go with a decorator, which additionally provides the functional context.

```
@functional_context
def _eval(self, context, functional):
...
```

It won't be crazy hard to add at this stage so I'll see if I can incorporate it into this PR.

alanmcruickshank: > Go with a decorator for sure, but use the decorator to subclass the context :slightly_smiling_face:

alanmcruickshank: On second thoughts I'm going to make an issue and approach that separately. This PR is also hella big.

src/sqlfluff/utils/functional/context.py
alanmcruickshank: I'm going to address this as a separate issue and moved the discussion here: https://github.com/sqlfluff/sqlfluff/issues/3745

src/sqlfluff/utils/functional/context.py
alanmcruickshank: https://github.com/sqlfluff/sqlfluff/issues/3745

PR #3747, opened 2022-08-15
New rule L069: Consistent syntax for sql type casting

### LINKED ISSUE #3713: Add a rule to enforce consistent syntax for SQL type casting
### Search before asking

### Description
Apart from supporting the SQL standard type casting variant `CAST(expression AS type)`, several dialects have other variants for type casting.

The other variants are listed below:

1. PostgreSQL https://www.postgresql.org/docs/current/sql-expressions.html#SQL-SYNTAX-TYPE-CASTS
```
expression::type
```
2. SQL Server https://docs.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver16
```
CONVERT(type, expression)
```
3. MySQL https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html
```
CONVERT(type, expression)
```
4. Redshift https://docs.aws.amazon.com/redshift/latest/dg/r_CAST_function.html
```
expression::type
CONVERT(type, expression)
```
5. Snowflake https://docs.snowflake.com/en/sql-reference/functions/cast.html
```
expression::type
```
6. Clickhouse https://clickhouse.com/docs/en/sql-reference/functions/type-conversion-functions/#castx-t
```
CAST(type, expression)
expression::type
```
7. Teradata https://docs.teradata.com/r/kmuOwjp1zEYg98JsB8fu_A/BQxvaT7YNqCkkfWB0FaYvA
```
CAST(expression AS type FORMAT expression)
```

Currently, there is no way to enforce consistent type casting in a codebase involving these dialects. We could have a rule specifying the type of casting we want to try to enforce(`STANDARD` or `NON-STANDARD`). This rule should be configurable. 

REF: https://sqlfluff.slack.com/archives/C01GPHR8J0G/p1659570939547909

### Use case

Consistent type casting for the above-listed dialects

### Dialect

postgres, tsql, clickhouse, redshift, mysql, teradata, snowflake

### Code of Conduct

bolajiwahab:

Fixes #3713 

Had to regenerate all fixtures that included `cast_expression`.

### Pull Request checklist

  - Other.


bolajiwahab:
@barrywhart and @alanmcruickshank  if you have some spare time, can you review this PR? Thanks

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -979,7 +979,19 @@ class ShorthandCastSegment(BaseSegment):
 
     type = "cast_expression"
     match_grammar: Matchable = Sequence(
-        Ref("CastOperatorSegment"), Ref("DatatypeSegment"), allow_gaps=True
```
barrywhart, line 982: @tunetheweb: Can you review the dialect change portion of this PR?

src/sqlfluff/rules/L013.py
```diff
@@ -64,6 +64,18 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
         ):
             return None
 
+        # Ignore if it's a cast_expression with non-function enclosed children
+        # For example, we do not want to ignore something like func()::type
+        # but we can ignore something like a::type
+        if children.children().select(
```
barrywhart, line 70: Does this fix an existing issue, or is this related to the ANSI dialect change? I'm a bit confused and worried why a new rule requires both dialect changes as well as changes to 3 other rules.

Since you've added logic to this rule, can you please add tests for it?

(Probably to the file `test/fixtures/rules/std_rule_cases/L013.yml`)

src/sqlfluff/rules/L069.py
```diff
@@ -0,0 +1,175 @@
...
+                expression = convert_content[1].raw
+
+                # get the data type
+                datatype = convert_content[0].raw
+
+                # Create fix to replace ``CONVERT`` with ``CAST``.
+                fix = LintFix.replace(
+                    context.segment,
+                    [
+                        CodeSegment(
```
barrywhart, line 111: Can you look into creating the new code using the individual segment types, rather than a single raw segment? During the fix process, we want to maintain a parse tree that matches what you'd get from parsing the fixed code. Otherwise, linting rules won't work well with one another.

src/sqlfluff/rules/L069.py
```diff
@@ -0,0 +1,175 @@
...
+                    description=f"Use 'CAST' instead of \n"
+                    f"'{context.segment.get_child('function_name').raw_upper}'.",
+                )
+
+            # casting_operator(::)
+            if context.segment.is_type("cast_expression"):
+                # get the expression and the datatype segment
+                expression_datatype_segment = FunctionalContext(
+                    context
+                ).segment.children(
```
barrywhart, line 129: Nitpick: This code is very similar to lines 79-94. Possible to avoid the duplication?

src/sqlfluff/rules/L069.py
```diff
@@ -0,0 +1,175 @@
...
+                    # update new_segment to include datatype
+                    # as well as accommodate multicast
+                    new_datatype = expression_datatype_segment[datatype_level].raw
+                    new_segment = f"CAST({new_segment} AS {new_datatype})"
+
+                # create fix to replace ``::`` with ``CAST``.
+                fix = LintFix.replace(
+                    context.segment,
+                    [
+                        CodeSegment(
```
barrywhart, line 161: Same comment as before: can you create individual segments?

barrywhart (changes requested):

src/sqlfluff/rules/L069.py
```diff
@@ -0,0 +1,175 @@
...
+    def _eval(self, context: RuleContext) -> Optional[LintResult]:
+        """Use ``CAST`` instead of ``CONVERT`` or ``::``."""
+        # Config type hints
+        self.allow_non_standard_type_casting: bool
+
+        # Rule disabled for teradata.
+        if context.dialect.name == "teradata":
+            return None
+
+        if not self.allow_non_standard_type_casting:
```
OTooleMichael, line 68: Don't this this True, false makes sense.

If False is the same as excluding the Rule in general.

PS this would be easier to see if it read

```python
if self.allow_non_standard_type_casting:
   return None
```

src/sqlfluff/rules/L069.py
```diff
@@ -0,0 +1,175 @@
...
+    document_groups,
+)
+
+
+@document_groups
+@document_fix_compatible
+@document_configuration
+class Rule_L069(BaseRule):
+    """Use ``CAST`` instead of ``CONVERT`` or ``::``.
+
```
OTooleMichael, line 21: Would be better to support
`consistent, shorthand, or cast` as options.
consistent is a very common approach from SQLFluff - meaning once we see a first type, `::` or `CAST` etc, then all the rest of the script gets changed to that style

OTooleMichael (changes requested):
This rule should support different opinions, and the concept of `consistent` (a common approach) - this also allows the introduction of a new rule without much change for current users

src/sqlfluff/rules/L013.py
bolajiwahab: It is related to the ANSI dialect change.
L013 enforces `Column expression without alias`. L013 was written to skip `cast_expression`(ShorthandCastSegment + DataTypeSegment). But now we have ExpressionSegment as part of `cast_expression` which means things like `SELECT CAST(foo_col AS INT)::VARCHAR` won't fail anymore, this breaks the logic of L013 because now we see a different `cast_expression` which can include functions. I hope that clarifies the logic above.

We already have a test for it, I was simply rewriting https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/rules/L013.py#L104. The test is can be found here https://github.com/sqlfluff/sqlfluff/blob/main/test/fixtures/rules/std_rule_cases/L013.yml#L32

src/sqlfluff/rules/L069.py
bolajiwahab: I am not sure how that would look. `cast_expression` comprises a single segment that is being changed. Also, something like `a::varchar` => `cast_expression` is being transformed into `cast(a as varchar)` => function, so I am not sure how we could use individual segment here.

src/sqlfluff/rules/L069.py
bolajiwahab: Yeah, that works too.

src/sqlfluff/rules/L069.py
bolajiwahab: Alright. Thanks for suggesting that. Will look into it.

src/sqlfluff/rules/L069.py
barrywhart: I mean, the rule heeds to create the same segments that would exist if you'd written the code by hand. You can use `sqlfluff parse` on a manually fixed file to figure out what it looks like. Other rules have examples of how to write the Python.

src/sqlfluff/rules/L069.py
bolajiwahab: Yes, I agree. The main issue is the bracketed content in `convert`. I don't see an easy way around it.

src/sqlfluff/rules/L069.py
bolajiwahab: Okay, could you point me to an example?

barrywhart: L042 and L043 do this. Unfortunately, these techniques are not well documented, partly because they are not well established. Most other autofix rules fix individual segments or at least a "flat" structure. In this case, the fix involves multiple, nested segments.

barrywhart: Also L032.

src/sqlfluff/rules/L069.py
barrywhart: You could create a separate function with the code inside it, e.g.:
```
def do_something(segments: Segments) -> Segments:
return segments.children(                    sp.and_(
                        sp.not_(sp.is_meta()),
                        sp.not_(
                            sp.is_type(
                                "start_bracket",
                                "end_bracket",
                                "whitespace",
                                "newline",
                                "casting_operator",
                            )
                        ),
                    )
                )
```

then call that function from both places.

src/sqlfluff/rules/L069.py
bolajiwahab: Okay. Thank you, I am checking them.

src/sqlfluff/rules/L069.py
bolajiwahab: Fixed by [ebeb4ad](https://github.com/sqlfluff/sqlfluff/pull/3747/commits/ebeb4ada735810c3d050d690d93e0d8fcc64d4fa)

src/sqlfluff/rules/L069.py
bolajiwahab: Fixed by [ebeb4ad](https://github.com/sqlfluff/sqlfluff/pull/3747/commits/ebeb4ada735810c3d050d690d93e0d8fcc64d4fa)

src/sqlfluff/rules/L069.py
bolajiwahab: Fixed by [ebeb4ad](https://github.com/sqlfluff/sqlfluff/pull/3747/commits/ebeb4ada735810c3d050d690d93e0d8fcc64d4fa)

src/sqlfluff/rules/L069.py
bolajiwahab: Fixed by [ebeb4ad](https://github.com/sqlfluff/sqlfluff/pull/3747/commits/ebeb4ada735810c3d050d690d93e0d8fcc64d4fa)

src/sqlfluff/rules/L069.py
```diff
@@ -29,147 +35,366 @@ class Rule_L069(BaseRule):
...
+                ),
+            )
+        )
+
+    @staticmethod
+    def _cast_fix_list(
+        context: RuleContext,
+        cast_arg_1: BaseSegment,
+        cast_arg_2: BaseSegment,
+        later_types=None,
```
barrywhart, line 91: What is `later_types`? Could you add a type annotation to help readers of the code?

src/sqlfluff/rules/L069.py
```diff
@@ -29,147 +35,366 @@ class Rule_L069(BaseRule):
...
+            "cast"
+            if context.segment.is_type("function")
+            and context.segment.get_child("function_name").raw_upper == "CAST"
+            else "convert"
+            if context.segment.is_type("function")
+            and context.segment.get_child("function_name").raw_upper == "CONVERT"
+            else "shorthand"
+            if context.segment.is_type("cast_expression")
+            else None
+        )
```
barrywhart, line 211: I find the multiple chained ternary `if / else` hard to read. What would you think of changing it to more traditional `if / else`, something like:
```
if context.segment.is_type("function")
            and context.segment.get_child("function_name").raw_upper == "CAST":
    current_type_casting_style = "cast"
elif context.segment.is_type("function")
            and context.segment.get_child("function_name").raw_upper == "CONVERT"
    current_type_casting_style = "convert"
etc.
```

src/sqlfluff/rules/L069.py
```diff
@@ -29,147 +35,366 @@ class Rule_L069(BaseRule):
...
-                # Create fix to replace ``CONVERT`` with ``CAST``.
-                fix = LintFix.replace(
-                    context.segment,
-                    [
-                        CodeSegment(
-                            raw=f"CAST({expression} AS {datatype})",
+            elif prior_type_casting_style == "convert":
+                if current_type_casting_style == "cast":
+                    cast_content = self._get_children(
+                        FunctionalContext(context).segment.children(
```
barrywhart, line 257: Nitpick: Maybe create the functional context once rather than repeating it several times, e.g.:
```
functional_context = FunctionalContext(context)
```

barrywhart (changes requested):
Great progress! A few suggestions.

Do you plan to make any other changes? (I'm asking so I know whether to do a more thorough review now.)

src/sqlfluff/rules/L069.py
```diff
@@ -0,0 +1,403 @@
...
+            )
+        ]
+        return fixes
+
+    @staticmethod
+    def _convert_fix_list(
+        context: RuleContext,
+        convert_arg_1: BaseSegment,
+        convert_arg_2: BaseSegment,
+        later_types=None,
```
barrywhart, line 135: What is `later_types`? Could you add a type annotation to help readers of the code?

src/sqlfluff/rules/L069.py
```diff
@@ -0,0 +1,403 @@
...
+                edits,
+            )
+        ]
+        return fixes
+
+    @staticmethod
+    def _shorthand_fix_list(
+        context: RuleContext, shorthand_arg_1: BaseSegment, shorthand_arg_2: BaseSegment
+    ) -> List[LintFix]:
+        """Generate list of fixes to convert CAST and CONVERT to ShortHandCast."""
```
barrywhart, line 176:
```suggestion
        """Generate list of fixes to convert CAST and CONVERT to ShorthandCast."""
```

test/fixtures/rules/std_rule_cases/L069.yml
```diff
@@ -0,0 +1,209 @@
...
+        cast(10 as text) as coo
+    from foo;
+  fix_str: |
+    select
+        convert(int, 1) as bar,
+        convert(text, convert(int, 100)),
+        convert(text, 10) as coo
+    from foo;
+
+# test_fail_inconsistent_type_casting_prior_3_argument_convert:
```
barrywhart, line 42: What's this commented-out test? Is this something you're still working on?

test/fixtures/rules/std_rule_cases/L069.yml
```diff
@@ -0,0 +1,209 @@
...
+    from foo;
+
+test_pass_convert:
+  pass_str: |
+    select convert(varchar, 1) as bar
+    from foo;
+
+test_pass_3_argument_convert:
+  pass_str: |
+    select convert(varchar, 1, 126) as bar
```
barrywhart, line 25: Please add another test with the same SQL string but with `preferred_type_casting_style` set to `cast`. In that case, what does the rule  do?
* Generate a lint warning but no autofix
* Pass (no lint warning, no autofix)

Which _should_ it do? I'm thinking the first, but I don't have a strong opinion.

test/fixtures/rules/std_rule_cases/L069.yml
```diff
@@ -0,0 +1,209 @@
...
+    from foo;
+
+test_fail_inconsistent_type_casting_prior_convert:
+  fail_str: |
```
barrywhart, line 29: Please add a similar test where there are comments in the way. What happens? E.g.
```
    select
        cast(10 as text) as coo,
        convert( -- Convert the value
            int /*
                  to an integer
                */, 1) as bar,
        100::int::text
    from foo;
```

bolajiwahab:
> Great progress! A few suggestions.
> 
> Do you plan to make any other changes? (I'm asking so I know whether to do a more thorough review now.)

No more changes.

src/sqlfluff/rules/L069.py
bolajiwahab: Yes, sure.

src/sqlfluff/rules/L069.py
bolajiwahab: I only used it to reduce the code but if it is hard to read then traditional it is then.

src/sqlfluff/rules/L069.py
bolajiwahab: `later_types` is any other type after the first type e.g in `1::int::text`, text is the later type. I couldn't get a better naming.

test/fixtures/rules/std_rule_cases/L069.yml
bolajiwahab: It is leftover which I forgot to clean up.

test/fixtures/rules/std_rule_cases/L069.yml
bolajiwahab: It currently does the second but the first one might be better.

bolajiwahab:
> Great progress! A few suggestions.
> 
> Do you plan to make any other changes? (I'm asking so I know whether to do a more thorough review now.)

@barrywhart I guess since you made a few suggestions then I will be adding a few more changes then.

barrywhart:
Thanks @bolajiwahab. Tag me when you're ready for another review. Great work!!

test/fixtures/rules/std_rule_cases/L069.yml
bolajiwahab: Implementing the first case is a bit complicated. It is easier when we have just one cast in a statement but in the case of multiple casts, some would be fixable while 3-args convert won't be fixable 
What do you suggest in cases like this:
```
select
        convert(int, 1, 126) as bar,
        100::int::text,
        cast(10 as text) as coo
    from foo;
```

test/fixtures/rules/std_rule_cases/L069.yml
bolajiwahab: Currently, we skip such convert. Since we do not exclude comments from the children, they make it into convert_content hence increasing the list to more than 2.
This is an edge case.

test/fixtures/rules/std_rule_cases/L069.yml
barrywhart: I think no fix, just a lint warning. A partial fix would look worse, I think. ❓

test/fixtures/rules/std_rule_cases/L069.yml
barrywhart: Ok. Is there a test for it? For future maintainers (you or others), it's helpful to have tests, perhaps with a comment saying, "Could we autofix this someday?"

bolajiwahab: Okay.

barrywhart:
@bolajiwahab: Are you planning to continue working on this PR?

bolajiwahab:
> @bolajiwahab: Are you planning to continue working on this PR?

Hi @barrywhart yeah, I will be pushing my latest change this week.

barrywhart:
Thanks for letting me know. There's no pressure, I was just looking over the list of PRs yesterday and was curious. 🙏

bolajiwahab:
> Thanks for letting me know. There's no pressure, I was just looking over the list of PRs yesterday and was curious. 🙏

Been quite hung up with a few things. I will find time this weekend to finish it up.

test/fixtures/rules/std_rule_cases/L069.yml
bolajiwahab: Hmm, this is harder to really implement as the offending casting style can be anywhere in the statement. We can generate a fix for a previous segment and find an offending segment in the later segments.

barrywhart: It's been so long, I can't remember the context on this. What would a partial fix look like in the above case?

barrywhart: SQLFluff now allows rules that are called *once* per file, passing it the root parse segment and letting the rule do its own traversal of the file. I don't know if that would be useful in the above case.

bolajiwahab: > SQLFluff now allows rules that are called once per file, passing it the root parse segment and letting the rule do its own traversal of the file. I don't know if that would be useful in the above case.

I am not sure it would help in the case of a statement.

barrywhart: Why do the fixed columns appear in a different order? And can you please remind me why they can't all be converted to `cast`? I'm very sorry, but I remember almost no details of this conversation from 45 days ago, even my own original question.

bolajiwahab: For the ordering, there is a different rule doing that, I will have to check for the rule name later.

Yes, I know it's quite long time ago. The original question was about throwing linting error if for instance cast_style is set to cast but since we cannot fix 3 argument convert, we should still throw linting error to the user and the user can decide to fix it manually since we don't auto-fix such. 
Now, I have implemented that but along the way I figured if we have several cast style for example 3 arguments convert and then casting operator while cast-style is set to cast, we fix the casting operator but leave out the convert(that is what I meant by partial fix). Now that I look at it again, partial fix might not have been the right word to use. We should still fix these ones that are fixable and leave anything not fixable out to the user. When we fix the ones matching and a user run lint again, the former convert would/should still raise a lint error.
Let me know my explanation is okay and what you think about my suggestions.
Thanks.

barrywhart: Probably L034 is the one changing the column order.

I think I understand now. Each issue found by a lint rule is handled separately. It is reasonable to fix some of these and not others. Often this is done either because there's no reasonable fix, or because autofixing certain issues is more complex and has not been implemented yet. So your description sounds like correct behavior.

bolajiwahab:
> A few lines of the rule are not covered by the tests. If it's reasonable to test those, that's good. If not, we add `# pragma: no cover`to get the coverage check to pass.

Yeah, looking into those.

src/sqlfluff/rules/L069.py
bolajiwahab: Type annotation added.

src/sqlfluff/rules/L069.py
bolajiwahab: Added.

bolajiwahab:
@barrywhart You may review the PR now.

barrywhart (approved):

barrywhart:
@OTooleMichael: You had requested changes on this PR. Can you review it again?

@tunetheweb: Can you please review the small dialect change to this PR?

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1075,7 +1075,19 @@ class ShorthandCastSegment(BaseSegment):
...
+        OneOf(
+            Ref("Expression_D_Grammar"),
+            Ref("CaseExpressionSegment"),
+        ),
+        AnyNumberOf(
+            Sequence(
+                Ref("CastOperatorSegment"),
+                Ref("DatatypeSegment"),
+                Ref("TimeZoneGrammar", optional=True),
+                allow_gaps=True,
```
tunetheweb, line 1087: Isn't this true by default? So do we need this?

```suggestion
```

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1075,7 +1075,19 @@ class ShorthandCastSegment(BaseSegment):
...
+            ),
```
tunetheweb, line 1088: Should this have a `min_times=1`? It's not really a Cast Segment if it's not. going a cast.

```suggestion
            ),
            min_times=1,
```

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1075,7 +1075,19 @@ class ShorthandCastSegment(BaseSegment):
...
+        ),
+        allow_gaps=True,
```
tunetheweb, line 1090: Again true by default so don't think we need this do we?

```suggestion
```

src/sqlfluff/dialects/dialect_ansi.py
bolajiwahab: Okay. I didn't know it's the default.

src/sqlfluff/dialects/dialect_ansi.py
bolajiwahab: What does min_times do? Maybe a reference I can check?

bolajiwahab: Okay. Thanks for the explanation.

PR #3753, opened 2022-08-16
When ignore=templating and fix_even_unparsable=True, provide defaults for missing vars

### LINKED ISSUE #3752: Jinja templater crashes and halts linting on undefined variable, even though ignore=templating is set
### Search before asking

### What Happened

I ran `sqlfluff fix` on a file with `ignore=templating` and "fix even unparsable". SQLFluff reported no fixable linting violations.

### Expected Behaviour

Since I specified `ignore=templating` and "fix even unparsable", I expected SQLFluff to fix the file the best it could.

### Observed Behaviour

This is the output I saw from the `fix` command:
```
(sqlfluff-3.9.11) m1 ➜  sqlfluff (main) ✗ sqlfluff fix test.sql --dialect bigquery --ignore=templating --FIX-EVEN-UNPARSABLE
All Finished 📜 🎉!
```

I ran `parse` on the same file, and this gives some insight what happened:
```
(sqlfluff-3.9.11) m1 ➜  sqlfluff (main) ✗ sqlfluff parse --dialect bigquery --ignore=templating test.sql 
...Failed to Parse...
==== parsing violations ====
L:   1 | P:   1 |  TMP | Unrecoverable failure in Jinja templating: 'Undefined' object is
                       | not iterable. Have you configured your variables?
WARNING: Parsing errors found and dialect is set to 'bigquery'. Have you configured your dialect correctly?
```

### How to reproduce

Use this `test.sql` file:
```
select * except(rnk) from
  (
    select tbl1.*,
    row_number() over (partition by tbl1.the_name,  {{ context_columns | join(', ') }} order by created_at desc) rnk
    from tbl1
    inner join tbl2
    on tbl1.the_name = tbl2.the_name
    and tbl1.run_id = tbl2.run_id
    where run_rnk = 1
  )
where rnk = 1
```

Run this command:
```
sqlfluff fix test.sql --dialect bigquery --ignore=templating --FIX-EVEN-UNPARSABLE
```

### Dialect

BigQuery

### Version

1.2.1, also the latest `main` (commit `f2ddd3a702c7ed7520733426f63cd368327e209a`)

### Configuration

n/a

### Code of Conduct

barrywhart:
Fixes #3752 

### Pull Request checklist

  - Other.


src/sqlfluff/core/templaters/jinja.py
```diff
@@ -393,7 +393,15 @@ def __getattr__(self, item):
...
-                live_context[val] = Undefined(name=val)
+                if not config or "templating" not in config.get("ignore"):
+                    live_context[val] = Undefined(name=val)
+                else:
+                    # When ignoring=templating is configured, provide a
+                    # "reasonable" default for undefined variables, rather than
+                    # trying to force them to fail. This won't always work
+                    # (i.e. there's no "universal default" value that will) work
+                    # in all cases, but this will probably work 90% of the time.
+                    live_context[val] = "a"
```
tunetheweb, line 404: Should we emit an error or warning here to make it clear we're doing this? Even if it's just in debug mode.

barrywhart: Sure, I'll add that.

barrywhart: Done!

src/sqlfluff/core/config.py
```diff
@@ -542,7 +542,7 @@ def __getstate__(self):
         del state["_plugin_manager"]
         return state
 
-    def __setstate__(self, state):
+    def __setstate__(self, state):  # pragma: no cover
```
barrywhart, line 545: AFAIK, this is unrelated to the rest of the PR, but it was coming up in the PR build as "not covered". 🤷‍♀️

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -383,28 +403,84 @@ def process(
 
         undefined_variables = set()
 
-        class Undefined:
```
barrywhart, line 386: I suggest reviewing this section with "Hide whitespace" enabled.

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -383,28 +403,84 @@ def process(
...
-            """Similar to jinja2.StrictUndefined, but remembers, not fails."""
+        if "templating" not in config.get("ignore"):
```
barrywhart, line 406: I suggest reviewing this section with "Hide whitespace" enabled. We retain the old version of `class Undefined`, and we add a new implementation for the `ignore=templating` case.

tunetheweb (approved):

src/sqlfluff/core/config.py
tunetheweb: FYI I got this, then it fixed itself while I was digging in codecov. So I suspect it might not need it again. But could be wrong!

```suggestion
    def __setstate__(self, state):
```

AydarDD:
@barrywhart  Hi, it's a great feature. Do we have any reasons why it does not work for the default python templating?

barrywhart:
There's no technical reason it couldn't. My assumption was that supplying values automatically for undefined variables is quite "aggressive" and potentially results I surprising, confusing, even undesirable. I also think it could result in a lot more support questions.

I'd strongly prefer people actually define their variables.

AydarDD:
@barrywhart Yes but why is it implemented only for Jinja? I think this feature should be available for all templating engines. There are multiple use cases when we want to use the default template values, for example, we want to automate CI tests of SQL queries with custom variables. Not sure how collaboration works for this repo, can I implement and create a PR to do the same for the python templating engine?

barrywhart:
Sure, feel free. The current implementation is very specific to Jinja.

PR #3760, opened 2022-08-18
ReflowSequence 1: Scope + Respacing

alanmcruickshank:
This is a start on #946 / #3673.

It also looks a bit different from when originally drafted. Some of that functionality was merged separately in #3766.

What it does do:
- Brings in frameworks for reflow in the `utils` library.
  - This is mainly based around a `ReflowSequence` which contains alternating `ReflowBlock` elements (which are the things we can reflow - essentially the things we _can't_ change) and `ReflowPoint` elements (which are the things that we can edit during a reflow).
- Migrates `L001` to the new reflow sequence framework.

Next steps would be:
- Bring the other "spacing" rules into the same framework: `L005`, `L006`, `L008`
- Bring `L007` into the framework (relationship between newlines and operators).
- Bring `L009` into the framework (file should end with newline).

The hard rules are going to be `L019` (leading/trailing commas) and `L003` (indentation) - but I think once they're in it will be pretty easy to build a `reflow` api (or even cli) command.

### DIFF
--- a/src/sqlfluff/rules/L001.py
+++ b/src/sqlfluff/rules/L001.py
@@ -1,11 +1,12 @@
 """Implementation of Rule L001."""
-from sqlfluff.core.rules import BaseRule, LintResult, LintFix, RuleContext
-from sqlfluff.core.rules.crawlers import SegmentSeekerCrawler
+from typing import List
+from sqlfluff.core.rules import BaseRule, LintResult, RuleContext
+from sqlfluff.core.rules.crawlers import RootOnlyCrawler
 from sqlfluff.core.rules.doc_decorators import (
     document_fix_compatible,
     document_groups,
 )
-from sqlfluff.utils.functional import FunctionalContext, sp
+from sqlfluff.utils.reflow.classes import ReflowSequence
 
 
 @document_groups
@@ -36,27 +37,15 @@ class Rule_L001(BaseRule):
     """
 
     groups = ("all", "core")
-    crawl_behaviour = SegmentSeekerCrawler(
-        {"newline", "end_of_file"}, provide_raw_stack=True
-    )
+    crawl_behaviour = RootOnlyCrawler()
 
-    def _eval(self, context: RuleContext) -> LintResult:
+    def _eval(self, context: RuleContext) -> List[LintResult]:
         """Unnecessary trailing whitespace.
 
         Look for newline segments, and then evaluate what
         it was preceded by.
         """
-        if len(context.raw_stack) > 0 and context.raw_stack[-1].is_type("whitespace"):
-            # Look for a newline (or file end), which is preceded by whitespace
-            deletions = (
-                FunctionalContext(context)
-                .raw_stack.reversed()
-                .select(loop_while=sp.is_type("whitespace"))
-            )
-            # NOTE: The presence of a loop marker should prevent false
-            # flagging of newlines before jinja loop tags.
-            return LintResult(
-                anchor=deletions[-1],
-                fixes=[LintFix.delete(d) for d in deletions],
-            )
-        return LintResult()
+        sequence = ReflowSequence.from_root(context.segment)
+        fixes = sequence.trailing_whitespace_fixes()
+        results = [LintResult(anchor=fix.anchor, fixes=[fix]) for fix in fixes]
+        return results
--- a/src/sqlfluff/utils/reflow/README.md
+++ b/src/sqlfluff/utils/reflow/README.md
@@ -95,7 +95,7 @@ Some rules add, remove or rearrange segments and need to do so in
 a way which is reflow aware.
 - In general, rules which only _remove_ segments as part of a fix
   should only readdress the _Spacing_ left in the gap. We should
-  not attempt to reflow and attempt to remove any _Line Breaks_
+  not attempt to reflow and not attempt to remove any _Line Breaks_
   which could now be possible due to the remaining shorter lines.
   - This should use the `.respace()` method.
 - For rules which _move_ an element (and not just a movement with
--- a/src/sqlfluff/utils/reflow/README.md
+++ b/src/sqlfluff/utils/reflow/README.md
@@ -1,0 +1,176 @@
+# Reflow Primer
+
+Several SQLFluff rules are concerned with layout and the placement
+of segments with respect to whitespace and newlines. This is reflow.
+
+The benefit of centralising reflow routines is that interdependencies
+can be resolved and duplicate code can be reduced.
+
+The challenge of centralising reflow routines is that separating out
+the functionality to allow rules to _selectively_ apply element of
+that logic without defaulting to a full reflow.
+
+In the interests of separating this out, there are a few distinct
+elements of reflow logic:
+1. _Indentation_. This is the most obvious one we think of. This
+   is concerned _only with the amount of leading whitespace at the_
+   _start of lines_.
+2. _Line Length_. This is concerned only with the total amount of
+   characters on a line (including the amount of leading whitespace).
+3. _Spacing_. This is concerned only with the amount of whitespace
+   between elements on a line.
+4. _Line Breaks_. This is concerned only with where line breaks are
+   introduced within sequences of segments (usually, but not
+   exclusively without reference to changes in indentation). This
+   would (as an example) include the enforcement of trailing or
+   leading commas.
+
+It follows that each of these elements has dependencies on each other:
+1. _Spacing_ will influence the _Line Length_.
+2. Changes in _Indentation_ will influence _Line Length_.
+3. Removal of _Line Breaks_ will influence _Line Length_ and may
+   necessitate changes in _Spacing_.
+4. Addition of _Line Breaks_ must take account of _Indentation_.
+5. Resolving excessing _Line Length_ usually requires changes to
+  _Line Breaks_ and potentially _Indentation_.
+
+## Application in rules
+
+### What is part of reflow and what is not?
+
+In general there are two phases to a reflow related rule:
+1. _Linting_ - working out whether there's an issue.
+2. _Fixing_ - resolving the issue.
+
+The intent is that _Fixing_ should always be in the reflow code.
+_Linting_ may or may not be in the reflow code and could also be
+encoded within the rule itself.
+
+### How to apply in a rule context
+
+When applying reflow in a rule context it makes sense to consider
+how widespread the impact we want to have is:
+- A rule designed around _Spacing_ may act in relative isolation,
+  but may also optionally need to resolve excessive _Line Length_.
+- A rule designed around _Indentation_ may act simply on it's own
+  but may also optionally need to resolve excessive _Line Length_.
+- A rule designed around the placement or presence of _Line Breaks_
+  will probably have to consider _Spacing_ (or at least decide
+  what spacing to impose at least for the moved elements) and if
+  introducing additional _Line Breaks_ it may need to account
+  for _Indentation_.
+- Resolving _Line Length_ will involve introducing additional
+  _Line Breaks_ (assuming that _Spacing_ has already been covered).
+
+As a rule of thumb:
+- If it's a rule about _Spacing_ it should only care about that
+  in full knowledge that it may make lines too long. Unless the
+  user has _also_ enabled long line checking, this is fine.
+- If it's a rule about _Indentation_ we should also not check
+  line length (for the same reasons).
+- If it's a rule about _Line Breaks_ in relation to other segments
+  (e.g. around commas or operators), we should aim to keep the
+  existing _Indentation_ and not reset that or consider any
+  _Line Length_ issues. We **should** consider _Spacing_, but
+  only around the elements moved (e.g. if we're shifting a trailing
+  comma to be a leading one, then we should also correct it's spacing
+  at the same time).
+- If it's a rule about _Line Length_ we will also need to consider
+  _Indentation_ and _Line Breaks_.
+
+### How to apply in a file reflow context
+
+When reflowing a whole file we have a few considerations:
+- Ideally we have all of the compulsory _Line Breaks_ and as few
+  others as possible without going over our maximum _Line Length_,
+  assuming correct _Indentation_ and _Spacing_.
+- If additional _Line Breaks_ are required, we add them at the
+  higher levels in the parse tree first before adding them deeper.
+
+Therefore the order of operations should be:
+1. Coerce the entire file to compulsory _Line Breaks_ **only**.
+   i.e. Strip all non-compulsory ones, add in any missing
+   compulsory ones.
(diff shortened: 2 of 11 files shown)


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1526,11 +1526,16 @@ def iter_patches(self, templated_file: TemplatedFile) -> Iterator[FixPatch]:
         else:
             # This segment isn't a literal, but has changed, we need to go deeper.
 
+            # If there's an end of file segment or indent, ignore it.
+            segments = self.segments
+            if self.segments[-1].is_type("end_of_file", "indent"):
+                segments = segments[:-1]
+
```
OTooleMichael, line 1533: Not sure I fully understand, but does this risk having different Trees before and after fixes?

ala we fix one rule, it deletes the indent (without reason)
next rule picks up on a missing indent
produces a fix that it didn't need to

src/sqlfluff/rules/L001.py
```diff
@@ -1,11 +1,13 @@
...
+from typing import List
 from sqlfluff.core.rules import BaseRule, LintResult, LintFix, RuleContext
-from sqlfluff.core.rules.crawlers import SegmentSeekerCrawler
+from sqlfluff.core.rules.crawlers import RootOnlyCrawler
 from sqlfluff.utils.functional import sp, FunctionalContext
 from sqlfluff.core.rules.doc_decorators import (
     document_fix_compatible,
     document_groups,
 )
+from ..utils.reflow.classes import ReflowSequence
```
OTooleMichael, line 9: why not `sqlfluff.utils` ?

src/sqlfluff/rules/L003.py
```diff
@@ -75,7 +75,10 @@ def finalise(self, line_no: int, templated_file: Optional[TemplatedFile]):
         # Generate our final line summary based on the current state
         is_comment_line = all(
             seg.is_type(
-                "whitespace", "comment", "indent"  # dedent is a subtype of indent
+                "whitespace",
+                "comment",
+                "indent",
+                "end_of_file",  # dedent is a subtype of indent
```
OTooleMichael, line 81: Comment now in the wrong place

src/sqlfluff/core/parser/segments/base.py
OTooleMichael: Also could this ever produce an empty segments?

`assert segments, 'Segments Cannot be empty'`

src/sqlfluff/utils/reflow/classes.py
```diff
@@ -0,0 +1,283 @@
...
+        reflow_logger.debug("TW: Checking %s", self)
+        if not self.class_types.intersection({"newline", "end_of_file"}):
+            return []
+        fixes = []
+        for idx, seg in enumerate(self.segments):
+            if (
+                seg.is_type("newline", "end_of_file")
+                and idx > 0
+                and self.segments[idx - 1].is_type("whitespace")
+            ):
```
OTooleMichael, line 92:
```suggestion
            if not (
                seg.is_type("newline", "end_of_file")
                and idx > 0
                and self.segments[idx - 1].is_type("whitespace")
            ):
                continue
```
Invert and save the indentation

src/sqlfluff/utils/reflow/classes.py
```diff
@@ -0,0 +1,283 @@
...
+                # Check evens are all blocks
+                assert all(
+                    isinstance(elem, ReflowPoint) for elem in self.elements[1::2]
+                )
+        except AssertionError as err:
+            reflow_logger.exception(
+                "Assertion check on ReflowSequence failed. " "Elements: %s",
+                self.elements,
+            )
+            raise err
```
OTooleMichael, line 167: I'd be inclined to do this in a util function 
`def _validate_reflow_sequence(elements: Sequence[Union[ReflowBlock, ReflowPoint]], root_segment: BaseSegment)`

src/sqlfluff/utils/reflow/classes.py
```diff
@@ -0,0 +1,283 @@
...
+        seg_buff = []
+        SeqClass: Union[None, Type[ReflowBlock], Type[ReflowPoint]] = None
+        for seg in segments:
+            NextClass: Union[Type[ReflowBlock], Type[ReflowPoint]]
+            # NOTE: Can indents be in a reflow block? I assume not.
+            # Placeholders certainly not because they also get indented.
+            if seg.is_type("whitespace", "newline", "end_of_file"):
+                NextClass = ReflowPoint
+            else:
+                NextClass = ReflowBlock
```
OTooleMichael, line 188:
```suggestion
            NextClass: Union[Type[ReflowBlock], Type[ReflowPoint]] = (
               ReflowPoint if seg.is_type("whitespace", "newline", "end_of_file")
               else ReflowBlock
            )
            # NOTE: Can indents be in a reflow block? I assume not.
            # Placeholders certainly not because they also get indented.

```

src/sqlfluff/utils/reflow/classes.py
```diff
@@ -0,0 +1,283 @@
...
+            elif SeqClass is ReflowPoint and NextClass is ReflowPoint:
+                # Extend the buffer if we're still in s spacing section.
+                seg_buff.append(seg)
+            elif SeqClass is ReflowBlock and (
+                seg.is_type("indent") or all(s.is_type("indent") for s in seg_buff)
+            ):
+                # Extend the buffer if this is an indent
+                # OR the existing block is all indent.
+                seg_buff.append(seg)
+            else:
```
OTooleMichael, line 208: These if else block and the fall through (`SeqClass = NextClass`) are hard to read 
```suggestion
            if _should_extend_buffer(seg, SeqClass, NextClass):
                seg_buff.append(seg)
                SeqClass = NextClass
                continue

```

OTooleMichael (changes requested):
Looks Good, On a surface scan I only really comment on deep nesting as this caused quite some trouble to read in the past with rules like L003

src/sqlfluff/rules/L001.py
alanmcruickshank: That's my editor being "helpful" - you're right.

src/sqlfluff/utils/reflow/classes.py
alanmcruickshank: why? I'm not particularly against it, I'm just curious.

src/sqlfluff/utils/reflow/classes.py
alanmcruickshank: Yeah I agree - rather than make a new function though I'm going to keep the conditions in the function, but structure it more like what you've put here.

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: Ah - this doesn't _mutate_ segments, it just creates a temporary version of `self.segments` that it used for patch iteration. The actual child segments remain unaltered.

The purpose of this is to ensure that an insertion at the end of a file gets treated properly using the `"end_point"` code path rather than the presence of the new `end_of_file` marker forcing it down the `"mid_point"` path which leads to bugs.

If it results in an empty `segments` list then that won't cause any issues here. We just won't iterate through anything.

alanmcruickshank: I've added a note to explain that more clearly.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1526,11 +1526,18 @@ def iter_patches(self, templated_file: TemplatedFile) -> Iterator[FixPatch]:
         else:
             # This segment isn't a literal, but has changed, we need to go deeper.
 
+            # If there's an end of file segment or indent, ignore it just for the
+            # purposes of patch iteration.
+            # NOTE: This doesn't mutate the underlying `self.segments`.
+            segments = self.segments
+            if self.segments[-1].is_type("end_of_file", "indent"):
+                segments = segments[:-1]
```
barrywhart, line 1534: I suspect the answer is "no", but is it possible (and do we care) if the last *two* segments in the list are:
* indent
* end_of_file

src/sqlfluff/core/parser/lexer.py
```diff
@@ -560,6 +561,15 @@ def elements_to_segments(
                         )
                     )
 
+        # Add an end of file marker
+        segment_buffer.append(
```
barrywhart, line 565: Related: Can you take a look at this code:
```
 ack --python is_final src
src/sqlfluff/core/rules/base.py
420:        Similar in spirit to BaseRule.is_final_segment(), but:
768:    def is_final_segment(self, context: RuleContext, filter_meta: bool = True) -> bool:

src/sqlfluff/rules/L052.py
275:        if not self.is_final_segment(context):
```

Seems like the function should be eliminated (or rewritten?) with the introduction of an `EndOfFile` segment.

src/sqlfluff/utils/reflow/classes.py
```diff
@@ -0,0 +1,290 @@
...
+    """Class for keeping track of editable elements in reflow.
+
+    It holds segments which can be changed during a reflow operation
+    such as whitespace and newlines.
+
+    It holds no configuration and is influenced by the blocks either
+    side.
+    """
+
+    def trailing_whitespace_fixes(self) -> List[LintFix]:
```
barrywhart, line 75: As more rules are incorporated into this, will each one become a function in this class? Or a separate class?

Just thinking that some of the rules (e.g. L003) are pretty big and wondering how the new code will be organized to keep it manageable. We can save this question for later if you prefer.

src/sqlfluff/utils/reflow/classes.py
```diff
@@ -0,0 +1,290 @@
...
+    """Class for keeping track of elements in a reflow operation.
+
+    It is assumed that there will be alternating blocks and points
+    (even if some points have no segments). This is validated on
+    construction.
+
+    We assume points on each end because this is the case with a file.
+    """
+
+    elements: Sequence[Union[ReflowBlock, ReflowPoint]]
```
barrywhart, line 139: Since both `ReflowBlock` and `ReflowPoint` inherit from `_ReflowElement`, would it be reasonable to say:
```
Sequence[_ReflowElement]
```
?

src/sqlfluff/utils/reflow/classes.py
alanmcruickshank: The idea is that it will all be in one class. I might see if I can incorporate another rule into this before merging so that you can see what I mean.

I agree that it's going to get big, especially when we bring in L003 - code organisation is going to get really important.

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: hmmmm good question. It's not happened yet, but I guess in theory it probably is. I'll see if I can account for that.

src/sqlfluff/utils/reflow/classes.py
barrywhart: FYI, someone is starting to look at creating a new, flow-related behavior. Remember how people have requested that column names and types should align in columns? The idea is that this would be a new option for L039, since that rule already enforces related behavior.

src/sqlfluff/core/parser/segments/base.py
barrywhart: Shameless sales pitch: IIRC, the Segments class in the functional API has a `last()` function which makes this easy.

alanmcruickshank: Good idea 👍

alanmcruickshank: Hmmm - actually from here in the codebase I'd rather not add more external dependents. If this was in the reflow module I would totally be with you.

src/sqlfluff/core/parser/lexer.py
alanmcruickshank: I can't see a reference to `is_final_segment` in L052, but I did find one use of the function on line 420 in `L003`. I've updated the rule to use `end_of_file` and got rid of `RuleContext.final_segment`.

src/sqlfluff/utils/reflow/classes.py
alanmcruickshank: Actually - on reflection now I'm with you. It will be useful later.

alanmcruickshank:
This PR is much smaller now that #3766 is merged. I'm going to leave it as a draft and bring a couple more rules into it so we can start to see the consolidation before marking it as properly ready for review.

src/sqlfluff/utils/reflow/README.md
```diff
@@ -0,0 +1,219 @@
...
+    in practice this needs to be broader).
+
+### What about non-reflow rules?
+
+Some rules add, remove or rearrange segments and need to do so in
+a way which is reflow aware.
+- In general, rules which only _remove_ segments as part of a fix
+  should only readdress the _Spacing_ left in the gap. We should
+  not attempt to reflow and attempt to remove any _Line Breaks_
+  which could now be possible due to the remaining shorter lines.
```
barrywhart, line 99: Do you mean **never remove line breaks", or that it would happen elsewhere/later?

src/sqlfluff/utils/reflow/README.md
```diff
@@ -0,0 +1,219 @@
+# Reflow Primer
```
barrywhart, line 1: I don't see any mention of templating in this file. Will computing and applying reflow fixes work similarly to the current rules and core linting engine? I.e.:
* Some rules have additional logic to behave intelligently in or near templated code.
* The core linting engine prevents changing templated code or copying code from a templated region to literal code, or to other templated regions.

src/sqlfluff/utils/reflow/README.md
alanmcruickshank: Correct. This was lazy use of the word `not`. Will correct it.

src/sqlfluff/utils/reflow/README.md
alanmcruickshank: Kind of yes.

The core linting engine code around not changing templated code still applies :+1:.

However, I do think that with the new markers added in #3766 and the pre-existing `TemplatedSegment`, that we can remove any template specific code in rules. I might be wrong, but this is my hope.

src/sqlfluff/utils/reflow/classes.py
alanmcruickshank: I've updated this now - so that this logic (and a chunk more) is now included in the new `.respace()` method. Proof of concept which allows L001, L011 and L012 to use the same respacing logic is running through tests now. Once I've got the tests passing for those I hope to be able to bring a lot more rules into use that logic quite swiftly.

alanmcruickshank:
I'm going to close this PR and open a new one with the updated work done.

PR #3766, opened 2022-08-21
Add end_of_file and template_loop markers

alanmcruickshank:
This is a smaller version of #3760.

This introduces (I think) the last of the probably parse tree changes to facilitate the reflow work (#946 / #3673).

This introduces two new meta segments:
- An `EndOfFile` segment, to explicitly signal the end of the file. This would be seen by users in their parse tree as a new meta segment. This means eventually we can get rid of some of the functions which try and infer the end of a file - but just having a meta segment which explicitly states where the end of the file is.
- A `LoopJump` segment, to explicitly signal where a jump has been made and no placeholder has been added (usually in a for loop). This means we know that a control structure (a `{% for ... %}` tag or similar) would have been there. This is especially useful for indentation so we know not to remove indents before these elements.

### DIFF
--- a/src/sqlfluff/core/parser/lexer.py
+++ b/src/sqlfluff/core/parser/lexer.py
@@ -12,7 +12,7 @@ from sqlfluff.core.parser.segments import (
     TemplateSegment,
     UnlexableSegment,
     EndOfFile,
-    LoopJump,
+    TemplateLoop,
 )
 from sqlfluff.core.parser.markers import PositionMarker
 from sqlfluff.core.errors import SQLLexError
@@ -374,7 +374,7 @@ class Lexer:
                     "      Backward jump detected. Inserting Loop Marker"
                 )
                 segment_buffer.append(
-                    LoopJump(
+                    TemplateLoop(
                         pos_marker=PositionMarker.from_point(
                             last_source_slice.stop,
                             element.template_slice.start,
--- a/src/sqlfluff/core/parser/lexer.py
+++ b/src/sqlfluff/core/parser/lexer.py
@@ -11,6 +11,8 @@ from sqlfluff.core.parser.segments import (
     Dedent,
     TemplateSegment,
     UnlexableSegment,
+    EndOfFile,
+    LoopJump,
 )
 from sqlfluff.core.parser.markers import PositionMarker
 from sqlfluff.core.errors import SQLLexError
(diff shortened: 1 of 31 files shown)


barrywhart:
The code changes look good. Delaying approval until tests pass.

barrywhart (approved):
Looks good once the tests pass.

alanmcruickshank:
@barrywhart - any better ideas for what to call the segment? `loop_jump` was the best I could think of but I'm not 100% sold on it. Any better ideas?

src/sqlfluff/core/parser/lexer.py
```diff
@@ -358,6 +360,29 @@ def elements_to_segments(
...
+                last_source_slice
+                and last_source_slice.stop > source_slice.start
+                and last_source_slice != source_slice
+            ):
+                # If we have, insert a loop marker to reflect that.
+                lexer_logger.debug(
+                    "      Backward jump detected. Inserting Loop Marker"
+                )
+                segment_buffer.append(
+                    LoopJump(
```
barrywhart, line 377: Possible alternative names: 
* `BackwardsTemplateJump`
* `BackwardsTemplateBranch`

barrywhart: I like including the word `Template` because it provides some additional context. We don't do that for other template-related meta segments, but it seems more valuable here because "loop" seems more plausible as something in SQL (recursive CTE loop or stored procedure loop).

barrywhart: Question: Is it possible we'll need to do something similar for **forward** jumps?

alanmcruickshank: Forward jumps are fine - because that would mean a skipped section and we already have placeholder code in for them. 

The reason for adding the backward jump is that almost every other instance of templated code already has a templated marker for it - either an explicit placeholder or a skip or similar. It's the backward loops that were missing one.

I like your point on including `Template` somehow... but I also like including `Loop` because these should only exist in loop cases. I'm going to go with `TemplateLoop`.

Will update PR.

wong-codaio:
Can't wait for this to be released! (I've a couple custom rule ideas that'll benefit from the `end_of_file` segment!)

alanmcruickshank:
> Can't wait for this to be released! (I've a couple custom rule ideas that'll benefit from the `end_of_file` segment!)

@wong-codaio - it's out! Came out on 1.3.0: https://github.com/sqlfluff/sqlfluff/releases/tag/1.3.0

wong-codaio:
> @wong-codaio - it's out! Came out on 1.3.0: https://github.com/sqlfluff/sqlfluff/releases/tag/1.3.0

This is a game changer @alanmcruickshank!  I've already implemented two custom rules based on `end_of_file` :D Thank you thank you

PR #3770, opened 2022-08-22
Add a file size check in bytes

### LINKED ISSUE #3693: Large files take several minutes before large_file_skip_char_limit stops parsing them
### Search before asking

### Description

large_file_skip_char_limit can prevent sqlfluff from hanging potentially forever on very large files, however when I timed the linting of two 2MB DBCR files it took over 2 minutes.

I suspect the issue is the current order of operations, as it seems files are fully loaded into memory and parsed for sqlfluff commands before this large file check occurs. If large file checking occurred earlier processing could go much quicker. 

Current workaround: Manually adding large files to .sqlfluffignore

### Use case

As a user with large sql migrations I want my linting too to run quickly while properly ignoring overly large files.

### Dialect

N/A

### Code of Conduct

alanmcruickshank:
This resolves #3693.

Is adds another config flag (`large_file_skip_byte_limit`) which checks the size of the file in bytes before loading it. Very similar functionality to `large_file_skip_char_limit`.

I've kept the original config flag but changed the default so that it's not applied. If anyone is still using it - it will work but they'll get a deprecation warning, point them to the new flag which should be much more efficient. In a future release we can remove the old flag.

In the process I also rename `SQLTemplaterSkipFile` to `SQLFluffSkipFile` - because we're not just calling it from the templater anymore - and there's nothing particularly templater specific about it.

### DIFF
--- a/src/sqlfluff/core/templaters/base.py
+++ b/src/sqlfluff/core/templaters/base.py
@@ -5,7 +5,7 @@ from bisect import bisect_left
 from typing import Dict, Iterator, List, Tuple, Optional, NamedTuple, Iterable
 from sqlfluff.core.config import FluffConfig
 
-from sqlfluff.core.errors import SQLTemplaterSkipFile
+from sqlfluff.core.errors import SQLFluffSkipFile
 
 # Instantiate the templater logger
 templater_logger = logging.getLogger("sqlfluff.templater")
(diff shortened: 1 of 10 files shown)


src/sqlfluff/core/templaters/base.py
```diff
@@ -162,7 +168,7 @@ def __init__(
             for idx, tfs in enumerate(self.sliced_file):
                 if previous_slice:
                     if tfs.templated_slice.start != previous_slice.templated_slice.stop:
-                        raise SQLTemplaterSkipFile(  # pragma: no cover
+                        raise SQLFluffSkipFile(  # pragma: no cover
                             "Templated slices found to be non contigious. "
```
barrywhart, line 172:
```suggestion
                            "Templated slices found to be non-contiguous. "
```

src/sqlfluff/core/templaters/base.py
```diff
@@ -37,8 +37,14 @@ def _wrapped(
     ):
         if config:
             limit = config.get("large_file_skip_char_limit")
+            if limit:
+                templater_logger.warning(
+                    "The config value large_file_skip_char_limit was found set. "
+                    "This feature will be removed in a future release, please "
+                    "use the more efficient large_file_skip_byte_limit instead."
```
barrywhart, line 44:
```suggestion
                    "use the more efficient 'large_file_skip_byte_limit' instead."
```

barrywhart (commented):
A couple of small suggestions, looks good overall.

barrywhart (approved):

PR #3803, opened 2022-08-29
Allow the use of libraries in macro definitions

bjgbeelen:

I ran into an issue where I used a library (`adapter.something`) inside a macro, but it failed the sqlfluff linting. The fix is easier than the recursive macro pattern, because I just had to make sure to load the libraries in the context before continuing with the macros.

Extended the unit test to see that it also works now when the library is mentioned inside of a macro.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/templaters/jinja.py
+++ b/src/sqlfluff/core/templaters/jinja.py
@@ -286,6 +286,10 @@ class JinjaTemplater(PythonTemplater):
         live_context = super().get_context(fname=fname, config=config)
         # Apply dbt builtin functions if we're allowed.
         if config:
+            # first make libraries available in the context
+            # so they can be used by the macros too
+            live_context.update(self._extract_libraries_from_config(config=config))
+
             apply_dbt_builtins = config.get_section(
                 (self.templater_selector, self.name, "apply_dbt_builtins")
             )
@@ -316,7 +320,6 @@ class JinjaTemplater(PythonTemplater):
                 )
             )
 
-            live_context.update(self._extract_libraries_from_config(config=config))
         return live_context
 
     def template_builder(
--- a/test/fixtures/templater/jinja_j_libraries/jinja.sql
+++ b/test/fixtures/templater/jinja_j_libraries/jinja.sql
@@ -1,3 +1,3 @@
 SELECT 56
-FROM {{ foo.schema }}.{{ foo.table("xyz") }}
+FROM {{ foo.schema }}.{{ table_proxy("xyz") }}
 WHERE {{ bar.equals("x", 23) }}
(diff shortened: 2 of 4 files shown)


bjgbeelen:
Just for reference that I indeed ran tests and linting locally first:
image

test/fixtures/templater/jinja_j_libraries/jinja.sql
```diff
@@ -1,3 +1,3 @@
 SELECT 56
-FROM {{ foo.schema }}.{{ foo.table("xyz") }}
+FROM {{ foo.schema }}.{{ table_proxy("xyz") }}
```
barrywhart, line 2: Could we add a test rather than change an existing one? In effect, this seems like we're removing an existing test, which would not be good.

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -286,6 +286,10 @@ def get_context(self, fname=None, config=None, **kw) -> Dict:
         live_context = super().get_context(fname=fname, config=config)
         # Apply dbt builtin functions if we're allowed.
         if config:
+            # first make libraries available in the context
+            # so they can be used by the macros too
+            live_context.update(self._extract_libraries_from_config(config=config))
```
barrywhart, line 291: I was wondering if changing the order could cause issues for other users, e.g. if a library module depends on a macro. I don't _think_ that could happen, though, because the modules are just Python; they don't work with a Jinja2 `Environment`.

barrywhart (changes requested):

test/fixtures/templater/jinja_j_libraries/jinja.sql
bjgbeelen: I thought about that, but my reasoning was that both the usage of `{{ foo.schema }}` and `{{ foo.table("xyz") }}` is actually testing the same thing, so in a way one of the two is superfluous. One reference to the `foo` package should suffice in testing that the library is available for the template.

By altering one reference to test the macro case, we keep it concise but still complete. 

But, if you still don't agree with this reasoning, I am happy to add it as a separate test

src/sqlfluff/core/templaters/jinja.py
bjgbeelen: As I understand it the library definition do represent macro's, but as part of an isolated package. So I don't think there is an option where a library refers to a macro that exists outside of the scope of the library.

bjgbeelen: (thinking about `dbt_utils` here which is a library, but contains macros)

test/fixtures/templater/jinja_j_libraries/jinja.sql
barrywhart: I'd prefer a second test so we have a test using dots. Thismis a common reason for defining macros in Python. Tests are cheap, and this test also serves as a working example.

bjgbeelen:
HI @barrywhart I moved the extra test to a separate folder. Ran everything successfuly locally as well

barrywhart (approved):
Looks great!

PR #3808, opened 2022-08-29
Indentation fixes and rule logging improvements

alanmcruickshank:
Found this as part of working on #3760 and I think it makes sense to merge separately before that. This is related to #3673, and also solves a long standing weird test case which I think is actually a bug, but that we hadn't found a solution for.

There are a few logging improvements in this PR but the main thrust is the treatment of indentation fixes. Previously they were always resolved using `create_before`, which led to strange issues where some fixes would be applied and not others. Changing them to `replace` unless there really was no previous indentation - while also adding a special case to the fix safety checks seems to resolve this. Previously any edits to newly created segments adjoining templated sections would be deemed unsafe - this naturally included indentation added before them.

This also makes takes one instance in `L003` where we weren't using `._coerce_indent_to()` and makes it use that function too. This ensures it behaves the same as above and also lays the groundwork for future simplification.


src/sqlfluff/core/rules/base.py
```diff
@@ -320,6 +320,17 @@ def get_fix_slices(
...
                 slice(anchor_slice.stop - adjust_boundary, anchor_slice.stop + 1),
             ]
+        elif (
+            self.edit_type == "replace"
+            and self.anchor.pos_marker.source_slice.stop
+            == self.anchor.pos_marker.source_slice.start
+        ):
+            # We're editing something with zero size in the source. This means
+            # it likely _didn't exist_ in the source and so can be edited safely.
+            # We return and empty set because this edit doesn't touch anything
```
barrywhart, line 330:
```suggestion
            # We return an empty set because this edit doesn't touch anything
```

barrywhart (approved):
Great work!

I've tackled existing bugs in the other rules, but I'd been afraid to touch L003 for a while. 😅

PR #3814, opened 2022-08-30
Remove a chunk of name references

alanmcruickshank:
This is another step toward #3692.

It mostly concerns operators, which were a lot of the remaining cases.

The was a choice to make here:
- Take the values which used to sit in `name` and make them the `type`. This preserves the information but results in a lot of parse tree changes.
- Remove the values which used to sit in `name`.

Ultimately, I went with the latter. The names aren't used anywhere and I think seeing `binary_operator` is more useful in the `yaml` parse trees than the name of the operator - that's obvious from the text. e.g. If we only care about equals signs we can just look for `comparison_operator` where the `.raw_upper` is `=`.

This removes the last of the defined `name=` options on segments, next step is to start removing the machinery.


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -782,9 +782,18 @@ def match(
...
-                return MatchResult(
-                    (cls(segments=m.matched_segments),), m.unmatched_segments
-                )
+                try:
+                    return MatchResult(
+                        (cls(segments=m.matched_segments),), m.unmatched_segments
+                    )
+                except TypeError as err:  # pragma: no cover
+                    # This is an error to assist with debugging dialect design.
+                    # It's most likely that the match_grammar has been seg on
```
barrywhart, line 791:
```suggestion
                    # It's most likely that the match_grammar has been set on
```

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -782,9 +782,18 @@ def match(
...
+                    # a raw segment which shouldn't happen.
+                    raise TypeError(
+                        f"Error in instantiating {cls.__module__}.{cls.__name__}. Have "
+                        f"you defined a match_grammar on a RawSegment? : {str(err)}"
+                    )
```
barrywhart, line 796:
```suggestion
                    ) from err
```
This will preserve the original exception (which I think will appear in the stack backtrace).
https://stackoverflow.com/questions/24752395/python-raise-from-usage

I haven't tested this, so probably worth a quick manual test if you accept this suggestion.

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -228,6 +228,42 @@ class LiteralKeywordSegment(KeywordSegment):
...
+class ComparisonOperatorSegment(CodeSegment):
+    """An comparison operator segment.
+
+    Defined here for type inheritance.
+    """
+
+    type = "comparison_operator"
+
+
+class CompositeComparisonOperatorSegment(BaseSegment):
```
barrywhart, line 258: What does the word "composite" mean here?

barrywhart (approved):
Looks good! A couple of minor suggestions.

src/sqlfluff/dialects/dialect_ansi.py
alanmcruickshank: Just that it's a operator made up of multiple raw segments. The reason for having two versions here is to have one which derives from `RawSegment` and one which derives from `BaseSegment`. There's probably scope in future for condensing these into two and rationalising the parse structure, but this is the closest to current live usage.

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: I think I'm going to leave this out. In some ways the stack trace is actively unhelpful for this error - it will just have a big old trace through a load of parsing code (in a very nested way). I kind of wanted to actually _hide_ the stack trace but that seemed too hard.

PR #3819, opened 2022-08-31
Typed matching and ripping out the rest of .name

### LINKED ISSUE #3692: Enhancement: Deprecate BaseSegment.name in favour of extending types.

alanmcruickshank:
I think this closes #3692, the final stage of ripping out `.name`.

This attacks the meat of replacing `NamedParser`, with a slightly simpler (although pretty similar) `TypedParser`, shifting a little logic up into the lexer and then just matching those types in the parser. It aligns any functions like `get_name` and `is_name` onto `get_type` and `is_type`.

This triggers a small number of parsing changes - related to the parsing of the `like_operator` which is an exception to what I said in #3814 where I've let it have a more specific type than just `comparison_operator`, where matching the `raw` to detect it feels wasteful because of the flexible regex.


src/sqlfluff/core/parser/parsers.py
```diff
@@ -43,21 +41,28 @@ def is_optional(self) -> bool:
     def _is_first_match(self, segment: BaseSegment):
         """Does the segment provided match according to the current rules."""
 
-    def _make_match_from_first_result(self, segments: Tuple[BaseSegment, ...]):
+    def _make_match_from_first_result(self, segment: BaseSegment):
```
barrywhart, line 44: Nitpick: Since the function now takes a single segment rather than a tuple of segments, we may want to rename it (the `first_result` bit no longer makes sense).

src/sqlfluff/core/parser/parsers.py
```diff
@@ -43,21 +41,28 @@ def is_optional(self) -> bool:
...
             type=self.type,
-            name=self.name,
             **self.segment_kwargs,
         )
         # Return as a tuple
-        return MatchResult((new_seg,), segments[1:])
+        return new_seg
+
+    def _match_single(self, segment: BaseSegment):
+        # Is the first one already of this type?
```
barrywhart, line 60:
```suggestion
        # Is the segment already of this type?
```

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -114,11 +121,9 @@
         allow_trailing=True,
     ),
     QuestionMarkSegment=StringParser("?", SymbolSegment, type="question_mark"),
-    AtSignLiteralSegment=NamedParser(
+    AtSignLiteralSegment=TypedParser(
         "at_sign_literal",
         ansi.LiteralSegment,
-        type="at_sign_literal",
-        trim_chars=("@",),
```
barrywhart, line 121: I'm curious -- what was `trim_chars` doing, and how do we no longer need it?

src/sqlfluff/dialects/dialect_exasol.py
```diff
@@ -188,15 +208,11 @@
...
+    FunctionScriptTerminatorSegment=TypedParser(
         "function_script_terminator",
         SymbolSegment,
-        name="function_script_terminator",
-        type="function_script_terminator",
-    ),
-    WalrusOperatorSegment=NamedParser(
-        "walrus_operator", SymbolSegment, type="assignment_operator"
     ),
+    WalrusOperatorSegment=StringParser(":=", SymbolSegment, type="assignment_operator"),
```
barrywhart, line 215: Not related to your changes, but I wonder if we should name this `AssignmentOperatorSegment` for consistency with `type`.

src/sqlfluff/utils/functional/segment_predicates.py
```diff
@@ -29,27 +29,29 @@ def _(segment: BaseSegment):
...
+        return segment.is_type(*seg_type)
 
     return _
 
 
 def is_keyword(*keyword_name) -> Callable[[BaseSegment], bool]:
     """Returns a function that determines if it's a matching keyword."""
-    return and_(is_type("keyword"), is_name(*keyword_name))
+    return and_(
+        is_type("keyword"), raw_upper_is(*[raw.upper() for raw in keyword_name])
```
barrywhart, line 53: Nitpick: For performance, I wonder if we'd want to `assert` that the strings in `keyword_name` are already uppercase (to avoid calling `upper()`)

barrywhart (approved):
A few minor suggestions, but looks 💯 overall.

src/sqlfluff/dialects/dialect_bigquery.py
alanmcruickshank: Ah - we still need it - it's just defined further up now. Given the segment in the lexer is of type `at_sign_literal` and of the class `LiteralSegment` - it won't mutate and these values wouldn't be applied anyway if set.

src/sqlfluff/dialects/dialect_exasol.py
alanmcruickshank: I see your point - but if I look at other dialects, they have other `assignment_operator` segments beyond just the walrus, so I think the current name is appropriate. TSQL seems to have _loads_.

src/sqlfluff/utils/functional/segment_predicates.py
alanmcruickshank: Good idea :+1:

src/sqlfluff/dialects/dialect_exasol.py
barrywhart: No worries, I'm not, as they say, _passionate_ about this!

src/sqlfluff/utils/functional/segment_predicates.py
alanmcruickshank: actually here - we're calling `.upper()` on the elements of `keyword_name`, which is going to be quite short relative to the segments. I think the performance hit here is minimal.

PR #3762, opened 2022-08-19
Issue 1582: Add align_alias configuration to L039 (unnecessary whitespace found)

mtsadler-branch:
**Goal**
fixes #1582 
  
--- 

**Changes**
- Added `align_alias` configuration, with default `False` (to favor existing behavior)
  
--- 
  
**Implementation:**
If `align_alias` is `True`, we loop over segments in `SelectClauseElementSegment`:
   - If `segment` is (`expression` or `column reference`) AND (is longest expression yet):
      - update max_len = expression.match_length  
   
  
Then we loop over segments in  `SelectClauseElementSegment` again:
   - If `segment` is (`expression` or `column reference`) AND (followed by WhiteSpaceSegment and AliasSegment):
      - pad expression's WhiteSpaceSegment with max_len - expression.match_length

**Relationship with Existing Rule:**
If we edited the segment's WhiteSpaceSegment, we should ignore it when removing unnecessary whitespace (in other words, if `align_aliases` and `_skip_aliases(context, seg)` then `continue`).

--- 

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.
   - Added test case for `align_alias=False` with extra spacing  
   - Added test case for `align_alias=True` with extra/uneven spacing

Issue #1582, opened 2021-10-10
Support aligning AS keyword in column aliases 


amardeep:
Tools such as Datagrip allow aligning AS keyword in column aliases. For eg., given input sql:
```sql
SELECT
  customer_id AS id,
  customer_name AS name,
  age
FROM customers
WHERE age < 60
```

it would reformat it to:
```sql
SELECT
  customer_id   AS id,
  customer_name AS name,
  age
FROM customers
WHERE age < 60
```

In most of our queries, we find this makes the file more readable.
It would be great if sqlfluff could support it to.

jpy-git:
Feels like this should be an option of L003, either align in multiples of indents from the left hand side of the page (current) or align on as (this proposal)

christopherridley:
Doesn't seem like L003 makes sense to me here. L003 is great to make sure the column names are aligned appropriately. L039 to ignore additional whitespace directly before the "AS" only between "SELECT" and "FROM" would be ideal here I believe. That would give people the flexibility to align the column aliases or not if they wish.

Olaktal:
+1 we align all our aliases and I think it makes SQL clearer 👌 
Any update on this feature ? Maybe it could be a configuration settings 🤔

RoyalTS:
Would love this (as an option to L039?) as well. Makes for much cleaner code!

christopherridley:
Seems like it doesn't really need to even be an option. I think if the logic is added to L039 to ignore additional whitespace directly before the "AS" only between "SELECT" and "FROM", it allows for people to do both styles. 

select
        company_id                      as company_id,
        deal_id                              as deal_id,
        _fivetran_synced             as _fivetran_synced

from source

or

select
        company_id as company_id,
        deal_id as deal_id,
        _fivetran_synced as _fivetran_synced

from source

moseleyi:
This is one of my main rules to do this when possible to increase the readability of the file!

PR #3824, opened 2022-09-01
Reflow centralisation 1: Scope + Respace

alanmcruickshank:
This is a tidier and more advanced version of #3760.

It makes progress on #946 & #3673.

The main element of this is a detail scope of how to achieve the aims of the reflow work. This is in the `README.md` file in the new `reflow` module. The tl;dr is:

> In the interests of separating this out, there are a few distinct elements of reflow logic:
> 1. _Indentation_. This is the most obvious one we think of. This is concerned _only with the amount of leading whitespace at the _start of lines_.
> 2. _Line Length_. This is concerned only with the total amount of characters on a line (including the amount of leading whitespace).
> 3. _Spacing_. This is concerned only with the amount of whitespace between elements on a line. **NB:** trailing whitespace is in this understanding, a special case of _Spacing_.
> 4. _Line Breaks_. This is concerned only with where line breaks are introduced within sequences of segments (usually, but not exclusively without reference to changes in indentation). This would (as an example) include the enforcement of trailing or leading commas.

This PR also starts a framework (based around a `ReflowSequence` class) for applying these operations. This is designed to be relatively idiomatic, and is modelled somewhat on the chaining approach offered by packages like `pandas`. As an illustration, here's a section of the refactored `L011`:

```python
...
elif self.aliasing != "implicit":
    self.logger.debug("Inserting AS keyword and respacing.")
    return LintResult(
        anchor=context.segment,
        # Work out the insertion and reflow fixes.
        fixes=ReflowSequence.from_around_target(
            context.segment.raw_segments[0],
            context.parent_stack[0],
            config=context.config,
            # Only reflow before, otherwise we catch too much.
            sides="before",
        )
        .insert(
            KeywordSegment("AS"),
            target=context.segment.raw_segments[0],
            pos="before",
        )
        .respace()
        .get_fixes(),
   )
```

The `ReflowSequence` keeps track of fixes which are generated at each step so that they can be combined intelligently and returned at the end of the transformations using `.get_fixes()`. Additionally the `.insert()`, `.without()` and `.replace()` methods also generate appropriate `LintFix` objects for the initial operations too at the same time as adjusting the `ReflowSequence` accordingly. This allows us to easily enrich that initial fix with the spacing fixes necessary.

So far, the only developed section is the elements to apply _Spacing_ (i.e. explicitly not attacking _Indentation_ or _Line Breaks_). In the interests of keeping PRs to a manageable size I'd propose considering merging this (or a more advanced version of it) before those elements are fully integrated.

Several rules have been migrated into the framework (reducing duplication, and reducing rule-specific code significantly):
- `L001` - trailing whitespace.
- `L005` - Commas should not have whitespace directly before them.
- `L006` - whitespace around operators.
- `L011` - implicit/explicit aliasing of tables.
- `L012` - implicit/explicit aliasing of columns (no code changes directly to `L012` but it inherits most from `L011`).
- `L015` - `DISTINCT` used with parentheses.
- `L023` - Single whitespace expected after `AS` in `WITH` clause.
- `L024` - Single whitespace expected after `USING` in `JOIN` clause (inherits from `L023`).
- `L039` - Unnecessary whitespace found.
  - Note that for this one, we needed to introduce the `DepthMap` class, which is primarily a look ahead to indentation. This class (and really the elements within it which are `DepthInfo` objects) hold the information on how deep the segment is within the tree and allow an interface to work out the common parents of two segments to obtain appropriate configuration for them. It's implemented here for the purposes of getting configuration to make sure that we strip whitespace _within_ operators, but not around them. e.g. `a   <   =  b` gets corrected to `a <= b` not `a<=b` or `a < = b`.
- `L048` - quoted literals surrounded by whitespace (used to inherit from `L006` for whitespace logic but now standalone).
- `L049` - comparisons with null should use `IS`/`IS NOT` (also used to inherit from `L006` for whitespace logic but now standalone).

Some rules do limited whitespace manipulation but I don't think do anything significant which would benefit from migration. Mostly this means limited insertion of a segment with appropriate whitespace around it. These are `L008`, `L017`, `L032`, `L033`, `L037`, `L043`, `L051`, 

I suggest that migrating `L052` is a sufficiently different exercise to mean that it should be done as a separate PR.

At this point I think the bones of the API are already in place and that there shouldn't be significant changes to the approach to support indentation rules and line breaks. The `DepthMap` object does the heavy lifting for allowing us to work out where line breaks should go - and a migration of the exsting `LineSummary` objects from `L003` will facilitate indentation.

Feedback welcome, both on the approach and also on whether this is an appropriate point to merge (either too much in one go, or not far enough yet).


pwildenhain:
Such cool work -- I can already tell how much simpler this is going to make adding new rules with fixes.

Shout out to think about including `L019` (comma placement) in this list. I know when I worked on it I ran into a lot of trouble re: line breaks and spacing

alanmcruickshank:
> Shout out to think about including `L019` (comma placement) in this list.

@pwildenhain - I do want to tackle `L019`, but I think the primary focus of that rule is about the positioning of _line breaks_ (similar to `L007` for operators) - which I think is probably going to be the hardest category of these rules. In my mind _spacing_ is the easiest (which are the rules highlighted above), then _indentation_ (aka `L003`) and then finally line breaks.

I also think if I wait until this is all done then it will take a long while, and I'd love to release this in incremental chunks. I think what I'm saying is: _I hear you - and soon - but not yet_.

barrywhart:
A bit related to @pwildenhain's question and the discussion about placement... this is probably a random suggestion, but a while back, I did some preliminary work with a package called `awesome-pattern-matching`. It seemed promising as a tool to help implement rules. A bit like the functional API, but more powerful. Think that could be useful for some of this work?

https://github.com/sqlfluff/sqlfluff/issues/2221

If you're writing lots of new code to recognize things to fix, it could be worth a look. See the linked issue -- it has several examples of rules that I refactored to use it. I found it nice to work with. I didn't do any performance testing, but it seemed well designed, and the author was responsive when I found an issue. (This was about a year ago.)

alanmcruickshank:
> I did some preliminary work with a package called `awesome-pattern-matching`. It seemed promising as a tool to help implement rules. A bit like the functional API, but more powerful. Think that could be useful for some of this work?

I think it's a still a promising angle - but I think it's orthogonal to what I'm doing here. I think whatever we do with reflow should work with `apm` or with the current parsing approach regardless.

On the bright side, centralising some of the whitespace code might make it easier to bring in `apm` (or similar) at a later date - just by having a more DRY codebase.

barrywhart:
No worries re: `apm`. I haven't looked closely at this code and thought combining rules into the reflow module might be requiring rework. 👍

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -0,0 +1,331 @@
...
+    The attributes exposed are designed to be "post configuration"
+    i.e. they should reflect configuration appropriately.
+
+    NOTE: These are the smallest unit of "work" within
+    the reflow methods, and may contain meta segments.
+    """
+
+    # Options for spacing rules are:
+    # - single: the default (one single space)
+    # - close: no whitespace
```
NiallRees, line 61: None or false would be much more explicit, let's stick to one of those?

alanmcruickshank: ah - there are going to be more than three unfortunately I think. If I can keep the number down I agree, but that might not be possible.

I'm working on how to handle comments effectively now and that might make things clearer.

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -0,0 +1,388 @@
...
+
+    NOTE: These are the smallest unit of "work" within
+    the reflow methods, and may contain meta segments.
+    """
+
+    # Options for spacing rules are:
+    # - single:         the default (one single space)
+    # - close:          no whitespace
+    # - close_inline:   no whitespace and no newlines
+    # - any:            don't enforce any spacing rules
```
NiallRees, line 59: This prevents reflow from being deterministic, right? This seems like a case where designing reflow for the existing rules could hold it back.

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -0,0 +1,388 @@
...
+    i.e. they should reflect configuration appropriately.
+
+    NOTE: These are the smallest unit of "work" within
+    the reflow methods, and may contain meta segments.
+    """
+
+    # Options for spacing rules are:
+    # - single:         the default (one single space)
+    # - close:          no whitespace
+    # - close_inline:   no whitespace and no newlines
```
NiallRees, line 58: How would it look if these were two configs, one for spacing, and one for newlines?

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: Not necessarily but perhaps. This certainly would _allow_ multiple situations in a rule context, but I'm thinking that when we get to full reflow we might do a bit of a reset, and at that point would could take a stricter line with comments. I.e. `any` would accept multiple options but it also implies that treating it as a `single` is acceptable.

I feel like there's a more elegant solution for comments that I can't see yet. This solution feels ok for now but maybe needs a rethink further down the line. Any better ideas?

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: So I think there _will_ be two configs. This is what I'm currently working with on my branch were I'm tackling newlines (as an example with commas):

```ini
[sqlfluff:layout:type:comma]
spacing_before = close
line_position = trailing
```

Which I think is what you're implying.

At the moment the only segments which use `close_inline` for their _spacing_ is `dot` - which maybe I should be able to deprecate later, but for now it was important to be able to _prevent_ newlines in that context. You're right though that this feels weird. I might review that part and see if I can get it to work without this option for the config.

Leave this one with me to see what other options might work. Worst case I don't think it's _awful_ if there's a little grey between the two config values, as long as their intent is different.

src/sqlfluff/utils/reflow/elements.py
NiallRees: Makes sense. I think as long as the non-any options are deterministic then it’s mission accomplished.

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: Ok - original `close_inline` was bad idea - agreed. I did that before I had the `spacing_within` config working, and that's a much better fit.

I've migrated over to that and so we now have:

```ini
[sqlfluff:layout:type:object_reference]
spacing_within = inline
```

rather than a spacing config tied to the `dot` as before. Does that feel cleaner @NiallRees ?

src/sqlfluff/rules/L005.py
```diff
@@ -41,20 +41,22 @@ class Rule_L005(BaseRule):
...
-            and context.segment.is_type("comma")
-            and anchor.is_type("whitespace")
-            and anchor.pos_marker.line_pos > 1
-        ):
-            return LintResult(anchor=anchor, fixes=[LintFix.delete(anchor)])
-        # Otherwise fine.
+        fixes = (
+            ReflowSequence.from_around_target(
+                context.segment,
+                context.parent_stack[0],
```
barrywhart, line 51: Nitpick: I notice the rules always pass the same 2nd and 3rd parameters:
```
context.parent_stack[0],
config=context.config,
```

Did you consider letting the rules just pass `context` and having it get `.parent_stack[0]` and `.config` from that?

src/sqlfluff/rules/L049.py
```diff
@@ -41,79 +42,60 @@ class Rule_L049(Rule_L006):
...
         """Relational operators should not be used to check for NULL values."""
         # Context/motivation for this rule:
         # https://news.ycombinator.com/item?id=28772289
         # https://stackoverflow.com/questions/9581745/sql-is-null-and-null
-        if len(context.segment.segments) <= 2:
-            return None  # pragma: no cover
 
         # Allow assignments in SET clauses
-        if context.parent_stack and context.parent_stack[-1].is_type(
+        if context.parent_stack and context.parent_stack[-2].is_type(
```
barrywhart, line 54: Is there a risk that `parent_stack[-2]` raises an `IndexError` (i.e. if there are less than 2 ancestors in the stack)?

src/sqlfluff/rules/L049.py
```diff
@@ -41,79 +42,60 @@ class Rule_L049(Rule_L006):
...
+            return None
+
+        # We only care about equality operators.
+        if context.segment.raw not in ("=", "!=", "<>"):
             return None
 
-        segment = FunctionalContext(context).segment
-        # Iterate through children of this segment looking for equals or "not
-        # equals". Once found, check if the next code segment is a NULL literal.
+        # We only care if it's followed by a null.
```
barrywhart, line 69: Seems a bit clearer...
```suggestion
        # We only care if it's followed by a NULL literal.
```

barrywhart (commented):
I've reviewed the first half of the PR, including all the changes to rule files. This looks really elegant! Just a few small-ish suggestions.

I'll try and review the rest of it later this weekend. 🙏

src/sqlfluff/core/default_config.cfg
```diff
@@ -56,6 +56,38 @@ indented_using_on = True
 indented_on_contents = True
 template_blocks_indent = True
 
+[sqlfluff:layout:type:comma]
+spacing_before = touch
```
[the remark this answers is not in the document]
alanmcruickshank: Check out the `ReflowPoint.respace_point()` function in `sqlfluff.utils.reflow.elements`. Most of the magic happens there.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -455,6 +455,24 @@ def raw_segments(self) -> List["RawSegment"]:
         """Returns a list of raw segments in this segment."""
         return self.get_raw_segments()
 
+    @cached_property
+    def raw_segments_with_stack(self) -> List[Tuple["RawSegment", List["BaseSegment"]]]:
+        """Returns a list of raw segments in this segment with the parents."""
```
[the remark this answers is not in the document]
alanmcruickshank: Good point. I'll see what I can do to make this clearer...

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -798,6 +816,7 @@ def _recalculate_caches(self):
             "raw_upper",
             "matched_length",
             "raw_segments",
+            "raw_segments_with_stack",
```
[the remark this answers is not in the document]
alanmcruickshank: I agree, but yeah I agree it's probably a separate PR. Fancy making an issue for it?

src/sqlfluff/core/rules/base.py
```diff
@@ -655,7 +658,7 @@ def matches_target_tuples(
         if seg.raw_upper in [
             elem[1] for elem in target_tuples if elem[0] == "raw_upper"
         ]:
-            return True
+            return True  # pragma: no cover
```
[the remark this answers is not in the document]
alanmcruickshank: Good idea - probably a separate PR I think though too. I think we _might_ be able to rip out quite a lot of `matches_target_tuples` now that the crawl behaviour PR is done.

src/sqlfluff/rules/L005.py
alanmcruickshank: I did, but I'd rather not pass around the whole context - it's quite powerful and it makes it a little harder to separate concerns, because you don't know which bits of the context they're using.

It was quite hard to unpick some of the older implementations for exactly that reason!

src/sqlfluff/rules/L049.py
alanmcruickshank: good point - I'll have a look.

src/sqlfluff/core/parser/segments/base.py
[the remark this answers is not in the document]
barrywhart: #3855

src/sqlfluff/utils/reflow/depthmap.py
```diff
@@ -0,0 +1,128 @@
...
+            stack_hash_set=set(stack_hashes),
+            stack_class_types=[seg.class_types for seg in stack],
+        )
+
+    def common_with(self, other: "DepthInfo") -> List[int]:
+        """Get the common depth and hashes with the other."""
+        # We use set intersection because it's faster and hashes should be unique.
+        common_hashes = self.stack_hash_set.intersection(other.stack_hashes)
+        common_depth = len(common_hashes)
+        return self.stack_hashes[:common_depth]
```
barrywhart, line 40: Is there an assumption here that `self` and `other` share a common prefix in terms of `stack_hashes`?

A doctest example, relevant assertion, or a bit more description could be helpful here. (Or can you think of a more descriptive function name?)

src/sqlfluff/utils/reflow/depthmap.py
```diff
@@ -0,0 +1,128 @@
...
+        return self.__class__(
+            stack_depth=self.stack_depth - amount,
+            stack_hashes=self.stack_hashes[:-amount],
+            stack_hash_set=self.stack_hash_set.difference(self.stack_hashes[-amount:]),
+            stack_class_types=self.stack_class_types[:-amount],
+        )
+
+
+class DepthMap:
+    """A sequence raw segments with depth and parent information.
```
barrywhart, line 56:
```suggestion
    """A sequence of raw segments with depth and parent information.
```
The class name mentions "map", but it's described as a sequence. That feels confusing to me. Is the description accurate? Also, is it describing more the external interface of the class or its implementation? (Hopefully the first.)

src/sqlfluff/utils/reflow/depthmap.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+@dataclass(frozen=True)
+class DepthInfo:
+    """An object to hold the depth information for a specific raw segment."""
+
+    stack_depth: int
+    stack_hashes: List[int]
+    # This is a convenience cache to speed up operations.
+    stack_hash_set: Set[int]
```
barrywhart, line 21: I notice `DepthMap` stores segment UUIDs, but here we're storing segment object hashes. Is there a reason for mixing approaches?

src/sqlfluff/utils/reflow/depthmap.py
```diff
@@ -0,0 +1,128 @@
...
+    def __init__(self, raws_with_stack: Sequence[Tuple[RawSegment, List[BaseSegment]]]):
+        # TODO: decide whether we need the raw segments?
+        # self.raw_segments = []
+        self.depth_info = {}
+        for raw, stack in raws_with_stack:
+            # self.raw_segments.append(raw)
+            self.depth_info[raw.uuid] = DepthInfo.from_raw_and_stack(raw, stack)
+
+    @classmethod
+    def from_parent(cls, parent: BaseSegment):
```
barrywhart, line 79: Does this work? (Found this pattern documented here: https://stackoverflow.com/questions/44640479/type-annotation-for-classmethod-returning-instance)
```suggestion
    def from_parent(cls: Type[T], parent: BaseSegment) -> T:
```

src/sqlfluff/utils/reflow/depthmap.py
```diff
@@ -0,0 +1,128 @@
...
+
+        NOTE: This is the most efficient way to construct a DepthMap
+        due to caching in the BaseSegment.
+        """
+        return cls(raws_with_stack=parent.raw_segments_with_stack)
+
+    @classmethod
+    def from_raws_and_root(
+        cls, raw_segments: Sequence[RawSegment], root_segment: BaseSegment
+    ):
```
barrywhart, line 90:
```suggestion
        cls: Type[T], raw_segments: Sequence[RawSegment], root_segment: BaseSegment
    ) -> T:
```

src/sqlfluff/utils/reflow/depthmap.py
```diff
@@ -0,0 +1,128 @@
...
+        `from_parent`.
+        """
+        buff = []
+        for raw in raw_segments:
+            stack = root_segment.path_to(raw)
+            # Don't include the final element of the stack.
+            buff.append((raw, stack[:-1]))
+        return cls(raws_with_stack=buff)
+
+    def get_depth_info(self, raw: RawSegment):
```
barrywhart, line 104:
```suggestion
    def get_depth_info(self, raw: RawSegment) -> DepthInfo:
```

src/sqlfluff/utils/reflow/depthmap.py
```diff
@@ -0,0 +1,128 @@
...
+                f"segment {raw} with UUID {raw.uuid}"
+            ) from err
+
+    def copy_depth_info(
+        self, anchor: RawSegment, new_segment: RawSegment, trim: int = 0
+    ):
+        """Copy the depth info for one segment and apply to another.
+
+        This mutates the existing depth map. That's ok because it's
+        an idempotent and uuids should be unique.
```
barrywhart, line 121:
```suggestion
        an idempotent operation and uuids should be unique.
```

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -0,0 +1,383 @@
...
+    # - any:            don't enforce any spacing rules
+    spacing_before: str
+    spacing_after: str
+    # The depth info is used in determining where to put line breaks.
+    depth_info: DepthInfo
+    # This stores relevant configs for segments in the stack.
+    stack_spacing_configs: Dict[int, str]
+
+    @classmethod
+    def from_config(cls, segments, config: ReflowConfig, depth_info: DepthInfo):
```
barrywhart, line 67:
```suggestion
    def from_config(cls: Type[T], segments, config: ReflowConfig, depth_info: DepthInfo) -> T:
```

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -0,0 +1,383 @@
...
+
+    It holds no configuration and is influenced by the blocks either
+    side.
+    """
+
+    def respace_point(
+        self,
+        prev_block: Optional[ReflowBlock] = None,
+        next_block: Optional[ReflowBlock] = None,
+        fixes: Optional[List[LintFix]] = None,
```
barrywhart, line 105: Nitpick: Is it useful to make `fixes` optional, since the list could be empty?

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -0,0 +1,383 @@
...
+            # Just check the most immediate parent for now for speed.
+            # TODO: Review whether this is enough.
+            # NOTE: spacing configs will be available on both sides if they're common
+            # so it doesn't matter whether we get it from prev_block or next_block.
+            within_constraint = prev_block.stack_spacing_configs.get(common[-1], None)
+            if not within_constraint:
+                pass
+            elif within_constraint in ("touch", "inline"):
+                # NOTE: inline is actually a more extreme version of "touch".
+                # i.e. inline, implies no spaces between either.
```
barrywhart, line 140: Can you explain this in more detail? Perhaps with an example?

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -0,0 +1,383 @@
...
+                if within_constraint == "inline":
+                    # If they are then strip newlines.
+                    strip_newlines = True
+                # If segments are expected to be touch within. Then modify
+                # constraints accordingly.
+                # NOTE: We don't override if it's already "any"
+                if pre_constraint != "any":
+                    pre_constraint = "touch"
+                if post_constraint != "any":
+                    post_constraint = "touch"
```
barrywhart, line 150: Random thought: Is this ordering going to occur other places? Wondering if it may be useful to define a class (perhaps an [IntEnum](https://docs.python.org/3/library/enum.html#enum.IntEnum)) that makes this more natural/reusable.

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -0,0 +1,383 @@
...
+            else:  # pragma: no cover
+                idx = prev_block.depth_info.stack_hashes.index(common[-1])
+                raise NotImplementedError(
+                    f"Unexpected within constraint: {within_constraint} for "
+                    f"{prev_block.depth_info.stack_class_types[idx]}"
+                )
+
+        reflow_logger.debug("Respacing: %s", self)
+        for idx, seg in enumerate(self.segments):
```
barrywhart, line 159: Nitpick: It doesn't appear that `idx` is used, thus the `enumerate()` could be removed.

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -0,0 +1,383 @@
...
+    """Class for keeping track of editable elements in reflow.
+
+    It holds segments which can be changed during a reflow operation
+    such as whitespace and newlines.
+
+    It holds no configuration and is influenced by the blocks either
+    side.
+    """
+
+    def respace_point(
```
barrywhart, line 101: Are there automated tests for this function? Would love to have some fairly "standalone" tests, as I've often found such tests useful with other lower-level components of the project.

Related: This function is pretty long. Maybe break it into smaller functions? (Seems good _eventually_, but it may make sense to wait until things have settled down more.)

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -0,0 +1,383 @@
...
+                segment_buffer.remove(ws)
+                new_fixes.append(LintFix("delete", ws))
+
+        # Is there a newline?
+        if self.class_types.intersection({"newline", "end_of_file"}):
+            # Most of this section should be handled as _Indentation_.
+            # BUT: There is one case we should handle here.
+            # If we find that the last whitespace has a newline
+            # before it, and the position markers imply there was
+            # a removal between them. Remove the whitespace.
```
barrywhart, line 203:
```suggestion
            # a removal between them, then remove the whitespace.
```

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -0,0 +1,383 @@
...
+                            raise ValueError(
+                                "Fixes detected, but none passed to .respace(). "
+                                "This will cause conflicts."
+                            )
+                        # Find the fix
+                        for fix in fixes:
+                            # Does it contain the insertion?
+                            # TODO: This feels ugly - eq for BaseSegment is different
+                            # to uuid matching for RawSegment. Perhaps this should be
+                            # more aligned. There might be a better way of doing this.
```
barrywhart, line 322: Would it be safe to use `is` here (checking for same exact object)?

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+
+from sqlfluff.utils.reflow.elements import ReflowElement, ReflowBlock, ReflowPoint
+
+# We're in the utils module, but users will expect reflow
+# logs to appear in the context of rules. Hence it's a subset
+# of the rules logger.
+reflow_logger = logging.getLogger("sqlfluff.rules.reflow")
+
+
+class ReflowSequence:
```
barrywhart, line 22: Nitpick: Could this be a dataclass? From the constructor, it seems that it could be.

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+        self.embodied_fixes: List[LintFix] = embodied_fixes or []
+
+    def get_fixes(self):
+        """Get the current fix buffer."""
+        return self.embodied_fixes
+
+    @staticmethod
+    def _validate_reflow_sequence(elements: Sequence[ReflowElement]):
+        assert elements, "ReflowSequence has empty elements."
+        # Check odds and events
```
barrywhart, line 60:
```suggestion
        # Check odds and evens
```

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+        return elem_buff
+
+    @classmethod
+    def from_raw_segments(
+        cls,
+        segments: Sequence[RawSegment],
+        root_segment: BaseSegment,
+        config: FluffConfig,
+        depth_map: Optional[DepthMap] = None,
+    ):
```
barrywhart, line 121:
```suggestion
        cls: Type[T],
        segments: Sequence[RawSegment],
        root_segment: BaseSegment,
        config: FluffConfig,
        depth_map: Optional[DepthMap] = None,
    ) -> T:
```

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+                # map should be constructed elsewhere and then passed in.
+                depth_map=depth_map,
+            ),
+            root_segment=root_segment,
+            reflow_config=reflow_config,
+            depth_map=depth_map,
+        )
+
+    @classmethod
+    def from_root(cls, root_segment: BaseSegment, config: FluffConfig):
```
barrywhart, line 144:
```suggestion
    def from_root(cls: Type[T], root_segment: BaseSegment, config: FluffConfig) -> T:
```

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+        )
+
+    @classmethod
+    def from_around_target(
+        cls,
+        target_segment: BaseSegment,
+        root_segment: BaseSegment,
+        config: FluffConfig,
+        sides: str = "both",
+    ):
```
barrywhart, line 161:
```suggestion
        cls: Type[T],
        target_segment: BaseSegment,
        root_segment: BaseSegment,
        config: FluffConfig,
        sides: str = "both",
    ) -> T:
```

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+                "Not expected removal of whitespace in ReflowSequence."
+            )
+        merged_point = ReflowPoint(
+            segments=list(self.elements[removal_idx - 1].segments)
+            + list(self.elements[removal_idx + 1].segments),
+        )
+        return ReflowSequence(
+            elements=list(self.elements[: removal_idx - 1])
+            + [merged_point]
+            + list(self.elements[removal_idx + 2 :]),
```
barrywhart, line 249: Nitpick: Are the two calls to `list()` call necessary here? Slicing a list already creates a new list.

https://discuss.python.org/t/does-the-list-slicing-produce-a-new-list/14627

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+            root_segment=self.root_segment,
+            reflow_config=self.reflow_config,
+            depth_map=self.depth_map,
+            # Generate the fix to do the removal.
+            embodied_fixes=[LintFix.delete(target)],
+        )
+
+    def insert(
+        self, insertion: RawSegment, target: RawSegment, pos="before"
```
barrywhart, line 258:
```suggestion
        self, insertion: RawSegment, target: RawSegment, pos: str="before"
```

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+        target_idx = self._find_element_idx_with(target)
+        # Are we trying to insert something whitespace-like?
+        if insertion.is_type("whitespace", "indent", "newline"):  # pragma: no cover
+            raise ValueError(
+                "ReflowSequence.insert() does not support direct insertion of "
+                "spacing elements such as whitespace or newlines"
+            )
+
+        # We're inserting something blocky. That means a new block AND a new point.
+        # It's possible we try to _split_ a point by targetting a whitespace element
```
barrywhart, line 275:
```suggestion
        # It's possible we try to _split_ a point by targeting a whitespace element
```

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+        )
+        if isinstance(self.elements[target_idx], ReflowPoint):
+            raise NotImplementedError(  # pragma: no cover
+                "Can't insert relative to whitespace for now."
+            )
+        elif pos == "before":
+            return ReflowSequence(
+                elements=list(self.elements[:target_idx])
+                + [new_block, ReflowPoint([])]
+                + list(self.elements[target_idx:]),
```
barrywhart, line 294: Same question as earlier, whether the `list()` calls are needed.

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+                embodied_fixes=[LintFix.create_before(target, [insertion])],
+            )
+        elif pos == "after":  # pragma: no cover
+            # TODO: This doesn't get coverage - should it even exist?
+            # Re-evaluate whether this code path is ever taken once more rules use
+            # this.
+            return ReflowSequence(
+                elements=list(self.elements[: target_idx + 1])
+                + [ReflowPoint([]), new_block]
+                + list(self.elements[target_idx + 1 :]),
```
barrywhart, line 308: Same question about `list()` calls.

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+                embodied_fixes=[LintFix.create_after(target, [insertion])],
+            )
+        raise ValueError(
+            f"Unexpected value for ReflowSequence.insert(pos): {pos}"
+        )  # pragma: no cover
+
+    def replace(
+        self, target: BaseSegment, edit: Sequence[BaseSegment]
+    ) -> "ReflowSequence":
+        """Returns a new reflow sequence elements replaced."""
```
barrywhart, line 322:
```suggestion
        """Returns a new reflow sequence with `edit` elements replaced."""
```

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+            # NOTE: if target raws has more than one segment we take the depth info
+            # of the first one. We trim to avoid including the implications of removed
+            # "container" segments.
+            self.depth_map.copy_depth_info(target_raws[0], edit_raw, trim=trim_amount)
+
+        # It's much easier to just totally reconstruct the sequence rather
+        # than do surgery on the elements.
+
+        # TODO: The surgery is actually a good idea for long sequences now that
+        # we have the handle the depth map.
```
barrywhart, line 347:
```suggestion
        # we have the depth map.
```

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -0,0 +1,450 @@
...
+            if isinstance(elem, ReflowPoint):
+                pre = None
+                post = None
+                if idx > 0:
+                    pre = cast(ReflowBlock, self.elements[idx - 1])
+                if idx < len(self.elements) - 1:
+                    post = cast(ReflowBlock, self.elements[idx + 1])
+                yield elem, pre, post
+
+    def respace(self, strip_newlines=False, filter="all") -> "ReflowSequence":
```
barrywhart, line 382:
```suggestion
    def respace(self, strip_newlines: bool=False, filter: str="all") -> "ReflowSequence":
```

src/sqlfluff/utils/reflow/README.md
```diff
@@ -0,0 +1,219 @@
...
+## Implications for configuration
+
+To be able to both reflow whole files with knowledge of appropriate
+_Spacing_ and _Line Breaks_, while also being able to lint effectively
+in a rules context - we need to have a common way of defining what
+our intended setup should be. On a very simple level, the understanding
+that commas usually have whitespace after them, but not before, or that
+bracket characters usually have whitespace outside them but not directly
+inside them. On a more complex level, that we might want to configure
+the precedence of operators in evaluating where to insert _Line Breaks_
```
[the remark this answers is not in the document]
alanmcruickshank: Exactly. This is my hope that we can get to this. I'll add this comment to the parent issue.

src/sqlfluff/utils/reflow/depthmap.py
alanmcruickshank: Yes there is, although I should probably make that more explicit. In theory they should at least share a `FileSegment` parent.

src/sqlfluff/utils/reflow/depthmap.py
alanmcruickshank: This is an awkward corner of `Segment` comparison, and I don't really like it - but it felt like too much to change it as part of this PR.

`RawSegment` classes (and their decendants) have uuids for comparison.
`BaseSegment` classes (and their decendants) have hashes for comparison.

The reasons behind that are quite convoluted (and maybe need a rethink at some point).

src/sqlfluff/utils/reflow/depthmap.py
barrywhart: Cool. My motivation for this comment is 90% about making the intent easier for humans to understand, and 10% about catching possible future errors.

src/sqlfluff/utils/reflow/depthmap.py
barrywhart: Thanks for the explanation! As I read farther in the PR, I saw you had a comment or two relating to this. Definitely not a blocker, just wanted to understand better. Perhaps someday, every segment could have a UUID. Anyway, not in scope for this work!

src/sqlfluff/utils/reflow/depthmap.py
```diff
@@ -0,0 +1,128 @@
...
+
+reflow_logger = logging.getLogger("sqlfluff.rules.reflow")
+
+
+@dataclass(frozen=True)
+class DepthInfo:
+    """An object to hold the depth information for a specific raw segment."""
+
+    stack_depth: int
+    stack_hashes: List[int]
```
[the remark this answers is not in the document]
alanmcruickshank: yeah that's probably a good idea 👍

src/sqlfluff/utils/reflow/depthmap.py
alanmcruickshank: The description _isn't_ accurate. It used to be but I got rid of that logic and didn't change the docstring. Good catch. I'll update it 👍 .

src/sqlfluff/utils/reflow/depthmap.py
alanmcruickshank: it doesn't, but this does:
```python
def from_parent(cls: Type["DepthMap"], parent: BaseSegment) -> "DepthMap":
   ....
```

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: Good catch!

src/sqlfluff/utils/reflow/sequence.py
alanmcruickshank: originally it was, but running the validation step was awkward. I migrated _away_ from it being dataclass for that reason.

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: Yeah - it should be split apart, and those parts should have some more sensible low level tests. I'm going to think about how to approach this. 🤔

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: Unfortunately not - we're not checking for _exactly_ the same segment, because that includes the location which might change when fixes are applied. We're checking for a segment which has the same signature rather than the same object.

src/sqlfluff/utils/reflow/sequence.py
alanmcruickshank: I did it this way because they're necessary for mypy type checking, but I think you're right - will find a different route.

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: Yeah I reckon we can do that.

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: Yeah, I did think about how to make it more reusable. I had a look at how to do the `Enum` route, but I'm not sure it's actually simpler. I agree it makes the ordering a little more obvious, but I think the precedence rules aren't simple enough for that to be enough.

Maybe when the logic has settled down a bit it would be a good option in future.

src/sqlfluff/core/rules/base.py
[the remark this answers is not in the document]
alanmcruickshank: I've made an issue for this idea: https://github.com/sqlfluff/sqlfluff/issues/3873

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: Good idea. I'll add a small example here, but I think your other note about more specific tests gets closer to the heart of it.

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: I've added some explicit test cases for `ReflowPoint.respace_point()` and `ReflowSequence.respace()` (see `respace_test.py`), and I've split out a lot of the logic in the bigger functions into smaller methods, however: getting sensible test harnesses around those smaller functions feels like quite a big job, as each one relies on quite a lot of config under the hood.

See what I've done and whether you think it's clear enough?

alanmcruickshank:
@barrywhart - I'd love a second review if you're able. I've addressed a lot of the smaller comments you had, and there are a few bigger changes of note:
1. I think I've brought all the relevant logic in from #3762 , including the test cases (and a few extras of my own work). The logic I've implemented is a lot more generic and so already supports a couple of different variations as you'll see in the code.
2. The big logic function are now split apart into smaller more manageable chunks. Each is a much better size now. I've also put more explicit tests around the public functions (`ReflowSequence.respace()` and the function it relies on: `ReflowPoint.respace_point()`). I don't have units tests around the smaller _private_ functions which support these two. Partly because the test harnesses would be _huge_ and partly because I wonder if the flexibility is useful for now.
3. I haven't changed anything in the space of documentation or commentary - however I do wonder what the right level of documentation is for this feature. The removed configs PR (#3874) helps a little, but I wonder what the right intro withing the main sqlfluff docs would be - and whether that needs to be in this PR or whether that can/should be separate.

Would love to see what you think and see whether you think this might be mergable?

test/fixtures/rules/std_rule_cases/L039.yml
```diff
@@ -112,3 +112,123 @@ test_redshift_at_time_zone:
...
+  fix_str: |
+    SELECT
+        a           AS first_column,
+        b           AS second_column,
+        (a + b) / 2 AS third_column
+    FROM foo AS bar
+  configs: &align_alias
+    layout:
+      type:
+        alias_expression:
```
barrywhart, line 154: Where's the corresponding logic for aligning aliases? Not sure where to find it after it was moved from the other PR.

test/utils/reflow/respace_test.py
```diff
@@ -0,0 +1,107 @@
...
+        new_seq = seq.respace(**kwargs)
+
+    assert new_seq.get_raw() == raw_sql_out
+
+
+@pytest.mark.parametrize(
+    "raw_sql_in,point_idx,kwargs,raw_point_sql_out,fixes_out",
+    [
+        # Basic cases
+        ("select    1", 1, {}, " ", {("replace", "    ")}),
```
barrywhart, line 57: For `replace`, should there be another value? (I.e. we're replacing X with Y.)

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -0,0 +1,606 @@
...
+
+        desired_space = " " * (
+            1 + max_desired_line_pos - whitespace_seg.pos_marker.working_line_pos
+        )
+        reflow_logger.debug(
+            "    desired_space: %r (based on max line pos of %s)",
+            desired_space,
+            max_desired_line_pos,
+        )
+        return desired_space
```
mtsadler-branch, line 289: This looks like the main chunk of code for identifying `align_aliases` violations

 (and then  `_handle_respace__inline_with_space` applies the fixes)

I dont have any concerns with this code.. 


This seems more reliable than the PR i put together, because you're actually checking `is_code()` which is an extra check to ensure `this segment has no subsegments`

My only concern is that the new default configs arent mentioned in `rules/config_info.py` ([example](https://github.com/sqlfluff/sqlfluff/pull/3762/files#diff-1fdcdfc3c128b96b64ddafed38baa65954da0cb213e7d72833307e5d7d9faaa5R37))

mtsadler-branch: I saw the `Reflow README` mentioned a new `layout` type of config, so wasnt sure if there was some default configuration you were doing somewhere else

barrywhart (approved):
Looks good to me!

test/fixtures/rules/std_rule_cases/L039.yml
alanmcruickshank: as we discussed via slack (and below) the logic is in ` _determine_aligned_inline_spacing()`

test/utils/reflow/respace_test.py
alanmcruickshank: Yes there is, but that's kind of tested using the `raw_point_sql_out` variable. The `fixes_out` option is much more just the check that the right kind and volume of fixes are generated.

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: Yeah - I think we totally need better documentation for this. I've got that as a separate task for the master issue (#3673).

Thanks for the review!

FabiAngel:
Hey guys I'm interested in https://github.com/sqlfluff/sqlfluff/issues/1582. As I understand correctly, that should have been solved by this issue, but I can't find the solution here nor in the sqlfluff docs. It's my first time here, so probably overseeing something. Can someone help?

Thanks!

Jaxsss:
> Hey guys I'm interested in #1582. As I understand correctly, that should have been solved by this issue, but I can't find the solution here nor in the sqlfluff docs. It's my first time here, so probably overseeing something. Can someone help?
> 
> Thanks!

Hey, I've also struggled searching the solution, I tried making my own rule copied from the solution from this issue, but I had to modify the entire library just like contributors do and it didn't work 100%. However after some 2 hours later I found the desired function in the [documentation](https://docs.sqlfluff.com/en/stable/layout.html#aligned-elements). It works just like I wanted it to work.

PR #3825, opened 2022-09-02
Fixed broken main branch

alanmcruickshank:
I broke the main branch through poor git hygiene 🤦 .

This PR is to fix it before I cause more damage.


barrywhart:
What's this PR about? I see two commits that cancel out?

alanmcruickshank:
> What's this PR about? I see two commits that cancel out?

I accidentally pushed to `main` and now the tests are failing. This PR is to test whether I fixed it.

greg-finley:
We should try to remember to remove these commits from the CHANGELOG when we cut a new release.

PR #3834, opened 2022-09-08
Snowflake: Fix syntax errors in tests

JamesRTaylor:
Fix a couple of minor syntax errors in Snowflake tests

No side effects

### Pull Request checklist
- [ x] Please confirm you have completed any of the necessary steps below.
I will update the yaml file in a second pass based on errors reported by unit tests. I'm not able to run the script locally (on macOS) to update the yaml files, so I'll do it by hand.


JamesRTaylor:
Found a few more syntax errors in the Snowflake tests, @barrywhart. Here are the fixes.

barrywhart (approved):
Thanks!

PR #3847, opened 2022-09-13
Reflow centralisation 2: Rebreak (operators & commas)

### LINKED ISSUE #3740: Rule L007 - Unexpected behavior in WHERE clause
### Search before asking

### What Happened

I encountered unexpected results with rule L007, particularly in where clauses and boolean operators and parentheses.

Example Input used
```
SELECT
    *
FROM
   foo
WHERE
    (g > z) AND
    (
      (a = 'a' AND b = 'b') OR
      (a = 'a' AND b = 'e') OR
      (a = 'c' AND b = 'g') OR
      (a = 'c' AND b = 'e') OR
      1 = 1
    );
```

### Expected Behaviour

I expect all boolean operators to be on start of the newline. 

```
  SELECT
      *
  FROM
     foo
  WHERE
      (g > z)
      AND (
        (a = 'a' AND b = 'b')
        OR (a = 'a' AND b = 'e')
        OR (a = 'c' AND b = 'g')
        OR  (a = 'c' AND b = 'e')
        OR 1 = 1
      );
```

### Observed Behaviour

As you can see below, when there are parens/parentheses involved, some SQLFluff allows boolean operators to remain prior to a newline.

Fix from SQLFluff:

```
SELECT *
FROM
    foo
WHERE
    (g > z)
    AND (
        (a = 'a' AND b = 'b')
        OR (a = 'a' AND b = 'e') OR
        (a = 'c' AND b = 'g') OR
        (a = 'c' AND b = 'e') OR
        1 = 1
    );
```

### How to reproduce

Have SQL fluff attempt to fix the input (using Python API):

```
SELECT
    *
FROM
   foo
WHERE
    (g > z) AND
    (
      (a = 'a' AND b = 'b') OR
      (a = 'a' AND b = 'e') OR
      (a = 'c' AND b = 'g') OR
      (a = 'c' AND b = 'e') OR
      1 = 1
    );
```

```
sqlfluff.fix(query, dialect='snowflake', exclude_rules=[
    'L032', 
    'L014', 
    'L016',
    'L027',
    'L029'
])
```
_Even with no rules excluded, the sqlfluff.fix produces the same output_

### Dialect

`snowflake`

### Version

sqlfluff 1.2.1 (using the Python API)

### Configuration

N/A

### Code of Conduct

### LINKED ISSUE #3833: L019 adds a comma between the new qualified alias and the unqualified reference after L028
### Search before asking

### What Happened

Currently this block:
```
with a as (SELECT 'test' as key,'value' AS value)

select
key,
value

from a
where value is not null


```
will return 

```
with a as (select
    'test' as key
, 'value' as value)

select
    a.key
    a., value

from a
where a.value is not null

```

### Expected Behaviour

I would expect the return to be 
```
with a as (select
    'test' as key
, 'value' as value)

select
    a.key
    ,  a.value

from a
where a.value is not null

```

### Observed Behaviour

```
with a as (select
    'test' as key
, 'value' as value)

select
    a.key
    a., value

from a
where a.value is not null

```

### How to reproduce

Create a sql file with these contents: 
```
with a as (SELECT 'test' as key,'value' AS value)

select
key,
value

from a
where value is not null
```

then run sqlfluff fix on it given the included sqlFluff config.

Below are the rule violations from the lint logs

```
L:   1 | P:  12 | L010 | Keywords must be lower case.
L:   1 | P:  12 | L036 | Select targets should be on a new line unless there is
                       | only one select target.
L:   1 | P:  29 | L029 | Keywords should not be used as identifiers.
L:   1 | P:  33 | L008 | Commas should be followed by a single whitespace unless
                       | followed by a comment.
L:   1 | P:  41 | L010 | Keywords must be lower case.
L:   1 | P:  44 | L029 | Keywords should not be used as identifiers.
L:   4 | P:   1 | L003 | Expected 1 indentation, found 0 [compared to line 03]
L:   4 | P:   1 | L028 | Unqualified reference 'key' found in single table
                       | select.
L:   4 | P:   4 | L019 | Found trailing comma. Expected only leading.
L:   5 | P:   1 | L003 | Expected 1 indentation, found 0 [compared to line 03]
L:   5 | P:   1 | L028 | Unqualified reference 'value' found in single table
                       | select.
L:   8 | P:   7 | L028 | Unqualified reference 'value' found in single table
                       | select.
L:   9 | P:   1 | L009 | Files must end with a single trailing newline.
```

### Dialect

snowflake

### Version

sqlfluff, version 1.3.0
```

### Configuration

[sqlfluff]
verbose = 0
nocolor = False
# Or run 'sqlfluff dialects'
dialect = snowflake
# One of [dbt|raw|jinja|python|placeholder]
templater = jinja
rules = all
exclude_rules = L031
recurse = 0
output_line_length = 80
runaway_limit = 10
ignore = None
ignore_templated_areas = True
encoding = autodetect
disable_noqa = False
fix_even_unparsable = False
# This limit skips files over a certain character length
# and warns the user what has happened.
# Set this to 0 to disable.
large_file_skip_char_limit = 20000
# If negative or zero, implies number_of_cpus - specifed_number.
# e.g. -1 means use all processors but one. 0  means all cpus.
processes = 1

[sqlfluff:indentation]
# See https://docs.sqlfluff.com/en/stable/indentation.html
indented_joins = False
indented_ctes = False
indented_using_on = True
indented_on_contents = True
template_blocks_indent = True

[sqlfluff:templater]
unwrap_wrapped_queries = True

[sqlfluff:templater:jinja]
apply_dbt_builtins = True

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 80
indent_unit = space
comma_style = leading

[sqlfluff:rules:L003]
hanging_indents = False

[sqlfluff:rules:L007]
operator_new_lines = after

[sqlfluff:rules:L010]
# Keywords
capitalisation_policy = lower
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L011]
aliasing = explicit

[sqlfluff:rules:L012]
aliasing = explicit

[sqlfluff:rules:L013]
#Column expression without alias. Use explicit AS clause
allow_scalar = False

[sqlfluff:rules:L014]
# Unquoted identifiers
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L015]
#DISTINCT used with parentheses. NOTE: No Config Argument was documented

[sqlfluff:rules:L016]
# Line length
ignore_comment_lines = False

[sqlfluff:rules:L017]
#Function name not immediately followed by parenthesis.

[sqlfluff:rules:L018]
#WITH clause closing bracket should be on a new line.

[sqlfluff:rules:L027]
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L026]
force_enable = False

[sqlfluff:rules:L028]
force_enable = False

[sqlfluff:rules:L029]
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L030]
# Function names
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L031]
force_enable = False

[sqlfluff:rules:L036]
wildcard_policy = single

[sqlfluff:rules:L038]
# Trailing commas

[sqlfluff:rules:L040]
# Null & Boolean Literals
capitalisation_policy = lower
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L042]
forbid_subquery_in = both

[sqlfluff:rules:L047]
prefer_count_1 = False
prefer_count_0 = False

[sqlfluff:rules:L051]
# Fully qualify JOIN clause

[sqlfluff:rules:L052]
multiline_newline = False

[sqlfluff:rules:L054]

[sqlfluff:rules:L057]
additional_allowed_characters = None
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L059]
ignore_words = None
ignore_words_regex = None
force_enable = False

[sqlfluff:rules:L062]
blocked_words = None
blocked_regex = None

[sqlfluff:rules:L063]
# Data Types
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L064]
force_enable = False

[sqlfluff:rules:L066]
min_alias_length = None
max_alias_length = None

```

- [] Yes I am willing to submit a PR!

### Code of Conduct

alanmcruickshank:
This builds on top of #3824 and is part of #3673.

It also resolves #3740 and resolves #3833.

This introduces `.rebreak()` (and a couple of functions we'll also rely on when we come to indentation like `.indent_to()`). Currently L007, L019 and L065 have been migrated across (trailing/leading/alone commas/operators). Likewise the old configs have been added to the removed list, so that users get a prompt to update their config.

Interestingly this also solves an edge case in L052 too.

Test cases (as per #3824) are relatively extensive but only on the public functions and not on the internal private ones, partly because test harnesses are hard and partly because those interfaces may change as this progresses.


src/sqlfluff/rules/L019.py
```diff
@@ -107,135 +68,19 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
+                config=context.config,
+            )
+            .rebreak()
+            .get_fixes()
+        )
+
+        if not fixes:
+            return LintResult()
+
+        # TODO: Still need to incorporate the nicer formatted description
```
barrywhart, line 84: Are you planning to address this "TODO" in the current PR?

src/sqlfluff/utils/reflow/depthmap.py
```diff
@@ -33,7 +33,7 @@ class DepthInfo:
     # This is a convenience cache to speed up operations.
     stack_hash_set: FrozenSet[int]
     stack_class_types: Tuple[FrozenSet[str], ...]
-    stack_positions: Tuple[str, ...]
+    stack_positions: Dict[int, Tuple[int, int, str]]
```
barrywhart, line 36: Think it's worth introducing a `NamedTuple` or `dataclass` here in place of the anonymous tuple structure, `Tuple[int, int, str]`?

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -28,11 +28,26 @@ class ReflowElement:
...
+
     @property
     def raw(self):
         """Get the current raw representation."""
         return "".join(seg.raw for seg in self.segments)
 
+    def num_newlines(self) -> int:
+        """How many newlines does this element contain?"""
+        return sum(
+            bool(seg.class_types.intersection({"newline", "end_of_file"}))
```
barrywhart, line 47: Any concern with equating end of file with newline? Especially if (as is typical) the last character in a file is a newline.

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -54,10 +69,15 @@ class ReflowBlock(ReflowElement):
     # - any:            don't enforce any spacing rules
     spacing_before: str
     spacing_after: str
+    # - None:           the default (no particular preference)
+    # - leading:        prefer newline before
+    # - trailing:       prefer newline after
+    line_position: Optional[str]
```
barrywhart, line 75: Should we consider making this an enum?

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -497,13 +523,128 @@ def _handle_respace__inline_without_space(
...
+                # Coerce existing indent to desired.
+                if indent_seg.raw == desired_indent:
+                    # Trivial case. Indent already correct
+                    return [], self
+                elif desired_indent == "":
+                    # Coerce to no indent. We don't want the indent. Delete it.
+                    new_indent = indent_seg.edit(desired_indent)
+                    idx = self.segments.index(indent_seg)
+                    return [LintFix.delete(indent_seg)], ReflowPoint(
+                        list(self.segments[:idx]) + list(self.segments[idx + 1 :])
```
barrywhart, line 576: Is it necessary to put `list()` around the two slices here?

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -497,13 +523,128 @@ def _handle_respace__inline_without_space(
...
+                    )
+
+                # Standard case of an indent change.
+                new_indent = indent_seg.edit(desired_indent)
+                idx = self.segments.index(indent_seg)
+                return [LintFix.replace(indent_seg, [new_indent])], ReflowPoint(
+                    list(self.segments[:idx])
```
barrywhart, line 583: Same question

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -497,13 +523,128 @@ def _handle_respace__inline_without_space(
...
+                    + [new_indent]
+                    + list(self.segments[idx + 1 :])
+                )
+            else:
+                # There is a newline, but no indent. Make one after the newline
+                idx = max(
+                    i for i, seg in enumerate(self.segments) if seg.is_type("newline")
+                )
```
barrywhart, line 591: Performance nitpick: Should we consider iterating the segments in reverse, then stopping on the first newline?

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -497,13 +523,128 @@ def _handle_respace__inline_without_space(
...
+                new_indent = WhitespaceSegment(desired_indent)
+                return [
+                    LintFix.create_after(self.segments[idx], [new_indent])
+                ], ReflowPoint(
+                    list(self.segments[: idx + 1])
```
barrywhart, line 596: Same question about `list()`

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -497,13 +523,128 @@ def _handle_respace__inline_without_space(
...
+                # it there.
+                new_segs: List[RawSegment]
+                if desired_indent == "":
+                    new_segs = [new_newline]
+                else:
+                    new_segs = [new_newline, indent_seg.edit(desired_indent)]
+                idx = self.segments.index(indent_seg)
+                fix = LintFix.replace(indent_seg, new_segs)
+                new_point = ReflowPoint(
+                    list(self.segments[:idx])
```
barrywhart, line 633: Same question about `list()`

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -20,6 +21,112 @@
...
+    def from_span(cls, span: _RebreakSpan, elements: ReflowSequenceType):
+        """Expand a span to a location."""
+        # First get the next newline.
+        prev_point_idx = span.start_idx - 1
+        next_point_idx = span.end_idx + 1
+
+        prev_nl_idx = prev_point_idx
+        next_nl_idx = next_point_idx
+        # We hop in 2s because we're checking two ahead.
+        while (
```
barrywhart, line 57: Could these two `while` loop be rewritten as `for` loops? Something like:
```
for prev_nl_idx in range(prev_point_idx, 2, -2):
```

For reference:
```
>>> list(range(13, 2, -2))
[13, 11, 9, 7, 5, 3]
```

Context: `for` loops are easier to read and it's clearer that they will definitely terminate.

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -20,6 +21,112 @@
...
+        # Here we use the newline index, not
+        # just the adjacent point, so that we can see past comments.
+        n_prev_newlines = elements[self.prev_nl_idx].num_newlines()
+        n_next_newlines = elements[self.next_nl_idx].num_newlines()
+        return (
+            # If there isn't a newline on either side then carry
+            # on. We're not interested in this yet.
+            not bool(n_prev_newlines or n_next_newlines)
+            # If there is a newline on BOTH sides. That's ok.
+            or bool(n_prev_newlines and n_next_newlines)
```
barrywhart, line 126: For readability, could we rewrite this using comparison operators? I find this a bit tricky to read.

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -199,25 +310,21 @@ def from_around_target(
         if sides in ("both", "before"):
             # Catch at least the previous segment
             pre_idx -= 1
-            while pre_idx - 1 > 0 and all_raws[pre_idx].is_type(
-                "whitespace", "newline", "indent"
-            ):
+            while pre_idx - 1 > 0 and not all_raws[pre_idx].is_code:
```
barrywhart, line 313: Nitpick: Could this and the next `while` loop be `for` loops?

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -458,3 +565,262 @@ def respace(
...
+            # Only evaluate blocks
+            if not isinstance(elem, ReflowBlock):
+                continue
+            # Does the element itself have config? (The easy case)
+            if elem.line_position:
+                # Blocks should only have one segment so it's easy to pick it.
+                spans.append(
+                    _RebreakSpan(elem.segments[0], idx, idx, elem.line_position)
+                )
+            # Do any of it's parents have config, and are we at the start
```
barrywhart, line 588:
```suggestion
            # Do any of its parents have config, and are we at the start
```

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -458,3 +565,262 @@ def respace(
...
+                                target, idx, end_idx, elem.line_position_configs[key]
+                            )
+                        )
+                        break
+                else:
+                    # If we find the start, but not the end, it's not a problem, but
+                    # we won't be rebreaking this span. This is important so that we
+                    # don't rebreak part of something without the context of what's
+                    # in the rest of it. Continue.
+                    continue
```
barrywhart, line 618: Does the `continue` here really do anything? We're at the end of the loop body already.

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -458,3 +565,262 @@ def respace(
...
+                    # to negotiate around. In this case, we _move the target_
+                    # rather than just adjusting the whitespace.
+
+                    # Delete the existing position of the target, and
+                    # the _preceding_ point.
+                    fixes.append(LintFix.delete(loc.target))
+                    for seg in elem_buff[loc.prev_point_idx].segments:
+                        fixes.append(LintFix.delete(seg))
+
+                    # We re-insert always reinsert after the first point, but respace
```
barrywhart, line 722:
```suggestion
                    # We always reinsert after the first point, but respace
```

test/fixtures/rules/std_rule_cases/L007.yml
```diff
@@ -98,20 +136,85 @@ passes_on_after_override:
 fails_on_before_override_with_comment_order:
   fail_str: |
     select
-        a
```
barrywhart, line 101: Why did we change an existing test rather than adding a new one?

test/fixtures/rules/std_rule_cases/L007.yml
```diff
@@ -98,20 +136,85 @@ passes_on_after_override:
...
+      /* foo */ 2
+  fix_str: |
+    SELECT
+      1
+      /* foo */ + 2,
+      1 -- foo
+      + 2,
+      1 /* foo */
+      + 2,
+      1 /* foo */ -- foo
```
barrywhart, line 190: These tests with interspersed comments remind me: sometimes I think it'd be interesting to have a "fuzz" tester that takes our existing test cases and throws randomly placed comments into them to see if something blows up. Think that may be useful?

test/rules/std_test.py
```diff
@@ -25,7 +25,6 @@
         # NB The newline before the comma, should report on the comma, not the newline
         # for clarity.
         ("L005", "whitespace_errors.sql", [(2, 9)]),
-        ("L019", "whitespace_errors.sql", [(4, 1)]),
```
barrywhart, line 28: What happened here?

test/utils/reflow/rebreak_test.py
```diff
@@ -0,0 +1,132 @@
...
+        ("select\n  a\n  , b", "select\n  a,\n  b"),
+        ("select\n    a\n    , b", "select\n    a,\n    b"),
+        ("select\n  a\n    , b", "select\n  a,\n    b"),
+        ("select\n  a\n  -- comment\n  , b", "select\n  a,\n  -- comment\n  b"),
+    ],
+)
+def test_reflow__sequence_rebreak_root(raw_sql_in, raw_sql_out, default_config, caplog):
+    """Test the ReflowSequence.rebreak() method directly.
+
+    Focussed around a whole segment.
```
barrywhart, line 45:
```suggestion
    Focused around a whole segment.
```

test/utils/reflow/rebreak_test.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    root = parse_ansi_string(raw_sql_in, default_config)
+    print(root.stringify())
+    seq = ReflowSequence.from_root(root, config=default_config)
+    for idx, elem in enumerate(seq.elements):
+        print(idx, elem)
+
+    with caplog.at_level(logging.DEBUG, logger="sqlfluff.rules.reflow"):
```
barrywhart, line 53: What are we using `caplog` for here? It doesn't look like we're doing anything with the captured logs.

test/utils/reflow/rebreak_test.py
```diff
@@ -0,0 +1,132 @@
...
+        ("select a,\n(b+c)", 4, "a,\n(", "a,\n("),
+        ("select a\n  , (b+c)", 6, "a\n  , (", "a,\n  ("),
+        # Here we don't have enough context to rebreak it so
+        # it should be left unaltered.
+        ("select a,\n(b+c)", 6, ",\n(b", ",\n(b"),
+        # This intentionally targets an incomplete span.
+        ("select a<=b", 4, "a<=", "a<="),
+    ],
+)
+def test_reflow__sequence_rebreak_target(
```
barrywhart, line 73: Nice test!

test/utils/reflow/rebreak_test.py
```diff
@@ -0,0 +1,132 @@
...
+    raw_sql_in, target_idx, seq_sql_in, seq_sql_out, default_config, caplog
+):
+    """Test the ReflowSequence.rebreak() method directly.
+
+    Focussed around a target segment. This intentionally
```
barrywhart, line 78:
```suggestion
    Focused around a target segment. This intentionally
```

test/utils/reflow/rebreak_test.py
```diff
@@ -0,0 +1,132 @@
...
+    root = parse_ansi_string(raw_sql_in, default_config)
+    print(root.stringify())
+    target = root.raw_segments[target_idx]
+    print("Target: ", target)
+    seq = ReflowSequence.from_around_target(target, root, config=default_config)
+    for idx, elem in enumerate(seq.elements):
+        print(idx, elem)
+    assert seq.get_raw() == seq_sql_in
+
+    with caplog.at_level(logging.DEBUG, logger="sqlfluff.rules.reflow"):
```
barrywhart, line 90: Same question about `caplog`.

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: This is a really good point. I've actually changed the handling of this in the next PR, but I think there's an argument for me to push that down into this one. I'll do that so that it's handled consistently from the start.

src/sqlfluff/rules/L019.py
alanmcruickshank: Good catch, yes. I'll add that in.

test/utils/reflow/rebreak_test.py
alanmcruickshank: Caplog here controls the level of logging shown by pytest on a fail. It's a fairly standard pattern: https://docs.pytest.org/en/7.1.x/how-to/logging.html#caplog-fixture

test/utils/reflow/rebreak_test.py
alanmcruickshank: Caplog here controls the level of logging shown by pytest on a fail. It's a fairly standard pattern: https://docs.pytest.org/en/7.1.x/how-to/logging.html#caplog-fixture

test/fixtures/rules/std_rule_cases/L007.yml
alanmcruickshank: I agree - and I'm thinking about doing something like that when we get to the full blown reflow command. That way we can randomly comment and space sql files and then check that the reflow command is deterministic. I'll add a task on the master issue to do that.

test/fixtures/rules/std_rule_cases/L007.yml
alanmcruickshank: It seemed better to extend the existing test (see the addition of a third comment), which make this test cover additional functionality, than just adding in another test and having two tests which test almost the same thing.

test/rules/std_test.py
alanmcruickshank: Ok - this is a good point to discuss.

That file effectively has a part which has this structure if you simplify it:
```sql
SELECT
    a
    , -- comment
    b
```

Previously that comma line would get flagged up as a _leading_ comma and so trigger `L019`.

In a situation like this there are 4 possible cases:
1. `inline` (i.e. no newlines either side): `a, b`
2. `leading`: `a\n, b` (newline before comma, not after)
3. `trailing`: `a,\n b` (newline after comma, not before)
4. `alone`: `a\n,\n b` (newlines both before _and_ after)

That line in `whitespace_errors.sql` is a situation which I've classed as `alone` (even though there's a comment on that line too which I think adds to the ambiguity).

I got a bit deeper into this in the next PR, which I think I'll bring more of into this PR, but where I'm aiming at is that if you configure yourself for `trailing` (the default), then it also allows `inline` _and_ `alone`, for commas this might initially seem weird but for operators I think it's easier to picture and I don't see why commas should be different. The extension of this is that I'm thinking to allow users to qualify that configuration with `:strict` which means the other cases aren't allowed anymore.

Given that I'm very slightly changing the behaviour of `L019` in doing this - this particular test no longer fails for L019 because I'm not allowing this case by default.

Do you think this is a sensible move?

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: Not for now - potentially in the future. I've brought in some of the additional commits from the next PR and you'll see I'm extending these configs a little to make them more flexible.

src/sqlfluff/utils/reflow/depthmap.py
alanmcruickshank: Yeah good idea.

alanmcruickshank: Done. Converted to frozen dataclass.

src/sqlfluff/utils/reflow/elements.py
alanmcruickshank: No - this is good feedback which I should have taken from the first PR too 🤦 . I think given your other comment, and that a `ReflowElement` is a frozen `dataclass` I've instead swapped the datatype to a `tuple` and then also removed all the unnecessary type changes. There are a couple still in there now where we do the working logic in a `list` and then convert to tuple on constructing a `ReflowElement` but mostly we can work on slicing in the native type 👍

src/sqlfluff/utils/reflow/sequence.py
alanmcruickshank: Fair point, I'll tidy this up and make it more explicit.

src/sqlfluff/utils/reflow/sequence.py
alanmcruickshank: Good idea.

test/utils/reflow/rebreak_test.py
barrywhart: Got it. Outside of SQLFluff, most things I work on don't involve so much logging (parsers will do that!). Good to know!!

test/rules/std_test.py
barrywhart: I think so. It's natural that as we think more deeply and rigorously about some of these corner cases, we'll refine it beyond what we did in the first version. 👍

src/sqlfluff/utils/reflow/sequence.py
barrywhart: Cool. This was more of a nitpick / readability thing. I had to look at the code a few times to be sure I wasn't missing something.

One really simple way to address this would be to change the `continue` to a `pass`, but maybe you have a better idea. 👍

src/sqlfluff/utils/reflow/sequence.py
alanmcruickshank: Check out the new version - is it a bit clearer?

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -20,6 +21,122 @@
...
+               is taken from to allow comparison.
+            strict (:obj:`bool`): If set to true, this will not allow
+               the case where there aren't newlines on either side.
+        """
+        # Here we use the newline index, not
+        # just the adjacent point, so that we can see past comments.
+        n_prev_newlines = elements[self.prev_nl_idx].num_newlines()
+        n_next_newlines = elements[self.next_nl_idx].num_newlines()
+        newlines_on_neither_side = n_prev_newlines + n_next_newlines == 0
+        newlines_on_both_sides = n_prev_newlines > 0 and n_next_newlines > 0
```
barrywhart, line 130: New version looks great! This line in particular is much easier to read now.

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -458,3 +627,322 @@ def respace(
...
+                                elem.line_position_configs[key].split(":")[0],
+                                elem.line_position_configs[key].endswith("strict"),
+                            )
+                        )
+                        break
+                # If we find the start, but not the end, it's not a problem, but
+                # we won't be rebreaking this span. This is important so that we
+                # don't rebreak part of something without the context of what's
+                # in the rest of it. We continue without adding it to the buffer.
+        return spans
```
barrywhart, line 692: A "bigger picture" question occurred to me as I was reading this function. Are there places in this new code where it would be helpful to introduce preconditions and/or postconditions? I.e. places where we check that things are in the expected state? Kind of analogous to the internal linter check that reruns the parser on "fixed" areas of the code to verify no parse errors were introduced.

In reflow, these checks might enforce things like:
* The alternating structure of a reflow sequence (whitespace, not whitespace)
* No adjacent whitespace segments (is that a rule?)

If we aren't already doing this, a little of this, applied strategically, could help catch bugs and aid in future maintenance.

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -458,3 +627,322 @@ def respace(
...
+                        + elem_buff[
+                            loc.prev_point_idx + 1 : loc.next_point_idx
+                        ]  # the target
+                        + elem_buff[loc.prev_code_pt_idx : loc.prev_point_idx + 1]
+                        + elem_buff[loc.next_point_idx + 1 :]
+                    )
+
+            elif loc.line_position == "alone":
+                # If we get here we can assume that the element is currently
+                # either leading or trailing and needs to be moved onto it's
```
barrywhart, line 907:
```suggestion
                # either leading or trailing and needs to be moved onto its
```

barrywhart:
Looks good overall! A few small, specific suggestions, and a couple questions.

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -20,6 +21,122 @@
...
+    line_position: str
+    strict: bool
+
+
+@dataclass(frozen=True)
+class _RebreakLocation:
+    """A location within a sequence to rebreak, with metadata."""
+
+    target: BaseSegment
+    prev_code_pt_idx: int
```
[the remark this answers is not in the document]
alanmcruickshank: Perhaps 🤔. The two would kind of be mirrors of each other however, so might not allow total reuse. I'll have a closer look to see what might be possible....

src/sqlfluff/utils/reflow/sequence.py
alanmcruickshank: So the first of these (the alternating structure), is checked on every construction of a `ReflowSequence`. That definitely is already enforced.

The second one - that's not really a rule, as much as an easy fix. It's picked up in `respace`, every time that's called, and so I'm not sure the others need to handle it. It might still be worth having it though 🤔 .

src/sqlfluff/utils/reflow/depthmap.py
```diff
@@ -12,16 +12,36 @@
...
+            return "solo"
+        elif path_step.idx == 0:
+            return "start"
+        elif path_step.idx == path_step.len - 1:
+            return "end"
+        else:
+            return ""  # NOTE: Empty string evaluates is falsy.
+
+    @classmethod
+    def from_path_step(cls, path_step: PathStep):
```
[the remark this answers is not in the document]
alanmcruickshank: That formulation on it's own doesn't work, I'll change it to: `def from_path_step(cls: Type["StackPosition"], path_step: PathStep) -> "StackPosition":` which does. Otherwise it errors that `T` is not defined.

src/sqlfluff/utils/reflow/sequence.py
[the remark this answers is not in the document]
alanmcruickshank: Ok - this is actually a good idea. Will update PR shortly - got something good working.

src/sqlfluff/utils/reflow/sequence.py
barrywhart: It's up to you. These were just examples, so there may be other situations where it'd be more valuable. Just a thought/question, don't worry too much about it -- can always add later if we see a need.

src/sqlfluff/utils/reflow/sequence.py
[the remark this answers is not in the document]
barrywhart: Oh cool. Over the years at various jobs, I've discovered many bugs in different copies of the same code. Actually fixed several in SelectCrawler recently, which was handling simple queries differently from the "main" query when CTEs were involved.

src/sqlfluff/utils/reflow/sequence.py
alanmcruickshank: So the existing `_validate_reflow_sequence` method checks for alternating points and blocks. I've added a clause in there to check for double whitespace, because I agree that should never happen.

src/sqlfluff/utils/reflow/depthmap.py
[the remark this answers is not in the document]
barrywhart: I think this is it.

https://twitter.com/willmcgugan/status/1397636465404915719?s=20&t=INxWfvN8AwLA8V7RwAKTjA

src/sqlfluff/utils/reflow/sequence.py
alanmcruickshank: I take it back - that check causes issues, because I'm using `respace` to _fix_ that when it happens, so I have to accept things like that initially. Good idea, but I'm going to leave it out for now.

barrywhart: Sounds good. Perhaps later (if it seems worthwhile), we can devise a way to check the parse tree after reflow-related fixes have been applied. Difficult to check prior to that.

Maybe won't be needed, just remembering last year, finding and fixing little issues left over when fixes were slightly wrong. 🙏

src/sqlfluff/utils/reflow/depthmap.py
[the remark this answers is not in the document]
alanmcruickshank: Yeah that works - it needs the `TypeVar` definition at the top of the file. I think being explicit here is overall simpler.

alanmcruickshank:
Ok, I think it's ready for a final review. Sorry for the minor git chaos at the end - there was an issue introduced by #3908 which clashed with this pr. The solution was stopping the reflow engine checking whitespace around template tags.

I might revisit that in a later development, and that will be necessary for full reflow - but I'm going to suggest we handle that then.

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -32,58 +32,61 @@ class _RebreakSpan:
...
         return cls(
             span.target,
-            prev_code_pt_idx,
-            prev_nl_idx,
-            prev_point_idx,
-            next_point_idx,
-            next_nl_idx,
-            next_code_pt_idx,
+            _RebreakIndices.from_elements(elements, span.start_idx, -1),
+            _RebreakIndices.from_elements(elements, span.end_idx, 1),
```
barrywhart, line 89: 🎉 😎

barrywhart (approved):
Looks great!

Issue #356, opened 2020-05-19
Rule L025 fails when using BigQuery UNNEST function with alias


johansjob:
**Problem**
BigQuery's unnest function with alias fails with rule L025, alias is never used.

**Example**
```
select ts from unnest(generate_timestamp_array('2020-01-01', '2020-01-30', interval 1 day)) as ts
```

**Expected**
Successful parsing :)

**Settings**
Version: 0.3.4
Dialect: BigQuery

johansjob:
Adding another example, the UNNEST breaks rule L027 as well.

```
with a as (
    select 1
)

select a.*, _t_start from a
left join unnest(generate_timestamp_array('2020-01-01', '2020-01-30', interval 1 day)) as _t_start on true
```

johansjob:
And another one, of course you can do the cross join variant as well..

```
with a as (
    select 1
)

select 
    a.*
    , _t_start 
from a, unnest(generate_timestamp_array('2020-01-01', '2020-01-30', interval 1 day)) as _t_start
```

alanmcruickshank:
Excellent examples. Thanks @johansjob.

I think this is a bug because this *should* work. Looking at `TableExpressionSegment`, there's an optional `AliasExpressionSegment` in there already. I'll need to examine some of the logs to work out what's going on.

barrywhart:
I'm taking an initial look at this and may attempt a fix. Here is `sqlfluff parse` output from the first test query. Based on the indentation, it looks like it's parsing correctly, in the sense that the `alias_expression` is nested inside the `table_expression`, at the same level as the `function` above. So my initial intuition is that this may be a bug in the lint rules rather than the parser. If so, I'm thinking of two general approaches to fixing it:
* Fix the rules specifically mentioned above
* Consider changing the `table_expression` grammar so these aliases are not easily mistaken for table aliases. If we can do this, it may help avoid similar bugs in the future. 🎉💯

```
[15](1, 1, 16)      |                table_expression:
[15](1, 1, 16)      |                    function:
[15](1, 1, 16)      |                        function_name:                        'unnest'
[21](1, 1, 22)      |                        start_bracket:                        '('
[22](1, 1, 23)      |                        [META] indent:
[22](1, 1, 23)      |                        expression:
[22](1, 1, 23)      |                            function:
[22](1, 1, 23)      |                                function_name:                'generate_timestamp_array'
[46](1, 1, 47)      |                                start_bracket:                '('
[...]

PR #3848, opened 2022-09-13
Allow standalone aliases in L027

olagjo:
This is related to #356.
The previous fix of this issue (https://github.com/sqlfluff/sqlfluff/pull/722/files) ended up passing a parameter
to rule L027 for `value_table_function_aliases`, but not actually exempting these aliases from the rule.

It seems to me that the test case then passes because the result of the `unnest` function call is not counted as a table
reference, rather than passing because it is ok to use "unqualified" references for value tables (they are impossible
to qualify).

As such, I have added a new testcase `test_ignore_value_table_functions_when_counting_unqualified_aliases` which fails
on main, but passes with this fix.

### DIFF
--- a/test/fixtures/rules/std_rule_cases/L027.yml
+++ b/test/fixtures/rules/std_rule_cases/L027.yml
@@ -61,12 +61,34 @@ test_allow_date_parts_as_function_parameter_snowflake:
     core:
       dialect: snowflake
 
-test_ignore_value_table_functions:
+test_ignore_value_table_functions_when_counting_tables:
+  # Allow use of unnested value tables from bigquery without counting as a
+  # table reference. This test passes despite unqualified reference
+  # because we "only select from one table"
+  pass_str: |
+    select
+        unqualified_reference_from_table_a,
+        _t_start
+    from a
+    left join unnest(generate_timestamp_array(
+            '2020-01-01', '2020-01-30', interval 1 day)) as _t_start
+        on true
+  configs:
+    core:
+      dialect: bigquery
+
+test_ignore_value_table_functions_when_counting_unqualified_aliases:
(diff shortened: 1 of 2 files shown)


test/fixtures/rules/std_rule_cases/L027.yml
```diff
@@ -61,12 +61,34 @@ test_allow_date_parts_as_function_parameter_snowflake:
     core:
       dialect: snowflake
 
-test_ignore_value_table_functions:
+test_ignore_value_table_functions_when_counting_tables:
```
olagjo, line 64: This, despite the slightly weird diff, is just the old test `test_ignore_value_table_functions` updated to
- Have a new name for the specific case it is testing
- Have make the reference to a column from `a` unqualified (to verify that this still passes lint, now that `_t_start` is ignored)

test/fixtures/rules/std_rule_cases/L027.yml
```diff
@@ -61,12 +61,34 @@ test_allow_date_parts_as_function_parameter_snowflake:
...
+        _t_start
+    from a
+    left join unnest(generate_timestamp_array(
+            '2020-01-01', '2020-01-30', interval 1 day)) as _t_start
+        on true
+  configs:
+    core:
+      dialect: bigquery
+
+test_ignore_value_table_functions_when_counting_unqualified_aliases:
```
olagjo, line 80: This is a new test, which does the same as the old test `test_ignore_value_table_functions`, but verifies that `_t_start` is still ignored when more than one proper table is joined in

barrywhart (approved):
Nice catch!

Embarrassed about the bug in my earlier PR. I am much more experienced now. 🤣

PR #3874, opened 2022-09-25
Support Removed warning on configs

alanmcruickshank:
This is a fairly separable part of #3847, and can be merged separately. It will also simplify the review of that PR to take these changes out of it.

This allows particular config values to be marked as deprecated, prompting the user to remove them if they are set in a config. Optionally it also allows a configurable message to be set, informing the user how to update their config accordingly.

### DIFF
--- a/src/sqlfluff/core/config.py
+++ b/src/sqlfluff/core/config.py
@@ -28,7 +28,7 @@ can still cache appropriately
 
 ConfigElemType = Tuple[Tuple[str, ...], Any]
 
-DEPRECATED_CONFIGS: Dict[Tuple[str, ...], Any] = {
+REMOVED_CONFIGS: Dict[Tuple[str, ...], Any] = {
     ("rules", "L003", "lint_templated_tokens"): ("No longer used."),
 }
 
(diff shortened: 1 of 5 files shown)


src/sqlfluff/core/config.py
```diff
@@ -267,13 +283,30 @@ def _incorporate_vals(ctx: dict, vals: List[Tuple[Tuple[str, ...], Any]]) -> dic
...
 
-    def load_default_config_file(self, file_dir: str, file_name: str) -> dict:
+    @staticmethod
+    def _validate_configs(configs: Iterable[ConfigElemType], file_path):
+        """Validate config elements against deprecated list."""
+        for k, _ in configs:
+            if k in DEPRECATED_CONFIGS:
+                formatted_key = ":".join(k)
+                raise SQLFluffUserError(
+                    f"Config file {file_path} set deprecated config "
```
barrywhart, line 293: The typical pattern I've seen for deprecation is to warn the user, then in a future release to remove it or raise an error. Wondering if we want to consider changing the behavior and/or use a term other than "deprecate".

barrywhart (approved):
Looks good overall. One question about the term "deprecated" and/or behavior of it.

src/sqlfluff/core/config.py
alanmcruickshank: Hmmmm - good idea. I'm going to change the wording. I think the more "brutal" approach is good because otherwise users may get strange linting results which they look at before any raised warnings.

PR #3875, opened 2022-09-25
Refactor path_to

alanmcruickshank:
This is another relatively separable part of #3847. It's a useful development alone and can be merged before the rest. That will also simplify the review of the other PR.

This changes the behaviour of `BaseSegment.path_to()`.

Previously it would return a `list` of segments. The first element of that list would be `self` and the last would be the segment passed as `other`. The others would be the segments between.

I noticed that many uses of `path_to` always cut the last element (i.e. slice `[:-1]`), and that there are use cases where we want to know more than just the lineage, but also where in the parent each child is.

This changes `path_to` to instead:
- ...return a `list` of `PathStep` (a `dataclass`), which combines the segment with a new `index` which states where in the parent the child was found. This is useful for understanding whether the segment was at the start or end of a parent.
- ...to return one _less_ element in that it will never return the `other` (i.e. the last element previously returned). This is partly because it was not often used, partly because anyone calling `path_to` already has access to that and mostly because it doesn't make sense to ask what index something is within itself.

### DIFF
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -78,6 +78,21 @@ class SourceFix:
         return hash((self.edit, self.source_slice.start, self.source_slice.stop))
 
 
+@dataclass()
+class PathStep:
+    """An element of the response to BaseSegment.path_to().
+
+    Attributes:
+        segment (:obj:`BaseSegment`): The segment in the chain.
+        idx (int): The index of the target within it's `segment`.
+        len (int): The number of childen `segment` has.
+    """
+
+    segment: "BaseSegment"
+    idx: int
+    len: int
+
+
 @dataclass
 class FixPatch:
     """An edit patch for a source file."""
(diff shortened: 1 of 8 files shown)


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -78,6 +78,21 @@ def __hash__(self):
         return hash((self.edit, self.source_slice.start, self.source_slice.stop))
 
 
+@dataclass()
+class PathStep:
+    """An element of the response to BaseSegment.path_to().
+
+    Attributes:
+        segment (:obj:`BaseSegment`): The segment in the chain.
+        idx (int): The index of the target within it's `segment`.
```
barrywhart, line 87:
```suggestion
        idx (int): The index of the target within its `segment`.
```

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -78,6 +78,21 @@ def __hash__(self):
         return hash((self.edit, self.source_slice.start, self.source_slice.stop))
 
 
+@dataclass()
```
barrywhart, line 81: Should we consider this? (Mainly for documentation; I assume these objects won't be changed after creation.)
```suggestion
@dataclass(frozen=True)
```

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1065,33 +1080,40 @@ def recursive_crawl(
...
+        # Return empty if we've found the segment.
         if self is other:
-            return [self]
+            return []  # pragma: no cover
 
         # Are we in the right ballpark?
         # NB: Comparisons have a higher precedence than `not`.
         if not self.get_start_loc() <= other.get_start_loc() <= self.get_end_loc():
-            return None
+            return []
```
barrywhart, line 1098: I have sometimes wondered if this result should be indicated a different way. It could be an error in some situations.

Also, does this result become ambiguous now that the last step is removed? It could be interpreted in two ways:
* There is no path to `other` from `self`.
* `other` is an immediate child of `self`.

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: Good idea

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: So my interpretation of this is that it's ok for `[]` to be the only result and not require an additional `None` result. The reason is that `[]` should always be interpreted as `other` _is not a descendent of_ `self`. That interpretation would include `other is self`, because something is not a descendent of itself.

I'll update the docstring to be clearer on that - but does that make sense?

alanmcruickshank: I've updated the docstring with the explanation:

```rst
The result of this should be interpreted as *the path from `self` to `other`*.
If the return value is `[]` (an empty list), that implies there is no path
from `self` to `other`. This would include the case where the two are the same
segment, as there is no path from a segment to itself.
```

barrywhart: Oh, I see -- if `other` is an immediate child, it returns a path containing `self`. The code is definitely good. In mathematical terms, the path we're returning would be described as a ["half open interval"](https://mathresearch.utsa.edu/wiki/index.php?title=Interval_Notation#:~:text=A%20half%2Dopen%20interval%20includes,0%20and%20less%20than%201.), which we could mention in the docs as `self, other)`. I think it would be cool add that additional info, because it's formal and precise about what's being returned. Sometimes words are not enough. 😛

alanmcruickshank: @barrywhart - I've added some better tests to spell it out better and a not in the docstring. Reckon that's better?

barrywhart (approved):
Looks great! Thanks for the explanation and the doc changes!

src/sqlfluff/core/parser/segments/base.py
barrywhart: 🎉♥️

PR #3877, opened 2022-09-25
Adjust TSQL Operators

alanmcruickshank:
This is part of #3824.

Given it might be a bit longer until that gets merged, it seems sensible to merge this part earlier because it's easily separable.

This makes more precise use of `RawEqualsSegment` in TSQL rather than the existing use of `EqualsSegment`.


barrywhart (approved):
Looks good!

PR #3878, opened 2022-09-25
raw_segments_with_ancestors

alanmcruickshank:
This is another relatively standalone part of #3824.

It's slightly more abstract than the others as nothing in the codebase _uses_ the method yet, but it is well tested, and takes a little more logic out of the parent PR to make that more manageable.

This has already been partly reviewed by @barrywhart as part of the original PR so should look familiar.


barrywhart (approved):
Looks good!

Not merging yet because update/CI build is still in progress.

alanmcruickshank:
Actually - hold off on merging - there's another commit that needs to be added here 🤦 .

alanmcruickshank:
There we go! Juggling too many branches. That was my fault. Ready to merge at will now.

PR #3898, opened 2022-09-30
Adding missing KeywordSegments for different file types in Athena dialect

CommonCrisis:
These types are missing --> therefore the linter is failing with:
`RuntimeError: Grammar refers to 'OrcKeywordSegment' which was not found in the athena dialect. Perhaps specify the keyword? https://github.com/sqlfluff/sqlfluff/wiki/Contributing-Dialect-Changes#keywords`

These are set here (`dialect_athena.py`) but are missing as KeywordSegment.

```
    FileFormatGrammar=OneOf(
        "SEQUENCEFILE",
        "TEXTFILE",
        "RCFILE",
        "ORC",
        "PARQUET",
        "AVRO",
        "JSONFILE",
        "ION",
        Sequence(
            "INPUTFORMAT",
            Ref("QuotedLiteralSegment"),
            "OUTPUTFORMAT",
            Ref("QuotedLiteralSegment"),
        ),
    ),
```

It actually works then if  the output format looks like this in the query
```
STORED AS INPUTFORMAT
  'org.apache.hadoop.hive.ql.io.orc.OrcInputFormat'
OUTPUTFORMAT
  'org.apache.hadoop.hive.ql.io.orc.OrcOutputFormat'
```

### Pull Request checklist

  - Other.


barrywhart (changes requested):
Can you add test cases for these additions? They would go in `test/fixtures/dialects/athena/`.

CommonCrisis:
Added a test for orc file types based on the AWS athena example in their docs

alanmcruickshank:
I think you might need to run `python test/generate_parse_fixture_yml.py --athena` to build the associated yaml files for your test case.

CommonCrisis:
Done

test/fixtures/dialects/athena/create_external_table_orc_file_type.sql
```diff
@@ -0,0 +1,13 @@
+CREATE EXTERNAL TABLE my_table(
```
barrywhart, line 1: This file doesn't appear to use the new `ORCFILE` keyword. Wasn't that the intent of adding this test file, to ensure that the new keyword parses correctly?

barrywhart (changes requested):

test/fixtures/dialects/athena/create_external_table_orc_file_type.sql
barrywhart: This PR has become confusing because there was no associated issue. In your PR, you mention that the linter is failing with a specific error, but _you didn't provide an example of a file that reproduced the failure_`. That's what we need here. I can see that this file would use ORC if it were executed in Athena, but that's not relevant. We need a test case that actually uses the `ORCFILE` keyword. You must've had a file like that which was failing, right? **That** file can be the test case. (Preferably a simplified/minimized version of it.)

CommonCrisis: Ahhh I know get whats happening. It is acutally not the sql query it is the implementation. The problem is `STORED AS INPUTFORMAT`. Once it is set it will check grammar and the keyword segements. It is failing because Orc is mentioned in grammar but not as Keyword.

So it is actually not about orc but the `INPUTFORMAT` key word.

Btw: The query is kind of the one that was failing - but now it's clear why.


I will remove the test and create one that checks for `STORED AS INPUTFORMAT`. The formats itself should be tested as well I guess but that should be another PR.

barrywhart (approved):

sxooler:
Missing segments for "ORC", "PARQUET",  "AVRO" and "ION" were added but all contain `FILE` suffix which is not correct, if I am not mistaken. So, e.g. `STORED AS PARQUETFILE` is accepted, but the correct form `STORED AS PARQUET` not. Could you confirm?

Issue #3512, opened 2022-06-29
redshift dialect: parse error while comparing column values in where clause


6e2f61:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Querying redshift system tables produces parsing error in where clause while comparing values.

### Expected Behaviour

Add support for the redshift system table select query syntax.

### Observed Behaviour

Part of query not currently supported, results in parse errors. Schema somehow is omitted by sqlfluff
```bash
== [app/sys_tbl_redshift_part1.sql] FAIL                                                                                                                                                             
L:   7 | P:   5 | L027 | Unqualified reference 'pg_namespace' found in select with more than one referenced table/view.
L:   7 | P:  17 |  PRS | Line 7, Position 17: Found unparsable section: '.oid = pg_constraint.connamespace'
WARNING: Parsing errors found and dialect is set to 'redshift'. Have you configured your dialect correctly?
All Finished 📜 🎉!
```

### How to reproduce

Create a test file containing as below and try to lint/parse with sqlfluff.
```sql
SELECT
    pg_namespace.nspname AS constraint_schema
    , pg_constraint.conname AS constraint_name
FROM
    pg_namespace, pg_constraint
WHERE
    pg_namespace.oid = pg_constraint.connamespace
```

### Dialect

redshift

### Version

`sqlfluff, version 1.0.0`
`Python 3.8.10`

### Configuration

```bash
# https://docs.sqlfluff.com/en/stable/configuration.html#config

[sqlfluff]
dialect = redshift
templater = jinja
rules = all
exclude_rules = L034
# Controls SQLFluff output, see max_line_length for SQL output
output_line_length = 118
ignore_templated_areas = true
sql_file_exts = .sql,.sql.j2,.dml,.ddl

[sqlfluff:indentation]
indented_joins = false
indented_using_on = true
template_blocks_indent = true

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 118
indent_unit = space
comma_style = trailing

# https://docs.sqlfluff.com/en/stable/rules.html#ruleref

[sqlfluff:rules:L003]
hanging_indents = true

[sqlfluff:rules:L007]
operator_new_lines = after

[sqlfluff:rules:L010]
# Keywords
capitalisation_policy = upper

[sqlfluff:rules:L011]
# of tables
aliasing = explicit

[sqlfluff:rules:L012]
# of columns
[...]

PR #3905, opened 2022-10-02
fix(redshift): Allow keywords in qualified references

tomasfarias:

Fixes: #3512

A couple of `.yml` test fixtures where automatically re-created by `tox -e generate-fixture-yml -- --dialect redshift`. AFAIK this should be harmless.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_redshift.py
+++ b/src/sqlfluff/dialects/dialect_redshift.py
@@ -177,27 +177,6 @@ redshift_dialect.replace(
             anti_template=r"^(" + r"|".join(dialect.sets("reserved_keywords")) + r")$",
         )
     ),
-    ColumnReferenceSegment=Delimited(
-        Sequence(
-            ansi.ColumnReferenceSegment,
-            AnyNumberOf(Ref("ArrayAccessorSegment")),
-        ),
-        delimiter=OneOf(
-            Ref("DotSegment"), Sequence(Ref("DotSegment"), Ref("DotSegment"))
-        ),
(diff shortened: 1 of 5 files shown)


src/sqlfluff/dialects/dialect_redshift.py
```diff
@@ -2401,3 +2401,11 @@ class FunctionSegment(ansi.FunctionSegment):
...
         ),
     )
+
+
+class ColumnReferenceSegment(postgres.ColumnReferenceSegment):
+    """A reference to column, field or alias.
+
+    We override this for Redshift to allow keywords in fully qualified column
+    names (using Full segments) like it's done in Postgres.
+    """
```
alanmcruickshank, line 2411: I think you're right that this should just refer to postgres - but not by inserting it here.

The redshift dialect already should inherit everything direct from postgres. The issue here is that it seems that someone is already overriding that on line 180 of `dialect_redshift.py`. I think if you delete those lines you should be able to get the same effect without needing to effectively double-override it here.

alanmcruickshank (changes requested):

src/sqlfluff/dialects/dialect_redshift.py
tomasfarias: Good call! I didn't spot line 180. I'll apply the changes and run the tests :+1:

tomasfarias:
Rebased my branch with `main`.

alanmcruickshank (approved):

PR #3908, opened 2022-10-02
Bring L008 into reflow work

alanmcruickshank:
This part of #3673.

It brings L008 into the reflow work, and uses logic which was all merged in #3824.


barrywhart (approved):
Looks good to me! I see the PR is still marked "Draft", so wasn't sure if you planned to make additional changes.

PR #3915, opened 2022-10-04
Map old configs to new configs

alanmcruickshank:
So #3824 and #3847 changed existing configs for some rules. This makes it less backward compatible.

This PR maps the old configs to the new configs, and warns the user to update their config. This also makes the config changes not breaking changes. It makes it easier to deploy parts of #3673 incrementally.

In addition I've cleaned up some incorrect references. For python 3.2+ we also don't need to subclass `logging.Filter` anymore and can just provide a callable.

Slightly annoyingly, the config validation is done before the full logging setup is done so the warnings don't come out in red 🤷 . I had a quick look at resolving that but it seemed overkill.

### DIFF
--- a/src/sqlfluff/core/config.py
+++ b/src/sqlfluff/core/config.py
@@ -17,7 +17,7 @@ import appdirs
 
 import toml
 
-# Instantiate the templater logger
+# Instantiate the config logger
 config_logger = logging.getLogger("sqlfluff.config")
 
 global_loader = None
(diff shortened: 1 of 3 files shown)


src/sqlfluff/core/config.py
```diff
@@ -294,17 +367,46 @@ def _incorporate_vals(ctx: dict, vals: List[ConfigElemType]) -> dict:
...
-                    " for more details."
-                )
+                removed_option = config_map[k]
+                # Is there a mapping option?
+                if removed_option.translation_func and removed_option.new_path:
+                    # Mutate and warn.
+                    v = removed_option.translation_func(v)
+                    k = removed_option.new_path
+                    formatted_new_key = ":".join(k)
+                    # NOTE: At the stage of emitting this warnig, we may not yet
```
barrywhart, line 386:
```suggestion
                    # NOTE: At the stage of emitting this warning, we may not yet
```

src/sqlfluff/core/config.py
```diff
@@ -294,17 +367,46 @@ def _incorporate_vals(ctx: dict, vals: List[ConfigElemType]) -> dict:
...
+                    # have set up red logging because we haven't yet loaded the config
+                    # file. For that reason, this error message has a bit more padding.
```
barrywhart, line 388: Should we consider adding some ASCII decoration as well? E.g.: `*********`, `======`, something like that.

barrywhart (approved):
Looks good! A couple minor things.

src/sqlfluff/core/config.py
alanmcruickshank: I think the newlines are enough for now - they space it out nicely.

PR #3924, opened 2022-10-06
Remove is_template

alanmcruickshank:
This appears to be hardly used as a kwarg. This is a draft PR to see if we can remove it. If tests fail I'll keep it but document it better.


alanmcruickshank:
Ok I get it. Will document this better in #3925

PR #3925, opened 2022-10-06
Refactor Lexer: Split apart elements_to_segments and refine placement of meta segments.

alanmcruickshank:
To support migration of L003 to reflow, this adds `Indent` and `Dedent` segments around template loop markers. They already exist around template block placeholders so it's odd that they're not _also_ here. This should allow us to remove a chunk of complexity in L003.

This logic is embedded in the lexer and there weren't many test cases to really cover some of these situations. This adds three more tests in the jinja templater suite to cover some more cases. On the logic side I've broken out the placeholder logic out of the bigger function (`.elements_to_segments()`) into a couple of smaller methods. One of them is then reused for the _end of file_ situation which had it's own custom logic (which also missed some of the advances made in the main logic).

This also resolves an edge case where `TemplateLoop` markers were being incorrectly inserted around some template tags that weren't loops.

I'm also _really_ surprised that this section of code wasn't causing any issues - but flicking through the outstanding bugs, I don't think it was. In any case this should prevent future bugs by being much more robust, well tested and consistent.

Relates to #3673


alanmcruickshank:
Actually - don't review this yet - I think it needs some more tests and more modularisation. I don't think these code paths are behaving as they should.

src/sqlfluff/core/parser/lexer.py
```diff
@@ -561,27 +633,16 @@ def elements_to_segments(
...
+            if last_source_slice and idx == len(elements) - 1:
+                template_segments, _ = _generate_placeholder_segments(
+                    slice(source_slice.stop, len(templated_file.source_str)),
+                    last_source_slice,
+                    slice(element.template_slice.stop, element.template_slice.stop),
+                    source_only_slices,
+                    templated_file,
+                    add_indents,
+                )
+                segment_buffer.extend(template_segments)
```
alanmcruickshank, line 645: This is the main _behaviour_ change here. The rest is test cases or reorganisation. Previously it had it's own logic. Now it reuses the logic from above.

src/sqlfluff/core/parser/segments/meta.py
```diff
@@ -17,6 +17,13 @@ class MetaSegment(RawSegment):
     is_meta = True
 
     def __init__(self, is_template=False, *args, **kwargs):
+        """Constructor for MetaSegment.
+
+        Args:
+            is_template (:obj:`bool`): A flag to indicate whether
+                this meta segment is related to a templated section.
+                This allows proper handing.
```
barrywhart, line 25:
```suggestion
                This allows proper handling.
```

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -740,7 +740,7 @@ test_pass_ignore_templated_whitespace:
     FROM my_table
   configs:
     core:
-      ignore_templated_areas: False
```
barrywhart, line 743: Did the capitalized versions work? Wasn't sure if this breaking things, or just inconsistent.

src/sqlfluff/core/parser/lexer.py
```diff
@@ -254,6 +255,242 @@ def search(self, forward_string: str) -> Optional[Tuple[int, int]]:
...
+    """
+    if (
+        last_source_slice
+        and last_source_slice.stop > source_slice.start
+        and last_source_slice.stop != source_slice.stop
+        and last_source_slice != source_slice
+    ):
+        # If we have, insert a loop marker to reflect that.
+        lexer_logger.debug("      Backward jump detected. Inserting Loop Marker")
+        # TemplateLoops should usually have a dedent before
```
barrywhart, line 278: Aren't we doing it **every** time? The word "usually" was confusing. 
```suggestion
        # TemplateLoops should have a dedent before
```

src/sqlfluff/core/parser/lexer.py
```diff
@@ -254,6 +255,242 @@ def search(self, forward_string: str) -> Optional[Tuple[int, int]]:
...
+        # and an indent after.
+        return [
+            Dedent(
+                is_template=True,
+                pos_marker=PositionMarker.from_point(
+                    last_source_slice.stop,
+                    templated_idx,
+                    templated_file,
+                ),
```
barrywhart, line 287: Nitpick: Since all 3 segments have the same `PositionMarker`, could we reuse or copy it rather than repeating these 5 lines each time?

src/sqlfluff/core/parser/lexer.py
```diff
@@ -254,6 +255,242 @@ def search(self, forward_string: str) -> Optional[Tuple[int, int]]:
...
+    templ_str = templated_file.templated_str[template_slice]
+    source_str = templated_file.source_str[source_slice]
+
+    # For reasons which aren't entirely clear right now, if there is
+    # an included literal, it will always be at the end. Let's see if it's
+    # there.
+    if source_str.endswith(templ_str):
+        existing_len = len(templ_str)
+    else:
+        existing_len = 0
```
barrywhart, line 362: I wonder if this logic is still needed. It potentially seems like a workaround from before `JinjaTracer`.

src/sqlfluff/core/parser/lexer.py
alanmcruickshank: Good point.

src/sqlfluff/core/parser/lexer.py
alanmcruickshank: It's still necessary for jinja tags which consume whitespace (like `{%- ... -%}`). Otherwise I'd agree.

src/sqlfluff/core/parser/lexer.py
alanmcruickshank: good point

test/fixtures/rules/std_rule_cases/L003.yml
alanmcruickshank: they work - it's just were getting yamllint warnings on them. I saw that when I got some actual yamllint fails - so I thought it would be better to correct them to the _recommended_ format while I was here.

src/sqlfluff/core/parser/lexer.py
barrywhart: Tricky!

barrywhart (approved):

test/fixtures/rules/std_rule_cases/L003.yml
barrywhart: Makes sense!

YAML is a funny mixture, because it has significant whitespace like Python, but its data types are from JSON.

PR #3936, opened 2022-10-08
Add templated block uuids

alanmcruickshank:
This relates to #3673, and in particular the pathway to more intelligent indentation and the migration of L003.

This also serves as the basis for how we treat templated elements in a script, especially the handling of blocks which cut across the parse tree (i.e. where the start of a jinja block, like `{% if ... %}` exists at a different level in the parsed structure than it's closing element). Having a reference to be able to associate them together is necessary in order to align their indentation.


src/sqlfluff/core/parser/lexer.py
```diff
@@ -315,12 +318,18 @@ def _generate_placeholder_segments(
     This code extracts them, adjusts the source_slice and generates
     appropriate template segments to insert.
 
+    Then block stack is to provide a consistent reference between
+    of the same expression (i.e. link an {% if .. %} and an
```
barrywhart, line 322:
```suggestion
    The block stack is to provide a consistent reference between template tags
    of the same expression (i.e. link an {% if .. %} and an
```

barrywhart (approved):
Looks good! I suggested a couple of changes to a comment, nothing major.

PR #3937, opened 2022-10-08
Enable dumping of performance information to csv.

alanmcruickshank:
To enable a bit more introspection on performance information - this enables dumping performance information to an external csv file. I'd like to do some analysis on the results from our large project a little before going much further with this as a feature - but eventually I'd love to validate each releases performance via some indices.

This might make @WittierDinosaur very happy (I hope).

To get a sense of file complexity - I've added methods for raw segments, and total segments. These get included in the csv.

For now this just add the ability to dump performance. I might backport it to some older releases locally to get an idea of performance changes over time, but I'm not going to try and re-release them (that sounds like too much work).


alanmcruickshank:
I've tested this feature on a subset of out main tails.com project, and the results looks like this: 
image

Linting and parsing are the main steps which take time. A future version of this might also include timing information for specific rules, but I think that can happen at a later date.

Interesting that for very large files the performance hit isn't crazily super-linear, and that (aside from a couple of exceptions), parsing is the main place to focus performance gains.

src/sqlfluff/cli/commands.py
```diff
@@ -514,6 +514,16 @@ def dump_file_payload(filename: Optional[str], payload: str):
     cls=DeprecatedOption,
     deprecated=["--disable_progress_bar"],
 )
+@click.option(
```
barrywhart, line 517: Should we consider adding this to `fix` and `parse` as well?

src/sqlfluff/core/linter/linting_result.py
```diff
@@ -146,6 +147,50 @@ def timing_summary(self) -> Dict[str, Dict[str, float]]:
...
+
+            for dir in self.paths:
+                for file in dir.files:
+                    writer.writerow(
+                        {
+                            "path": file.path,
+                            "source_chars": (
+                                len(file.templated_file.source_str)
+                                if file.templated_file
+                                else ""
```
barrywhart, line 173: Nitpick: Since these output fields are all integers, would `0` make more sense than `""` as the `else` value?

barrywhart:
Related to performance: I know we tried Pypy once, and it wasn't really any faster. It might be worthwhile to try it again occasionally. Some of the design changes we've made over time have made the code less dynamic, which could may Pypy more effective. It excels on simpler code. Also on code with fewer object allocations, but that's difficult to avoid with a parser.

src/sqlfluff/core/linter/linting_result.py
alanmcruickshank: My intent here was to output something more like a `null` value than a 0. It didn't take zero time, it didn't happen at all. I think zeros could mess up averages further down the line.

src/sqlfluff/cli/commands.py
alanmcruickshank: I considered this 🤔 - it's a good question.

For `parse` I thought that we get that information from `lint` already so it felt like duplication. Arguably some rules might behave differently during `fix` than during `lint` - so there's a good argument for `fix`.

On balance though, I thought for the purpose we need it for now, we get the bulk of the time from the parsing and linting steps, which are covered nicely by `lint`. Yes we get slightly different information from `fix` but not drastically.

So, my roundabout answer I think is that I'm tempted not to for now. There will almost certainly be another iteration of this (hence the `beta` warning), where we might want some more metadata. At that stage I think we re-evaluate this question again.

barrywhart: No worries, very nitpicky question.

barrywhart (approved):

PR #3939, opened 2022-10-08
T-SQL: Allow arbitrary expressions in PARTITION BY clause

### LINKED ISSUE #3938: T-SQL: PARTITION BY does not allow expression
### Search before asking

### What Happened

The query below could not be parsed.

### Expected Behaviour

The query should be parseable.

### Observed Behaviour

`sqlfluff lint -d tsql` yields:

```
== [test.sql] FAIL
L:   1 | P:  18 |  PRS | Line 1, Position 18: Found unparsable section: '()
                       | OVER(PARTITION BY CASE WHEN t.col1 = ...'
All Finished 📜 🎉!
```

### How to reproduce

```sql
SELECT ROW_NUMBER() OVER(PARTITION BY CASE WHEN t.col1 = 'value' THEN 1 END ORDER BY t.col2) rn
FROM mytable t;
```

### Dialect

`tsql`

### Version

1.3.2

### Configuration

n/a

### Code of Conduct

borchero:

This PR resolves #3938.

The `PARTITION BY` clause now allows for arbitrary expressions. Looking at the documentation, this seems to be the case in T-SQL (see https://learn.microsoft.com/en-us/sql/t-sql/queries/select-over-clause-transact-sql?view=sql-server-ver15#partition-by). However, I am not sure whether `ExpressionSegment` is too broad.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -2037,19 +2037,7 @@ class PartitionClauseSegment(ansi.PartitionClauseSegment):
...
-                    Bracketed(
-                        Ref("SelectStatementSegment"),
-                    ),
-                    Ref("FunctionSegment"),
-                    Ref("VariableIdentifierSegment"),
-                    "NULL",
-                ),
-            ),
-        ),
+        Delimited(OptionallyBracketed(Ref("ExpressionSegment"))),
```
alanmcruickshank, line 2040: I'm tempted to go for something like `OneOf(Ref("ColumnReferenceSegment"), Ref("ExpressionSegment"))` as a hybrid of the old and the new. I imagine, often this will only be a single column, and it seems a bit far to complicate the parse tree with expressions when it's not required - but for everything else (functions, select statements etc...) then I think reaching for just `expression` is the right option.

What do you think?

borchero: Fair enough, I've been approaching the grammar very much from a theoretical perspective but it seems to be useful to have this simplification in practice. Will change this!

barrywhart (approved):

PR #3942, opened 2022-10-08
Reflow 3: Reindent

### LINKED ISSUE #1284: Add possibility to apply main indentation rule without any exceptions
For the indentation main rule "For Keywords within a statement, the first root keyword of each line should be aligned" there is an exception now: "if two (or more) of these keywords are on the same line, then the second (and any further) keywords won’t raise a violation, provided that the first was correctly aligned".

Please add the functionality to fix code according to this rule without this exception.
Parameterization seems good to me.

### LINKED ISSUE #2677: L003: Jinja for-loop indentation not stable across runs (dbt, snowflake)

### LINKED ISSUE #2853: L003  over indents - Confused with Functions at CASE STMTs
### Search before asking

### What Happened

Closing / Dedents if all on one line are not understood. As the algo only sees the closing line (the one being computed for faults) as far as the first code char.

```
    SELECT
        MIN(MAX(COALESCE(
            a
        ))) AS c,
        MAX(CASE
            WHEN b THEN 1
            WHEN c THEN 1
        END) as agg
    FROM my_tbl
   ```
   ---> [wrongly corrects to]
   ```
    SELECT
        MIN(MAX(COALESCE(
            a
            ))) AS c,
        MAX(CASE
            WHEN b THEN 1
            WHEN c THEN 1
            END) as agg
    FROM my_tbl
   ```

### Expected Behaviour

In the samle above things are already correct. no need for L003 to do anything.

Under the hood I think the algo is not quite right.
Im guessing it grew organically - After a lot of read and refactoring just to be able to read it. It looks to be half and half using but also subverting the crawler.

### Observed Behaviour

At first I wondered if the over indenting was intentional, and am surprised not to find ti reported, but reading the code it claims to address this exact thing but actually its not possible based on how the code is trying

### How to reproduce

```
    SELECT
        MIN(MAX(COALESCE(
            a
        ))) AS c,
        MAX(CASE
            WHEN b THEN 1
            WHEN c THEN 1
        END) as agg
    FROM my_tbl
``` 

but also any query with many closing brackets on one line.

### Dialect

any

### Version

0.11.0

### Configuration

na

### Code of Conduct

### LINKED ISSUE #2867: L003: Fix applies strange indentation after WHERE ... BETWEEN ... (should there be limits on "hanging indent"?)
### Search before asking

### What Happened

To replicate, create test.template.sql
```
SELECT 1 + 1 AS col_1
FROM A_TABLE
WHERE THE_COL_DATE BETWEEN DATEADD(DAY, -1, '2022-01-01') AND DATEADD(DAY, -1, '2022-01-10')
  AND COL_3 > 2

```

Notice the last AND is on a new row with two spaces before.

when running: `sqlfluff fix -f test.template.sql ` this is the result:
```
SELECT 1 + 1 AS col_1
FROM A_TABLE
WHERE THE_COL_DATE BETWEEN DATEADD(DAY, -1, '2022-01-01') AND DATEADD(DAY, -1, '2022-01-10')
                                                                      AND COL_3 > 2

```

Normally i would expect the last AND to be tabbed in one time.

### Expected Behaviour

I would expect the last AND to be tabbed in one time, like this:

```
SELECT 1 + 1 AS col_1
FROM A_TABLE
WHERE THE_COL_DATE BETWEEN DATEADD(DAY, -1, '2022-01-01') AND DATEADD(DAY, -1, '2022-01-10')
    AND COL_3 > 2

```

### Observed Behaviour

Multiple tabs are applied before the last AND.

### How to reproduce

Mentioned above.

### Dialect

tested with snowflake dialect

### Version

sqlfluff, version 0.11.0

### Configuration

[sqlfluff]
verbose = 1
dialect = snowflake
templater = jinja
exclude_rules = L027,L031,L032,L036,L044,L046,L034,L050
output_line_length = 121
sql_file_exts=.sql

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 250
indent_unit = space
comma_style = trailing
allow_scalar = True

[sqlfluff:rules:L042]
forbid_subquery_in = both

capitalisation_policy = upper

[sqlfluff:rules:L014]

[sqlfluff:rules:L030]  # function names

### Code of Conduct

### LINKED ISSUE #3069: L003: Doubled indents are tolerated
### Search before asking

### What Happened

L003 seems to have an odd behaviour where double indents are tolerated (and sometimes even `fix` ed to).

Consider the following two examples:

```sql
SELECT GREATEST(1, 2,
        3, 4,
    5, 6
)
```

After running `sqlfluff fix`

```sql
SELECT GREATEST(1, 2,
        3, 4,
        5, 6
)
```

Everything gets indented to 2x 4 spaces.

Interestingly, this seems to be specific to 2x. For example tripple (or quadruple) indents get indented correctly

```sql
SELECT GREATEST(1, 2,
            3, 4,
    5, 6
);

SELECT GREATEST(1, 2,
                    3, 4,
    5, 6
);
```

After running `sqlfluff fix`

```sql
SELECT GREATEST(1, 2,
    3, 4,
    5, 6
);

SELECT GREATEST(1, 2,
    3, 4,
    5, 6
);
```

### Expected Behaviour

Doubled indents should not be tolerated (and certainly not suggested as a fix 😉 )

### Observed Behaviour

SQLFluff does not complain about doubled indents.

### How to reproduce

https://online.sqlfluff.com/fluffed?sql=H4sIANP3U2IC_wt29XF1DlFwD3J1DHENDtEw1FEw0uHlUoACYx0FEyjXVEfBjJdL05qXi5crGL8uynTiNQEAsiH7abAAAAA%3D&dialect=ansi

### Dialect

ansi

### Version

0.12 -> webversion

### Configuration

default

### Code of Conduct

### LINKED ISSUE #3133: L003: Jinja directive outside of SELECT interfers with Jinja conditionals inside

### LINKED ISSUE #3185: L003: Unfixable Jinja templates (auto fix loop)
### Search before asking

### What Happened

On certain types of templated queries the auto fixer of L003 runs into a loop where indents can added and removed on each time the fix is run. The linter will always report a linting error and users have no other choice than ignoring the rule for that file (or line).

### Expected Behaviour

Stable behaviour of L003 with one result being considered "correct". This should also be the result of Auto Fix.

### Observed Behaviour

Take the following example:

```sql
{% for item in some_list -%}
    SELECT *
    FROM some_table
{{ "UNION ALL\n" if not loop.last }}
{%- endfor %}
```

The Autofix will indent line 4 to

```sql
{% for item in some_list -%}
    SELECT *
    FROM some_table
    {{ "UNION ALL\n" if not loop.last }}
{%- endfor %}
```

However, a second call will revert back to original.

### How to reproduce

Run lint/fix on

```sql
{% for item in some_list -%}
    SELECT *
    FROM some_table
{{ "UNION ALL\n" if not loop.last }}
{%- endfor %}
```

with the config provided below.

### Dialect

ansi

### Version

`main` branch from Github, commit `656a958146e9d415a39e77da7a063cd304fd56d3`

### Configuration

```
some_list = ["abc", "def"]
```

### Code of Conduct

### LINKED ISSUE #3246: L003: Fix indentation doesn't work on first try
### Search before asking

### What Happened

Given the following SQL:
```
create procedure Testy as
begin
  begin try
    if @something = 1
      begin
        set @result = 1
      end
    else
      begin
        set @result = 0
      end

    declare @something int

    set @something = 5

    select *
      from Dbo
      where Blah is not null
  end try
  begin catch
    select * from Dbo
  end catch
end
```

Note that almost everything (starting with `begin try`) needs another indentation level.

### Expected Behaviour

with `sqlfluff fix`, the indentation is corrected

### Observed Behaviour

For most lines, it is fixed correctly. However, some lines (`end`, `set @something`, `where`) are not. Also note that their indentation is 1 space, which is not even a valid value.
```
create procedure Testy as
begin
  begin try
    if @something = 1
      begin
        set @result = 1
      end
    else
      begin
        set @result = 0
      end

    declare @something int

    set @something = 5

    select *
    from Dbo
 where Blah is not null
  end try
  begin catch
    select * from Dbo
  end catch
end

```

### How to reproduce

See above

### Dialect

T-SQL

### Version

Latest `main`

### Configuration

```
[sqlfluff:rules]
tab_space_size = 2
indent_unit = space
```

### Code of Conduct

### LINKED ISSUE #3247: Indentation stays incorrect
### Search before asking

### What Happened

This can happen in combination with #3246, making it really annoying.

For `order by` clauses with several fields, the indentation seems to be ambiguous: It can be normally indented (one indent more than `order`), or aligned with the previous field.
If the indentation of the second field is aligned, but the `order by` clause is wrong, `sqlfluff fix` switches between to indentations, but cannot fix the whole statement.

It seems that this is related to some interference between L003 and L039. I can work around this by excluding L039, but this can only be temporary.

As an example, take this:
```
begin
  begin
    select Blah
    from Blubb
 order by Field1 desc,
             Field2 desc
  end
end
```
Note that `Field2` would be correctly aligned with `Field1` if `order by Field1` were correctly indented.

### Expected Behaviour

`order by` should be indented correctly. Then, I'd prefer to have `Field2` just one more indent than the previous line, but it is also acceptable if the fields are aligned.

### Observed Behaviour
After `sqlfluff fix`:
```
begin
  begin
    select Blah
    from Blubb
 order by Field1 desc,
         Field2 desc
  end
end
```
`Field2` is neither correctly indented nor aligned.

After the next `sqlfluff fix`, we're back at the beginning:
```
begin
  begin
    select Blah
    from Blubb
 order by Field1 desc,
             Field2 desc
  end
end
```

### Other observations
Correctly indenting `Field2` manually also solves this problem.
Removing just one space character from `Field2`'s indentation will take two iterations, but then the whole statement is correctly formatted and stable. Maybe it is also relevant that the broken indentations are 17 (aligned with previous fields) and 13, both odd numbers?

### How to reproduce

See above

### Dialect

T-SQL

### Version

Latest `main`

### Configuration

```
[sqlfluff:rules]
tab_space_size = 2
indent_unit = space
```

### Code of Conduct

### LINKED ISSUE #3836: Inappropriate hanging indent introduced by L003
### Search before asking

### What Happened

`WHEN` conditions within `CASE` statements get dedented to be inline with the `CASE`.

I've tried configuring handing indents but that doesn't seem to resolve the problem. 

```
select
1,
case 
when 1 = 1
then 2
when 2 = 1
then 3 end as foo
from bar
```

the

### Expected Behaviour

Given the following query

```
SELECT
 foo
,  CASE
 WHEN 1 = 1
 THEN 2
 END AS example
FROM tbl
```

The query should return

```
SELECT
    foo
    ,  CASE
           WHEN 1 = 1
           THEN 2
    END AS example
FROM tbl
```

### Observed Behaviour

```
SELECT
    foo
    ,  CASE
       WHEN 1 = 1
          THEN 2
    END AS example
FROM tbl
```

### How to reproduce
```
import textwrap
from sqlfluff.core import Linter
from sqlfluff.core.config import FluffConfig

linter = Linter(
    config=FluffConfig(
        configs={
            "rules": {
                "comma_style": "leading",
                "tab_space_size": 2,
                "L003": {"hanging_indents": True, "indent_unit": "space", "tab_space_size": 2},
            },
        }
    )
)
```

```
linter.lint_string_wrapped(
    textwrap.dedent(
        """\
        SELECT
          foo
          , case when 1 = 1 then 2 end AS example
        FROM tbl
        """
    )
).as_records()
```

### Dialect

snowflake

### Version

1.2.1

### Configuration
```
[sqlfluff]
verbose = 0
nocolor = False
# Or run 'sqlfluff dialects'
dialect = snowflake
templater = jinja
rules = all
exclude_rules = L031,L057,L051,L060
recurse = 0
output_line_length = 80
runaway_limit = 10
ignore = None
ignore_templated_areas = True
encoding = autodetect
disable_noqa = False
fix_even_unparsable = False
# This limit skips files over a certain character length
# and warns the user what has happened.
# Set this to 0 to disable.
large_file_skip_char_limit = 20000

[sqlfluff:indentation]
# See https://docs.sqlfluff.com/en/stable/indentation.html
indented_joins = False
indented_ctes = False
indented_using_on = True
indented_on_contents = True
template_blocks_indent = True

[sqlfluff:templater]
unwrap_wrapped_queries = True

[sqlfluff:templater:jinja]
apply_dbt_builtins = True

[sqlfluff:templater:dbt]
project_dir = .
profiles_dir = .
profile = snowflake-dbt
target = prod

[sqlfluff:templater:dbt:context]
# Put vars defined in dbt_project.yml here


# Put standard jinja params here

[sqlfluff:rules]
tab_space_size = 2
max_line_length = 80
indent_unit = space
comma_style = leading
allow_scalar = True

[sqlfluff:rules:L003]
hanging_indents = True

[sqlfluff:rules:L007]
operator_new_lines = after

[sqlfluff:rules:L010]
# Keywords
capitalisation_policy = upper
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L011]
aliasing = explicit

[sqlfluff:rules:L012]
aliasing = explicit

[sqlfluff:rules:L014]
# Unquoted identifiers
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L016]
# Line length
ignore_comment_lines = False

[sqlfluff:rules:L027]
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L026]
force_enable = False

[sqlfluff:rules:L028]
force_enable = False

[sqlfluff:rules:L029]
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L030]
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L031]
force_enable = False

[sqlfluff:rules:L036]
wildcard_policy = single

[sqlfluff:rules:L038]

[sqlfluff:rules:L040]
capitalisation_policy = upper
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L042]
forbid_subquery_in = join

[sqlfluff:rules:L047]
prefer_count_1 = False
prefer_count_0 = False

[sqlfluff:rules:L051]

[sqlfluff:rules:L052]
multiline_newline = False

[sqlfluff:rules:L054]

[sqlfluff:rules:L057]
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L059]
ignore_words = None
ignore_words_regex = None
force_enable = False

[sqlfluff:rules:L062]
blocked_words = None
blocked_regex = None

[sqlfluff:rules:L063]
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L064]
force_enable = False

[sqlfluff:rules:L066]
min_alias_length = None
max_alias_length = None
```

### Code of Conduct

### LINKED ISSUE #3865: New rule suggestion: Alignment of start and `endX` jinja lines
### Search before asking

### Description

Currently I see no rule that has requirements around the start/end blocks for `if` and `for` blocks within jinja/dbt code. This means that the below code will lint with no errors when run with the dbt templater (when the if condition is set to `true` the only errors generated are due to the leading comma, which is expected).

```jinja
select *
  {% if false -%}
            , current_date() AS cur_date
          {%- endif %}
from dual
```

when I would at a minimum expect it to look like
```jinja
select *
  {% if false -%}
            , current_date() AS cur_date
  {%- endif %}
from dual
```

Despite this it is common standard that the start/end clauses are aligned when multiline and it seems to be the case at least within `dbt` code examples that the inner code is also starts at the same indentation as the block itself (but this would potentially make sense as a different rule maybe?). This would lead to the below

```jinja
select *
{% if false -%}
, current_date() AS cur_date
{%- endif %}
from dual
```

Outside of the block itself I am unsure what standards exist around where the block itself is aligned to given the context of the rest of the code.

### Use case

This would allow for better linting and fixing of jinja heavy code beyond the rule for spaces in {{ func }} calls, enabling less time manually reviewing these components of the code.

### Dialect

jinja/dbt, in theory it applies to all database dialects that support these.

### Code of Conduct

### LINKED ISSUE #3988: L001: Using jinja & dbt_utils results in unwanted unneccessary trailing whitespace with sqlfluff 1.3.2
### Search before asking

### What Happened

I'm trying to move to latest dbt & sqlfluff version. I now use `dbt 1.2.1`  & `sqlfluff 1.2.1 `, and trying to move to `dbt 1.3.0` & `sqlfluff 1.3.2` . However when using the latest versions I keep running into a lot of issues like I mention in `Observed Behavior` that were not present in the versions I currently use.

### Expected Behaviour

The expected behavior would be to have no linting error in this scenario like I used to have in version `sqlfluff 1.2.1`.

### Observed Behaviour

My `dim_dates.sql` file that I'm linting:

```sql
{{ config(materialized="table") }}

with
date_spine as (
    select date_actual
    from {{ ref('dates') }}
),

calculated as (
    select
        date_actual,
        extract(isodow from date_actual) as day_of_week,
        extract(day from date_actual) as day_of_month,
        extract(doy from date_actual) as day_of_year,
        to_char(date_actual, 'TMDay') as day_name,
        extract(month from date_actual) as month_actual,
        to_char(date_actual, 'TMMonth') as month_name,
        extract(year from date_actual) as year_actual,
        extract(week from date_actual) as week_of_year,
        'Q' || to_char(date_actual, 'Q') as quarter_actual,
        to_char(date_actual, 'iyyy/IW') as year_week,
        to_char(date_actual, 'yyyy/mm') as year_month,
        to_char(date_actual, 'yyyy/"Q"Q') as year_quarter,
        case when extract(isodow from date_actual) in (6, 7) then 0 else 1 end as is_weekday,
        case when extract(isodow from date_actual) in (6, 7) then 1 else 0 end as is_weekend,
        {{ dbt_utils.current_timestamp() }} as dbt_created_at
    from date_spine
)

select *
from calculated
```

The result of linting:

```bash
== [models/datamarts/dim_dates/dim_dates.sql] FAIL                                                                                                                                                                       
L:  26 | P:   1 | L001 | Unnecessary trailing whitespace.
```

It is complaining about the following. It does not allow a space on the line with column `dbt_created_at`. It now is:

```sql
        {{ dbt_utils.current_timestamp() }} as dbt_created_at
```

When running `sqlfluff fix` it turns into:

```sql
{{ dbt_utils.current_timestamp() }} as dbt_created_at
```

### How to reproduce

You need two sql files & a dbt package.

`dates.sql`:

```sql
{{ config(materialized="table") }}

with
date_spine as (
  {{ dbt_utils.date_spine(
      start_date="to_date('01/01/2010', 'mm/dd/yyyy')",
      datepart="day",
      end_date="CURRENT_DATE + INTERVAL '12 years'"
     )
  }}
),

all_dates as (
    select
        date_day as date_actual,
        {{ dbt_utils.current_timestamp() }} as dbt_created_at
    from date_spine
)

select *
from all_dates

```

`dim_dates.sql`:

```sql
{{ config(materialized="table") }}

with
date_spine as (
    select date_actual
    from {{ ref('dates') }}
),

calculated as (
    select
        date_actual,
        extract(isodow from date_actual) as day_of_week,
        extract(day from date_actual) as day_of_month,
        extract(doy from date_actual) as day_of_year,
        to_char(date_actual, 'TMDay') as day_name,
        extract(month from date_actual) as month_actual,
        to_char(date_actual, 'TMMonth') as month_name,
        extract(year from date_actual) as year_actual,
        extract(week from date_actual) as week_of_year,
        'Q' || to_char(date_actual, 'Q') as quarter_actual,
        to_char(date_actual, 'iyyy/IW') as year_week,
        to_char(date_actual, 'yyyy/mm') as year_month,
        to_char(date_actual, 'yyyy/"Q"Q') as year_quarter,
        case when extract(isodow from date_actual) in (6, 7) then 0 else 1 end as is_weekday,
        case when extract(isodow from date_actual) in (6, 7) then 1 else 0 end as is_weekend,
        {{ dbt_utils.current_timestamp() }} as dbt_created_at
    from date_spine
)

select *
from calculated
```

`packages.yml`:

```yml
packages:
  - package: dbt-labs/dbt_utils
    version: 0.8.6
```

My guess would be that this is irrespective of sqlfluff configuration, but will deliver my `.sqlfluff` below.

### Dialect

postgres

### Version

Python version: `3.10.6`
sqlfluff version: `1.3.2`
dbt version: `1.3.0`

### Configuration

`.sqlfluff`:

```
[sqlfluff]
verbose = 0
nocolor = False
# Or run 'sqlfluff dialects'
dialect = postgres
templater = dbt
rules = all
exclude_rules = L031
recurse = 0
output_line_length = 160
runaway_limit = 10
ignore = None
ignore_templated_areas = True
encoding = autodetect
disable_noqa = False
fix_even_unparsable = False

[sqlfluff:indentation]
# See https://docs.sqlfluff.com/en/stable/indentation.html
indented_joins = False
indented_ctes = False
indented_using_on = True
template_blocks_indent = True

[sqlfluff:templater:dbt]
project_dir = .
profiles_dir = ~/.dbt

[sqlfluff:templater:jinja]
apply_dbt_builtins = True

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 160
indent_unit = space
comma_style = trailing
allow_scalar = True

[sqlfluff:rules:L003]
hanging_indents = True

[sqlfluff:rules:L007]
operator_new_lines = after

[sqlfluff:rules:L010]
# Keywords
capitalisation_policy = lower
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L011]
aliasing = explicit

[sqlfluff:rules:L012]
aliasing = explicit

[sqlfluff:rules:L014]
# Unquoted identifiers
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L016]
# Line length
ignore_comment_lines = True
ignore_comment_clauses = True

[sqlfluff:rules:L026]
force_enable = False

[sqlfluff:rules:L028]
force_enable = True
single_table_references = unqualified

[sqlfluff:rules:L029]
ignore_words = date,year_month,name
ignore_words_regex = None

[sqlfluff:rules:L030]
# Function names
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L031]
force_enable = False

[sqlfluff:rules:L038]
# Trailing commas

[sqlfluff:rules:L040]
# Null & Boolean Literals
capitalisation_policy = lower
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L042]
forbid_subquery_in = both

[sqlfluff:rules:L047]
prefer_count_1 = False
prefer_count_0 = False

[sqlfluff:rules:L051]
# Fully qualify JOIN clause

[sqlfluff:rules:L052]
multiline_newline = False

[sqlfluff:rules:L054]
group_by_and_order_by_style = explicit

[sqlfluff:rules:L057]
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L059]
ignore_words = None
ignore_words_regex = None
force_enable = False

[sqlfluff:rules:L062]
blocked_words = None

[sqlfluff:rules:L063]
# Data Types
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L064]
force_enable = False
```

`.sqlfluffignore`:

```
target/
dbt_packages/
macros/
.cache/
venv/
```

### Code of Conduct

### LINKED ISSUE #3989: Bug: Self crawling rules don't respect unparsable code.
### Search before asking

### What Happened

After upgrading from 1.2.1 to 1.3.0, previously linted files are showing `L003` errors.

### Expected Behaviour

Linting should not introduce regressions when upgrading sqlfluff within a major version (1.2.1 → 1.x).

Quite some tools (e.g. Flyway for database migrations) produce and verify file check sums, making it impossible to "fix" an older SQL file.

### Observed Behaviour

```bash
$ sqlfluff lint --nocolor --verbose --dialect postgres 2022-10-25.indentation-regression.sql
==== sqlfluff ====
sqlfluff:                1.3.0 python:                 3.8.10
implementation:        cpython verbosity:                   1
dialect:              postgres templater:               jinja
==== readout ====

=== [ path: 2022-10-25.indentation-regression.sql ] ===

== [2022-10-25.indentation-regression.sql] FAIL                                                                                                                                   
L:   6 | P:   2 | L003 | Expected 1 indentation, found 0 [compared to line 05]
L:   8 | P:   1 | L003 | Expected 1 indentation, found 0 [compared to line 05]
==== summary ====
violations:        2 status:         FAIL
All Finished!
```

### How to reproduce

Reproduce with the following file:
```sql
CREATE OR REPLACE AGGREGATE range_agg_preserve_gaps(TSTZRANGE)  -- noqa
(
    sfunc = tstzrange_arr_merge,
    stype = TSTZRANGE[],
    initcond = '{}'
);

CREATE TABLE test (my_col TEXT);
```

Even when excluding the whole (unparseable) `AGGREGATE` block, the regression remains for the flawless `CREATE TABLE`:
```sql
-- noqa: disable=all
CREATE OR REPLACE AGGREGATE range_agg_preserve_gaps(TSTZRANGE)
(
    sfunc = tstzrange_arr_merge,
    stype = TSTZRANGE[],
    initcond = '{}'
);
-- noqa: enable=all

CREATE TABLE test (my_col TEXT);
```

Result:
```
== [2022-10-25.indentation-regression.sql] FAIL                                                                                                                                   
L:  10 | P:   1 | L003 | Expected 1 indentation, found 0 [compared to line 06]
==== summary ====
```

### Dialect

postgres

### Version

Regression starts with 1.3.0
```
sqlfluff:                1.3.0 python:                 3.8.10
implementation:        cpython verbosity:                   1
dialect:              postgres templater:               jinja
```

...but continues through 1.3.2:
```
sqlfluff:                1.3.2 python:                 3.8.10
implementation:        cpython verbosity:                   1
dialect:              postgres templater:               jinja
```

### Configuration

_none_

### Code of Conduct

alanmcruickshank:
Ok. This is a bit of a monster. Make sure to read this description in  detail before reviewing to understand what's going on. This makes progress on #3673.

*Potential things that might seem odd*:
- This version of `L003` effectively brings together the functionality of `L002` & `L004` (and some of the functionality of `L050`). I haven't resolved that overlap here, I suggest that happens separately as part of a wider rules reshuffle. I've made a note about that in #4030.

This PR:
- fixes #3185 & fixes #2677 
- fixes #2867 
- fixes #3865
- fixes #3069 
- fixes #3836 [by also tweeking indentation in case statements and refining the reordering of meta segments. It's still necessary, but now we need to ensure we can respect order too].
- fixes #3988
- fixes #1284 
- fixes #2853
- fixes #3246 & fixes #3247
- fixes #3133
- fixes #3989 (now that #4032 is merged)
- Introduces `.reindent()` and the associated configuration to achieve that.
  - This also removes the option of hanging indents and therefore dramatically simplifies the algorithm for reindenting code.
  - If a hanging indent is detected on the way up, or the way down, new line breaks are introduced appropriately to avoid a hanging indent.
  - There's currently an exception in the code for semicolons (`;`) so that we don't conflict with L052, but I might revisit this at a later date.
- Migrates L003 to use `.reindent()`.
  - As part of that, the engine to generate `description` strings for the violations is totally new. Because the new indentation algorithm isn't really _comparing_ one line to another, the text is now much more explicit e.g. `Expected indent of 8 spaces`, so that the user knows exactly what the linter wants.
- Includes the now closed #3999, which rethinks how placeholders are included in templated code (this is most of the changes in the lexer):
  - This breaks apart the joined templated elements so that we now may have several in a row rather than one big lump. e.g. `{% if false %}...unused code...{% endif %}` would now be three placeholders: `{% if false %}`, `...unused code...` & `{% endif %}`, which allows much more granular placement of `Indent` and `Dedent` segments.
  - This significantly refines the fixing routines around templated elements to account for the change and make them more precise.
  - This intentionally includes placeholders for consumed whitespace (i.e. the whitespace which might happen before `   {%- ...`), and then takes these into account when running indentation code so that we can accurately indent _in the source file_.
- Given the new indentation logic, many test cases are updated to account for that. I've also introduced much more extensive testing of the underlying code to get there as part of `reindent_test.py`.


plugins/sqlfluff-templater-dbt/test/fixtures/dbt/dbt_project/snapshots/issue_1771.sql.after
```diff
@@ -1,6 +1,6 @@
 {% snapshot dim_aggregated_brand_hierarchy_snapshot %}
 
-{{
+    {{
```
alanmcruickshank, line 3: This is the only slight strangeness in the new algorithm, is that it wants to indent this tag because it's within the snapshot, but the tag itself has a newline within it and it's not smart enough to add newlines and indents *inside* the jinja tag.

Any ideas on good solutions here?

test/fixtures/linter/autofix/ansi/002_indentation/after.sql
```diff
@@ -8,14 +8,18 @@ SELECT
...
+    ) as actually_ok,
+    a.line + (
+        a.with
+        + a.bad_hanging_indent
+    ) as problem,
     a.line + (
         a.something_indented_well
         + least(
-    	    a.good_example,
+            a.good_example,
```
barrywhart, line 22: I wonder what changed here? Tab vs space, maybe?

test/fixtures/linter/autofix/ansi/004_indentation/after.sql
```diff
@@ -9,20 +9,23 @@ WITH audience_counts AS (
         list_emails.active != 'D'
     GROUP BY
         user_id,
-        list_id)
+        list_id
+)
```
barrywhart, line 13: I'm curious about this change. I tend to prefer fewer lines as long as things don't get too cluttered. (So I can see more code on the screen without scrolling.)

test/fixtures/linter/autofix/ansi/004_indentation/after.sql
```diff
@@ -9,20 +9,23 @@ WITH audience_counts AS (
...
     audience,
     CASE
         WHEN audience > 0 AND audience <= 200 THEN '< 200'
-        WHEN audience > 200
-             AND audience <= 3000
-             -- NB: This one is a hanging indent, which should pass.
-             AND audience <= 2000 THEN '200 - 2,000'
+        WHEN
+            audience > 200
+            AND audience <= 3000
```
barrywhart, line 23: Is there a possible configuration which would have done this?
```
            audience > 200 AND audience <= 3000
```

test/fixtures/linter/autofix/ansi/007_with_clause/after.sql
```diff
@@ -1,5 +1,5 @@
 -- Dealing with complicated indents before with clauses.
-    	WITH cte as (
-            select a from tbla
-        )
-        select a from cte
+WITH cte as (
+    select a from tbla
+)
+select a from cte
```
barrywhart, line 5: Nice! The old version is 🤯.

test/fixtures/linter/autofix/bigquery/003_templating/after.sql
```diff
@@ -14,14 +14,14 @@ SELECT
...
+    {% for action in considered_actions %}
+        {% if loop.first %}
+            {{action}}_raw_effect_sizes
+        {% else %}
+        JOIN
+            {{action}}_raw_effect_sizes
+            USING
+                ({{corr_states}})
+        {% endif %}
+    {% endfor %}
```
barrywhart, line 26: 💯

test/fixtures/linter/autofix/bigquery/004_templating/after.sql
```diff
@@ -27,8 +27,10 @@ raw_effect_sizes AS (
             -- NOTE: The L003 fix routine behaves a little strangely here around the templated
             -- code, specifically the indentation of STDDEV_POP and preceding comments. This
             -- is a bug currently with no obvious solution.
-            , SAFE_DIVIDE(SAFE_MULTIPLY(CORR({{metric}}_rate_su, {{action}}), STDDEV_POP({{metric}}_rate_su)),
-                STDDEV_POP({{action}})) AS {{metric}}_{{action}}
+            , SAFE_DIVIDE(
+                SAFE_MULTIPLY(CORR({{metric}}_rate_su, {{action}}), STDDEV_POP({{metric}}_rate_su)),
+                STDDEV_POP({{action}})
+            ) AS {{metric}}_{{action}}
```
barrywhart, line 33: 👍

test/fixtures/linter/autofix/bigquery/006_fix_ignore_templating/after.sql
```diff
@@ -11,7 +11,8 @@ SELECT * EXCEPT({% include query %}) FROM
             FROM tbl1
             {% endif %}
         INNER JOIN tbl2
-            ON tbl1.the_name = tbl2.the_name
+            ON
+                tbl1.the_name = tbl2.the_name
```
barrywhart, line 15: Possible to keep these on the same line?

I realize some of these questions are nitpicky. I wonder what the overall effect of these changes looks like on large, real-world SQL files. My main concern is if we're adding lots of additional lines. I once used a formatter that easily increased the number of lines 20-40%. Made the code "prettier" but actually much harder to read, because lots of scrolling.

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -227,19 +227,29 @@ test_pass_indented_on_contents_default:
         s.b
     FROM r
     JOIN s
-        ON r.a = s.a
```
barrywhart, line 230: Since it would all fit on one line, could we generate:
```
        ON r.a = s.a AND true
```
?

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -714,16 +747,20 @@ test_tsql_function:
...
+        SET
+            @ISOweek = DATEPART(wk, @DATE) + 1
             - DATEPART(wk, CAST(DATEPART(yy, @DATE) AS char(4)) + '0104');
         --Special cases Jan 1-3 may belong to the previous year
         IF (@ISOweek = 0)
-            SET @ISOweek = dbo.ISOWEEK(CAST(DATEPART(yy, @DATE) - 1
+            SET @ISOweek = dbo.ISOWEEK(CAST(
+                DATEPART(yy, @DATE) - 1
                 AS char(4)
-                ) + '12' + CAST(24 + DATEPART(day, @DATE) AS char(2))) + 1;
```
barrywhart, line 723: 👍

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -848,41 +885,51 @@ test_fail_fix_template_indentation_2:
 test_pass_tsql_update_indent:
   pass_str: |
     update Extracts.itt_parm_base
-        set DateF = convert(varchar, @from_date, 112),
-            DateT = convert(varchar, @to_date, 112)
+    set
+        DateF = convert(varchar, @from_date, 112),
+        DateT = convert(varchar, @to_date, 112)
```
barrywhart, line 890: 👍

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -1179,3 +1215,104 @@ test_tsql_outer_apply_indentation_fix:
...
+        SELECT *
+    FROM some_table
+    {{ 'UNION ALL\n' if not loop.last }}
+    {%- endfor %}
+  fix_str: |
+    {% for item in [1, 2] -%}
+        SELECT *
+        FROM some_table
+        {{ 'UNION ALL\n' if not loop.last }}
+    {%- endfor %}
```
barrywhart, line 1248: 💯

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -1179,3 +1215,104 @@ test_tsql_outer_apply_indentation_fix:
...
+
+test_pass_consuming_whitespace_stable:
+  # Test for stability in fixes with loops and consuming tags.
+  # https://github.com/sqlfluff/sqlfluff/issues/3185
+  pass_str: |
+    {% for item in [1, 2] -%}
+        SELECT *
+        FROM some_table
+        {{ 'UNION ALL\n' if not loop.last }}
+    {%- endfor %}
```
barrywhart, line 1258: Nice!

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -1179,3 +1215,104 @@ test_tsql_outer_apply_indentation_fix:
...
+    foo
+    , CASE
+    WHEN 1 = 1
+    THEN 2
+    END AS example
+    FROM tbl
+  fix_str: |
+    SELECT
+      foo
+      , CASE
```
barrywhart, line 1286: Possible to put the entire `CASE` on one line?
```
      , CASE WHEN 1 = 1 THEN 2 END AS example
```

test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -1179,3 +1215,104 @@ test_tsql_outer_apply_indentation_fix:
...
+                    when 1 = 1
+                        {% if json_key in ["b"] %}
+                            then 0
+                        {% else %}
+                            then 1
+                        {% endif %}
+                    else null
+                end as {{ json_key }}_suffix{% if not loop.last %}, {% endif %}
+            {% endfor %}
+    )
```
barrywhart, line 1315: 👍

barrywhart (commented):
I only reviewed the test cases for now. Overall, things look good. A few questions, mainly around "could we put more code on one line"?

test/fixtures/linter/autofix/ansi/002_indentation/after.sql
alanmcruickshank: yeah - `reindent` also coerces indents to be of the right composition too.

test/fixtures/linter/autofix/ansi/004_indentation/after.sql
alanmcruickshank: Not yet - but let's handle this case as part of "long lines", which I think is the next step. For now, reindent only _adds_ newlines, it never takes them away.

test/fixtures/linter/autofix/ansi/004_indentation/after.sql
alanmcruickshank: So specific case with brackets I think is right, it ensures that opening bracket are at the same indent as their closing token.

test/fixtures/linter/autofix/bigquery/006_fix_ignore_templating/after.sql
alanmcruickshank: Yeah - this is another example of where I think an implicit indent might be appropriate. I'll add `ON` clauses into #4054 . I think being able to configure this one would be very useful. A newline there is _strictly_ the most correct, but I also grant that it feels like an unnecessary newline for many users.

test/fixtures/rules/std_rule_cases/L003.yml
alanmcruickshank: See comment on currently not _removing_ newlines. Let's handle that as part of "long lines".

test/fixtures/rules/std_rule_cases/L003.yml
alanmcruickshank: Ditto never removing newlines.

alanmcruickshank:
@barrywhart - Thanks for the first review. I think the only tricky points are:
1. Sometimes there being an opportunity to _remove_ newlines and condense expressions onto fewer lines. I'm planning on addressing this in the next phase of reflow (potentially still within `.reindent()`) but I knot that's still to do.
2. The places which would be covered by implicit indents. #4054 .

Both of which I'm going to suggest as builds _on top of this work after merging_, given the current outputs are functional, if not yet optimal.

I'll keep working on both of those points, but if you're able to do a code review in the meantime then we can get this merged 👍 .

test/fixtures/linter/autofix/ansi/004_indentation/after.sql
barrywhart: Oh cool. Sorry, some of my feedback comes from not following the big picture plan.

I'll try and review the code part of the PR this weekend.

barrywhart:
> if you're able to do a code review in the meantime then we can get this merged 👍 .

Definitely!

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+
+    We do this to ensure that templated lines are _somewhat_ consistent.
+
+    Total consistency is very hard, given templated elements
+    can be used in a wide range of places. What we do here is
+    to try and take a somewhat rules based approach, but also
+    one which should fit mostly with user expectations.
+
+    To do this we have three scenarios:
+    1. Template tags area already on the same indent.
```
barrywhart, line 213:
```suggestion
    1. Template tags are already on the same indent.
```

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+       does so without requiring intervention on the parsed
+       file.
+    3. Template tags which actively cut across the tree (i.e.
+       start and end tags aren't at the same level and can't
+       be hoisted). In this case the tags should be indented
+       at the lowest indent of the matching set.
+
+    In doing this we have to attempt to match up template
+    tags. This might fail. As we battle-test this feature
+    there may be some interesting bugs which come up!
```
barrywhart, line 227: 🍿

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+        reflow_logger.debug("    Overlap: %s, Limit: %s", overlap, limit_indent)
+        # Is there a mutually agreeable option?
+        if overlap:
+            # Go for the deeper option if there's flexibility, because this
+            # will usually involve moving the fewest options.
+            best_indent = max(overlap)
+            reflow_logger.debug(
+                "    Case 2: Best: %s, Overlap: %s", best_indent, overlap
+            )
+        # If no overlap, it's case 3
```
barrywhart, line 334: Hurray! I remember talking about this. Very good stuff.

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+
+    We do this to ensure that lines with comments are aligned to
+    the following non-comment element.
+    """
+    reflow_logger.debug("# Revise comment lines.")
+    # new_lines: List[_ReindentLine] = []
+    comment_line_buffer: List[int] = []
+
+    # Slice to avoid copying
+    for idx, line in enumerate(lines[:]):
```
barrywhart, line 366: How is this different from just `enumerate(lines)`? As written, it constructs a shallow copy of `lines`, but in either case, the underlying list elements are the same.

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+                    "  Comment Only Line: %s. Anchoring to %s", comment_line_idx, idx
+                )
+                # Mutate reference lines to match this one.
+                lines[
+                    comment_line_idx
+                ].initial_indent_balance = line.initial_indent_balance
+            # Reset the buffer
+            comment_line_buffer = []
+
+    # Any trailing comments should be anchored the baseline.
```
barrywhart, line 383:
```suggestion
    # Any trailing comments should be anchored to the baseline.
```

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+) -> List[LintFix]:
+    """Evalute a single set of indent points on one line.
+
+    NOTE: This mutates the given `elements` and `forced_indents` input to avoid
+    lots of copying.
+
+    Order of operations:
+    1. Evaluate the starting indent for this line.
+    2. For point which aren't line breaks in the line, we evaluate them
+       to see whether they *should* be. We seperately to missing indents
```
barrywhart, line 517: This reads a little funny -- can you review for correctness?
> We seperately to missing indents

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+        # That could be added here, but it's an edge case which can be done
+        # at a later date easily. For now it won't get picked up.
+        for indent_seg in elements[0].segments[::-1]:
+            if indent_seg.is_type("whitespace"):
+                break
+        # Handle edge case of no whitespace, but with newline.
+        if not indent_seg.is_type("whitespace"):
+            indent_seg = None
+        else:
+            NotImplementedError(
```
barrywhart, line 577: I think you want to _raise_ the error, otherwise execution will continue below.
```suggestion
            raise NotImplementedError(
```

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+    initial_point = cast(ReflowPoint, elements[indent_points[0].idx])
+
+    if current_indent != desired_starting_indent:
+        reflow_logger.debug(
+            "    Correcting indent @ line %s. Existing indent: %r -> %r",
+            elements[indent_points[0].idx + 1].segments[0].pos_marker.working_line_no,
+            current_indent,
+            desired_starting_indent,
+        )
+        # Initial point gets special handling it it has no newlines.
```
barrywhart, line 615: The word "it" is duplicated. Did you mean this?
```suggestion
        # Initial point gets special handling if it has no newlines.
```

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+                + indent_points[-1].indent_trough
+            )
+        else:
+            closing_trough = (
+                indent_points[-1].initial_indent_balance
+                + indent_points[-1].indent_impulse
+            )
+
+        # On the way up we're looking for whether the ending balance
+        # was an untaken indent on the way up.
```
barrywhart, line 652: "On the way up" is repeated at the beginning and end of the sentence. Remove one of them?

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+        if (
+            # Edge case: if closing_balance > starting balance
+            # but closing_trough isn't, then we shouldn't insert
+            # a new line. That means we just dropped back down to
+            # close the untaken newline.
+            closing_trough > starting_balance
+            and closing_trough in indent_points[-1].untaken_indents
+        ):
+            # It was! Force a new indent there.
```
barrywhart, line 661: Given the "if" comment is pretty complex, I'm unsure what exactly is meant by "It was" here.

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+                if ip.closing_indent_balance == closing_trough:
+                    target_point_idx = ip.idx
+                    desired_indent = single_indent * (
+                        ip.closing_indent_balance - len(ip.untaken_indents)
+                    )
+                    break
+            else:  # pragma: no cover
+                raise NotImplementedError("We should always find the relevant point.")
+            reflow_logger.debug(
+                "    Detected missing +ve line break @ line %s. Indenting to %r",
```
barrywhart, line 672: What's `+ve`?

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+            )
+            elements[target_point_idx] = new_point
+            fixes += new_fixes
+            # Keep track of the indent we forced
+            forced_indents.append(closing_balance)
+
+    # Or the way down.
+    elif closing_balance < starting_balance:
+        # On the way down we're looking for indents which *were* taken on
+        # The way up, but currently aren't on the way down. We slice so
```
barrywhart, line 688:
```suggestion
        # the way up, but currently aren't on the way down. We slice so
```

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -27,3 +36,792 @@ def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> st
...
+            # It's negative, is it untaken?
+            if (
+                ip.initial_indent_balance in ip.untaken_indents
+                and ip.initial_indent_balance not in forced_indents
+            ):
+                # Yep, untaken.
+                continue
+            # Edge Case: Semicolons. For now, semicolon placement is a little
+            # more complicated than what we do here. For now we don't (by
+            # default) introduce missing -ve indents before semicolons.
```
barrywhart, line 703: What's `-ve`?

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -74,6 +79,58 @@ def get_fixes(self) -> List[LintFix]:
         """Get the current fix buffer."""
         return self.embodied_fixes
 
+    def get_results(self) -> List[LintResult]:
+        """Generate a list of LintResult from the current fix buffer.
+
+        For some fixes, we do a bit of a shuffle with the anchor so
+        that we reference the most sensible position in the source file
+        to make sense to the user. For example if we've due to create
```
barrywhart, line 87:
```suggestion
        to make sense to the user. For example if we're due to create
```

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -74,6 +79,58 @@ def get_fixes(self) -> List[LintFix]:
...
+
+        NOTE: This generates one result per fix. Depending on your use
+        case, this may not always be appropriate. In those cases, call
+        `ReflowSequence.get_fixes()` and construct :obj:`LintResult`
+        objects directly.
+        """
+        results = []
+        segments = None
+        for fix in self.get_fixes():
+            # Anchoring on the segment AFTER makes is a lot more
```
barrywhart, line 100: The word "makes" here confuses me. Is this correct?

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -74,6 +79,58 @@ def get_fixes(self) -> List[LintFix]:
...
+                if not segments:
+                    segments = list(
+                        chain.from_iterable(elem.segments for elem in self.elements)
+                    )
+
+                target = fix.anchor.segments[-1] if fix.anchor.segments else fix.anchor
+                try:
+                    idx = segments.index(target)
+                except ValueError:
+                    # the anchor isn't there any more, it's a good anchor.
```
barrywhart, line 118: This comment confuses me. If it's not there, how is it good? 🤔

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -155,7 +212,13 @@ def _elements_from_raw_segments(
         for seg in segments:
             # NOTE: end_of_file is block-like rather than point-like.
             # This is to facilitate better evaluation of the ends of files.
-            if seg.is_type("whitespace", "newline", "indent"):
+            # NOTE: This also allows us to include literal placeholders for
+            # whitespace only strings.
+            if seg.is_type("whitespace", "newline", "indent") or (
+                seg.is_type("placeholder")
+                and cast(TemplateSegment, seg).block_type == "literal"
+                and cast(TemplateSegment, seg).source_str.isspace()
```
barrywhart, line 220: I noticed a few cases of things like this in the PR. Random thought -- worth creating helper functions for these to keep the conditions a bit shorter and more readable? Judgement call, because there aren't a LOT of these.

src/sqlfluff/core/linter/linted_file.py
```diff
@@ -395,40 +394,15 @@ def _generate_source_patches(
...
-
-                # We have a single occurrence of the thing we want to patch. This
-                # means we can use its position to place our patch.
-                new_source_slice = slice(
-                    patch.source_slice.start + positions[0],
-                    patch.source_slice.start + positions[0] + len(patch.templated_str),
+                # more carefully mapped. This used to happen with greedy template
+                # element matching, but should now never happen. In the event that
+                # it does, we'll warn but carry on.
+                linter_logger.warning(
```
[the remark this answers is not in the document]
alanmcruickshank: Yeah - I agree. I'm expecting that during testing of a potential pre-release, we might find some more of these, in the future though - yes.

src/sqlfluff/core/rules/base.py
```diff
@@ -332,6 +373,31 @@ def get_fix_slices(
             # We return an empty set because this edit doesn't touch anything
             # in the source.
             return set()
+        elif (
+            self.edit_type == "replace"
+            and all(edit.is_type("raw") for edit in cast(List[RawSegment], self.edit))
+            and all(edit._source_fixes for edit in cast(List[RawSegment], self.edit))
+        ):
+            # This is just source fixes. We can go directly to the touched raw
+            # slices and only if they overlap directly.
```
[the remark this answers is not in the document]
alanmcruickshank: Good question 🤔 - I'm not sure what I meant here. I think this might be an outdated comment. I'll revise it.

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -196,28 +308,50 @@ def indent_to(
...
+                # we checked self.num_newlines() above).
+                for idx in range(len(self.segments) - 1, -1, -1):
                     if self.segments[idx].is_type("newline"):
                         break
                 new_indent = WhitespaceSegment(desired_indent)
+                # Rather than doing a `create_after` here, we're
+                # going to do a replace. This is effectively to give a hint
+                # to the linter that this is safe to do before a templated
+                # placeholder. This solves some potential bugs - although
+                # it feels a bit like a workaround.
```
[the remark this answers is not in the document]
alanmcruickshank: Good question. I had a think about it when I wrote it and I don't think we should.

Some of the changes here allow us to fix things _much closer_ to templated segments, but still safely. I think allowing us to do a `create_after` directly before a templated section starts to become dangerous.

At the very least we should let this logic settle in a bit first.

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: I'm doing a shallow copy, because I'm editing `lines` within the loop. It felt like bad practice to modify any object which I'm directly iterating over without doing a shallow copy. I'll add a comment.

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: This whole point is badly worded. Thanks for catching it - I think I probably wrote this while everything wasn't quite working yet. 😵

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: GREAT CATCH. I found another instance of this elsewhere in the code while debugging and it took me _ages_ to debug.

src/sqlfluff/utils/reflow/sequence.py
alanmcruickshank: Good question - I'll see what I can do. I hear you that I do this same check a few times.

src/sqlfluff/utils/reflow/reindent.py
barrywhart: I saw that other loops were editing `lines` in the loop. At a glance, I didn't notice that here. Definitely making a copy is good if you're adding or deleting items. Probably even if just replacing items, although that may technically work without it. 👍

src/sqlfluff/core/rules/base.py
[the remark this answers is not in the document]
alanmcruickshank: Yeah no idea. I've updated it to something more descriptive of what's going on:
```python
# As an exception to the general rule about "replace" fixes (where
# they're only safe if they don't touch a templated section at all),
# source-only fixes are different. This clause handles that exception.

# So long as the fix is *purely* source-only we can assume that the
# rule has done the relevant due diligence on what it's editing in
# the source and just yield the source slices directly.

# More complicated fixes that are a blend or source and templated
# fixes are currently not supported but this (mostly because they've
# not arisen yet!), so further work would be required to support them
# elegantly.
```

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: Ah - that's an abbreviation of _negative_. i.e. a point where the indent balance decreases.

src/sqlfluff/utils/reflow/sequence.py
alanmcruickshank: I've revised this comment to:
```python
            # In the case of *creations*, even if we are actually
            # creating *after* something, it makes sense to anchor
            # on the segment which follows the creation point and not
            # the segment which precedes it. That's because when
            # displaying the position of a linting result in the CLI
            # we display the position of the *start* of the anchor.
            # By anchoring on the segment *following* the insertion point
            # we display the position in a way which is much more
            # interpretable by the user.
```

Is that clearer?

src/sqlfluff/utils/reflow/sequence.py
alanmcruickshank: I've updated the comment to:
```python
                    # If the anchor for the fix wasn't found, then that
                    # means that it has been replaced or removed itself.

                    # The most likely scenario for this is where the anchor
                    # for the fix is also part of the problem - therefore
                    # using it as the position for the linting result is ok.
                    # We'll set the idx to None to signal this so that in
                    # the next clause we just set the anchor of the linting
                    # result to the anchor of the fix.
```

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: abbreviation for _positive_. See other comment on whether I need to make the comments and logging here clearer.

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: Yeah - I'll rework all the comments in this area.

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: I'll rework all the comments in this area.

plugins/sqlfluff-templater-dbt/test/fixtures/dbt/dbt_project/snapshots/issue_1771.sql.after
alanmcruickshank: good plan. I've updated the code and added a test case which does exactly that.

src/sqlfluff/utils/reflow/sequence.py
barrywhart: That's good!

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -526,13 +543,248 @@ def _map_line_buffers(elements: ReflowSequenceType) -> List[_IndentLine]:
...
+    # close the untaken newline.
+    if closing_trough <= starting_balance:
+        return [], []
+
+    # On the way up we're looking for whether the ending balance
+    # was an untaken indent or not. If it *was* untaken, there's
+    # a good chance that we *should* take it.
+    if closing_trough not in indent_points[-1].untaken_indents:
+        # NOTE: I don't think we should ever get here. It's probably
+        # an error if we do.
```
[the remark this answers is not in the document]
alanmcruickshank: I've done a bit more debugging and added a better comment here. It's actually a very normal clause to end up in.

alanmcruickshank:
Apart from keeping this up to date with master - this PR is ready to go.

I'm not going to merge it yet because that starts the clock ticking for `2.0.0`. Will merge once everything else is teed up ready to go.

PR #3945, opened 2022-10-09
Consolidate comma style configs

alanmcruickshank:
Most of the references to `comma_style` were removed in #3847 . At that stage I didn't however clear up a few of the other references, in particular scattered across the config files.

`L022` was one still using the old config which raises risks for confusion. This PR migrates that rule off the old config and onto the new one. It _doesn't_ migrate the rule onto the reflow codebase itself because we don't yet have the capability to insert blank lines in that code (I'll make a note to cover that later).

While I was in there I simplified the nesting of the logic a little, but the intent is unchanged.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -124,7 +124,6 @@ For example:
    tab_space_size = 4
    max_line_length = 80
    indent_unit = space
-   comma_style = trailing
    allow_scalar = True
    single_table_references = consistent
    unquoted_identifiers_policy = all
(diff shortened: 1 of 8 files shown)


greg-finley:
I think @barrywhart is a better reviewer for now? I feel confident on CI and dialect PRs, but I don't know the rules code at all.

alanmcruickshank:
> I think @barrywhart is a better reviewer for now? I feel confident on CI and dialect PRs, but I don't know the rules code at all.

No problem at all. Thanks for taking a look 😄

barrywhart (approved):

PR #3972, opened 2022-10-18
Snowflake partition nonreserved keyword

### LINKED ISSUE #3670: partition is not a snowflake reserved keyword.
### Search before asking

### What Happened

We are using `sqlfluff` with templater `dbt` and dialect `snowflake`. 
Naming a column 'PARTITION' causes a parsing error. 
`PARTITION BY` is a reserved keyword but not the word partition itself see snowflake docs: https://docs.snowflake.com/en/sql-reference/reserved-keywords.html
It is even the suggested column name for partitioning when using external tables. Can this be updated?

### Expected Behaviour

No error is thrown by pre-commit hooks running sqlfluff with templater dbt and dialect snowflake when a column is named partition.

### Observed Behaviour

error is currently  thrown:
Screenshot 2022-07-26 at 11 25 42
n

### How to reproduce

Using the .sqlfluff below try to create a table with a column named 'partition'

### Dialect

dbt-snowflake

### Version

1.2.1

### Configuration

[sqlfluff]
dialect = snowflake
templater = dbt
exclude_rules = L031, L034, L039,  L036, L044, L013, L014

[sqlfluff:templater:dbt]
project_dir = ./booking/
profiles_dir =./booking/

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 100

[sqlfluff:rules:L010]
# Keywords
capitalisation_policy = upper

[sqlfluff:rules:L030]
# Function names
capitalisation_policy = upper

[sqlfluff:rules:L040]
# Null & Boolean Literals
capitalisation_policy = upper

[sqlfluff:rules:L057]
# Special Characters in Identifiers
additional_allowed_characters = ['.','(',')','-']

target/
# dbt <1.0.0
dbt_modules/
# dbt >=1.0.0
dbt_packages/
macros/

### Code of Conduct

YilangHe:

fixes #3670 
Move `PARTITION` from `snowflake_reserved_keywords` to `snowflake_unreserved_keywords` in file [dialect_snowflake_keywords.py](https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/dialects/dialect_snowflake_keywords.py#L1)
No

### Pull Request checklist

  - Other.


barrywhart (dismissed):
Looks good to me. I notice that `PARTITION` is a reserved keyword in ANSI and BigQuery, but not in most other dialects. (I checked several dialects but not all.)

Is there a reason for having it reserved in some dialects but not others? @tunetheweb

YilangHe:
> Looks good to me. I notice that `PARTITION` is a reserved keyword in ANSI and BigQuery, but not in most other dialects. (I checked several dialects but not all.)
> 
> Is there a reason for having it reserved in some dialects but not others? @tunetheweb

Yes, here is a list of keywords for snowflake specifically
https://docs.snowflake.com/en/sql-reference/reserved-keywords.html

tunetheweb:
Exactly. It's not even a keyword in other dialects so ideally we shouldn't even have it as an UNRESERVED word in them.

But `PartitionClauseSegment` is defined in ANSI and inherited by them. So choices are either add as unreserved keyword (so they don't fail), or change `PartitionClauseSegment` in every other dialect to be `Nothing()` so it never stumbles across this keyword.

Or do neither, and then you get a horrible "PartitionClauseSegmentKeyword not found" error if the parser ever stumbles upon this code path - usually incorrectly for other grammar that's not been implemented hence why it's getting desperate and trying other code paths it probably shouldn't be going down.

Or, if it really is a BigQuery specific thing then maybe shouldn't be in ANSI at all so not inherited by other dialects.

There's quite a few keywords in this situation and we've done a mixture of above solutions. SQL dialects are messy 😁, and our implementation tries to balance simplicity and perfection.

barrywhart:
I think we're getting a parse error on this file -- see the CI build results (Check generated YAML files / ymlchecks).

```
sqlfluff.core.errors.SQLParseError: Could not parse: test/fixtures/dialects/snowflake/snowflake_window_function_ignore_nulls.sql
```

YilangHe:
> I think we're getting a parse error on this file -- see the CI build results (Check generated YAML files / ymlchecks).
> 
> ```
> sqlfluff.core.errors.SQLParseError: Could not parse: test/fixtures/dialects/snowflake/snowflake_window_function_ignore_nulls.sql
> ```

Hi @barrywhart, 
Thank you so much for approving the PR and giving the feedback!
Actually, I didn't change the code in that file and the yml file is not generated through my test. I wondering is that a previous problem or I need to do some modifications to that.

barrywhart:
I think your changes to the Snowflake dialect may have broken the parsing of this file. You'll need to figure out if that's the case, and if there's a way to implement your change without breaking this or other existing tests.

barrywhart:
To clarify the issue with this PR...

If I check out your branch and run:
```
sqlfluff parse test/fixtures/dialects/snowflake/snowflake_window_function_ignore_nulls.sql
```

I get a parse error.

If I am on the `main` branch, the same command works. So there is an issue with this PR.

barrywhart:
The failing query uses `PARTITION BY` in a window function. My guess is that the area of the dialect that parses window function clauses needs to be modified.
```
 SELECT
  FIRST_VALUE(foo) IGNORE NULLS over (
              PARTITION BY buzz
              ORDER BY bar
              ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
            ) AS bat
            from some_table

```

YilangHe:
> barrywhart







```python
class WindowSpecificationSegment(BaseSegment):  
    """Window specification within OVER(...)."""  
  
    type = "window_specification"  
        Ref("SingleIdentifierGrammar", optional=True),  # "Base" window name  
        Ref("PartitionClauseSegment", optional=True),  
        Ref("OrderByClauseSegment", optional=True),  
        Ref("FrameClauseSegment", optional=True),  
        optional=True,  
        ephemeral_name="OverClauseContent",  
    )
```
It seems like the parser match `PARTITION` into the `SingleIdentifierGrammar` which at the end is a `NakedIdentifierSegment` and eventually fails the match. 

An ideal way is that we could backtrack and start to match from `PartitionClauseSegment`. I wonder how `optional` is defined in sqlfluff in this case? Thx

barrywhart:
@tunetheweb or @alanmcruickshank : Any insights or advice? This is more complex than I typically get with dialect PRs.

tunetheweb:
> An ideal way is that we could backtrack and start to match from PartitionClauseSegment. I wonder how optional is defined in sqlfluff in this case? Thx

Unfortunately SQLFluff does not presently work like this. I've often wished it would, but also wonder what the performance implications of that would be?

Anyway, for not, what you have to do is explicitly exclude `PARTITION` from matching here like this:

```
class WindowSpecificationSegment(BaseSegment):
    """Window specification within OVER(...)."""

    type = "window_specification"
        Ref("SingleIdentifierGrammar", optional=True, exclude=Ref.keyword("PARTITION")),  # "Base" window name
```

I think it should be safe to add this to Ansi definition rather than overwriting it in Snowflake since `PARTION` is part of `WindowSpecificationSegment` anyway.

YilangHe:
> > An ideal way is that we could backtrack and start to match from PartitionClauseSegment. I wonder how optional is defined in sqlfluff in this case? Thx
> 
> Unfortunately SQLFluff does not presently work like this. I've often wished it would, but also wonder what the performance implications of that would be?
> 
> Anyway, for now, what you have to do is explicitly exclude `PARTITION` from matching here like this:
> 
> ```
> class WindowSpecificationSegment(BaseSegment):
>     """Window specification within OVER(...)."""
> 
>     type = "window_specification"
>     match_grammar: Matchable = Sequence(
>         Ref("SingleIdentifierGrammar", optional=True, exclude=Ref.keyword("PARTITION")),  # "Base" window name
> ```
> 
> I think it should be safe to add this to Ansi definition rather than overwriting it in Snowflake since `PARTION` is part of `WindowSpecificationSegment` anyway.

Hi @tunetheweb, 
Thank you so much for your advice! I keep `PARTITION` inside `unreserved_keyword` and disallow `PARTITION` as the window name. It seems like solves the problem, I will try to submit a new PR

tunetheweb:
Looks like you need to reformat with black to get the linting to pass but at least the other tests are passing!

YilangHe:
fixed the format problem with black

greg-finley:
@YilangHe next time, could you check this box when opening a PR? Maintainers often need to merge latest `main` changes into your branch before we can merge your PR. Thanks!

Screen Shot 2022-10-27 at 10 12 47 AM

YilangHe:
> @YilangHe next time, could you check this box when opening a PR? Maintainers often need to merge latest `main` changes into your branch before we can merge your PR. Thanks!
> 
> Screen Shot 2022-10-27 at 10 12 47 AM

Got it, thx for letting me know! Is there any further modification needed for this PR?

greg-finley (approved):
Thanks for the contribution! Nice job.

PR #3991, opened 2022-10-26
Hive: Add support for EXCHANGE PARTITION in ALTER TABLE

nahuelverdugo:
Added support for parsing `Hive` queries when a partition is exchanged between tables in an `ALTER TABLE` statement. More info in the Hive [doc](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL#LanguageManualDDL-ExchangePartition).

Some minor changes in the contribution docs after following it.

I don't think so.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_hive.py
```diff
@@ -1017,3 +1017,55 @@ class SortByClauseSegment(ansi.OrderByClauseSegment):
...
+                        Ref("BracketedColumnReferenceListGrammar"),
+                        optional=True,
+                    ),
+                ),
+                # Rename
+                Sequence(
+                    "RENAME",
+                    OneOf("AS", "TO", optional=True),
+                    Ref("TableReferenceSegment"),
+                ),
```
greg-finley, line 1060: This part seems all the same as ANSI. I wonder if we could redefine that as `AlterTableOptionsGrammar` or something in ANSI and then just extend in Hive to add the "exchange" part, similar to: https://github.com/sqlfluff/sqlfluff/blob/72fed8048248426a7b6e95880cec87641249db06/src/sqlfluff/dialects/dialect_hive.py#L157-L166

greg-finley (commented):
Nice start! But I feel we can reduce code duplication a bit.

src/sqlfluff/dialects/dialect_hive.py
nahuelverdugo: Hi @greg-finley thank you for the suggestion, should I do the same replacement in other dialects where the `ansi` `AlterTableStatementSegment` class is overwritten?

greg-finley (approved):
Thanks for your contribution, and for addressing the feedback! 🙌

src/sqlfluff/dialects/dialect_hive.py
greg-finley: If you find similar ways that they are copying the same code, sure! It can be a separate PR.

PR #3999, opened 2022-10-27
Rework placeholder insertion in templates

alanmcruickshank:
Placeholder insertion and templating is really dense. It was also causing issues for the reindentation work where we need to be able to access potential unrendered whitespace in some cases. This PR reworks the lexer to rethink how we generate segments. I think it's a much more efficient was of working through the segments - and while it's a lot more lines of code - a lot of that is comments! I've tried to document this much more thoroughly.

This is quite deep in the codebase, so I'm putting it up as a draft first while I test coverage etc...

I'll test the changes against the reindentation code in parallel to make sure it does what I need there too.


alanmcruickshank:
I'm closing this PR and folding it into #3942 instead. They kind of need to happen together.

Issue #1741, opened 2021-10-25
Add rule to check number of columns in set operations


erevear:
Add rule to check that both queries in a set operation contain the same number of columns.

For example:

```
SELECT a, b, c
FROM test_table

UNION ALL

SELECT a, b
FROM test_table_2
```

The above produces a compilation error in Snowflake (`"invalid number of result columns for set operator input branches, expected 3, got 2 in branch 1"`), however it passes linting and parsing. 

Seems similar to rule L044, and could raise a parsing error.

barrywhart:
I like the idea. I suggest that it should raise a linting error, not a parsing error. For technical reasons, rules don't raise parse errors. Parse errors are generally raised only by the parser, when the SQL does not obey the grammar (i.e. the dialect).

jpy-git:
Good idea!
A clarification would be if one of the statements contains a wildcard we can't determine the number of columns, so let's not raise the lint error in that case as it would already be covered by L044.

PR #4028, opened 2022-11-04
Rule: check number of columns in set operations match

erevear:

Fixes https://github.com/sqlfluff/sqlfluff/issues/1741

Somewhat related to L044 (check that wildcards resolve to columns), however, if wildcard columns can't be resolved, doesn't fail, leaves that check to L044

### Pull Request checklist

  - Other.


src/sqlfluff/rules/L070.py
```diff
@@ -0,0 +1,249 @@
...
+
+@document_groups
+class Rule_L070(BaseRule):
+    """Queries within set query produce different numbers of columns.
+
+    **Anti-pattern**
+
+    When writing set expressions
+    , all queries must return the
+    same number of columns
```
barrywhart, line 18: Is there a reason for the lines being so short? Also, the sentence needs a `.` at the end.

src/sqlfluff/rules/L070.py
```diff
@@ -0,0 +1,249 @@
...
+    groups = ("all",)
+    crawl_behaviour = SegmentSeekerCrawler({"set_expression"}, provide_raw_stack=True)
+
+    def _find_all_ctes_utils(self, query, cte_dict):
+        """Generate a list of all ctes in a query."""
+        # cte_dict = cte_dict | query.ctes
+        cte_dict.update(query.ctes)
+        for cte_name, cte in query.ctes.items():
+            cte_dict = self._find_all_ctes_utils(cte, cte_dict)
+        return cte_dict
```
barrywhart, line 66: Nitpick: Since you're updating the dictionary in place, you don't need to return it from the function.

src/sqlfluff/rules/L070.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def _find_all_ctes(self, context: RuleContext, cte_dict):
+        for i in range(len(context.parent_stack)):
+            if context.parent_stack[i].type != "common_table_expression":
+                parent_query = SelectCrawler(
+                    context.parent_stack[i], context.dialect
```
barrywhart, line 72:
```suggestion
    def _find_all_ctes(self, context: RuleContext, cte_dict):
        for seg in context.parent_stack:
                    if  not seg.is_type("common_table_expression"):
                        parent_query = SelectCrawler(
                            seg, context.dialect
```

src/sqlfluff/rules/L070.py
```diff
@@ -0,0 +1,249 @@
...
+        all_ctes,
+    ):
+        """Attempt to resolve the wildcard to a list of selectables."""
+        process_queries = query.selectables
+        # if one of the source queries for a query within the set is a
+        # set expression, just use the first query. If that first query isn't
+        # reflective of the others, that will be caught when that segment
+        # is processed
+        if query.selectables[0].parent:
+            if query.selectables[0].parent.type == "set_expression":
```
barrywhart, line 93:
```suggestion
            if query.selectables[0].parent.is_type("set_expression"):
```

src/sqlfluff/rules/L070.py
```diff
@@ -0,0 +1,249 @@
...
+        for i in range(len(context.parent_stack)):
+            if context.parent_stack[i].type != "common_table_expression":
+                parent_query = SelectCrawler(
+                    context.parent_stack[i], context.dialect
+                ).query_tree
+                cte_dict.update(self._find_all_ctes_utils(parent_query, cte_dict))
+
+        return cte_dict
+
+    def __resolve_wildcard(
```
barrywhart, line 78: This function is really large and complex. Could you look at making it shorter by moving parts of it to separate helper functions?

src/sqlfluff/rules/L070.py
```diff
@@ -0,0 +1,249 @@
...
+                                        select_info_target,
+                                        parent_query,
+                                        resolve_targets,
+                                        all_ctes,
+                                    )
+                            else:
+                                cte = None
+                                cte_child = query.lookup_cte(wildcard_table)
+                                if wildcard_table.upper() in all_ctes:
+                                    cte = all_ctes[wildcard_table.upper()]
```
barrywhart, line 147:
```suggestion
                                cte = all_ctes.get(wildcard_table.upper())
                                cte_child = query.lookup_cte(wildcard_table)
```

test/fixtures/rules/std_rule_cases/L070.yml
```diff
@@ -0,0 +1,482 @@
+rule: L070
```
barrywhart, line 1: I think I found a bug in nested CTE handling. This query should report an issue but does not:
```
WITH y1 AS (
    WITH x AS (
        SELECT a, b FROM MyTable
    )
    SELECT * FROM x
), y2 AS (
    WITH x AS (
        SELECT a, b, c FROM MyTable
    )
    SELECT * FROM x
)
SELECT * FROM y1
UNION ALL
SELECT * FROM y2
```

This similar query where I rename the second `x` CTE as `x2` **does** report an issue:
```
WITH y1 AS (
    WITH x AS (
        SELECT a, b FROM MyTable
    )
    SELECT * FROM x
), y2 AS (
    WITH x2 AS (
        SELECT a, b, c FROM MyTable
    )
    SELECT * FROM x2
)
SELECT * FROM y1
UNION ALL
SELECT * FROM y2
```

barrywhart (changes requested):

test/fixtures/rules/std_rule_cases/L070.yml
barrywhart: BTW, I think rule L044 handles this case correctly. If you see opportunities to improve `SelectCrawler` to make it easier to handle this case, that could be worthwhile. This class has been useful but is still a work in progress.

erevear: Yes, I missed this case; I'll dig into whether there's a better way to handle finding the ctes in the SelectCrawler

erevear: @barrywhart I added a method in the SelectCrawler to build out the parent query in cases where the parent stack is added as an argument. This helps find ctes in the parent query - via the query lookup_ctes method - that are out of the scope of the segment with the set expression. Do you think this is a reasonable solution in this case?
https://github.com/sqlfluff/sqlfluff/pull/4028/files#diff-352afddb37ded0c1b8d8e0d6b5c4116308ec37b157d98bd3453eab3174e3b314R440

test/fixtures/rules/std_rule_cases/L070.yml
```diff
@@ -0,0 +1,482 @@
...
+
+test_pass_known_number_of_result_columns_1:
+  pass_str: |
+    select a, b
+    from t
+    union all
+    select c, d
+    from k
+
+test_fail_unknown_number_of_result_columns_1:
```
barrywhart, line 11: I agree this test case should fail, but if I read it correctly, the number of result columns is known (1 for the first select, 2 for the second). Should the test case be renamed?

test/fixtures/rules/std_rule_cases/L070.yml
```diff
@@ -0,0 +1,482 @@
...
+  fail_str: |
+    select a
+    from t
+    union all
+    select c, d
+    from k
+
+test_pass_known_number_of_result_columns_2:
+  desc: Columns are specified in CTE so * in final query will return only columns specified earlier.
```
barrywhart, line 20: 👍

test/fixtures/rules/std_rule_cases/L070.yml
```diff
@@ -0,0 +1,482 @@
...
+            a, b
+        from
+            t
+    )
+    select * from cte
+    union
+    select c, d
+    from t2
+
+test_fail_unknown_number_of_result_columns_2:
```
barrywhart, line 33: Same comment as above -- test name seems wrong.

test/fixtures/rules/std_rule_cases/L070.yml
```diff
@@ -0,0 +1,482 @@
...
+            t
+    )
+    select *
+    from cte
+    union
+    select d, e
+    from t
+
+# pass3;
+# fail3;
```
barrywhart, line 48: Can these lines be removed?

test/fixtures/rules/std_rule_cases/L070.yml
```diff
@@ -0,0 +1,482 @@
...
+        , cte2.*
+    from cte1
+    join cte2
+    using (a)
+
+    union
+    select e, f, g, h
+    from cte3
+
+test_fail_unknown_number_of_result_columns_3:
```
barrywhart, line 85: Test name

test/fixtures/rules/std_rule_cases/L070.yml
```diff
@@ -0,0 +1,482 @@
...
+            a, b
+        from
+            t
+    )
+    union
+    select c, d
+    from t2
+
+
+test_pass_known_number_of_result_columns_6:
```
barrywhart, line 122: This is a fun test!

test/fixtures/rules/std_rule_cases/L070.yml
```diff
@@ -0,0 +1,482 @@
...
+  pass_str: |
+    select a from t1
+    union all
+    select b from t2
+
+test_pass_unknown_wildcard_number_of_result_columns_8:
+  pass_str: |
+    select a from t1
+    union all
+    select * from t2
```
barrywhart, line 159: I wonder how this query behaves with [L044](https://docs.sqlfluff.com/en/stable/rules.html#sqlfluff.rules.Rule_L044): "Query produces an unknown number of result columns." That rule may report an issue, but arguably we know it's okay for the same reason the query passes this rule (i.e. it must return one column or the `UNION ALL` would fail at runtime).

Nothing for you to do here, just thinking...

test/fixtures/rules/std_rule_cases/L070.yml
```diff
@@ -0,0 +1,482 @@
...
+
+test_pass_known_number_of_result_columns_17:
+  pass_str: |
+    SELECT *
+    FROM table1 as table2
+    UNION
+    SELECT e, f
+    FROM table3
+
+test_fail_unknown_number_of_result_columns_7:
```
barrywhart, line 287: Test name?

test/fixtures/rules/std_rule_cases/L070.yml
```diff
@@ -0,0 +1,482 @@
...
+    from @public.mytable1
+    union
+    select mycolumn2
+    from table2
+  configs:
+    core:
+      dialect: snowflake
+
+
+test_pass_cte_no_select_final_statement:
```
barrywhart, line 462: Should we add a similar failure case (i.e. where the SELECT is inside an UPDATE)?

src/sqlfluff/utils/analysis/select_crawler.py
```diff
@@ -232,6 +232,7 @@ def __init__(
         dialect: Dialect,
         parent: Optional[Query] = None,
         query_class: Type = Query,
+        parent_stack: Optional[Tuple[BaseSegment, ...]] = None,
```
barrywhart, line 235: This looks good to me. Could you look at adding test(s) for it in `test/utils/analysis/test_select_crawler.py`?

barrywhart (changes requested):
Looks good overall. A few suggestions about tests, particularly for the new SelectCrawler behavior.

src/sqlfluff/utils/analysis/select_crawler.py
erevear: @barrywhart Yes! Added a test case for crawling the parent queries

barrywhart (approved):
Looks great!

Thanks for adding those tests.

PR #4032, opened 2022-11-07
Filter out issues in unparsable sections

alanmcruickshank:
This partially addresses #3989. It fully addresses it in combination with #3942.

Rule crawlers (#3717) introduced more advanced crawling mechanisms and would filter out unparsable sections of code _before linting them_, however for rules which handle the own crawling this isn't sufficient - because they might still recurse down to sections of unparsable code. This PR filters violations after parsing to prune any which anchor on (or modify) unparsable code if that's set on the crawler.

### DIFF
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -595,6 +595,9 @@ class BaseRule:
             # Check whether this should be filtered out for being unparsable.
             # To do that we check the parents of the anchors (of the violation
             # and fixes) against the filter in the crawler.
+            # NOTE: We use `.passes_filter` here to do the test for unparsable
+            # to avoid duplicating code because that test is already implemented
+            # there.
             anchors = [lerr.segment] + [fix.anchor for fix in lerr.fixes]
             for anchor in anchors:
                 if not self.crawl_behaviour.passes_filter(anchor):  # pragma: no cover
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -549,7 +549,7 @@ class BaseRule:
                 memory = res.memory
                 self._adjust_anchors_for_fixes(context, res)
                 self._process_lint_result(
-                    res, templated_file, ignore_mask, new_lerrs, new_fixes
+                    res, templated_file, ignore_mask, new_lerrs, new_fixes, tree
                 )
             elif isinstance(res, list) and all(
                 isinstance(elem, LintResult) for elem in res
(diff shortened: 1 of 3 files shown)


src/sqlfluff/core/rules/base.py
```diff
@@ -586,13 +586,37 @@ def _log_critical_errors(error: Exception):  # pragma: no cover
...
     def _process_lint_result(
-        self, res, templated_file, ignore_mask, new_lerrs, new_fixes
+        self, res, templated_file, ignore_mask, new_lerrs, new_fixes, root
     ):
         self.discard_unsafe_fixes(res, templated_file)
         lerr = res.to_linting_error(rule=self)
         ignored = False
         if lerr:
-            if ignore_mask:
+            # Check whether this should be filtered out for being unparsable.
```
barrywhart, line 595: It's not clear to me how this code is checking for "unparsable". I see "unparsable" in the comment, but not in the code. How does this situation arise, and how does this code detect and address it? A bit more detailed comments seems like it would be helpful here.

What test(s) are hitting this code? Should we update or add tests for this change?

Do we need to consider `--FIX-EVEN-UNPARSABLE` here? I.e. if that is set, don't filter out the fix?

alanmcruickshank: The current configuration for whether rules should work on unparsable sections is in the crawler (see `BaseCrawler`). All of them have the optional attribute of `works_on_unparsable`. That feeds through to `.passes_filter()` (check out the implementation of `BaseCrawler.passes_filter()`). The `passes_filter` method is the one that's called in this section of code - to avoid needing to redefine whether a rule works on unparsable code or not.

The reason I've done it this way is because unless we leverage the existing unparsable logic then we end up redefining it in another place. This way feels more DRY.

alanmcruickshank: I hear you on tests though... I'll see what I can do.

barrywhart: Definitely agree re: not repeating code. Do you think it'd be worth adding something similar to your PR comment to the code itself, to help future maintainers connect the dots?

src/sqlfluff/core/rules/base.py
```diff
@@ -586,13 +586,40 @@ def _log_critical_errors(error: Exception):  # pragma: no cover
...
+            # and fixes) against the filter in the crawler.
+            # NOTE: We use `.passes_filter` here to do the test for unparsable
+            # to avoid duplicating code because that test is already implemented
+            # there.
+            anchors = [lerr.segment] + [fix.anchor for fix in lerr.fixes]
+            for anchor in anchors:
+                if not self.crawl_behaviour.passes_filter(anchor):  # pragma: no cover
+                    # NOTE: This clause is untested, because it's a hard to produce
+                    # edge case. The latter clause is much more likely.
+                    linter_logger.info("Fix skipped due to anchor not passing filter.")
```
barrywhart, line 606: For possible future debugging purposes, do you think it may be useful to log the segment here and in the similar logic below?

barrywhart (approved):

src/sqlfluff/core/rules/base.py
alanmcruickshank: Yeah good idea.

PR #4038, opened 2022-11-07
Update CI to use Python 3.11

greg-finley:
Was just curious if it works yet. If so, why not?

### DIFF
--- a/.github/workflows/check-yaml-files-generation.yml
+++ b/.github/workflows/check-yaml-files-generation.yml
@@ -5,10 +5,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - name: Set up Python 3.10
+    - name: Set up Python 3.11
       uses: actions/setup-python@v3
       with:
-        python-version: '3.10'
+        python-version: '3.11'
     - name: Install dependencies
       run: |
         pip install -Ur requirements.txt -Ur requirements_dev.txt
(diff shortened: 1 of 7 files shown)


src/sqlfluff/utils/analysis/select_crawler.py
```diff
@@ -145,13 +145,9 @@ def as_json(self) -> Dict:
         if self.query_type != QueryType.Simple:
             result["query_type"] = self.query_type.name
         if self.selectables:
-            result["selectables"] = [
-                s.as_str() for s in self.selectables
-            ]  # type: ignore
+            result["selectables"] = [s.as_str() for s in self.selectables]
```
greg-finley, line 148: 3.11 thinks that this ignore is not needed, but someone with an old mypy would get the opposite error when they try to commit.

alanmcruickshank: what's the error? There might be another way around the type hinting here to make both of them happy.

alanmcruickshank:
Great idea @greg-finley - the potential speed improvements in 3.11 would be a good addition to the test suite.

src/sqlfluff/utils/analysis/select_crawler.py
alanmcruickshank: actually - I'm running an older version so I can check myself! Have just pushed a fix which should work for both.

alanmcruickshank (approved):
LGTM

PR #4045, opened 2022-11-08
Use Python 3.11 for Windows tests

greg-finley:
Let's see if it works ...


greg-finley:
I don't immediately know how to solve the issues. I guess just give up for now?

alanmcruickshank:
dbt core doesn't support py11 yet https://pypi.org/project/dbt-core/.

It's only recently that it's fully supported py10 😄 . I think probably leave it for now.

PR #4046, opened 2022-11-08
Bump after_n_builds to 10

greg-finley:
Since we added a Python 3.11 required CI test.

https://sqlfluff.slack.com/archives/G01G58WRJNR/p1667926639893419?thread_ts=1667922270.894929&cid=G01G58WRJNR


tunetheweb (approved):
LGTM is tests pass - and more importantly don't hang :-)

Issue #4048, opened 2022-11-08
Snowflake: COMMENT clause in ALTER TABLE statement throwing parsing violation


srjonemed:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

`sqlfluff parse` throws PRS error for sql containing ALTER TABLE statement including a COMMENT section 

### Expected Behaviour

above sql runs fine on snowflake, so expecting sqlfluff should pass the parsing

### Observed Behaviour

`sqlfluff parse` throws PRS error for following sql:
`ALTER TABLE rpt_enc_table
ADD
encounter_count INTEGER COMMENT 'count of encounters past year'
;`

`==== parsing violations ====
L:   3 | P:  24 |  PRS | Line 3, Position 24: Found unparsable section: " COMMENT 'count of
                       | encounters past year'..."`

### How to reproduce

run sqlfluff parse with above sql example

### Dialect

snowflake

### Version

0.10.1

### Configuration

[sqlfluff]
verbose = 0
nocolor = False
dialect = snowflake
templater = jinja
rules = L001,L002,L003,L004,L005,L006,L007,L008,L009,L010,L011,L012,L013,L016,L017,L018,L019,L022,L023,L024,L025,L029,L030,L036,L037,L038,L039,L041,L044,L045,L046,L047,L048
exclude_rules = L029,L031
recurse = 0
output_line_length = 80
runaway_limit = 10
ignore_templated_areas = True
ignore = parsing, templating
# Comma separated list of file extensions to lint.

# NB: This config will only apply in the root folder.
sql_file_exts = .sql,.dml,.ddl

[sqlfluff:indentation]
indented_joins = False
template_blocks_indent = True

[sqlfluff:templater]
unwrap_wrapped_queries = True

[sqlfluff:templater:jinja]
apply_dbt_builtins = True

[sqlfluff:templater:jinja:macros]
# Macros provided as builtins for dbt projects
dbt_ref = dbt_source = dbt_config = dbt_var = dbt_is_incremental =
# Some rules can be configured directly from the config common to other rules.
[sqlfluff:rules]
tab_space_size = 4
max_line_length = 140
indent_unit = space
comma_style = leading
allow_scalar = True
single_table_references = consistent
unquoted_identifiers_policy = all

[sqlfluff:templater:jinja:context]
env_name=integration

# Some rules have their own specific config.
#[sqlfluff:rules:L003]
[...]

PR #4049, opened 2022-11-08
Snowflake: ADD and DROP without COLUMN

greg-finley:

Closes https://github.com/sqlfluff/sqlfluff/issues/4048

[Docs](https://docs.snowflake.com/en/sql-reference/sql/alter-table.html) say `COLUMN` keyword isn't required for `ADD` or `DROP`:

Screen Shot 2022-11-08 at 11 06 59 AM

No

### Pull Request checklist

  - Other.

### DIFF
--- a/test/fixtures/dialects/snowflake/snowflake_alter_table_column.yml
+++ b/test/fixtures/dialects/snowflake/snowflake_alter_table_column.yml
@@ -3,7 +3,7 @@
 # computed by SQLFluff when running the tests. Please run
 # `python test/generate_parse_fixture_yml.py`  to generate them after adding or
 # altering SQL files.
-_hash: ca8548d7f42a1ff1964b273ff4ae16434c6b5b94eee29bc8779240e1aecfa31f
+_hash: 22878f1f16fcd23b97eeabe2abce943e02764130cdad64ba43b7353535333d37
 file:
 - statement:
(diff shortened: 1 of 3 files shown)


test/fixtures/dialects/snowflake/snowflake_alter_table_column.yml
```diff
@@ -613,14 +642,14 @@ file:
       - naked_identifier: my_table
     - alter_table_table_column_action:
         keyword: add
-        column_definition:
+        column_reference:
```
greg-finley, line 645: does this matter?

srjonemed:
ah good catch, thanks for the fix!

test/fixtures/dialects/snowflake/snowflake_alter_table_column.yml
tunetheweb: Yes it does. It now thinks that `PRIMARY` is a column name, rather than a column definition, meaning it will be subject to the identifier rules, rather than the keyword rules - for example rules L010 and L014.

Though it looks to me like it didn't parse this part properly anyway since it thinks `PRIMARY` is an identifier and `KEY` is a data type. Still I think we've made it worse with this change.

tunetheweb (approved):
Nice! LGTM.

PR #4053, opened 2022-11-09
Allow warnings for specific rules

### LINKED ISSUE #201: Add severity level for lint violations
Most linters tag each violation with a severity level. Text editor integrations have settings to surface violations with severity above one or another level. Fluff should have these levels, even if it as much as a warning for a violation and an error for a parse failure! 

This could be surfaced to the user in the lint output (as text and/or color?) and in the json/yaml output.

alanmcruickshank:
This resolves #201 (THE OLDEST OUTSTANDING ISSUE 🚀 ) - only took over 2 yrs.

This delivers almost exactly against the spec outlined in that original issue.

Rules can be configured as warnings, in which case they're displayed as such, but don't fail the file.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -229,6 +229,37 @@ The default values can be seen in `Default Configuration`_.
 
 See also: `Ignoring Errors & Files`_.
 
+Downgrading rules to warnings
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To keep displaying violations for specific rules, but not have those
+issues lead to a failed run, rules can be downgraded to *warnings*.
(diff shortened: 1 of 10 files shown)


tunetheweb:
Change itself looks good. However it feels a bit hidden away with just a comment in the default config file. Could we add some more documentation to https://docs.sqlfluff.com/en/stable/configuration.html?

alanmcruickshank:
@tunetheweb - what do you reckon the documentation I've put in? It's fairly brief but I think to the point.

tunetheweb:
Looks great to me!

tunetheweb (approved):

nolanbconaway:
Heck yeah!!!

PR #4054, opened 2022-11-09
Implicit Indents

alanmcruickshank:
## This is an experiment.

The current reindent branch #3942 has one bit I'm not really happy with and that's queries like:

```sql
SELECT *
FROM foo
WHERE a
    AND b
-- ... which gets corrected to...
SELECT *
FROM foo
WHERE
    a
    AND b
-- ... which is _fine_, but not quite what I think many users would expect.
```

Right now, because there's a line break after `a` and no linebreak before it - this gets picked up as a missing indent between `WHERE` and `a` and one would be forced in. If this were a normal indent, that would be correct - like in this bracketed example:

```sql
SELECT (a
    + b);

-- ...corrects to ...
SELECT (
    a
    + b
);
-- ...which I think is correct.
```

So I think where I'm going is that the indent in the `WHERE` clause example is a bit different - it behaves like a taken indent even when it's strictly not. This PR introduces the idea of an _implicit_ indent which behaves a little different and allows situations like this. If it kind of works I plan to allow their behaviour to be configurable, ether:
- `allow` - where both a line break is allowed, but not required.
- `force` - where there _must_ be a line break unless the whole thing fits on one line (i.e. treat it like a normal indent).
- `never` - where there's never a line break there.

I'm expecting this might break several tests to start with - hence draft PR.


NiallRees:
I'm personally onboard with the behaviour in the first example being the default:
```
SELECT *
FROM foo
WHERE a
    AND b
-- ... which gets corrected to...
SELECT *
FROM foo
WHERE
    a
    AND b
-- ... which is _fine_, but not quite what I think many users would expect.
```

alanmcruickshank:
@NiallRees - have you got any other ideas of SQL constructs which should also have this treatment? I've only updated tests for `WHERE` so far, but I'm sure there are others.

NiallRees:
My personal choice would be that where clauses and case whens should default to having newlines. Others we should use the max line length to create newlines.

alanmcruickshank:
> My personal choice would be that where clauses and case whens should default to having newlines.

@NiallRees - do you mean default to **not** having newlines? This is all about intentionally not forcing a newline immediately after `WHERE` or `WHEN` (thanks for the suggestion of `CASE` statements, that's a good one).

NiallRees:
> > My personal choice would be that where clauses and case whens should default to having newlines.
> 
> @NiallRees - do you mean default to **not** having newlines? This is all about intentionally not forcing a newline immediately after `WHERE` or `WHEN` (thanks for the suggestion of `CASE` statements, that's a good one).

Some examples of my preference:
```
select *
from a
where a.a = true
```
^ should not be changed


```
select *
from a
where a.a = true and b.b = false
```
^ should be changed to:
```
select *
from a
where 
    a.a = true
    and b.b = false
```

```
select case when a = b then 'a' end as col
from table
```
^ should not be changed, contains only one branch (a single `when`)


```
select case when a = b then 'a' else 'b' end as col
from table
```
^ should be changed to the following on the basis that it has more than one branch (a `when` and an `else`):
```
select
    case
        when a = b then 'a'
        else 'b'
    end as col
from table
```

alanmcruickshank:
@NiallRees to be clear - this PR is about which of these is correct:

```
select *
from a
where 
    a.a = true
    and b.b = false
```

or

```
select *
from a
where a.a = true
    and b.b = false
```

This isn't about forcing a newline around the `and`.

barrywhart:
Aww, darn -- I like the second way. (In fact I asked about this on the indentation PR.)

Not so much about "correctness", I just like fewer lines of code in general.

NiallRees:
To try and put some reasoning behind it, I prefer the first because I think identical grammars should be consistently indented. Placing the first on the where clause breaks that

alanmcruickshank:
Ok - this is useful feedback. It says that this feature is useful, but that it definitely needs to be configurable.

We can argue later about what the default should be.

alanmcruickshank:
@barrywhart - given the convo so far - I also think let's keep this as a separate feature and get the reflow PR merged before bolting this on as an addition.

barrywhart:
Sure, no problem.

I mentioned on the other PR, I'd be curious how much impact this has on a typical, medium complex SQL file. I'm less concerned about any particular decision like this and more curious what the overall impact is in terms of SQL file length and how that affects someone working on a query that's too long to see in its entirety without scrolling.

alanmcruickshank:
It's taken a bit of time to get this into a state that it works nicely with the other reflow PRs, but this is getting close now.

Notably, I've checked some of the other style guides:
- https://docs.telemetry.mozilla.org/concepts/sql_style.html
- https://about.gitlab.com/handbook/business-technology/data-team/platform/sql-style-guide/
- https://github.com/dbt-labs/corp/blob/main/dbt_style_guide.md

...and none recommend this style.

I quite like it, but I've set it to be disabled by default for now - so users will have to explicitly enable it if they want to use it.

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -1167,10 +1191,29 @@ def _match_indents(
         if matched_indents[indent_level] == [newline_idx]:
             matched_indents.pop(indent_level)
             reflow_logger.debug(
-                "    purging balance of %s, it references only the " "final element",
+                "    purging balance of %s, it references only the final element.",
                 indent_level,
             )
 
+    # ADDITIONALLY - if implicit indents are allowed we want we should
```
barrywhart, line 1198:
```suggestion
    # ADDITIONALLY - if implicit indents are allowed we should
```

test/fixtures/rules/std_rule_cases/L016.yml
```diff
@@ -455,3 +455,79 @@ test_comment_move_mid_query:
...
+  configs:
+    core:
+      max_line_length: 45
+    indentation:
+      allow_implicit_indents: True
+
+test_fix_no_implicit_indent:
+  # Test explicitly preventing implicit indents.
+  fail_str: |
+    SELECT
```
barrywhart, line 498: I notice the `fail_str` for this test is different from the one for `test_fix_implicit_indent`, e.g. the indentation of the first `THEN`. Is that intentional? I had assumed they'd be the same, in order to illustrate the different outcomes.

barrywhart: Assuming the two are logically the same even despite the indentation differences, I notice the output is 15 lines vs 20 lines. In that sense, I like the test because a 33% increase in lines is a high price to pay! It definitely reminds me of some annoying things Black does occasionally. FWIW, I code on a notebook, often a 13" notebook. 🙂

alanmcruickshank: They are supposed to be the same. Good catch.

Now you can see why I like this feature. :)

barrywhart: Me too!

barrywhart (approved):

PR #4060, opened 2022-11-11
Fix type error in `get_rules` hook of plugin example

Samyak2:

The `get_rules` function in the plugin example template had a type error. The function returns all the rule classes that are defined in the plugin. The return type was incorrectly annotated as `List[BaseRule]` which means a list of *instances* of `BaseRule` and not classes derived from `BaseRule`. Making it `List[Type[BaseRule]]` instead fixes the error.

For reference, the type error shown by pyright was:
```
Expression of type "list[Type[Rule_Example_L001]]" cannot be assigned to return type "List[BaseRule]"
...
  "Type[type]" is incompatible with "Type[BaseRule]"
```

None (AFAICT)

### Pull Request checklist

  - Other.

### DIFF
--- a/plugins/sqlfluff-plugin-example/src/example/rules.py
+++ b/plugins/sqlfluff-plugin-example/src/example/rules.py
@@ -12,13 +12,13 @@ from sqlfluff.core.rules.doc_decorators import (
     document_fix_compatible,
     document_groups,
 )
-from typing import List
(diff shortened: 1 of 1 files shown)


barrywhart (approved):

barrywhart:
Thanks!

Issue #4031, opened 2022-11-07
Reorganize rule names


pwildenhain:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

Let's think about a nice way to rename rules/codes, so that we don't have to look up what the rule code is everytime we think about them 😵 

Here's an example of what `pylint` messages look like:

```sh
$ pylint etl.py 
************* Module etl
etl.py:20:61: C0303: Trailing whitespace (trailing-whitespace)
etl.py:1:0: C0114: Missing module docstring (missing-module-docstring)
etl.py:7:0: W0622: Redefining built-in 'compile' (redefined-builtin)
etl.py:7:0: E0401: Unable to import 'parse' (import-error)

------------------------------------------------------------------
Your code has been rated at 5.56/10 (previous run: 9.95/10, -4.40)
```

You can see that the have "common sense" names as the end of each linting message.

Here's an example of what the current output looks like for `sqlfluff lint`:

```sh
$ sqlfluff lint models/output/
=== [dbt templater] Sorting Nodes...                                                       
=== [dbt templater] Compiling dbt project...                                               
=== [dbt templater] Project Compiled.                                                      
== [models/output/redcap_import.sql] FAIL                                                  
L:   6 | P:   5 | L019 | Found leading comma. Expected only trailing.                      
L:   6 | P:   6 | L008 | Commas should be followed by a single whitespace unless
                       | followed by a comment.
L:   9 | P:  40 | L012 | Implicit/explicit aliasing of columns.
L:  17 | P:   6 | L003 | Expected 1 indentations, found 1 [compared to line 16]
```

For the record I think on the whole, our output is better organized/nicer to look at than pylint. But I think it would be greatly improved if we could implement common sense names like this:

```sh
$ sqlfluff lint models/output/
=== [dbt templater] Sorting Nodes...                                                       
=== [dbt templater] Compiling dbt project...                                               
=== [dbt templater] Project Compiled.                                                      
== [models/output/redcap_import.sql] FAIL                                                  
L:   6 | P:   5 | L019 | Found leading comma. Expected only trailing. (wrong-comma-style)                  
L:   6 | P:   6 | L008 | Commas should be followed by a single whitespace unless
                       | followed by a comment. (comma-missing-whitespace)
L:   9 | P:  40 | L012 | Implicit/explicit aliasing of columns. (wrong-column-alias-style)
[...]

PR #4067, opened 2022-11-15
Reflow 4: Long Lines

### LINKED ISSUE #365: "sqlfluff fix": Questionable reformatting of long line
I'm testing `sqlfluff fix --rules L001,L003,L006,L008,L010,L011,L016,L018,L019,L022,L023,L030` on some of our projects.

This was a case where the reformatting seemed a bit ugly, and maybe there's a better way.

Given this original line as input (part of a `WHERE` clause:
```
recommendation_list[ORDINAL(1)] = 'uses_small_subject_line' AND uses_small_subject_line != CAST(effect_size_list[ORDINAL(1)] AS FLOAT64)
```

It got reformatted as:
```
    recommendation_list[
            ORDINAL(1)
    ] = 'uses_small_subject_line' AND uses_small_subject_line != CAST(
            effect_size_list[ORDINAL(1)] AS FLOAT64
    )
```

I think splitting the lines around the `AND` would be much more readable, something like:
```
recommendation_list[ORDINAL(1)] = 'uses_small_subject_line' AND
    uses_small_subject_line != CAST(effect_size_list[ORDINAL(1)] AS FLOAT64)
```

I didn't have a max line length set, so I assume it was using the default.

If you like, let me know and I can provide a full example query. I'm just doing initial triage right now...

### LINKED ISSUE #946: Resolve whitespace interdependencies between rules.

### LINKED ISSUE #2466: Prefer to break lines at mathematical operators over function calls
Currently, `sqlfluff` breaks long lines up like this, for example:
```sql
select
  blah = ISNULL(
        count, '0'
      ) * 10000 + ISNULL(planned, 100)
from blah
```

However, I think it would be easier to read if it's broken up at the operators, not at the parameter list:
```sql
select
  blah = ISNULL(count, '0') * 10000
    + ISNULL(planned, 100)
from blah
```
(When I manually change it to this formatting, `sqlfluff` accepts it.)
I'd go so far as to say that even between mathematical operators, there should be a preference to break up at `+` over `*`.

What do you think?

alanmcruickshank:
This brings L016 into reflow. It depends heavily on #3942 so that should be merged before this comes in. This is part of #3673. This also resolves #946.

In essence this makes long line checks part of `.reindent()` and therefore currently part of `L003`. That shouldn't matter with #4031 but it will feel a little strange for now.

Test cases added for:
- fixes #2466
- fixes #365 

This PR is about 50% code for long lines and 50% refinements to position markers. Some inaccuracy in how we generate the `source_slice` for fixes around templated elements and for meta segments was causing a real headache here (which is partly why it's taken so long). I'm sure there are still a couple of oddities in that code, but this definitely moves us a good few steps forward in that regard.

One thing that this _doesn't_ currently do is add brackets to expressions which get broken across multiple lines. I'm not 100% sure on whether that should always be the case, but it does occasionally lead to some slightly unexpected outcomes, for example if set with a very short line length it would flatten as follows:

```sql
-- from:
select
    some + long = variable + expression
from foo;

-- with a medium setting on line length it would get corrected to:
select
   some + long
   = variable + expression
from foo;
-- (assuming _leading_ operators, and I think this correction is good progress because it respects
-- operator precedence).

-- but with a very short limit it corrects to:
select
    some
    + long
    = variable
    + expression
from foo;
-- which is as correct as we can get without inserting brackets.

-- but a better fix would be:
select
   (
       some
       + long
   ) = (
      variable
      + expression
   )
from foo;
```

I don't think this is an awful situation, and it's worth noting that if the bracketed example was given, sqlfluff will think it's correct _but it won't proactively add brackets as part of a long line fix where none currently exist_.


test/fixtures/linter/autofix/ansi/001_long_line/violations.json
```diff
@@ -2,7 +2,7 @@
     "violations":{
         "linting":{
             "L016": [
-                [3, 199]
+                [3, 9]
```
barrywhart, line 5: Can you explain this change?

test/fixtures/linter/autofix/ansi/023_l016_confuses_l041/after.sql
```diff
@@ -1,10 +1,10 @@
...
-        )
+    *
+FROM
+    superverylongtablenamereallyreally1
+WHERE
+    long_varname_to_trigger_Rule_L016_id in (
+        SELECT distinct id
+        FROM superverylongtablenamereallyreally2
+        WHERE deletedat IS NULL
+    )
```
barrywhart, line 10: 👍

test/fixtures/linter/autofix/ansi/026_l016_line_length_includes_earlier_fixes/after.sql
```diff
@@ -2,10 +2,10 @@ SELECT
...
     abs(
         (
-            round(
-                foo_bar_report.metricx-xxx_yyy_report.metricx
-            )/foo_bar_report.metricx
-        )*100
+            round(foo_bar_report.metricx-xxx_yyy_report.metricx)
+            /foo_bar_report.metricx
+        )
+        *100
```
barrywhart, line 8: Why was `*100` moved to a separate line, even though it fits?

Not necessarily a problem, just curious.

test/fixtures/rules/std_rule_cases/L016.yml
```diff
@@ -4,7 +4,7 @@ test_pass_line_too_long_config_override:
   # Long lines (with config override)
   pass_str: "SELECT COUNT(*) FROM tbl\n"
   configs:
-    rules:
+    indentation:
```
barrywhart, line 7: `indentation` feels like a strange place for `max_line_length`. Should we consider a more inclusive term such as `formatting`?

test/fixtures/rules/std_rule_cases/L016.yml
```diff
@@ -325,15 +330,129 @@ test_fail_templated_comment_clause:
       jinja:
         context: {}
 
-test_pass_ignore_templated_comment_clause:
+test_pass_ignore_templated_comment_lines:
+  # NOTE: This is potentially a behaviour change in 2.0.0.
+  # This was erroneously using the `ignore_comment_clauses`
+  # config when this query contains no comment clauses.
```
barrywhart, line 336: >contains no comment clauses

Is that because it's a Jinja comment rather than a SQL comment?

test/fixtures/rules/std_rule_cases/L016.yml
```diff
@@ -325,15 +330,129 @@ test_fail_templated_comment_clause:
...
+    indentation:
+      max_line_length: 0
+
+test_pass_check_off_2:
+  # Tests that we can disable the check
+  pass_str: |
+    select my_really_really_really_really_really_really_really_really_really_really_really_long_var from tbl
+  configs:
+    indentation:
+      max_line_length: -1
```
barrywhart, line 443: What's the difference between 0 and -1?

test/fixtures/rules/std_rule_cases/L016.yml
```diff
@@ -325,15 +330,129 @@ test_fail_templated_comment_clause:
...
+
+test_comment_move_mid_query:
+  # Tests that we can disable the check
```
barrywhart, line 446: >disable the check

What do you mean here? Is code plus comment allowed to exceed the max line limit?

test/rules/std_test.py
```diff
@@ -76,9 +76,11 @@
         (
             "L016",
             "block_comment_errors.sql",
-            [(1, 121), (2, 99), (4, 88)],
+            # Two errors on line one, because we call out the missing
+            # indent *and* the long line as two issues.
+            [(1, 7), (1, 8), (2, 5), (4, 5)],
         ),
-        ("L016", "block_comment_errors_2.sql", [(1, 85), (2, 86)]),
+        ("L016", "block_comment_errors_2.sql", [(1, 1), (2, 1)]),
```
barrywhart, line 83: What changed here?

src/sqlfluff/core/default_config.cfg
```diff
@@ -114,6 +118,33 @@ spacing_within = inline
 spacing_before = any
 spacing_after = any
 
+# Setting a selection of clauses to "alone" ensures line breaks
+# are preferred around them. Setting this to "alone:strict" would always
```
barrywhart, line 122: Does "preferred" mean the linter is allowed to add line breaks there? Does it mean anything else/additional?

src/sqlfluff/core/parser/markers.py
```diff
@@ -91,6 +91,31 @@ def from_point(
...
         )
 
+    @classmethod
+    def from_points(
+        cls,
+        start_point_marker: "PositionMarker",
+        end_point_marker: "PositionMarker",
+    ):
+        """Construct a position marker from the section between two points."""
+        return cls(
```
barrywhart, line 101: Should we add sanity checks that:
* The two slices don't overlap
* The end slice starts after the start slice ends

Looking over this file, other factory functions could use similar checks also.

src/sqlfluff/core/parser/markers.py
```diff
@@ -91,6 +91,31 @@ def from_point(
...
+            slice(
+                start_point_marker.source_slice.start,
+                end_point_marker.source_slice.stop,
+            ),
+            slice(
+                start_point_marker.templated_slice.start,
+                end_point_marker.templated_slice.stop,
+            ),
+            # The position markers should be the same so just pick
+            # the first.
```
barrywhart, line 111: Which ones should be the same? What does "the first" mean?

src/sqlfluff/rules/L003.py
```diff
@@ -73,8 +73,14 @@ def _eval(self, context: RuleContext) -> List[LintResult]:
...
+        # NOTE: Until rule consolidation, we're going to filter out
+        # any long line fixes. After that we'll consolidate the two
+        # into one rule.
+        results = (
             ReflowSequence.from_root(context.segment, context.config)
             .reindent()
             .get_results()
         )
+        return [
+            res for res in results if not res.description.startswith("Line is too long")
```
barrywhart, line 85: Could there be value in formalizing the source / reason for a rule? I.e. something that is guaranteed to be unique and not change such as an `Enum`. In the past, we've occasionally changed descriptions, which would have broken this code in a non-obvious way.

Related: I see code in L016 which does:
```
res.source == "reflow.long_line"
```

Is that doing something similar? Could we use that here?

test/fixtures/linter/autofix/ansi/001_long_line/violations.json
alanmcruickshank: L016 used to anchor on the final newline at the end of the line. It now anchors on the first non-indent character of the line. That felt a little more consistent.

test/fixtures/rules/std_rule_cases/L016.yml
alanmcruickshank: good point 🤔 . Rather than making a new section for it - I wonder whether I should just bring it up to the parent `sqlfluff` section. 👍

test/fixtures/rules/std_rule_cases/L016.yml
alanmcruickshank: No difference. Both work. It was easy enough to support both conventions - as I realise different tools sometimes use 0 or -1 for a _"don't use this"_ value.

test/rules/std_test.py
alanmcruickshank: Same as previous comment. Long line checks now anchor on the first non-indent char rather than the closing newline.

src/sqlfluff/core/default_config.cfg
alanmcruickshank: I'll clarify this. What it actually means is that if there's a newline on _one_ side then it will introduce one on the other too - but if there isn't a newline _on either side_ then it won't force new line breaks on either without the `strict` option.

src/sqlfluff/rules/L003.py
alanmcruickshank: Ah yes - we should do exactly do that. Must have missed this bit. Good catch!

src/sqlfluff/rules/L016.py
```diff
@@ -28,620 +23,144 @@ class Rule_L016(BaseRule):
...
+                # is a comment_clause.
+                raw_idx = raw_segments.index(res.anchor)
+                for _, seg in enumerate(raw_segments[raw_idx:], raw_idx):
+                    path = context.segment.path_to(seg)
+                    if (
+                        seg.pos_marker.working_line_no
+                        != res.anchor.pos_marker.working_line_no
+                    ):
+                        # We've gone past the end of the line. Stop looking.
+                        break
```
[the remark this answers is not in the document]
alanmcruickshank: Hmmmm - good idea. I'll investigate that option.

src/sqlfluff/rules/L042.py
```diff
@@ -562,7 +561,7 @@ class SegmentCloneMap:
     """Clones a segment tree, maps from original segments to their clones."""
 
     def __init__(self, segment: BaseSegment):
-        segment_copy = copy.deepcopy(segment)
+        segment_copy = segment.copy()
```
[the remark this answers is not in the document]
alanmcruickshank: Ah yes -  this was causing bugs because the deep copy was copy too much. Copying the `segment` copies the `PositionMarker` which copies the `TemplatedFile`. That's too deep. The `.copy()` method on the segment manages that depth more elegantly by copying the segments but not their positions (which are immutable - so would be recreated if edited anyway).

src/sqlfluff/utils/reflow/helpers.py
```diff
@@ -25,3 +31,21 @@ def pretty_segment_name(segment: BaseSegment) -> str:
         # Reference other segments just by their type.
         # (With underscores as spaces)
         return segment.get_type().replace("_", " ")
+
+
+def deduce_line_indent(raw_segment: RawSegment, root_segment: BaseSegment) -> str:
+    """Given a raw segment, deduce the indent of it's line."""
+    seg_idx = root_segment.raw_segments.index(raw_segment)
```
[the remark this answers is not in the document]
alanmcruickshank: This is actually the root segment rather than the parent. The root segment gets passed around for doing exactly this kind of operation.

src/sqlfluff/utils/reflow/helpers.py
```diff
@@ -25,3 +31,21 @@ def pretty_segment_name(segment: BaseSegment) -> str:
...
+    indent_seg = None
+    for seg in root_segment.raw_segments[seg_idx::-1]:
```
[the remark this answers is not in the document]
alanmcruickshank: Oh interesting! Great idea.

src/sqlfluff/core/parser/markers.py
alanmcruickshank: Will clarify....

src/sqlfluff/utils/reflow/helpers.py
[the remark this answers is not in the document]
alanmcruickshank: Ok - so because we're not starting at the end of the whole series, `reversed()` isn't a great solution - but we can do something with `range()` and a lookup to approximate the same effect.

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -585,8 +579,16 @@ def _deduce_line_current_indent(
...
+            reflow_logger.warning(
+                "Segment position marker: %s: [SRC: %s, TMP:%s]",
+                indent_seg.pos_marker,
+                indent_seg.pos_marker.source_slice,
+                indent_seg.pos_marker.templated_slice,
+            )
         raise NotImplementedError(
-            "Unexpected templated indent. Report this as a bug on GitHub."
+            "Unexpected templated indent. Report this as a bug on "
+            f"GitHub. Segment: {indent_seg}"
```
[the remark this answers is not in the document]
alanmcruickshank: Good idea.

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -925,3 +935,455 @@ def lint_indent_points(
...
+            # Then check for cases where we have no other options.
+            elif not matched_indents:
+                # NOTE: In this case we have no options for shortening the line.
+                # We'll still report a linting issue - but no fixes are provided.
+                reflow_logger.debug("    Handling as unfixable line.")
+                fixes = []
+
+            # Lastly deal with the "normal" case.
+            else:
+                # As a blunt solution, force indents at the lowest available
```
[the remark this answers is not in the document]
alanmcruickshank: Ah - I mean here that it's not a very smart solution, but that it works as a first pass. I'll choose some different wording.

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -925,3 +935,455 @@ def lint_indent_points(
...
+        else:  # pragma: no cover
+            raise NotImplementedError(
+                "Unexpected line position: %s", span.line_position
+            )
+        # NOTE: Operator precedence here is hard coded. It could be
+        # moved to configuration in the layout section in the future.
+        # Operator precedence is fairly consistent between dialects
+        # so for now it feels ok that it's coded here - it also wouldn't
+        # be a breaking change at that point so no pressure to release
+        # it early.
```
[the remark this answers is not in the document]
alanmcruickshank: I see where you're going and I don't see why not. I think this code would still be robust if we did go down that road so I think we're future proof for now.

src/sqlfluff/rules/L016.py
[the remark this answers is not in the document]
alanmcruickshank: Because of the way the code is structed, I actually think I'll leave it like this. I tried a couple of ways of re-doing it, but I think while they avoid the duplication they're harder to understand and reason about. This way is at least clear even if it does involve a little duplication.

src/sqlfluff/core/default_config.cfg
alanmcruickshank: This is corrected to:

```
# By setting a selection of clauses to "alone", we hint to the reflow
# algorithm that in the case of a long single line statement, the
# first place to add newlines would be around these clauses.
# Setting this to "alone:strict" would always _force_ line breaks
# around them even if the line isn't too long.
```

src/sqlfluff/core/parser/markers.py
alanmcruickshank: Good idea, but I think this needs to be a separate issue. Some of these assumptions aren't actually true. I think they *should* be but even in the vanilla code before this PR they're not. I'll make a separate issue.

alanmcruickshank: See #4183 4183

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -925,3 +935,455 @@ def lint_indent_points(
...
+        # of indent and dedent tokens. That's a job for another day
+        # however.
+        if seg.is_type("indent"):
+            continue
+        # Get the source position.
+        source_slice = seg.pos_marker.source_slice
+        # Is there a newline in the source string?
+        source_str = seg.pos_marker.source_str()
+        if "\n" in source_str:
+            # There is. Stop here. It's probably a complicated
```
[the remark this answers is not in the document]
alanmcruickshank: It's a good idea, but I think the position marker logic needs some love before I'm confident enough in that. See #4183

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -925,3 +935,455 @@ def lint_indent_points(
...
+                # distraction. Ignore it. It's probably a meta.
+                continue
+            # Otherwise if we're literal, use the raw length
+            # because it might be an edit.
+            elif seg.pos_marker.is_literal():
+                char_len += len(seg.raw)
+                last_source_slice = source_slice
+            # Otherwise assume it's templated code.
+            else:
+                char_len += slice_length(source_slice)
```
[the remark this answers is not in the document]
alanmcruickshank: It's a good idea, but I think the position marker logic needs some love before I'm confident enough in that. See #4183

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -925,3 +935,455 @@ def lint_indent_points(
...
+            continue
+
+        # As usual, indents are referred to by their "uphill" side
+        # so what number we store the point against depends on whether
+        # it's positive or negative.
+        indent_stats = e.get_indent_impulse()
+        # TODO: Better isolation and test coverage of this
+        # indexing maths. Very easy to get wrong. Necessary
+        # so we don't mistake empty elements, but potentially
+        # fragile nonetheless.
```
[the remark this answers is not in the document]
alanmcruickshank: Good idea. Have gone down the doctest route because I think it's neater.

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -925,3 +935,455 @@ def lint_indent_points(
...
+                line_no,
+                line_len,
+                line_length_limit,
+            )
+
+            # Potential places to shorten the line are either indent locations
+            # or segments with a defined line position (like operators).
+
+            # NOTE: We need the closing indent too for the next operations because
+            # it might be an option contain the closing indent for a potential
```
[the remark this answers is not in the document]
barrywhart: we'll miss the any locations -> we'll miss any locations

src/sqlfluff/core/parser/markers.py
```diff
@@ -107,11 +107,12 @@ def from_points(
                 start_point_marker.templated_slice.start,
                 end_point_marker.templated_slice.stop,
             ),
-            # The position markers should be the same so just pick
-            # the first.
+            # The templated file reference from the point markers
```
barrywhart, line 110: Should this be:
```suggestion
            # The templated file references from the point markers
```

src/sqlfluff/rules/L003.py
```diff
@@ -81,6 +81,4 @@ def _eval(self, context: RuleContext) -> List[LintResult]:
             .reindent()
             .get_results()
         )
-        return [
-            res for res in results if not res.description.startswith("Line is too long")
-        ]
+        return [res for res in results if res.source != "reflow.long_line"]
```
barrywhart, line 84: 👍

src/sqlfluff/rules/L016.py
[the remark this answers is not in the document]
barrywhart: No problem, it was kind of a nitpick.

src/sqlfluff/utils/reflow/helpers.py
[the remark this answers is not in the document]
barrywhart: Oh, the `.index()` call confused me (made me think `raw_segment` must be a direct child of `root_segment`, but I guess `.raw_segments` recursively returns **all** raw segments? It's a bit subtle. Should we consider renaming `.raw_segments` as `.raw_descendants` or `.raw_segment_descendants`?

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -1197,9 +1239,11 @@ def lint_line_length(
             # Potential places to shorten the line are either indent locations
             # or segments with a defined line position (like operators).
 
-            # NOTE: We need the closing indent too for the next operations because
-            # it might be an option contain the closing indent for a potential
-            # indent earlier in the line.
+            # NOTE: We make a buffer including the closing point, because the we're
```
barrywhart, line 1242:
```suggestion
            # NOTE: We make a buffer including the closing point, because we're
```

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -1197,9 +1239,11 @@ def lint_line_length(
...
+            # looking for pairs of indents and dedents. The closing dedent for one
+            # of those pairs might be in the closing point so if we don't have it
+            # then we'll miss the any locations which have their closing dedent at
```
barrywhart, line 1245:
```suggestion
            # then we'll miss any locations which have their closing dedent at
```

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -1309,20 +1353,24 @@ def lint_line_length(
 
             # Lastly deal with the "normal" case.
             else:
-                # As a blunt solution, force indents at the lowest available
-                # option.
-                reflow_logger.debug("    Handling as normal line.")
+                # For now, the algorithm we apply isn't particularly elegant
+                # and just finds the "outermost" opportunity to add additional
+                # line breaks and adds them. It doesn't add more them at more
+                # levels or selectively to add indents preferentially to an
```
barrywhart, line 1359: Can you review these two lines? This part reads strangely:
>add more them at more
>levels or selectively to add indents

barrywhart (approved):
A few more wording issues with comments. One possible suggestion about renaming a recursive property or otherwise clarifying some subtle code. Overall looks good, approving!

src/sqlfluff/utils/reflow/helpers.py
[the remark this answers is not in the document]
alanmcruickshank: Potentially yes - although that turns into a much much bigger fix. I'm going to leave it for now, but I think you're right that it could do with some clarity improvements.

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: Yeah I'll rethink. It's correct, but unclear.

PR #4083, opened 2022-11-18
Switch reflow buffer from LintFix to LintResult

alanmcruickshank:
This is *not* a breaking change on existing released features so can be merged before some of the other things. It does bring in a couple of features from #3942, but I'll resolve conflicts with that separately.

What this does do is change the `embodied_fixes` feature of `ReflowSequence` from a list of `LintFix` to a list of `LintResult`.

# Why?

Four reasons:
1. There's existing weird code paths in some rules to _backward engineer_ some rule descriptions in some code and it's hacky. This allows the generation of those descriptions inside the reflow code to make them much more consistent and not how to work backwards.
2. The long lines PR (#4067) is going to need the ability to group multiple fixes together before returning and `LintResult` objects are the most sensible way to do this.
3. The reindent PR (#3942) currently solves this using descriptions on `LintFix` objects, which I think it also a bit of a hack. This allows us to not need that anymore.
4. Looking ahead to the rules reorg (#4031), there's already quite a few rules which do different things, and the errors they return aren't very consistently formatted. We're going to have to address that at some point.

# Effect

While this isn't a *breaking* change, it does change the text description returned by several rules. IMHO I think they're much more consistent, and importantly that sets the groundwork for merging some of these rules together in the near term.

As examples of the new return values we have:

```
L006:
- Expected single whitespace between binary operator '+' and numeric literal.

L007:
- Found trailing binary operator '+'. Expected only leading near line breaks.

L008:
- Expected single whitespace between comma ',' and numeric literal.

L019:
- Found leading comma ','. Expected only trailing near line breaks.

L023:
Expected single whitespace between 'as' keyword and start bracket '('.

L039:
- Unexpected whitespace before comma ','.
- Expected only single space before binary operator '+'. Found '   '.

L048:
- Expected single whitespace between quoted literal and 'AS' keyword

L065:
- Set operators should always have a line break both before and after.
```

This PR adds the code paths to generate these descriptions centrally.


src/sqlfluff/cli/formatters.py
```diff
@@ -22,10 +22,25 @@
...
+    For longer sections it will split at an appropriate point.
+    >>> split_string_on_spaces("abc def ghi", line_length=7)
+    ['abc def', 'ghi']
+
+    After splitting, multi-space sections should be intact.
+    >>> split_string_on_spaces("a '   ' b c d e f", line_length=11)
+    ["a '   ' b c", 'd e f']
+    """
     line_buff = []
     str_buff = ""
```
barrywhart, line 40: Not part of this PR, but I notice that this function builds a string through repeated concatenation, which can be slow (n^2) behavior. Should we consider using `io.StringIO` to build the string instead? AFAIK, that's the canonical Python way to build a string incrementally.

src/sqlfluff/rules/L005.py
```diff
@@ -43,20 +43,22 @@ class Rule_L005(BaseRule):
     groups = ("all", "core")
     crawl_behaviour = SegmentSeekerCrawler({"comma"})
 
-    def _eval(self, context: RuleContext) -> Optional[LintResult]:
+    def _eval(self, context: RuleContext) -> List[LintResult]:
```
barrywhart, line 46: Nit: `BaseRule._eval()` returns `EvalResultType`. The rule implementations (like this one) often return something else. Would be cool at some point if the rules were consistent.

src/sqlfluff/utils/reflow/helpers.py
```diff
@@ -0,0 +1,27 @@
...
+from sqlfluff.core.rules.base import LintFix, LintResult
+from sqlfluff.core.parser import BaseSegment
+
+
+def fixes_from_results(results: Iterable[LintResult]) -> List[LintFix]:
+    """Return a list of fixes from an iterable of LintResult."""
+    return list(chain.from_iterable(result.fixes for result in results))
+
+
+def pretty_segment_name(segment: BaseSegment) -> str:
```
barrywhart, line 15: Nice!

src/sqlfluff/utils/reflow/respace.py
```diff
@@ -228,28 +250,38 @@ def handle_respace__inline_with_space(
...
-        )
         segment_buffer.pop(ws_idx)
-        return segment_buffer, new_fixes
+        if next_block:
+            description = (
+                "Unexpected whitespace before "
+                f"{pretty_segment_name(next_block.segments[0])}."
+            )
+        else:  # pragma: no cover
+            # This clause isn't has no test coverage because next_block is
```
barrywhart, line 273:
```suggestion
            # This clause has no test coverage because next_block is
```

src/sqlfluff/utils/reflow/sequence.py
```diff
@@ -68,53 +69,24 @@ def __init__(
         # Alternatively pictured: This is the list of fixes required
         # to generate this sequence. We can build on this as we edit
         # the sequence.
-        self.embodied_fixes: List[LintFix] = embodied_fixes or []
+        # Rather than saving *fixes* directly, we packages them into
```
barrywhart, line 72:
```suggestion
        # Rather than saving *fixes* directly, we package them into
```

barrywhart (approved):
Looks good overall. A few typo and grammar fixes. A couple of lower-priority suggestions.

src/sqlfluff/cli/formatters.py
alanmcruickshank: Good point - I've just done some experiments, and it's a little tricky because we also need to call `len()` on it to do the stacking and that doesn't work easily on `StringIO`. Good idea though.

PR #4195, opened 2022-12-21
Codecov migration

alanmcruickshank:
This might fail but I hope it works.

The reason codecov isn't reporting is that the `Python 3.10 Windows tests` job isn't reporting. The error we're getting is `{'detail': ErrorDetail(string='Missing "owner" argument. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}`. We're currently using the v1 uploader and that has been officially sunset: https://github.com/codecov/codecov-action.

This updates to v3 (and hopefully solves our codecov issues).

See here for example: https://github.com/sqlfluff/sqlfluff/actions/runs/3749312506/jobs/6367654414

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -63,6 +63,7 @@ indented_joins = False
 indented_ctes = False
 indented_using_on = True
 indented_on_contents = True
+allow_implicit_indents = False
 template_blocks_indent = True
 # This is a comma seperated list of elements to skip
 # indentation edits to.
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -78,6 +78,8 @@ jobs:
     - name: Upload Coverage Report
       uses: codecov/codecov-action@v3
       with:
+        # Public token to allow coverage from forks to work.
+        token: fd44b020-c716-4f72-b5f3-03fa20119956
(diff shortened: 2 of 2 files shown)


alanmcruickshank:
Up to date thread on codecov: https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954

Suggests adding a token to resolve. So that this works for PRs from forks, I'm going to add a token which is public. I think the security risks of doing so are minimal. FerretDB have done exactly that: https://github.com/FerretDB/FerretDB/blob/f4de6e41fc6d0ab9f00ac9e2facdda877fca86ee/.github/workflows/go.yml#L23-L25

src/sqlfluff/core/default_config.cfg
```diff
@@ -63,6 +63,7 @@ indented_joins = False
 indented_ctes = False
 indented_using_on = True
 indented_on_contents = True
+allow_implicit_indents = False
```
barrywhart, line 66: Did you mean to include this line? Seems unrelated to coverage checks.

barrywhart (approved):

src/sqlfluff/core/default_config.cfg
alanmcruickshank: oops - no I didn't but it's not harmful 🤷

PR #4198, opened 2022-12-22
Fixed False Positive for L037

### LINKED ISSUE #4194: L037 False Positive with explicit ORDER BY expressions and dbt
### Search before asking

### What Happened

Using sqlfluff with dbt I had trouble getting the following SQL query to work:

```sql
SELECT * FROM a_table
ORDER BY a ASC, b DESC
```

sqlfluff would errantly throw and L037 violation of `Ambiguous order by clause. Order by clauses should specific order direction for ALL columns or NO columns`

### Expected Behaviour

No L037 should be triggered

### Observed Behaviour

I traced the issue back to this line in the code:

https://github.com/sqlfluff/sqlfluff/blob/edb79809d2d7e16933e977d72f7a7d7b491c8cbc/src/sqlfluff/rules/L037.py#L65

I _think_ this can be refactored to just be `if found_column_reference and ordering_reference:` to resolve. The issue _may_ have something to do with the fact that (at least using dbt) the parsed segments of the above query look something like:

```
<ColumnReferenceSegment: >
<Dedent: >
<TemplateLoop: >
<Indent: >
<WhitespaceSegment: ' '>
<KeywordSegment: 'ASC'>
```

So the `TemplateLoop` I think throws off the current logic

### How to reproduce

`sqlfluff --dialect redshift <the_file>` where <the_file> contains:

```sql
SELECT * FROM a_table
ORDER BY a ASC, b DESC
```

### Dialect

redshift

### Version

1.4.2

### Configuration

[sqlfluff]
templater = dbt

[sqlfluff:templater:dbt]
project_dir = <the_folder>

[sqlfluff:templater:jinja]
apply_dbt_builtins = True

### Code of Conduct

WillAyd:
fixes #4194


alanmcruickshank (changes requested):
It looks like your fix *also* handles the clause just after your edits labelled `# Special handling for last column` (because it's now no longer covered in tests). Can you validate that you think that's correct functioning of your changes? If it is, then delete those lines.

Otherwise the logic here looks sound. Thanks for putting this together.

WillAyd:
Yea saw that with codecov - didn't realize that about the last column. That helps simplify the implementation a bit in the latest commit

src/sqlfluff/rules/L037.py
```diff
@@ -62,12 +62,7 @@ def _get_orderby_info(segment: BaseSegment) -> List[OrderByColumnInfo]:
...
-            elif column_reference and child_segment.type not in [
-                "keyword",
-                "whitespace",
-                "indent",
-                "dedent",
-            ]:
+            if child_segment.raw == ",":
                 result.append(
                     OrderByColumnInfo(
                         column_reference=column_reference, order=ordering_reference
```
barrywhart, line 68: Getting a mypy error now.
```
src/sqlfluff/rules/L037.py:68: error: Argument "column_reference" to "OrderByColumnInfo" has incompatible type "Optional[Any]"; expected "BaseSegment"  [arg-type]
```

barrywhart:
There was a spurious dbt test failure, so I restarted that build. There is also a mypy failure, apparently legitimate.

barrywhart:
Since the original issue occurs with dbt, do you think it would be worth adding a more specific test that uses the dbt templater? (`plugins/sqlfluff-templater-dbt/test/`)

WillAyd:
Nice idea. Latest commit gives that a shot. FYI wasn't able to test locally - I think the docker steps in the contributing guide need some updating

barrywhart:
What issues did you have testing on Docker?

You can also run the dbt tests using tox. I think the command is `tox -e dbt130`.

I use both methods occasionally.

WillAyd:
> What issues did you have testing on Docker?
> 
> You can also run the dbt tests using tox. I think the command is `tox -e dbt130`.
> 
> I use both methods occasionally.

Looks like the shell script uses `docker-compose` to launch instead of the newer `docker compose`. Easy enough to fix, but then inside the container trying to run things gives you an error about the `dbt` templater not being available and the warning:

`WARNING  sqlfluff.config:config.py:911 Starting in sqlfluff version 0.7.0 the dbt templater is distributed as a separate python package. Please pip install sqlfluff-templater-dbt to use it.
`

`pip install sqlfluff-templater-dbt` fixes that, but then get a lot of errors like:

```sh
E       assert [SQLTemplater...pendencies.')] == []
E         Left contains one more item: SQLTemplaterError('dbt error: dbt found 1 package(s) specified in packages.yml, but only 0 package(s) installed in dbt_packages. Run "dbt deps" to install package dependencies.')
E         Full diff:
E         - []
E         + [SQLTemplaterError('dbt error: dbt found 1 package(s) specified in packages.yml, but only 0 package(s) installed in dbt_packages. Run "dbt deps" to install package dependencies.')]

plugins/sqlfluff-templater-dbt/test/templater_test.py:422: AssertionError
```

Found the yml file from there, so did

```sh
pushd ./plugins/sqlfluff-templater-dbt/test/fixtures/dbt/dbt_project/
dbt deps
popd
```

After all that got it to work

barrywhart:
@WillAyd: See PR #4201. I think I fixed all the issues you found. Thanks for pointing those out! They're the kind of issues mainly seen with a fresh clone, which I rarely do. 🙏

plugins/sqlfluff-templater-dbt/test/fixtures/dbt/dbt_project/models/my_new_project/L037_test.sql
```diff
@@ -0,0 +1,8 @@
+select
+    birth_date,
+    name
+from cows
+order by
```
barrywhart, line 5: I think the original issue you mentioned involved a `TemplateLoop`. Should the SQL file have a `{% for %}` loop in order to reproduce the issue?

WillAyd: Interestingly enough the original issue I found this in had no templating. Not sure why that was in the parse results - maybe a separate issue?

barrywhart: @alanmcruickshank: Any ideas on this?

To me, as long as the PR fixes _an_ issue that is covered by newly added tests, that's progress, and if something arises later, we can address it at that time.

alanmcruickshank: agreed

alanmcruickshank (approved):
Looking great 👍 . Thanks for this

PR #4199, opened 2022-12-22
Tweak Coveralls settings

barrywhart:
### Pull Request checklist

  - Other.


.github/workflows/ci-tests.yml
```diff
@@ -74,7 +74,7 @@ jobs:
         pip install tox
     - name: Run the tests
       run: |
-        tox -e py -- -n 2 test
+        tox -e py -- --cov=sqlfluff -n 2 test
```
barrywhart, line 77: For test runs that upload coverage checks, we want to specify **which** code we're targeting for coverage. Previously, we were always specifying `sqlfluff`, even for the dbt templater tests, which **should've** been targeting `sqlfluff_templater_dbt`.

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -32,12 +32,10 @@
...
-if DBT_VERSION_TUPLE >= (1, 3):
-    COMPILED_SQL_ATTRIBUTE = "compiled_code"
-    RAW_SQL_ATTRIBUTE = "raw_code"
-else:
-    COMPILED_SQL_ATTRIBUTE = "compiled_sql"
-    RAW_SQL_ATTRIBUTE = "raw_sql"
+COMPILED_SQL_ATTRIBUTE = (
+    "compiled_code" if DBT_VERSION_TUPLE >= (1, 3) else "compiled_sql"
+)
+RAW_SQL_ATTRIBUTE = "raw_code" if DBT_VERSION_TUPLE >= (1, 3) else "raw_sql"
```
barrywhart, line 38: The `if/else` was causing uncovered lines. Rather than adding `pragma: no cover`, I changed to use the ternary operator.

plugins/sqlfluff-templater-dbt/test/conftest.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+@pytest.fixture(scope="session", autouse=True)
+def dbt_flags():
+    """Set dbt flags for dbt templater tests."""
+    # Setting this to True disables some code in dbt-core that randomly runs
+    # some test code in core/dbt/parser/models.py, ModelParser. render_update().
+    # We've seen occasional runtime errors from that code:
+    # TypeError: cannot pickle '_thread.RLock' object
+    dbt.flags.USE_EXPERIMENTAL_PARSER = True
```
barrywhart, line 13: Possible fix for the sporadic build failure I mentioned on Slack.

tox.ini
```diff
@@ -142,7 +142,9 @@ testpaths = test
 
 [coverage:run]
 source = src/sqlfluff
-omit = src/sqlfluff/__main__.py
+omit =
+    src/sqlfluff/__main__.py
+    plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/osmosis/*.py
```
barrywhart, line 147: `osmosis` is vendored code, so naturally it's not 100% covered by our tests.

.github/workflows/ci-tests.yml
```diff
@@ -201,7 +201,7 @@ jobs:
         pip install tox
     - name: Run the tests
       run: |
-        tox -e ${{ matrix.dbt-version }} -- plugins/sqlfluff-templater-dbt
+        tox -e ${{ matrix.dbt-version }} -- --cov=sqlfluff_templater_dbt plugins/sqlfluff-templater-dbt
```
barrywhart, line 204: As mentioned above, this change means now we're checking coverage for the dbt templater, too.

.github/workflows/ci-tests.yml
```diff
@@ -243,20 +243,13 @@ jobs:
...
         mkdir temp_pytest
-        python -m tox -e winpy -- -n 2 test
+        python -m tox -e winpy -- --cov=sqlfluff -n 2 test
     - name: Upload Coverage Report
       uses: codecov/codecov-action@v3
       with:
         # Public token to allow coverage from forks to work.
         token: fd44b020-c716-4f72-b5f3-03fa20119956
         files: ./coverage.xml
-    - name: Coveralls Parallel
```
barrywhart, line 253: Removed this because I've confirmed we're hitting 100% coverage without running Windows tests. Removing this means we see coverage results **faster** on PR builds, because the Windows build almost always finishes after the Linux one.

README.md
```diff
@@ -1,7 +1,5 @@
 ![SQLFluff](https://raw.githubusercontent.com/sqlfluff/sqlfluff/main/images/sqlfluff-wide.png)
 
-[![Coverage Status](https://coveralls.io/repos/github/sqlfluff/sqlfluff/badge.svg?branch=main)](https://coveralls.io/github/sqlfluff/sqlfluff?branch=main)
```
barrywhart, line 3: Moving the Coverall badge into the section with the other ones.

.github/workflows/ci-tests.yml
alanmcruickshank: Sounds sensible. We can always bring it back if we end up with more windows specific code.

tox.ini
```diff
@@ -33,7 +33,8 @@ commands =
     # Clean up from previous tests
     python {toxinidir}/util.py clean-tests
     # Run tests
-    pytest -vv -rsfE --cov=sqlfluff --cov-report=xml --cov-report=lcov {posargs: {toxinidir}/test} -m "not integration_test"
+    pytest -vv -rsfE --cov-report=xml --cov-report=lcov {posargs: {toxinidir}/test} -m "not integration_test"
+    python test/patch_lcov.py
```
alanmcruickshank, line 37: I don't think if I understand *why* we need to do this 🤔 . I don't think it's harmful, so no need to disapprove, but if you're able to add some commentary here (and/or in the `.py` file) then I think that makes this more future proof.

alanmcruickshank (approved):
One place where I think some more explanation would make this more future proof - but otherwise looks good. 👍

tox.ini
barrywhart: Will do.

I'm adding this to the docstring in the `patch_lcov.py` script.

```
Context: When the CI build runs tests, it uses tox, which installs SQLFluff
in a virtual environment. Thus, iThe lcov file generated by the tests contains
paths to the virtual environment. This script replaces those paths with
paths relative to the repo root. This allows the lcov file to be used by
Coveralls. Without this, Coveralls has valid coverage info, but it generates
URLs that point to source files that don't exist in the SQLFluff GitHub repo.

For example, we want to change this:
SF:.tox/py/lib/python3.10/site-packages/sqlfluff/__init__.py

to this:
SF:src/sqlfluff/__init__.py
```

PR #4202, opened 2022-12-24
Lint and fix areas of templated code that are not actually executed

barrywhart:
### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/templaters/base.py
+++ b/src/sqlfluff/core/templaters/base.py
@@ -90,6 +90,7 @@ class TemplatedFileSlice(NamedTuple):
     slice_type: str
     source_slice: slice
     templated_slice: slice
+    slice_idx: Optional[int] = None
 
 
 class RawSliceBlockInfo(NamedTuple):
@@ -142,7 +143,7 @@ class TemplatedFile:
         # If we get here and we don't have sliced files, then it's raw, so create them.
         self.sliced_file: List[TemplatedFileSlice] = sliced_file or [
             TemplatedFileSlice(
-                "literal", slice(0, len(source_str)), slice(0, len(source_str))
+                "literal", slice(0, len(source_str)), slice(0, len(source_str)), 0
             )
         ]
         self.raw_sliced: List[RawFileSlice] = raw_sliced or [
--- a/src/sqlfluff/core/templaters/slicers/tracer.py
+++ b/src/sqlfluff/core/templaters/slicers/tracer.py
@@ -182,6 +182,7 @@ class JinjaTracer:
                     else len(self.raw_str),
                 ),
                 slice(self.source_idx, self.source_idx + target_slice_length),
+                slice_idx,
             )
         )
         if slice_type in ("literal", "templated"):


src/sqlfluff/core/templaters/jinja.py
```diff
@@ -513,9 +517,46 @@ def slice_file(
...
-        tracer = analyzer.analyze(make_template)
-        trace = tracer.trace(append_to_templated=kwargs.pop("append_to_templated", ""))
-        return trace.raw_sliced, trace.sliced_file, trace.templated_str
+        tracer_probe = analyzer.analyze(make_template)
+        tracer_copy = copy.deepcopy(tracer_probe)
+        append_to_templated = kwargs.pop("append_to_templated", "")
+        trace = tracer_probe.trace(append_to_templated=append_to_templated)
+        print(f"Yielding trace for {trace.templated_str!r}")
+        yield trace.raw_sliced, trace.sliced_file, trace.templated_str
+        return
```
barrywhart, line 526: Remove this `return` to enable "multi-linting"

test/fixtures/linter/autofix/ansi/028_l003_lint_unreached_code/after.sql
```diff
@@ -0,0 +1,7 @@
+{% if True %}
```
barrywhart, line 1: @alanmcruickshank: Could you take a look at the failing test and let me know what you think?

The linter returns a proper list of the issues in both parts of the file (reachable and unreachable):
```
$ sqlfluff lint test/fixtures/linter/autofix/ansi/028_l003_lint_unreached_code/before.sql
== [test/fixtures/linter/autofix/ansi/028_l003_lint_unreached_code/before.sql] FAIL                                                                                                                 
L:   2 | P:   1 | L003 | Expected indent of 4 spaces.
L:   5 | P:   1 | L003 | Expected indent of 4 spaces.
L:   6 | P:   1 | L003 | Expected indent of 4 spaces.
All Finished 📜 🎉!
```

But when running `sqlfluff fix`, only the first fix is applied, resulting in:
```
{% if True %}
    SELECT 1
{% else %}
    SELECT c
FROM t
WHERE c < 0
{% endif %}
```

My intuition is that the differing `sliced_file` contents and/or related things like placeholder segments prevent applying all the fixes directly. _If_ that's the case, one possible workaround would be to apply the fixes separately and combine the results.

I'd like to get your thoughts on this before beginning a deeper dive.

alanmcruickshank: Linking this up to what I posted on slack:

> Without actually running the code I'm not 100% sure, but I have a theory on why this might happen. When running the fix logic - the fixes don't patch the file - they overwrite it with a reconstructed version. So if we read the file twice, fix it twice and then try and write them again - only the fixes from one branch will be applied. The others with just be overwritten by the second branch which is written.

>
 In my mind, there are three ways around this problem.

> 1. Change the fix logic to try and patch rather than overwrite. Deep down - I think this is a losing battle, because all the references will be wrong if the file has been also modified by the other branch.
> 2. If we're in a fixing mode - explicitly lint the other branch after it has already been fixed by the first. This is probably the easier to implement - but feels clunky given the UX.
> 3. It might be possible to somehow condense or coalesce the two results into one file before writing. This is probably my preferred option. Details...

> If neither branch has applied their fixes - then the two `TemplatedFile` references should be compatible. They won't be identical (because you'll have fiddled the source file to force the branch evaluation) - but you'd be able to check whether they were otherwise identical. With a little finesse, you could take two `LintedFile` references, both with a set of fixes, different fixed files, but compatible raw files, and turn them into one entity with a combined set of fixes, which would enable the reconstructed file to contain both. You're effectively making a combined result file - before doing any interaction with the filesystem - and then only writing once.

> Logic would be something like:
> 1. Lint, fix etc... get as far as where we try to apply fixes.
> 2. In the fix application logic, (or perhaps just before), we group together any `LintedFile` objects which reference the same source file.
> 3. Identify which is the unaltered variant, and use that as a the base model.
> 4. Identify sections of the raw unaltered variant which aren't covered in the rendered version (code which you've already written - so maybe this needs to be encoded into a structure which can be used again later) - and importantly we also need to be able to map the location of these sections in both the raw source file and the altered source file.
Somewhere in the file reconstruction logic - whenever we reach an unrendered section in the unaltered variant - we look in the other variants to see if they want to make any changes there (once changes are remapped) and then bring them into the reconstructed source file at that stage.
> 5. File the fixed file once using this combined version.

> the place you'd need to make most of these changes would probably be in `LintedFile.fix_string` - although this logic would now need to be "multi-variant-aware". We still need to run `_generate_source_patches()` for each variant.
> The edits would need to then make `_slice_source_file_using_patches` and `_build_up_fixed_source_string` work with patches from multiple variants to then construct the combined file.

alanmcruickshank:
Just picking up on this for anyone watching it. Regardless of the implementation here, this feature would likely mean longer runtimes (because we end up parsing and linting multiple versions of the same file). In the best case scenario we can use the "base" variant as a hint for parsing the other versions, but it's still likely going to be longer.

I'll pick this up once I've made a fair inroad into that performance frontier.

barrywhart:
@alanmcruickshank: A followup to your performance comment -- that is true, but we have coverage information on the rendered template and can skip the whole process of generating and linting variants if the initial rendering covered everything. So in most cases, there'll be no impact on performance. I think the draft PR does this.

alanmcruickshank:
Oh, totally agree for files that are fully covered in the first pass, it's the others that will experience the slowdown

peterbud:
Closed by #7518

PR #4208, opened 2022-12-26
don't pickle the templater

alanmcruickshank:
In testing the 2.0.0a1 release, I repeatedly experienced pickling issues with the dbt templater.

The problem is in multi-threaded mode, that the `FluffConfig` object gets passed around, and the new dbt templater contains some non-pickle-happy objects. In child processes, the templater shouldn't be required (the same as the plugin manager), so this PR strips it out of the `FluffConfig` object before pickling.

### DIFF
--- a/src/sqlfluff/core/config.py
+++ b/src/sqlfluff/core/config.py
@@ -797,6 +797,12 @@ class FluffConfig:
         state = self.__dict__.copy()
         # Remove the unpicklable entries.
         del state["_plugin_manager"]
+        if "templater_obj" in state["_configs"]["core"]:
+            # If it's already removed, don't try again.
+            # The dbt templater doesn't pickle well, but isn't required
+            # within threaded operations. If it was, it could easily be
+            # rehydrated within the thread.
+            del state["_configs"]["core"]["templater_obj"]
         return state
 
     def __setstate__(self, state):  # pragma: no cover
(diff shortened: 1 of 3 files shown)


src/sqlfluff/core/config.py
```diff
@@ -797,6 +797,12 @@ def __getstate__(self):
         state = self.__dict__.copy()
         # Remove the unpicklable entries.
         del state["_plugin_manager"]
+        if "templater_obj" in state["_configs"]["core"]:
```
barrywhart, line 800: Nitpick: IIRC, you could use `pop()` and pass a default value. Same behavior, but just one line of code.

src/sqlfluff/core/linter/runner.py
```diff
@@ -37,7 +37,7 @@ def __init__(
 
     pass_formatter = True
 
-    def iter_rendered(self, fnames: List[str]) -> Iterator[Tuple]:
+    def iter_rendered(self, fnames: List[str]) -> Iterator[Tuple[str, RenderedFile]]:
```
barrywhart, line 40: What's this change? Was the existing type annotation wrong?

barrywhart (approved):
A couple of questions, but looks reasonable to merge.

src/sqlfluff/core/linter/runner.py
alanmcruickshank: It wasn't wrong, just a bit vague. A more specific one would have been useful in debugging so I thought I'd improve it while I was there.

src/sqlfluff/core/config.py
alanmcruickshank: Oohhh great idea. Thanks for the tip. I did think my approach was a little clumsy.

barrywhart:
Update looks good! ✅

Issue #4205, opened 2022-12-25
ANSI: SELECT ... FOR UPDATE not supported


Yiwen-Gao:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

I have a SELECT statement that retrieves the oldest element from a table, but sqlfluff is unable to lint it correctly.
```sql
SELECT job
FROM queue
ORDER BY last_update_time ASC
FOR UPDATE SKIP LOCKED
LIMIT 1;
```

### Expected Behaviour

I think sqlfluff should support the `FOR UPDATE` clause because I believe it's part of the [ANSI SQL standard](https://cockroachlabs.com/blog/select-for-update/#what-is-select-for-update).

### Observed Behaviour

When I try to lint, sqlfluff fails to parse `FOR UPDATE` correctly.
```
L: 209 | P:   1 |  PRS | Line 209, Position 1: Found unparsable section: 'FOR
                       | UPDATE SKIP LOCKED'
```

### How to reproduce

- Use any SELECT query with a FOR UPDATE clause
- Lint the query with sqlfluff

### Dialect

Postgres

### Version

sqlfluff: 1.4.5
Python: 3.8

### Configuration

[sqlfluff]
dialect=postgres
exclude_rules=
templater=placeholder
large_file_skip_byte_limit=0

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

Yiwen-Gao:
I see there's a `ForClauseSegment` for MySQL--would it be better to add it for Postgres or all dialects?
https://github.com/sqlfluff/sqlfluff/blob/83636f24350483de31438441e050a8c54b61a187/src/sqlfluff/dialects/dialect_mysql.py#L1619-L1635

PR #4209, opened 2022-12-26
postgres: add row-level locks to SELECT statements

Yiwen-Gao:

These changes address the problem raised in https://github.com/sqlfluff/sqlfluff/issues/4205 and enable sqlfluff to parse Postgres SELECT queries that have row-level locks. For example, ones of the form:
```sql
SELECT * 
FROM mytable 
FOR UPDATE SKIP LOCKED;
``` 
I saw that https://github.com/sqlfluff/sqlfluff/commit/e85fcbb1be134a46fe3444fc850caa0b3287d5ee added row-level locks to MySQL, but none of the other dialects support this feature yet. Should I make a follow-up issue to add locks to `dialect_ansi.py`?

N/A

### Pull Request checklist


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1171,6 +1209,9 @@ class SelectStatementSegment(ansi.SelectStatementSegment):
             Ref("LimitClauseSegment", optional=True),
             Ref("NamedWindowSegment", optional=True),
         ]
+    ).copy(
+        insert=[Ref("ForClauseSegment", optional=True)],
+        before=Ref("LimitClauseSegment", optional=True),
```
Yiwen-Gao, line 1214: The `ForClauseSegment` is in MySQL's `UnorderedSelectStatementSegment`'s match and parse grammars, but I only added it to Postgres's `SelectStatementSegment`'s parse grammar for my test cases to pass. 
https://github.com/sqlfluff/sqlfluff/blob/83636f24350483de31438441e050a8c54b61a187/src/sqlfluff/dialects/dialect_mysql.py#L1543-L1578
Let me know if I'm missing anything.

greg-finley:
Check the CI failure, looks like a linting thing.

I agree it's better to put in ANSI if it applies to many dialects. Do you feel that's worth a shot? We could try to remove the overrides in MySQL. Otherwise with your change we'd have overrides in both MySQL and Postgres that really belong in ANSI.

greg-finley:
Thanks for doing the research, let’s go with this for now then!

greg-finley (approved):

PR #4213, opened 2022-12-28
don't consider templated whitespace

alanmcruickshank:
Found this bug while testing the new 2.0.0a1 release (#4203). I think this is the last one we need before releasing that, but will confirm on that thread once I've done a little more testing.

This PR fixes a bug where the reindentation logic would panic when being passed a templated indent. The code which finds indent segments wasn't robust to that chance and so wasn't checking whether indents were templated or not. This fixes that (includes test case).

I improved some of the reflow logging too while I was in there which was hard to debug with.

### DIFF
--- a/test/fixtures/rules/std_rule_cases/L003.yml
+++ b/test/fixtures/rules/std_rule_cases/L003.yml
@@ -1401,3 +1401,19 @@ test_fail_deny_implicit_indent:
   configs:
     indentation:
       allow_implicit_indents: False
+
+test_pass_templated_newlines:
+  # NOTE: The macro has many newlines in it,
+  # and the calling of it is indented. Check that
+  # this doesn't panic.
+  pass_str: |
+    {% macro my_macro() %}
+      
+      macro
+      + with_newlines
+
+    {% endmacro %}
+
+    SELECT
+        {{ my_macro() }} as awkward_indentation
+    FROM foo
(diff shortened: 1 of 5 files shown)


test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -1401,3 +1401,19 @@ test_fail_deny_implicit_indent:
   configs:
     indentation:
       allow_implicit_indents: False
+
+test_pass_templated_newlines:
+  # NOTE: The macro has many newlines in it,
+  # and the calling of it is indented. Check that
+  # this doesn't panic.
+  pass_str: |
```
barrywhart, line 1409: Can you add this test case (or at least confirm it works now)? I noticed this issue yesterday.

```
{% if False %}
SELECT 1
{% else %}
SELECT c
FROM t
WHERE c < 0
{% endif %}
```

In the `main` branch, it only finds and fixes *2* of the 3 bad indentations in the `{% else %}` block:
```
{% if False %}
SELECT 1
{% else %}
SELECT c
    FROM t
    WHERE c < 0
{% endif %}
```

This seems like it may be related to what you've fixed.

barrywhart (approved):
The PR looks good. I had found a linting/fixing issue yesterday which may be related to this one -- can you take a quick look at that before merging, and maybe add a test case for it if it seems appropriate? (Or we could create a separate issue if it's not related.)

test/fixtures/rules/std_rule_cases/L003.yml
alanmcruickshank: hmmm - doesn't solve that bug. I'll raise it as a new issue.

PR #4217, opened 2022-12-28
Fix #4215

### LINKED ISSUE #4215: Indentation bug on 2.0.0a1 release
### Search before asking

### What Happened

As raised on #4213 . This query is improperly fixed:

```sql
{% if False %}
SELECT 1
{% else %}
SELECT c
FROM t
WHERE c < 0
{% endif %}
```

on current main it fails as

```sql
{% if False %}
SELECT 1
{% else %}
SELECT c
    FROM t
    WHERE c < 0
{% endif %}
```

### Expected Behaviour

expected fix is that all three incorrectly indented lines should be fixed.

### Observed Behaviour

the `SELECT` line is not fixed.

### How to reproduce

see description above

### Dialect

ansi

### Version

ce5c5f228d92df339ef448e66dd3ef8ad859a9e2

### Configuration

default config

### Code of Conduct

alanmcruickshank:
Templated code checks were a bit aggressive, and for the places that a `replace` is kind of behaving like a `create_before` or `create_after`, we can utilise that fact and relax a little. This fixes #4215.

### DIFF
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -488,6 +488,11 @@ class BaseRule:
     _works_on_unparsable = True
     _adjust_anchors = False
     targets_templated = False
+    # Some fix routines do their own checking for whether their fixes
+    # are safe around templated elements. For those - the default
+    # safety checks might be inappropriate. In those cases, set
+    # template_safe_fixes to True.
+    template_safe_fixes = False
 
     # Lint loop / crawl behavior. When appropriate, rules can (and should)
     # override these values to make linting faster.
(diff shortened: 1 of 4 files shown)


barrywhart (approved):
The PR looks good. What are the SQLFluff checks? They are kind of distracting, given that most of our files deliberately have issues!

alanmcruickshank:
@barrywhart - can you re-review. My original approach didn't work.

We don't run an SQLFluff checks... Can you see where they're coming from?

barrywhart (approved):
Looks good!

I don't see the SQLFluff checks in the output now. Did something change? I've only seen them a few times, so they're either sporadic, or we only see output in some cases (if something is wrong?).

I'll take a look through our CI checks. @tunetheweb, @greg-finley, do either of you know about this? Something was checking for SQLFluff rule violations in some of the `.sql` files in the project.

tunetheweb:
Not sure what you mean? I see the checks on this PR fine.

We have a Critical CI Checks to test the rules don’t crash out on any of our SQL files, but not that they actually can be fixed.

barrywhart:
@tunetheweb: On an earlier build of the PR, I was seeing complaints about individual SQL files -- not failed tests, more like comments or annotations on the PR. It perhaps looked like someone had added a SQLFluff GitHub Action to the repo. Odd...

PR #4239, opened 2023-01-03
Push indents to after comments

### LINKED ISSUE #4223: 2.0.0a1 - Comments cause inconsistent indentation of CASE
### Search before asking

### What Happened

Given this query:

```sql
select
    case
        when a
        then b
    end as foo,
    case
        when a -- bar
        then b
    end as bar
from c
```

the fix is:

```sql
select
    case
        when a
            then b
    end as foo,
    case
        when
            a -- bar
        then b
    end as bar
from c
```

### Expected Behaviour

The comment causes a different treatment of the fix. This is true with and without implicit indents. The indentation without the comment is the "correct" one (at least as per https://github.com/dbt-labs/corp/blob/main/dbt_style_guide.md#example-sql).

### Observed Behaviour

see above.

### How to reproduce

`sqlfluff fix -r L003 -x f -d ansi test.sql --logger rules -vvvv`

### Dialect

snowflake

### Version

2.0.0a1

### Configuration

default

### Code of Conduct

alanmcruickshank:
This resolves #4223 . This fix is a little unorthodox - but it works.

When assessing the position of indent effects, we push the effects of any `Indent` or `Dedent` segments to _after_ any comment segments. Within a line it shouldn't make a difference, but for end of line comments, it makes sure that any tokens are available in the same `IndentPoint` that forms the start of the next line. That's the main purpose of what we're doing.

I considered actually moving the `Indent` and `Dedent` tokens in the lexing or parsing stage, but that felt like too much "hacking". Keeping this logic in the reflow interpretation code feels acceptable.


test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -1516,3 +1516,42 @@ test_fail_case_else_end_clause:
...
+        case
+            when a -- bar
+                then b
+        end as bar
+    from c
+
+test_fail_coverage_indent_trough:
+  # This test primarily tests the handling of closing trough indents
+  fail_str: |
+    WITH bar as (SELECT 1
```
barrywhart, line 1549: In the past, L003 only handled indentation. Any concerns that here it is moving code between lines?

alanmcruickshank: I don't think so - this is functionality that is already out in the wild in the treatment of existing hanging indents in `2.0.0a1`. All this PR does is fix a bug with it.

barrywhart (approved):

PR #4278, opened 2023-01-14
Parser Performance: Cache segment string repr to reduce function calls

WittierDinosaur:
The current version of __repr__ in raw.py doesn't cache the string. This means that every comparison causes a rebuild, and also recalculates the pos marker. This PR caches it as a property. I tested the parser performance of this PR on a large file (~2300 lines) using CProfile:

Before:
```
424944039 function calls (413269403 primitive calls) in 221.592 seconds
```
After
```
213425726 function calls (201751080 primitive calls) in 110.296 seconds
```

Not that I'm aware of

### Pull Request checklist

  - Other.


.gitignore
```diff
@@ -34,6 +34,7 @@ test-reports
 # Ignore root testing sql & python files
 /test*.sql
 /test*.py
+/test*.txt
```
alanmcruickshank, line 37: Why this?

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -58,12 +58,13 @@ def __init__(
...
-
-    def __repr__(self):
-        return "<{}: ({}) {!r}>".format(
+        self.representation = "<{}: ({}) {!r}>".format(
             self.__class__.__name__, self.pos_marker, self.raw
         )
 
+    def __repr__(self):
+        return self.representation
+
```
alanmcruickshank, line 67: Looks great 👍 . Can you add a short comment here to note the intent of `self.representation` for future-us? Feels like something we'll totally forget.

alanmcruickshank (commented):
This looks like such an excellent idea! Good thinking.

Couple of notes on the details, but otherwise I think good to go.

.gitignore
WittierDinosaur: Unrelated change - but when I use cProfile, I tend to output to .txt, so adding this (currently unused) namespace to give me somewhere in root to dump my outputs

src/sqlfluff/core/parser/segments/raw.py
WittierDinosaur: Done 👍

.gitignore
alanmcruickshank: Ok - I don't have a good reason for us to use txt files for much, so makes sense.

alanmcruickshank (approved):

PR #4279, opened 2023-01-14
Move ISSUE from Snwoflake reserved keywords to unreserved ones

KaoutherElhamdi:
[Slack thread](https://sqlfluff.slack.com/archives/C01GPHR8J0G/p1673275543319059)
There is no issue created for this purpose while the change is pretty small. (If it's mandatory I can create a quick one)

**SQLFLUFF** struggles to parse a Snowflake query that includes `ISSUE` as table name or column name.

For example : `select * from issue` 
Will throw a parsing error : `Found unparsable section : 'from issue'`

This is due to the fact that the keyword `ISSUE` is considered as a reserved word in **Snowflake**.
Otherwise `ISSUE` is not a reserved keyword (unless it is used in SHOW query) (check [Snowflake doc](https://docs.snowflake.com/en/sql-reference/reserved-keywords.html))
This means we can move ISSUE from reserved to unreserved keywords.


test/dialects/snowflake_test.py
```diff
@@ -17,6 +17,7 @@
             "create table orders_clone_restore clone orders at (timestamp => "
             "to_timestamp_tz('04/05/2013 01:02:03', 'mm/dd/yyyy hh24:mi:ss'));",
         ),
+        ("SelectStatementSegment", "SELECT issue FROM issue;"),
```
tunetheweb, line 20: Could you move this test to one of the select statement files in `test/fixtures/dialects/snowflake`?

We try to avoid python tests unless really necessary as SQL tests are 1) Easier to write, 2) Allow us to see the parse tree YML by automatically generating that allowing us to track any changes in that in future PRs.

More info here: https://github.com/sqlfluff/sqlfluff/wiki/Contributing-Dialect-Changes#adding-test-cases-for-your-changes

tunetheweb (changes requested):

tunetheweb (approved):
LGTM!

PR #4285, opened 2023-01-15
Fix #4252 (Resolve multiple sensible indents)

### LINKED ISSUE #4252: 2.0.0a2 Comment in a complex, nested sub-query breaks indentation of following SQL
### Search before asking

### What Happened

Took me a while to figure out what the issue here was, probably there is a simpler MWE, but I was only able reproduce it in this specific scenario. 

Sometimes a comment can cause problems with following indentation. Consider the following example, everything after the comment has 1 indent too much

```sql
WITH cte AS (
    SELECT *
    FROM (
        SELECT *
        FROM table
        WHERE
            NOT bool_column AND NOT bool_column
            AND some_column >= 1  -- This is a comment
        )
    ),

    SELECT *
    FROM cte
;

    SELECT *
    FROM table3
;
```

However, if I remove the comment everything gets linted correctly

```sql
WITH cte AS (
    SELECT *
    FROM (
        SELECT *
        FROM table
        WHERE
            NOT bool_column AND NOT bool_column
            AND some_column >= 1
    )
),

SELECT *
FROM cte
;

SELECT *
FROM table3
;
```

### Expected Behaviour

see above

### Observed Behaviour

see above

### How to reproduce

see above

### Dialect

BigQuery

### Version

```
sqlfluff, version 2.0.0a2
```

### Configuration

default

### Code of Conduct

alanmcruickshank:
This resolves #4252.

The issue arises from a bug in #4239. That PR did not account for both the point before and after the comment having important indent stats. This introduces a method for combining them so all the information is encoded.


test/fixtures/rules/std_rule_cases/L003.yml
```diff
@@ -1570,3 +1570,26 @@ test_fail_coverage_indent_trough:
...
+  # https://github.com/sqlfluff/sqlfluff/issues/4252
+  pass_str: |
+    WITH cte AS (
+        SELECT *
+        FROM (
+            SELECT *
+            FROM table
+            WHERE
+                NOT bool_column AND NOT bool_column
+                AND some_column >= 1
```
WittierDinosaur, line 1585: This doesn't have a comment in it?

alanmcruickshank: 🤦 . Good catch

WittierDinosaur (approved):
LGTM

PR #4303, opened 2023-01-19
toml: only install `toml` dependency if < Python 3.11 (otherwise use builtin `tomllib`)

kevinmarsh:

Now that Python 3.11+ supports reading `.toml` files with [the builtin `tomllib` module](https://docs.python.org/3/library/tomllib.html), we can just use that instead of installing the 3rd party package `toml`. On Python 3.10 and lower, `toml` will still be installed.

No, I've just tested this on Python 3.8 and Python 3.11, so will see what the full `ci-tests` returns (might need to tweak the `# pragma: no cover` comment).

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/config.py
+++ b/src/sqlfluff/core/config.py
@@ -4,6 +4,7 @@ import logging
 import os
 import os.path
 import configparser
+import sys
 from dataclasses import dataclass
 
 import pluggy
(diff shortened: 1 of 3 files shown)


src/sqlfluff/core/config.py
```diff
@@ -15,7 +16,14 @@
 
 import appdirs
 
-import toml
+if sys.version_info >= (3, 11):
+    import tomllib
+
+    tomllib_filemode = "rb"
```
kevinmarsh, line 22: Hmm actually I can just use `.loads` instead of `.load` and that way we don't need this binary flag, just retesting locally then will push

greg-finley (approved):
Nice, thank you!

PR #4327, opened 2023-01-28
Parser: Optimise lookahead_match

WittierDinosaur:
This PR optimises lookahead_match. It was doing a full lookahead, trimming the upper raw, and then iterating again through the segments catching all matches. This is unnecessary. The new version iterates through the matchers on each trimmed segment, allowing immediate exit on match.

The performance improvements are substantial. Tested on `CREATE USER danny;\n` * 2000

Profile before:
```
89404611 function calls (89139730 primitive calls) in 39.543 seconds
```

Profile after:
```
41622541 function calls (41357663 primitive calls) in 19.628 seconds
```
Though it actually runs in less than half of this time without the Profiler also running

No

### Pull Request checklist

  - Other.


src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -343,54 +343,40 @@ def _trim_elem(seg):
...
             )
 
-            while match_queue:
+            simple_match = None
+            for idx, seg in enumerate(segments):
+                for matcher, simple in simple_matchers:
+                    # Simple will be a tuple of options
+                    assert simple
+                    for simple_option in simple:
+                        if simple_option == _trim_elem(seg):
```
alanmcruickshank, line 360: I think we could further optimise this by computing `_trim_elem(seg)` outside this inner `for` loop, because it's always the same for a given `seg`.

src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -343,54 +343,40 @@ def _trim_elem(seg):
...
-                    # excluded. Log but then move on to the next item on the list.
-                    parse_match_logging(
-                        cls.__name__,
-                        "_look_ahead_match",
-                        "NM",
-                        parse_context=parse_context,
-                        v_level=4,
-                        _so=queued_option,
-                    )
+                # queued_matcher, queued_buff_pos, queued_option = match_queue.pop(0)
```
alanmcruickshank, line 367:
```suggestion
```

alanmcruickshank (approved):
Nice. I think the performance improvement here comes from reversing the sequence of the loops.
- previously we looped through all all indices for each matcher.
- now we're looping through each matcher for each index.

Thinking of the likely way this code path should work I think your suggestion is much better 👍 

I've added two nits (one commented line of code which I think we can remove), and one potential further performance improvement. Otherwise happy to go though.

src/sqlfluff/core/parser/grammar/base.py
alanmcruickshank: Maybe even better would be caching the trimmed element on the segment so that it's just a cache fetch.

PR #4314, opened 2023-01-25
Revise comment effects on indentation.

alanmcruickshank:
# NOTE: This PR has got too messy. I'm going to split it up and then close this one.

This resolves #4294 and resolves #4291 .

It builds on the logic introduced in #4239 , makes it more robust and more modular. The problems raised in both of the issues above I think arise from loopholes in this logic which I've now closed. Also been able to simplify a little in the process.

PR #4337, opened 2023-02-01
Resolve #4294: Comments affecting indentation

### LINKED ISSUE #4294: 2.0.0a3: Inline comments confusing `WHERE` indentation and line breaks
### Search before asking

### What Happened

The indentation of a `WHERE` statement (probably affects other areas as well) is getting indented differently based of the existence of an inline comment.

### Expected Behaviour

```sql
SELECT *
FROM table
WHERE 
    TRUE  -- some comment
    AND TRUE
```

This expected behaviour is based on the lint result of the same query without the comment

```sql
SELECT *
FROM table
WHERE
    TRUE
    AND TRUE
```

### Observed Behaviour

```sql
SELECT *
FROM table
WHERE TRUE  -- some comment
AND TRUE
```

### How to reproduce

Run lint/fix on any of the queries above.

### Dialect

BigQuery (but probably applies to all dialects)

### Version

sqlfluff, version 2.0.0a3

### Configuration

default

### Code of Conduct

alanmcruickshank:
This is part of #4314 - that PR is getting messy so this merges the easy bit. It resolves #4294.

This PR is mostly logging improvements, but the substantive part is to simplify the code path through `_lint_line_untaken_positive_indents` in the reindentation code to remove some routes which affect comment handling. Solving that however leads to some knock-on issues in L016. The ultimate solution to that for now is to better segregate those two rules rather than relying on the same `.reindent()` method. When addressing the rules reorg - we might re-merge them - but for now this makes a lot more sense.


greg-finley (approved):

greg-finley:
@alanmcruickshank CI failure here too

PR #4338, opened 2023-02-01
reindent refactor

alanmcruickshank:
This is some of the code elements of #4314 . This PR should have no substantive effect, it's just pure refactoring. It does however support some of the comment logic changes intended in the original PR.


WittierDinosaur:
Looks good - but you've lost a bit of code coverage

WittierDinosaur:
Looks good - but you've lost a bit of code coverage

alanmcruickshank:
@WittierDinosaur - right you are. I think that clause makes sense to keep - so I've marked it as no cover. Could you give me a review and I'll get this in?

WittierDinosaur (approved):
LGTM

Issue #3557, opened 2022-07-05
[postgres] Index USING → false negative


enote-kane:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Creating an index using a specific index method (which is not a function) gets mis-interpreted as a function name.

### Expected Behaviour

No error

### Observed Behaviour

```
sqlfluff lint --nocolor --verbose --dialect postgres 2022-07-05.index-using.sql
==== sqlfluff ====
sqlfluff:                1.1.0 python:                 3.8.10
implementation:        cpython verbosity:                   1
dialect:              postgres templater:               jinja
rules:                                all
==== readout ====

=== [ path: 2022-07-05.index-using.sql ] ===

== [2022-07-05.index-using.sql] FAIL                                                                                                                                              
L:   1 | P:  47 | L017 | Function name not immediately followed by parenthesis.
==== summary ====
violations:        1 status:         FAIL
All Finished!
```

### How to reproduce

Use the following file `2022-07-05.index-using.sql`:
```sql
CREATE INDEX my_json_idx ON my_table USING GIN (metrics_ jsonb_path_ops);
```

### Dialect

postgres

### Version

```
sqlfluff:                1.1.0 python:                 3.8.10
implementation:        cpython verbosity:                   1
dialect:              postgres templater:               jinja
rules:                                all
```

### Configuration

defaults

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

tunetheweb:
While it arguably might not technically be a function, is there any reason not to treat it as such for rule L017? Is it better to have no spaces here or not?

enote-kane:
@tunetheweb Well, I think detecting the construct `GIN (...)` as a function here may lead to further bugs in more complex index definition cases.

So I think it would be a much more future-proof solution to detect and support the `USING ...` construct.

Adding a more advanced case shows that this is problematic:
```sql
CREATE INDEX my_json_idx ON my_table USING GIN (metrics_ jsonb_path_ops);
CREATE INDEX my_json_idx ON my_table USING BTREE (name_ COLLATE icu_std text_pattern_ops (op_variable = 'value') DESC NULLS LAST);
```

...result:
```
[...]

PR #4356, opened 2023-02-05
Improving Postgres create index statement

### LINKED ISSUE #4355: Postgres CREATE INDEX opclass syntax fail parsing
### Search before asking

### What Happened

A parsing error happens in the following code:

```
CREATE INDEX opclass_index ON schema.opclass_table (col varchar_pattern_ops);
```

### Expected Behaviour

No parsing error

### Observed Behaviour

```
Found unparsable section: 'CREATE
                       | INDEX opclass_index ON schema.opc...'
```

### How to reproduce

1. Past the above code to a file called `file.yaml`
2. Run `sqlfluff lint -d postgres file.yaml`

### Dialect

Postgres

### Version

```
sqlfluff, version 1.4.5
```

### Configuration

default

### Code of Conduct

jmpfar:

Fixes bugs and adding new paths to the `CREATE INDEX` statement in Postgres:

* Changed code to use `optional` matchers rather than `AnyNumberOf`. This prevents matching incorrect syntax.
* Fix a bug where operator classes parameters were not matched properly
* Fix a bug where the `NULLS` keyword was matched by the Opclass name.
  Postgres implements `NULLS` `FIRST`/`LAST` matching using one token lookahead. Worked around this issue by 
 disallowing `NULLS` in Opclass name.
* Added `NULLS DISTINCT` / `NULLS NOT DISTINCT` syntax handling
* Added `TABLESPACE` syntax handling
* Added `WHERE` syntax handling

For reference see https://www.postgresql.org/docs/current/sql-createindex.html

fixes #4355
Also related to #3557 

I added an explicit exclude for the `NULLS` non-reserved keyword in the opclass name. This is different than how postgres handles it, which can reject some valid syntax.

The [relevant syntax](https://www.postgresql.org/docs/current/sql-createindex.html) looks like so:
```
[ COLLATE collation ] [ opclass [ ( opclass_parameter = value [, ... ] ) ] ] [ ASC | DESC ] [ NULLS { FIRST | LAST } ]
```

The issue is that `NULLS` is not a reserved keyword, and therefore can be used in `opclass` name. both are optional and therefore there is a precedence issue here where `NULLS` will be parsed by the `opclass` name.

Postgres handles this using a lookahead on an extra token, see https://github.com/postgres/postgres/blob/f282b026787da69d88a35404cf62f1cc21cfbb7c/src/backend/parser/parser.c#L206. 
I opted to exclude as I did not find an easy way to implement lookahead, and I thought the improvements are better than the minor lost edge cases.

This implementation choice means that the valid although weird syntax will be rejected currently: 
```
CREATE INDEX x ON y (col NULLS);
CREATE INDEX x ON y (col NULLS NULLS FIRST);
```

### Pull Request checklist

  - Other.


greg-finley:
> I added an explicit exclude for the NULLS non-reserved keyword in the opclass name. This is different than how postgres handles it, which can reject some valid syntax.

Is there any way around this? Ideally we should be able to parse as much as possible, erroring on parsing illegal syntax vs not being able to parse legal syntax.

jmpfar:
> > I added an explicit exclude for the NULLS non-reserved keyword in the opclass name. This is different than how postgres handles it, which can reject some valid syntax.
> 
> Is there any way around this? Ideally we should be able to parse as much as possible, even enabling parsing illegal syntax vs not being able to parse legal syntax.

So apparently Postgres uses a one token lookahead to handle this case (see [here](https://github.com/postgres/postgres/blob/10a082bf72150f49c34fcf152565c77d80f31a9c/src/backend/parser/gram.y#L767-L776) and [here](https://github.com/postgres/postgres/blob/f282b026787da69d88a35404cf62f1cc21cfbb7c/src/backend/parser/parser.c#L206)). Without a lookahead, the optional `NULLS`token, not being a keyword will be slurped by the opclass grammar, which is also optional and can match identifiers (including `NULLS`).

I did not find a better way of implementing this, as I couldn't find a way to add lookahead parsing, but maybe you can direct me at the right direction to fixing such an issue or have a better idea of working around it.

Btw notice, that the caveat statements mentioned above are also unparsable in the current version of sqlfluff, while this fix adds a lot of new supported syntax to `CREATE INDEX` (opclass, tablespace, nulls first/last, nulls distinct, where and rejecting of some invalid syntax)

greg-finley (approved):
Gotcha, I don't know any better answer. Thanks for the contribution! Sorry for the delayed review.

PR #4362, opened 2023-02-06
[draft] V2 Rules interface

alanmcruickshank:
This PR is the first step in #4031. It's draft for now while I work through it, but eventually will:
- [x] Include working example plugins for both the V1 and V2 interface.
  - [x] Rename the existing rules plugin to "V1".
  - [x] Add a new plugin clearly labelled as "V2".
  - [x] Update references to those example plugins to make sense.
- [ ] Update the `sqlfluff rules` CLI command to display V2 rules using their names.
- [ ] Update docs with guidance on how to use `rules` and `exclude_rules`.
- [ ] Add tests for test selection using aliases, codes and groups.

This PR won't migrate existing rules into the new format - I propose we do that as a separate step.


alanmcruickshank:
Referencing the discussion via slack. I'm going to park this PR in favour of #4399. This is getting too complicated and the additional complexity of breaking the 1-2-1 relationship between rules and crawlers is too much.

PR #4372, opened 2023-02-08
Add ruff linter

greg-finley:

Closes https://github.com/sqlfluff/sqlfluff/issues/4334

This linter is gaining tons of traction with other open-source projects. It's also extremely fast: https://github.com/charliermarsh/ruff#ruff

No

### Pull Request checklist

  - Other.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,25 +5,25 @@ repos:
       # If adding any exceptions here, make sure to add them to .editorconfig as well
       - id: end-of-file-fixer
         exclude: |
-            (?x)^(
-              test/fixtures/linter/sqlfluffignore/|
-              test/fixtures/config/inheritance_b/example.sql|
-              test/fixtures/config/inheritance_b/nested/example.sql|
-              plugins/sqlfluff-templater-dbt/test/fixtures/dbt/templated_output/trailing_newlines.sql|
-              plugins/sqlfluff-templater-dbt/test/fixtures/dbt/dbt_project/models/my_new_project/trailing_newlines.sql
-            )$
+          (?x)^(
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -33,35 +33,35 @@ repos:
...
-          utf8,
-        ]
+        args: [--file-encoding, utf8]
         files: docs/source/.*\.rst$
   - repo: https://github.com/adrienverge/yamllint.git
     rev: v1.26.3
     hooks:
       - id: yamllint
         args: [-c=.yamllint]
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
```
barrywhart, line 63: The README for ruff advertises:
> [Near-parity](https://github.com/charliermarsh/ruff#how-does-ruff-compare-to-flake8) with the built-in Flake8 rule set

In light of that, should we remove flake8 (line 50 above)? It would be odd to run two highly overlapping style checker tools.

If we consider doing this, we should make sure something is still enforcing/running Black at commit time. You'll see references to flake8 and Black in this file, also in `requirements_dev.txt`, possibly other places.

greg-finley: For now I was thinking just having them to run together, since Ruff runs nearly instantly, and it doesn't introduce thrash (they don't try to change things back and forth). What do you think?

barrywhart: There's no harm to adding it, so I'm happy with it. Just pointing out it may not add a lot of value because of the overlap. I like trying new tools, though! In fact, I had already run ruff on one of my projects -- can't remember which one.

barrywhart (approved):

Issue #4349, opened 2023-02-04
Support for arrays in TOML configuration


alistairewj:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

Currently configuration for a few settings requires a string with comma delimited values, e.g.

```toml
[tool.sqlfluff.core]
rules = "L010, L035"
```

toml files also support arrays

```toml
[tool.sqlfluff.core]
rules = ["L010", "L035"]
```

However this results in an error:

```python
<snip>
    return [s.strip() for s in raw_str.split(",") if s.strip()]
AttributeError: 'list' object has no attribute 'split'
```

Presumably as the attribute is always assumed to be a string: https://github.com/sqlfluff/sqlfluff/blob/f3762975a655166a3fac95c8ef757adc30423741/src/sqlfluff/core/config.py#L672-L684

I'm on v1.4.5 but it doesn't seem to be different on 2.x.x. Happy to make a PR if you can suggest how to account for the non-TOML config file types, which I don't routinely use.

### Use case

It would be useful to specify these attributes using arrays. Would allow for nicer formatting of the config file, which seems to be something people might care about around here :)

### Dialect

N/A

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

PR #4387, opened 2023-02-13
Add support for arrays in TOML configuration

AidanHarveyNelson:
fixes https://github.com/sqlfluff/sqlfluff/issues/4349

Added a check in the `FluffConfig` class to validate if the config is a str before trying to split the string.

Unsure if this should be implemented here or change the Toml loader function to change all lists into a concatenated string. Based on the previous test case checking for a list in the Toml config I think this is the suitable implementation of it.

If advise that this should reside in the toml loader then I can update the code to complete that instead.

No

### Pull Request checklist

  - Other.


src/sqlfluff/core/config.py
```diff
@@ -755,9 +755,13 @@ def __init__(
...
-                self._configs["core"][out_key] = _split_comma_separated_string(
-                    self._configs["core"][in_key]
-                )
+                # Checking if key is string as can potentially be a list to
+                if isinstance(self._configs["core"][in_key], str):
+                    self._configs["core"][out_key] = _split_comma_separated_string(
+                        self._configs["core"][in_key]
+                    )
+                else:
+                    self._configs["core"][out_key] = self._configs["core"][in_key]
```
pdebelak, line 764: Should this also check that the resulting config value is a list and raise an error otherwise?

Also, this makes the config for these values inconsistent with certain rule configs like `ignore_words` for rule L010. I wonder if this logic should get moved into `_split_comma_separated_string` and to make `BaseRule.split_comma_separate_string` call that as well so they share the logic.

Something like this maybe?

```python
def _split_comma_separated_string(raw: Union[str, List[str]]) -> List[str]:
    if isinstance(raw, str):
        return [s.strip() for s in raw.split(",") if s.strip()]
    if isinstance(raw, list):
        return raw
    raise SQLFluffUserError(f"Expected list or comma separated string. Got {type(raw)} instead for value {raw}.")
```

AidanHarveyNelson:
Apologies for the delay in updating this. I have taken the feedback and updated the function and used it for BaseRule as well

test/core/config_test.py
```diff
@@ -144,6 +144,7 @@ def test__config__load_toml():
             "testing_bar": 7.698,
             "testing_bool": False,
             "testing_arr": ["a", "b", "c"],
+            "rules": ["L010", "L035"],
```
WittierDinosaur, line 147: Can you update these to the new codes?

AidanHarveyNelson: @WittierDinosaur I have updated the codes to `LT03` and `LT09`. I believe this is what you meant but please clarify if this is not the correct change.

alanmcruickshank (approved):
LGTM. Thanks for this.

Issue #4211, opened 2022-12-27
Parsing ERROR while using nested arrays


VaKVa:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Using nested arrays leads to an error

### Expected Behaviour

No errors while using nested arrays

### Observed Behaviour

While running "sqlfluff fix...":
![image](https://user-images.githubusercontent.com/78371300/209633696-ff8b2417-c813-4ec6-9211-cf55e9b5ac86.png)

While running "sqlfluff lint...":
![image](https://user-images.githubusercontent.com/78371300/209633782-3fb1d321-e73f-44b2-8234-639a980ead25.png)

While running "sqlfluff parse...":
![image](https://user-images.githubusercontent.com/78371300/209634023-1cc59956-c586-4623-a54f-c55fe41ed666.png)

### How to reproduce

SELECT pn[1].qqq AS pl
FROM sks

### Dialect

Athena

### Version

sqlfluff: version 1.3.2

### Configuration

[sqlfluff]
dialect = athena
exclude_rules = L009, L031, L033, L025, L034, L022, L023
templater = dbt
sql_file_exts = .sql,.sql.j2,.dml,.ddl
runaway_limit = 15
large_file_skip_byte_limit = 26000

[sqlfluff:rules]
tab_space_size = 2
max_line_length = 999
indent_unit = space
comma_style = leading
allow_scalar = True
single_table_references = consistent
unquoted_identifiers_policy = all

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

Issue #4193, opened 2022-12-21
PRS error in ARRAY(ROW(id varchar)) construnction


VaKVa:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Unparsable section in such construction: THEN ARRAY[CAST(ROW(ARRAY[CAST(ROW('') AS ROW(id varchar))], CAST(ROW('') AS ROW(id varchar)), 'Approved') AS ROW(approvers ARRAY(ROW(id varchar)), performer ROW(id varchar), approvalStatus varchar))] 

### Expected Behaviour

No parsing errors

### Observed Behaviour

![image](https://user-images.githubusercontent.com/78371300/208962333-967ed78a-76b6-4b82-811f-03357f87f9aa.png)

### How to reproduce

SELECT 
CAST(ROW(ARRAY[CAST(ROW('') AS ROW(id varchar))], CAST(ROW('') AS ROW(id varchar)), 'Approved') AS ROW(approvers ARRAY(ROW(id varchar)), performer ROW(id varchar), approvalStatus varchar)) as test

### Dialect

Athena

### Version

sqlfluff: version 1.3.2

### Configuration

[sqlfluff]
dialect = athena
exclude_rules = L009, L031, L033, L025, L034, L022, L023
templater = dbt
sql_file_exts = .sql,.sql.j2,.dml,.ddl
runaway_limit = 15
large_file_skip_byte_limit = 26000

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

Issue #4187, opened 2022-12-20
Athena: PRS error with element_at(), while trying to get subelement from the structure


VaKVa:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

In Athena parsing, the error happens while trying to get subelement from the structure.

### Expected Behaviour

No parsing errors

### Observed Behaviour

--1st part:
                     COALESCE(
                        element_at(rq.hiring_managers, 1), 
                        element_at(rq.hiring_managers, 2), 
                        rq.creator_id), 
-- 2nd part:    COALESCE(
                        element_at(pl.hiring_managers, 1).id, 
                        element_at(pl.hiring_managers, 2).id)
                )

The 1st part is work perfectly, while the 2nd (when we get .id from the pl.hiring_managers) is throwing the parsing error

### How to reproduce

                     COALESCE(
                        element_at(rq.hiring_managers, 1), 
                        element_at(rq.hiring_managers, 2), 
                        rq.creator_id), 
                     COALESCE(
                        element_at(pl.hiring_managers, 1).id, 
                        element_at(pl.hiring_managers, 2).id)
                       )

### Dialect

Athena

### Version

sqlfluff: version 1.3.2

### Configuration

[sqlfluff]
dialect = athena
exclude_rules = L009, L031, L033, L025, L034, L022, L023
templater = dbt
sql_file_exts = .sql,.sql.j2,.dml,.ddl
runaway_limit = 15
large_file_skip_byte_limit = 26000

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

PR #4391, opened 2023-02-15
fix(athena): resolve errors parsing around maps, structs, and arrays

timcosta:
### Fixes various PRS errors and invalid types
Specifically fixes:
* #4211 
* #4193
    * Mostly, it still has a syntax error where `ARRAY[ROW(id varchar)]` should be `ARRAY<ROW(id varchar)>` due to usage of array as a type and not a function
* #4187 

Test cases were added for the reproductions provided in the above issues, and additional test cases were added from various pages on the Athena documentation site.

Added support for prepared statements.

Yes. This will be a slightly breaking change in some scenarios. None of the existing tests were broken, however:
* Types that athena doesnt recognize were checked and removed
* `location` no longer regex matches the string for an s3 uri because that should be a linting rule, not a parsing failure
* `double`s no longer support precision, because thats not a real thing
* `decimal` rules changed slightly to reflect that argument 2 is optional but argument 1 is not
* `uniontype` was removed because its not a real thing
* `?` was added as a segment to the ANSI dialect due to its presence in [a_expr](https://www.cockroachlabs.com/docs/stable/sql-grammar.html#a_expr) but it is not used in the ANSI dialect yet 

### Pull Request checklist

  - Other.


amiraliucsc (approved):
LGTM

amiraliucsc (approved):
LGTM

WittierDinosaur (approved):
LGTM

WittierDinosaur (approved):
LGTM

PR #4399, opened 2023-02-16
Rule names, aliases and more complicated selection.

### LINKED ISSUE #4031: Reorganize rule names

alanmcruickshank:
This sets the ground work for #4031.

It doesn't actually do the reorg - but it builds all the capability required for the reorg by adding selection by name, groups and aliasing. It also doesn't add the _use_ of names to the CLI output, either in `sqlfluff rules` or in the linting output. I'll do that as the next step.

I've chosen to scale back the ambition from what I started in #4362 - because that was getting _far_ to complicated.

This will instead:
- [x] Tidy up the existing rules plugin example.
- [x] Allow selection of rules by `name`, `alias`, `group` and `code`. Including test cases for this.
- [x] Improve documentation on how to use `rules` and `exclude_rules`.


src/sqlfluff/core/rules/base.py
```diff
@@ -1115,76 +1118,134 @@ def get_rulelist(self, config) -> List[BaseRule]:
...
             # Allow variable substitution in making the description
-            kwargs["description"] = self._register[k]["description"].format(**kwargs)
-            rule_kwargs[k] = kwargs
+            kwargs["description"] = self._register[code].description.format(**kwargs)
+            # Instantiate when ready
+            instantiated_rules.append(rule_class(**kwargs))
 
-        # Instantiate in the final step
-        return [self._register[k]["cls"](**rule_kwargs[k]) for k in keylist]
+        return instantiated_rules
```
pwildenhain, line 1248: I like the look of this better 😅

pwildenhain (approved):
I like that you limited the scope of this PR, definitely made it a lot easier to follow

I didn't even think about how name collisions between aliases/groups/names would even be a thing, I like your solution for handling it 

Great work Alan 🙌

PR #4400, opened 2023-02-17
Add rule names to CLI outputs

alanmcruickshank:
This sets completes the CLI elements of #4031. It builds on the work done in #4399 (and I'm assuming we'll agree and merge that before moving on to this). I'll keep this PR in draft until that one is done, but I thought I'd open in early so people can feedback on the output (and also so I can make sure that tests pass 😺 ).

Specifically this PR adds out two CLI improvements to support named rules. To allow testing of those, it adds names, groups, and aliases (in the form of their future new code for now) - although I anticipate that we will do the actual recoding as a final separate PR.

## `sqlfluff rules`

Firstly this adds names, groups and aliases (where present) to the `sqlfluff rules` command. The new output looks like this:

image

**NOTE**: The codes are still the existing codes rather than the proposed codes (as I've said above), but see that the new proposed codes are currently in as aliases.

This makes the CLI output much more verbose, but I thought it a good idea to be explicit on names, codes, groups and aliases in this CLI command to promote their use.

## linting output

Secondly it adds the rule name to the linting output. The new output looks like this:

image

**NOTE**: I haven't added a name for L065 yet (so it shows no name - and that would also be true for a rule plugin which doesn't yet define a name), and for L001 there is a name, but it's only on a new line because there isn't space on the existing line. If there _is_ space, then it should be on the same line.

Rather than trying to build the name into a [table format as suggested in the original issue](https://github.com/sqlfluff/sqlfluff/issues/4031#issuecomment-1315482188), this simply add the rule name to the end of the output which is closer to the pylint format. The main reason for that is that the current output conforms to an 80 character line width limit. If we try and fit another column within that 80 char limit there really isn't much space left for the description itself.


pwildenhain (approved):
We're getting so close! Great work 👏🏻👏🏻

PR #4409, opened 2023-02-19
SQLite Refactor to reduce statement and keyword scope

### LINKED ISSUE #4398: Incorrect identification of sqlite keywords
### Search before asking

### What Happened

```sql
CREATE TABLE users (
    user_id INTEGER PRIMARY KEY AUTOINCREMENT,
    username TEXT NOT NULL UNIQUE,
    password TEXT NOT NULL,
    email TEXT NOT NULL UNIQUE
);
```

This is returning "L029 Keywords should not be used as identifiers" in line 4 pos 5 (the `password` name). This is valid sqlite, and isn't listed as a keyword here: https://www.sqlite.org/lang_keywords.html. Removing the "UNIQUE" on the previous line (to check if it was incorrectly pointing there instead) changed nothing. Playing around in the online generator shows that it specifically doesn't like "password" as a column name. 

### Expected Behaviour

No errors

### Observed Behaviour

L029 | 4 / 5 | Keywords should not be used as identifiers.

### How to reproduce

```sql
CREATE TABLE users (
    user_id INTEGER PRIMARY KEY AUTOINCREMENT,
    username TEXT NOT NULL,
    password TEXT NOT NULL,
    email TEXT NOT NULL UNIQUE
);
```

### Dialect

sqlite

### Version

online.sqlfluff.com

### Configuration

na

### Code of Conduct

WittierDinosaur:
This PR reduces the scope of sqlite's dialect coverage. SQLite doesn't actually support very much, I've restricted the statements to only real sqlite statements, and removed keywords in kind. 

Fixes #4398

No. While I can't guarantee zero regressions, there are none in the code base. It should having the added side effect of improving parsing speed for sqlite.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -568,6 +568,7 @@ class CompositeComparisonOperatorSegment(BaseSegment):
     ),
     # This is a placeholder for other dialects.
     SimpleArrayTypeGrammar=Nothing(),
+    AutoIncrementGrammar=Sequence("AUTO_INCREMENT"),
```
tunetheweb, line 571: I'm not sure any other dialect supports this other than MySQL so wonder if we should just remove this and move it to MySQL? Or set this to Nothing in ANSI (and remove from other keyword files)? Other dialects do have a similar concept (`AUTOINCREMENT`/`IDENTITY` in Snowflake, `SERIAL` in Posgres, nothing in BigQuery)

tunetheweb (approved):
Approved with one suggestion.

src/sqlfluff/dialects/dialect_sqlite.py
```diff
@@ -149,3 +431,42 @@ class TableConstraintSegment(ansi.TableConstraintSegment):
...
+        Ref("DropTableStatementSegment"),
+        Ref("DropTriggerStatementSegment"),
+        Ref("DropViewStatementSegment"),
+        Ref("ExplainStatementSegment"),
+        Ref("InsertStatementSegment"),
+        Ref("SelectableGrammar"),
+        Ref("TransactionStatementSegment"),
+        Ref("UpdateStatementSegment"),
+        Bracketed(Ref("StatementSegment")),
+    )
```
alanmcruickshank, line 472: For this grammar, because most is a copy of the root one, but with some things removed - you could use the `.copy()` method to avoid repeating all the duplicated ones here, and just remove the ones you don't need.

Something like:
```python
parse_grammar = ansi.StatementSegment.parse_grammar.copy(remove=...)
```

Given just how _many_ you're removing - I can see that it might actually be _less_ duplication to keep things how you've done it already - but worth considering.

src/sqlfluff/dialects/dialect_sqlite.py
```diff
@@ -29,9 +41,240 @@
...
+        Sequence("UNION", OneOf("DISTINCT", "ALL", optional=True)),
+        Sequence(
+            OneOf(
+                "INTERSECT",
+                "EXCEPT",
+            ),
+            Ref.keyword("ALL", optional=True),
+        ),
+        exclude=Sequence("EXCEPT", Bracketed(Anything())),
+    )
```
alanmcruickshank, line 241: This is one is a good candidate for `.copy(remove=...)` too. I think you're just removing `MINUS` right?

src/sqlfluff/dialects/dialect_sqlite.py
```diff
@@ -29,9 +41,240 @@
...
         "PRIMARY", "KEY", Sequence("AUTOINCREMENT", optional=True)
     ),
+    TemporaryTransientGrammar=Ref("TemporaryGrammar"),
+    DateTimeLiteralGrammar=Sequence(
+        OneOf("DATE", "DATETIME"),
+        TypedParser(
+            "single_quote", ansi.LiteralSegment, type="date_constructor_literal"
+        ),
+    ),
+    BaseExpressionElementGrammar=OneOf(
```
alanmcruickshank, line 51: Ditto this one.

alanmcruickshank (commented):
This definitely works as it is - but you might be able to make this more concise by using the `grammar.copy(remove=...)` method.

It's not particularly subtle so definitely won't work with all of what you've done here - but it might make what you've done a bit more DRY in some of the cases.

src/sqlfluff/dialects/dialect_sqlite.py
WittierDinosaur: I agree in principle, but like you've said, I'm removing many. This also affords sqlite some extra regression protection by stopping it getting anything we may add to ansi in the future. It's very very lightweight compared to every other dialect, which is why I think being explicit is cleaner.

src/sqlfluff/dialects/dialect_ansi.py
WittierDinosaur: Agree - will spin that up in a separate PR. Gonna merge as-is for now as it's blocking a couple of tickets I want to clear out

PR #4414, opened 2023-02-20
Enable noqa using aliases and groups

alanmcruickshank:
This is part of #4031 and builds on #4399 .

The original PR added the option to use aliases and groups in rule selection, this extends that functionality to the `--noqa` option.

To do that involved a bit more than expected, because we need to be able to pass around an accurate `reference_map` (so allow any references to be interpreted). Rule lists are generated in the main thread and then passed around, complicating the issue.

This PR introduces a `RulePack` class which replaces all the locations where previously we would use a list of instantiated rules (`List[BaseRule]`). This contains the same list of rules, but also the reference map used in creation. This means any time we have a list of rules, we have a map of references to use in any other selection locations.

### DIFF
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -262,7 +262,7 @@ class Linter:
     def parse_noqa(
         comment: str,
         line_no: int,
-        reference_map: Dict[str, List[str]],
+        reference_map: Dict[str, Set[str]],
     ):
         """Extract ignore mask entries from a comment string."""
         # Also trim any whitespace afterward
@@ -310,7 +310,7 @@ class Linter:
                         unexpanded_rules = tuple(
                             r.strip() for r in rule_part.split(",")
                         )
-                        expanded_rules = []
+                        expanded_rules: List[str] = []
                         for r in unexpanded_rules:
                             matched = False
                             for expanded in (
@@ -423,7 +423,7 @@ class Linter:
     def extract_ignore_from_comment(
         cls,
         comment: RawSegment,
-        reference_map: Dict[str, List[str]],
+        reference_map: Dict[str, Set[str]],
     ):
         """Extract ignore mask entries from a comment segment."""
         # Also trim any whitespace afterward
@@ -438,7 +438,7 @@ class Linter:
     def extract_ignore_mask_tree(
         cls,
         tree: BaseSegment,
-        reference_map: Dict[str, List[str]],
+        reference_map: Dict[str, Set[str]],
     ) -> Tuple[List[NoQaDirective], List[SQLBaseError]]:
         """Look for inline ignore comments and return NoQaDirectives."""
         ignore_buff: List[NoQaDirective] = []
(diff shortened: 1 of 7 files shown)


src/sqlfluff/core/linter/linter.py
```diff
@@ -261,7 +262,7 @@ def _parse_tokens(
     def parse_noqa(
         comment: str,
         line_no: int,
-        rule_codes: List[str],
+        reference_map: Dict[str, List[str]],
```
barrywhart, line 265: Nit: IIUC, this function only accesses the _keys_ of the dictionary. Should we consider this instead?

```
rule_codes: typing.KeysView
```

If we do this, I don't think the function itself requires any changes.

src/sqlfluff/core/rules/base.py
```diff
@@ -955,6 +956,26 @@ class RuleManifest:
...
+    This contains a set of rules, post filtering but also contains the mapping
+    required to interpret any noqa messages found in files.
+
+    The reason for this object is that rules are filtered and instantiated
+    into this pack in the main process when running in multi-processing mode so
+    that user defined rules can be used without reference issues.
+    """
+
+    rules: List[BaseRule]
+    reference_map: Dict[str, List[str]]
```
barrywhart, line 972: Can you add a comment and/or example of the contents of this dictionary? It would make the code a lot more understandable. As written, I'm unsure:
* Whether the rule code is a key or a value
* Why multiple values are supported

Is ordering important for the values (i.e. the use of `List` rather than `Set`)?

src/sqlfluff/core/rules/base.py
```diff
@@ -1177,7 +1189,27 @@ def get_rulelist(self, config) -> List[BaseRule]:
...
+        return {**alias_map, **reference_map}
+
+    def get_rulepack(self, config) -> RulePack:
+        """Use the config to return the appropriate rules.
+
+        We use the config both for allowlisting and denylisting, but also
+        for configuring the rules given the given config.
+
+        Returns:
+            :obj:`list` of instantiated :obj:`BaseRule`.
```
barrywhart, line 1201: Needs updating (or removing)... now returns `RulePack`

alanmcruickshank: I'll just remove it now - I think the type hints are enough.

src/sqlfluff/core/rules/base.py
alanmcruickshank: Ah good point. I think a `Set` would be more appropriate. I'll add some docs to explain exactly what it is.

src/sqlfluff/core/linter/linter.py
alanmcruickshank: Unfortunately we _also_ need the values of the map to be able to extract interpret any group references. This mapping is used to turn `--noqa:my_group` into `--noqa:L043,L088` (if `reference_map` is `{"my_group":{"L043", "L088"}}`).

barrywhart: Oh cool. I _thought_ it must be using the values somewhere, but I overlooked it during review.

barrywhart (approved):

PR #4415, opened 2023-02-20
Deprecate doc decorators (replace with metaclass)

alanmcruickshank:
This is part of #4031.

This deprecates the `doc_decorators` for rules, and instead builds the functionality directly into the `BaseRule` using a metaclass (similar to what we did for the segment metaclass). This reduces lines of code by eliminating the necessity to add multiple decorators to all rules.

In the process, this also adds `aliases` & `name` to the documented values (along with config values, fix capability and groups). All of these are triggered directly by the presence of the relevant config values in the rule attributes. The only _additional_ value is to add an `is_fix_compatible` attribute on the `BaseRule` which can be overridden as a flag for whether the rule is fixable.

I think this solution is much more robust in many ways than the original implementation (the docstrings for the original admit that too!). This will have merge conflicts with some of my other open PRs, but I'd suggest that this one goes first.

### DIFF
--- a/plugins/sqlfluff-plugin-example/src/example/rules.py
+++ b/plugins/sqlfluff-plugin-example/src/example/rules.py
@@ -10,11 +10,6 @@ from sqlfluff.core.rules import (
     RuleContext,
 )
 from sqlfluff.core.rules.crawlers import SegmentSeekerCrawler
-from sqlfluff.core.rules.doc_decorators import (
-    document_configuration,
-    document_fix_compatible,
-    document_groups,
-)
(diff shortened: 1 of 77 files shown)


alanmcruickshank:
> This looks good.
> 
> I think the best "test" of this code would be to build the docs before and after these changes, then run a "before and after" diff to verify they are identical. Can you (or have you) done that?

I've manually tested the docs build yes - unfortunately they're no the _same_ in this example because I'm also adding auto-documentation for groups and for aliases. The new docs look like this:

image

PR #4417, opened 2023-02-21
Postgres: AS MATERIALIZED support

saintamh:

Make the `NOT` optional in expressions like `WITH name AS NOT MATERIALIZED (...)`.

See https://www.postgresql.org/docs/15/queries-with.html#id-1.5.6.12.7

No

### Pull Request checklist

  - Other.


alanmcruickshank (approved):
Looks sensible 👍 . Assuming the tests pass, I think this is good.

I'm reviewing before the CI suite has completed, so if that fails we might still need some rework.

Thanks for contributing this. 🎉

PR #4420, opened 2023-02-21
Handle untaken positive indents with taken negative pair.

### LINKED ISSUE #4234: 2.0.0a1 Multiline `SELECT` are not correctly indented in CTA
### Search before asking

### What Happened

Consider the following two examples:

```sql
CREATE TABLE mytable
AS
(SELECT
    id,
    user_id
FROM another_table
)
;
```

This one ☝️ does not throw any lint errors which is wrong I think. The `SELECT` should be on a new, indented line.

Compare this to below example that uses only a single SELECT target.

```sql
CREATE TABLE mytable
AS
(SELECT id
FROM table
)
;
```

which gets linted to

```sql
CREATE TABLE mytable
AS
(
    SELECT id
    FROM table
)
;
```

### Expected Behaviour

see above

### Observed Behaviour

see above

### How to reproduce

see above

### Dialect

ansi

### Version

2.0.0a1

### Configuration

default

### Code of Conduct

### LINKED ISSUE #4328: 2.0.0a4 L003 line break within datatype
### Search before asking

### What Happened

I tried to format the following tsql statement:
```sql
DECLARE @SqlTruncate nvarchar(400) = IIF(@myvar = 123 OR @myvar = 2
, 'yes'
, 'no'
)

```

### Expected Behaviour

I expected the statement to stay as is.

### Observed Behaviour

In fact the SQL was formatted the following way:
```sql
DECLARE @SqlTruncate nvarchar(
  400) = IIF(@myvar = 123 OR @myvar = 2
, 'yes'
, 'no'
)
```

### How to reproduce

```
PS > cat .\q1.sql
DECLARE @SqlTruncate nvarchar(400) = IIF(@myvar = 123 OR @myvar = 2
, 'yes'
, 'no'
)
PS > sqlfluff.exe fix --dialect tsql .\q1.sql
== [q1.sql] FAIL
L:   1 | P:  31 | L003 | Expected line break and indent of 2 spaces before '400'.
L:   2 | P:   1 | L003 | Expected indent of 2 spaces.
L:   3 | P:   1 | L003 | Expected indent of 2 spaces.
==== fixing violations ====
Attempting fixes...
Persisting Changes...
== [q1.sql] PASS
All Finished 📜 🎉!
PS C:\Users\fmoeller\Desktop\RepositoryVorlage>
```

### Dialect

tsql

### Version

```
==== sqlfluff ====
sqlfluff:              2.0.0a4 python:                 3.11.0
implementation:        cpython verbosity:                   2
dialect:                  tsql templater:                 raw
```

### Configuration

default

### Code of Conduct

alanmcruickshank:
Fixes #4234 
Fixes #4328 

This resolves #4234. Implementation as suggested in that issue.

This introduces the idea of an `imbalanced` indent, i.e. one where the positive indent is _untaken_ but the negative one isn't. When they're found, this keeps track of them at the mapping phase so they can be forced when we do the linting.

One challenge which this one is that it requires somewhat special handling for bracketed expressions. As per discussion on slack, I think this is ok - and consistent with major style guides. There's a slight collision here with the `implicit` indent functionality so I've added another test case to explicitly handle that scenario.

### DIFF
--- a/src/sqlfluff/utils/reflow/reindent.py
+++ b/src/sqlfluff/utils/reflow/reindent.py
@@ -533,6 +533,10 @@ def _crawl_indent_points(
     cached_point: Optional[_IndentPoint] = None
     for idx, elem in enumerate(elements):
         if isinstance(elem, ReflowPoint):
+            # NOTE: The following line should never lead to an index error
+            # because files should always have a trailing IndentBlock containing
+            # an "end_of_file" marker, and so the final IndentPoint should always
+            # have _something_ after it.
             following_class_types = elements[idx + 1].class_types
             indent_stats = IndentStats.from_combination(
                 cached_indent_stats,
@@ -662,7 +666,15 @@ def _map_line_buffers(
     # Buffers to keep track of indents which are untaken on the way
     # up but taken on the way down. We track them explicitly so we
     # can force them later.
+
+    #: dict of ints: maps indentation balance values to the last
+    #: index location where they were seen. This is a working buffer
+    #: and not directly returned by the function.
     untaken_indent_locs = {}
+    #: list of ints: a list of element indices which contain untaken
+    #: positive indents, that should be forced later because their
+    #: corresponding negative indent _was_ taken. Several edge cases
+    #: are excluded from this list and so not included. See code below.
     imbalanced_locs = []
 
     for indent_point in _crawl_indent_points(
@@ -730,7 +742,10 @@ def _map_line_buffers(
                     # First check for bracket special case. It's less about whether
                     # the section _ends_ with a lone bracket, and more about whether
                     # the _starting point_ is a bracket which closes a line. If it
-                    # is, then skip this location. (Special case 2)
+                    # is, then skip this location. (Special case 2).
+                    # NOTE: We can safely "look ahead" here because we know all files
+                    # end with an IndentBlock, and we know here that `loc` refers to
+                    # an IndentPoint.
                     if "start_bracket" in elements[loc + 1].class_types:
                         continue
 
--- a/src/sqlfluff/utils/reflow/reindent.py
+++ b/src/sqlfluff/utils/reflow/reindent.py
@@ -681,7 +681,7 @@ def _map_line_buffers(
         # Bracketed expressions are a bit odd here.
         # e.g.
         #   WHERE (
-        #       foo = bar 
+        #       foo = bar
         #   )
         #   LIMIT 1
         #
@@ -690,13 +690,10 @@ def _map_line_buffers(
         # one there even though there _is_ a line break after the closing
         # bracket.
         following_class_types = elements[indent_point.idx + 1].class_types
-        preceding_class_types = elements[indent_point.idx - 1].class_types
         if (
             indent_point.indent_trough
-            # End of file ends case
+            # End of file ends case. (Special case 1)
             and "end_of_file" not in following_class_types
-            # Bracket special case.
-            and "end_bracket" not in preceding_class_types
         ):
             passing_indents = list(
                 range(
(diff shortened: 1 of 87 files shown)


src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -533,8 +533,10 @@ def _crawl_indent_points(
     cached_point: Optional[_IndentPoint] = None
     for idx, elem in enumerate(elements):
         if isinstance(elem, ReflowPoint):
+            following_class_types = elements[idx + 1].class_types
```
barrywhart, line 536: Is there a possibility this will cause an `IndexError`?

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -641,11 +643,28 @@ def _crawl_indent_points(
...
+) -> Tuple[List[_IndentLine], List[int]]:
+    """Map the existing elements, building up a list of _IndentLine.
+
+    Returns:
+        :obj:`tuple` of a :obj:`list` of :obj:`_IndentLine` and a
+            :obj:`list` of :obj:`int`. The first is the main output
+            and is designed to be used in assessing indents and
+            their effect through a SQL file. The latter is a list of
+            "imbalanced" indent locations, where the positive indent
+            is untaken, but it's corresponding negative indent *is*
```
barrywhart, line 655:
```suggestion
            is untaken, but its corresponding negative indent *is*
```

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -641,11 +643,28 @@ def _crawl_indent_points(
...
+            taken.
+
+    """
     # First build up the buffer of lines.
     lines = []
     point_buffer = []
+    # Buffers to keep track of indents which are untaken on the way
+    # up but taken on the way down. We track them explicitly so we
+    # can force them later.
+    untaken_indent_locs = {}
```
barrywhart, line 665: Maybe consider documenting what are the keys/values in these two variables, not just types but what they _mean_. (E.g. if it's just an integer, is that an index into some list, or a character position...).

barrywhart:
Does this PR have any impact on these other issues?
- #4328 
- #4234 

I think they are somewhat related in the general sense of being indent/dedent issues in the new reflow code.

I had started looking at #4328 a few days ago, but since you're in this area of the code anyway, if it makes sense to look at that one as well, feel free. 🙏

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -654,18 +673,87 @@ def _map_line_buffers(
         point_buffer.append(indent_point)
 
         if not indent_point.is_line_break:
+            # If it's not a line break, we should still check whether it's
```
barrywhart, line 676: See my question about #4234 and #4328. The latter issue, at least, seems like it may involve the same logic (line break / dedent interaction).

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -654,18 +673,87 @@ def _map_line_buffers(
...
+                    if i not in untaken_indent_locs:
+                        continue
+
+                    loc = untaken_indent_locs[i]
+
+                    # First check for bracket special case. It's less about whether
+                    # the section _ends_ with a lone bracket, and more about whether
+                    # the _starting point_ is a bracket which closes a line. If it
+                    # is, then skip this location. (Special case 2)
+                    if "start_bracket" in elements[loc + 1].class_types:
```
barrywhart, line 734: Possible `IndexError`?

barrywhart (changes requested):
Suggested a punctuation fix in a comment. Otherwise a few minor questions. Looks good overall.

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: Not here - there's always a trailing `IndentBlock`. I'll add a comment to that effect...

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: Yeah - good idea. The naming isn't super clear here.

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: yep - agreed. Thanks for updating the linked issues 👍

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: Same as above - we know that files end with a block, and we're in a point so I think we're safe. I'll add a note.

barrywhart (approved):
Looks good!

PR #4423, opened 2023-02-23
Handle long lines without trailing newlines gracefully (#4386)

### LINKED ISSUE #4386: 2.0.0a4 - L016 doesn't catch long lines which end with an end_of_file before newline.
### Search before asking

### What Happened

Testing L016 on sample files - L016 doesn't trigger on lines which end with `end_of_file` before a newline (i.e. the last line of a file without a trailing newline).

This isn't a critical bug - and only affects users who don't have the trailing newline rule enabled. If that rule is enabled, then this will be fixed and then caught.

### Expected Behaviour

L016 should trigger on all lines - even if they're the final one.

### Observed Behaviour

Last line skipped

### How to reproduce

see above.

### Dialect

all

### Version

2.0.0a4

### Configuration

default

### Code of Conduct

alanmcruickshank:
This resolves #4386 . Slight tweak to the long line logic so that we do still process the final line of files where the final line has no trailing newline.

### DIFF
--- a/src/sqlfluff/utils/reflow/reindent.py
+++ b/src/sqlfluff/utils/reflow/reindent.py
@@ -1419,9 +1419,23 @@ def lint_line_length(
         # Are there newlines in the element?
         # If not, add it to the buffer and wait to evaluate the line.
         # If yes, it's time to evaluate the line.
-        if not isinstance(elem, ReflowPoint) or not has_untemplated_newline(
-            cast(ReflowPoint, elem)
+
+        if isinstance(elem, ReflowPoint) and (
+            # Is it the end of the file?
+            # NOTE: Here, we're actually looking to see whether we're
+            # currently on the _point before the end of the file_ rather
+            # than actually on the final block. This is important because
+            # the following code assumes we're on a point and not a block.
+            # We're safe from indexing errors if we're on a point, because
+            # we know there's always a trailing block.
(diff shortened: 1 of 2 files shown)


src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -1419,9 +1419,23 @@ def lint_line_length(
...
+            # than actually on the final block. This is important because
+            # the following code assumes we're on a point and not a block.
+            # We're safe from indexing errors if we're on a point, because
+            # we know there's always a trailing block.
+            "end_of_file" in elements[i + 1].class_types
+            # Or is there a newline?
+            or has_untemplated_newline(cast(ReflowPoint, elem))
         ):
+            # In either case we want to process this, so carry on.
+            pass
```
barrywhart, line 1436: Nit: Since the `if` results in a `pass`, should we consider putting a `not` in front of the condition so we don't need an `else`? Makes the code simpler. Not sure if it'd make the code harder to comment clearly.

barrywhart (approved):

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: It was originally that way around, but yes I think this way is easier to explain (and understand) even if it's a bit less conventional

PR #4424, opened 2023-02-23
Proposed graceful handling of noqa by L016 (#4248)

### LINKED ISSUE #4248: `noqa` comments on long lines incorrectly moved to its own line
### Search before asking

### What Happened

Run `fix` on the following SQL using `ansi` dialect using default config:

```sql
SELECT
    col1,
    col2,
    col3
FROM
    really_really_really_really_really_really_long_schema_name.TABLE1 -- noqa: L014
```

The `noqa` is moved to the new line, where it no longer effects the line it needs to effect.

### Expected Behaviour

I think there’s a few cases:

1. if `ignore_comment_clauses` and/or `ignore_comment_lines` is set and the comment is the only reason the line length is exceeded, then to me it should not be a violation.
2. If those are not set, or it’s exceeded anyway I think this should be unfixable, as it’s not clear how the line should be split and where the `noqa` should go. The alternative is that we add `noqa` to all new lines, but that seems like it might be overuse of `noqa`.

### Observed Behaviour

SQL is incorrectly fixed and then flags on the rule that should be `noqa`d in subsequent runs.

### How to reproduce

As per above

### Dialect

All, but above example works in ANSI.

### Version

Main branch for upcoming version 2.

### Configuration

Default

### Code of Conduct

alanmcruickshank:
This is a fairly blunt (but IMHO, _safe_) treatment of `-- noqa` comments by L016. This resolves #4248.

We already detect trailing inline comments in L016 so that we can handle them differently. This explicitly handles two cases as unfixable:
1. Any comments which are, just by their own length, going to just cause another error if we move them.
2. Any `noqa` comments.

Effectively - if a line is too long with a `noqa`, then we won't touch it - we will still report the error. The code diff looks worse than it is, just because I've added some `if` statements and indented pre-existing code by another step.


WittierDinosaur (approved):
LGTM

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -1654,77 +1657,93 @@ def lint_line_length(
...
-                            elements[last_indent_idx + 1].segments[0],
-                            [
-                                comment_seg,
-                                NewlineSegment(),
-                                WhitespaceSegment(current_indent),
-                            ],
-                        )
+
+                # Sense check a few edge cases:
+                if "noqa" in line_buffer[-1].segments[-1].raw:
```
WittierDinosaur, line 1664: Approved, but is this an ideal way to check noqa? Wouldn't this also fail on SELECT * FROM myschema.mynoqatable? Would we not want to verify type as a Comment?

alanmcruickshank: ah - good thought - but we've just confirmed that it's a comment on line 1659, so I think this is safe here.

PR #4426, opened 2023-02-24
Adjust MySQL dialect to support combination of not-null, default and …

FabianScheidt:
MySQL allows to automatically initialize and update datetime columns with the current date and time. I noticed that SQLFluff fails to parse these statements if they include all three of `NOT NULL`, `DEFAULT` and `ON UPDATE`. I assume this happens because in the current implementation `DEFAULT` is not related to the defined default value.

 This commit mandates the `DEFAULT` keyword to appear together with the default value.

With proper SQL there should be no side effects.

### Pull Request checklist


greg-finley (approved):
Awesome, thanks!

PR #4427, opened 2023-02-25
Recode Aliasing Rules

alanmcruickshank:
This makes progress on #4031. Similar to #4413.

It migrates _alias_ related rules, as another bundle which can be migrated as a relatively cohesive bunch. It follows the same logic as the #4413 .

I'm thinking that the plugin is called `aliasing`, rather than `aliases` or just `alias` - but open to discussion on that. I'm mostly going for `aliasing` because it seems to go nicely with `capitalisation`.


WittierDinosaur (approved):
LGTM

pwildenhain:
I agree that aliasing sounds better than alias for a rule group

PR #4430, opened 2023-02-25
L046 and L056 recode

alanmcruickshank:
This makes progress on #4031. Similar to #4413 and #4427.

It migrates two "loner" rules which don't really fit neatly with any other rules. One is a TSQL specific rule (`L056`, now `TQ01`) and one is our only jinja specific rule (`L046`, now `JJ01`). Eventually I would like more jinja rules, so having a place for them seems sensible. On the TSQL rule, I could imagine spinning that off to be it's own package eventually, but for backward compatibility, I've kept it bundled in SQLFluff for now - just in it's own plugin. It follows the same logic as the #4413.

This PR also alters the plugin name for the `capitalisation` plugin, as per my note in the `setup.cfg`:
```cfg
sqlfluff =
    sqlfluff = sqlfluff.core.plugin.lib
    # NOTE: We namespace the rules plugins with `rules`, because some
    # of them might later collide with other types of plugins. In particular
    # `tsql` may eventually refer to a dialect plugin and `jinja` may refer
    # to a templater plugin.
    sqlfluff_rules_capitalisation = sqlfluff.rules.capitalisation
    sqlfluff_rules_jinja = sqlfluff.rules.jinja
    sqlfluff_rules_tsql = sqlfluff.rules.tsql
```


WittierDinosaur (approved):
LGTM

PR #4432, opened 2023-02-26
Layout Rules Recode (part 1)

alanmcruickshank:
Another step in #4031. This takes the first step in the `layout` rules bundle. It _does not complete that move_, but this PR is already getting pretty big given how far some of these dependencies go. I'm going to propose this as a sensible first step before I tackle the rest.

This bundle is where I propose to merge quite a few rules together to enable _fewer_, but _more powerful and flexible_ rules. Eventually there should just be `layout.spacing`, `layout.indent`, `layout.line-length` and perhaps a couple of others.

So far, this PR:
- Merges `L001` and `L039` into one rule (`LT01` : `layout.spacing`). NOTE: This rule will likely get some more additions eventually (see below). I don't think merging these two is yet controversial, one was "trailing whitespace" and the other was "too much whitespace" - really they're both "too much whitespace".
- Merges `L002`, `L003` and `L004` into one rule (`LT02` : `layout.indent`). In particular here, by merging these rules we actually lose a little bit of functionality if people (for example) didn't want to lint the _size_ of their indents, but ones that didn't contain a mix of tabs and spaces. I don't think that's a problem per-se now the indentation algorithm is much better, but it's a potential watchout. For context: L002 was mixed indents (i.e. tabs + spaces), L003 was incorrect indent size, L004 was incorrect indent type (i.e. specifying spaces, but using tabs).
- Updates the API to return rule names in the provided dicts (and show them in github annotations)
- Updated `noqa` glob matching to use sets for deduplication.

Scope for things still to do (as per the [rules doc](https://docs.google.com/spreadsheets/d/1xGig5Eq0vTdc7d82wwTXSD4X1BUmk1zgiMUuoVPRfaU)):
 - Bring together the rest of `layout.spacing`: L005, L006, L008, L015, L017, L023, L024, L048 & L071
 - Anything with the rest of the layout rules: L007, L009, L016, L018, L019, L022, L036, L041 and L065.


src/sqlfluff/rules/layout/LT01.py
```diff
@@ -29,14 +33,20 @@ class Rule_L039(BaseRule):
...
     is_fix_compatible = True
 
     def _eval(self, context: RuleContext) -> Optional[List[LintResult]]:
         """Unnecessary whitespace."""
         sequence = ReflowSequence.from_root(context.segment, config=context.config)
-        results = sequence.respace(filter="inline").get_results()
+        results = sequence.respace().get_results()
 
         # For now, respace rules are separate for creation and reduction.
         # That shouldn't be true in future.
```
pwildenhain, line 52: Is this comment still relevant?

src/sqlfluff/rules/layout/LT02.py
```diff
@@ -0,0 +1,65 @@
...
+        To set the default tab size, set the `tab_space_size` value
+        in the appropriate configuration. To correct indents to tabs
+        use the `indent_unit` value set to `tab`.
+
+        """
+        return (
+            ReflowSequence.from_root(context.segment, context.config)
+            .reindent()
+            .get_results()
+        )
```
pwildenhain, line 65: I would call this a **HUGE** improvement over how `L003` used to look. Well done Alan 👏🏻

pwildenhain (approved):
Looking great so far, I love the simplification, really makes the package feel smoother to develop and use

src/sqlfluff/rules/layout/LT01.py
alanmcruickshank: It is _for now_. But I'd like to break that constraint in the next PR by bringing in the other spacing rules that do create whitespace when it's missing. That one could reduce the number of rules significantly.

src/sqlfluff/rules/layout/LT02.py
alanmcruickshank: lol Good :)

PR #4444, opened 2023-02-28
Structure and Ambiguous rule bundles

alanmcruickshank:
Another step in https://github.com/sqlfluff/sqlfluff/issues/4031. These ones are all straightforward recodes. All the rules are still intact and just moved into new bundles as per the [rules spreadsheet](https://docs.google.com/spreadsheets/d/1xGig5Eq0vTdc7d82wwTXSD4X1BUmk1zgiMUuoVPRfaU).

This also updates rule glob matching expressions in `rules` and `exclude_rules` to also match on aliases, names and groups - mostly because it supports better backward compatibility and it's not significantly more complicated to execute.


test/fixtures/config/glob_exclude/test.sql
```diff
@@ -2,9 +2,9 @@
 /*
 Denylist glob test
 
-This query violates L027, L044, L050, L051, and L052.
-When we exclude L05*,L027 in the config we expect L027, L050, L051,
-and L052 to be ignored by the linter.
+This query violates L027, AM04 (ex L044), L050, AM05 (ex L051), and L052.
```
WittierDinosaur, line 5: Not entirely sure we want zombie refs around where we don't need them?

alanmcruickshank: Ah - I mention them, because the globs in this test (`L04*`), do still match these rules _because_ of their aliases. It's not as much that they're there as old refs, it's because those refs persist as aliases which can be matched by globs (as demonstrated in this test).

alanmcruickshank: @WittierDinosaur - do you reckon that makes sense? I think otherwise the test is kind of confusing to understand.

WittierDinosaur: Okay, I guess so. But you don't do that in the glob_include tests

alanmcruickshank: Hmmmm good point. I'll update the comments to be more explicit on both.

WittierDinosaur (approved):
LGTM

PR #4446, opened 2023-03-01
[draft] References rule bundle

alanmcruickshank:
Another step in https://github.com/sqlfluff/sqlfluff/issues/4031. These ones are all straightforward recodes. All the rules are still intact and just moved into new bundles as per the [rules spreadsheet](https://docs.google.com/spreadsheets/d/1xGig5Eq0vTdc7d82wwTXSD4X1BUmk1zgiMUuoVPRfaU).

(currently in draft for testing. No functionality change here, just recode)


src/sqlfluff/core/default_config.cfg
```diff
@@ -281,7 +281,7 @@ require_final_semicolon = False
 # GROUP BY/ORDER BY column references
 group_by_and_order_by_style = consistent
 
-[sqlfluff:rules:L057]
+[sqlfluff:rules:RF05]
```
pwildenhain, line 284: Are there any plans to allow using the new "readable" rule names in the config rather than the codes?

pwildenhain (approved):
🌈 ✨ Nicely done

src/sqlfluff/core/default_config.cfg
alanmcruickshank: I'll slack you about this. Short answer - not at the moment because if we accept more than one way of specifying config, we then have to deal with resolving priority.

PR #4448, opened 2023-03-01
Convention rules bundle

alanmcruickshank:
yet... another... step in https://github.com/sqlfluff/sqlfluff/issues/4031. These ones are all straightforward recodes. All the rules are still intact and just moved into new bundles as per the [rules spreadsheet](https://docs.google.com/spreadsheets/d/1xGig5Eq0vTdc7d82wwTXSD4X1BUmk1zgiMUuoVPRfaU).

No major functionality changes here. Just recoding.


pwildenhain (approved):
Inch by inch (or in your case 2.5cm by 2.5cm), we're getting there 📏

PR #4456, opened 2023-03-03
Layout Rules Recode (part 2)

alanmcruickshank:
This is a follow on from #4432 . It's part of #4031 .

This PR is primarily about consolidating all the other spacing rules into `LT01`. This removes _a lot_ of individual rules, but results in much less code to basically do the same thing. We rely on layout configs to allow users to tweek formats for their setup. Specifically, `LT01` now combines:
- L001: Trailing Whitespace
- L005 & L008: Space around commas
- L006: Space around operators
- L023: Space after AS in WITH clause
- L024: Space immediately after USING
- L039: Unnecessary Whitespace
- L048: Spacing around quoted literals
- L071: Spacing around brackets

In the process this also revises a couple of spacing configurations. In particular I've changed `inline` to be a qualifier so that we can have `single:inline` and `touch:inline` (where the latter was the previous meaning of just `inline`). This also allows a little more modularity in the code which handles it.

In the original issue - I suggested that we should also combine L015 and L017 into here too, but on closer inspection I think that's a bad idea. I'll handle them separately.


src/sqlfluff/utils/reflow/respace.py
```diff
@@ -29,47 +29,77 @@ def determine_constraints(
...
+    elif modifier == "inline":
+        strip_newlines = True
+    else:  # pragma: no cover
+        raise NotImplementedError(
+            f"Unexpected within constraint modifier: {pre_constraint}"
+        )
+
     post_constraint = next_block.spacing_before if next_block else "single"
+    # Unless align, split.
+    if post_constraint.startswith("align"):
```
WittierDinosaur, line 48: Can you not just loop through the constraints rather than repeating the code block?

alanmcruickshank: yeah that's probably a better idea... 🤔

WittierDinosaur (approved):
LGTM

docs/source/configuration.rst
```diff
@@ -291,7 +291,7 @@ above:
 .. code-block:: cfg
 
     [sqlfluff]
-    warnings = L019, L006
+    warnings = L019, LT01
```
pwildenhain, line 294: Oh interesting, I didn't know we were downgrading this to a warning

alanmcruickshank: ... we're not - it's just an example 😄

pwildenhain: 🙃

PR #4462, opened 2023-03-05
Rules Reorg Mopup

alanmcruickshank:
Almost done with #4031. This picks up a selection of rules that haven't been recoded yet and puts them into existing bundles. No significant functionality changes.
- L009 -> LT12
- L015 -> ST08 (this one was originally slated to be a layout rule, but I think it's actually structural)
- L016 -> LT05
- L017 -> LT06
- L018 -> LT07
- L022 -> LT08
- L036 -> LT09
- L041 -> LT10
- L065 -> LT11

After this PR (and #4456), we're almost done with recode. It leaves only L031 and L050 to go - which are slightly more complicated cases.


WittierDinosaur (approved):
Files changed 104 💀 but LGTM

Issue #4030, opened 2022-11-05
2.0.0 Release Checklist


alanmcruickshank:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

#3970 kicked off the timeline for 2.0.0 by explicitly mentioning it in the docs. This issue lays out the things that need to happen before release.

# Why 2.0.0?

There are a few backward incompatible changes coming and planned, which should trigger a major release. Those are:
- Reflow (#3673 ). In particular because of the impacts on configuration and on the parse tree.
  - #3942
  - #4067
- `sqlfluff format`, planned in #2436.
- Reorganisation of rule codes and categories: #4031

# How should we go about the release?

We should merge things as incrementally as possible. Pre-releases are a good thing. But the first full release should contain all the breaking bits.

There *should be a migration guide*.

# What's the checklist:

Before release we need:
- [x] Reindent merged #3942 
- [x] A solution to long lines (L016). #4067 
- [x] Rules reorganised, especially once reflow is all sorted. Old ones removed. Nice deprecation warnings etc... (#4031)
- [x] `sqlfluff format` command, working and integrated with rules. #4473 
- [ ] Migration guide.
- [x] Fix for #3989
- [x] Fix for #3988 
- [x] DBT templater changes: #3976
- [x] Named config references where possible for rules (re convo between @alanmcruickshank and @pwildenhain ) #4465  #4470 

### Dialect

All

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

PR #4465, opened 2023-03-05
Rule config lookup improvements & config warnings

alanmcruickshank:
This PR is part of #4030 and stems from a conversation between @alanmcruickshank and @pwildenhain .

It sets the groundwork for being able to configure rules using the `name` in the config rather than the `code`. This does all the logic changes, but doesn't actually flip the switch yet (see `BaseRule.get_config_ref()` still returns `self.code` for now).

When the method is switched to return `self.name` for the rules which implement it, rules will be default look for their config at `sqlfluff:rules:my_rule_name`, which is much easier to interpret and understand when reading a config file. To avoid the ambiguity of having multiple potential config settings, where a rule does define a name - it can _only be configured by the name_. If a rule does not define a name, _it can only be configured by the code_. This means there is no ambiguity, but still allows backward compatibility for plugins which don't yet define a `name`.

This PR moves the `code` and `description` extraction logic into the metaclass introduced in #4415, this should have a minor performance gain but mostly a simpler ability to extract the `code` and `description` before the instantiation of the rule.

This PR _also_ does explicit checking against the provided configuration. If any rule config sections are found which _don't_ correspond to a valid rule reference, a warning will be shown. This has several upsides:
- on releasing 2.0.0, this is a convenient way to warn people they need to update their config files and provides hints on how to do so.
- if config files contain errors in the section headings, they will now be warned to the user rather than silently ignored.
- internally within the project we can be more confident that there aren't errors in our config because they will be explicitly raised.

I chose to _warn_ rather than error so that people can continue to still use the tool.

Additionally, this solves two other small issues:
1. As brought up in [this comment](https://github.com/sqlfluff/sqlfluff/pull/4468#discussion_r1126445313), I've added some validation for rule names, enforcing _lowercase_ and _snake_case_
2. I've also closed a small loophole with group inheritance where groups for `CPO2` were not showing in the docs.

My intent is to stage a second PR to actually do the swap of the return value of `get_config_ref`, because that PR will be very large as it will require updating all of the configuration files in the project.


alanmcruickshank:
hmmmm - needs more tests. Will do that before putting this up for review again.

src/sqlfluff/core/rules/base.py
```diff
@@ -499,6 +499,7 @@ class RuleMetaclass(type):
         flags=re.MULTILINE,
     )
     _valid_classname_regex = regex.compile(r"Rule_?([A-Z]{1}[a-zA-Z]+)?_([A-Z0-9]{4})")
+    _valid_rule_name_regex = regex.compile(r"[a-z\.\_]+")
```
WittierDinosaur, line 502: thoughts on `a-z[a-z\.\_]+`?

alanmcruickshank: yeah, that would be better... I'll update it.

src/sqlfluff/core/rules/base.py
```diff
@@ -614,6 +684,12 @@ class BaseRule(metaclass=RuleMetaclass):
     # referred to this rule.
     aliases: Tuple[str, ...] = ()
 
+    # NOTE: code and description are provided here as hints, but should not
+    # be set directly. They are set automatically by the metaclass based on
+    # the class _name_ when defined.
+    code: str = ""
```
WittierDinosaur, line 690: for hints, do you need to set a default value? Can't you simply
```
code: str
```

alanmcruickshank: Good question - I'll try it 🤷

alanmcruickshank: looks like that works. Change coming through now...

WittierDinosaur (approved):
LGTM

PR #4468, opened 2023-03-06
Recode L050

alanmcruickshank:
This is part of #4031 . Initially I thought we might need to modify L050 because it does overlap somewhat with `LT01` (as it also catches initial trailing whitespace). However - I think that small overlap is ok (and also not new).

This is a less breaking change. Just recode L050 > LT13 and be done 👍


src/sqlfluff/rules/layout/LT13.py
```diff
@@ -63,7 +63,9 @@ class Rule_L050(BaseRule):
             foo
     """
 
-    groups = ("all",)
+    name = "layout.start-of-file"
```
WittierDinosaur, line 66: shouldn't this be snake case?

WittierDinosaur: also, possibly a good test to add, as an easy mistake to make when writing a new rule

alanmcruickshank: oh - good catch. Yes it should 👍

alanmcruickshank: I like the idea of validating rule names too - I'll do that as a new PR though.

PR #4470, opened 2023-03-06
Named Config (part 2)

alanmcruickshank:
This is the second part of #4465, also part of the critical path for #4030. Very little functionality change here, mostly just the impact of swapping rule config lookups from codes to names. All the functionality change is in the first PR.

In the process I've reshuffled the `default_config.cfg` file so that rules from similar groupings are next to eachother.

You'll see there is _a lot_ of find-and-replace here 💀 .


pwildenhain (approved):
This is awesome ✨ 

Makes both the config and our tests so much more readable! 2.0.0 here we come 😎

PR #4473, opened 2023-03-06
Add a sqlfluff format CLI command

### LINKED ISSUE #2436: Add 'reflow' capability for configurable, black-like deterministic formatting

alanmcruickshank:
Ok this is a long time coming. This resolves #2436 and adds the `sqlfluff format` CLI command.

It's basically a variant on `sqlfluff fix`, but with a hardcoded set of rules. This has a nice upside that if someone is using `exclude_rules`, they can still exclude rules - but they can't enable additional ones, so it remains fairly safe.

As a first pass I've added the following rules to the command:
- All of the capitalisation rules: `capitalisation`
- All of the layout rules: `layout`
- `ambiguous.union`
- `convention.not_equal`
- `convention.coalesce`
- `convention.select_trailing_comma`
- `convention.is_null`
- `jinja.padding`
- `structure.distinct`

I haven't added _extensive_ testing, because the majority of the machinery is the same as `sqlfluff fix`. There should be enough here to get coverage.

How to review (@barrywhart):
1. I've moved several common `click` arguments into `@lint_options`. No changes to the functionality, just a reorg. Those options are `--processes` and `--disable-progress-bar`. That happens in the first commit in the PR afc6ec3ac913ccad4fb3b4c10fe6772cc8e5f10e
2. I've broken out the original `fix` command into two branches (and two methods), one for stdin fixing (`_stdin_fix()`) and one for file based fixing (`_paths_fix()`) - both are used in `fix` and `format`. No functionality changes - just refactor.
3. Introduce the `cli_format()` method to support `sqlfluff format` cli command. It's mostly a copy paste of `sqlfluff fix` but with several things removed or changed to simplify things. Specifically, it hard codes the `rules` argument and always sets the `force` argument.

### DIFF
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -704,6 +704,163 @@ def do_fixes(lnt, result, formatter=None, **kwargs):
     return False  # pragma: no cover
 
 
+def _stdin_fix(linter, formatter, fix_even_unparsable):
+    """Handle fixing from stdin."""
+    exit_code = EXIT_SUCCESS
+    stdin = sys.stdin.read()
+
+    result = linter.lint_string_wrapped(stdin, fname="stdin", fix=True)
+    templater_error = result.num_violations(types=SQLTemplaterError) > 0
+    unfixable_error = result.num_violations(types=SQLLintError, fixable=False) > 0
+    if not fix_even_unparsable:
+        exit_code = formatter.handle_files_with_tmp_or_prs_errors(result)
+
+    if result.num_violations(types=SQLLintError, fixable=True) > 0:
+        stdout = result.paths[0].files[0].fix_string()[0]
+    else:
+        stdout = stdin
+
+    if templater_error:
+        click.echo(
+            formatter.colorize(
+                "Fix aborted due to unparsable template variables.",
+                Color.red,
+            ),
+            err=True,
+        )
+        click.echo(
+            formatter.colorize(
+                "Use --FIX-EVEN-UNPARSABLE' to attempt to fix the SQL anyway.",
+                Color.red,
+            ),
+            err=True,
+        )
+
+    if unfixable_error:
+        click.echo(
+            formatter.colorize("Unfixable violations detected.", Color.red),
+            err=True,
+        )
+
+    click.echo(stdout, nl=False)
+    sys.exit(EXIT_FAIL if templater_error or unfixable_error else exit_code)
+
+
+def _paths_fix(
+    linter,
+    formatter,
+    paths,
+    processes,
+    fix_even_unparsable,
+    force,
+    fixed_suffix,
+    bench,
+    show_lint_violations,
+    warn_force: bool = True,
+):
+    """Handle fixing from paths."""
+    # Lint the paths (not with the fix argument at this stage), outputting as we go.
+    click.echo("==== finding fixable violations ====")
+    exit_code = EXIT_SUCCESS
+
+    with PathAndUserErrorHandler(formatter):
+        result = linter.lint_paths(
+            paths,
+            fix=True,
+            ignore_non_existent_files=False,
+            processes=processes,
+        )
+
+    if not fix_even_unparsable:
+        exit_code = formatter.handle_files_with_tmp_or_prs_errors(result)
+
+    # NB: We filter to linting violations here, because they're
+    # the only ones which can be potentially fixed.
+    if result.num_violations(types=SQLLintError, fixable=True) > 0:
+        click.echo("==== fixing violations ====")
+        click.echo(
+            f"{result.num_violations(types=SQLLintError, fixable=True)} fixable "
+            "linting violations found"
+        )
+        if force:
+            if warn_force:
+                click.echo(
+                    f"{formatter.colorize('FORCE MODE', Color.red)}: "
+                    "Attempting fixes..."
+                )
+            success = do_fixes(
+                linter,
+                result,
+                formatter,
+                types=SQLLintError,
+                fixed_file_suffix=fixed_suffix,
+            )
+            if not success:
+                sys.exit(EXIT_FAIL)  # pragma: no cover
+        else:
+            click.echo(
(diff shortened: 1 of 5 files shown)


src/sqlfluff/cli/commands.py
```diff
@@ -695,154 +704,67 @@ def do_fixes(lnt, result, formatter=None, **kwargs):
     return False  # pragma: no cover
 
 
-@cli.command(cls=DeprecatedOptionsCommand)
```
barrywhart, line 698: @alanmcruickshank: Can you suggest how to review this? It looks like a lot of changes starting here, but I'm guessing some of this was code reorg?

src/sqlfluff/cli/commands.py
```diff
@@ -937,6 +861,223 @@ def fix(
...
+@core_options
+@lint_options
+@click.option(
+    "-x",
+    "--fixed-suffix",
+    default=None,
+    help="An optional suffix to add to fixed files.",
+)
+@click.option(
+    "--FIX-EVEN-UNPARSABLE",
```
barrywhart, line 975: In the spirit of `format` being a simpler, safer command, I wonder if we should consider omitting this option, which is known to be unsafe.

src/sqlfluff/cli/commands.py
```diff
@@ -937,6 +861,223 @@ def fix(
...
+        "Enables fixing of files that have templating or parse errors. "
+        "Note that the similar-sounding '--ignore' or 'noqa' features merely "
+        "prevent errors from being *displayed*. For safety reasons, the 'fix'"
+        "command will not make any fixes in files that have templating or parse "
+        "errors unless '--FIX-EVEN-UNPARSABLE' is enabled on the command line"
+        "or in the .sqlfluff config file."
+    ),
+)
+@click.option(
+    "--show-lint-violations",
```
barrywhart, line 988: Is this option relevant for `format`?

src/sqlfluff/rules/jinja/JJ01.py
```diff
@@ -48,6 +48,9 @@ class Rule_JJ01(BaseRule):
     # sections.
     crawl_behaviour = SegmentSeekerCrawler({"raw"})
     targets_templated = True
+    # NOTE: This rule does return fixes, but when is_fix_compatible
+    # is set, it spirals.
+    # is_fix_compatible = True
```
barrywhart, line 53: ???

alanmcruickshank: I've made a separate issue for this: https://github.com/sqlfluff/sqlfluff/issues/4474.

I left the comment there in case someone else tries to do the same as what I tried.

src/sqlfluff/cli/commands.py
alanmcruickshank: probably not.

src/sqlfluff/cli/commands.py
alanmcruickshank: Good point. I'll add a summary in the PR above.

barrywhart (approved):
🎉💯

PR #4488, opened 2023-03-09
Clickhouse support all join types 

### LINKED ISSUE #3506: ClickHouse dialect does not support all join types
### Search before asking

### What Happened

ClickHouse supports more join types than ANSI SQL, including `semi join` and `anti join`. Their docs show a join clause syntax like
```
SELECT <expr_list>
FROM <left_table>
[GLOBAL] [INNER|LEFT|RIGHT|FULL|CROSS] [OUTER|SEMI|ANTI|ANY|ASOF] JOIN <right_table>
(ON <expr_list>)|(USING <column_list>) ...
```
([Source](https://clickhouse.com/docs/en/sql-reference/statements/select/join/#supported-types-of-join))

However, SQLFluff misinterprets the `semi` in `semi join`, for example, as a table alias. It doesn't seem to support join types beyond typical ones like `left join` or `inner join`.

### Expected Behaviour

The tool should understand all of the join types in the syntax above.

### Observed Behaviour

A join query like this one:

```sql
select
    example1.a,
    b
from example1
    semi join example2 on (example1.a = example2.a);
```

yields these warnings (note how it thinks `semi` is an alias for `example1`, while it's really a join type like `left join` or `inner join`):

```
❯ sqlfluff lint example.sql --dialect clickhouse
== [example.sql] FAIL
L:  19 | P:   5 | L027 | Unqualified reference 'b' found in select with more than
                       | one referenced table/view.
L:  21 | P:   5 | L011 | Implicit/explicit aliasing of table.
L:  21 | P:   5 | L025 | Alias 'semi' is never used in SELECT statement.
L:  21 | P:   5 | L031 | Avoid aliases in from clauses and join conditions.
L:  21 | P:  10 | L051 | Join clauses should be fully qualified.
All Finished 📜 🎉!
```

### How to reproduce

```sql
-- example.sql
select
    example1.a,
    b
from example1
    semi join example2 on (example1.a = example2.a);
```

Command:

```
sqlfluff lint example.sql --dialect clickhouse
```

### Dialect

ClickHouse

### Version

```
❯ sqlfluff --version
sqlfluff, version 1.0.0

❯ python --version
Python 3.7.13
```

### Configuration

Just the stock configuration, no other config files.

### Code of Conduct

konnectr:

Closes #3506

The reserved names in Clickhouse will match the reserved names in ansi.

### Pull Request checklist

  - Other.


konnectr:
I have a question, how best to cover this case.
At Clickhouse:
`SELECT * FROM test1 LEFT ANTI JOIN test2 USING ty1,ty2;`
equal
`SELECT * FROM test1 ANTI LEFT  JOIN test2 USING ty1,ty2;`

I have made a dialect for the moment only for the first option. What is the best way to support two cases?

src/sqlfluff/dialects/dialect_clickhouse.py
```diff
@@ -25,7 +28,96 @@
...
+                Sequence(
+                    Ref.keyword("RIGHT"),
+                    OneOf(
+                        "OUTER",
+                        "ANTI",
+                        "SEMI",
+                        "ANY",
+                        optional=True,
+                    ),
+                ),
```
tunetheweb, line 65:
```suggestion
                Sequence(
                    "LEFT",
                    OneOf(
                        "OUTER",
                        "ANTI",
                        "SEMI",
                        "ANY",
                        "ASOF",
                        optional=True,
                    ),
                ),
                Sequence(
                    "ANTI",
                    "LEFT",
                ),
                Sequence(
                    "RIGHT",
                    OneOf(
                        "OUTER",
                        "ANTI",
                        "SEMI",
                        "ANY",
                        optional=True,
                    ),
                ),
                Sequence(
                    "ANTI",
                    "RIGHT",
                ),
```

src/sqlfluff/dialects/dialect_clickhouse.py
```diff
@@ -25,7 +28,96 @@
...
+
+    match_grammar = OneOf(
+        Sequence(
+            Ref("JoinTypeKeywords", optional=True),
+            Ref("JoinKeywordsGrammar"),
+            Indent,
+            Ref("FromExpressionElementSegment"),
+            Dedent,
+            Conditional(Indent, indented_using_on=True),
+            # NB: this is optional
```
tunetheweb, line 88: Hence why we have optional attribute so don't think you need this.

```suggestion
```

tunetheweb (changes requested):
LGTM but a couple of minor suggestions

tunetheweb:
> I have a question, how best to cover this case. At Clickhouse: `SELECT * FROM test1 LEFT ANTI JOIN test2 USING ty1,ty2;` equal `SELECT * FROM test1 ANTI LEFT JOIN test2 USING ty1,ty2;`
> 
> I have made a dialect for the moment only for the first option. What is the best way to support two cases? @tunetheweb

Made a suggestion in the code review as to how it could be covered. I don't know if only `ANTI` is supported like this, or if the full pre join keyword should have been repeated?

src/sqlfluff/dialects/dialect_clickhouse.py
konnectr: I changed the Sequence

src/sqlfluff/dialects/dialect_clickhouse.py
```diff
@@ -71,6 +213,7 @@ class FromExpressionElementSegment(ansi.FromExpressionElementSegment):
                 Ref("SamplingExpressionSegment"),
                 Ref("JoinLikeClauseGrammar"),
                 Ref.keyword("Final"),
```
tunetheweb, line 215: Not your issue but just noticed this. IT doesn't need the Ref.keyword and `Final` should be in uppercase.

```suggestion
                FINAL,
```

tunetheweb (approved):

src/sqlfluff/dialects/dialect_clickhouse.py
```diff
@@ -70,7 +212,8 @@ class FromExpressionElementSegment(ansi.FromExpressionElementSegment):
             exclude=OneOf(
                 Ref("SamplingExpressionSegment"),
                 Ref("JoinLikeClauseGrammar"),
-                Ref.keyword("Final"),
+                FINAL,
```
tunetheweb, line 215:
```suggestion
                "FINAL",
```

tunetheweb:
@konnectr it looks like ST07 has an issue with `USING` clauses that do not uses brackets - like some of your queries here. This was causing one of the CI jobs to fail. I've therefore disabled this rule for the Clickhouse dialect in https://github.com/sqlfluff/sqlfluff/pull/4488/commits/ef40cc90919a2110a00bccec404fe731a82cab33

Hoping CI passes now!

PR #4490, opened 2023-03-10
Update reserved keywords in Athena language

ValentinCrr:

Based on [this documentation](https://docs.aws.amazon.com/athena/latest/ug/reserved-words.html), "time" is not a reserved keyword, but rather a DDL reserved keyword.

Hence, TIME keyword was moved from "reserved" to "unreserved keywords" in Athena dialect.

No

### Pull Request checklist

  - Other.


tunetheweb (approved):
Change looks good to me.

However, there are quite a few other "keywords" in a similar situation. Should we move them too? `DATE` is one that maybe should be moved even if the others aren't?

BTW if you do decide to move more, I don't think we need a test for ALL of them. A test for `time` is good as that specifically caused you an issue, but wouldn't bother adding a specific test for the rest (though maybe one for `date` too?)

tunetheweb:
> That would indeed be interesting to move them all.

Thanks!

> Should I include all keywords from List of reserved keywords in DDL statements but not present in List of reserved keywords in SQL SELECT statements ?

Maybe. I say use your judgement as an Athena user.

PR #4497, opened 2023-03-12
Bug in operator precedence

alanmcruickshank:
Found a small bug in operator precedence while testing 2.0.0a6. This resolves that.

Lowercase `and` operators were picked up with a different precedence to uppercase `AND` because we were using `.raw` and not `.raw_upper`.


WittierDinosaur (approved):
LGTM

PR #4499, opened 2023-03-12
Fix #4433 (more untaken positive indents)

### LINKED ISSUE #4433: 2.0.0a5 Multiline `SELECT` are (still) not correctly indented in CTA
### Search before asking

### What Happened

This is a continuation of https://github.com/sqlfluff/sqlfluff/issues/4234 but I chose to create a new issue instead as this was tested on latest pre-release.

The issue wrong indentation still appears if the queries get more complex. Consider the following three examples (note the slightly different indentation in Example 3)

**Example 1:**
```sql
CREATE TABLE mytable
AS
(SELECT
    id,
    user_id
FROM another_table
)
;
```

**Example 2:**
```sql
CREATE TABLE mytable
AS
(SELECT
    id,
    user_id
FROM another_table
WHERE TRUE
)
;
```

**Example 3:**
```sql
CREATE TABLE mytable
AS
(SELECT
    id,
    user_id
FROM another_table
WHERE 
    TRUE
)
;
```

Examples 1&2 work perfectly, however Example 3 stays as it is.

### Expected Behaviour

See above

### Observed Behaviour

See above

### How to reproduce

See above

### Dialect

BigQuery

### Version

2.0.0a5

### Configuration

default

### Code of Conduct

alanmcruickshank:
This resolves #4433 . It makes the "untaken positive indent" code (introduced in #4420), and makes it a little more powerful, but keeping an exception for any indents which neighbour any placeholders.

This seems safe for now, although I wonder if there's a more elegant solution in the long run.


WittierDinosaur (approved):
LGTM

PR #4506, opened 2023-03-13
Revise templating and lexing of calls.

### LINKED ISSUE #4013: Internal error when linting dbt model that contains a call statement
### Search before asking

### What Happened

I used `sqlfluff lint --dialect snowflake myfile.sql` using the requirements below and got internal errors when linting a dbt model. It seems to be a different error than the one reported in #3611 
```
dbt-core==1.3
dbt-snowflake==1.3.0
streamlit==1.12.0
sqlfluff==1.4.1
sqlfluff-templater-dbt==1.4.1
```

I got a warning/error with this traceback:
```
root@645cc3519edf:/app# sqlfluff lint --dialect snowflake models/error_fluff.sql 
=== [dbt templater] Compiling dbt project...
=== [dbt templater] Project Compiled.
WARNING    Unable to lint models/error_fluff.sql due to an internal error. Please report this as an issue with your query's contents and stacktrace below!
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/linter/runner.py", line 108, in run
    yield partial()
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/linter/linter.py", line 769, in lint_rendered
    parsed = cls.parse_rendered(rendered)
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/linter/linter.py", line 378, in parse_rendered
    tokens, lvs, config = cls._lex_templated_file(
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/linter/linter.py", line 159, in _lex_templated_file
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/parser/lexer.py", line 564, in lex
    segments: Tuple[RawSegment, ...] = self.elements_to_segments(
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/parser/lexer.py", line 620, in elements_to_segments
    placeholders, source_slice, block_stack = _generate_placeholder_segments(
  File "/usr/local/lib/python3.8/site-packages/sqlfluff/core/parser/lexer.py", line 446, in _generate_placeholder_segments
    block_uuid = block_stack.pop()
IndexError: pop from empty list
```

### Expected Behaviour

- linting does not fail or
- error is more informative

### Observed Behaviour

I obtained the error & traceback above when linting this file:

```
{{
  config(
    materialized = 'view'
 )
}}
{% call statement('unique_keys', fetch_result=True) %}
  select 'tests' as key_name
{% endcall %}
{% set unique_keys = load_result('unique_keys') %}

with temp as (
    select 1 as dummy
    union all 
    select 2
    union all 
    select 99
),
prep as (
select 
    row_number() over (order by dummy asc ) as id,
    {%- if execute %}
    {%- for key_name in unique_keys['data'] %}
    {%- if key_name[0] not in ('tests')%}
        'this_was_a_test_indeed'
    {%- else %}
    coalesce('this was not a test','error')
    {%- endif %}
        as was_it_a_test,
    {%- endfor %}
    {% endif %}

    1 as dummy
from temp)
select * from prep
```

removing the call statement block resolves the linting error (so this one doesn't fail):
```
{{
  config(
    materialized = 'view'
 )
}}
{% set unique_keys = load_result('unique_keys') %}

with temp as (
    select 1 as dummy
    union all 
    select 2
    union all 
    select 99
),
prep as (
select 
    row_number() over (order by dummy asc ) as id,
    {%- if execute %}
    {%- for key_name in unique_keys['data'] %}
    {%- if key_name[0] not in ('tests')%}
        'this_was_a_test_indeed'
    {%- else %}
    coalesce('this was not a test','error')
    {%- endif %}
        as was_it_a_test,
    {%- endfor %}
    {% endif %}

    1 as dummy
from temp)
select * from prep
```

### How to reproduce

To reproduce:
I'm running this on docker, but I would imagine this error is independent of the environment:

docker-compose:
```
version: '3.3'
services:
    frosty-local:
        command: /bin/bash
        build: .
        volumes:
            - ./:/app/
            # - ~/.dbt:/root/.dbt
            - ~/.ssh:/root/.ssh

        environment:
          - several vars exported to set up the profiles.yml     
        

        ports:
            - "8085:8085"
            - "8501:8501"
```  

Dockerfile:

```
FROM python:3.8.2-buster

RUN apt-get update -yqq \
  && apt-get install -yqq \
    postgresql \
    libssl-dev \
    less \
    vim \
    libffi-dev \
    libpq-dev

# upgrade pip
RUN pip install --upgrade pip

RUN echo 'alias ll="ls -lari"' >> ~/.bashrc

# Install dependencies:
COPY requirements.txt .
RUN pip install -r requirements.txt

ENV DBT_PROFILES_DIR=/app/profiles
RUN mkdir -p /app
COPY ./ /app/
RUN chmod 777 -R /app
WORKDIR /app/
RUN dbt deps
EXPOSE 8501
```   
error_fluff.sql:
```
{{
  config(
    materialized = 'view'
 )
}}
{% call statement('unique_keys', fetch_result=True) %}
  select 'tests' as key_name
{% endcall %}
{% set unique_keys = load_result('unique_keys') %}

with temp as (
    select 1 as dummy
    union all 
    select 2
    union all 
    select 99
),
prep as (
select 
    row_number() over (order by dummy asc ) as id,
    {%- if execute %}
    {%- for key_name in unique_keys['data'] %}
    {%- if key_name[0] not in ('tests')%}
        'this_was_a_test_indeed'
    {%- else %}
    coalesce('this was not a test','error')
    {%- endif %}
        as was_it_a_test,
    {%- endfor %}
    {% endif %}

    1 as dummy
from temp)
select * from prep
```

command:
`sqlfluff lint --dialect snowflake models/error_fluff.sql`

### Dialect

sqlfluff-templater-dbt & dbt-snowflake

### Version

1.4.1

### Configuration

sqlfluff:
```
[sqlfluff]
verbose = 0
nocolor = False
# Or run 'sqlfluff dialects'
dialect = None
templater = dbt
rules = L001,L002,L003,L004,L005,L006,L008,L009,L010,L011,L012,L014,L015,L016,L017,L018,L019,L020,L021,L022,L023,L025,L027,L028,L029,L030,L032,L036,L038,L039,L040,L041,L042,L043,L045,L046,L048,L049,L051,L053,L054,L055,L057,L058,L059,L060,L061,L062,L063,L065
# all
exclude_rules = L007,L013,L024,L026,L031,L033,L034,L035,L037,L044,L047,L050,L052,L056,L064,L066,
recurse = 0
output_line_length = 80
runaway_limit = 10
ignore = None
ignore_templated_areas = True
encoding = autodetect
disable_noqa = False
fix_even_unparsable = False
# The more efficient check is the _byte_ limit check which
# is enabled by default. The previous _character_ limit check
# is still present for backward compatibility. This will be
# removed in a future version.
# Set either to 0 to disable.
large_file_skip_char_limit = 0
large_file_skip_byte_limit = 20000
# e.g. -1 means use all processors but one. 0  means all cpus.
processes = 1

[sqlfluff:indentation]
# See https://docs.sqlfluff.com/en/stable/layout.html#configuring-indent-locations
indented_joins = True
indented_ctes = True
indented_using_on = True
indented_on_contents = True
template_blocks_indent = True

# Layout configuration
# See https://docs.sqlfluff.com/en/stable/layout.html#configuring-layout-and-spacing
[sqlfluff:layout:type:comma]
spacing_before = touch
line_position = trailing

[sqlfluff:layout:type:binary_operator]
line_position = leading

[sqlfluff:layout:type:statement_terminator]
spacing_before = touch
line_position = trailing

[sqlfluff:layout:type:end_of_file]
spacing_before = touch

[sqlfluff:layout:type:set_operator]
line_position = alone:strict

[sqlfluff:layout:type:start_bracket]
spacing_after = touch

[sqlfluff:layout:type:end_bracket]
spacing_before = touch

[sqlfluff:layout:type:start_square_bracket]
spacing_after = touch

spacing_before = touch

[sqlfluff:layout:type:casting_operator]
spacing_before = touch
spacing_after = touch

[sqlfluff:layout:type:comparison_operator]
spacing_within = touch
line_position = leading

[sqlfluff:layout:type:object_reference]
spacing_within = inline

[sqlfluff:layout:type:comment]
spacing_before = any
spacing_after = any

# Template loop tokens shouldn't dictate spacing around them.
[sqlfluff:layout:type:template_loop]
spacing_before = any
spacing_after = any

[sqlfluff:templater]
unwrap_wrapped_queries = True

[sqlfluff:templater:jinja]
apply_dbt_builtins = True

[sqlfluff:rules]
tab_space_size = 4
# Max line length is set by default to be in line with the dbt style guide.
# https://github.com/dbt-labs/corp/blob/main/dbt_style_guide.md
max_line_length = 80
indent_unit = space
allow_scalar = True

[sqlfluff:rules:L003]
hanging_indents = True

[sqlfluff:rules:L010]
# Keywords
capitalisation_policy = upper
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L011]
aliasing = explicit

[sqlfluff:rules:L012]
aliasing = explicit

[sqlfluff:rules:L014]
# Unquoted identifiers
ignore_words = None
ignore_words_regex = None
# unquoted_identifiers_policy = aliases

[sqlfluff:rules:L016]
# Line length
ignore_comment_lines = True
max_line_length = 100

[sqlfluff:rules:L027]
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L026]
force_enable = False

[sqlfluff:rules:L028]
force_enable = False

[sqlfluff:rules:L029]
ignore_words = email,first_name,last_name,value
ignore_words_regex = (^.*_$),(^.*\$.*$)

[sqlfluff:rules:L030]
# Function names
# consistent
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L031]
force_enable = False

[sqlfluff:rules:L036]
wildcard_policy = single

[sqlfluff:rules:L038]
# Trailing commas

[sqlfluff:rules:L040]
# Null & Boolean Literals
capitalisation_policy = lower
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L042]
forbid_subquery_in = both
# join

[sqlfluff:rules:L047]
prefer_count_1 = False
prefer_count_0 = False

[sqlfluff:rules:L051]
# Fully qualify JOIN clause

[sqlfluff:rules:L052]
multiline_newline = False

[sqlfluff:rules:L054]

[sqlfluff:rules:L057]
additional_allowed_characters = None
ignore_words = email,first_name,last_name,metadata$filename,metadata$file_row_number 
ignore_words_regex = (^.*_$)

[sqlfluff:rules:L059]
ignore_words = None
ignore_words_regex = None
force_enable = False

[sqlfluff:rules:L062]
blocked_words = None
blocked_regex = None

[sqlfluff:rules:L063]
# Data Types
ignore_words = None
ignore_words_regex = None

[sqlfluff:rules:L064]
force_enable = False

[sqlfluff:rules:L066]
min_alias_length = None
max_alias_length = None
```

. sqlfluffignore
```
target/
dbt_packages/
/app/dbt_packages/
macros/
models/challenge_20*
```

### Code of Conduct

alanmcruickshank:
This resolves #4013. It adds a couple of test cases which cover this specific scenario more completely. While we were testing the _templating_ of call clauses, we weren't testing their lexing too. This adds better coverage for that and also removes the special handling of `call` clauses which was causing the issue.


test/core/templaters/jinja_test.py
```diff
@@ -866,7 +866,8 @@ def test__templater_jinja_slice_template(test, result):
 
 
 def _statement(*args, **kwargs):
-    return "_statement"
+    # NOTE: The standard dbt statement() call returns nothing.
```
barrywhart, line 869: Should this be a _new_ test case rather than changing the existing one? In other words, just because dbt's `statement()` call returns nothing, it seems like it should still be valid for some other Jinja or dbt macro to return something. As written, I worry we may be breaking something or creating an opening for future issues.

alanmcruickshank: Good question 🤔 . I think I'll keep this test the same, but I will add a more general test for jinja `call` blocks 👍

alanmcruickshank: That was a *great* call. It was waaaay more tricky than expected to do that, but I'm confident that the result is much better. I now understand a bit more why it was originally done the way you did.

Turns out that the conflict was that the lexer assumes that blocks must have zero length. The tracer puts any `call` results in the `block_start`. When we renamed the `block_start` to a `templated` section - that made the code _work_ but block counting in the lexer would fail. I've update the code now to allow `block_start` sections to have length, and then block count effectively when they do.

barrywhart: In light of the discussion above, are you planning to undo the change of `"_statement"` -> `""`?

barrywhart: (Having a bit of trouble comprehending the test additions vs changes.)

alanmcruickshank: In this case I don't think so. The example query further down won't parse properly if we do that. What I wanted to do instead was have another test case where we do a `{% call something_else() %}`, rather than just `statement()` (which isn't something supported out of the box in jinja - or supported by the context of tools like airflow).

In short:
- `{% call statement(...) %}` I think should always return nothing, but just so the test cases are representative of likely use cases.
- We do _also_ need test cases which also cover situations where `{% call xyz() %}` does return a string with length, which is why I assed the jinja tests with `JINJA_MACRO_CALL_SQL` and the additional test cases in `lexer_test.py`.

barrywhart (approved):

PR #4508, opened 2023-03-13
Bigquery spacing

alanmcruickshank:
This resolves a few issues in bigquery specific spacing raised during testing by @tunetheweb .

Specifically:
- spacing after `EXCEPT` and `REPLACE`.
- spacing around angle brackets.
- spacing around array type specifiers
- spacing within qualified function names.

In doing this I've also revised the parsing of array types more generally so that the "type" part of it is a distinct segment. This is partly necessary for the spacing rules to apply properly, but also I think it's more representative of how array literals are structured.

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -101,6 +101,12 @@ spacing_after = touch
 [sqlfluff:layout:type:end_square_bracket]
 spacing_before = touch
 
+[sqlfluff:layout:type:start_angle_bracket]
+spacing_after = touch
+
+[sqlfluff:layout:type:end_angle_bracket]
+spacing_before = touch
+
 [sqlfluff:layout:type:casting_operator]
 spacing_before = touch
 spacing_after = touch:inline
@@ -123,17 +129,9 @@ spacing_within = touch:inline
 spacing_after = touch:inline
 
 [sqlfluff:layout:type:function_name]
+spacing_within = touch:inline
 spacing_after = touch:inline
 
-[sqlfluff:layout:type:select_except_clause]
-# This prevents space between EXCEPT following brackets:
-# e.g. SELECT * EXCEPT()
-spacing_within = touch:inline
-
-[sqlfluff:layout:type:select_replace_clause]
-# This prevents space between REPLACE following brackets:
-# e.g. SELECT * REPLACE()
-spacing_within = touch:inline
 
 [sqlfluff:layout:type:snowflake_semi_structured_expression]
 spacing_within = touch:inline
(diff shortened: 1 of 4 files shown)


src/sqlfluff/core/default_config.cfg
```diff
@@ -101,6 +101,12 @@ spacing_after = touch
 [sqlfluff:layout:type:end_square_bracket]
 spacing_before = touch
 
+[sqlfluff:layout:type:start_angle_bracket]
+spacing_after = touch
+
+[sqlfluff:layout:type:end_angle_bracket]
+spacing_before = touch
```
tunetheweb, line 108: As discussed it is possible to make this an optional `touch`? The following should not touch before `LANGUAGE` for example:

```sql
CREATE TEMPORARY FUNCTION DoSomething(param1 STRING, param2 STRING)
RETURNS ARRAY<STRING> LANGUAGE js AS """Some JS""";

SELECT DoSomething(col1) FROM table1
```

Big query does seem to accept it without the space but it looks weird.

src/sqlfluff/core/default_config.cfg
```diff
@@ -123,17 +129,12 @@ spacing_within = touch:inline
 spacing_after = touch:inline
 
 [sqlfluff:layout:type:function_name]
-spacing_after = touch:inline
-
-[sqlfluff:layout:type:select_except_clause]
-# This prevents space between EXCEPT following brackets:
-# e.g. SELECT * EXCEPT()
 spacing_within = touch:inline
+spacing_after = touch:inline
```
tunetheweb, line 133: I prefer this and think it should be the default as that's what the BigQuery docs have, but it means we need to update `test/fixtures/linter/autofix/bigquery/006_fix_ignore_templating/after.sql` (and probably before.sql too).

tunetheweb (changes requested):

src/sqlfluff/core/default_config.cfg
alanmcruickshank: That config applies to the spacing between `STRING` and `>`. The problematic one for that query is the `spacing_after` for `[sqlfluff:layout:type:array_type]`, as discussed via slack I've updated that to be more liberal for now until we have a better option.

src/sqlfluff/core/default_config.cfg
alanmcruickshank: yep - updated 👍

alanmcruickshank:
@tunetheweb - I've taken the `ARRAY` and `STRUCT` bits out of here so we can merge this and I'll have another go at them in a more consistent way.

tunetheweb (approved):

PR #4509, opened 2023-03-13
Add a deprecation warning for removed config option.

alanmcruickshank:
As part of revising layout config, I removed the `inline` option. @tunetheweb was still using it 😄 .

This adds a more helpful deprecation message and also auto-translates the value in the short term.

I've tested locally, but not adding to test suite because this should eventually be removed.


WittierDinosaur (approved):
LGTM

tunetheweb (approved):

PR #4511, opened 2023-03-13
Array and Struct Spacing

alanmcruickshank:
This is a continuation of #4508 .

To enable correct spacing of `ARRAY` and `STRUCT` declarations, this adds some more structures. While I'm adding them to enabled the spacing, this also allows a much more consistent treatment of both expressions across all dialects. There was quite a difference between many, with all of them trying to achieve roughly the same thing.

I do wonder whether `athena` should depend more directly on `hive`, but in any case this adds:
- `array_literal` and `struct_literal` which take the form `[x,y,z]` or `(x,y,z)` (NOTE: they _don't_ include the `ARRAY` or `STRUCT` specifiers). I expect that `struct_literal` without that qualification will be rare, but the structure stands.
- `array_type` and `struct_type` which take the form of `ARRAY` or `STRUCT` in most dialects, but some also allow type qualification, such as `ARRAY<int>`. `STRUCT<...>` is also supported in some, but the `<...>` portion is wrapped in a `struct_type_schema` segment to isolate it from spacing rules.
- `typed_struct_literal` and `typed_array_literal` which are the combination of the above, i.e. `ARRAY<int>[1,3,4]` or `STRUCT<...>(...)`.

The spacing rules are then defined within `typed_x_literal` and `x_type` which achieves the spacing desired (as shown by the bigquery tests).

### DIFF
--- a/src/sqlfluff/dialects/dialect_athena.py
+++ b/src/sqlfluff/dialects/dialect_athena.py
@@ -247,7 +247,27 @@ class ArrayTypeSegment(ansi.ArrayTypeSegment):
     """Prefix for array literals specifying the type."""
 
     type = "array_type"
-    match_grammar = Ref.keyword("ARRAY")
+    match_grammar = Sequence(
+        Ref.keyword("ARRAY"),
+        Ref("ArrayTypeSchemaSegment", optional=True),
+    )
+
+
+class ArrayTypeSchemaSegment(ansi.ArrayTypeSegment):
+    """Prefix for array literals specifying the type."""
+
+    type = "array_type_schema"
+    match_grammar = Bracketed(
+        Delimited(
+            Sequence(
+                Ref("DatatypeSegment"),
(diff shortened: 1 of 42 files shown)


src/sqlfluff/dialects/dialect_athena.py
```diff
@@ -244,6 +243,61 @@
 )
 
 
+class ArrayTypeSegment(ansi.ArrayTypeSegment):
+    """Prefix for array literals specifying the type."""
+
+    type = "array_type"
+    match_grammar = Sequence(
+        Ref.keyword("ARRAY"),
```
tunetheweb, line 251: Nit but doesn't require `Ref.keyword`

```suggestion
        "ARRAY",
```

tunetheweb (approved):
A lot of change here,  but ran through my large BigQuery code base that uses a lot of Structs and looks better now.

Issue #3559, opened 2022-07-06
Per file dialect selection


Clepech:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

Allow dialect selection per sql file.

The dialect could be declared as a header comment in each file, For example:

```sql
-- sqlfluff dialect : postgres

SELECT * FROM table
```

This declaration will override general configuration

### Use case

We run sqlfluff on a repo containing many SQL files with different dialects. Our current use is to set sqlfluff (as a pre-commit) with the most used dialect and ignore other dialects. This request aims to run sqlfluff on currently ignored files.

### Dialect

All dialect

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

tunetheweb:
Best practice would surely be to separate the files into different folders?:

- `sql/postgres/`
- `sql/bigquery/`

Within eaxch folder you could have a `.sqlfluff` file that sets the dialect and, at the top level, `sql` folder you could have a common `.sqlfluff` with the rest of the config.

anneadb:
This feature would be useful for my team as well: We have a lot of data processing pipelines that have stages in multiple separate database systems. It would not make sense to have subfolders depending on the dialect for each pipeline. Plus, then we would need to copy paste the .sqlfluff configuration file into each folder instead of defining it for an entire project.

tunetheweb:
> Plus, then we would need to copy paste the .sqlfluff configuration file into each folder instead of defining it for an entire project.

That's not true. Maybe my comment above wasn't clear, but as SQLFluff supports inheritance of config, you can set it up like this:"

- `sql/.sqlfluff` -- common sqlfluff config (lots of lines)
- `sql/postgres/.sqlfluff` -- a simple, 2-line config file setting the dialect to set it to postgres
- `sql/postgres/postgres-sql-file1.sql`
- `sql/postgres/postgres-sql-file2.sql`
- ...
- `sql/bigquery/.sqlfluff` -- a simple, 2-line config file setting the dialect to bigquery
- `sql/bigquery/postgres-sql-file1.sql`
- `sql/bigquery/bigquery-sql-file2.sql`
- ...

anneadb:
[...]

PR #4518, opened 2023-03-14
Per file dialect selection fix

LauraRichter:

This change addresses this issue: https://github.com/sqlfluff/sqlfluff/issues/3559

The issue author asked for it to be possible to set the dialect per file, and a suggestion in the issue discussion was that it could be set using an in-file configuration directive. From the documentation, it reads like this should already be possible:
https://docs.sqlfluff.com/en/stable/configuration.html
with syntax like this:
```
-- sqlfluff:dialect:bigquery
```

But it doesn't work. Why?
It is because the config parameter `core.dialect` will be changed to the value provided in the file, BUT, the parameter `core.dialect_obj` is set when the FluffConfig is initialised, before the file is read in, and not updated by the inline config.

So this pull request makes the behaviour consistent - now an inline dialect directive changes both the dialect and the dialect_obj

No

### Pull Request checklist

  - Other.


alanmcruickshank (commented):
Looks like a very sensible approach 👍 . Thanks for contributing this. I'll review properly once test suite has passed.

greg-finley:
Thanks for the contribution! I was able to link to the related issue manually, but you could link it next time in the PR description with one of the [magic words](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

LauraRichter:
> magic words

Oh I see! Thanks!

LauraRichter:
> @LauraRichter - there's one line not covered by your tests, are you able to add an additional test case which covers it? Everything else looks good.

Will do!

greg-finley (approved):
Thanks, nice work! 👍

PR #4530, opened 2023-03-15
Bigquery hyphentated identifiers

### LINKED ISSUE #4510: Bigquery project name fails to parse when it includes a hyphen and numeric segment
### Search before asking

### What Happened

This issue has come up before (https://github.com/sqlfluff/sqlfluff/pull/903) and was seemingly fixed, but it turns out the merged fix only works in some cases. Specifically it will parse a project name like `project-abc` just fine, but fails to parse one like `project-123`. From some basic experimentation I did, this will happen if any sub-element (i.e. after splitting on hyphens) of the project name begins with a numeric character. So for instance `project-a1` is fine, but not `project-1a`

### Expected Behaviour

No parse error

### Observed Behaviour

Described above.

### How to reproduce

Minimal example:

```
import sqlfluff
sqlfluff.parse(
    "SELECT * FROM a-1a.b.c",
    dialect="bigquery",
)
```

results in:

```
APIParsingError: Found 1 issues while parsing string.
Line 1, Position 16: Found unparsable section: '-1a.b.c'
```

### Dialect

bigquery

### Version

1.4.5

### Configuration

Default configuration

### Code of Conduct

alanmcruickshank:
This resolves #4510. Possible knock-on to other dialects given part of the fix is a very small adjustment to the regex which matches numeric literals.


test/fixtures/dialects/bigquery/create_table_hyphen_project.yml
```diff
@@ -14,7 +14,7 @@ file:
     - table_reference:
       - naked_identifier: project
       - dash: '-'
-      - naked_identifier: name
+      - naked_identifier_part: name
```
tunetheweb, line 17: This change means some rules (e.g. CP02, RF03, RF04...etc.) will no longer match here?

alanmcruickshank: ah good point - will revise

tunetheweb (approved):
Approve if CI passed

PR #4532, opened 2023-03-15
Attempt in-house coverage

alanmcruickshank:
This is as per https://hynek.me/articles/ditch-codecov-python/

This PR:
- Modularises python as dbt tests into reusable workflows (and seperate files).
- Adds local coverage testing, including a check that we can use for approving PRs. NB: I'll need to disable the existing checks to merge this, because it's still awaiting a bunch of tests that I've moved around or renamed.
- Consolidates the two smaller workflows into ci tests (the yaml check and the rules check) - this avoids drift between the triggering rules. The original aim was to they could be re-run separately - but that's very achievable still, even if they're in the same workflow.

Once this is approved, I'll force merge it and update the checks.


greg-finley (approved):
🎉

PR #4547, opened 2023-03-16
Rethink test segregation in CI

alanmcruickshank:
This stems from a discussion on slack about the duration of the test suite. Previously, the longest duration was around 16 minutes. The longest parts are below, taken from this example: https://github.com/sqlfluff/sqlfluff/actions/runs/4441331387
- The python matrix tests (without coverage): 7-11mins
- Python 3.11 tests (with coverage): 13mins
- Python 3.10 windows test (with coverage): 16mins
- _"Critical errors test"_: 8mins

This PR does a few things:
- It expands the use of [pytest marks](https://docs.pytest.org/en/7.1.x/how-to/mark.html), from just the existing `integration_test` mark to:
  - A `fix_suite` mark which maps roughly to the original `integration_test` mark.
  - A `parse_suite` mark which covers the similar parsing suite for parsing structure tests and parsing/lexing success tests.
  - A `rules_suite` mark which covers the rules yaml tests.
  - An `integration` mark which also covers the above three suites.
- For the default python matrix tests, it excludes the `integration` mark, which reduces runtimes to ~2mins without coverage, ~3min for linux with coverage and ~4min for windows with coverage.
- Creates new test jobs for each of the three other suites:
  - _"Dialect parsing"_ - which covers the `parse_suite`, taking ~8mins
  - _"Rules yaml test cases"_ - which covers the `rules_suite`, taking ~4mins
  - _"Dialect fixing"_ - which covers the `fix_suite`, taking ~6mins
- Additionally, the existing _"Critical errors test"_, (now called the _"Dialect fixing"_ job), was running the `.parse()` methods twice, once in the `lex_and_parse.()` method and once in `Parser.parse()` method. A small refactor of `dialects_test.py` removes this duplication.

**The critical path of the whole CI run takes around 9mins, rather than around 17mins**. 🎉🎉

The limiting factor is the `parse_suite` (and then the coverage reporting job briefly afterwards). The next longest job is the `fix_suite` which is also very sensitive to parsing times. Reducing parsing times (@WittierDinosaur 😄 ), is the most sensible thing to reduce times further and could get us as far as 5mins (which is the runtime of the tests on the dbt templater plugin - and the next longest job).

Additionally, if merged, the `rules_suite` and `parse_suite` should be made _Required_ tests.

### DIFF
--- a/.github/workflows/ci-test-python.yml
+++ b/.github/workflows/ci-test-python.yml
@@ -58,6 +58,14 @@ jobs:
       if: ${{ !inputs.coverage }}
       run: tox -e py${{ steps.py_version.outputs.PYVERSION }} -- -n 2 test -m "${{ inputs.marks }}"
 
+    - name: Rename coverage files with suffix
+      # NOTE: We do this because we're using the same tox environment for multiple
+      # test jobs and we need to make sure that their coverage files don't collide.s
+      if: ${{ !inputs.coverage }}
+      run: |
+        COVSUFFIX=$(echo "${{ inputs.marks }}" | sed -e 's/ /-/g')
+        for file in .coverage.*; do mv "$file" "$file.$COVSUFFIX"; done;
+
     - name: Coveralls Parallel (coveralls)
       uses: coverallsapp/github-action@master
       if: ${{ inputs.coverage }}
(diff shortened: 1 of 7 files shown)


tunetheweb (approved):
LGTM

Great speed improvement!

PR #4558, opened 2023-03-18
Address #1630: Optionally move comments after long line

### LINKED ISSUE #1630: Optionally allow rule L016 to move online comments AFTER the current line
I ran `sqlfluff lint` on

```sql
SELECT
    CASE
        WHEN o IS NULL
        AND z='str str str str str str' THEN i -- comment comment comment comment comment comment comment comment comment comment
    END
```

The output has a changed order of rows.

## Expected Behaviour
Order of rows and commands remains unchanged

## Observed Behaviour
```sql
SELECT CASE
        WHEN o IS NULL
            -- comment comment comment comment comment comment comment comment comment comment
             AND z = 'str str str str str str' THEN i
    END
```

## Steps to Reproduce
Run lint on code above

## Dialect
postgres

## Version

## Configuration
```
[sqlfluff]
verbose = 0
nocolor = False
dialect = postgres
templater = jinja
rules = None
exclude_rules = L011,L031,L034
recurse = 0
output_line_length = 80
runaway_limit = 10
ignore_templated_areas = True
encoding = autodetect

[sqlfluff:indentation]
indented_joins = False
indented_using_on = True
template_blocks_indent = True

[sqlfluff:templater]
unwrap_wrapped_queries = True

[sqlfluff:templater:jinja]
apply_dbt_builtins = True

[sqlfluff:rules]
tab_space_size = 4
max_line_length = 120
indent_unit = space
comma_style = trailing
allow_scalar = True

[sqlfluff:rules:L003]
lint_templated_tokens = True

operator_new_lines = after

[sqlfluff:rules:L016]
ignore_comment_lines = False

[sqlfluff:rules:L038]

[sqlfluff:rules:L042]
forbid_subquery_in = join

prefer_count_1 = False
prefer_count_0 = False

```

alanmcruickshank:
This resolves #1630 . It's also a good opportunity to add a little more modularity and clean up some duplicated code in `lint_line_length()`. This extracts the routines for moving a trailing comment into `_fix_long_line_with_comment()` and also adds a configuration flag to that method to optionally move comments _after_ rather than _before_ the line they're currently on.

### DIFF
--- a/src/sqlfluff/utils/reflow/reindent.py
+++ b/src/sqlfluff/utils/reflow/reindent.py
@@ -1593,56 +1593,44 @@ def _fix_long_line_with_comment(
 
     # Start of file case.
     if last_indent_idx is None:
-        elements = (
-            [
-                line_buffer[-1],
-                ReflowPoint((NewlineSegment(),)),
-            ]
-            + line_buffer[:-2]
-            + elements[last_elem_idx + 1 :]
-        )
-        fixes.append(
-            LintFix.create_before(
-                first_seg,
-                [
-                    comment_seg,
-                    NewlineSegment(),
-                ],
-            )
-        )
-        return elements, fixes
-
-    # Normal "before" case
+        new_point = ReflowPoint((NewlineSegment(),))
+        prev_elems = []
+        anchor = first_seg
     else:
-        fixes.append(
-            # NOTE: This looks a little convoluted, but we create
-            # *before* a block here rather than *after* a point,
-            # because the point may have been modified already by
-            # reflow code and may not be a reliable anchor.
-            LintFix.create_before(
-                elements[last_indent_idx + 1].segments[0],
-                [
-                    comment_seg,
-                    NewlineSegment(),
-                    WhitespaceSegment(current_indent),
-                ],
+        new_point = ReflowPoint(
+            (
+                NewlineSegment(),
+                WhitespaceSegment(current_indent),
             )
         )
-        elements = (
-            elements[: last_indent_idx + 1]
-            + [
-                line_buffer[-1],
-                ReflowPoint(
-                    (
-                        NewlineSegment(),
-                        WhitespaceSegment(current_indent),
-                    )
-                ),
-            ]
-            + line_buffer[:-2]
-            + elements[last_elem_idx + 1 :]
+        prev_elems = elements[: last_indent_idx + 1]
+        anchor = elements[last_indent_idx + 1].segments[0]
+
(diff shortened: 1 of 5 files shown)


src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -1546,12 +1546,108 @@ def _match_indents(
...
+        return elements, []
+
+    comment_seg = line_buffer[-1].segments[-1]
+    first_seg = line_buffer[0].segments[0]
+    last_elem_idx = elements.index(line_buffer[-1])
+
+    assert trailing_comments in (
+        "after",
+        "before",
+    ), f"Unexpected value for `trailing_comments`: {trailing_comments!r}"
```
barrywhart, line 1576: Is the value also checked elsewhere (farther upstream)? I can't recall if we do anything systematic about checking user config settings.

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -1546,12 +1546,108 @@ def _match_indents(
...
+
+    # The simpler case if if we're moving the comment to the line
+    # _after_. In that case we just coerce the point before it to
+    # be an indent.
+    if trailing_comments == "after":
+        anchor_point = cast(ReflowPoint, line_buffer[-2])
+        results, new_point = anchor_point.indent_to(current_indent, before=comment_seg)
+        # Mutate elements
+        elements[last_elem_idx - 1] = new_point
```
barrywhart, line 1585: Similar to the other PR, we're mutating **and** returning a list. Could we mutate without returning?

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: No, not at the moment. I think that might increase the scope of this PR a lot though, I'll raise another issue for it.

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: In this case, we actually _need_ to return it (see the later clause of the function), but I can update this section so that is doesn't mutate, and then we're entirely in _just return the value_.

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: see: https://github.com/sqlfluff/sqlfluff/issues/4578

barrywhart (approved):

PR #4560, opened 2023-03-19
Indents on window functions

### LINKED ISSUE #4033: Indentation of window functions
### Search before asking

### What Happened

As mentioned in #1839 by @alexscott-ff, the following query is currently acceptable by the indentation engine:

```sql
SELECT
    field_one,
    -- Shouldn't line continuations be indented?
    min(event_tracking_activity_id)
    OVER (PARTITION BY some_id_used_to_partition) = event_tracking_activity_id
    AS is_first_val
FROM
    my_table
```

This feels strange because the `OVER` clause isn't indented.

### Expected Behaviour

I would expect a better indentation to be:

```sql
SELECT
    min(event_tracking_activity_id)
        OVER (PARTITION BY some_id_used_to_partition) = event_tracking_activity_id
        AS is_first_val
FROM
    my_table
```

which I think is also congruent with an indentation for normal functions as follows:

```sql
SELECT
    min(event_tracking_activity_id)
        AS is_first_val
FROM
    my_table
```

The `Indent` and `Dedent` tokens should be reviewed to achieve the desired outcome.

### Observed Behaviour

see above

### How to reproduce

lint using #3942

### Dialect

ansi

### Version

main

### Configuration

ansi, L003 only

### Code of Conduct

alanmcruickshank:
This resolves #4033. It builds on #4706.

On the downside, this leads to a slightly _longer_ default fix for long lines. This is for the following reason:

```sql
-- Take the following long line:
select
    my_function(col1 + col2, arg2, arg3) over (partition by col3, col4 order by col5 rows between unbounded preceding and current row) as my_relatively_long_alias
from foo;

-- ...the first fix is to add an indents before `over` and before `as` (as the _outermost_ indents) to give this:
select
    my_function(col1 + col2, arg2, arg3)
        over (partition by col3, col4 order by col5 rows between unbounded preceding and current row) 
        as my_relatively_long_alias
from foo;

-- ... which because it's still long is then fixed to:
select
    my_function(col1 + col2, arg2, arg3)
        over (
            partition by col3, col4
            order by col5 rows between unbounded preceding and current row
        ) 
        as my_relatively_long_alias
from foo;
```

The final result is definitely _correct_, but there's a more condensed layout, which is _also_ acceptable, just not suggested by the fixing routine because of the priority logic, is:

```sql
select
    my_function(col1 + col2, arg2, arg3) over (
        partition by col3, col4
        order by col5 rows between unbounded preceding and current row
    ) as my_relatively_long_alias
from foo;
```

I think this is ok - and the fix presented is definitely _correct_, it's just that some users might be confused with the suggested layout if they otherwise prefer the secondary one.


alanmcruickshank:
Fixing this relies on #4564 - I'll pick this up again once that's merged.

alanmcruickshank:
I suspect this is falling foul of the same bug as #4598 and so should be unblocked by #4622

alanmcruickshank:
Now that #4706 is merged, I'm going to repurpose this to be an PR just about the window functions part. I've pushed that variant of it to this branch for testing.

WittierDinosaur (approved):
Nice!

PR #4561, opened 2023-03-19
Remove matches_target_tuples (#3873)

### LINKED ISSUE #3873: Refactor/remove matches_target_tuples
Nitpick: Looping and checking `elem[0]` seems messy. I wonder if we should consider splitting `target_tuples` into multiple, separate function parameters, `raw_upper`, `type`, `parenttype`.

_Originally posted by @barrywhart in https://github.com/sqlfluff/sqlfluff/pull/3824#discussion_r973606062_

alanmcruickshank:
This resolves #3873.

Brought up in reviewing the rule crawler work, we identified that `BaseSegment.matches_target_tuples()` was an overly complicated (and arguably too _powerful_) method that could/should be replaced.

This removes the last references to it and replaces those instances with much simpler calls.

### DIFF
--- a/src/sqlfluff/rules/capitalisation/CP05.py
+++ b/src/sqlfluff/rules/capitalisation/CP05.py
@@ -50,13 +50,8 @@ class Rule_CP05(Rule_CP01):
             "data_type",
         }
     )
-    _target_elems: List[Tuple[str, str]] = [
-        ("parenttype", "data_type"),
-        ("parenttype", "datetime_type_identifier"),
-        ("parenttype", "primitive_type"),
-        ("type", "data_type_identifier"),
-    ]
-    _exclude_elements: List[Tuple[str, str]] = []
+    _exclude_types = ()
+    _exclude_parent_types = ()
     config_keywords = [
         "extended_capitalisation_policy",
         "ignore_words",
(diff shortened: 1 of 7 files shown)


src/sqlfluff/rules/capitalisation/CP05.py
```diff
@@ -65,20 +60,13 @@ class Rule_CP05(Rule_CP01):
     _description_elem = "Datatypes"
 
     def _eval(self, context: RuleContext) -> List[LintResult]:
-        """Inconsistent capitalisation of keywords.
+        """Inconsistent capitalisation of datatypes.
 
         We use the `memory` feature here to keep track of cases known to be
-        INconsistent with what we've seen so far as well as the top choice
+        Inconsistent with what we've seen so far as well as the top choice
```
barrywhart, line 66:
```suggestion
        inconsistent with what we've seen so far as well as the top choice
```

src/sqlfluff/rules/capitalisation/CP05.py
```diff
@@ -50,13 +50,8 @@ class Rule_CP05(Rule_CP01):
...
         }
     )
-    _target_elems: List[Tuple[str, str]] = [
-        ("parenttype", "data_type"),
-        ("parenttype", "datetime_type_identifier"),
-        ("parenttype", "primitive_type"),
-        ("type", "data_type_identifier"),
-    ]
-    _exclude_elements: List[Tuple[str, str]] = []
+    _exclude_types = ()
```
barrywhart, line 53: It looks like the exclusion moved farther down in the rule, around line 89. Any concerns about that? Anything worth adding a comment about?

alanmcruickshank: For this rule, I don't think we're changing the logic at all.

- The clause starting `# Skip if not an element of the specified type/name` wasn't being called at all, because it was matching against `_exclude_elements`, which was being set to an empty list. I've just removed it.
- The second check, I've just refactored - no change in logic.
- As for `_exclude_types` and `_exclude_parent_types`, they will never actually be called by `CP05`, because it overrides `_eval` and then only calls `_handle_segment` from the parent which doesn't use those checks. I'm setting them to empty here more as a tidy up. This last point is probably worth stating in a comment - I'll add that.

PR #4563, opened 2023-03-19
Don't force newlines before commas

alanmcruickshank:
Found this while investigating #4033. Commas behave similarly to semicolons in that we should allow them to trail (unless the leading/trailing comma rule decides to move them). I'm raising this as a draft to check for knock-on effects.


alanmcruickshank:
Hmmmm, ok - coverage is hard here. I'm going to just loop this into another PR and close this one.

PR #4564, opened 2023-03-19
Refactor lint_line_length and fix comma bug

alanmcruickshank:
This includes the changes made in the (now closed) PR: #4563
This also includes the changes made in the (separate) PR: #4562
This relates to #4033, although doesn't actually change the default indentation yet. This fixes all the bugs that I found in getting us ready for that. I'll actually change the indentation as a separate PR.

Similarly to #4558 this extracts a chunk of `.lint_line_length()` and moves it into smaller functions to make the code more manageable.

This PR also:
- Fixes a problem where dedents before commas forced unexpected newlines (so I've added them to the special case which already handled semi-colons).
- Prunes long line fixes from splitting short sections which don't need splitting. e.g. the second line of the new test case where we don't need to split `my_function(col1 + col2, arg2, arg3)`, it's the later part that is the problem.

### DIFF
--- a/src/sqlfluff/utils/reflow/reindent.py
+++ b/src/sqlfluff/utils/reflow/reindent.py
@@ -1545,6 +1545,126 @@ def _match_indents(
     return matched_indents
 
 
+def _fix_long_line_with_fractional_targets(
+    elements: ReflowSequenceType, target_breaks: List[int], desired_indent: str
+) -> Tuple[ReflowSequenceType, List[LintResult]]:
+    """Work out fixes for splitting a long line at locations like operators.
+
+    This is a helper function within .lint_line_length().
+    """
+    line_results = []
+    for e_idx in target_breaks:
+        e = cast(ReflowPoint, elements[e_idx])
+        new_results, new_point = e.indent_to(
+            desired_indent,
+            after=elements[e_idx - 1].segments[-1],
+            before=elements[e_idx + 1].segments[0],
+        )
+        # NOTE: Mutation of elements.
+        elements[e_idx] = new_point
+        line_results += new_results
+    return elements, line_results
+
+
+def _fix_long_line_with_integer_targets(
+    elements: ReflowSequenceType,
+    target_breaks: List[int],
+    line_length_limit: int,
+    inner_indent: str,
+    outer_indent: str,
+    allow_implicit_indents: bool,
+) -> Tuple[ReflowSequenceType, List[LintResult]]:
+    """Work out fixes for splitting a long line at locations like indents.
+
+    This is a helper function within .lint_line_length().
+    """
+    line_results = []
+    # Create a stash of indent_stats. We're going to need them
+    # twice, so we generate them one for later use.
+    _indent_stats_cache: Dict[int, IndentStats] = {}
+    for e_idx in target_breaks:
+        # Generate indent stats for it.
+        e = cast(ReflowPoint, elements[e_idx])
+        # We need to check for negative sections so they get the right
+        # indent (otherwise they'll be over indented).
+        # The `desired_indent` above is for the "uphill" side.
+        following_class_types = elements[e_idx + 1].class_types
+        indent_stats = e.get_indent_impulse(
+            allow_implicit_indents, following_class_types
+        )
+        # Cache them for later
+        _indent_stats_cache[e_idx] = indent_stats
+
+    # If we can get to the uphill indent of later break, and still be within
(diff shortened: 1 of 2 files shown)


src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -751,14 +751,6 @@ def _map_line_buffers(
...
                         continue
 
-                    # Second, check for placeholders. Indents around placeholders
-                    # are trickier to reason about. For now, don't force untaken
-                    # indents around placeholders.
-                    if "placeholder" in elements[loc + 1].class_types or (
-                        loc >= 1 and "placeholder" in elements[loc - 1].class_types
-                    ):
-                        continue
-
```
alanmcruickshank, line 761: This was only included recently (in #4499 last week), and after these changes isn't covered any more by tests. I'm removing it as no longer required.

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -1546,6 +1537,126 @@ def _match_indents(
...
+        else:
+            new_indent = inner_indent
+
+        new_results, new_point = e.indent_to(
+            new_indent,
+            after=elements[e_idx - 1].segments[-1],
+            before=elements[e_idx + 1].segments[0],
+        )
+        # NOTE: Mutation of elements.
+        elements[e_idx] = new_point
```
barrywhart, line 1645: Same question about mutation vs returning a list.

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -1793,55 +1910,34 @@ def lint_line_length(
...
-                    indent_stats = e.get_indent_impulse(
-                        allow_implicit_indents, following_class_types
+                # Is one of the locations the final element? If so remove it.
+                # There's already a line break there.
+                if i in target_breaks:
+                    target_breaks.remove(i)
+
+                # Is it an "integer" indent or a fractional indent.
+                # Handle accordingly.
+                if target_balance % 1 == 0:
```
barrywhart, line 1928: This seems wrong. **Any** integer returns 0 when you do `% 1` on it.

alanmcruickshank: ah yes - but fractional values don't. I'll update the comments to be clearer.

src/sqlfluff/utils/reflow/reindent.py
```diff
@@ -1546,6 +1537,126 @@ def _match_indents(
...
+    line_results = []
+    for e_idx in target_breaks:
+        e = cast(ReflowPoint, elements[e_idx])
+        new_results, new_point = e.indent_to(
+            desired_indent,
+            after=elements[e_idx - 1].segments[-1],
+            before=elements[e_idx + 1].segments[0],
+        )
+        # NOTE: Mutation of elements.
+        elements[e_idx] = new_point
```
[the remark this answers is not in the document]
alanmcruickshank: yeah, that's fair. Good idea.

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: I've updated the comment to:
```sql
                # Is it an "integer" indent or a fractional indent?
                # Integer indents (i.e. 1.0, 2.0, ...) are based on Indent and
                # Dedent tokens. Fractional indents (i.e. 1.5, 1.52, ...) are
                # based more on rebreak spans (e.g. around commas and operators).
                # The latter is simpler in that it doesn't change the indents,
                # just adds line breaks. The former is more complicated.
                # NOTE: Both of these methods mutate the `elem_buffer`.
                if target_balance % 1 == 0:
                    ...
```

src/sqlfluff/utils/reflow/reindent.py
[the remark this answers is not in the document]
alanmcruickshank: I've updated it to only mutate and not return 👍

src/sqlfluff/utils/reflow/reindent.py
alanmcruickshank: I've updated it to only mutate and not return 👍

src/sqlfluff/utils/reflow/reindent.py
barrywhart: Sounds good. There's also `math.modf()`, but this way is fine now that you added the comment. I never think about using `%` with floats. 🤯

PR #4575, opened 2023-03-20
Added keywords to DB2 dialect from IBM docs

aurany:
- The purpose of this PR is to add DB2 keywords from [IBM docs](https://www.ibm.com/docs/en/db2/11.5?topic=sql-reserved-schema-names-reserved-words).
- Instead of adding reserved and unreserved keywords separately we reuse reserved keywords from ansi and add DB2 specific keywords to ansi unreserved list.


aurany:
> Are there some examples of SQL that wouldn't parse before this change but do parse after this change? If so, we could add some test fixtures?

I don't think so. We only add unreserved keywords to be more restrictive. Reserved keywords are left unchanged.

greg-finley (approved):
Thanks for the contribution!

PR #4576, opened 2023-03-20
Try parse caching

alanmcruickshank:
This is a draft PR, which in testing on some files has _drastically_ reduced parsing time.

Part of the issue in parsing was that we were re-matching the same segments against the same matchers over and over again. This gets around that by caching results which can be reused if we try the same match again.

Draft until tests are passing.


WittierDinosaur (approved):
LGTM

PR #4577, opened 2023-03-20
Use terminators in BaseExpression

alanmcruickshank:
On heavily nested expressions, being able to shortcut when a "long enough" match is found helps significantly. On combination with #4576, this has a very significant effect on heavily nested files.

It's a slightly dialled back version of what I think could be achieved, but it's a step in the right direction and does have some impact already.


WittierDinosaur (approved):
LGTM - though this may just want deprecating if we get inheritance going properly

PR #4585, opened 2023-03-21
Fix ignored implicit indent after `GROUP BY` clause

### LINKED ISSUE #4539: 2.0.0 - Implicit indent not working on group by
### Search before asking

### What Happened

SQLFluff v2.0.0 gives an LT02 error when parsing a multi-line `group by` with a hanging/implicit indent with `allow_implicit_indents = True`.

### Expected Behaviour

I expected the SQL to parse without errors.

### Observed Behaviour

SQLFluff gave the following error message:

```
L:   7 | P:   9 | LT02 | Expected line break and indent of 4 spaces before 'a'.
                       | [layout.indent]
```

Related to https://github.com/sqlfluff/sqlfluff/issues/4467 ?

### How to reproduce

SQL file:
```
select
    a,
    b

from my_table

group by a,
    b
```

### Dialect

`ansi`

### Version

```
> sqlfluff --version
sqlfluff, version 2.0.0

> python --version
Python 3.9.9
```

### Configuration

```
[sqlfluff]
dialect = ansi

[sqlfluff:indentation]
allow_implicit_indents = True
```

### Code of Conduct

borchero:

In the spirit of #4583, this ought to resolve #4539.

Different formatting when using `allow_implicit_indents = true`.

### Pull Request checklist

  - Other.


test/fixtures/rules/std_rule_cases/LT02-indent.yml
```diff
@@ -1815,6 +1815,84 @@ test_fix_untaken_positive_4433:
...
+  pass_str: |
+    SELECT t.col1, t.col2
+    FROM t
+    GROUP BY t.col1, t.col2
+  configs:
+    core:
+      dialect: ansi
+    indentation:
+      allow_implicit_indents: true
+
```
alanmcruickshank, line 1829: I'm a little confused about these test cases - they would still pass even _without_ implicit indents. 🤔 

To test the changes that you're making don't you need a line break before `t.col2` in the `GROUP BY` clause?

fredriv: Was thinking the same thing 🙂

fredriv:
The implicit indent wouldn't require you to take it, so with this change you'd be able to choose between

```
GROUP BY t.col1, t.col2

GROUP BY t.col1,
    t.col2

GROUP BY
    t.col1,
    t.col2
```
where the middle alternative is not allowed today. Right?

WittierDinosaur:
I think as we roll more implicit indents out, we need a way to select in the config which implicit indents we enable. This one I think is the first controversial implementation. @alanmcruickshank what are your thoughts on this?

alanmcruickshank:
> I think as we roll more implicit indents out, we need a way to select in the config which implicit indents we enable. This one I think is the first controversial implementation. @alanmcruickshank what are your thoughts on this?

I think you're right that this should be a configurable one. Happy to support it as an option, but I don't think it should be a default. The `Conditional` grammar should already allow this kind of thing.  What I'm not sure about is whether this is an isolated example (which would be configured with something like `implicit_group_by`) or whether this is an example of a wider pattern which might apply to other places like `ORDER BY`, `SELECT` or even more generally within delimited places (like lists or arrays) (which would be blanket enabled with something like `implicit_delimited`).

@fredriv @borchero - do either of you have views on whether this is an isolated example or something more general?

Issue #4559, opened 2023-03-18
Implicit indents are not allowed within WHERE clauses


borchero:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

When setting `allow_implicit_indents = true` in the SQLFluff configuration, the option is not honored in `WHERE` clauses when running `sqlfluff fix`.

_Note:_ This is likely related to #4539. Thus, I believe this issue concerns clauses other than `WHERE` and `GROUP BY`.

### Expected Behaviour

The query should not be modified.

### Observed Behaviour

The query is modified to disallow implicit indents:

```sql
SELECT t.col1
FROM t
WHERE
    t.col2 = 'foo'
    AND t.col3 = 'bar'
```

### How to reproduce

Use the following SQL script:

```sql
SELECT t.col1
FROM t
WHERE t.col2 = 'foo'
    AND t.col3 = 'bar'
```

### Dialect

`tsql`

### Version

SQLFluff version 2.0.1

### Configuration

```
[sqlfluff]
dialect = tsql

[sqlfluff:indentation]
allow_implicit_indents = true
```

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

alanmcruickshank:
I think this is a `tsql` specific bug, due to that dialect having it's own implementation of some clauses. This should be an easy fix.

Issue #4579, opened 2023-03-21
`allow_implicit_indents = True` not working for case when


kieronellis:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

SQLFluff fix is moving the first `when` statement to a new line, despite setting `allow_implicit_indents = True`.

### Expected Behaviour

As the documentation for [implicit indents](https://docs.sqlfluff.com/en/stable/layout.html#implicit-indents) states:
```
An implicit indent is exactly like a normal indent, but doesn’t have to be actually taken to influence the indentation of lines after it - it just needs to be left un-closed before the end of the line. These are normally available in clauses which take the form of KEYWORD <expression>, like WHERE clauses or CASE expressions.

-- This WHERE clause here takes advantage of an implicit indent.
SELECT *
FROM my_table
WHERE condition_a
   AND condition_b;

-- With implicit indents disabled (which is currently the
-- default), the above formulation is not allowed, and instead
-- there should be a newline immediately after `WHERE` (which
-- is the location of the _implicit_ indent).
SELECT *
FROM my_table
WHERE
   condition_a
   AND condition_b;
```

Therefore the following SQL should not be formatted:
```
select
  case old_col when 'aaaaaaaaaaaaaaaaaaaaaaaaa-a' then 'aaaaaaaaaaaaaaaaaaaaaaaaa1'
    when 'aaaaaaaaaaaaaaaaaaaaaaaaa-b' then 'aaaaaaaaaaaaaaaaaaaaaaaaa2'
    when 'aaaaaaaaaaaaaaaaaaaaaaaaa-c' then 'aaaaaaaaaaaaaaaaaaaaaaaaa3'
    when 'aaaaaaaaaaaaaaaaaaaaaaaaa-d' then 'aaaaaaaaaaaaaaaaaaaaaaaaa4'
  end as new_col
from my_table
```

### Observed Behaviour

Instead the above SQL gets formatted to 
```
select
  case old_col
    when 'aaaaaaaaaaaaaaaaaaaaaaaaa-a' then 'aaaaaaaaaaaaaaaaaaaaaaaaa1'
    when 'aaaaaaaaaaaaaaaaaaaaaaaaa-b' then 'aaaaaaaaaaaaaaaaaaaaaaaaa2'
[...]

PR #4587, opened 2023-03-21
Fix ignored implicit indent for `CASE .. WHEN` statements

borchero:

This should resolve https://github.com/sqlfluff/sqlfluff/issues/4559#issuecomment-1477811414.

Different formatting of `CASE .. WHEN` expression when `allow_implicit_indents = true`.

### Pull Request checklist

  - Other.


borchero:
Resolves #4590.

alanmcruickshank:
I think this also fixes #4579 .

@borchero - I don't know if I'm entirely on board with this layout, but I'm up for exploring it. See comment here: https://github.com/sqlfluff/sqlfluff/issues/4579#issuecomment-1480235760.

Of the indents you propose, I think I have different views:

## Implicit indent between `CASE` and `WHEN` in `CASE WHEN`

Adding the implicit indent in the situation where there isn't an expression between `CASE` and `WHEN`:

```sql
-- That would allow this...
SELECT
    CASE WHEN a THEN b
        WHEN c THEN d
        ELSE e
    END AS f
FROM foo;
-- ...which I think is ok.
-- It _also_ allows this though (which I think is kind of gross):
SELECT
    CASE WHEN a
            THEN b
        WHEN c THEN d
        ELSE e
    END AS f
FROM foo;
```

## Implicit indent between `CASE` and `WHEN` in `CASE <expression> WHEN`

Similarly to the above, but even more convoluted.

```sql
-- That would allow this...
SELECT
    CASE x + y WHEN a THEN b
        WHEN c THEN d
        ELSE e
    END AS f
FROM foo;
-- ...which I'm not sure is great and _also_ allows this (which is worse):
SELECT
    CASE x + y WHEN a
            THEN b
        WHEN c THEN d
        ELSE e
    END AS f
FROM foo;
```

## Implicit indent just before `THEN` in `CASE`

This one I think is fairly uncontroversial...

```sql
-- That would allow this...
SELECT
    CASE
        WHEN a THEN
            b
            + c
        WHEN c THEN d
        ELSE e
    END AS f
FROM foo;
```

...however, I do wonder if this one is more about configuring whether the default is for indents after `THEN` or before `THEN` in clauses like this.

@borchero - would you be able to add a wider range of test cases to flesh out what you're suggesting?

fredriv:
I tried out this PR on my examples from https://github.com/sqlfluff/sqlfluff/issues/4589, and while it now no longer requires a line break it still complains about the indentation, e.g. it wants to change from

```
select
    a,
    case
        when a = 1 then
            case
                when b then 'bar'
            end

        when a = 2 then case
            when b then 'bar'
        end
    end as c

from my_table
```

to
```
select
    a,
    case
        when a = 1 then
                case
                    when b then 'bar'
                end

        when a = 2 then case
                    when b then 'bar'
                end
    end as c

from my_table
```

So it seems like it wants an extra level or two of indentation in the nested case / when statements?

alanmcruickshank:
Yeah, I wondered if this might happen @fredriv. I think it's what I meant when I said:
> ...however, I do wonder if this one is more about configuring whether the default is for indents after THEN or before THEN in clauses like this.

I think a better solution to your specific problem would be to optionally allow a configuration which moves the normal `Indent` (not an implicit one), from _before_ the `THEN` to _after_ the `THEN`. We could implement that very similarly to the switches [described here](https://docs.sqlfluff.com/en/stable/layout.html#configuring-indent-locations) in the docs.

fredriv:
Added another more concrete example from our codebase to the other PR though: https://github.com/sqlfluff/sqlfluff/issues/4589#issuecomment-1480340967

I tried playing around a bit with the `WhenClauseSegment` of `dialect_ansi.py`, and by replacing
```
        Indent,
        "THEN",
        ImplicitIndent,
        Ref("ExpressionSegment"),
        Dedent,
        Dedent,
```
with
```
        "THEN",
        Indent,
        Ref("ExpressionSegment"),
        Dedent,
```

I got it to accept
```
select
    a,
    case
        when product_type = 'Product' then
            case
                when measured_weight > 0 then measured_weight
                else quantity * item_weight
            end
        else 0
    end as product_weight

from my_table;
```

Is that what you meant @alanmcruickshank (just making it configurable, of course)?

fredriv:
The existing `ImplicitIndent` after `THEN` works if I move `then` to the line after, but then it requires extra indentation for the `case` body - any way to avoid that? 🤔 
```
select
    a,
    case
        when product_type = 'Product'
            then case
                    when measured_weight > 0 then measured_weight
                    else quantity * item_weight
                end
        else 0
    end as product_weight

from my_table;
```

alanmcruickshank:
> Is that what you meant @alanmcruickshank (just making it configurable, of course)?

You've got the effect of what I'm suggesting yes, but not necessarily the implementation.

The default is:

```python
        "WHEN",
        Sequence(
            ImplicitIndent,
            Ref("ExpressionSegment"),
            Dedent,
        ),
        Indent,
        "THEN",
        ImplicitIndent,
        Ref("ExpressionSegment"),
        Dedent,
        Dedent,
    )
```

I think you're wanting an indentation more like:

```python
        "WHEN",
        Sequence(
            ImplicitIndent,
            Ref("ExpressionSegment"),
            Dedent,
        ),
        # No indent here
        "THEN",
        ImplicitIndent,
        Ref("ExpressionSegment"),
        Dedent,
        # No dedent here
    )
```

The functionality we have at the moment for switching indents like this is the `Conditional` grammar, as demonstrated in the `WithCompoundStatementSegment`:

```python
class WithCompoundStatementSegment(BaseSegment):
    """A `SELECT` statement preceded by a selection of `WITH` clauses.

    `WITH tab (col1,col2) AS (SELECT a,b FROM x)`
    """

    type = "with_compound_statement"
    # match grammar
        "WITH",
        Ref.keyword("RECURSIVE", optional=True),
        Conditional(Indent, indented_ctes=True),
        Delimited(
            Ref("CTEDefinitionSegment"),
            terminator=Ref.keyword("SELECT"),
        ),
        Conditional(Dedent, indented_ctes=True),
        OneOf(
            Ref("NonWithSelectableGrammar"),
            Ref("NonWithNonSelectableGrammar"),
        ),
    )
```

You can see that theres an `Indent` - `Dedent` pair which is only enabled if `indented_ctes` is `True` (and `indented_ctes` is assumed to be a configuration value in the `indentation` section).

I wonder if the solution here is making the `Indent` and `Dedent` that I mentioned above, conditional on some configuration value. I don't think that value should be set by default, because it's not the typical style that most style guides follow - but I'm very happy for it to be a configurable option for projects that prefer that style of formatting.

fredriv:
Sounds good! I could have a go at creating a PR if you want @alanmcruickshank? 🙂

alanmcruickshank:
> Sounds good! I could have a go at creating a PR if you want @alanmcruickshank? 🙂 

I would love that! ♥️ Always keen for new contributers.

PR #4601, opened 2023-03-23
--bench output with rule timings

alanmcruickshank:
This adds more granularity to the `--bench` output so that rule timings are visible (from performance session with @WittierDinosaur). Also adds rule timings to the timing csv output for performance tuning.


WittierDinosaur (approved):
LGTM

PR #4612, opened 2023-03-24
.simple() matching for TypedMatcher

alanmcruickshank:
This relates to #3390, although probably not obviously.

`RegexMatcher` is bad, because it doesn't prune nicely using the `.simple()` route. Much of the way to get around that is through a greater use of the `TypedMatcher`, which means we can push the regex operations up into the lexer. **However**, up till now the `TypedMatcher` didn't support `simple` matching _either_. This implements that feature.

As a side effect, some of the handing for _non-simple_ matching can be removed from the base grammar - resulting in a simpler codebase. While it's still _possible_ to end up with some non-simple options (because we haven't get removed all of the `RegexMatcher` options), they now aren't actually used in any of the relevant cases. Because the aim is to remove them in the long run, instead of keeping the handling logic (_"just in case"_), I've instead asserted that all the options _are_ simple and then we'll get an error if they are used in that position. If that does happen it should be a signal for us to migrate things faster to the `TypedMatcher`.

This change on it's own I think results in a 5-10% improvement in the parsing speed. It's not massive, but every little helps, and it enables some simplification in the codebase.

### DIFF
--- a/src/sqlfluff/core/parser/grammar/anyof.py
+++ b/src/sqlfluff/core/parser/grammar/anyof.py
@@ -1,6 +1,8 @@
 """AnyNumberOf and OneOf."""
 
-from typing import List, Optional, Tuple
+import logging
+
+from typing import List, Optional, Tuple, Set
 
 from sqlfluff.core.parser.context import ParseContext
 from sqlfluff.core.parser.grammar.base import (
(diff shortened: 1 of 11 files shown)


WittierDinosaur (approved):
Looks good!

PR #4598, opened 2023-03-23
Enable configuring whether to require indent before THEN

fredriv:
### Brief summary of the change made

This PR adds the configuration option `indented_then` (defaults to `True`) to enable configuration of whether to require an indent or not before `THEN` in a `CASE` clause. Added a test case in `LT02-indent.yml` and updated the documentation in `layout.rst`.

Fixes #4589.

### Are there any other side effects of this change that we should be aware of?

None.

### Pull Request checklist
- [X] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #4622, opened 2023-03-26
Sequence Meta Handling

alanmcruickshank:
The handling of metas and conditionals in the `Sequence` grammar had some holes in it (it also needed an overhaul to be more efficient in how it stepped through the options). This upgrades and refines that. I've also added some more comments to explain better what's happening.

In doing that I found some other issues:
- Tracking of jinja blocks nested within other blocks could lead to some strange indentation issues. I've extended the existing `block_stack` to a more feature rich `BlockTracker` class which provides a similar interface. All these changes are in `lexer.py`. I've added an additional test case to cover this.
- The block tracker also allows better identification of indents and dedents which are linked to templating. Affected indents now also have their `block_uuid` attribute set. Additionally for both indents and for placeholders, the start of the `block_uuid` is also shown in the parse output (via `._suffix()`) for better debugging and understanding.
- The slight shift caused by the main changes, uncovered issues in our template tag indentation code (i.e. in `reindent.py`). On reflection, I'm impressed the original code mostly worked - because I think that was an accident. This revision of that code is more robust to more complicated jinja handling. For _almost all_ cases, I've managed to maintain the existing functionality and keep this as a "backend only" change - however, there is one more complicated test case (`bigquery/003_templating`) where I've had to make slight adjustments to the test case output to make everything consistent, in particular it's now consistent with `bigquery/004_templating` - which I think it actually an improvement . Future iterations of this might refine that, but I think this is as close as I can get.

I would love to merge each of these changes separately, but each affects the other so I think the only sensible route is to merge them all together. Sorry that it's a bit of a biggie.

Given _almost all_ of the changes are "backend only", and the one test case that's substantively changing is now more consistent with others, I'm ok with this _not_ triggering more than a bugfix release.

This should also resolve the bug uncovered by #4598


WittierDinosaur (approved):
LGTM - I'm comfortable with the parser stuff. Less so with the reflow stuff, but I've never been near that code. All tests passing, so happy to merge

src/sqlfluff/core/parser/grammar/sequence.py
```diff
@@ -15,15 +15,54 @@
...
     BaseGrammar,
+    MatchableType,
     cached_method_for_parse_context,
 )
 from sqlfluff.core.parser.grammar.conditional import Conditional
 from os import getenv
 
 
+def _all_remaining_metas(
+    remaining_elements: Iterable[MatchableType], parse_context: ParseContext
```
barrywhart, line 30: Since we iterate over it multiple times, I _think_ `remaining_elements` should be a `Sequence` instead. Iterables don't necessarily allow iterating over them more then once (e.g. generators).

src/sqlfluff/core/parser/grammar/sequence.py
```diff
@@ -72,18 +111,46 @@ def match(self, segments, parse_context):
...
+                        self._elements[idx:], parse_context
                     )
-                else:
-                    pre_nc = ()
-                    mid_seg = unmatched_segments
-                    post_nc = ()
+                    if remaining_metas is not None:
+                        # We're safe. Claim them and return.
+                        meta_post_nc += remaining_metas
+                        early_break = True
```
barrywhart, line 124: Is this flag actually used anywhere?

src/sqlfluff/core/parser/lexer.py
```diff
@@ -20,12 +20,63 @@
...
+    _stack: List[UUID] = []
+    _map: Dict[Tuple[int, int], UUID] = {}
+
+    def enter(self, src_slice: slice) -> None:
+        """Add a block to the stack."""
+        key = to_tuple(src_slice)
+        uuid = self._map.get(key, None)
+
+        if not uuid:
+            uuid = uuid4()
```
barrywhart, line 51: In the JinjaAnalyzer, I initially used UUIDs, then later I realized that a simple counter had advantages:
* Same results every time
* Easier to read in debugging output (e.g. "0000001" vs a long random string)

Worth considering here?

barrywhart (approved):
Looks good! A few minor suggestions.

src/sqlfluff/core/parser/grammar/sequence.py
alanmcruickshank: yeah I think that's a good idea

src/sqlfluff/core/parser/grammar/sequence.py
alanmcruickshank: Yeah, it's used just up above as a way of "double breaking", i.e. a way to break both out of the `while` loop and _also` the `for` loop.

src/sqlfluff/core/parser/lexer.py
alanmcruickshank: Based on a quick investigation, I think you're right that we _could_ do that, but I don't know if it's actually _simpler_. Good shout however.

barrywhart: Mainly useful for debugging. No meaningful difference from a technical point of view.

PR #4658, opened 2023-04-02
Remove three RegexParsers

alanmcruickshank:
This makes a little progress on #3390.

The most substantive change here (at least as far as parse tree changes - is that the "root" matcher (`code`) now has it's own type rather than just showing up as `raw`). That means we can use it for matching using the `TypedParser` and remove a couple of the `RegexParser` usages.

The segments migrated from `RegexParser` to `TypedParser` are:
- `FunctionNameIdentifierSegment`
- `VersionIdentifierSegment`
- `NakedIdentifierFullSegment`

### DIFF
--- a/test/fixtures/dialects/exasol/AlterTableDistributePartition.yml
+++ b/test/fixtures/dialects/exasol/AlterTableDistributePartition.yml
@@ -3,7 +3,7 @@
 # computed by SQLFluff when running the tests. Please run
 # `python test/generate_parse_fixture_yml.py`  to generate them after adding or
 # altering SQL files.
-_hash: 45da3e37ea8947db3458800b70ff413f4f91f92fc0b62d6df43b7f9611fc1d9e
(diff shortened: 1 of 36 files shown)


test/fixtures/dialects/exasol/AlterTableDistributePartition.yml
```diff
@@ -62,10 +62,10 @@ file:
         - comma: ','
         - column_reference:
             naked_identifier: DISTRIBUTE
-        - raw: BY
-        - raw: shop_id
+        - code: BY
```
WittierDinosaur, line 65: This doesn't look more wrong than it was, but this does look wonky

WittierDinosaur (approved):
LGTM

test/fixtures/dialects/exasol/AlterTableDistributePartition.yml
alanmcruickshank: Agreed. But totally separate to this change 😄

PR #4671, opened 2023-04-05
Fix bug in templated with clauses LT07

alanmcruickshank:
Did some linting of our wider project and found this bug. Took the opportunity to rationalise how LT07 works more generally. I think it was much more convoluted than it needed to be.

Includes a test case to demonstrate the issue which was originally unexpectedly making this rule flag up errors.


WittierDinosaur:
Code looks reasonable - but linting & mypy errors

WittierDinosaur (approved):
LGTM

PR #4677, opened 2023-04-05
Make version number an argument not an option in release script.

alanmcruickshank:
For the release script I always have to use `--help` to remember the command and the option name. The script doesn't really work _without using the option_ however - so it seems more sensible to make it an argument so that it's always required. I think the command is also too long. This shortens things:
```bash
# from...
python util.py prepare-release --new_version_num=1.1.1
# ...to...
python util.py release 1.1.1
```

I've also updated the CI script, but I'm hoping this is easier for me to remember when kicking this off manually in future.

### DIFF
--- a/.github/workflows/create-release-pull-request.yaml
+++ b/.github/workflows/create-release-pull-request.yaml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Prepare release
         run: |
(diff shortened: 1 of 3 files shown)


greg-finley (approved):
Nice. Feel free to liberally change this up to meet your needs. 🙂

PR #4685, opened 2023-04-06
Safety valve for fixes in CV03

### LINKED ISSUE #4681: CV03 corrupts SQL when applied to looped sql.
### Search before asking

### What Happened

User @iKwesi found that running `fix` on a statement where the final select target in a loop and contains a trailing comma results in a fix which corrupts the SQL.

This minimal query demonstrates:
```sql
SELECT
    {% for col in ['a', 'b', 'c'] %}
        {{col}},
    {% endfor %}
FROM tbl
```

Results in:
```
> sqlfluff fix test.sql -d bigquery -r CV03 -x f
== [test.sql] FAIL
L:   3 | P:  16 | CV03 | Trailing comma in select statement forbidden
                       | [convention.select_trailing_comma]
==== fixing violations ====
```

After applying the fix we get:

```sql
SELECT
    {% for col in ['a', 'b', 'c'] %}
        {{col}}{% endfor %}
FROM tbl
```

...which is no longer valid SQL.

### Expected Behaviour

In this case, the rule should still flag the issue - but _it should not attempt the fix_, because it cannot uniquely identify the final comma.

### Observed Behaviour

Fix corrupts SQL.

### How to reproduce

See above.

### Dialect

all - but only really relevant for dialects which allow this (e.g. bigquery).

### Version

Confirmed on 2.0.3

### Configuration

occurs on default config

### Code of Conduct

alanmcruickshank:
This resolves #4681 . When a comma is in a loop - it's not safe to remove. This prevents that.

### DIFF
--- a/src/sqlfluff/rules/convention/CV03.py
+++ b/src/sqlfluff/rules/convention/CV03.py
@@ -79,8 +79,8 @@ class Rule_CV03(BaseRule):
                                     # Not safe to fix
                                     self.logger.info(
                                         "Preventing deletion of %s, because source "
-                                        "position is the same as %s. Templated positions "
-                                        "are %s and %s.",
+                                        "position is the same as %s. Templated "
+                                        "positions are %s and %s.",
                                         last_content,
                                         seg,
                                         last_content.pos_marker.templated_position(),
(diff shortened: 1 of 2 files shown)


greg-finley:
Check CI

greg-finley (approved):

PR #4695, opened 2023-04-08
Rethink Rule Docs

alanmcruickshank:
We've hacked around `autodoc` for a while now for the rule docs, and it's started to get quite hard. This takes a more robust approach to rule documentation.

The main rules page now has a table, for users to look up rules (grouped by bundle):

image

Each bundle now gets it's own page, and the rules appear in the table of contents by name, but can be referred to by _code and name_:

image

It's not perfect, and I think there's a lot more we could do with this but I think it's a big step forward.

Technically:
- The `conf.py` file loads the sqlfluff plugin host and processes the rules to generate the `rst` files. These are loaded into a folder which is `.gitignored` so that it's not committed but is then imported by the rest of the project.
- The additional `sqlfluff` domain plugin for sphinx handles the cross referencing and in the indices.


noel (commented):
This is indeed a huge improvement. One thing to note is that this change will break the link to the docs in the vs code plugin. We will need to coordinate with @RobertOstermann to see what it would take to support the new urls.

alanmcruickshank:
> This is indeed a huge improvement. One thing to note is that this change will break the link to the docs in the vs code plugin. We will need to coordinate with @RobertOstermann to see what it would take to support the new urls.

Good catch! Do you know what format the links in the VScode plugin are? I might be able to make this change still support that link format. If you (or @RobertOstermann ) is able to find a couple of examples I should be able to set them up for testing.

noel:
@alanmcruickshank this is the pattern being used


https://github.com/sqlfluff/vscode-sqlfluff/blob/master/src/features/commands/showDocumentation.ts

docs/source/conf.py
```diff
@@ -26,7 +26,7 @@
 # -- Project information -----------------------------------------------------
 
 project = "SQLFluff"
-copyright = "2019, Alan Cruickshank"
+copyright = "2023, Alan Cruickshank"
```
barrywhart, line 29: Would it make sense to always use the current year?

barrywhart (approved):
Looks good to me!

I noticed we have ads on the rule page -- when was that added? Any updates on sponsorship?

docs/source/conf.py
alanmcruickshank: Probably - Seemed more legit to update it manually rather than have it automated in some way.

alanmcruickshank:
> Looks good to me!
> 
> I noticed we have ads on the rule page -- when was that added? Any updates on sponsorship?

Ads? I can't see any - can you screenshot them?

alanmcruickshank:
> @alanmcruickshank this is the pattern being used
> 
> https://github.com/sqlfluff/vscode-sqlfluff/blob/master/src/features/commands/showDocumentation.ts

It's going to be difficult to support the old links fully. I'll put together a PR to support the new links for the plugin and add a redirect so the old links _work_, they just won't be quite as specific.

noel:
> > Looks good to me!
> > I noticed we have ads on the rule page -- when was that added? Any updates on sponsorship?
> 
> Ads? I can't see any - can you screenshot them?

image

noel:
> > @alanmcruickshank this is the pattern being used
> > https://github.com/sqlfluff/vscode-sqlfluff/blob/master/src/features/commands/showDocumentation.ts
> 
> It's going to be difficult to support the old links fully. I'll put together a PR to support the new links for the plugin and add a redirect so the old links _work_, they just won't be quite as specific.

I see, because the new links are just taking you to the group?
Can we add anchors on each rule then we can support Rule_AL03?  Even if we are not backward compatible it is a nice experience to go right to the specific rule

barrywhart:
@alanmcruickshank: Noel's screenshot matches what I'm seeing. It's near the bottom of the page, after the docs but before some GitHub stuff.

alanmcruickshank:
> > > @alanmcruickshank this is the pattern being used
> > > https://github.com/sqlfluff/vscode-sqlfluff/blob/master/src/features/commands/showDocumentation.ts
> > 
> > 
> > It's going to be difficult to support the old links fully. I'll put together a PR to support the new links for the plugin and add a redirect so the old links _work_, they just won't be quite as specific.
> 
> I see, because the new links are just taking you to the group? Can we add anchors on each rule then we can support Rule_AL03? Even if we are not backward compatible it is a nice experience to go right to the specific rule

I've managed to get the original rule anchors to work so that we don't need to change the VSCode plugin - everything is also on one page as we discussed via slack today 👍

PR #2566, opened 2022-02-05
Capitalization rules (L010, L014, L030, L040) should ignore templated code

barrywhart:
### Brief summary of the change made
Fixes #2565

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

Issue #4641, opened 2023-03-30
sqlfluff is not testing sql code added by macros (dbt-databricks)


mvdwielen:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Installed sqlfluff:
pip install dbt-databricks sqlfluff-templater-dbt

Added a .sqlfluff file to root of dbt project folder:
![image](https://user-images.githubusercontent.com/62689682/228937363-a1348990-9e9c-44bf-9923-137d91cd32ff.png)

Added a .sqlfluffignore file to root of dbt project folder:
![image](https://user-images.githubusercontent.com/62689682/228937517-4c2b12f8-971b-4b8c-b035-735801e62260.png)

Added a test.sql file to the models folder:
![image](https://user-images.githubusercontent.com/62689682/228937639-004dea1b-7ba8-436e-872f-712887285a97.png)

Added a test.sql file to the macro folder:
![image](https://user-images.githubusercontent.com/62689682/228937754-7f871ee3-b6e0-469b-84d5-347bb5f13650.png)

Run sqlfluff lint:
sqlfluff lint models\test.sql

The compiled model contains the following SQL:
![image](https://user-images.githubusercontent.com/62689682/228938011-af1a408d-2456-4a2b-ae7e-62421e1eb2e1.png)

sqlfluff only reports a rule failed on the first line of the test.sql model while the rule should fail for line 1 as well as line 4
![image](https://user-images.githubusercontent.com/62689682/228938087-c2f7a050-3302-4d8a-96a8-9147af29fe66.png)

It seems that sqlfluff is not testing the SQL code added by macro's. I have tested this for different models and macros but it only reports failed rules on SQL which is in the model file but it ignores SQL added by macros.

### Expected Behaviour

sqlfluff should test all sql of the compiled model (so also the sql added by macros) and not only the sql in the model file itself.

### Observed Behaviour

sqlfluff only tests sql which is in the model file

### How to reproduce

see steps in 'what happened'

### Dialect

databricks

### Version

sqlfluff, version 2.0.2

Python 3.10.0

Name: sqlfluff-templater-dbt
Version: 2.0.2
Summary: Lint your dbt project SQL
Home-page: https://github.com/sqlfluff/sqlfluff
Author: Alan Cruickshank
Author-email: alan@designingoverload.com
License: MIT License
[...]

PR #4697, opened 2023-04-09
Capitalization rules ignore templated code only if configured to

james-johnston-thumbtack:
The pull request at https://github.com/sqlfluff/sqlfluff/pull/2566 made sense for users who set ignore_templated_areas to true because those macros are not under their control.  However, if a user is writing their own macros and would like to lint those, that commit made the capitalization rules completely unusable.

For example, the following code in a single file would not trigger any violations, even if ignore_templated_areas is set to false:

    {% macro abc() %}
        sElEcT a, B fRoM abc;
    {% endmacro %}
    {{ abc() }}

That is... a surprising result...

This commit will now only skip templated areas if we are actually configured to do that.  Hopefully this will allow their use in templated areas if desired, while still avoiding false positives when we ignore template areas as the previous commit had aimed to fix.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/capitalisation/CP01.py
+++ b/src/sqlfluff/rules/capitalisation/CP01.py
@@ -86,10 +86,11 @@ class Rule_CP01(BaseRule):
         if parent.get_type() == "function_name" and len(parent.segments) != 1:
             return [LintResult(memory=context.memory)]
 
-        return [self._handle_segment(context.segment, context.memory)]
+        return [self._handle_segment(context.segment, context)]
 
-    def _handle_segment(self, segment, memory) -> LintResult:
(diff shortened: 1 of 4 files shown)


james-johnston-thumbtack:
This one may also fix the latest capitalization issues reported by the user for #4641:

> When I deliberately upper case a quoted identifier in the model.sql then rule CP02 | Unquoted identifiers must be lower case will fail, when I upper case a quoted identifier in the SQL added by the macro the rule will not fail.

james-johnston-thumbtack:
linting error fixed on this one now, too

alanmcruickshank (approved):
Nice. 👍

PR #4698, opened 2023-04-09
Fix LT01 spacing check in templated areas

james-johnston-thumbtack:
When both the LT01 rule is on, and ignore_templated_areas is False, the LT01 rule yields a large number of false positives.  For example, a template like this:

    {{ 'SELECT 1;' }}

would fail with a false positive error like this, if you run with ignore_templated_areas set to false:

    L:   1 | P:   1 | LT01 | Expected single whitespace between 'SELECT'
        keyword and numeric literal. [layout.spacing]

In respace.py, the original "if" block that bypasses template segments was added recently in:

https://github.com/sqlfluff/sqlfluff/commit/0b1f2aab2ac5eceeae5c38b049d3b8e40c338ab7

That commit was mainly focused on indentation.  I'm not sure why the change in rescape.py was made... but deleting it doesn't cause any test cases to fail, and it helps my new test cases pass.

This probably makes progress on issue #4641, I am not sure.  It definitely makes progress on my own test cases.

This code is very hairy and unfamiliar to me, and I'm relying on existing test cases to ensure I'm not breaking something.  I could very easily be doing something wrong here...

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/utils/reflow/respace.py
+++ b/src/sqlfluff/utils/reflow/respace.py
@@ -113,14 +113,8 @@ def process_spacing(
 
     # Loop through the existing segments looking for spacing.
     for seg in segment_buffer:
-        # If it has a position marker, but it's not literal, then
-        # it's a templated element and so we shouldn't consider it
-        # here.
-        if seg.pos_marker and not seg.pos_marker.is_literal():
-            continue
-
         # If it's whitespace, store it.
-        elif seg.is_type("whitespace"):
+        if seg.is_type("whitespace"):
             last_whitespace.append(seg)
 
         # If it's a newline, react accordingly.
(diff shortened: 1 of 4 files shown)


alanmcruickshank (changes requested):
@james-johnston-thumbtack  - thanks for putting this together - and also for delving into the reflow code! I'm glad it's not just me in there 😄 .

Would you be able to add one or two test cases which have a mix to templated and untemplated code, which also set `ignore_templated_areas` to `true` to make sure that the correct filtering is happening in the default case? I worry that this code is not set up well to handle that case, but I'm hoping that I'm wrong. In any case, an additional test case which explicitly covers that scenario would probably be enough.

james-johnston-thumbtack:
@alanmcruickshank good call on testing a mixture.  Do these added test cases cover what you're looking for?

alanmcruickshank (approved):
Yep 👍

PR #4702, opened 2023-04-09
Suggested started config file

alanmcruickshank:
This comes from a discussion with @noel .

As suggested:
- This adds a default minimal config for a starter project (@WittierDinosaur, hopefully you agree with the settings).
- This also adds a note above the full default config, suggesting users don't copy it.
- It adds a section further up in the config page to explain what to do instead.


docs/source/configuration.rst
```diff
@@ -80,10 +83,61 @@ For example, a snippet from a :code:`pyproject.toml` file:
...
+    # While implicit indents are not enabled by default. Many of the
+    # SQLFluff maintainers do use them in their projects.
+    allow_implicit_indents = true
+
+    # The default configuration for capitalisation rules is "consistent"
+    # which will auto-detect the setting from the rest of the file. This
+    # is less desirable in a new project and you may find this (slightly
+    # more strict) setting more useful. 
+    [sqlfluff:rules:capitalisation.keywords]
+    capitalisation_policy = lower
```
WittierDinosaur, line 126: Do you think everything being `lower` is the most common way? I'd think upper for keywords and types is more common (at least everywhere i've worked).

noel: I think it lower is indexed preferred because these days we have syntax highlighting with pretty colors so it’s not strictly necessary. 

https://stackoverflow.com/questions/608196/why-should-i-capitalize-my-sql-keywords-is-there-a-good-reason

noel:
Should we encourage a new user to enable all rules? Are there any we should exclude?

noel (commented):
I am wondering if this is “too minimal“
Are there any other defaults we should override?
This was my attempt, but maybe I have too much

https://github.com/datacoves/balboa/blob/main/transform/.sqlfluff

alanmcruickshank:
@noel @WittierDinosaur - I've tried to walk a midline between both of your suggestions. Adding some extra links and bringing in some more from the [balboa repo](https://github.com/datacoves/balboa/blob/main/transform/.sqlfluff). It's not everything but I think it's a good starter set.

Would love views on whether we think this is close enough to "best practice" that it's good to go.

WittierDinosaur (approved):
Begrudging tick. It's a good baseline, even if i disagree on the caps

noel:
I like it. better UX for newbies

PR #4703, opened 2023-04-09
Don't do newline conversion on write

alanmcruickshank:
I ran into an issue while linting a large chunk of our project where some windows encoded files (using `\r\n`) would end up with doubled newlines. This was because the newlines were being converted on write. This ensures no conversion is done.

I'm not sure how to get a good test case around this - any suggestions welcome. Otherwise I'll suggest merging this without a test case.


WittierDinosaur (approved):
LGTM

PR #4706, opened 2023-04-09
Indent Aliases

alanmcruickshank:
I'm having real trouble getting #4560 to do what I want it to, so I've separated this part off it - and I'm hoping that I might be able to get it to a mergable state first and then come back to window functions.

This also fixes a long standing bug in respace logic where we couldn't effectively detect the start and end of rebreak spans because of non-code characters. This also resolves _part_ of #4033 .


src/sqlfluff/utils/reflow/elements.py
```diff
@@ -662,7 +671,12 @@ def respace_point(
             if last_whitespace:
                 ws_idx = self.segments.index(last_whitespace)
                 if ws_idx > 0:
-                    prev_seg = self.segments[ws_idx - 1]
+                    for prev_seg in self.segments[ws_idx - 1 :: -1]:
```
barrywhart, line 674: Performance: Should we consider looping with a range of indicates rather than actually slicing the list?

src/sqlfluff/utils/reflow/elements.py
```diff
@@ -662,7 +671,12 @@ def respace_point(
...
+                        # Skip past any indents
+                        if prev_seg.is_type("indent"):
+                            continue
```
barrywhart, line 677: Nit: The `continue` / `break` together reads a little strange. No big deal, though.

barrywhart (approved):

Issue #4624, opened 2023-03-27
JinjaAnalyzer: Provide better support for custom Jinja tags / statements provided by custom Jinja extensions


james-johnston-thumbtack:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

I would like to be able to more easily load a custom Jinja extension, perhaps as part of a custom templater plugin.  Most importantly, I'd like to be able to instruct Jinja how to parse and trace the new custom Jinja tags that this extension introduces.

An example of what the Jinja file I am trying to parse looks like.  This is from a custom system (not my creation) for running database migrations.

```
{% up 'add a column to test_table' %}
ALTER TABLE test_table ADD COLUMN test_column_two bigint;
{% down %}
ALTER TABLE test_table DROP COLUMN test_column_two;
{% end %}
```

This file contains custom Jinja tags: "up", "down", and "end".  A file could contain multiple migrations in the form of multiple up/down/end blocks.

The solution I have come up with so far involves creating a simple templater plugin that looks like this:
```python
from __future__ import annotations

from typing import cast

import jinja2.parser
from jinja2 import Environment, nodes
from jinja2.ext import Extension
from jinja2.sandbox import SandboxedEnvironment
from sqlfluff.core import FluffConfig
from sqlfluff.core.templaters.jinja import JinjaTemplater


# Jinja extension.  In summary, the parsing step will follow through to the "down" and "end" tags.
# When rendering, both the up & down migrations are rendered.
class DBMigrationExtension(Extension):
    SPLIT_STRING = "============== DBMigrationExtension Splitter =============="
    tags = {"up"}

    def __init__(self, environment: jinja2.environment.Environment):
        super().__init__(environment)
        # note: the actual migration system doesn't actually look at rendered template output.
        # instead, it's using this extension to capture the migrations into a trace.
        environment.extend(migrations=[])

    def parse(self, parser: jinja2.parser.Parser) -> jinja2.nodes.Node | list[jinja2.nodes.Node]:
        lineno = next(parser.stream).lineno

        # Parse the title of this migration
        args = [parser.parse_expression()]

        up_body = parser.parse_statements(("name:down",), drop_needle=True)
        sep = [nodes.Output([nodes.TemplateData(self.SPLIT_STRING)])]
        down_body = parser.parse_statements(("name:end",), drop_needle=True)

        rv = nodes.CallBlock(
[...]

PR #4708, opened 2023-04-10
Preserve zero-length template segments

james-johnston-thumbtack:
I am making a custom templater that will intentionally yield zero length slices with a special slice_type.  This is going to be used to signal to some rules information about special markers in the templated result. Unfortunately, SQLFluff strips these right out unless they are prefixed with "block".  Thus, I had to implement this fragile and hacky workaround from my templater when emitting slices:

```python
[
    TemplatedFileSlice("block_start", ...),
    TemplatedFileSlice("block_my_custom_slice_type", ...),
    TemplatedFileSlice("block_end", ...),
]
```

Prefixing the slice type with "block" was necessary to trick this line of code in lexer.py in _handle_zero_length_slice:

    if tfs.slice_type.startswith("block"):
        <snip>

The next problem was that the yield statement was failing:

    yield TemplateSegment.from_slice(...block_uuid=block_stack.top())

Since the block_stack was empty, this would crash.  Thus the final hack: first emit some fake block start/end slices to put something on the block stack, and then the custom "block" slice type in the middle.

This all seems very hacky and fragile, so I'd rather just pass through these segments as-is without going down the "block" branch of code.

This commit removes the "if" condition that excluded these segments. A new test function is also introduced that allows us to continue to add more test cases where we want to specify an exact TemplatedFile input to the lexer, rather than depending on a particular templater implementation in the test.

Hopefully not?  All tests pass...

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/parser/lexer.py
+++ b/src/sqlfluff/core/parser/lexer.py
@@ -434,16 +434,15 @@ def _handle_zero_length_slice(
         # Move on
         return
 
-    # We've got a zero slice. This could be a block, unrendered templates
-    # or unrendered code (either because of loops of consumption).
-    if not is_zero_slice(tfs.source_slice):
-        yield TemplateSegment.from_slice(
-            tfs.source_slice,
-            tfs.templated_slice,
-            tfs.slice_type,
-            templated_file,
-        )
-    return
+    # Always return the slice, even if the source slice was also zero length.  Some
+    # templaters might want to pass through totally zero length slices as a way of
+    # marking locations in the middle of templated output.
+    yield TemplateSegment.from_slice(
+        tfs.source_slice,
+        tfs.templated_slice,
+        tfs.slice_type,
+        templated_file,
+    )
 
 
 def _iter_segments(
--- a/test/core/parser/lexer_test.py
+++ b/test/core/parser/lexer_test.py
@@ -2,8 +2,10 @@
 
 import pytest
 import logging
+from typing import Any, Dict, Tuple, NamedTuple, List, Union
 
 from sqlfluff.core.parser import Lexer, CodeSegment, NewlineSegment
+from sqlfluff.core.parser.segments.meta import TemplateSegment
 from sqlfluff.core.templaters import JinjaTemplater
 from sqlfluff.core.parser.lexer import (
     StringLexer,
@@ -172,6 +174,19 @@ def test__parser__lexer_trim_post_subdivide(caplog):
         assert len(res.elements) == 3
 
 
+class _LexerSlicingCase(NamedTuple):
+    name: str
+    in_str: str
+    context: Dict[str, Any]
+    # (
+    #     raw,
+    #     source_str (if TemplateSegment),
+    #     block_type (if TemplateSegment),
+    #     segment_type
+    # )
+    expected_segments: List[Tuple[str, Union[str, None], Union[str, None], str]]
+
+
 def _statement(*args, **kwargs):
     return ""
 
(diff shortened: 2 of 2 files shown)


alanmcruickshank (approved):
Thanks for this - this looks like a good step forward.

I'd be curious for some additional test cases which more explicitly support what your custom templater is planning on relying on - that was we can make sure it doesn't get broken in a future release.

james-johnston-thumbtack:
@alanmcruickshank yes, I suppose I left some context out!

This is a continuation of the work in https://github.com/sqlfluff/sqlfluff/issues/4624

Basically, my templater derives from JinjaTemplater and:

- Load a custom extension in the Jinja environment for parsing directives related to running database migrations.
- The extension has special Jinja statements / directives supported for _optionally_ outputting some tracing information when they are encountered.  For example, when trace mode is on, `{% down %}` renders as `__db_migration_extension_down_migration`, else it renders as the empty string.
- In order to avoid screwing up the template slicing and subsequent parsing, the JinjaTemplater.slice_file function is overridden, and the `template_builder` / `make_template` functions are patched/overridden so that we can change the Jinja environment configuration.  The [first rendering](https://github.com/sqlfluff/sqlfluff/blob/2460e2a6e4476a218f6cd3d2fc3c7a0c77ebefa5/src/sqlfluff/core/templaters/slicers/tracer.py#L70-L77) done by JinjaAnalyzer with the "alternate codes" does _not_ get the tracing strings from my extension, so as to avoid messing up the parsing any further, and let it generate slices as normal.  The [second rendering](https://github.com/sqlfluff/sqlfluff/blob/2460e2a6e4476a218f6cd3d2fc3c7a0c77ebefa5/src/sqlfluff/core/templaters/slicers/tracer.py#L124), however, is traced.
- After getting that rendered output, `slice_file` then searches it for the custom markers that my extension produced.  These markers are removed from the string (again to avoid messing up the SQLFluff lexer/parser).  However, we record the positions they had and insert them into the list of TemplatedFileSlice in the form of a zero-length slice with a custom slice type.  This seems reasonably clean, except for the part where I had to trick the Jinja lexer to not discard them... that last bit is what motivated this CR... maybe it was one hack too far... :)

I think the trickiest parts most likely to break here are that it makes some assumptions about the number / order of template renderings that take place, and the fact that nothing of consequence has happened with the second rendering before I can modify it and remove the marker strings.

However, none of this is as fragile IMHO as the issues originally described in https://github.com/sqlfluff/sqlfluff/issues/4624 which details the workarounds I took to get the directives parsing reasonably in the first place.

The end result (just where am I going with all this?) is now I can lint the structure of my directives deep inside of a macro.  For example:

```
{% macro set_not_null(schema, table, column) %}
    {% up 'Set {}.{}.{} as NOT NULL'.format(schema, table, column) %}
        BEGIN;
        ALTER TABLE {{ schema }}.{{ table }} ALTER COLUMN {{ column }} SET NOT NULL;
    {% down %}
        ALTER TABLE {{ schema }}.{{ table }} ALTER COLUMN {{ column }} DROP NOT NULL;
        COMMIT;
    {% end %}
{% endmacro %}

{{ set_not_null('public', 'foo', 'baz') }}
```
With this technique, I can have custom lint rules that do things like:
```
L:  11 | P:  43 | TTMigrate_T007 | Multi-statement migration does not end with COMMIT.
                       | [migration.no_transaction]
L:  11 | P:  43 | TTMigrate_T007 | Multi-statement migration does not start with START
                       | TRANSACTION. [migration.no_transaction]
L:  11 | P:  43 | TTMigrate_T011 | Transaction in migration was not finished.
                       | [migration.unfinished_migration_transaction]
```
These rules examine the custom meta placeholders that the templater originated via the slice, and can even look deep inside the macro, included files, etc.  It's still not perfect since as you can see, the line numbers don't show the original offending line (which could even be in a different file).  But good enough for our purposes.

PR #4716, opened 2023-04-11
Add support for PG CREATE/ALTER/DROP PUBLICATION stmts

james-johnston-thumbtack:
This commit adds support for the CREATE/ALTER/DROP PUBLICATION statements in PostgreSQL 15.  It someone closely-ish follows the upstream PG grammar in gram.y.  Notable deviations include:

1.  The upstream grammar has the concept of continuation objects in the CREATE statement, where "b" is a continuation statement in "TABLE a, b, TABLE c".  It's assumed that "b" is a table by looking at the previous object, "TABLE a".

    I think it will be easier to use this tree if we instead group the related objects under a parent PublicationObjectsSegment node.

2.  The ATLER PUBLICATION ... OWNER TO statement is a bit messy in that I reproduced the CURRENT_ROLE, CURRENT_USER, SESSION_USER as is seen elsewhere in this file.  However, the upstream grammar defines a RoleSpec which is used in many different places.

    An area of future improvement could be to refactor this grammar and use a shared role spec everywhere similar to what the upstream grammar does.  In fact, I think these keywords are in the ANSI standard, so it really should be an improvement made to the ANSI dialect...

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -258,6 +258,12 @@ postgres_dialect.add(
         Ref("NakedIdentifierFullSegment"),
     ),
     CascadeRestrictGrammar=OneOf("CASCADE", "RESTRICT"),
+    ExtendedTableReferenceGrammar=OneOf(
+        Ref("TableReferenceSegment"),
+        Sequence("ONLY", Ref("TableReferenceSegment")),
+        Sequence("ONLY", Bracketed(Ref("TableReferenceSegment"))),
+        Sequence(Ref("TableReferenceSegment"), Ref("StarSegment")),
+    ),
     RightArrowSegment=StringParser("=>", SymbolSegment, type="right_arrow"),
 )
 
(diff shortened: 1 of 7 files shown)


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -258,6 +258,12 @@
         Ref("NakedIdentifierFullSegment"),
     ),
     CascadeRestrictGrammar=OneOf("CASCADE", "RESTRICT"),
+    ExtendedTableReferenceGrammar=OneOf(
+        Ref("TableReferenceSegment"),
+        Sequence("ONLY", Ref("TableReferenceSegment")),
+        Sequence("ONLY", Bracketed(Ref("TableReferenceSegment"))),
```
WittierDinosaur, line 264: These 2 lines can be condensed to `Sequence("ONLY", OptionallyBracketed(Ref("TableReferenceSegment"))),`

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -736,6 +742,39 @@ class ArrayTypeSegment(ansi.ArrayTypeSegment):
...
+class DefinitionParametersSegment(BaseSegment):
+    """List of definition parameters.
+
+    https://github.com/postgres/postgres/blob/4380c2509d51febad34e1fac0cfaeb98aaa716c5/src/backend/parser/gram.y#L6313
+    """
+
+    type = "definition_parameters"
+    match_grammar: Matchable = Bracketed(
+        Delimited(
+            Sequence(
```
WittierDinosaur, line 771: This Sequence isn't needed

james-johnston-thumbtack:
I just realized the DROP EXTENSION I copied from should probably use DropBehaviorGrammar instead of CascadeRestrictGrammar...

WittierDinosaur (approved):
LGTM

PR #4717, opened 2023-04-11
Add ExpressionSegment to CREATE TABLE ... DEFAULT / Fix multiple parse issues in Expression_A_Grammar

### LINKED ISSUE #4552: SQLite unable to parse CREATE TABLE .. DEFAULT with parentheses
### Search before asking

### What Happened

When parsing a CREATE TABLE expression which has a column definition including a DEFAULT statement, if the default value is bracketed in parentheses, the statement is determined to be unparsable by sqlfluff.

Parentheses are optional for default literal values, but required for defaults that use functions like `datetime()` or `randomblob()`.

```
sqlite> CREATE TABLE foo (id TEXT DEFAULT datetime('now'));
Error: in prepare, near "(": syntax error (1)
sqlite> CREATE TABLE foo (id TEXT DEFAULT (datetime('now')));
sqlite> -- succeeded
```

### Expected Behaviour

```
All Finished 📜 🎉!                                                                                               
```

### Observed Behaviour

```
== [stdin] FAIL                                                                                                                                  
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: "CREATE
                       | TABLE foo (id TEXT DEFAULT (''))"
```

### How to reproduce

```
echo "CREATE TABLE foo (id TEXT DEFAULT (''));" | sqlfluff lint --dialect sqlite - 
```

Removing the parentheses (changing `('')` to `''`) restores the expected behavior.

### Dialect

sqlite

### Version

Both sqlfluff version 2.0.0 and 1.4.5 are affected.
Python version: 3.11.2

### Configuration

No configuration

### Code of Conduct

james-johnston-thumbtack:
**Add ExpressionSegment to CREATE TABLE ... DEFAULT**

When creating a table with a DEFAULT in PostgreSQL, you can specify an expression for the default value.  It doesn't even have to be in parenthesis.

However, the existing grammar doesn't allow for these cases.  This commit adds ExpressionSegment to the end of the existing grammar, which is similar to what exists for ALTER TABLE.

This would fix #4552.

**Fix multiple parse issues in Expression_A_Grammar**

This changes the grammar to look more like the original Cockroach Labs diagram from the link.  This fixes a few test cases that the original code struggled with, as well as some totally inaccurate outputs it was yielding.

This work was necessary in order to unblock fixing the CREATE TABLE ... DEFAULT constraint.

More information about the test cases can be found in the commit message.

This commit also increases the recursion limit... I saw from an earlier PR that there was some controversy there... my personal take is maybe it makes sense to prioritize correctness over performance for now until e.g. the parser could handle recursive grammars better (I don't think they are "bad" or "buggy"?)... but if there are some good suggestions to keep the grammar non-recursive while keeping the new test cases working... we could alternatively try that...

I put a more detailed discussion of this in the commit message as well, including some observations I had taking some measurements.

### Pull Request checklist

  - Other.


WittierDinosaur:
I've done some perf testing, and I'm happy this doesn't significantly impact performance. I'm not overly keen on adding the recursion hack in. We should either set a static one, or not, but we shouldn't hack it like this. @alanmcruickshank what do you think about this?

alanmcruickshank:
> I've done some perf testing, and I'm happy this doesn't significantly impact performance. I'm not overly keen on adding the recursion hack in. We should either set a static one, or not, but we shouldn't hack it like this. @alanmcruickshank what do you think about this?

I think I agree on the recursion hack. Given the relative simplicity of the queries we're trying to parse, I think it's the sign of an overcomplicated solution if we need to increase the recursion limit.

I haven't been able to fully review the whole PR here yet, but would I be accurate in saying we just want to enable multiple consecutive sign indicators before numeric literals? Or have I missed something else?

james-johnston-thumbtack:
@alanmcruickshank the commit message for https://github.com/sqlfluff/sqlfluff/pull/4717/commits/511affb16453fe3d425a17143e1caa5bbb9903a9 enumerates the different test cases which did not do the right thing.  Some of them like the consecutive sign indicators have parse errors.  But it is not just that.  Others were "succeeding" to parse but resulting in totally non-sensical results, like successfully parsing `FALSE NOT NULL`, or parsing the middle `NOT` in `FALSE AND NOT NOT NOT (TRUE)` as a column reference (oops).

The `FALSE NOT NULL` "successful" parse is what started me on this whole rabbit hole.  All I wanted to do was enable expressions for `CREATE TABLE` with `DEFAULT` constraint.  Sadly this meant that the test case that had `CREATE TABLE ... DEFAULT FALSE NOT NULL` started to fail because it only parsed one constraint instead of two.... Then after looking at the grammar I realized there are multiple other issues as well as noted earlier.

Also I just now discovered that the codebase I am parsing (which has several years worth of migrations) uses the tilde operator (for some reason some people here decided long ago that making lots of use of bit fields in an integer would be a great idea).  I noticed that is commented out for some reason, and so now there are parse errors on that, too.  So I will also have to come back here and try to turn it on.... I am guessing my chances of success here will be higher if it's more closely following the Cockroach parser. (Was it commented out to work around an issue resulting from deviations from the Cockroach parser?)

Also, maybe worth noting that the a_expr grammar might continue to grow for some dialects, for example note that the [latest version](https://www.cockroachlabs.com/docs/dev/sql-grammar.html#a_expr) of the Cockroach DB adds a `qual_op`, and a `row OVERLAPS row`.  Might be easier to keep this grammar updated if it aligns better with the grammar we're following?

> We should either set a static one, or not, but we shouldn't hack it like this

Another idea I thought of might be if we automatically scale the Python recursion limit proportional to the configured `recurse` SQLFluff configuration option.  The default `recurse` is `0` (unlimited), but that is clearly a lie: there's still a limit.  We could deprecate the unlimited option (since there's no such thing), and set this configuration option to some value large enough for common use cases.  Then set the Python recursion limit to the value of `recurse` times some constant value that accounts for the number of typical stack frames for each parse recursion level.

I kind of held back on proposing a configuration approach initially because it seems a bit out of scope of this PR, but it could be a good follow-up if we were otherwise open to having recursion?

james-johnston-thumbtack:
I added another commit to the branch that adds support for the tilde (~) bitwise operator, along with some basic test cases for it.  This gets another group of SQL files from my codebase to parse without error.

Another issue I found is that LT01 wants to join two unary operators together, e.g. `SELECT 3 * -n;` --> `SELECT 3 *-n` (or `SELECT 3 | ~n` --> `SELECT 3 |~n`).  Without spacing, PG considers that to be a different (and invalid) operator.  So, we probably need to track that for follow up (somehow the `spacing_after` config option needs to be constrained to apply to either include or exclude certain neighboring segments).  However, given that the current status quo is that the parser dies with a parse error on those inputs I just gave, I think it shouldn't be a blocking issue for merging this one since it's still incremental improvement.

james-johnston-thumbtack:
The next commit fixes what is arguably a regression in the first branch commit: the fact that numeric_literal no longer grouped together the sign with a signed number.  From the latest commit message:

--------
**SignedSegmentGrammar: only match one sign / preserve QualifiedNumericLiteralSegment in expressions**

Numeric literals only have up to one sign, not an unlimited number:

 - http://web.cecs.pdx.edu/~len/sql1999.pdf -- defines:
```
     <signed numeric literal> ::= [ <sign> ] <unsigned numeric literal>
     <sign> ::= <plus sign> | <minus sign>
```

 - PostgreSQL has a similar rule:
     https://github.com/postgres/postgres/blob/4380c2509d51febad34e1fac0cfaeb98aaa716c5/src/backend/parser/gram.y#L16782-L16788

So it's a little unusual for us to lump several signs together in the
same "numeric_literal", which at that point requires doing parsing of
unary operators just to interpret the numeric_literal (e.g. a rule could
not do "int(num_lit.raw)" to get the integer value).

This commit changes the SignedSegmentGrammar so that it only matches one
sign, not many signs.  This means the QualifiedNumericLiteralSegment
will match with one sign only.

This was required for the second part of this commit: the expression
parser will now no longer match on a unary + or - operator if it can
also be interpreted as a QualifiedNumericLiteralSegment.  This fixes a
regression from the earlier commit in this branch where we no longer
saw QualifiedNumericLiteralSegment in the parse trees.  While that was
still a valid parse tree, it's probably the case most people would
expect us to use QualifiedNumericLiteralSegment and group a single sign
with it, so we should do that.

The end result is that "SELECT 3 + - - +4;" now parses as:

    numeric_literal:                      '3'
    whitespace:                           ' '
    binary_operator:                      '+'
    whitespace:                           ' '
    sign_indicator:                       '-'
    whitespace:                           ' '
    sign_indicator:                       '-'
    whitespace:                           ' '
    numeric_literal:
        sign_indicator:                   '+'
        numeric_literal:                  '4'

james-johnston-thumbtack:
@alanmcruickshank @WittierDinosaur the last commit eliminates most recursion while keeping the new test cases in a healthy state!  I feel a lot better about this branch now.  What do you think?

The key insight is to refactor the beginning of a_expr that contains unary operators into its own grammar.  We then consider the concept of "tail recursion", which in this context just means that it's a recursive call to a_expr at the end of the matched sequence.  In that case we just replace those calls with the tail recursive grammar.

It's similar in concept to the baseline commit except that the unary operators now are in their own separate reusable grammar, which eliminates the edge cases when those elements were not correctly repeated before referencing Expression_C_Grammar in the baseline code on main.

Here's the latest commit message:

------

**Remove most expression recursion by using tail recursion**

The earlier commit in this branch fixed a lot of issues, but it had the
downside of re-introducing recursion, which SQLFluff is particularly
sensitive to since it makes recursive function calls.

This commit refactors the a_expr gramamr into smaller pieces.  In
particular, the recursive element necessary to support unary operators
has been moved into its own grammar.

The idea now is that whenever a part of the grammar _ends_ with a
single recursive call to a_expr, we instead call a special
"tail recursive" grammar that includes the non-repeating elements that
prefix the a_expr (e.g. unary operators).

If the recursive call is in the middle of the grammar, such as with

    a IN (b)

then you can't use this, because you can't have more things that match
after the reference to the tail recursion grammar.  (In this case, the
closing parenthesis prevents it.)

This commit is sufficient to avoid recursion errors on the test cases
that have long expressions, so it reverts the call to
sys.setrecursionlimit.

WittierDinosaur (approved):
I'm far more comfortable with this version of the PR. Great work!!

WittierDinosaur:
But alas - merge conflicts

james-johnston-thumbtack:
@WittierDinosaur merge conflict was only the checksum in one YAML file.  No conflicts in the rest of the YAML file.

james-johnston-thumbtack:
@WittierDinosaur merge conflicts resolved again, it was only YAML checksum conflicts again

PR #4723, opened 2023-04-12
Keep out zero length keywords

james-johnston-thumbtack:
Keep out the zero length keyword.  Their inclusion is due to the fact
that some keywords are specified in multiline strings, and the trailing
newline results in an empty string.

I think their inclusion is almost certainly accidental, so this code
strips the input string before we split it in order to avoid that
problem.

### Pull Request checklist
- [ ] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #4725, opened 2023-04-12
Improve support for EXCLUDE table constraints in PG

james-johnston-thumbtack:
The `EXCLUDE` constraint type doesn't really work and can't even parse simple examples.  There was zero test coverage.  Some of the problems with the existing grammar:

- It uses a `FunctionSegment` to refer to the index access method, instead of what should be just a simple identifier string.
- The `WHERE` clause is required for some reason.
- Other parts of the grammar were missing/incomplete.

I went through the upstream grammar file and updated this section of the parser accordingly.

Some common segment types are introduced which I will update other existing PG segments to reuse in a later commit.  We can make improvements to other areas where common segments can be used.

Most importantly, there are a few new test cases.  One of them exhaustively checks every option.  Congratuations, SQLFluff now parses this better than IntelliJ, the latter showing a syntax error on the test case!  (But PG15 does parse all examples without a syntax error.)

### Pull Request checklist

  - Other.


test/fixtures/dialects/postgres/postgres_alter_publication.yml
```diff
@@ -114,10 +114,10 @@ file:
             quoted_literal: "'insert,update'"
         - comma: ','
         - definition_parameter:
-            parameter: publish_via_partition_root
-            comparison_operator:
+          - parameter: publish_via_partition_root
+          - comparison_operator:
               raw_comparison_operator: '='
-            boolean_literal: 'TRUE'
+          - parameter: 'TRUE'
```
WittierDinosaur, line 120: Any easy way to remove the regression here?

test/fixtures/dialects/postgres/postgres_create_publication.yml
```diff
@@ -300,10 +300,10 @@ file:
             quoted_literal: "'insert,update'"
         - comma: ','
         - definition_parameter:
-            parameter: publish_via_partition_root
-            comparison_operator:
+          - parameter: publish_via_partition_root
+          - comparison_operator:
               raw_comparison_operator: '='
-            boolean_literal: 'TRUE'
+          - parameter: 'TRUE'
```
WittierDinosaur, line 306: This and below should also eval to boolean literal

james-johnston-thumbtack:
NOTE: here is a draft PR stacked on top of this one to expand use in other areas of the PG parser: https://github.com/sqlfluff/sqlfluff/pull/4726

test/fixtures/dialects/postgres/postgres_create_publication.yml
james-johnston-thumbtack: That will be a deviation from [how the PG parser](https://github.com/postgres/postgres/blob/4380c2509d51febad34e1fac0cfaeb98aaa716c5/src/backend/parser/gram.y#L6331-L6336) does things.  It might be a usability improvement to try to parse more literal types than what the PG parser does, but just wanted to double-check that this makes sense & that you're aware it's different from normal parsing done by PG?  And also would like to confirm what you want to do for other values than booleans.

If I parse: `alter table abc set (abc = testing, abc = null, abc = true, abc = 5, abc = 'hello', abc = 3.5);` using a utility [`pgsql-parser`](https://www.npmjs.com/package/pgsql-parser) that uses the real PG parser, I get the arg values parsing as: `"TypeName": { "names": [ { "String": { "str": "testing"`, and then a simpler `"arg": { "String": { "str": "null"` for the second one (because it was a reserved keyword), and the third boolean one as well, as well as the 5th string literal one.  The numbers get passed through as numbers.  Everything else is ultimately a string.  (I like the comment warning that "any simple identifier will be returned as a type name!" ... it smells like a bit of a hack to me with the comment... like they really wanted just a simple string for my first case as well...)

A good example of what the unquoted keyword / parameter name is for is like: https://www.postgresql.org/docs/current/sql-createtable.html --> "vacuum_index_cleanup" option.  It can be set as `(vacuum_index_cleanup = auto)`, for example.

So anyway, it's not really giving special treatment to booleans; it's just passing them through as strings.  Same for every other reserved keyword I guess.

I could try selecting a value with OneOf from either LiteralSegment or the ParameterSegment... but I want to confirm if that would yield the results you are looking for?

- `null` maps to null literal
- `true`/`false` maps to boolean literal
- number and string literals presumably continue to map as they are done in this commit (since after all that's what PG did, too)
- do you want reserved keywords to map to `keyword` or to `parameter`?  I started [this](https://github.com/sqlfluff/sqlfluff/pull/4723) to try to match with all keywords because that's what the upstream grammar does, but then I realized we can just be generic and probably just use `parameter`; I was going to abandon the main focus of that PR.  But now I am wondering if we should merge that other PR more or less as-is, and use it here to map reserved (& non-reserved???) keywords to "keyword"?
- what do you want other arbitrary unquoted strings to evaluate to? `(some_key = veryspecial)` -- do we continue to try to make a ParameterSegment out of it? is there another segment type that would be more appropriate? should I define a new segment type?

Let me know what you want to do for other data types as well, and confirm you still want to parse boolean literals, and I can make adjustments!

james-johnston-thumbtack: A few thoughts:

- Side thought: An interesting conflict would arise if an "enum" included "boolean" values.  For example, `vacuum_index_cleanup` can be `auto`, `off`, or `on`.  But imagine if they used true/false instead?  Then we'll force inconsistency within the enum.  Hopefully they don't do that and follow the on/off example..... you make a good case, especially since many params are actually booleans, so I will see about parsing booleans and nulls as their respective types.
- I poked around some more with pgsql-parser.  And since you prompted it, re-reviewed the capitalization rules.  I think CP02 (incorrect capitalization of unquoted identifiers) is the best one to apply to both parameter name and value.  What I found in the PG parse tree is that if it's unquoted, it'll always parse as a lowercase string.  But if you use double quotes, then it keeps the case.  Same as any other identifier like a table name.  Shouldn't be very surprising.
- Regarding keywords: using `vacuum_index_cleanup` as an example.  Each possible value is its own keyword categorization, so I think we probably want to enforce consistency within the enum values.  For example, I am guessing that we do not want to force lower-case `auto` while at the same time forcing upper-casing `OFF`, since those were all part of the same enum and `OFF` is not really acting as a keyword in the normal sense here anyway.
    - `auto` is not a keyword at all.
    - `off` is a non-reserved keyword.
    - `on` is a reserved keyword.
- I noticed that has a `properties_naked_identifier` that is also crawled.  It turns out it is only defined by Spark SQL dialect.  It is the same pattern as naked_identifier except that it also matches on keywords.

Here is what I'm currently thinking / would propose:

1.  Define `properties_naked_identifier` in PG dialect using RegexParser, and same as naked_identifier, except but without the anti_template that excludes keywords.  (Should we put it in ANSI dialect too?).  The name chosen because Spark SQL already did it and CP02 already matches with it.  
2. The parameter name always matches as `properties_naked_identifier`.
3.  Use `LiteralGrammar` as the first matching priority for the parameter value.  If it's not a literal, then try a `properties_naked_identifier`.

Your booleans and nulls get kept as boolean/null literals, but all the other keywords become properties_naked_identifier, which CP02 knows about.

Does that sound good?

WittierDinosaur: Generally like the approach, with one caveat - we're actively trying to deprecate the RegexParser. We really want all regexing to be done in the lexer. `TypedParser` may be able to achieve what you're after

james-johnston-thumbtack: Sounds good.  The latest commit takes that approach, and we're back to boolean literals!  I was able to use `TypedParser` as suggested for the new properties_naked_identifier.

WittierDinosaur:
Merge conflicts again

james-johnston-thumbtack:
@WittierDinosaur merge conflicts resolved.  Conflict in dialect_postgres.py was due to inserting two unrelated classes at the same position in the file.  Conflict in YAML was only checksum.

PR #4739, opened 2023-04-13
fix(athena): map type matching failed, array type only contains a datatype

timcosta:
The athena dialect was erroring when parsing `CAST(JSON_PARSE(mytable.mycolumn) AS MAP<VARCHAR, VARCHAR>)` because the `Sequence` for it was improperly constructed. This also fixes the array type which can only have a single type as the contents between angle braces.

These changes were verified against the grammar shown in https://docs.aws.amazon.com/athena/latest/ug/data-types.html.

None that I'm aware of, at worst it will fail poorly formed queries that had passed beforehand.

### Pull Request checklist

  - Other.


WittierDinosaur (approved):
Seems sensible - LGTM

PR #4746, opened 2023-04-13
Improve interval expressions on MySQL

james-johnston-thumbtack:
There are some initial efforts here, but it can be improved a lot:

1.  Add the keywords for the datetime_units set.  This is necessary for the ANSI-provided DatetimeUnitSegment to work properly with MySQL.  Previously we were falling back to ANSI-provided keywords, which is incomplete for MySQL.

2.  There is no concept of an "interval literal" in MySQL.  There are only interval expressions.  Therefore, delete INTERVAL from DateTimeLiteralGrammar.

    Also delete DATETIME from the literal: it's the name of a data type but it's not part of a date/time _literal_.  You'll get a straight parsing syntax error if you try (I did).

3.  An interval expression consists of an expression segment and the unit keyword.  That's what the upstream grammar says.  So we can simplify it.

    Also the existing grammar we had did not correctly require a DatetimeUnitSegment after the QuotedLiteralSegment.

4.  The PURGE statement references a DateTimeLiteralGrammar, but it's actually supposed to reference an expression, according to the upstream grammar.  Also, the purge_binary_logs.sql fixture had statements with literals using DATETIME.  As mentioned before, those are invalid and don't actually parse (I tested it).  So, delete those.

5.  There were some hacks that removed some things from the reserved keywords.  The interval units can be removed, since we have now implemented them properly.

    I don't know why INDEX was being removed.  I deleted it and none of the test cases started failing.  What I do know is that removing the keyword creates a parsing ambiguity with ALTER TABLE ... ADD, such that it considers "INDEX" (in "ADD INDEX") to be the name of a column that it will add.

6.  Regarding test coverage: I believe the existing test cases in interval.sql should be adequate.

### Pull Request checklist

  - Other.


WittierDinosaur (approved):
LGTM

PR #4747, opened 2023-04-13
Recognize quoted data types

### LINKED ISSUE #4621: [Postgres] unparseable error on various character types
### Search before asking

### What Happened

SQLFluff cannot parse all of the character data types that are documented at https://www.postgresql.org/docs/current/datatype-character.html

Example SQL (it is an integration test for a schema linter):

```sql
CREATE TABLE tbl (
    -- All forms of character data types listed at:
    -- https://www.postgresql.org/docs/current/datatype-character.html
    --
    -- None of them are allowed except for the text type (col_text).
    col_char_varying_unlimited character varying,
    col_char_varying_limited character varying(50),
    col_varchar_unlimited varchar,
    col_varchar_limited varchar(50),

    col_character_default character,
    col_character_specified character(50),
    col_char_default char,
    col_char_specified character(50),

    col_text text, -- the only allowed type

    col_system_char "char",
    col_name name
);
```

### Expected Behaviour

SQLFluff should parse the above SQL snippet successfully.

### Observed Behaviour

It fails with this error:

```
% sqlfluff lint .
== [tests/integration/schema_rules/use_text_type/test.sql] FAIL
                       | TABLE tbl (\n    -- All forms of c...'
WARNING: Parsing errors found and dialect is set to 'postgres'. Have you configured your dialect correctly?
All Finished 📜 🎉!
```

### How to reproduce

Try to parse it with `sqlfluff lint`

Some additional investigation seems to suggest that these types are problematic to parse:

- `"char"` (documented in table 8.5)
- `character varying` (with no upper bound given; it's a special PG extension from the standard)
- `character varying(50)`

The `character varying` options result in `LT01 | Unexpected whitespace before 'varying' keyword.` and an offer to "fix" automatically.  If I choose to do so, it deletes the space between the words, like so: `charactervarying`... which is invalid SQL.

There is an old related issue to parsing the latter types: https://github.com/sqlfluff/sqlfluff/issues/733 ... I guess this is a regression?

### Dialect

postgres

### Version

```
 % sqlfluff --version
sqlfluff, version 2.0.1
% python --version
Python 3.11.1
```

### Configuration

from `pyproject.toml`
```
[tool.sqlfluff.core]
dialect = "postgres"
templater = "raw"
exclude_rules = "ST06"
max_line_length = 100
```

### Code of Conduct

james-johnston-thumbtack:
The ANSI standard allows for quoted data types, as do many implementations.  However, SQLFluff does not.  Therefore the following trivial SQL would fail to parse:

    CREATE TABLE foo (
        col "my type"
    );

This commit fixes the issue by changing DatatypeIdentifierSegment so that it can also match with SingleIdentifierGrammar.  The initial data_type_identifier type segment is left as the default, since there is a capitalization rule that refers to it.  However, such a rule would be inappropriate to apply to quoted data types... so this commit just leaves the normal quoted_identifier segment type if the data type was quoted.

In order for this change to work, we have to move NOT from the list of ANSI non-reserved keywords to the list of ANSI reserved keywords. According to http://web.cecs.pdx.edu/~len/sql1999.pdf it is in fact a reserved word.  If we don't do this, one of the test cases will fail to parse correctly.

Fixes #4621 

### Pull Request checklist

  - Other.


james-johnston-thumbtack:
this fixes https://github.com/sqlfluff/sqlfluff/issues/4621

WittierDinosaur:
Looks good, but can you add the test case from #4621 into postgres' tests?

WittierDinosaur (approved):
LGTM

PR #4757, opened 2023-04-14
Fix issue with macros triggering LT01

alanmcruickshank:
Found an issue in testing where some macros would trigger LT01 unnecessarily. This was because of incorrect handling of templated newlines. This fixes that.


WittierDinosaur (approved):
LGTM

PR #4758, opened 2023-04-14
Allow comment match with preceding line

alanmcruickshank:
Another one found during testing against our project. IMHO I think the comment indentation rules are _too strict_ right now. While I agree that the default _fix_ anchor should be the _following_ line - if a comment currently matches the indentation of the line before, then I don't think that should trigger a linting error. This opens up that.

### DIFF
--- a/test/fixtures/rules/std_rule_cases/LT02-indent.yml
+++ b/test/fixtures/rules/std_rule_cases/LT02-indent.yml
@@ -346,12 +346,22 @@ test_pass_indented_from_with_comment:
     -- Comment
     JOIN t2 USING (user_id)
 
-test_fail_indented_from_with_comment_fix:
-  fail_str: |
+test_fail_indented_from_with_comment_alternate:
+  # This shows the alternative position of comments being allowed.
+  pass_str: |
     SELECT *
     FROM
         t1
         -- Comment
+    JOIN t2 USING (user_id)
+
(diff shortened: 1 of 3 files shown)


test/fixtures/rules/std_rule_cases/LT02-indent.yml
```diff
@@ -346,12 +346,22 @@ test_pass_indented_from_with_comment:
...
+    JOIN t2 USING (user_id)
+
 test_fail_indented_from_with_comment_fix:
+  # This shows the fix still returns to the primary location.
   fail_str: |
     SELECT *
     FROM
         t1
-        -- Comment
+            -- Comment
```
WittierDinosaur, line 364: Seems inconsistent where this is indented to, should it not have one more indent?

alanmcruickshank: Our current default indentation for comments is that they anchor to the line _after_ (i.e. it assumes that a comment comes before the line it describes). That's already true in 2.0.5.

All this PR does is _allow_ an additional location. If it's not in either, the fix routine will put it in the same place as before, which is to match the _following_ line in all cases.

WittierDinosaur: Okay, makes sense

WittierDinosaur (approved):
LGTM

PR #4761, opened 2023-04-15
Remove the majority of greedy matchers

WittierDinosaur:
Greedy matchers are less performant than their non-greedy equivalent. they're also more prone to over-matching. This PR removes most uses of StartsWith() - with a view to deprecate in the medium term. Of all of these changes, only one change has happened to the parse fixtures - and it's an improvement! There are a few StartsWith instances left, these have proven tricker to excise, so I figured I'd spin them off into a separate PR.

This also speeds up parsing of some of my large files by 5-10%

It changes the parse tree position of some trailing whitespace. But this is due to an inconsistency between GreedyMatch and Sequence. GreedyMatch discards trailing whitespace, Sequence does not. Either are fine, but we should be consistent - and hopefully, we should be able to remove GreedyMatch soon

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -1390,12 +1390,7 @@ class PartitionClauseSegment(BaseSegment):
     """A `PARTITION BY` for window functions."""
 
     type = "partitionby_clause"
-    match_grammar: Matchable = StartsWith(
-        "PARTITION",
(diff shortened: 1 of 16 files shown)


alanmcruickshank (approved):
This will make the parsing step slightly less resilient (i.e. it reduces the ability for the parser to parse some sections and not others), but I don't think that's regularly used, and if you're seeing a performance gain off the back of this - perhaps it's worth it 👍

PR #4760, opened 2023-04-15
Fix #4582: Comments after end of line

alanmcruickshank:
This fixes #4582.

Multiline comments would obscure the closing comment of a line. I did a little bit of tidy up here too to mitigate the potential performance implications of this change.

Also:
- Precalculate `IndentStats` on instantiation to simplify logic later on (and probably also get a performance boost).
- Moved the bracket check that was is `get_indent_impulse` into `_crawl_indent_points` because that's the only function that was actually using the results.

PR #4762, opened 2023-04-16
Remove indent hack

alanmcruickshank:
This takes a small part of #4760 and makes a separate PR for it. I thought it would be a simple change, but the knock on implications are surprising. One to investigate.


alanmcruickshank:
I'm going to close this because I think it's too hard 😭

PR #4768, opened 2023-04-16
Fix #4540: Untaken indents evaluation order.

### LINKED ISSUE #4540: 2.0.0 - Indentation error in CTE
### Search before asking

### What Happened

SQLFluff v2.0.0 gives an LT02 indentation error for a select statement inside a CTE which does not produce the same error with the "bare" select statement.

### Expected Behaviour

I expected the CTE to parse without errors. The same SQL parses without errors using SQLFluff v1.4.5.

### Observed Behaviour

SQLFluff gives the following linting errors:
```
L:   5 | P:  12 | LT02 | Expected line break and indent of 8 spaces before
                       | 'row_number'. [layout.indent]
L:   6 | P:   1 | LT02 | Expected indent of 12 spaces.
                       | [layout.indent]
L:   7 | P:   1 | LT02 | Expected indent of 8 spaces.
                       | [layout.indent]
```

### How to reproduce

SQL to reproduce:

```
-- Gives LT02 error when inside a CTE
with cte as (
    select a
    from b
    qualify row_number() over (
        partition by a
    ) = 1
)

select * from cte;

-- No error for the same select outside of CTE
select a
from b
qualify row_number() over (
    partition by a
) = 1;
```

### Dialect

`snowflake`

### Version

```
> sqlfluff --version
sqlfluff, version 2.0.0

> python --version
Python 3.9.9
```

### Configuration

```
[sqlfluff]
dialect = snowflake
```

### Code of Conduct

alanmcruickshank:
I think the part of code which I've covered here is now getting pretty mature, so this bug was a little hard to pinpoint - but I think the crux is that we should evaluate untaken indents _outward in_ (i.e. backwards) rather than forwards. This means that the right priority is given in locating potential indents to force. This resolves #4540 .

### DIFF
--- a/src/sqlfluff/utils/reflow/reindent.py
+++ b/src/sqlfluff/utils/reflow/reindent.py
@@ -884,53 +884,54 @@ def _map_line_buffers(
             )
             # There might be many indents at this point, but if any match, then
             # we should still force an indent
-            if any(i in indent_point.untaken_indents for i in passing_indents):
-                for i in passing_indents:
-                    # If we don't have the location of the untaken indent, then
-                    # skip it for now. TODO: Check this isn't a bug when this happens.
-                    # It seems very rare for now.
-                    if i not in untaken_indent_locs:
+
+            # NOTE: We work _inward_ to check which have been taken.
+            for i in reversed(passing_indents):
+                # Was this outer one untaken?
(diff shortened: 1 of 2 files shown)


WittierDinosaur (approved):
LGTM

PR #4771, opened 2023-04-17
Snowflake Execute Task with Schema

Thashin:
In Snowflake, allow for the execution of task by specifying the schema name and task name

### DIFF
--- a/src/sqlfluff/dialects/dialect_snowflake.py
+++ b/src/sqlfluff/dialects/dialect_snowflake.py
@@ -5736,7 +5736,7 @@ class ExecuteTaskClauseSegment(BaseSegment):
     match_grammar = Sequence(
         "EXECUTE",
         "TASK",
-        Ref("ParameterNameSegment"),
+        Ref("SchemaReferenceSegment"),
     )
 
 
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -5736,7 +5736,7 @@ class ExecuteTaskClauseSegment(BaseSegment):
     match_grammar = Sequence(
         "EXECUTE",
         "TASK",
-        Ref("ParameterNameSegment"),
+        Ref("SchemaReferenceSegment"),
```
alanmcruickshank, line 5739: `SchemaReferenceSegment` feels odd here because we're not referring to a schema. Perhaps `ObjectReferenceSegment` might be more appropriate (and I think effectively identical in meaning).

alanmcruickshank (commented):
Once suggestion on an alteration to your suggested change - looks like one CI check is failing too. `ymlchecks` is suggesting that you need to regenerate the parsing fixtures, now that you've updated the dialect.

src/sqlfluff/dialects/dialect_snowflake.py
Thashin: Fixed, cheers.

alanmcruickshank (approved):
LGTM

Issue #4645, opened 2023-03-31
LT01, BigQuery: Whitespaces in BigQuery `SAFE.` prefix 


dmohns:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Consider the following SQL

```sql
SELECT SAFE.STRING(JSON '1') AS string;
```

yields

```
L:   1 | P:  12 | LT01 | Expected single whitespace between 'SAFE' keyword and
                       | dot '.'. [layout.spacing]
L:   1 | P:  13 | LT01 | Expected single whitespace between dot '.' and function
                       | name identifier. [layout.spacing]
```

SQLFluff want `SAFE.STRING` to appear as `SAFE . STRING` which is odd (I think).

### Expected Behaviour

The `SAFE.` prefix of BigQuery functions should not be surrounded by spaces.

### Observed Behaviour

```sql
SELECT SAFE . STRING(JSON '1') AS string;
```

### How to reproduce

see above

### Dialect

BigQuery

### Version

2.0.2

### Configuration

default

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

fmms:
I believe this might be very similar to #4614

katieclaiborne:
We were able to resolve the "function name identifier" issue with the configuration below.
```
[sqlfluff:layout:type:function_name]
spacing_before = touch:inline
```

katieclaiborne:
No luck on the "keyword" issue, though. Could this have been an unintended consequence of adding [SAFE](https://github.com/sqlfluff/sqlfluff/blob/402a47c4cbe5b280acdbeed53e0c5ee906a20f4f/src/sqlfluff/dialects/dialect_bigquery_keywords.py#L242) to the `bigquery_unreserved_keywords` list in #3048?

We tried the configuration below, which changed the issue to "Unexpected whitespace before naked identifier".
```
[sqlfluff:layout:type:keyword]
spacing_after = touch
```

Running SQLFluff fix removed the spaces after all other keywords.

Before:
```sql
safe.parse_date('%Y-%m-%d', raw_date_field) as parsed_date_field
```
After:
```sql
[...]

PR #4773, opened 2023-04-17
Inroduce `SAFE` prefix segment

dmohns:

Makes progress in: https://github.com/sqlfluff/sqlfluff/issues/4645

The idea here is to introduce a BigQuery specific `SafePrefixSegment` to better control the whitespacing within the function call. This

- **does** resolve the whitespace issue between `dot` and `function_name`
- **does not** resolve the whitespace issue between `SAFE` and `dot`

**Open Question:**

It's not clear to me why this still happens

```
"Expected single whitespace between 'SAFE' keyword and dot '.'."
```

Shouldn't

```
Sequence("SAFE", Ref("DotSegment"), allow_gaps=False)
```

prevent exactly that?

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -824,7 +824,10 @@ class NormalizeFunctionNameSegment(BaseSegment):
 
 
 class FunctionNameSegment(ansi.FunctionNameSegment):
-    """Function name, including any prefix bits, e.g. project, schema or SAFE keyword."""
+    """Describes the name of a function.
+
+    This includes any prefix bits, e.g. project, schema or SAFE keyword.to
+    """
 
     match_grammar: Matchable = Sequence(
         # Project name, schema identifier, etc.
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -823,30 +823,17 @@ class NormalizeFunctionNameSegment(BaseSegment):
     )
 
 
-class SafePrefixSegment(BaseSegment):
-    """SAFE prefix to a function name.
-
-    BigQuery Function names can be prefixed by the keyword SAFE to
-    return NULL instead of error.
-    https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-reference#safe_prefix
-    """
-
-    type = "safe_prefix"
-    match_grammar: Matchable = Sequence("SAFE", Ref("DotSegment"), allow_gaps=False)
-
-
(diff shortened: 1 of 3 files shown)


alanmcruickshank (commented):
The reason the layout configs aren't getting applied as you expect is the introduction of the `safe_prefix` segment. You _could_ create a new layout config for `safe_prefix` and add that to the default config, but I think you might find it easier to just add the optional `SAFE` keyword option into the `FunctionNameSegment` and then you can use the existing layout config for that segment which says there shouldn't be spaces within it.

dmohns:
> The reason the layout configs aren't getting applied as you expect is the introduction of the `safe_prefix` segment. You _could_ create a new layout config for `safe_prefix` and add that to the default config, but I think you might find it easier to just add the optional `SAFE` keyword option into the `FunctionNameSegment` and then you can use the existing layout config for that segment which says there shouldn't be spaces within it.

thanks @alanmcruickshank I wasn't aware about "layout config", could you point we to where they are defined for a given segment? The way you suggested is of course much cleaner. Changed the PR accordingly.

There is one small problem though, now the SQL as getting fixed to by current version (2.0.5) is

```
SELECT SAFE . STRING(JSON '1')
```

is producing a syntax error.

alanmcruickshank:
> There is one small problem though, now the SQL as getting fixed to by current version (2.0.5) namely
> 
> ```sql
> SELECT SAFE . STRING(JSON '1')
> ```

I think what to do here @dmohns depends on whether that example parses in bigquery. If it _does_ then I think the bigquery version of `FunctionNameSegment` should have the `allow_gaps=False` removed. If it _doesn't_ parse (i.e. isn't not valid SQL anyway) - then I think this PR is good to go 👍

dmohns:
> > There is one small problem though, now the SQL as getting fixed to by current version (2.0.5) namely
> > ```sql
> > SELECT SAFE . STRING(JSON '1')
> > ```
> 
> I think what to do here @dmohns depends on whether that example parses in bigquery. If it _does_ then I think the bigquery version of `FunctionNameSegment` should have the `allow_gaps=False` removed. If it _doesn't_ parse (i.e. isn't not valid SQL anyway) - then I think this PR is good to go 👍

I checked this. Turns out that actually both of these

```sql
SELECT * FROM project  .   dataset   .   table
```

and

```sql
SELECT SAFE . STRING(JSON '1')
```

are valid BigQuery SQLs.

I will update the segment definition accordingly.

alanmcruickshank:
> But this should be a separate issue/PR.

I'll make one. Feel like addressing it yourself? You know exactly how to now 😄

alanmcruickshank (approved):
Thanks for this!

PR #4776, opened 2023-04-17
Fix #2855: (Tech debt: check consistency in TemplatedFile init)

### LINKED ISSUE #2855: Tech debt: Eliminate the "check_consistency" flag in "TemplatedFile.__init__", requires cleaning up hardcoded test data in jinja_test.py
### Search before asking

### What Happened

We recently added some sanity checks to the slice information in `TemplatedFile.__init__()`. Those checks are very helpful, but they cause some tests in `test/core/templaters/jinja_test.py` to fail due to some bad hardcoded data. For more info, see the comment around line 108 of that file:
```
        # NOTE: The "check_consistency" flag should always be True when using
        # SQLFluff in real life. This flag was only added because some legacy
        # templater tests in test/core/templaters/jinja_test.py use hardcoded
        # test data with issues that will trigger errors here. It would be cool
        # to fix that data someday. I (Barry H.) started looking into it, but
        # it was much trickier than I expected, because bits of the same data
        # are shared across multiple tests.
```

### Expected Behaviour

I expected the tests in `jinja_test.py` to pass the sanity checks.

### Observed Behaviour

The sanity checks failed.

### How to reproduce

Find and replace occurrences of `check_consistency=True` in `test/core/templaters/jinja_test.py`, replacing them with `check_consistency=False`, then run those tests:
```
py.test -v test/core/templaters/jinja_test.py
```

You'll see a number of failures.

### Dialect

n/a

### Version

Was running from not-yet-merged PR #2849. (Expect this to be merged soon due to some high-priority fixes.)

### Configuration

n/a

### Code of Conduct

alanmcruickshank:
This resolves #2855. There was an unnecessary flag to cover some misshapen test cases which I've fixed here. It's a low risk PR because only really covers the test suite and not the core codebase.

### DIFF
--- a/src/sqlfluff/core/templaters/base.py
+++ b/src/sqlfluff/core/templaters/base.py
@@ -120,7 +120,6 @@ class TemplatedFile:
         templated_str: Optional[str] = None,
         sliced_file: Optional[List[TemplatedFileSlice]] = None,
         raw_sliced: Optional[List[RawFileSlice]] = None,
-        check_consistency=True,
     ):
         """Initialise the TemplatedFile.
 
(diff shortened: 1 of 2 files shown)


barrywhart (approved):
Thanks for the cleanup! 🎉🥳

PR #4795, opened 2023-04-20
fix: correct macos/osx config file location

jpuris:

Use linux style xdg on mac (#955) had introduced a change where macOS config files no longer use the `~/Library/.. ` config directory and instead prefers the linux variant. The documentation was not updated, however.

This PR seeks to amend that.

No. This is only a documentation fix.


WittierDinosaur (approved):
LGTM

Issue #4787, opened 2023-04-19
Bigquery: Object references with spaces don't parse in SQLFluff


alanmcruickshank:
> > There is one small problem though, now the SQL as getting fixed to by current version (2.0.5) namely
> > ```sql
> > SELECT SAFE . STRING(JSON '1')
> > ```
> 
> I think what to do here @dmohns depends on whether that example parses in bigquery. If it _does_ then I think the bigquery version of `FunctionNameSegment` should have the `allow_gaps=False` removed. If it _doesn't_ parse (i.e. isn't not valid SQL anyway) - then I think this PR is good to go 👍

I checked this. Turns out that actually both of these

```sql
SELECT daniel_test    .    AddFourAndDivide(5, 10);
```

and

```sql
SELECT SAFE . STRING(JSON '1')
```

are valid BigQuery SQLs.

I will update the segment definition accordingly.

I also noted that, actually

```sql
SELECT * FROM project  .   dataset   .   table
```

is a valid BigQuery SQL but currently doesn't parse. But this should be a separate issue/PR.

_Originally posted by @dmohns in https://github.com/sqlfluff/sqlfluff/issues/4773#issuecomment-1514366405_

dmohns:
Please assign it to me 😉

nathangriffiths-cdx:
From the docs: https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical

> You can separate tokens with comments or whitespace such as spaces, backspaces, tabs, or newlines.

So in fact all of these are valid SQL:

```
-- spaces 
SELECT * FROM project  .  dataset   .   table
```
```
-- tabs
SELECT * FROM project[TAB].[TAB]dataset[TAB].[TAB]table
```
```
-- newlines
SELECT 
* 
FROM 
project 
.
dataset
.
table
```

I don't imagine that many people actually write SQL like that though.

PR #4796, opened 2023-04-20
fix: #4787 BigQuery allow spaces in object references

dmohns:

Resolves: #4787 

BigQuery allows whitespaces in object references. However, currently SQL that "utilize" this currently don't parse. This PR tries resolve this and allow spaces (at least) in table references.

Unfortunately, this seem to entirely break to logic around templated areas. I will add a (temporary) [commit](https://github.com/sqlfluff/sqlfluff/pull/4796/commits/d8f08a23abd8cabb25d96208a6963e4663b7ca64) to this branch that demonstrate this (by making the corresponding tests pass). I'm not really sure, how this should be addressed, so looking forward to get some hints/guidance.

Currently, yes. It entirely breaks the logic around templates areas, see above.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -1266,7 +1266,10 @@ class TableReferenceSegment(ObjectReferenceSegment):
             Ref("JoinLikeClauseGrammar"),
             BracketedSegment,
         ),
-        allow_gaps=False,
+        # BigQuery allows whitespaces between the `.` of a table refrencex.
+        # Keeping the explicit `allow_gaps=True` here to
+        # make the distinction to other dialects clear.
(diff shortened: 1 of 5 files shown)


greg-finley:
I keep seeing this one but not sure how to assess. Does it do more good than harm?

dmohns:
> I keep seeing this one but not sure how to assess. Does it do more good than harm?

I think this would be nice addition, but definitely "just" nice-to-have. 

However, the problem is, to get this, we need to identify the part that is breaking this: https://github.com/sqlfluff/sqlfluff/pull/4796/commits/d8f08a23abd8cabb25d96208a6963e4663b7ca64

Unfortunately, this might be a bit out of my scope, as I think it requires deeper understand of the templating logic.

PR #4835, opened 2023-05-02
docs(templater): Refactor templater configuration docs

dmohns:

This PR tries to improve the documentation about configuration of the various templaters.

Currently, the documentation can be a little hard to read and understand. For example

- Placeholder templater docs is a subsection to Jinja. In the middle of other Jinja documentation, which can be distracting.
- Alone from the configuration documentation It's not immediately clear which templaters exists. Users have to find the  Internals documentation to learn about available templaters.
- All available configuration options for the Jinja templater are not available at one glance.

See the current structure:

image

To address these point, this PR groups all templating into a section of the configuration docs, like so

image

Then, all the specifics are handled in the corresponding sections.

**Note:** I tried to leave to documentation itself, as untouched as possible. Only the general section is new. For existing ones, just changing the hierarchy levels. I'm planing to add further changes to the documentation itself, but I don't wanna mix this in here, which is focusing on the structure on the docs mainly.

No, documentation only change.

### Pull Request checklist

  - Other.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -448,7 +448,7 @@ For example, if passed the following *.sql* file:
 Jinja templater
 ^^^^^^^^^^^^^^^
 
-The Jinja template uses Jinja2_ to render templates.
+The Jinja templater uses Jinja2_ to render templates.
 
 .. _Jinja2: https://jinja.palletsprojects.com/
 
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -454,11 +454,29 @@ The Jinja template uses Jinja2_ to render templates.
 
 There are multiple, complementary ways of configuring the Jinja templater.
 
-- Reading variables and Jinja macros from config, see
-  `Complex Jinja Variable Templating`_ and
-  `Jinja Macro Templating (from config)`_
-- Loading macros from a path, see `Jinja Macro Templating (from file)`_
-- Using a library, see `Library Templating`_
+- Reading variables and Jinja macros directly from the SQLFLuff config file
+- Loading macros from a path
+- Using a library
+
+.. list-table:: Overview of Jinja templater's configuration options
+   :header-rows: 1
+
+   * - Configuration
+     - Variables
+     - Macros
+     - Documentation
+   * - Config file
+     - ✅
+     - ✅
+     - `Complex Jinja Variable Templating`_ and `Jinja Macro Templating (from config)`_
+   * - Macro Path
+     - ❌
+     - ✅
+     - `Jinja Macro Templating (from file)`_
+   * - Library
+     - ✅
+     - ✅
+     - `Library Templating`_
 
 For example, a snippet from a :code:`.sqlfluff` file that uses all config
 options:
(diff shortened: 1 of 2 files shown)


docs/source/configuration.rst
```diff
@@ -403,112 +427,49 @@ For example, if passed the following *.sql* file:
...
-For example a query in SQLAlchemy can look like this:
+.. _Jinja2: https://jinja.palletsprojects.com/
 
-.. code-block:: sql
+There are multiple, complementary ways of configuring the Jinja templater.
 
-    SELECT * FROM table WHERE id = :myid
+- Reading Variables and macros from config
+- Loading macros from a path.
+- Using a library
```
alanmcruickshank, line 441: Should some of these be links to the relevant sections below?

docs/source/configuration.rst
```diff
@@ -356,18 +356,42 @@ layout: :ref:`layoutconfig`.
...
+SQL at run time.
+This can range from very simple placeholder templating to complex Jinja
+templating.
 
-Variables are available in the *jinja*, *python* and *placeholder* templaters.
+SQLFluff supports templated sections in SQL, see :ref:`templater`.
+This is achieved by the following set of operations:
+
+1. SQLFluff pre-renders the templated SQL
+2. SQLFluff applies the lint and fix operations to the rendered file
```
alanmcruickshank, line 374: This would be a great moment to add a note saying something like: 
> SQLFluff linting requires an understanding of the SQL syntax, which means linting relies on having a valid SQL file to parse. The implication of this for the templater is that sections of the raw file which aren't included in the rendered file (e.g. because they're in an :code:`{% if ... %}` statement), will _not be linted_. This is functionality we hope to support in future.

Whether that's in as a note, or as part of point 2 here I think doesn't particularly matter.

docs/source/configuration.rst
```diff
@@ -356,18 +356,42 @@ layout: :ref:`layoutconfig`.
...
+3. SQLFluff backports the rule violations to the templated section of the SQL.
+
+SQLFluff may not have access to the fully operational templater.
+But instead uses a set of user-provided dummy values to render the template,
+see below.
```
alanmcruickshank, line 379: I think the word "operational" here might be confusing. Perhaps better would be to say:

> SQLFluff does not automatically have access to the _same environment_ as your production template setup. This means it is necessary to either provide that environment, or to provide dummy values in order to effectively render the template and still generate valid SQL.

alanmcruickshank (commented):
Thanks for taking this on! I really like what you're trying to achieve here. I've got a couple of suggestions to improve clarity and linking, but mostly this looks great ❤️

docs/source/configuration.rst
dmohns: Yeah, good idea. I added a note similar to what you suggested.

dmohns:
I'm adding a small overview to compare the different configuration types.

image

WDYT?

dmohns:
Hey, this has been stale for a while, any further suggestions, comments on it?

alanmcruickshank (approved):
Sorry - I had to take a couple of weeks away from the project. Looks great. Thanks for putting this together 👍

Issue #4811, opened 2023-04-25
Support for SHOW statement


dogversioning:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Several databases (mysql, athena, for two) have support for a sql statement like `SHOW TABLES FROM schema_or_db_name '*regex*'` or `SHOW VIEWS IN schema_or_db_name like '*regex*' to get information about what tables are present. It would be great to have SQLFluff support this - I've tried this for both dialects with jinja templates and raw queries and consistently get an 'unparseable section' error message.

I'm checking the 'would submit a PR' box but I might need some help with it.

### Expected Behaviour

SQL statements formatted in this way should pass linitng.

### Observed Behaviour

Error messages like the following:
```L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: "SHOW
                       | TABLES FROM `test_schema` 'Test__*'..."```

### How to reproduce

Create four files files:

`test_table.sql`: 
```SHOW TABLES FROM `schema_name` 'prefix__*';

```
`test_table.sql.jinja`: 
```SHOW TABLES FROM `{{schema_name}}` '{{prefix}}__*';

```

`test_view.sql`: 
```SHOW VIEWS IN "schema_name" LIKE 'prefix__*';

```
`test_view.sql.jinja`: 
```SHOW VIEWS IN "{schema_name}" LIKE '{prefix}__*';

```

### Dialect

athena, mysql

### Version

2.0.2

### Configuration

pyproject.toml:
```[tool.sqlfluff.core]
templater = "jinja"
dialect = "athena"
sql_file_exts = ".sql,.sql.jinja"
# this rule overfires on athena nested arrays
exclude_rules="references.from,structure.column_order"
max_line_length = 88

[tool.sqlfluff.indentation]
template_blocks_indent = false

[tool.sqlfluff.rules.layout.long_lines]
ignore_comment_lines = true

[tool.sqlfluff.rules.capitalisation.keywords]
capitalisation_policy = "upper"

[tool.sqlfluff.templater.jinja.context]
schema_name = "test_schema"
table_name = "test_table"
prefix = "Test"
dataset = [["foo","foo"],["bar","bar"]]
table_cols = ["a","b"]
col_type_list = ["a string","b string"]```

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

[...]

PR #4854, opened 2023-05-09
Add show tables/views to athena

dogversioning:

This PR makes progress on #4811 by  addressing the Athena side of the bug report. It adds a new Show segment and the associated tests to drive it.

I narrowed the scope to just one DB since this is my first PR over here, I have a couple style questions that I thought would be easier to ask in this context.

Nope, just grammar

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_athena.py
+++ b/src/sqlfluff/dialects/dialect_athena.py
@@ -662,13 +662,13 @@ class ShowStatementSegment(BaseSegment):
         OneOf(
             Sequence(
                 "TABLES",
-                Sequence("IN", Ref("TableReferenceSegment"), optional=True),
-                Ref("ExpressionSegment", optional=True),
+                Sequence("IN", Ref("DatabaseReferenceSegment"), optional=True),
+                Ref("QuotedLiteralSegment", optional=True),
             ),
             Sequence(
                 "VIEWS",
-                Sequence("IN", Ref("TableReferenceSegment"), optional=True),
-                Sequence("LIKE", Ref("ExpressionSegment"), optional=True),
+                Sequence("IN", Ref("DatabaseReferenceSegment"), optional=True),
+                Sequence("LIKE", Ref("QuotedLiteralSegment"), optional=True),
             ),
         ),
     )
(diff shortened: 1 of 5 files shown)


src/sqlfluff/dialects/dialect_athena.py
```diff
@@ -643,3 +644,31 @@ class IntervalExpressionSegment(BaseSegment):
...
+        OneOf(
+            Sequence(
+                "TABLES",
+                Sequence("IN", Ref("TableReferenceSegment"), optional=True),
+                Ref("ExpressionSegment", optional=True),
+            ),
+            Sequence(
+                "VIEWS",
+                Sequence("IN", Ref("TableReferenceSegment"), optional=True),
+                Sequence("LIKE", Ref("ExpressionSegment"), optional=True),
```
dogversioning, line 671: so i'm mostly concerned about this approach - i'm currently trying to address just my use case, but if one were to fully cover the show keyword in Athena, there would be 8 nested bespoke clauses that are all based off the same keyword. should these all be in the same segment? or should they be broken out a bit more?

WittierDinosaur: I would say for this, it's fine to have all of these wrapped in a big OneOf. Mainly because the individual statements are so small for each type. It's primarily just about style preference at this point, and I don't have one unless one style becomes significantly less performant than another.

src/sqlfluff/dialects/dialect_athena.py
```diff
@@ -643,3 +644,31 @@ class IntervalExpressionSegment(BaseSegment):
...
+    https://docs.aws.amazon.com/athena/latest/ug/ddl-reference.html
+    """
+
+    type = "show_statement"
+    match_grammar = Sequence(
+        "SHOW",
+        OneOf(
+            Sequence(
+                "TABLES",
+                Sequence("IN", Ref("TableReferenceSegment"), optional=True),
```
WittierDinosaur, line 665: I think both of these `TableReferenceSegment` should be `DatabaseReferenceSegment`

src/sqlfluff/dialects/dialect_athena.py
dogversioning: ok - so thinking about the next steps in this area, if i'm gonna start walking down DBs that support these (i'm sure of hive & mysql, but i could poke around in some of the others), would it be better to do one PR per DB type, or just try to hit things in one broad stroke?

WittierDinosaur (approved):
LGTM

WittierDinosaur (approved):
LGTM

PR #4858, opened 2023-05-10
Add link to Trino keywords (Athena v3)

KulykDmytro:

Athena v2 uses Presto, Athena v3 uses Trino as backend.
So i've added link to reserved keywords for Trino


greg-finley (approved):
Thanks!

Issue #4841, opened 2023-05-04
AttributeError: module 'dbt.flags' has no attribute 'PROFILES_DIR' with dbt 1.5


johnpkennedy:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

After upgrading to **dbt 1.5** and **SQLFluff 2.1.0**, I get the following error when I run `sqlfluff lint`:

```
AttributeError: module 'dbt.flags' has no attribute 'PROFILES_DIR'
```

### Expected Behaviour

I don't get this error on SQLFluff 2.0.7 with dbt 1.4.6.

### Observed Behaviour

```
% sqlfluff lint
=== [dbt templater] Sorting Nodes...                                                                     
file /Users/johnkennedy/Projects/maple-data/dbt/dbt_packages/audit_helper/integration_tests/analyses/compTraceback (most recent call last):
  File "/Users/johnkennedy/.local/share/virtualenvs/dbt--3QRDvUu/bin/sqlfluff", line 8, in <module>
    sys.exit(cli())
  File "/Users/johnkennedy/.local/share/virtualenvs/dbt--3QRDvUu/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/johnkennedy/.local/share/virtualenvs/dbt--3QRDvUu/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/johnkennedy/.local/share/virtualenvs/dbt--3QRDvUu/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/johnkennedy/.local/share/virtualenvs/dbt--3QRDvUu/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/johnkennedy/.local/share/virtualenvs/dbt--3QRDvUu/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/johnkennedy/.local/share/virtualenvs/dbt--3QRDvUu/lib/python3.10/site-packages/sqlfluff/cli/commands.py", line 605, in lint
    result = lnt.lint_paths(
[...]

PR #4886, opened 2023-05-25
profiles dir env var or default

JasonGluck:
Fix Issue  #4841.

### DIFF
--- a/plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
+++ b/plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
@@ -13,6 +13,7 @@ from dbt.version import get_installed_version
 from dbt.config.runtime import RuntimeConfig as DbtRuntimeConfig
 from dbt.adapters.factory import register_adapter, get_adapter
 from dbt.compilation import Compiler as DbtCompiler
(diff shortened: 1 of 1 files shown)


greg-finley:
@JasonGluck Thanks so much! Are you confident in this change? (Just wondering why marked as draft)

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -193,7 +194,7 @@ def _get_profiles_dir(self):
                 self.sqlfluff_config.get_section(
                     (self.templater_selector, self.name, "profiles_dir")
                 )
-                or flags.PROFILES_DIR
+                or (os.getenv("DBT_PROFILES_DIR") or default_profiles_dir())
```
greg-finley, line 197: outer parens maybe unnecessary but also fine

JasonGluck:
> @JasonGluck Thanks so much! Are you confident in this change? (Just wondering why marked as draft)

Tested this locally  both the with envvar set and without (default) and it worked. Put in draft because I have not yet contributed and wanted to make sure I was following the process properly.

elyobo:
Thanks @JasonGluck!

PR #4905, opened 2023-06-07
Draft: Use dbtRunner for dbt 1.5+ for slimmer integration.

alanmcruickshank:
This prototypes the `dbtRunner` class as a way to invoke dbt more simply without as much fragile code.

The code is better. This is good.

Performance however is poor so far. Needs more work before I'm happy to merge it, although I might merge parts of it in the mean time.


alanmcruickshank:
I'm going to abandon this PR. It's too far behind `main` and I don't think it's easily feasible to integrate in it's current form.

If we wanted to have another go at this, I think it's one to start from scratch.

PR #4925, opened 2023-06-21
Security option for library path

alanmcruickshank:
This resolves a potential security risk, and also adds a bunch of documentation to cover it.

It enables a CLI option to override the `library_path` option in the templater to prevent unsecured code being called.

### DIFF
--- a/docs/source/production.rst
+++ b/docs/source/production.rst
@@ -31,8 +31,8 @@ tool functions in a secure environment.
    be able to access and edit the :ref:`config-files`. It's important to note
    that because of :ref:`in_file_config`, that users who can edit SQL files
    which are designed to be linted, will also have access to the vast majority
-   of any configuration options available in :ref:`config-files`. This means that
-   there is minimal additional protection from restricting access to
+   of any configuration options available in :ref:`config-files`. This means
+   that there is minimal additional protection from restricting access to
    :ref:`config-files` for users who already have access to edit the linting
    target files (as described above).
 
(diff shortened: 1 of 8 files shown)


docs/source/production.rst
```diff
@@ -21,34 +80,34 @@ more.
 
 .. _diff-quality:
 
-Using SQLFluff on changes using `diff-quality`
-----------------------------------------------
+Using SQLFluff on changes using ``diff-quality``
+------------------------------------------------
```
greg-finley, line 84: I think your IDE is thrashing with a change made recently by @tunetheweb . I think we want single backticks not double?

alanmcruickshank: for sphinx (rst) - I think we **do** want the double ones. Single ones seem to make the text _italic_ in the docs, double ticks make is `code`.

src/sqlfluff/cli/commands.py
```diff
@@ -382,8 +392,17 @@ def get_config(
     from_root_kwargs = {}
     if "require_dialect" in kwargs:
         from_root_kwargs["require_dialect"] = kwargs.pop("require_dialect")
+    library_path = None
+    if "library_path" in kwargs:
+        library_path = kwargs.pop("library_path")
```
greg-finley, line 397: I think also works?

```suggestion
    library_path = kwargs.pop("library_path", None)
```

greg-finley (approved):

src/sqlfluff/cli/commands.py
alanmcruickshank: Ah yes - thanks for that 👍

PR #4942, opened 2023-07-01
Move from `make_template` to `render_func` in jinja and dbt

alanmcruickshank:
This sets the groundwork for #4905. Rather than passing around a `make_template` function (which returns a jinja `Template` object) in the jinja (and therefore also dbt) templaters, this simplifies the interface to pass around a `render_func` method (which accepts a string and returns a string). The `make_template` function feels inappropriate for a standardised interface because it pre-supposes that we're using jinja directly.

This should be an internal-only change, but does change the interface for any dependent templaters as it changes the interface required by a templater plugin.

Suggest that this is released as a minor release rather than bugfix.

**For reviewers:** I'd suggest first reviewing the changes to the `python` templater, then the `jinja` templater (potentially by looking at the fist two commits in isolation). Once you get a feel for what I'm doing there, the `dbt` changes will make more sense. In particular when you look at some of my changes to the test suite, there's a blend of _necessary_ changes, and also some simplifications/clarifications made possible by the new approach.


plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -616,7 +621,23 @@ def make_template(in_str):
             #       production, slice_file() does not usually use this string,
             #       but some test scenarios do.
             setattr(node, RAW_SQL_ATTRIBUTE, source_dbt_sql)
-            compiled_sql = compiled_sql + "\n" * n_trailing_newlines
+
+            # So for files that have no templated elements in then render_func
```
barrywhart, line 625: Possible typo:
then -> them

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -494,25 +494,30 @@ def _unsafe_process(self, fname, in_str=None, config=None):
...
                         # Yes. Capture the important arguments and create
-                        # a make_template() function.
+                        # a render_func() closure with overwrites the variable
+                        # from within _unsafe_process when from_string is run.
                         env = args[0]
                         globals = args[2] if len(args) >= 3 else kwargs["globals"]
 
-                        def make_template(in_str):
+                        # Overwrite the outer render_func
+                        def render_func(in_str):
```
barrywhart, line 517: Nit: I think `render()` or `render_template()` would be a clearer name for the function implementations. `render_func` seems like a reasonable name for the variable(s) that hold a reference to the function.

alanmcruickshank: I think I see what you mean. In this case, the definition is to overwrite a variable and so they have to be the same. I'm tempted to keep this as is for now just because then at least we use a consistent name for the same thing throughout. Unless you think it's worth changing all references to `render_func` to just `render`? I'm open to that last suggestion if you think it would be easier to understand across the board. I went with `render_func` on balance just because it's often passed around in a context where it's not 100% obvious that it's a function.

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
alanmcruickshank: Good catch.

alanmcruickshank:
> @noel: Note that this will require updating the templater in dbt-core.

Agreed - I propose that this feature would be released as a `2.2.0` release rather than as part of `2.1.4` (which I think should come out first with some of the other performance improvements).

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
barrywhart: I do think it's worth renaming, because templaters are part of the public-facing API of SQLFluff, so clarity around function names and purposes is pretty important.

I see your point about "the function itself" versus "variable referring to the function". There are really two options:
* Rename it **everywhere**, as you mention.
* Rename the functions, then add an explicit assignment statement to overwrite the variable. Example:

```
nonlocal render_func
...
def render(in_str):
    ...

render_func = render
```

There may be a third option. Give me a few minutes -- I'll look into that and explain it if it seems like a good candidate.

barrywhart: After a closer look, I'm withdrawing my suggestion about renaming. I was somehow thinking that this was a normal, public-facing function on every templater class. That's only true for the Python templater. For Jinja and dbt, it's a dynamically created function. I should have realized this, since I created a lot of this code in the first place. 😭

Looks good to me!

alanmcruickshank:
This one is ready to rock, but I won't merge it straight away - there's probably enough to get another bugfix release out before merging this one.

PR #4944, opened 2023-07-02
Re-enable dbt 1.1 & 1.2

alanmcruickshank:
This re-enables support and tests for dbt 1.1 and 1.2. It effectively selectively reverts the fixes made in #4886, which fixed support for 1.5, but broke support for 1.1 & 1.2.

Combined with #4941 and #4939, this means we should have full compatibility with dbt 1.1.0 up to the latest 1.5.2. Anything pre 1.1 doesn't compile anymore so I'm very comfortable not supporting that 😄 .

Note that I'm added `dbt110` back into the coverage group so that the clauses for 1.1 and 1.2 are covered.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -87,20 +87,20 @@ jobs:
     secrets:
       gh_token: ${{ secrets.github_token }}
 
-  # dbt-tests-nocov:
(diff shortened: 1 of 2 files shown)


WittierDinosaur (approved):
Nice!

PR #4981, opened 2023-07-18
Terminator inheritance

alanmcruickshank:
As discussed with @WittierDinosaur - this implements terminator inheritance. Effectively allowing smart inheritance of some terminators such that the `AnyOf` grammar can terminate earlier by finding appropriate terminator tokens.

I've also added `--parser-statistics` option which outputs the relevant statistics to debug this option. I'm not expecting many end users to use the option - it's much more developer focussed so I think it's ok that it's uses the logging output rather than something more elegant.

I tested this on the `benchmarks\bench_002\bench_002_pearson.sql` file and managed to save about 10-15% on time on a parsing operation. I suspect on heavily nested expressions the improvement should be much bigger. Would be good to test this on some heavier files - that might also hint to some other terminators that need adding - but the ones I've done already seem like a sensible first step.


src/sqlfluff/core/parser/grammar/delimited.py
```diff
@@ -151,15 +151,18 @@ def match(
...
                     match, _ = self._longest_trimmed_match(
                         segments=seg_content,
                         matchers=elements,
                         parse_context=ctx,
                         # We've already trimmed
                         trim_noncode=False,
-                        terminators=delimiter_matchers
-                        if elements != delimiter_matchers
-                        else None,
+                        terminators=terminators,
```
WittierDinosaur, line 165: should this not be `ctx.terminators`?

WittierDinosaur:
This is a pretty elegant implementation. The only thing I think we're missing here is the option to intentionally not inherit terminators in any grammar/segment. Tests are failing, I imagine this is why? Not sure. Slightly around tomorrow, but then I'm AFK until Tuesday

src/sqlfluff/core/parser/grammar/delimited.py
alanmcruickshank: No - but you're right that there is duplication. In fact we shouldn't need to specify the terminators here _at all_ because they're fetched from the context within that function anyway. 👍 

Will tidy up

alanmcruickshank:
> 

The functionality exists, but the only use of it is by default within brackets. You're probably right that we might need to do some tweeking with this in some cases to get it just right. I've just solved one round of bugs and will see where else we need to do that.

WittierDinosaur:
~20% faster on complex files!!

alanmcruickshank:
> ~20% faster on complex files!!

That's epic. I do wonder whether a couple more terminators could bring that down a lot. Pick one of your big ones and run it with `--parse-statistics` - then paste the logs back here - there should be a section which looks a bit like this (but probably with bigger numbers):
```
WARNING    ==== Parse Statistics ====
WARNING    ltm_calls: 4973
WARNING    complete_match: 1141
WARNING    unterminated_match: 1795
WARNING    ltm_calls_w_ctx_terms: 3698
WARNING    terminated_match: 2037
WARNING    ## Tokens following un-terminated matches
WARNING    Adding terminator clauses to catch these may improve performance.
WARNING    6: 'DESC'
WARNING    1: 'CROSS'
WARNING    ==== End Parse Statistics ====
```

Depending on that the `Tokens following un-terminated matches` are - I might be able to take a few more bites in the same PR.

WittierDinosaur:
The top 
```
WARNING    ==== Parse Statistics ==== 
WARNING    ltm_calls: 55974 
WARNING    unterminated_match: 29378 
WARNING    ltm_calls_w_ctx_terms: 47994 
WARNING    terminated_match: 25000 
WARNING    complete_match: 1596 
WARNING    ## Tokens following un-terminated matches 
WARNING    Adding terminator clauses to catch these may improve performance. 
WARNING    2014: '.' 
WARNING    804: '=' 
WARNING    431: 'LEFT' 
WARNING    278: '(' 
WARNING    199: 'AND' 
WARNING    189: 'WHEN' 
WARNING    156: '/' 
WARNING    151: 'THEN' 
WARNING    109: 'END' 
```

src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -239,9 +239,20 @@ def _longest_trimmed_match(
...
 
+        NOTE: This matching method is the workhorse of the parser. It's performance
+        can be monitored using the `parse_stats` object on the context.
         """
         terminated = False
 
+        terminators = terminators or []
+        parse_context.increment("ltm_calls")
+        if terminators:
+            parse_context.increment("ltm_calls_w_terminator")
```
WittierDinosaur, line 250: this may be redundant now that we are just using the context

alanmcruickshank:
> The top
> 
> ```
> WARNING    ==== Parse Statistics ==== 
> WARNING    ltm_calls: 55974 
> WARNING    unterminated_match: 29378 
> WARNING    ltm_calls_w_ctx_terms: 47994 
> WARNING    terminated_match: 25000 
> WARNING    complete_match: 1596 
> WARNING    ## Tokens following un-terminated matches 
> WARNING    Adding terminator clauses to catch these may improve performance. 
> WARNING    2014: '.' 
> WARNING    804: '=' 
> WARNING    431: 'LEFT' 
> WARNING    278: '(' 
> WARNING    199: 'AND' 
> WARNING    189: 'WHEN' 
> WARNING    156: '/' 
> WARNING    151: 'THEN' 
> WARNING    109: 'END' 
> ```

I've just added another commit which should handle the case of "=" in there (probably also the `/` which is there too). I've made a guess on how to solve the `.` case too - although that one is one that I'm less confident in. Are most of your dots in identifiers? Has my most recent commit helped or hindered?

There are a couple of edge cases where this is currently too eager - I've definitely noticed something up with `VALUES` clauses and the commas which isn't parsing right at the moment. Will need more digging to resolve.

alanmcruickshank:
@WittierDinosaur - tests are passing now and I've done some tidyup. The changes highlighted a few places in the dialect which were being parsed inconsistently - which is why you'll see some changes to the yml tests.

I'm getting ~20% improvement in parsing speed on the `bench_002` file. I wonder whether you might be getting even better results in some others. I'm wondering whether to just bank these gains or whether to keep improving before merge - given this isn't yet a _massive_ code change I'm tempted to get this reviewed and in and then we can further optimise on other files.

WittierDinosaur (approved):
Huge!

PR #4988, opened 2023-07-20
Reduce copying in the parse phase

alanmcruickshank:
In investigating #4981 I also found that the `.copy()` method was being called a lot and causing a meaningful overhead during the parsing phase of operation. This was because on any segment `match()` operation, the resulting segments we always copied to ensure isolation.

At the reposition phase, the only _necessary_ copying I think is of the meta segments which may get moved around - but during parsing, the location of any substantive segments should not change - and so shouldn't be attempted. The test suite will confirm whether that's a safe assumption - if not we'll need to take a different tack on this approach.

On large files, this saves almost 50% of the parsing time 😱 .


WittierDinosaur (approved):
LGTM - 30% improvement on big files

Issue #4968, opened 2023-07-14
Snowflake: Found unparsable section: 'AS cross'


cbrswt:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

SQLFluff is throwing parsing error on `AS cross`, treating `cross` AS _code_ instead of _naked\_identifier.

### Expected Behaviour

SQLFluff is able to parse `AS cross` for Snowflake dialect.

### Observed Behaviour

SQLFluff is raising parsing error:
```bash
L:   5 | P:  23 |  PRS | Line 5, Position 23: Found unparsable section: 'AS cross'
```

### How to reproduce

1. Create a following dbt model in `.sql` file:
```sql
{{ config(materialized = "table") }}

SELECT
    customer_id,
    TRIM(value:cross) AS cross
FROM
    {{ source('jaffle_shop', 'customers') }}
```
2. Run `sqlfluff lint path/to/the/model.sql`

### Dialect

Snowflake

### Version

```bash
$ pip freeze | grep "sqlfluff"
sqlfluff==2.1.2
sqlfluff-templater-dbt==2.1.2
```

but it happens on `2.1.1` too.

### Configuration

```toml
[tool.sqlfluff.core]
templater = "dbt"
dialect = "snowflake"
large_file_skip_byte_limit = 25000
processes = 0
exclude_rules = "LT05,AL07,ST06"

[tool.sqlfluff.templater.jinja]
apply_dbt_builtins = true
load_macros_from_path = "macros"

[tool.sqlfluff.layout.type.alias_expression]
spacing_before = "align"
align_within = "select_clause"
align_scope = "bracketed"

[tool.sqlfluff.rules.'capitalisation.keywords']
capitalisation_policy = "upper"

[tool.sqlfluff.rules.'capitalisation.identifiers']
extended_capitalisation_policy = "lower"
ignore_words_regex = "^(METADATA\\$\\w+)$"

[tool.sqlfluff.rules.'references.keywords']
ignore_words = "date,value,name,url"

[tool.sqlfluff.rules.'capitalisation.functions']
extended_capitalisation_policy = "upper"

[tool.sqlfluff.rules.'capitalisation.literals']
capitalisation_policy = "upper"

[tool.sqlfluff.rules.'references.special_chars']
additional_allowed_characters = "$"

[tool.sqlfluff.rules.'capitalisation.types']
[...]

PR #4996, opened 2023-07-24
Correct Snowflake `CROSS JOIN` syntax

tunetheweb:

Better fix for #4968 

The [snowflake keyword snytax](https://docs.snowflake.com/en/sql-reference/reserved-keywords) says this about `CROSS`:

```
Cannot be used as table name or alias in a FROM clause.
```

So looks like it can be used as a column name as requested by #4968 and implemented in #4975.

This PR excludes it from table aliases only.

No more than original fix :-)

### Pull Request checklist

  - Other.


Martin-Carlsson:
Thank you 🙏

Issue #4891, opened 2023-05-30
-- noqa ignored in multiline comment


daviewales:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

I have a very long URL inside a multiline comment.
I don't want to trigger a rule failure, so I tried putting `-- noqa: LT05` at the end of the line.
But because we are already inside a multiline comment, the `noqa` declaration was ignored.

### Expected Behaviour

There should be a way to specify `noqa` for comment lines. (See also #4881 which would be a better solution in this particular case.)

### Observed Behaviour

`sqlfluff lint` outputs the following failure even when `-- noqa: LT05` is appended to the long comment line:

    L:   2 | P:   1 | LT05 | Line is too long (106 > 88).

### How to reproduce

Run `sqlfluff lint` on the following SQL file:

``` sql
/*
* https://example.com/1111111111111111111111111111111111111111111111111111111111111111111111 -- noqa: LT05
*/
```

### Dialect

tsql

### Version

sqlfluff, version 2.1.0
Python 3.10.10

### Configuration

``` cfg
[sqlfluff]
dialect = tsql
max_line_length = 88

# ST06 reorders columns, which seems presumptuous
exclude_rules = ST06

[sqlfluff:indentation]
allow_implicit_indents = True

[sqlfluff:rules:references.special_chars]
allow_space_in_identifier = True
additional_allowed_characters = ['-/()?,']
```

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

daviewales:
I've tracked it down to this line:
https://github.com/sqlfluff/sqlfluff/blob/7e4843427b8e34ccdd0909b55f16842159027f80/src/sqlfluff/core/linter/linter.py#L451
If I just remove that check, it happily follows `-- noqa` directives at the end of lines inside multiline comments.
I'm not sure if this has unintended consequences however.

From what I can tell, each line of a multiline comment is passed one at a time to `parse_noqa()`, which allows the function to process it the same as an inline comment.

Happy to make a PR with this change if there are no objections.

Robert-Kosmala-FLYR:
hi @daviewales,

[...]

PR #5010, opened 2023-07-27
Refactor: Extract noqa methods and tests.

alanmcruickshank:
This PR looks big, but contains no functional changes - it's just an extraction refactor.

It supports #2337 and #4891, but does not make progress on adding those features.

The rationale for doing this is:
- To make progress on #2337, we'll need to change how ignores are applied and managed. That means having it isolated and interacting with a single interface class (the new `IgnoreMask`) is going to be a necessary step.
- The `Linter()` class, and by extension the `linter.py` and `linter_test.py` files are already very bloated (over 1000 lines for `linter.py` as an example). The noqa methods within `Linter()` are also very self isolated and only called in that context. Likewise, the tests for noqa methods within `linter_test.py` were also substantial.
- There is already a submodule within the test suite for `test/core/parser/...`, but linter tests were all in files directly within `test/core/...`.

This PR does the following:
1. Extracts all the noqa methods from `Linter()` and `LintedFile()` and moves that logic into a new `IgnoreMask` class which is instantiated with the `LintedFile()` but holds an interface to any ignore logic. All of this now resides in `sqlfluff.core.linter.noqa` as a new module (along with the logic for `NoQaDirective` from `sqlfluff.core.linter.common`.
2. Moves `linter_test.py` and `linted_file_test.py` into a new submodule, and splits out any noqa tests into `noqa_test.py`. The final `linter_test.py` is now almost half the length.
3. Updates the docstrings the comments to reflect these changes (and catches some docstrings which were already out of date).


test/core/linter/linter_test.py
```diff
@@ -963,59 +501,6 @@ def test__attempt_to_change_templater_warning():
...
-    try:
-        linter.LintedFile._safe_create_replace_file(
-            str(p), str(p), case["update"], case["encoding"]
-        )
-    except:  # noqa: E722
-        pass
-    actual = p.read_text(encoding=case["encoding"])
-    assert case["expected"] == actual
-
-
```
alanmcruickshank, line 1018: Worth noting that this test isn't noqa related, but should really by in `linted_file_test.py` so I've moved it while we're here. It's not been deleted 👍

WittierDinosaur (approved):
All looks relatively straightforward. Also, pretty confident in the lack of breaking changes, as the tests that already exist are passing

PR #5013, opened 2023-07-27
Assorted typehints

alanmcruickshank:
Another one from exploring mypyc - these are a set of missing typehints that I added in the process.

No functionality. Just type hints 🎸


WittierDinosaur (approved):
Lovely

PR #5014, opened 2023-07-27
A step towards mypy strict

WittierDinosaur:
This PR is a step towards mypy strict compatibility
Before
`Found 5463 errors in 155 files (checked 209 source files)`
After
`Found 1050 errors in 135 files (checked 210 source files)`

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/parser/parsers.py
+++ b/src/sqlfluff/core/parser/parsers.py
@@ -9,6 +9,7 @@ import regex
 from typing import Collection, Type, Optional, Tuple, Union
 
 from sqlfluff.core.parser.context import ParseContext
+from sqlfluff.core.parser.grammar.base import SimpleHintType
 from sqlfluff.core.parser.matchable import Matchable
 from sqlfluff.core.parser.match_result import MatchResult
 from sqlfluff.core.parser.segments import RawSegment, BaseSegment
@@ -123,9 +124,7 @@ class TypedParser(BaseParser):
             **segment_kwargs,
         )
 
-    def simple(
-        cls, parse_context: ParseContext, crumbs=None
-    ) -> Tuple[frozenset, frozenset]:
+    def simple(cls, parse_context: ParseContext, crumbs=None) -> SimpleHintType:
         """Does this matcher support a uppercase hash matching route?
 
         TypedParser segment doesn't support matching against raw strings,
@@ -197,9 +196,7 @@ class MultiStringParser(BaseParser):
             **segment_kwargs,
         )
 
-    def simple(
-        self, parse_context: "ParseContext", crumbs=None
-    ) -> Tuple[frozenset, frozenset]:
+    def simple(self, parse_context: "ParseContext", crumbs=None) -> SimpleHintType:
         """Return simple options for this matcher.
 
         Because string matchers are not case sensitive we can
--- a/src/sqlfluff/core/parser/parsers.py
+++ b/src/sqlfluff/core/parser/parsers.py
@@ -9,7 +9,7 @@ import regex
 from typing import Collection, Type, Optional, Tuple, Union
 
 from sqlfluff.core.parser.context import ParseContext
-from sqlfluff.core.parser.grammar.base import SimpleHintType
+from sqlfluff.core.parser.grammar.types import SimpleHintType
 from sqlfluff.core.parser.matchable import Matchable
 from sqlfluff.core.parser.match_result import MatchResult
 from sqlfluff.core.parser.segments import RawSegment, BaseSegment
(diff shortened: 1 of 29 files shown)


src/sqlfluff/core/parser/parsers.py
```diff
@@ -123,15 +123,17 @@ def __init__(
             **segment_kwargs,
         )
 
-    def simple(cls, parse_context: ParseContext, crumbs=None):
+    def simple(
+        cls, parse_context: ParseContext, crumbs=None
+    ) -> Tuple[frozenset, frozenset]:
```
alanmcruickshank, line 128: To keep things consistent around this more convoluted type, I recently added `SimpleHintType` to characterise the return values of `simple` types. It's currently in `grammar/base.py`, although for this implementation that definition should probably move somewhere more central. I think it's a complicated enough type that we should use that single defined type for consistency.

I won't comment all the other like this - but there are a few.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1056,14 +1056,14 @@ def iter_unparsables(self):
         for s in self.segments:
             yield from s.iter_unparsables()
 
-    def type_set(self):
+    def type_set(self) -> set:
```
alanmcruickshank, line 1059: This one is a set of strings. `Set[str]` would be a more helpful hint.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1074,7 +1074,7 @@ def get_child(self, *seg_type):
                 return seg
         return None
 
-    def get_children(self, *seg_type):
+    def get_children(self, *seg_type) -> list:
```
alanmcruickshank, line 1077: `List["BaseSegment"]` would be more helpful here.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1088,7 +1088,7 @@ def select_children(
         stop_seg: Optional["BaseSegment"] = None,
         select_if: Optional[Callable[["BaseSegment"], Any]] = None,
         loop_while: Optional[Callable[["BaseSegment"], Any]] = None,
-    ):
+    ) -> list:
```
alanmcruickshank, line 1091: Ditto `List["BaseSegment"]` here too.

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -139,25 +139,25 @@ def source_fixes(self) -> List[SourceFix]:
...
 
-    def is_type(self, *seg_type):
+    def is_type(self, *seg_type) -> bool:
         """Extend the parent class method with the surrogate types."""
         if self._surrogate_type and self._surrogate_type in seg_type:
             return True
         return self.class_is_type(*seg_type)
 
-    def get_raw_segments(self):
+    def get_raw_segments(self) -> list:
```
alanmcruickshank, line 156: This would be more helpful as `List["RawSegment"]`

src/sqlfluff/core/parser/segments/raw.py
```diff
@@ -177,16 +177,16 @@ def raw_trimmed(self):
             return raw_buff
         return raw_buff
 
-    def raw_list(self):  # pragma: no cover TODO?
+    def raw_list(self) -> list:  # pragma: no cover TODO?
         """Return a list of the raw content of this segment."""
         return [self.raw]
```
alanmcruickshank, line 182: This method can just be deleted. It's not used anymore. It has a counterpart in `BaseSegment` which should also go. They're replaced by `.raw_segments`.

src/sqlfluff/rules/references/RF04.py
```diff
@@ -110,7 +110,7 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
         else:
             return None
 
-    def _init_ignore_string(self):
+    def _init_ignore_string(self) -> list:
```
alanmcruickshank, line 113: This would be more helpful as `List[str]`

src/sqlfluff/rules/convention/CV09.py
```diff
@@ -101,7 +101,7 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
 
         return None
 
-    def _init_blocked_words(self):
+    def _init_blocked_words(self) -> list:
```
alanmcruickshank, line 104: Ditto more helpful as `List[str]`

alanmcruickshank (changes requested):
I've run through all the changes. Looks like a pretty good haul!

There a few a places where a more specific type would be more helpful (although I've tried to not make it more specific everywhere, just the ones which people are likely to encounter and use as documentation). Where that's the case I've tried to suggest what the type hint should be instead so you don't have to figure it out yourself.

src/sqlfluff/core/parser/segments/base.py
WittierDinosaur: changing this actually causes errors in vanilla mypy - i think adding these types will show edge cases in our code that we'll need to address

alanmcruickshank (approved):
LGTM 👍

PR #5019, opened 2023-07-28
More mypy strict

WittierDinosaur:
More mypy strict compatibility

Now up to
`Found 751 errors in 99 files (checked 211 source files)`

No


alanmcruickshank (commented):
Looks pretty good! Linting failing but otherwise I think this is good to go 👍

alanmcruickshank (approved):
Nice 🚀

PR #5020, opened 2023-07-28
Streamline imports to reduce initial load times #4917

### LINKED ISSUE #4917: Initial SQLFluff invocation loads many unnecessary plugins
### Search before asking

### What Happened

I am using sqlfluff.parse to perform validations on LLM generated SQL statements. Seems like on the first time it is called it is taking too long to run.

### Expected Behaviour

sqlfluff.parse should not take over 2 seconds on first invocation

### Observed Behaviour

The first time parse is called, it takes quite long (2-3 seconds) while following executions are much much shorter (around 1/10 of second).
This is obviously better than all invocation taking that long, but I wonder if there is a way to reduce the time of the first call as well, or maybe even know what's causing it to take so long.
I further tested this behavior by adding sqlfluff.parse("") after module load, which took the initial long time and afterwards later invocations were fast again.

### How to reproduce

Wrap each invocation of sqlfluff.parse with something simple like:
start_time = timeit.default_timer()
sqlfluff.parse("")
end_time = timeit.default_timer()
print(f"1 - The function took {end_time - start_time} seconds to complete.")

start_time = timeit.default_timer()
sqlfluff.parse("")
end_time = timeit.default_timer()
print(f"2 - The function took {end_time - start_time} seconds to complete.")

And we'll be able to see that the second time works much faster

### Dialect

ansi

### Version

sqlfluff, version 2.1.1
Python 3.9.16

### Configuration

Using default

### Code of Conduct

alanmcruickshank:
This PR resolves #4917 (or at least makes as much progress on it as I think is reasonably possible).

The problem was in how plugins are loaded.

1. All runs of SQLFluff (whether CLI or API) create a `FluffConfig`
2. Any `FluffConfig` will load the plugin manager.
   - This was the first issue, it was recursively re-loading itself many times, because each plugin _also_ loads the plugin manager **on load** 🙄 . I've added some sensible caching to streamline this.
3. The plugin manager also loads all of the plugins. This means that the `__init__.py` of every registered plugin is called. Most plugins so far import all of the submodules within that file - and so this means the entirety of each plugin is **also** loaded on every run.
   - My second change was to move _most_ of the import methods within the plugins into one of the methods. This means that submodules are only loaded on demand. This is most obvious on the rule plugins.
4. The `FluffConfig` finally calls `get_templaters()` on all of the plugins so that in know what templaters it can choose from (and this kind of makes sense because it wants to select by name and it needs to load them all to know the names).
   - For the `dbt` templater this was the biggest hit, because the `dbt` package takes _a long time_ to load. This would mean a performance hit _whether or not the dbt templater was selected_ - as long as it's _installed_, `dbt` was getting loaded.
   - This is my third change (and probably the one with the biggest diff), where I've moved _all_ of the dbt imports into methods on the dbt templater. These are only called on _instatiation_ rather than on module load, so the performance hit of dbt is only taken if we've actually selected the dbt templater.

On my machine, this takes a cold `sqlfluff.parse("")` from taking ~5s with the dbt templater installed, down to ~0.2s. The improvement is less if the dbt plugin isn't installed, but the end result is a similar order of magnitude.

The OP of the issue (#4917) noted that any subsequent runs were much faster - that's still true (0.02s vs 0.1s on my machine in testing this PR) - and the multiple is _large_ (like x5), but a 0.1s import time for a python module seems fairly reasonable. The next thing to take a go at would be dialect expansion, but that's a job for another day.

This PR strongly recommends a specific import order for plugins. I've managed to make it so that plugins which _do_ import things in the way that we used to recommend do still work _however_ there will now be a warning point them to the docs (and so that we cover that eventuality and ensure that the old way is still supported, the example plugin keeps the _not_ recommended path). For rules that are imported in the _wrong_ order - we also skip the validation of config keywords as they won't yet be loaded. I've added details to the docs and to the example plugin to explain that.


docs/source/developingplugins.rst
```diff
@@ -8,6 +8,37 @@ to make linting Rules pluggable, which enable users to implement rules that
 are just too "organization specific" to be shared, or too platform specific
 to be included in the core library.
 
+.. note::
+
+    We recommend that the module in a plugin which defines all
+    of the hook implementations (anything using the ``@hookimpl`` decorator)
+    must be able to fully import before any rule implementations are imported.
+    More specifically, the SQLFluff must be able to both *import* **and**
```
WittierDinosaur, line 16: `the SQLFluff` doesn't read right

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -1,79 +1,52 @@
...
-from sqlfluff.cli.formatters import OutputStreamFormatter
-from sqlfluff.core import FluffConfig
 from sqlfluff.core.cached_property import cached_property
 from sqlfluff.core.errors import SQLTemplaterError, SQLFluffSkipFile, SQLFluffUserError
 
 from sqlfluff.core.templaters.base import TemplatedFile, large_file_check
 
 from sqlfluff.core.templaters.jinja import JinjaTemplater
 
+if TYPE_CHECKING:  # pragma: no cover
```
WittierDinosaur, line 41: remind me what this is for?

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -109,19 +85,42 @@ def __init__(self, **kwargs):
...
+    @cached_property
+    def dbt_version(self):
+        """Gets the dbt version."""
+        return self._dbt_version.to_version_string()
+
+    @cached_property
+    def dbt_version_tuple(self):
         """Gets the dbt version."""
-        return DBT_VERSION_STRING
+        return (int(self._dbt_version.major), int(self._dbt_version.minor))
```
WittierDinosaur, line 113: pretty sure the wrapping brackets are redundant when returning tuples

src/sqlfluff/rules/aliasing/__init__.py
```diff
@@ -2,16 +2,20 @@
...
 
 @hookimpl
 def get_rules():
-    """Get plugin rules."""
+    """Get plugin rules.
+
+    NOTE: Rules are imported only on fetch to manage import times
+    when rules aren't used.
+    """
+    from sqlfluff.rules.aliasing.AL01 import Rule_AL01
```
WittierDinosaur, line 13: can we not do `from sqlfluff.rules.aliasing import Rule_AL01, Rule_AL02...` here?

src/sqlfluff/rules/ambiguous/__init__.py
```diff
@@ -5,16 +5,20 @@
...
 
 @hookimpl
 def get_rules():
-    """Get plugin rules."""
+    """Get plugin rules.
+
+    NOTE: Rules are imported only on fetch to manage import times
+    when rules aren't used.
+    """
+    from sqlfluff.rules.ambiguous.AM01 import Rule_AM01
```
WittierDinosaur, line 16: etc

test/core/plugin_test.py
```diff
@@ -1,29 +1,61 @@
...
+
+    To do this we clear the plugin manager cache and also forcibly
+    unload the example plugin modules if they are already loaded.
+    This ensures that we can capture any warnings raised by importing the
+    module.
+    """
+    purge_plugin_manager()
+    try:
+        del sys.modules["sqlfluff_plugin_example"]
+    except KeyError:
```
WittierDinosaur, line 28: why are we doing try, excepts in a test?

docs/source/developingplugins.rst
alanmcruickshank: good catch - I'll remove the `the`. It's not necessary.

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
alanmcruickshank: Ah - this imports things for when `mypy` runs, but otherwise not. That's why it's not covered in tests. https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
alanmcruickshank: Right you are. Will fix 👍

src/sqlfluff/rules/aliasing/__init__.py
alanmcruickshank: Unfortunately not because they're all in their own source files. `Rule_AL01` comes from `AL01.py` and `Rule_AL02` comes from `AL02.py` etc...

src/sqlfluff/rules/ambiguous/__init__.py
alanmcruickshank: as above

test/core/plugin_test.py
alanmcruickshank: Ah - good question. It's because we may or may not be able to _remove_ the module depending on whether it's already been imported by _a different test_. The imported modules are shared between tests.

I'll add a comment to that effect.

WittierDinosaur (approved):
LGTM

PR #5022, opened 2023-07-30
Handle windows paths better in config files.

alanmcruickshank:
This is a small fix to handle Windows path better in config files (because they've got colons in them). In particular is also helps when the paths are provided in the new `_from_file()` methods (and where the config file itself also has no path).


src/sqlfluff/core/config.py
```diff
@@ -1211,16 +1239,14 @@ def process_inline_config(self, config_line: str, fname: str):
...
-        config_val = (tuple(config_path[:-1]), config_path[-1])
+        config_val = split_colon_separated_string(config_line)
         # Validate the value
         ConfigLoader._validate_configs([config_val], fname)
         # Set the value
         self.set_value(*config_val)
         # If the config is for dialect, initialise the dialect
-        if config_path[:-1] == ["dialect"]:
-            self._initialise_dialect(config_path[-1])
+        if config_val[0] == ("dialect",):
```
WittierDinosaur, line 1248:
```suggestion
        if config_val[0] == "dialect":
```

src/sqlfluff/core/config.py
```diff
@@ -296,6 +296,32 @@ def split_comma_separated_string(raw: Union[str, List[str]]) -> List[str]:
...
+            and element[0] == "\\"
+            and config_path[-1]
+            and len(config_path[-1]) == 1
+        ):
+            config_path[-1] = config_path[-1] + ":" + element
+        else:
+            # Otherwise just add it to the path.
+            config_path.append(element)
+
+    return (tuple(config_path[:-1]), config_path[-1])
```
WittierDinosaur, line 322:
```suggestion
    return tuple(config_path[:-1]), config_path[-1]
```

src/sqlfluff/core/config.py
alanmcruickshank: Unfortunately this one _does_ need to be a `tuple`. I got it wrong the first time and it caused a weird bug. I'll add a comment to explain.

WittierDinosaur (approved):
LGTM

PR #5025, opened 2023-07-31
Fix: Add exception to the warning & config for the BaseRule.

alanmcruickshank:
While building docs, the warning for rule configuration occasionally flags and that's because of the import path for the `BaseRule` (which shouldn't apply any configuration _anyway_). This provides an exception for that and explicitly _doesn't_ check for any of this when we're instantiating the `BaseRule`.


WittierDinosaur (approved):
LGTM

PR #5029, opened 2023-07-31
Warn on unused `noqa` directives

### LINKED ISSUE #2337: Warn on unused inline ignores

alanmcruickshank:
This resolves #2337 (an oldie). I've tried to mirror the `mypy` syntax as closely as possible.

My algorithm for what counts as "used" has a few loopholes in it, but I think they're manageable for now (e.g. We count an `enable` as _used_ when it successfully re-enables something disabled by a `disable` - but this only evaluation happens if we get close enough to it in the file (usually because there was an error within the span, or after it. If we never get there, we never evaluate it and _both_ the enable and disable will be deemed unnecessary).

The new test case in `commands_test.py` is probably the best place to see what the effect of this is.

@tunetheweb I know you had views on the original PR here so would be curious for your review on this one.

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -23,6 +23,8 @@ ignore = None
 # Warn only for rule codes (one of more rule codes, seperated by commas: e.g. LT01,LT02)
 # Also works for templating and parsing errors by using TMP or PRS
 warnings = None
+# Whether to warn about unneeded '-- noqa:' comments.
+warn_unused_ignores = False
 # Ignore linting errors found within sections of code coming directly from
 # templated code (e.g. from within Jinja curly braces. Note that it does not
 # ignore errors from literal code found within template loops.
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -359,6 +359,12 @@ def lint_options(f: Callable) -> Callable:
             "future releases without warning."
         ),
     )(f)
+    f = click.option(
+        "--warn-unused-ignores",
+        is_flag=True,
+        default=False,
+        help="Warn about unneeded '-- noqa:' comments.",
+    )(f)
     return f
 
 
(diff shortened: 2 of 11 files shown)


src/sqlfluff/core/default_config.cfg
```diff
@@ -23,6 +23,8 @@ ignore = None
 # Warn only for rule codes (one of more rule codes, seperated by commas: e.g. LT01,LT02)
 # Also works for templating and parsing errors by using TMP or PRS
 warnings = None
+# Whether to warn about unneeded '-- noqa:' comments.
+warn_unused_ignores = False
```
tunetheweb, line 27: I wonder if this should be defaulted to True?

I know I had concerns, but as long as it's a warning and doesn't exit error I think that's OK.

tunetheweb (approved):
I like it. One comment.

src/sqlfluff/core/default_config.cfg
alanmcruickshank: Good question 🤔 . I'll merge it default "off" for now - because it's new and that matches the behaviour of mypy most closely. It also has the benefit of allowing more extensive testing for now.

PR #5030, opened 2023-08-01
MyPy on errors, helpers, markers & context + remove ParseContext.denylist

alanmcruickshank:
This makes progress on adding more mypy type annotations.

In the process I did some testing on the `denylist` in `ParseContext`, and found that now that we've implemented more effective caching in #4576, this older mechanism for caching is an unecessary overhead.

It was poorly covered in the test suite and I've done testing on large files and performance is slightly better without it now.


WittierDinosaur (approved):
LGTM

PR #5032, opened 2023-08-01
Fix for multiprocessing warnings.

alanmcruickshank:
This resolves another issue created by #5020 .

When running in a multiprocessing configuration, the error about import order was flagging unnecessarily. This further restricts that validation so that it only happens in the main process.


WittierDinosaur (approved):
LGTM

PR #5033, opened 2023-08-01
strict mypy: match_wrapper & match_logging

alanmcruickshank:
Strict mypy for `match_wrapper.py` and `match_logging.py`. This is a relatively minimal PR. There's a larger version of this which would include `ephemeral.py` too, but that's a much bigger change.


WittierDinosaur (approved):
LGTM

PR #5036, opened 2023-08-02
mypy strict in config

pwildenhain:
There's a lot of `Dict[str, Any]` running around for the config. I know that it can't _really_ be any, since it's more like an arbitrarily nested dict of dicts. But `--strict` doesn't like having just `dict` as a type, so if anyone has other suggestions besides `Dict[str, Any]` I'm happy to implement it.

I'm not great with all the intricacies of typing, and this article was very helpful: https://realpython.com/python-type-checking/


alanmcruickshank (approved):
LGTM! 👍

PR #5044, opened 2023-08-03
Mypy: Ephemeral + Tuple Return on .parse()

alanmcruickshank:
This is another internal refactor as part of the mypy saga.

Importantly I found a bit of confusion in what the `BaseSegment.parse()` method should return, where in most cases it just returns a single segment, in the case of the `EphemeralSegment`, it returns a tuple. What I've done is update _all_ of them to return a tuple - which allows us to do a little less type munging.

The general upshot of this is a lot less type conversion. Stricter mypy yes, and some more awkward type _assertions_ but a lot let conversion overhead.


alanmcruickshank:
@pwildenhain - given you seem keen on this one, I've added you as a reviewer :) If you think it looks fine, then feel free to merge at will so you can crack on with yours.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -821,31 +797,10 @@ def match(
...
-                except TypeError as err:  # pragma: no cover
-                    # This is an error to assist with debugging dialect design.
-                    # It's most likely that the match_grammar has been set on
-                    # a raw segment which shouldn't happen.
-                    raise TypeError(
-                        f"Error in instantiating {cls.__module__}.{cls.__name__}. Have "
-                        f"you defined a match_grammar on a RawSegment? : {str(err)}"
-                    )
+                return MatchResult(
+                    (cls(segments=m.matched_segments),), m.unmatched_segments
```
WittierDinosaur, line 802: not relevant to your PR, but is this cls() needed?

alanmcruickshank: Ah - yes it is. But I'll add a comment to explain why.

WittierDinosaur (approved):
LGTM

PR #5045, opened 2023-08-03
Mypy gain: Remove unnecessary tuple construction in MatchResult

alanmcruickshank:
This is a performance improvement which stems from the `mypy --strict` work. We were doing some type munging on every instantiation of a `MatchResult` - this is a very simple change but it saves one of the top items on the `cProfile` output. Saves about 10% on big files.


WittierDinosaur (approved):
Lovely

PR #5046, opened 2023-08-03
Less copying in the ParseContext

alanmcruickshank:
Another possible performance improvement from the mypy work. Saves ~10% on large files.

In the ParseContext object, we made copies as we move up and down the parsing structure. This does some more "smart" management of the context (and uses `contextlib` instead) to avoid all the copying, and just use context managers properly instead. I've also consolidated the way that people interact with the context too.

At this point, the distinction between `ParseContext` and `RootParseContext` makes no sense - so I've merged the two together and simplified the whole module as a result. There is now only `ParseContext`.

I didn't manage _total_ isolation, so I still do some tuple manipulation to add some safety, because getting lists to play nice was too hard.


WittierDinosaur (approved):
Much neater 👍

PR #5047, opened 2023-08-03
:white_check_mark: Add strict typing for errors module

pwildenhain:
Squashing a small one here 😁 

I needed to use `Tuple[Any, ...]` for typing `__reduce__` since everything is a subclass of  `SQLBaseError` and `mypy` didn't like us having different return values for inherited methods

I also this that `mypy` helped us find a bug -- the `__reduce__` for `SQLBaseError` was referencing `self.pos` which doesn't exist. I replaced it here with `None` which seems appropriate

### DIFF
--- a/src/sqlfluff/core/errors.py
+++ b/src/sqlfluff/core/errors.py
@@ -47,11 +47,12 @@ class SQLBaseError(ValueError):
             self.line_pos = line_pos
         super().__init__(self.desc())
 
-    def __reduce__(self):  # pragma: no cover
+    def __reduce__(
+        self,
+    ) -> Tuple[type["SQLBaseError"], Tuple[Any, ...]]:  # pragma: no cover
         """Prepare the SQLBaseError for pickling."""
         return type(self), (
             self.description,
-            self.pos,
             self.line_no,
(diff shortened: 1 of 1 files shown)


src/sqlfluff/core/errors.py
```diff
@@ -47,11 +47,13 @@ def __init__(
...
 
-    def __reduce__(self):  # pragma: no cover
+    def __reduce__(
+        self,
+    ) -> Tuple[Type["SQLBaseError"], Tuple[Any, ...]]:  # pragma: no cover
         """Prepare the SQLBaseError for pickling."""
         return type(self), (
             self.description,
-            self.pos,
+            None,
```
WittierDinosaur, line 56: This is a positional arg, does anything tnat's using this expect something for it?

pwildenhain: I don't think so? This method is just for pickling, and the `pos` is just used creating for the initial instance., after which is splits that into line number and position. 

We also type the `pos` arg as `Optional` so that means it should be good to be `None`

WittierDinosaur (approved):
LGTM

PR #5048, opened 2023-08-03
Add mypy strict typing for sqlfluff.core.rules

pwildenhain:
FYI @alanmcruickshank I'll hold off on doing anymore work on this until you've got #5044 settled, since it seems like that's going to shake things up slightly

### DIFF
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -24,6 +24,9 @@ import pathlib
 import regex
 import re
 from typing import (
+    TYPE_CHECKING,
+    Sequence,
+    Sized,
     cast,
     Iterable,
     Optional,
@@ -52,6 +55,16 @@ from sqlfluff.core.rules.config_info import get_config_info
 from sqlfluff.core.plugin.host import plugins_loaded, is_main_process
 from sqlfluff.core.templaters.base import RawFileSlice, TemplatedFile
 
+# Best solution for generic types on older python versions
+# https://github.com/python/typeshed/issues/7855
+if TYPE_CHECKING:  # pragma: no cover
+    from sqlfluff.core.config import FluffConfig
+    from sqlfluff.core.plugin.hookspecs import PluginSpec
+
+    _LoggerAdapter = logging.LoggerAdapter[logging.Logger]
+else:
+    _LoggerAdapter = logging.LoggerAdapter
+
 # The ghost of a rule (mostly used for testing)
 RuleGhost = namedtuple("RuleGhost", ["code", "name", "description"])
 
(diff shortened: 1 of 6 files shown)


alanmcruickshank:
I've just had a quick flick through this, and I'll give it a full review this weekend. There's a few things in here I want to take a closer look at. Well done for tackling this module! Not an easy one to do 🚀

src/sqlfluff/core/rules/base.py
```diff
@@ -97,7 +110,7 @@ def __init__(
         self,
         anchor: Optional[BaseSegment] = None,
         fixes: Optional[List["LintFix"]] = None,
-        memory=None,
+        memory: Optional[Iterable[Any]] = None,
```
alanmcruickshank, line 113: This is a tricky one. The original _intent_ of `memory` was that it was whatever data structure a given rule wanted it to be. That isn't very helpful for mypy though. Does `Optional[Any]` work here?

src/sqlfluff/core/rules/base.py
```diff
@@ -237,18 +250,21 @@ def is_just_source_edit(self) -> bool:
...
-    def __repr__(self):
+    def __repr__(self) -> str:
         if self.edit_type == "delete":
             detail = f"delete:{self.anchor.raw!r}"
         elif self.edit_type in ("replace", "create_before", "create_after"):
             if hasattr(self.edit, "raw"):
-                new_detail = self.edit.raw  # pragma: no cover TODO?
+                seg = cast(BaseSegment, self.edit)  # pragma: no cover
+                new_detail = seg.raw  # pragma: no cover TODO?
             else:
```
alanmcruickshank, line 260: Now that we've got stronger typing in place, I think that we can remove this first clause. `edit` is defined as `Optional[Iterable[BaseSegment]]` so it should never have a `raw` attribute (which is from the days when we accepted a single segment).

(We also coerce it to a `list` in the `__init__` method)

src/sqlfluff/core/rules/base.py
```diff
@@ -260,13 +276,19 @@ def __repr__(self):
...
         A fix is equal to another if is in the same place (position), with the
         same type and (if appropriate) the same edit values.
 
         """
+        # We have to assert this here rather in the type annotation so we don't
+        # violate the Liskov substitution principle.
+        # More context here: https://stackoverflow.com/a/37557540/11381493
+        if not isinstance(other, LintFix):  # pragma: no cover
+            return NotImplemented
+
```
alanmcruickshank, line 291: You learn something every day! 😄

alanmcruickshank (commented):
Actually - on closer inspection, a lot of this makes sense. Sounds like you've had to wrestle with a few tricky cases here. Two point that I do have questions about, but the rest looks great!

alanmcruickshank (approved):
LGTM 👍

PR #5055, opened 2023-08-05
:arrow_up: Bump mypy version in pre-commit

pwildenhain:
Right now it errors for me on the older version of `mypy`, when I bring it up to the current version it passes just fine


WittierDinosaur (approved):
LGTM

PR #5063, opened 2023-08-07
Move `_prune_options` within `_longest_trimmed_match`.

alanmcruickshank:
Hopefully a nice small and tidy PR.

This moves the `_prune_options` method (and `_first_non_whitespace` along with it) from the `AnyOf` grammar up into `BaseGrammar`. I've also moved it _inside_ `_longest_trimmed_match()` which means it's automatically used in `Delimited` as well as `AnyOf` or `AnyNumberOf`. Performance impact should be positive, but small.

Part of this is for simplification, but I also hope to take another pass at the parse statistics to support better optimisation there, and having the pruning operation within that step should help significantly with doing that.

This led to finding an issue in the `ansi` dialect (related to some `clickhouse` statements) - so that's why I've touched one of the dialect and yaml.

### DIFF
--- a/src/sqlfluff/core/parser/grammar/anyof.py
+++ b/src/sqlfluff/core/parser/grammar/anyof.py
@@ -1,6 +1,6 @@
 """AnyNumberOf and OneOf."""
 
-from typing import List, Optional, Tuple, Set
+from typing import List, Optional, Tuple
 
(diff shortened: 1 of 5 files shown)


WittierDinosaur (approved):
Much tidier - LGTM

PR #5067, opened 2023-08-07
✅ Strictly type dialect

pwildenhain:
Starting to type this out, but I can tell I'm going to have to understand the parser/dialects a lot better to do this right. 

Right the `--strict` check succeeds in just the dialects, but fails the overall `mypy` check in the project, the next step is to look into those 👍🏻


src/sqlfluff/core/dialects/base.py
```diff
@@ -230,7 +231,8 @@ def add_update_segments(self, module_dct: Dict[str, Any]) -> None:
                 if k not in self._library:
                     self.add(**{k: v})
                 else:
-                    self.replace(**{k: v})
+                    non_seg_v = cast(Union[Matchable, SegmentGenerator], v)
+                    self.replace(**{k: non_seg_v})
```
alanmcruickshank, line 235: This looks wrong. We've just asserted that it's a `BaseSegment` above, so it can't also be a `Union[Matchable, SegmentGenerator]` here. I'm assuming you put this in because you were otherwise getting errors - do you know what errors you were getting? There might be another solution.

alanmcruickshank (changes requested):
I wonder whether some of the errors you're getting in mypy might be resolved by some of the things I've noted in a first pass through.

This definitely won't be the easiest module, but well done for taking it on!

src/sqlfluff/core/dialects/base.py
pwildenhain: This must've been left over from when I was trying to get stuff to work. I removed it without issue

src/sqlfluff/core/dialects/base.py
```diff
@@ -255,7 +257,7 @@ def get_segment(self, name: str) -> Type["BaseSegment"]:
         """
         if name not in self._library:  # pragma: no cover
             raise ValueError(f"Element {name} not found in dialect.")
-        segment = self._library[name]
+        segment = cast(Type["BaseSegment"], self._library[name])
```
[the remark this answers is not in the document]
pwildenhain: I think this works, I had to create a separate variable and wrap it in `type` in order to get `mypy` to narrow down the types (see [this issue](https://github.com/python/mypy/issues/10978) for more details). but that seems to do it!

pwildenhain:
Ok so there's still a `mypy` error in here around unsafe overload variants, but I can't seem to figure out how to give `mypy` what it wants. Since the `Dialect.sets()` return type is very different for brackets, overloading seemed like the best bet rather than `cast`ing everywhere in the codebase.

Calling for help on this one 🏳️

alanmcruickshank:
> Since the `Dialect.sets()` return type is very different for brackets, overloading seemed like the best bet rather than `cast`ing everywhere in the codebase.

I see what you mean 🤔 .

I'm wondering whether a different approach might be better here. Rather than overloading `sets()` and making it more complicated, if _all_ the other uses of sets return string, but the bracket sets have a different format - should we just make a totally seperate flow for bracket sets, i.e. that `Dialect()` implements `sets() -> Set[str]` and also something like `bracket_sets() -> Set[Tuple[str, str, str, bool]]`. That would make type hints easier - and perhaps also more robust.

Thoughts?

pwildenhain:
@alanmcruickshank Should be all set for a review, great recommendation 👍🏻 

FYI I'll be on vacation for the next week starting tomorrow, so if we don't get this merged before then, feel free to take over this branch and make any changes needed to merge.

alanmcruickshank (approved):
Nice 👍

Issue #4938, opened 2023-06-29
[postgres] Unparsable section for expressions in LIMIT or OFFSET


enote-kane:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Lint the following:
```sql
SELECT col_a
FROM test_table
LIMIT 10 * 10 OFFSET (5 * 100)
```

### Expected Behaviour

Passes

### Observed Behaviour

Raises an error:
```
L:   3 | P:   9 |  PRS | Line 3, Position 9: Found unparsable section: ' * 10
                       | OFFSET (5 * 100)'
```

### How to reproduce

Use this SQL:

```sql
SELECT col_a
FROM test_table
LIMIT 10 * 10 OFFSET (5 * 100)
```

### Dialect

postgres

### Version

2.1.1

### Configuration

_default_

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

joaostorrer:
This can be closed, since it was fixed and already merged

PR #5068, opened 2023-08-07
fixing limit handling for bracketed arithmathic operations

adityapat3l:
Adding logic to fix issue [#4938 ](https://github.com/sqlfluff/sqlfluff/issues/4938).

Handles cases for postgres well.

File now parses:

example.sql

```
SELECT col_a
FROM test_table
LIMIT 2 * 5 * 10 OFFSET (5 + 10);


SELECT col_a
FROM test_table
LIMIT ALL OFFSET 10 - 5;
```

```
% sqlfluff lint --dialect postgres example.sql
All Finished 📜 🎉!                    
```

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -1429,6 +1429,58 @@ class SelectStatementSegment(ansi.SelectStatementSegment):
     )
 
 
+class LimitClauseSegment(BaseSegment):
+    """A `LIMIT` clause like in `SELECT`."""
+
+    type = "limit_clause"
+    match_grammar: Matchable = Sequence(
+        "LIMIT",
+        Indent,
+        OneOf(
+            OptionallyBracketed(
+                Sequence(
+                    # Allow a number by itself OR
+                    Ref("NumericLiteralSegment"),
+                    # Arithmetic operations followed by a number any number of times
+                    AnyNumberOf(
+                        Sequence(
+                            Ref("ArithmeticBinaryOperatorGrammar"),
+                            Ref("NumericLiteralSegment"),
+                        ),
+                        optional=True,
+                    ),
+                )
+            ),
+            "ALL",
+        ),
+        OneOf(
+            Sequence(
+                "OFFSET",
+                OptionallyBracketed(
+                    Sequence(
+                        # Allow a number by itself OR
+                        Ref("NumericLiteralSegment"),
+                        # Arithmetic operations followed by a number any number of times
+                        AnyNumberOf(
+                            Sequence(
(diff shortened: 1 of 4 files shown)


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1429,6 +1429,58 @@ class SelectStatementSegment(ansi.SelectStatementSegment):
...
+                    Ref("NumericLiteralSegment"),
+                    # Arithmetic operations followed by a number any number of times
+                    AnyNumberOf(
+                        Sequence(
+                            Ref("ArithmeticBinaryOperatorGrammar"),
+                            Ref("NumericLiteralSegment"),
+                        ),
+                        optional=True,
+                    ),
+                )
```
alanmcruickshank, line 1452: I think if we're going to extend this we should do something more generic. This also doesn't cover all the possibilities.

Keeping the `Ref("NumericLiteralSegment")` though is still sensible because that's a much faster parse option.

Otherwise I think is should be:
```suggestion
                OneOf(
                    # Allow a number by itself OR
                    Ref("NumericLiteralSegment"),
                    # An arbitrary expression
                    Ref("ExpressionSegment"),
                )
```

src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -1429,6 +1429,58 @@ class SelectStatementSegment(ansi.SelectStatementSegment):
...
+                        Ref("NumericLiteralSegment"),
+                        # Arithmetic operations followed by a number any number of times
+                        AnyNumberOf(
+                            Sequence(
+                                Ref("ArithmeticBinaryOperatorGrammar"),
+                                Ref("NumericLiteralSegment"),
+                            ),
+                            optional=True,
+                        ),
+                    )
```
alanmcruickshank, line 1471: Same comment here as above - that we should allow either a numeric literal or an expression.

alanmcruickshank (changes requested):
Thanks @adityapat3l - I've made a suggestion which should also work, but is even more robust to the use of functions and more complex expressions.

Additionally - I think this might be more appropriate in the `ansi` dialect rather than just in `postgres`. For example, the same syntax is supported in MySQL too (https://dev.mysql.com/doc/refman/8.0/en/select.html).

Would you be able to shift your edits from here into `dialect_ansi.py` instead?

src/sqlfluff/dialects/dialect_postgres.py
adityapat3l: Thank you @alanmcruickshank! I'll get this sorted and come back to you

alanmcruickshank (approved):
Nice. LGTM 👍 . Thanks for putting this together

PR #5076, opened 2023-08-09
Parent references and more efficient path_to

alanmcruickshank:
@WittierDinosaur has been suggesting that segments should maintain a reference to their parent for a while. This PR implements that.

I found during some performance investigations on the most recent release that some rules take disproportionately longer than others. Especially RF01, and most of that time was spent in `path_to`. The lack of a reference to the parent means that the algorithm used for searching here was especially slow.

This introduces a `weakref` to the parent segment on `BaseSegment`. It's not populated on instantiation (because that proved difficult) and also includes a bit of validation on read because we definitely move segments around and there's a risk that the parent reference becomes out of date.

That parent reference can then be used to speed up the `path_to` method and in my testing on some large files, this roughly halves the runtime of RF01, AL04, AL05 & AM04 🚀 (which are also some of the currently slowest rules). This same functionality should enable some speedups in the reflow libraries too, but I'll handle that in a second PR (currently this doesn't show a significant gain there).

In the process, I also added a new pickle test for segments (similar to the ones done for errors), and that triggered me to break apart the segment tests into smaller file. The only change to the test suite is that new test - otherwise it's the same tests, just in smaller files (similar to #5078).


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -849,6 +880,42 @@ def _preface(self, ident, tabsize) -> str:
...
+        have an alternative way of assessing position, and ideally also
+        set the parent of any segments found without them.
+
+        NOTE: We only store a weak reference to the parent so it might
+        not be present. We also validate here that it's _still_ the parent
+        and potentially also return None if those checks fail.
+        """
+        if not self._parent:
+            return None
+        _parent = self._parent()
```
WittierDinosaur, line 914: maybe i'm missing something about weakref, but why is this a function call?

test/core/parser/segments/segments_base_test.py
```diff
@@ -228,35 +210,12 @@ def test__parser__base_segments_base_compare():
...
-                PathStep(test_seg, 0, 2, (0, 1)),
-                PathStep(test_seg.segments[0], 0, 1, (0,)),
-            ],
-        ),
-        (raw_seg_list[1], [PathStep(test_seg, 1, 2, (0, 1))]),
-    ]
+def test__parser__base_segments_pickle_safe(raw_seg_list):
+    """Test pickling and unpickling of BaseSegment."""
+    test_seg = BaseSegment([BaseSegment(raw_seg_list)])
+    test_seg.populate_parents()
```
WittierDinosaur, line 216: i think populate_parents is confusing as a method name. It is populating the children's parents, but running this as a function on test_seg implies you're setting test_seg's parent. Not the end of the world, but maybe there's a more clear name

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: That's just how `weakref` works 🤷: https://docs.python.org/3/library/weakref.html

test/core/parser/segments/segments_base_test.py
alanmcruickshank: Hmmm that is a good point 🤔 . I'm going to see if I can come up with a better name...

alanmcruickshank: How about `set_as_parent()`? That feels more descriptive.

WittierDinosaur: Yeah, I prefer that

WittierDinosaur (approved):
lgtm

PR #5079, opened 2023-08-09
New Rule: AL08 - column aliases must be unique

alanmcruickshank:
I've been meaning to add this for ages, strangely never been requested.

This is a new rule to catch duplicate columns. A sibling to https://docs.sqlfluff.com/en/stable/rules.html#rule-aliasing.unique.table

Optional case sensitivity which defaults to case insensitive, also quote stripping included.

### DIFF
--- a/src/sqlfluff/core/rules/config_info.py
+++ b/src/sqlfluff/core/rules/config_info.py
@@ -7,8 +7,9 @@ This mapping is used to validate rule config inputs, as well
 as document rule configuration.
 """
 
-from sqlfluff.core.plugin.host import get_plugin_manager
 from typing import Any, Dict
+
+from sqlfluff.core.plugin.host import get_plugin_manager
 
 STANDARD_CONFIG_INFO_DICT: Dict[str, Dict[str, Any]] = {
     "tab_space_size": {
@@ -45,10 +46,6 @@ STANDARD_CONFIG_INFO_DICT: Dict[str, Dict[str, Any]] = {
         "definition": (
             "Run this rule even for dialects where this rule is disabled by default."
         ),
-    },
-    "case_sensitive": {
-        "validation": [True, False],
-        "definition": ("Whether this rule is case sensitive or not."),
     },
     "unquoted_identifiers_policy": {
         "validation": ["all", "aliases", "column_aliases"],
--- a/src/sqlfluff/core/rules/config_info.py
+++ b/src/sqlfluff/core/rules/config_info.py
@@ -46,6 +46,10 @@ STANDARD_CONFIG_INFO_DICT: Dict[str, Dict[str, Any]] = {
             "Run this rule even for dialects where this rule is disabled by default."
         ),
     },
+    "case_sensitive": {
+        "validation": [True, False],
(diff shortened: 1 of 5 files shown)


noel:
Nice addition

src/sqlfluff/core/rules/config_info.py
```diff
@@ -46,6 +46,10 @@
             "Run this rule even for dialects where this rule is disabled by default."
         ),
     },
+    "case_sensitive": {
```
WittierDinosaur, line 49: are any dialects case insensitive? If not, i'm not sure we should offer this config. We don't consider case when we parse

test/fixtures/rules/std_rule_cases/AL08.yml
```diff
@@ -0,0 +1,70 @@
...
+test_pass_alias_case:
+  pass_str: |
+    select
+      foo,
+      b as FOO
+
+  configs:
+    rules:
+      aliasing.unique.column:
+        case_sensitive: true
```
WittierDinosaur, line 70: can you add a fail case like:
```
select
  a.foo
  , b as foo
  from a
```
assuming current logic causes this to fail

pwildenhain (commented):
I'm also really surprised this has t already been requested, makes a lot of sense. Great rule. After Danny's questions are addressed, I think this is good to go ☑️

src/sqlfluff/core/rules/config_info.py
alanmcruickshank: To my knowledge, most dialect aren't case sensitive _unless quoted_ and then they are. It's a good point though, that maybe I should just handle quoted and unquoted differently.

Quoted is case sensitive. Unquoted is not. MySQL and postgres fold to lowercase, snowflake folds to uppercase:
- https://dev.mysql.com/doc/refman/8.0/en/identifier-case-sensitivity.html
- https://www.postgresql.org/docs/current/sql-syntax-lexical.html
- https://docs.snowflake.com/en/sql-reference/identifiers-syntax

There's a good argument that while two quoted columns `"foo"` and `"FOO"` might be _valid_ and actually refer to different columns - it's probably still not _recommended_. I'll update the docstring and take away the config.

test/fixtures/rules/std_rule_cases/AL08.yml
alanmcruickshank: lol - it didn't, but it does now!

src/sqlfluff/core/rules/config_info.py
alanmcruickshank: docstring now says:

```python
class Rule_AL08(BaseRule):
    """Column aliases should be unique within each clause.

    Reusing column aliases is very likely a coding error. Note that while
    in many dialects, quoting an identifier makes it case-sensitive
    this rule always compares in a case-insensitive way. This is because
    columns with the same name, but different case, are still confusing
    and potentially ambiguous to other readers.

    In situations where it is *necessary* to have columns with the same
    name (whether they differ in case or not) we recommend disabling this
    rule for either just the line, or the whole file.
    """
```

WittierDinosaur (approved):
LGTM

PR #5081, opened 2023-08-10
Remove the internal cProfiler option

alanmcruickshank:
The internal `--profiler` option on `sqlfluff parse` is both too specific and also too blunt to be useful. I've been increasingly not using it, and instead reverting to running external profiling commands.

I think we might need something _instead_ of this, but it's probably not baked into the CLI, and instead is likely to be a more granular view focussed around specific steps in the process.

Until then - I think it's the right time to say bye bye to this CLI option as part of the next minor release.

### DIFF
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -13,10 +13,6 @@ import yaml
 
 import click
 
-# For the profiler
-import pstats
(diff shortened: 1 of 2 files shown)


WittierDinosaur (approved):
Didn't even know this existed 😂

PR #5082, opened 2023-08-10
Optimisation on select analysis

alanmcruickshank:
Debugging performance of AL05, I found this optimisation. It cuts the runtime of AL05 on my test file (a big mess query) by 10x!! 🎉 🚀 

In essence we were recursively exploring far too much of the parse tree and then going back to work out which elements to remove at the end of that recursion. The more efficient way to solve that is just to limit recursion in the first place. Double saving of 1) faster tree traversal and 2) not needing to prune them back off again later.

Across a blended project it has the following impacts on time spent:
- AL04 -75%
- AL05 -68%
- AM04 -60%
- RF01 -35%
- ST05 -47%

Especially exciting that these are some of the longest running rules. Overall reduction in linting time of ~30% (with the caveat that results may vary depending on project makeup and the rules enabled on the project).

### DIFF
--- a/src/sqlfluff/utils/analysis/select.py
+++ b/src/sqlfluff/utils/analysis/select.py
@@ -34,7 +34,11 @@ def get_select_statement_info(
     # Sometimes there is no select clause (e.g. "SELECT *" is a select_clause_element)
     if not sc:
         return None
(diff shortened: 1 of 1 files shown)


WittierDinosaur (approved):
Absolutely huge!

PR #5094, opened 2023-08-15
Strict mypy in parser.segments

alanmcruickshank:
This is another biggie.

At it's core, this makes `segments.base` compliant with strict mypy. In effect this now makes `parser.segments` fully strict mypy compliant 🚀 ✅ .

However in the process I've also:
- Extracted several things from `segments/base.py` (because it was already too big): `fix.py`, `file.py` and `bracketed.py` (those extractions are mostly unchanged)
- Also made `meta.py` fully mypy compliant.
- Simplified and added tests for the methods which convert segments to tuples and dicts: `to_tuple()` and `as_record()`. Both were in dire need of some attention. They also have proper typing in place now.
- Adds proper typing for `SegmentMetaclass` (which taught me a lot about typing 😬 )

The rest is mostly fallout from those changes and some tidyup. I don't believe that any of these changes should present anything noticeable to the end user - all under the hood.


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -743,34 +637,49 @@ def is_optional(cls) -> bool:
...
-                # This looks like a list of elements.
-                keys = [e[0] for e in elem]
-                # Any duplicate elements?
-                if len(set(keys)) == len(keys):
-                    # No, we can use a mapping tuple
-                    elem = {e[0]: cls.structural_simplify(e[1]) for e in elem}
-                else:
-                    # Yes, this has to be a list :(
-                    elem = [cls.structural_simplify(e) for e in elem]
-        return elem
```
alanmcruickshank, line 773: This old method was _a mess_. I've basically totally re-written it to make it more manageable (and testable).

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1545,32 +1465,35 @@ def compute_anchor_edit_info(
...
             # parse() on a "backup copy" of the segment.
-            r_copy = deepcopy(segment)
-            for seg in r_copy.segments:
-                seg.pos_marker = replace(
-                    seg.pos_marker,
-                    templated_file=self.pos_marker.templated_file,
-                )
-            r_copy.parse(ctx)
+            segment_copy = segment.copy()
+            segment_copy.parse(ctx)
```
alanmcruickshank, line 1483: The builtin `.copy()` method is now a lot more robust, so it should work fine in place of `deepcopy()`.

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -1512,17 +1512,27 @@ def get_eventual_alias(self) -> AliasInfo:
...
+                tbl_expression = _bracketed.get_child("table_expression")
         # For TSQL nested, bracketed tables get the first table as reference
         if tbl_expression and not tbl_expression.get_child("object_reference"):
-            if tbl_expression.get_child("bracketed"):
-                tbl_expression = tbl_expression.get_child("bracketed").get_child(
-                    "table_expression"
-                )
-        ref = tbl_expression.get_child("object_reference") if tbl_expression else None
+            _bracketed = tbl_expression.get_child("bracketed")
+            if _bracketed:
```
WittierDinosaur, line 1524: Based on the previous behaviour, i wonder if this should be an assert?

alanmcruickshank: Based on my reading of the old behaviour it only tries to `.get_child("table_expression")` if it's already confirmed that `tbl_expression.get_child("bracketed")` is non null (albeit in an inefficient way by calling `tbl_expression.get_child("bracketed")` twice).

That means it did previously allow `tbl_expression.get_child("bracketed")` to be `None`. Did you mean something else @WittierDinosaur ?

WittierDinosaur: Ah, I just completely read the old code wrong - you're right

WittierDinosaur (approved):
Great stuff

Issue #4139, opened 2022-12-06
Allow not specifying parameters names when using `placeholder` templater


FBorowiec:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

_First let me thank the creators and maintainers of `SQLFluff`!
I learned a lot about SQL thanks to you :)_

I am now using the following configuration inside a `pyproject.toml` (thanks a lot for supporting that format btw <3):

```toml
[tool.sqlfluff.core]
templater = "placeholder"
dialect = "postgres"

[tool.sqlfluff.templater.placeholder]
param_style = "pyformat"
```

But as far as I understood I need to specify all the parameters inside the `sql` queries I have in my project.
So for example the query:

```sql
INSERT INTO my_table (
    var1,
    var2)
VALUES (
    %(var1)s,
    %(var2)s);
```

would require me to update the `pyproject.toml` with the following lines:

```toml
[tool.sqlfluff.templater.placeholder]
param_style = "pyformat"
var1 = 1  # <-- Added a random value
var2 = 2  # <-- Added a random value
```

My use case though is that I load each `sql` query in `python` in the following manner:

```python
def execute_query(**kwargs) -> None:
  # kwargs could be = {"var1": 1, "var2": 2}

  # Connecting to the database with psycopg2
  db_connection = psycopg2.connect(
      host="hostname",
      dbname="database",
      port="5432",
      user="postgres",
      password="",
  )
  cursor = db_connection.cursor()

  with open("path/to/query.sql", "r", encoding="utf-8") as query_file:
      query: str = query_file.read()
      with db_connection:
          cursor.execute(query, kwargs)
          db_connection.commit()
```

And I don't know the values of each parameter beforehand.
If I don't specify them `SQLFluff` will crash with the following error:

```bash
"sqlfluff.core.errors.SQLTemplaterError: Failure in placeholder templating: 'var1'.
Have you configured your variables?"
```

Is there a way not to have to specify each of the parameters inside the
`pyproject.toml` file?

Thanks a lot!

### Use case

I would like to not have to specify each of the parameters inside the `pyproject.toml` file.

### Dialect

`postgres`

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

[...]

Issue #2108, opened 2021-12-13
Placeholder templater should optionally supply default values for missing parameters


barrywhart:
Capturing some thoughts from a PR (#2107) that we decided was too specific an implementation, but could potentially be re-implemented in a more generic way.

I've fixed some issues as pointed out by @jpy-git, and added some basic support for Postgres data types.

The author of the PR, @danicc097, described his use case:

>I'm using [sqlc](https://github.com/kyleconroy/sqlc) with Postgres. Variables use `@` prefix, so I'd have queries like:

```sql
-- name: UpdateUserById :one
UPDATE
  "users"
SET
  "password" = CASE WHEN @password_do_update::boolean THEN
    @password
  ELSE
    "password"
  END,
  "salt" = CASE WHEN @salt_do_update::boolean THEN
    @salt
  ELSE
    "salt"
  END,
  "username" = CASE WHEN @username_do_update::boolean THEN
    @username
  ELSE
    "username"
  END,
  "email" = CASE WHEN @email_do_update::boolean THEN
    @email
  ELSE
    "email"
  END
WHERE
  "id" = @id
RETURNING
  *;
```

>Creating dummy values in the config seems unneccesary in these cases, except common things like "id", etc.  Regarding usefulness, if a user opts in to autofill parameters it will look for the Postgres "::" syntax (else notify it will be replaced with text) and parse simple builtin types and arrays. Its just a matter of extending the recognized types with dummy values in the templater. Of course, as it is now, the implementation is limited to Postgres (else it will always autofill a string); I don't know if for other databases it's worth both the verbosity of typing and effort to make its type parser in the templater, maybe someone has more experience in other dialects.

==

Here's one high-level thought on how we might design this generically:
Let the user define (in `.sqlfluff`) a set of prefixes or suffixes that implicitly specify the type and default value of the variable, e.g. (this is adapted from the original PR):
```
[sqlfluff:templater:placeholder:context:autofill:prefixes]
str="'string'"
date=CURRENT_DATE
tbl=table_$1
col=column_$2
```

In other words, a missing variable with a "str" prefix gets a quoted string as a default value,  "date" prefix gets a date literal, and "tbl" and "col" prefixes get a not-quoted name followed by the variable name (to avoid having every table or column name be the same, which could confuse the linting rules.)

==

Another idea (more complex):
[...]

PR #5101, opened 2023-08-17
Allow not specifying parameters names when using placeholder templater

shyaginuma:

fixes https://github.com/sqlfluff/sqlfluff/issues/4139, fixes https://github.com/sqlfluff/sqlfluff/issues/2108 (actually enhancement)

From my perspective, nothing. But this change was different from the intent of `placeholder` templater, I'd like to discuss.

### Pull Request checklist

  - Other.


src/sqlfluff/core/templaters/placeholder.py
```diff
@@ -153,14 +153,7 @@ def process(
...
             last_literal_length = span[0] - last_pos_raw
-            try:
-                replacement = str(context[param_name])
-            except KeyError as err:
-                # TODO: Add a url here so people can get more help.
-                raise SQLTemplaterError(
-                    "Failure in placeholder templating: {}. Have you configured your "
-                    "variables?".format(err)
-                )
+            replacement = str(context.get(param_name, ""))
```
shyaginuma, line 156: It might be better to set custom default value instead of blank string, but I'd like to hear your opinion.

greg-finley:
n.b., to close two issues, next time put "fixes #4139, fixes #2108" ("fixes" twice) in the PR description https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

shyaginuma:
now, I think it's ready for review. Could someone take a look?

src/sqlfluff/core/templaters/placeholder.py
alanmcruickshank: @shyaginuma - I think it would be safer to replace with the name of the variable as a string. That way it's transparent and most likely to still parse. This is similar to what we've done with the dbt `ref()` macro.

i.e. `SELECT :foo, 1` would get parsed as `SELECT "foo", 1` rather than `SELECT , 1` (the latter would fail to parse, but the earlier one would pass).

How does that sound to you? I think most SQL dialects allow quoted literals in most cases that would accept strings, dates or other datatypes.

@barrywhart - do you have any views as you put up one of the original issues?

barrywhart: IIRC, when I implemented the similar behavior for the Jinja and dbt templaters, I had it simply return the unquoted string. Either behavior may be reasonable, but I think we should be consistent. Can someone confirm the dbt templater's behavior?

alanmcruickshank (changes requested):
Hi @shyaginuma - could you resolve the thread above about how to resolve missing variables? To be consistent with other places where we do something similar, I think we should resolve any missing variables as the name of the variable rather than an empty string, I think that gives us the best chance of the SQL still parsing afterward.

Otherwise I think this is good to go. Thanks for putting this together! 🏆

src/sqlfluff/core/templaters/placeholder.py
shyaginuma: @alanmcruickshank Thanks for your suggestion! Fixed in [`31bd5f1`](https://github.com/sqlfluff/sqlfluff/pull/5101/commits/31bd5f12b063a999c0eba0ecbdd296a109fc7ab5), could you take a look again?

alanmcruickshank (approved):
LGTM! 🚀 Thanks for the somewhat longer review process on this one.

shyaginuma:
Thanks for the review!! 😄

PR #5102, opened 2023-08-17
RF01 & recursive_crawl improvements

alanmcruickshank:
This is a fairly small PR, aimed at improving performance of RF01.

Includes one syntax simplification in RF01 (which is trivial), but more importantly updating the `.recursive_crawl()` method to filter based on the same type sets we now use in rule crawling.

Early testing suggests this more than halves the time spent in RF01 in long queries. Across a blended project the impact is more like 10%. Similar gains on ST05.

### DIFF
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -1145,17 +1145,29 @@ class BaseSegment(metaclass=SegmentMetaclass):
             recurse_into: :obj:`bool`: When an element of type "seg_type" is
                 found, whether to recurse into it.
             no_recursive_seg_type: obj: `str`: a type of segment
-                not to recurse further into.
+                not to recurse further into. It is highly recommended
+                to set this argument where possible, as it can significantly
+                narrow the search pattern.
         """
-        # Check this segment
+        # Check whether the types we're looking for are in this segment
+        # at all. If not, exit early.
+        if not self.descendant_type_set.intersection(*seg_type):
+            return
+
+        # Assuming there is a segment to be found, first check self.
         if self.is_type(*seg_type):
             match = True
             yield self
         else:
             match = False
+
+        # Then handle any recursion.
         if recurse_into or not match:
-            # Recurse
             for seg in self.segments:
+                # Don't recurse if the segment is of a type we shouldn't
+                # recurse into.
+                # NOTE: Setting no_recursive_seg_type can significantly
+                # improve performance in many cases.
                 if not seg.is_type(no_recursive_seg_type):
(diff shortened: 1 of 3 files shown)


src/sqlfluff/rules/references/RF01.py
```diff
@@ -91,30 +91,30 @@ def _eval(self, context: RuleContext) -> EvalResultType:
         ):
             return LintResult()
 
+        if FunctionalContext(context).parent_stack.any(sp.is_type(*_START_TYPES)):
```
pwildenhain, line 94: Good refactor here

src/sqlfluff/rules/structure/ST05.py
```diff
@@ -531,13 +531,21 @@ def _create_table_ref(table_name: str, dialect: Dialect) -> TableExpressionSegme
 
 
 def _get_case_preference(root_select: Segments):
-    first_keyword = root_select.recursive_crawl(
-        "keyword",
-        recurse_into=False,
-    ).first()[0]
-    if first_keyword.raw[0].islower():
```
pwildenhain, line 538: I'm sure mypy also appreciates this refactor too

pwildenhain (approved):
🚀

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1145,17 +1145,29 @@ def recursive_crawl(
...
         if self.is_type(*seg_type):
             match = True
             yield self
         else:
             match = False
+
+        # Check whether the types we're looking for are in this segment
+        # at all. If not, exit early.
+        if not self.descendant_type_set.intersection(seg_type):
+            return
```
WittierDinosaur, line 1162:
```suggestion
            return None
```

src/sqlfluff/rules/references/RF01.py
```diff
@@ -91,30 +91,30 @@ def _eval(self, context: RuleContext) -> EvalResultType:
...
+            return LintResult()
```
WittierDinosaur, line 95: My only issue with this is the mypy strict stuff. Between this and the below, we have 3 types of return, List[LintResult] (violations), LintResult, or None - could do with a bit of cleaning. Really we should only have one Falsy return type

WittierDinosaur (changes requested):
Two small things to keep us honest with mypy

src/sqlfluff/rules/references/RF01.py
alanmcruickshank: That's a good point - I've made them all lists, which I think is best in this case

WittierDinosaur (approved):
LGTM

PR #5104, opened 2023-08-17
Select Crawler Refactor: Part 1

alanmcruickshank:
This is a draft for now while I see how the test suite reacts.

The next big frontier for performance in linting is the `SelectCrawler`. It's very wasteful in terms of how much work it does. I'm slowly working through and pulling bits out of it, but it's going to take a few passes.

This PR:
- Modularises the test suite to make it a bit more DRY.
- Removes the `parent_stack` option to `SelectCrawler`. It was a hack and can be much better solved now that we've got fairly reliable access to the parent segment anyway.
- Removes the use of `parent_query` in AM07 but just detecting a `WITH` statement early and using that instead where we can.
- Simplifies `ST03` to use a much more minimal set of methods and avoid too much jumping around (it also uses fewer lines of code).

### DIFF
--- a/src/sqlfluff/rules/structure/ST03.py
+++ b/src/sqlfluff/rules/structure/ST03.py
@@ -46,7 +46,7 @@ class Rule_ST03(BaseRule):
     name = "structure.unused_cte"
     aliases = ("L045",)
     groups = ("all", "core", "structure")
-    crawl_behaviour = SegmentSeekerCrawler({"statement"})
+    crawl_behaviour = SegmentSeekerCrawler({"with_compound_statement"})
 
     @classmethod
     def _find_all_ctes(cls, query: Query) -> Iterator[Query]:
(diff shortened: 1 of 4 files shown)


src/sqlfluff/rules/structure/ST03.py
```diff
@@ -46,41 +44,32 @@ class Rule_ST03(BaseRule):
...
-    def _find_all_ctes(cls, query: Query) -> Iterator[Query]:
-        if query.ctes:
-            yield query
-        for query in query.ctes.values():
-            yield from cls._find_all_ctes(query)
-
-    @classmethod
-    def _visit_sources(cls, query: Query) -> None:
-        for selectable in query.selectables:
-            for source in query.crawl_sources(selectable.selectable, pop=True):
```
alanmcruickshank, line 61: Removing the reference to `.crawl_sources()` should have a performance boost. I can't remove the `crawl_sources()` method entirely yet though because it's still used in a few places.

WittierDinosaur (approved):
LGTM - changes all seem sensible

PR #5107, opened 2023-08-18
✅ Strict MyPy for sqlfluff.core.parser

alanmcruickshank:
This finishes off the effort to get the parser full `mypy --strict` compliant.

Notable things I've done along the way:
- `ExpandedDialectElementType` and `MatchableType` were defined to be the same thing. We now just have the latter (defined centrally in `sqlfluff.core.parser.types`).
- The bracket pairs set now gets it's own custom type: `BracketPairTuple` (defined centrally in `sqlfluff.core.parser.types`).
- Much more explicit `__init__()` methods across the whole module.
- Simplification of the `Ref` grammar, in particular to do fewer checks at runtime, and instead do them on instantiation.
- I've updated the coverage rules to handle `@overload` better.
- Updates the `mypy` tox step to do strict checks on the parser. NOTE: This involved changing how `sqlfluff` is imported for that job, but hopefully not in a controversial way. I was otherwise getting strange errors.

The rest is fairly self explanatory. A few other small methods got rewrites, but nothing earth shattering.

Annoyingly, this doesn't mean that I can compile with `mypyc` yet, but I still think this was a worthwhile effort. 🚀


src/sqlfluff/core/parser/grammar/anyof.py
```diff
@@ -1,51 +1,71 @@
 """AnyNumberOf, OneOf, OptionallyBracketed & AnySetOf."""
 
-from typing import List, Optional, Tuple
+from typing import FrozenSet, List, Optional
+from typing import Sequence as SequenceType
+from typing import Tuple, Union, cast
```
WittierDinosaur, line 5: Any reason for multiple calls to the same package?

src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -121,30 +134,42 @@ class BaseGrammar(Matchable):
...
-    def _resolve_ref(elem):
+    def _resolve_ref(elem: str) -> "BaseGrammar":
+        ...
+
+    @overload
+    @staticmethod
+    def _resolve_ref(elem: Matchable) -> Matchable:
+        ...
+
+    @overload
```
WittierDinosaur, line 154: Not sure I understand these overloads? Can you please explain?

src/sqlfluff/core/parser/grammar/conditional.py
```diff
@@ -41,15 +41,16 @@ class Conditional(BaseGrammar):
...
+    def __init__(
+        self,
+        meta: Type[Indent],
+        config_type: str = "indentation",
+        **rules: Union[str, bool]
+    ):
+        assert issubclass(
+            meta, Indent
+        ), "Conditional is only designed to work with Indent/Dedent segments."
+        self._Meta = meta
```
WittierDinosaur, line 53: accidental capitalisation?

src/sqlfluff/core/parser/grammar/anyof.py
alanmcruickshank: The `isort` library did this automatically 🤔 . It's set up using the `--profile black`, which is described here (https://pycqa.github.io/isort/docs/configuration/profiles.html). I don't hate this layout, but I think it's driven by the `Sequence as SequenceType` line in there. Without that it reverts to the normal format.

This feels like it might be a but in `isort` but given it's being automatically applied. I'm not against it. Views?

src/sqlfluff/core/parser/grammar/conditional.py
alanmcruickshank: I think I was trying to be clever to indicate that it's a class rather than an instance - but I think it might be more unhelpful. I'll revert to lowercase.

src/sqlfluff/core/parser/grammar/base.py
alanmcruickshank: Yeah maybe they're a bit overkill. I'll remove them. I don't think they're _necessary_ per se.

alanmcruickshank: Actually - I tried, they are necessary - but I have managed to reduce them a bit. In essence it's a hint to mypy to say, when you provide `None` you get `None`, otherwise you will always get at least _something_. If I remove them all I get awkward errors.

WittierDinosaur (approved):
LGTM

PR #5110, opened 2023-08-20
Select Crawler Refactor: Part 2

alanmcruickshank:
This PR follows on from #5104 .

As a next step this removes the use of `visit_segments` (which doesn't filter) and replaces it with `recursive_crawl` with much more specific filtering. That triggered some follow on issues with `RF03` which I've also addressed and tried to do that in the simplest way.

Draft for now, I might go a little further before putting it up for review...

### DIFF
--- a/src/sqlfluff/rules/references/RF03.py
+++ b/src/sqlfluff/rules/references/RF03.py
@@ -15,7 +15,6 @@ from sqlfluff.core.rules import (
     RuleContext,
 )
 from sqlfluff.core.rules.crawlers import SegmentSeekerCrawler
-from sqlfluff.utils.functional import sp, FunctionalContext
 from sqlfluff.dialects.dialect_ansi import IdentifierSegment, ObjectReferenceSegment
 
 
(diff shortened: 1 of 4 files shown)


src/sqlfluff/rules/references/RF03.py
```diff
@@ -254,40 +261,43 @@ def _validate_one_reference(
...
+                    SymbolSegment(raw=".", type="symbol"),
+                ],
+            )
+        ]
+    return LintResult(
+        anchor=ref,
+        fixes=fixes,
+        description="{} reference {!r} found in single table "
+        "select.".format(this_ref_type.capitalize(), ref.raw),
+    )
```
alanmcruickshank, line 303: Note for reviewers: This whole section is exactly the same in function as before. All I've done is swapped the condition of the first `if` statement from `!=` to `==`, and by doing that have a first (shorter) `return None` section, and then everything else can be un-nested by one step.

src/sqlfluff/rules/references/RF03.py
```diff
@@ -112,20 +121,17 @@ def _visit_queries(self, query: Query, visited: set) -> Iterator[LintResult]:
...
+                    possible_ref_tables += list(
+                        self._iter_available_targets(query.parent)
                     )
-                    if table_search_root.selectables
-                    else []
-                )
-                filtered_query_list = [q for q in query_list if isinstance(q, str)]
-                if len(filtered_query_list) != 1:
+                if len(possible_ref_tables) > 1:
                     # If more than one table name is visible, check for and report
```
alanmcruickshank, line 130: There was no need to run `SelectCrawler.get()` here (which incurs a significant penalty), we can instead just get all the information we need from the `Query` object which has already been parsed.

src/sqlfluff/rules/references/RF03.py
```diff
@@ -93,14 +92,24 @@ def _eval(self, context: RuleContext) -> EvalResultType:
         if context.dialect.name in self._dialects_with_structs:
             self._is_struct_dialect = True
 
-        if not FunctionalContext(context).parent_stack.any(sp.is_type(*_START_TYPES)):
+        # If any of the parents would have also triggered the rule, don't fire
+        # because they will more accurately process any internal references.
+        if not any(seg.is_type(*_START_TYPES) for seg in context.parent_stack):
```
alanmcruickshank, line 97: No need to convert to the functional context here, doing it with builtins is more readable and simpler.

src/sqlfluff/utils/analysis/select_crawler.py
```diff
@@ -177,40 +178,32 @@ def crawl_sources(
...
-                continue
-
+            # Crawl efficiently, don't recurse here. We do that later.
+            # What do we have?
+            # 1. If it's a table reference, work out whether it's to a CTE
+            #    or to an external table.
             if seg.is_type("table_reference"):
-                if not seg.is_qualified() and lookup_cte:
-                    cte = self.lookup_cte(seg.raw, pop=pop)
+                _seg = cast(ObjectReferenceSegment, seg)
```
WittierDinosaur, line 189: why is this cast necessary?

alanmcruickshank: It's so that mypy doesn't complain about the `.is_qualified()` call, which is only available on an object reference.

WittierDinosaur (approved):
LGTM

PR #5112, opened 2023-08-21
Refactor AM07

alanmcruickshank:
This is part of cleaning up the select crawlers.

As part of that I've had a bunch of issues with AM07. The original code was quite convoluted and hard to follow, in particular there was a `list` being passed around and mutated in several methods.

This clears up that rule to make it easier to work with in future. I haven't changed any of the tests and you'll note that this PR only touches one file. Specifically:

- I've removed all mutation within methods. Everything is now handled with return values.
- All returns are now fully typed, and rather than returning a `list` they return just an `int` and `bool`.
- I've removed the need to pass around as many values. In particular we were passing around a `context` and a `crawler` which weren't' necessary.
- There's more logging, and better comments.

This *looks* like a total re-write, and it basically is - but the logic flow is the same, just in a very different structure.

### DIFF
--- a/src/sqlfluff/rules/ambiguous/AM07.py
+++ b/src/sqlfluff/rules/ambiguous/AM07.py
@@ -1,5 +1,5 @@
 """Implementation of Rule AM07."""
-from typing import Any, List, Optional, Set, Tuple
+from typing import Optional, Set, Tuple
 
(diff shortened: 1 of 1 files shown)


greg-finley (approved):
Love it! Thanks for the detailed PR description. 👍

PR #5113, opened 2023-08-21
Simplify rules with improvement to SegmentSeeker

alanmcruickshank:
Noticed this while working on some other rules. A couple of rules could seek more effectively rather than just rejecting segments based on their parents. This adds that simple functionality, which should reduce overheads.

### DIFF
--- a/src/sqlfluff/core/rules/crawlers.py
+++ b/src/sqlfluff/core/rules/crawlers.py
@@ -51,13 +51,22 @@ class SegmentSeekerCrawler(BaseCrawler):
     """
 
     def __init__(
-        self, types: Set[str], provide_raw_stack: bool = False, **kwargs: Any
+        self,
+        types: Set[str],
+        provide_raw_stack: bool = False,
+        allow_recurse: bool = True,
+        **kwargs: Any
     ) -> None:
         self.types = types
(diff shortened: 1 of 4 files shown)


src/sqlfluff/core/rules/crawlers.py
```diff
@@ -51,13 +51,22 @@ class SegmentSeekerCrawler(BaseCrawler):
...
+        provide_raw_stack: bool = False,
+        allow_recurse: bool = True,
+        **kwargs: Any
     ) -> None:
         self.types = types
         # Tracking a raw stack involves a lot of tuple manipulation, so we
         # only do it when required - otherwise we skip it. Rules can explicitly
         # request it when defining their crawler.
         self.provide_raw_stack = provide_raw_stack
+        # If allow_recurse is false, then one a segment matches, none of it's
```
pwildenhain, line 65: *once a segment

pwildenhain (approved):
Love the simplification! -- question: when do we still need `FunctionalContext`?

alanmcruickshank:
> Love the simplification! -- question: when do we still need `FunctionalContext`?

I think there are some places where it totally makes sense. It probably needs a bit of work for it to be really useful, but it's most useful in iterating through sequences looking for particular types.

PR #5115, opened 2023-08-22
Select Crawler Refactor: Part 3

alanmcruickshank:
This follows on from #5104 & #5110. I think this is also my last biggie on this patch of the codebase (probably). See below for performance numbers.

This PR:
- Fully removes `visit_segments` from the `SelectCrawler`, which was the main performance concern.
- Extends the test suite to better cover `Query.children`. Covering it directly was very verbose, so I've introduced a separate `subqueries` property, which is the combination of `ctes` and `subqueries`. In doing that I've also added several more test cases to make sure it's properly covered.
- Renames `as_json` to `as_dict`, because it returns a ... `dict` and not json.
- Totally rewrites the code that was in the `__init__` method of the `SelectCrawler` and moves that into a constructor of the `Query` object. The old method was very convoluted and based on `visit_segments`. It also used several internal book-keeping functions which were harder to reason about. Where possible, I've consolidated clauses together to remove duplication.
- Totally removes the remaining shell which was `SelectCrawler`, and remove all the references to it. Replace them with constructors on `Query`. As a follow-on, I've renamed the source files for `select_crawler.py` to `query.py`.
- Removes the `parent` attribute on `Selectable`. It was only used in one place, and that was easily solvable without it. It was being set fairly inconsistently in the first place, so I suspect it added more confusion that benefit. There's still a `parent` attribute on `Query` which can be used instead in most cases.

I've run through the performance impacts of these changes and some are pretty big 🚀:
- `AL05`: -50%
- `AM04`: -25% (which brings it down low enough that it's not a problem)
- `AM07`: -60% (which brings this down to low enough that it's negligible)
- `RF01`: -40%
- `RF03`: -60%
- `ST03`: -90% !? (which brings this down to low enough that it's negligible)
- `ST05`: -40%

Based on the performance numbers the next most sensible place to focus would be on the reflow rules (notably `LT01`), but potentially also `JJ01`, `RF01`, `ST05` or `ST09` (two of which were still part of this PR, but I think further work would focus directly on those rules rather than on the analysis docs).


greg-finley (approved):

barrywhart:
I don't have time to review, but thanks for taking on this cleanup work! 🎉 My creation of SelectCrawler was more a case of need and opportunity than of a clear vision for how it should work. It's always felt like something of a messy but useful "evil".

PR #5116, opened 2023-08-22
Optimise crawl behaviour of JJ01

alanmcruickshank:
As noted in #5115, `JJ01` is also a fairly slow rule at the moment. On inspection this is because there's a lot of duplication going on as it crawls through the file.

This PR switches from crawling every `raw` segment in the file to only triggering _once_ on the root segment of the file, and then iterating through the templated slices from there. There might be a case in the future for a more generic mechanism, but for now this model is much more efficient.

The only complexity in here is that source fixes _must_ be attached to a `raw` segment rather than a `BaseSegment` (for reasons which I think are fairly sensible), but that means when we identify a source fix in this rule we _then_ try and identify an appropriate "host" segment for it. Feels potentially a little backward, but remarkably efficient in testing.

Reduces time spent in JJ01 to negligible, even across the whole project. 🚀

### DIFF
--- a/src/sqlfluff/rules/jinja/JJ01.py
+++ b/src/sqlfluff/rules/jinja/JJ01.py
@@ -1,16 +1,9 @@
 """Implementation of Rule JJ01."""
-from typing import Tuple
-from sqlfluff.core.parser.segments import SourceFix
+from typing import List, Tuple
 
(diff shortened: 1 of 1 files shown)


greg-finley (approved):
🚢

PR #5117, opened 2023-08-22
Ensure Unparsable can be given position.

alanmcruickshank:
This fixes an issue which was an oversight from some of the previous `mypy` work. If we try to apply fixes to an unparsable segment, then this fails (because we try to set position). No coverage because it's a very niche area.


pwildenhain (approved):
🐛

PR #5118, opened 2023-08-22
Reduce copying in LintFix instantiation

alanmcruickshank:
After some very high level performance investigations into `LT02`, I found that almost 20% of the time was being spent in copying stuff. The use of `deepcopy` was the culprit. Since this original code was written, I think the `BaseSegment.copy()` method has been cleaned up enough that it's safe to use instead.

No additional tests because the code is already well covered.


greg-finley (approved):
😍

PR #5119, opened 2023-08-22
Reduce copying on  _position_segments (improves `fix`)

alanmcruickshank:
Another performance one here. Triggered by the existing bigquery test `autofix/bigquery/004_templating`.

That test takes about 8s on my machine, of which most is copying. This significantly reduces the time spent applying fixes and the associated copying. Fix time is now ~3s so more than 50% improvement on linting. Specifically we now only copy the child segments once (we were copying them twice (at least?), first on the copy of the parent, and then again when we reposition the children).

This change will have the most effect when running `fix` on large files with lots of problems.


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -564,39 +566,45 @@ def _position_segments(
...
-                    line_no,
-                    line_pos,
-                )
-            )
-            line_no, line_pos = repositioned_seg.pos_marker.infer_next_position(
-                repositioned_seg.raw, line_no, line_pos
+            assert new_position
+
+            # Otherwise update the working position.
+            new_position = new_position.with_working_position(line_no, line_pos)
```
greg-finley, line 586: is this always going to run, since not under an `else` (matching `if` on line 543?) Was that the intention?

alanmcruickshank: Hmmm, good point. I don't think it needs to. I'll change that!

alanmcruickshank: I checked 😢 . Unfortunately this is still needed to be here. I tried moving it up and the tests fail. I'll leave it where it is for now.

greg-finley (approved):

PR #5135, opened 2023-08-27
Import rationalisation

alanmcruickshank:
I've been playing with some import tools and there are two that I'd love to add, both to apply them to the project but also adding them to CI:
1. `isort`, which sorts imports. This PR adds `isort` to CI and autofixes the current `sqlfluff.core` module. Eventually we can apply it to the rest of the project, but it's already a fairly large PR.
2. `importlinter`, which check which modules depend on eachother. The SQLFluff code base is already a little too convoluted in it's imports for my liking. This introduces the tool and two initial contracts (`sqlfluff.core` must not depend on parts of `sqlfluff` outside `core`, and `sqlfluff.api` must not depend on `sqlfluff.cli`). Neither of these rules are very controversial (although the first one wasn't entirely true before this PR), but this paves the way for stricter rules in the future.


alanmcruickshank:
This relies on #5134 before checks will pass.

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -747,6 +747,14 @@ class FileSegment(BaseFileSegment):
...
     )
 
+    def get_table_references(self) -> Set[str]:
+        """Use parsed tree to extract table references."""
+        references = set()
+        for stmt in self.get_children("statement"):
+            stmt = cast(StatementSegment, stmt)
+            references |= stmt.get_table_references()
+        return references
+
```
pwildenhain, line 757: Is there a reason this is in this PR? Did I miss something?

pwildenhain (approved):
Love `isort`, I personally like having a standard for import order, I think it really helps with code readability. Looks like there are still some `isort` errors, but overall it looks good.

src/sqlfluff/dialects/dialect_ansi.py
alanmcruickshank: Ah - it's just a _move_. It was on `BaseFileSegment` and now it's on `FileSegment` in the ansi dialect, so that we the definition file for `BaseFileSegment` doesn't need to import `ansi_dialect`.

PR #5148, opened 2023-09-02
Grammar .copy() assert no unexpected kwargs

alanmcruickshank:
While we can do need an arbitrary `**kwargs` argument to the grammar `.copy()` method, we never validate that for any given grammar, that we only pass appropriate values. This method is only called on import of grammars and so I think some validation is necessary.

In particular, I found an example where in the `hive` dialect we were passing an unused kwarg which was being silently ignored.

This adds an `assert` into the base `.copy()` method, and then fixes the one place where that assert was being violated.

NOTE: Most of the lines of code changed here are actually just `isort` running on the hive dialect. The rest of the changes are pretty minimal.

### DIFF
--- a/src/sqlfluff/core/parser/grammar/base.py
+++ b/src/sqlfluff/core/parser/grammar/base.py
@@ -913,6 +913,7 @@ class BaseGrammar(Matchable):
                 Elements are searched for individually.
 
         """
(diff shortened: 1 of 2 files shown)


greg-finley (approved):
Nice, like the safety of this!

Issue #5124, opened 2023-08-24
Pathway to "parse-less" parsing.


alanmcruickshank:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

This is an idea that has been kicking around for a while, but might now be possible to implement. In essence we current have a `match` phase and a `parse` phase to parsing. We flipflop between each for each segment as we recursively work through a given file.

However, that means we "touch" each potential segment many times and we do a lot of copying in the parsing phase which is quite inefficient.

I think there are a few steps to move us forward on this, roughly in this order:
1. There needs to be a rethink of how we handle and signal "unparsable" sections. This is currently tied to the `parse` phase of parsing, so we'd need to decouple this out. In particular, working through the question of where to mark the "unparsable" section. This should probably move into one of the grammars - or maybe within the `MatchResult` class. Perhaps returning the "best successful" variant even if it's eventually unsuccessful.
   - In essence, where segments have _two_ grammars currently (the `match_grammar` and `parse_grammar`), there is enough information in there to suggest how to appropriately create _"unparsable"_ sections. The same logic should apply, but without the two step process.
   - Where there are two grammars, the `match_grammar` is usually some kind of greedy match up to some terminator. I think we should be able to leverage the existing terminator infrastructure to intelligently first match the main body, _and then look ahead_ to any future terminators and create unparsable sections when they're in the way. We make a lot of progress on this with #5167.
3. I think the matching phase should do less copying, perhaps operating on slices of the original sequence of segments and then only crystalising into creating segments later in the process. Maybe we should do this before the part about unparsables above.
   - This would mean having a single `segments` object which everything is referenced to, but that the return object from `.match()` wouldn't yet have done any segment mutation, that would be saved for a second step in the process. i.e. `.match()` matches _in theory_, but doesn't create new segments _yet_. We only do that once when we're comfortable with the correct path.
4. Once we've got comfortable with both of the above, we can slowly start removing `parse_grammars` and only have `match_grammars`. In particular when doing this we need to check on performance and on behaviour around unparsables. To do this we're going to need a better test suite around unparsable code.

#### Notes
- Pathway to no `parse_grammar`:
  - #5186 (`StatementSegment`)
  - #5187 (`AnyNumberOf`)
  - will need a final PR to handle `FileSegment`, but that requires the other two. This will also get rid of `.expand()` and `.parse()` (potentially renaming `.match` to `.parse`).
[...]

PR #5149, opened 2023-09-02
Tests for unparsable sections

alanmcruickshank:
This is part of #5124 . Before I start changing the unparsable logic - I thought I should put some tests around the existing logic.

Not loads of tests, but this does give us enough to act as a comparison point. Ideally I'd like it to be a little more specific in future, but at the very least I don't want it to get more generic (i.e. just getting a totally unparsable statement).

### DIFF
--- a/test/dialects/unparsable_test.py
+++ b/test/dialects/unparsable_test.py
@@ -1,11 +1,10 @@
 """Test the behaviour of the unparsable routines."""
 
-import logging
 from typing import Any, Optional
 
 import pytest
 
-from sqlfluff.core import FluffConfig, Linter
+from sqlfluff.core import FluffConfig
 from sqlfluff.core.parser import BaseSegment, Lexer, RawSegment
 from sqlfluff.core.parser.context import ParseContext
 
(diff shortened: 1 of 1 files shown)


test/dialects/unparsable_test.py
```diff
@@ -0,0 +1,165 @@
...
+        Seg = config.get("dialect_obj").get_root_segment()
+    # We only allow BaseSegments as matchables in this test.
+    assert issubclass(Seg, BaseSegment)
+    assert not issubclass(Seg, RawSegment)
+
+    # Lex the raw string.
+    lex = Lexer(config=config)
+    seg_list, vs = lex.lex(raw)
+    assert not vs
+    print(seg_list)
```
greg-finley, line 152:
```suggestion
```

test/dialects/unparsable_test.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # Construct an unparsed segment
+    seg = Seg(seg_list, pos_marker=seg_list[0].pos_marker)
+    # Perform the match (THIS IS THE MEAT OF THE TEST)
+    ctx = ParseContext.from_config(config)
+    result = seg.parse(parse_context=ctx)
+
+    print(result)
```
greg-finley, line 160:
```suggestion
```

greg-finley (approved):

PR #5150, opened 2023-09-03
`terminator` > `terminators` on Delimited grammar

alanmcruickshank:
When we introduced deeper support for terminators in the context, it highlighted a few of the older grammars which already had _some_ support for terminators but with a slightly different syntax. This updates `Delimited` to use `terminators` and not `terminator` (note the plural - it's also a list rather than a single grammar). No functionality change for now, this almost entirely just syntax. It does use the parent `terminators` attribute from `AnyOf` in the background though.

The other grammar which needs this treatment is `StartsWith`, but I'll handle that as a separate PR (because it's a more tricky beast).

### DIFF
--- a/src/sqlfluff/core/parser/grammar/delimited.py
+++ b/src/sqlfluff/core/parser/grammar/delimited.py
@@ -1,6 +1,6 @@
 """Definitions for Grammar."""
 
-from typing import Optional, Tuple, Union, Sequence
+from typing import Optional, Sequence, Tuple, Union
 
 from tqdm import tqdm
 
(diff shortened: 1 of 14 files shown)


src/sqlfluff/core/parser/grammar/delimited.py
```diff
@@ -38,9 +38,8 @@ def __init__(
         *args: Union[MatchableType, str],
         delimiter: Union[MatchableType, str] = Ref("CommaSegment"),
         allow_trailing: bool = False,
-        # NOTE: Other grammars support terminators (plural)
-        # TODO: Align these to be the same eventually.
-        terminator: Optional[Union[MatchableType, str]] = None,
+        terminators: Optional[Sequence[Union[MatchableType, str]]] = None,
```
greg-finley, line 41: Love paying down tech debt!

greg-finley (approved):

PR #5152, opened 2023-09-03
`terminator` > `terminators` on StartsWith

alanmcruickshank:
This is a sister PR to #5150 . Context is the same:

> When we introduced deeper support for terminators in the context, it highlighted a few of the older grammars which already had some support for terminators but with a slightly different syntax.

This updates StartsWith to use `terminators` and not `terminator` (note the plural - it's also a list rather than a single grammar). Interestingly, this raised a bug in the function of `_look_ahead_match()` in the case that we used grammars which would match on `simple` but not in full (e.g. sequences), but where a later terminator might still match. This PR also resolves that bug.

`StartsWith` was also used a lot in template inheritance, and the method for setting the terminators was a little strange (and upredictable). I've extended the `.copy()` method for `StartsWith` to accept two additional kwargs `terminators` or `add_terminators` which allow users to modify the terminators when copying grammars during inheritance. Where previous grammars suggested an insertion point (i.e. used a `before` on the copy), I've removed this: it's not necessary for terminators.

To get full test coverage, this also includes #5153 which adds test coverage for a selection of `__repr__` methods (some which existed before and some not).


src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -135,24 +134,10 @@ class BaseGrammar(Matchable):
...
-    @staticmethod
-    def _resolve_ref(
-        elem: Union[None, str, MatchableType]
-    ) -> Union[None, MatchableType]:
         """Resolve potential string references to things we can match against."""
-        if elem is None:
-            return None
-        elif isinstance(elem, str):
+        if isinstance(elem, str):
             return Ref.keyword(elem)
```
alanmcruickshank, line 141: For reviewers: The rest of this refactor means that `_resolve_ref()` no longer needs to handle `None`. I think that's best, and allows this simplification.

src/sqlfluff/dialects/dialect_hive.py
```diff
@@ -979,24 +975,6 @@ class SetExpressionSegment(ansi.SetExpressionSegment):
...
-        "HAVING",
-        "QUALIFY",
-        # For window functions
-        "WINDOW",
-        Ref("FrameClauseUnitGrammar"),
-        "SEPARATOR",
-    )
-
-
 class ClusterByClauseSegment(ansi.OrderByClauseSegment):
```
alanmcruickshank, line 978: For reviewers: The ANSI dialect changed the grammar it used a while ago, and it's not one that allows terminators, so this code was redundant.

src/sqlfluff/dialects/dialect_sparksql.py
```diff
@@ -1825,25 +1825,6 @@ class GroupByClauseSegment(ansi.GroupByClauseSegment):
...
-        "QUALIFY",
-        # For window functions
-        "WINDOW",
-        Ref("FrameClauseUnitGrammar"),
-        "SEPARATOR",
-    )
-    parse_grammar = ansi.OrderByClauseSegment.parse_grammar
-
-
 class WithCubeRollupClauseSegment(BaseSegment):
```
alanmcruickshank, line 1828: Also redundant - see my not on the `hive` dialect.

greg-finley (approved):

PR #5153, opened 2023-09-03
`__repr__` methods for parsers

alanmcruickshank:
Draft PR while I check for existing coverage


alanmcruickshank:
I've merged this into #5152

PR #5162, opened 2023-09-05
Remove configurable `enforce_whitespace_preceding_terminator`

alanmcruickshank:
The justification for this isn't totally obvious.
- We're extending the use of terminators across dialects.
- Some terminators make sense to require whitespace before them, some don't.
- Currently this is configured on the _grammar_ (i.e. it's applied in the same way for all terminators).
- This doesn't make sense as we have more and more diverse things used as terminators.
- That means that we need a better way of controlling whether whitespace is required. That would either involve _per terminator configuration_ (which adds a lot of overhead), _OR_ we have some way of inferring whether whitespace is required based on the terminator itself.
- This PR implements the latter, inferring that if it's a _keyword_ (i.e. contains only letters), then it likely _does_ require whitespace before it (specifically the test case which highlights this is a snowflake one where `my_field:data:from:to:blah` was matching a `FROM` to terminate the clause). Symbols (like commas or semicolons) don't.

This means we can remove the current `enforce_whitespace_preceding_terminator` flag (which is a super long name anyway), but also opens the door for more flexible terminator usage across dialects.

In the process I've also fixed the `__repr__` method for `Ref` which was broken, and also un-nested the code flow a little in `greedy_match()`.

There were two locations in the grammar where I had to adjust the dialects to respond to this, both I think were incorrectly specified in the past and were just taking advantage of a loophole in their function. I think the new specs are much better. Only one parse yaml test changes as a result and it's just that the `select_statement` segments are no longer nested within `bracketed`.

### DIFF
--- a/src/sqlfluff/core/parser/grammar/base.py
+++ b/src/sqlfluff/core/parser/grammar/base.py
@@ -1009,7 +1009,7 @@ class Ref(BaseGrammar):
 
     def __repr__(self) -> str:
         return "<Ref: {}{}>".format(
-            ", ".join(str(self._elements)), " [opt]" if self.is_optional() else ""
+            str(self._ref), " [opt]" if self.is_optional() else ""
         )
 
     @match_wrapper(v_level=4)  # Log less for Ref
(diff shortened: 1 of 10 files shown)


greg-finley (approved):
Nice

PR #5167, opened 2023-09-05
Introduce `parse_mode` and remove `StartsWith` & `EphemeralSegment`.

alanmcruickshank:
Ok, this is a big one. There wasn't an easy way to make this much smaller.

Worth noting that it currently includes two currently unmerged PRs (#5180 & #5182), so some of the diff will look friendlier once they're merged.

For reviewers, think of this as three parts:
1. Adding new functionality to the `BaseGrammar` and a lot more to `Sequence`. This includes introducing the concept of `ParseModes` which dictate how greedy a grammar should be. As part of this I also needed to extract a couple more methods out of `BaseGrammar` and into `match_algorithms.py`. I would suggest focussing most of any review time on this work, in particular on the new tests I've introduced to cover these changes.
2. Working through each of the dialects to migrate from any instances of `StartsWith`, to using `Sequence(..., parse_mode=...)`. It was also necessary to migrate any usage of `ephemeral_name` (i.e. `EmphemeralSegment`) to using `parse_mode=` too (which was also the intent of doing this but I was hoping to do it as a separate step). This was _very tedious_, but it's worth noting that apart from a few examples in the `exasol` dialect where I needed to do a bit more of a rethink - none of the other parse test cases have changed. These changes cover many files, but are mostly just different ways of saying the same thing. Focus time here on `ANSI` and on `Exasol` where the biggest changes are.
3. Removing `StartsWith` and `EphemeralSegment` entirely including all references and test cases. This is a fairly easy to review (it's just deletes).

While this is quite a _large_ change, I don't think any of it is sufficiently breaking to users that it should trigger a significant release. In effect, this changes very slightly the handling of unparsable sections (which will now be trimmed of whitespace), and also the parse structure of some exasol expressions.

On the performance front, I've not done _extensive_ testing, but this does seem to reduce the time of the parsing test suite overall. In local testing it appears faster for some files and slower for others, although there's a lot more optimisation that could be done here. I would assume roughly performance neutral. In particular, I believe the treatment by the `Bracketed` grammar currently duplicates some logic, but this PR is already really big so I'm hesitant to take it even further when it's mergable in it's current state.

NOTE: I'm checking in on slack with the Exasol community on the implications for the `AM04` test case that I've had to change. Will report back when I have an answer.


alanmcruickshank:
Tests are passing at the moment because while this functionality is implemented, it's not actually _used_ in any of the dialects. Once #5162 is merged, then this can be used in several dialects straight away and should enabled single phase parsing in about _half_ of the current places. The other half will need a small extension to enable similar logic to work on `AnyNumberOf` (and therefore remove `GreedyUntil`). I'll do that as a second PR though.

alanmcruickshank:
I've done a bit more investigation into potential duplication of logic in `Bracketed`, and while it's definitely there, I think it's going to be easier to solve once this is merged and we're in single pass matching. There's a chunk of code in there designed to deal with matching sections of code which have already been parsed before and that's leading to some of the bloat. I'm going to leave this PR as is.

Yes it might be a slight performance drop, but it should unlock a bigger performance win in the longer run.

greg-finley (approved):

PR #5171, opened 2023-09-05
Flip the if in sequence and un-nest

alanmcruickshank:
Simple code flow refactor of the `match()` method of `Sequence`. All I've done is flipped some `if` statements to deal with the negative cases first and then reduce nesting afterward.


pwildenhain (approved):
Nice find!

PR #5172, opened 2023-09-05
Pull terminator setting up into the base grammar

alanmcruickshank:
This resolves some currently duplicated code in a few places across the grammars. With this PR, all grammars delegate terminator management to the base grammar. This reduces duplication and redundancy across the grammars.

### DIFF
--- a/src/sqlfluff/core/parser/grammar/base.py
+++ b/src/sqlfluff/core/parser/grammar/base.py
@@ -900,7 +900,7 @@ class BaseGrammar(Matchable):
         before: Optional[Any] = None,
         remove: Optional[List[MatchableType]] = None,
         terminators: List[Union[str, MatchableType]] = [],
-        add_terminators: List[Union[str, MatchableType]] = [],
+        replace_terminators: bool = False,
         # NOTE: Optionally allow other kwargs to be provided to this
         # method for type compatibility. Any provided won't be used.
         **kwargs: Any,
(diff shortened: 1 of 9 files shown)


src/sqlfluff/core/parser/grammar/base.py
```diff
@@ -876,6 +899,8 @@ def copy(
         at: Optional[int] = None,
         before: Optional[Any] = None,
         remove: Optional[List[MatchableType]] = None,
+        terminators: List[Union[str, MatchableType]] = [],
+        add_terminators: List[Union[str, MatchableType]] = [],
```
greg-finley, line 903: With these situations where we are raising a runtime error if both `terminators` and `add_terminators` are set, I try to look for ways to refactor to ensure that it's impossible instead.

What about one variable for the terminators and one variable about whether to add to existing?
```suggestion
        terminators: List[Union[str, MatchableType]] = [],
        add_to_existing_terminators: bool = False,
```

Or the `terminators` argument itself is a tuple or dataclass or something that has both the list and a boolean or enum expressing whether to add or override.

greg-finley (approved):
Approving in case you disagree with the comment or want to tackle it in a separate PR.

src/sqlfluff/core/parser/grammar/base.py
alanmcruickshank: I like that idea. I'll do something like that 👍

PR #5173, opened 2023-09-06
Adding "create table as" for greenplum dialect

bpfaust:

Attempts to add the "create table as" syntax for greenplum based on the same syntax from postgres.  Main addition is the DistributedBySegment grammar which should be at the end of creation statements within greenplum.

Also added support for non-boolean table parameters (ie. compression algorithm names)

None - should be self-contained to the new greenplum dialect.
yaml test generator is failing due to parse errors - would appreciate help in determining why

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_greenplum.py
```diff
@@ -167,13 +248,77 @@ class CreateTableStatementSegment(postgres.CreateTableStatementSegment):
...
+        ),
+        "AS",
+        OneOf(
+            OptionallyBracketed(Ref("SelectableGrammar")),
+            OptionallyBracketed(Sequence("TABLE", Ref("TableReferenceSegment"))),
+            Ref("ValuesClauseSegment"),
+            OptionallyBracketed(Sequence("EXECUTE", Ref("FunctionSegment"))),
         ),
+        Ref("WithDataClauseSegment", optional=True),
+        Ref("DistributedBySegment"),
```
WittierDinosaur, line 323: this is probably your issue - this wants to be optional

src/sqlfluff/dialects/dialect_greenplum.py
```diff
@@ -167,13 +248,77 @@ class CreateTableStatementSegment(postgres.CreateTableStatementSegment):
...
+                Ref("TemporaryGrammar"),
+            ),
+            "UNLOGGED",
+            optional=True,
+        ),
+        "TABLE",
+        Ref("IfNotExistsGrammar", optional=True),
+        Ref("TableReferenceSegment"),
+        AnyNumberOf(
+            Sequence(
```
WittierDinosaur, line 278: You don't need this Sequence(), you can stick the optional=True in Bracketed()

bpfaust: Ok - this was copied straight from the postgres dialect file so I left it alone

alanmcruickshank:
I've had a play around with this PR and the problem is that the `distributed_by` is being claimed by the `StatementSegment`. While you've updated the various terminator grammars, unfortunately the one you'd need isn't a grammar 🤦 (no reason you should know that either).

This is going to be easier once I've merged #5186. I'll give you a hand here once that's merged.

alanmcruickshank:
@bpfaust - I've just merged all the recent changes in to see how the tests behave. To see what I meant in the previous comment, check out how the `postgres` grammar treats the `SelectStatementSegment` and how it changes the `terminators` on it. We'll probably need to do something similar within this dialect to also add the new `DISTRIBUTED BY` segment as a terminator to that too.

src/sqlfluff/dialects/dialect_greenplum.py
alanmcruickshank: I sorted this.

src/sqlfluff/dialects/dialect_greenplum.py
alanmcruickshank: And sorted this.

alanmcruickshank (approved):
I realise I've made a bunch of changes here, but I think we're otherwise good to go.

PR #5179, opened 2023-09-07
Quality of life improvements on parse fixture script

alanmcruickshank:
The parsing script fails quite noisily when it fails on a file and that can make it quite hard to pinpoint the issue. This makes it a bit neater and easier to find the issue.


greg-finley (approved):
😍

PR #5177, opened 2023-09-07
Refactor of Sequence match

alanmcruickshank:
This ports some optimisations from #5143 over to the old `MatchResult` format, because I'd like to build on them for #5167.

Same functionality ⚒️ 
Fewer lines of code ✂️ 
Better comments 📜 
Less duplication, less nesting and simpler code flow 😸 

I'm hoping this should be fairly uncontroversial.

PR #5180, opened 2023-09-08
Output coverage report direct to PR

alanmcruickshank:
Coveralls is being buggy again.

This posts the coverage report from out internal coverage report directly onto the PR using a comment. Tested on #5177.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -383,8 +383,8 @@ jobs:
           script: |
             // First list the existing comments
 
-            const trigger_str = 'Coverage Results'
-            console.log("Getting existing comments...")
+            const trigger_str = 'Coverage Results';
+            console.log("Getting existing comments...");
 
             const comments = await github.paginate(
               github.rest.issues.listComments,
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-tests.yml
```diff
@@ -358,14 +358,81 @@ jobs:
...
+            const comments = await github.paginate(
+              github.rest.issues.listComments,
+              {
+                owner: 'sqlfluff',
+                repo: 'sqlfluff',
+                issue_number: context.issue.number
+              }
+            )
+
+            comment_id = null
```
greg-finley, line 398: I slightly prefer `let comment_id = null`

.github/workflows/ci-tests.yml
```diff
@@ -358,14 +358,81 @@ jobs:
...
+            for (const comment of comments) {
+              if (comment.body.indexOf(trigger_str) >= 0) {
+                console.log("Found target comment ID: %d", comment.id)
+                comment_id = comment.id
+              } else {
+                console.log("Comment ID %d not valid with body:\n%s.", comment.id, comment.body)
+              }
+            }
+
+            previous_outcome = "${{ steps.report_coverage.outcome }}"
```
greg-finley, line 410: I slightly prefer `const previous_outcome`

.github/workflows/ci-tests.yml
```diff
@@ -358,14 +358,81 @@ jobs:
...
+            console.log("Previous coverage step outcome: %s", previous_outcome)
+            if (previous_outcome == "success") {
+              status_emoji = "✅"
+            } else {
+              status_emoji = "⚠️"
+            }
+
+            const { promises: fs } = require('fs')
+            let content = await fs.readFile('coverage-report.txt', 'utf8')
```
greg-finley, line 419: `content` could be a `const`

greg-finley (approved):
Nice 👍 

Left some JS nits 😉 . Ideally would recommend to end all lines with semicolons also.

alanmcruickshank:
> > Nice 👍
> > Left some JS nits 😉 . Ideally would recommend to end all lines with semicolons also.
> 
> Since it works and maybe is unlikely to change, I would say feel free to ignore JS nits though if you prefer

They're all good suggestions and my JS-Fu is poor. I will take all the advice I can get!

.github/workflows/ci-tests.yml
```diff
@@ -358,14 +358,81 @@ jobs:
...
+            console.log("Got %d comments", comments.length)
+
+            for (const comment of comments) {
+              if (comment.body.indexOf(trigger_str) >= 0) {
+                console.log("Found target comment ID: %d", comment.id)
+                comment_id = comment.id
+              } else {
+                console.log("Comment ID %d not valid with body:\n%s.", comment.id, comment.body)
+              }
+            }
```
greg-finley, line 408: Fine, one more way that's slightly more JS-thonic

```js
comments.forEach(comment => {
  if (comment.body.indexOf(trigger_str) >= 0) {
    console.log("Found target comment ID: %d", comment.id);
    comment_id = comment.id;
  } else {
    console.log("Comment ID %d not valid with body:\n%s.", comment.id, comment.body);
  }
});
```

.github/workflows/ci-tests.yml
alanmcruickshank: On this one, I'm not sure I really get the impact of `let`, `const` and perhaps their combination. When should I use each?

greg-finley: `foo = true` could work for either setting or changing a variable, but `let foo = true` is saying that this is the first time we are setting this (and might help you find accidental variable shadowing)

`const foo = true` means `foo` cannot change later

.github/workflows/ci-tests.yml
```diff
@@ -358,14 +358,84 @@ jobs:
...
+        if: always()
+        with:
+          script: |
+            // First list the existing comments
+
+            const trigger_str = 'Coverage Results';
+            console.log("Getting existing comments...");
+
+            const issue_number = context.issue.number;
+            console.log("Issue number: " + issue_number);
```
greg-finley, line 390: @alanmcruickshank it's failing because
`Issue number: undefined`

greg-finley: Here's one suggestion, but I feel it wouldn't necessarily pick the right issue number: https://github.com/actions/github-script/issues/203#issuecomment-1027833024

greg-finley: We should probably do this? https://github.com/actions/github-script/issues/203#issuecomment-1034195974

PR #5183, opened 2023-09-09
Update coverage job to run in the right conditions

alanmcruickshank:
This is a bugfix for #5180. It wasn't updating the comment on coverage fails.


greg-finley (approved):
Ah, such a weird syntax. Thanks!

PR #5186, opened 2023-09-10
Terminators on `Anything()` + Strip _most_ of the other `parse_grammar`

alanmcruickshank:
This is another step on #5124.

It only adds a little more functionality (making the `Anything()` grammar terminator aware), but mostly cleans out a bunch of grammars using `parse_grammar` following #5167.

In particular it removes `parse_grammar` from `StatementSegment` (and all the dialect versions of that). That one didn't need any special features to remove now that we're part way through the work.

After this I believe that the only remaining parse grammars are in the bigquery dialect, which will need a little more work.

### DIFF
--- a/test/dialects/ansi_test.py
+++ b/test/dialects/ansi_test.py
@@ -1,7 +1,8 @@
 """Tests specific to the ansi dialect."""
 
-import pytest
 import logging
+
+import pytest
 
 from sqlfluff.core import FluffConfig, Linter
 from sqlfluff.core.parser import Lexer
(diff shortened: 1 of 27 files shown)


test/dialects/ansi_test.py
```diff
@@ -163,12 +164,12 @@ def test__dialect__ansi_specific_segment_not_match(
         ("SELECT 1 + (2 ", [(1, 12)]),
         # Set expression with inappropriate ORDER BY or LIMIT. Error
         # raised on the UNION.
-        ("SELECT * FROM a ORDER BY 1 UNION SELECT * FROM b", [(1, 27)]),
-        ("SELECT * FROM a LIMIT 1 UNION SELECT * FROM b", [(1, 24)]),
-        ("SELECT * FROM a ORDER BY 1 LIMIT 1 UNION SELECT * FROM b", [(1, 35)]),
+        ("SELECT * FROM a ORDER BY 1 UNION SELECT * FROM b", [(1, 28)]),
+        ("SELECT * FROM a LIMIT 1 UNION SELECT * FROM b", [(1, 25)]),
+        ("SELECT * FROM a ORDER BY 1 LIMIT 1 UNION SELECT * FROM b", [(1, 36)]),
```
alanmcruickshank, line 169: NOTE: The only change here is that the unparsable segment now starts at the code element, not the whitespace before. I think this is an improvement.

greg-finley (approved):
Nice! It was a bit confusing to know when to put `parse_grammar` in there, so I'm glad to see it's going away.

PR #5187, opened 2023-09-11
Parse modes for `AnyNumberOf`

alanmcruickshank:
Another step on #5124 .

Combined with #5186 , this removes the last parts of `parse_grammar` (apart from the `FileSegment` which is the final puzzle piece and will come last). We do this by adding greedy support to `AnyNumberOf` (and therefore to `OneOf` and all the other similar grammars).

Interestingly, most of the effects of this are in the `bigquery` dialect.


pwildenhain (approved):
💯

PR #5190, opened 2023-09-11
Refactor the Bracketed grammar `.match()`

alanmcruickshank:
This was referenced as necessary after #5167 where we refactored `Sequence()` but not `Bracketed()`. Single pass matching opens up some really nice optimisations for bracketed matching because we no longer need to deal with the possibility of brackets _already being there_.

(I think).

Draft while I check this is a good idea.


alanmcruickshank:
I'm closing this. It's now been effectively included in #5230.

PR #5194, opened 2023-09-12
Revise segment whitespace validation

alanmcruickshank:
Within segments, there is a convention that they do not start or end with whitespace. If that was going to occur, the same whitespace segments should be delegated up to a parent segment such that they aren't at the ends (except for the outer `FileSegment`, or more specifically, any segment which sets `can_start_end_non_code`).

There was a consistency check whenever fixes were applied which made sure that was the case. _However_ it was only applied if there was a `parse_grammar`. This is bad a) because nothing has a parse grammar anymore and b) because it was mostly not applied anyway c) because it _also_ wasn't being applied on new segments.

This PR resolves that by making the test not dependent on `parse_grammar`, and also applying it more liberally. It's not got a lot of computational overhead so, not a big deal to apply in more places. There's a new method for it (`validate_non_code_ends()`), and it's applied both on `BaseSegment` instantiation and on `apply_fixes()`.

Working through the implications of _better validation_ meant I found several bugs an inconsistencies which needed to be resolved:
- `Delimited` was consuming extra whitespace.
- `LintFix` objects of type `replace` weren't appropriately passing whitespace segments up to parent segments.
- `UnparsableSegments` were consuming too much whitespace

On a somewhat unrelated note it also highlighted a test case for the ansi dialect which doesn't parse in the ANSI dialect. That's also now resolved.

### DIFF
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -1499,7 +1499,7 @@ class BaseSegment(metaclass=SegmentMetaclass):
 
         # Rather than fix that here, we simply assert that it has been
         # done. This will raise issues in testing, but shouldn't in use.
-        if not r.can_start_end_non_code and seg_buffer:
+        if not self.can_start_end_non_code and seg_buffer:
             assert self._find_start_or_end_non_code(seg_buffer) is None, (
                 "Found inappropriate fix application: inappropriate "
                 "whitespace positioning. Post `_choose_anchor_segment`. \n\n"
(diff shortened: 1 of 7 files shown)


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1304,14 +1300,21 @@ def parse(
                     # Incomplete match.
                     # For now this means the parsing has failed. Lets add the unmatched
                     # bit at the end as something unparsable.
-                    # TODO: Do something more intelligent here.
+                    # NOTE: Don't claim any additional whitespace in the failed match.
+                    _idx = 0
+                    for _idx in range(len(m.unmatched_segments)):
```
greg-finley, line 1305:
```suggestion
                    for _idx in range(len(m.unmatched_segments)):
```

Seems like no need to set this to zero before doing the for loop.

greg-finley (approved):

src/sqlfluff/core/parser/segments/base.py
alanmcruickshank: I've got into the habit of doing it this way in case the tuple is empty, perhaps over cautiously. I'll leave it as is for now because I might rethink this whole method anyway

PR #5196, opened 2023-09-12
Validate fix parsing based on match_grammar

alanmcruickshank:
This is the third of a few of these changes (all relating to #5124). This updates the parsing checks after fixes to use `.match()` instead of `.parse()`.

Some of our internal consistency checks only work if there's a parse_grammar. This obviously presents some issues because these won't fail anymore, but that means they also don't actually work.

The effect of this PR is that we're much more extensively and effectively testing the impact on fixes to make sure we don't break parsing of the file in question (or at least don't make it worse).

This is going to look like a strange combination of things. Here's why:
1. The core of this PR is the changes to `linter.py` and `segments/base.py`. It takes the `_validate_segment_after_fixes` and uses `.match()` rather than `.parse()`. This is a **great thing**, because it means we can much more accurately identify segments that do and don't parse.
2. There are a chunk of dialect changes which I found in this process where exactly that was happening - fixes were being applied that were breaking SQL and we hadn't noticed 🤯. Particularly in the layout config which you'll see I've edited a few times. I've resolved _most_ of this in #5205.
3. One of the big places that it was happening is when an inner segment becomes unparsable, but the outer one is fine, just because of exactly how fixes were being applied. In those cases, the "unparsable" warning just _bubbles up_ and we re-test the outer segment until we get something that works _OR_ we get to the root segment. If we get to the root, rather than raising an exception, we skip the rule, so that the process can continue.

**For Reviewers**: Makes sure you've got the "show whitespace" review option _off_ otherwise this will be really hard to read.

### DIFF
--- a/src/sqlfluff/core/linter/linted_dir.py
+++ b/src/sqlfluff/core/linter/linted_dir.py
@@ -110,7 +110,7 @@ class LintedDir:
             raise ValueError(
                 ".tree() cannot be called when a LintedDir contains more than one file."
             )
-        elif not self.files:
+        elif not self.files:  # pragma: no cover
             raise ValueError(
                 "LintedDir has no parsed files. There is probably a parsing error."
             )
--- a/src/sqlfluff/core/linter/linted_dir.py
+++ b/src/sqlfluff/core/linter/linted_dir.py
@@ -110,4 +110,8 @@ class LintedDir:
             raise ValueError(
                 ".tree() cannot be called when a LintedDir contains more than one file."
             )
+        elif not self.files:
+            raise ValueError(
+                "LintedDir has no parsed files. There is probably a parsing error."
+            )
         return self.files[0].tree
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -1546,7 +1546,7 @@ class BaseSegment(metaclass=SegmentMetaclass):
         if not rematch.is_complete():
             return False
         opening_unparsables = set(segment.recursive_crawl("unparsable"))
-        closing_unparsables = set()
+        closing_unparsables: Set[BaseSegment] = set()
         for seg in rematch.matched_segments:
             closing_unparsables.update(seg.recursive_crawl("unparsable"))
         # Check we don't introduce any _additional_ unparsables.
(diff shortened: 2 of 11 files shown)


src/sqlfluff/core/linter/linted_dir.py
```diff
@@ -110,4 +110,8 @@ def tree(self) -> Optional[BaseSegment]:
             raise ValueError(
                 ".tree() cannot be called when a LintedDir contains more than one file."
             )
+        elif not self.files:
+            raise ValueError(
+                "LintedDir has no parsed files. There is probably a parsing error."
+            )
```
alanmcruickshank, line 116: This change was just to help debugging. I think it's still useful though.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1345,156 +1345,169 @@ def apply_fixes(
...
-                            )
-                            if f.edit_type == "delete":
-                                # We're just getting rid of this segment.
-                                pass
-                            elif f.edit_type in (
-                                "replace",
-                                "create_before",
-                                "create_after",
+        if not fixes or self.is_raw():
+            return self, [], [], True
```
alanmcruickshank, line 1349: I've un nested a clause here by bringing this further up and then returning.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1345,156 +1345,169 @@ def apply_fixes(
...
+
+        # If any fixes applied, do an intermediate reposition. When applying
+        # fixes to children and then trying to reposition them, that recursion
+        # may rely on the parent having already populated positions for any
+        # of the fixes applied there first. This ensures those segments have
+        # working positions to work with.
+        if fixes_applied:
+            seg_buffer = list(
+                self._position_segments(tuple(seg_buffer), parent_pos=self.pos_marker)
+            )
```
alanmcruickshank, line 1443: This section is unchanged - it's just one less indentation.

alanmcruickshank: Definitely review this with the "hide whitespace" github option on, otherwise it's really hard to read.

greg-finley (approved):

Issue #5195, opened 2023-09-12
sqlfluff errors on v2.2.0 (pip3/Ubuntu) with "name 'completion' is not defined"


jdahlbom:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

I upgraded from 2.1.3 to latest version and running sqlfluff started giving this error:
Traceback (most recent call last):
  File "/home/x/.local/bin/sqlfluff", line 5, in <module>
    from sqlfluff.cli.commands import cli
  File "/home/x/.local/lib/python3.8/site-packages/sqlfluff/cli/commands.py", line 23, in <module>
    from sqlfluff.cli.autocomplete import shell_completion_enabled, dialect_shell_complete
  File "/home/x/.local/lib/python3.8/site-packages/sqlfluff/cli/autocomplete.py", line 16, in <module>
    def dialect_shell_complete(ctx, param, incomplete) -> List[completion.CompletionItem]:
NameError: name 'completion' is not defined

I have installed sqlfluff via pip3 on Ubuntu LTS 20.04

Version 2.1.4 is the last tagged version that works as expected, and starting from version 2.2.0 the above error
occurs whenever I try to run sqlfluff parse or sqlfluff lint

### Expected Behaviour

Running sqlfluff lint <file> should start processing the file(s).

### Observed Behaviour

The sqlfluff breaks with this error message:
Traceback (most recent call last):
  File "/home/x/.local/bin/sqlfluff", line 5, in <module>
    from sqlfluff.cli.commands import cli
  File "/home/x/.local/lib/python3.8/site-packages/sqlfluff/cli/commands.py", line 23, in <module>
    from sqlfluff.cli.autocomplete import shell_completion_enabled, dialect_shell_complete
  File "/home/x/.local/lib/python3.8/site-packages/sqlfluff/cli/autocomplete.py", line 16, in <module>
    def dialect_shell_complete(ctx, param, incomplete) -> List[completion.CompletionItem]:
NameError: name 'completion' is not defined

### How to reproduce

Running on Ubuntu:
pip3 install sqlfluff==2.1.4
sqlfluff lint <file.sql>
--> Should start running the linting

pip3 uninstall sqlfluff
pip3 install sqlfluff==2.2.0
sqlfluff lint <file.sql>
Breaks with the error about the completion.

### Dialect

Redshift, but the error occurs even when not defining the dialiect and using default configuration.

[...]

PR #5197, opened 2023-09-12
Update click dependency to >=8.1.7

jdahlbom:
Updates the requirements.txt to specify version `click>=8.1.7`
Makes progress on #5195  (does not fix the broken code logic there, just bypasses the issue for users)
Any future users running `pip3 install sqlfluff` with existing old version of `click` will have required click version autoinstalled.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ appdirs
 backports.cached-property; python_version < '3.8'
 # To get the encoding of files.
 chardet
-click
+click>=8.1.7
 colorama>=0.3
 # Used for diffcover plugin
 diff-cover>=2.5.0


alanmcruickshank:
@jdahlbom , thanks for putting this together. For context there's a bit of history on this where we used to require a more recent version, but that was challenging for other users for different reasons: https://github.com/sqlfluff/sqlfluff/issues/2543

I'll bring this up with the other maintainers and work out what to do here.

Thanks for your patience 🙏

jdahlbom:
> Is there a reason you choose 8.1.7? Would be good to know what the minimum version we need is to avoid the issue, but also not create undue constraints.

No, that was the most recent version at the moment I created the issue and it could well be that ^8.1.0 or ^8 would do just as well to fix this issue.

alanmcruickshank:
I've suggested an alternative solution to this problem with #5293. If that one is merged then I'll close this PR. Until then I'll leave this open.

alanmcruickshank:
I'm going to close this PR as I think the alternative solution is simpler and doesn't change our dependency makeup.

Issue #4922, opened 2023-06-15
Namespaced types are being split


eliasmistler:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

When using type names in a schema, spaces are added around the `.`

```sql
CREATE TYPE my_schema.my_enum_type AS ENUM('val1', 'val2');

ALTER COLUMN produced_or_retailed TYPE my_schema.my_enum_type;
```

### Expected Behaviour

`my_schema.my_enum_type` remains as a namespaced type name

### Observed Behaviour

gets split to `my_schema . my_enum_type`:

```
L: 197 | P:  56 | LT01 | Expected single whitespace between naked identifier and
                       | dot '.'. [layout.spacing]
L: 197 | P:  57 | LT01 | Expected single whitespace between dot '.' and data type
                       | identifier. [layout.spacing]
L: 198 | P:  45 | LT01 | Expected single whitespace between naked identifier and
                       | dot '.'. [layout.spacing]
L: 198 | P:  46 | LT01 | Expected single whitespace between dot '.' and data type
                       | identifier. [layout.spacing]
```

### How to reproduce

- copy the example SQL into a new file
- set up `pre-commit` with the following config (`.pre-commit-config.yaml`):
```yaml
repos:
  - repo: https://github.com/sqlfluff/sqlfluff
    rev: 2.0.7
    hooks:
      - id: sqlfluff-lint
      - id: sqlfluff-fix
```
- run `pre-commit install` and then `pre-commit`

### Dialect

PostgreSQL

### Version

2.0.7

### Configuration

```toml
[tool.sqlfluff.core]
dialect = "postgres"
exclude_rules = ["ST07", "RF02", "AL03", "AM04"]
max_line_length = 120
[tool.sqlfluff.indentation]
allow_implicit_indents = true
[tool.sqlfluff.layout.type.alias_expression]
spacing_before = "align"
align_within = "select_clause"
align_scope = "bracketed"
[sqlfluff.rules.capitalisation.keywords]
capitalisation_policy = "upper"
[sqlfluff.rules.capitalisation.identifiers]
capitalisation_policy = "lower"
[sqlfluff.rules.capitalisation.functions]
extended_capitalisation_policy = "lower"
[sqlfluff.rules.capitalisation.literals]
capitalisation_policy = "upper"
[sqlfluff.rules.capitalisation.types]
[...]

PR #5205, opened 2023-09-13
Dialect spacing & quoting issues

alanmcruickshank:
This is a selection of issues found while putting together #5196. I'm separating them out because they're not really core functionality changes, more dialect improvements and rule test cases.

Most of these issues haven't been reported yet, although this does make some progress on #4922 (although that issue has other things that also need to be solved).


pwildenhain (approved):
Nice work 😄

PR #5209, opened 2023-09-14
Position assertions in BaseSegment

alanmcruickshank:
This is a tiny nit, but I've added some debug strings to these assertions to help understand what's going if they fail. They only matter if there's a bug elsewhere so they're really only to help other developers.


pwildenhain (approved):
🔍 ✔️

PR #5217, opened 2023-09-18
Parser test nits

alanmcruickshank:
This is almost entirely a _find-and-replace_ job (with a couple of other nits).

In particular, lots of tests refer to `seg_list` or `raw_seg_list`, and both are **always tuples**. This renames them so they're not labelled as lists.

Other than that, this fixes some typos, adds a few more test cases to `.simple()`. Hopefully non-controversial, and would otherwise complicate a separate PR.

### DIFF
--- a/test/core/parser/segments/segments_file_test.py
+++ b/test/core/parser/segments/segments_file_test.py
@@ -1,14 +1,6 @@
 """Test the BaseFileSegment class."""
 
-import pytest
-
 from sqlfluff.core.parser import BaseFileSegment
-
-
-@pytest.fixture(scope="module")
-def raw_segments(generate_test_segments):
-    """Construct a list of raw segments as a fixture."""
-    return generate_test_segments(["foobar", ".barfoo"])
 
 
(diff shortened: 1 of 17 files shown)


test/core/parser/segments/segments_file_test.py
```diff
@@ -6,14 +6,14 @@
 
 
 @pytest.fixture(scope="module")
-def raw_seg_list(generate_test_segments):
+def raw_segments(generate_test_segments):
     """Construct a list of raw segments as a fixture."""
     return generate_test_segments(["foobar", ".barfoo"])
```
pwildenhain, line 11: Isn't a fixture like this already defined in the conftest.py for segments?

pwildenhain (approved):
Looks good, just one question 🙋‍♂️ which may be useful

test/core/parser/segments/segments_file_test.py
alanmcruickshank: Ohh, good point. Will remove it.

PR #5224, opened 2023-09-21
Resolve #3176: Snowflake unparsable array casting

### LINKED ISSUE #3176: Snowflake - Unparsable array casting SQL
### Search before asking

### What Happened

```
CREATE OR REPLACE TABLE foo.foo AS
	SELECT raw:foo::array[0]::string  as channel
	FROM foo.bar
	;
```
The index of the array is unparsable.

### Expected Behaviour

-

### Observed Behaviour

-

### How to reproduce

-

### Dialect

Snowflake

### Version

0.13.0

### Configuration

-

### Code of Conduct

andychannery:

Resolves #3176. 

Previously, using bracket notation (i.e. '[ ]') to access a casted array (casted using '::' notation) resulted in an unparsable error. Also added corresponding test cases.

Currently, array slices seem to be allowed during array access (i.e. my_array[0:3]) but I believe Snowflake does not support this. Rather, they use [ARRAY_SLICE()](https://docs.snowflake.com/en/sql-reference/functions/array_slice).
Since my changes depend upon `ArrayAccessorSegment`, these changes will also allow array slicing (i.e. foo::array[0:3]). 

### Pull Request checklist

  - Other.


andychannery:
First time contributor here!

Open to any advice and feedback.

While working on resolving the reported bug, I noticed that the snowflake dialect currently allows array slicing (which I believe is not actually supported by Snowflake). Assuming I'm correct, would it be preferable that I also include changes to prevent array slicing within the same PR or should I open another PR?

alanmcruickshank:
> First time contributor here!
> 
> Open to any advice and feedback.
> 
> While working on resolving the reported bug, I noticed that the snowflake dialect currently allows array slicing (which I believe is not actually supported by Snowflake). Assuming I'm correct, would it be preferable that I also include changes to prevent array slicing within the same PR or should I open another PR?

Heya, thanks for contributing this. I think it's ok to leave it. The SQLFluff dialects tend to be a little less strict than the native parsers, which allows us to work with _slightly broken SQL_ and still improve it.

alanmcruickshank (approved):
Pending passing tests, I think this looks great. Thanks for contributing 👍

PR #5226, opened 2023-09-21
Spark Accessor Grammars

alanmcruickshank:
This came up while working on the next phase of #5124.

The way that JSON accessing is done in the SparkSQL dialect was very fragile. I've instead copied some of the same methodology from the Snowflake dialect which I think is more robust and ported it across to SparkSQL.

In the process, the `Accessor_Grammar` was (I think) the only grammar with an underscore, so I've also renamed it in all dialects to `AccessorGrammar` in line with the convention.

### DIFF
--- a/test/generate_parse_fixture_yml.py
+++ b/test/generate_parse_fixture_yml.py
@@ -98,7 +98,7 @@ def generate_one_parse_fixture(
 
     try:
         tree = parse_example_file(dialect, sqlfile)
-    except SQLParseError as err:
+    except Exception as err:
         # Catch parsing errors, and wrap the file path only it.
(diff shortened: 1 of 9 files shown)


test/generate_parse_fixture_yml.py
```diff
@@ -98,7 +98,7 @@ def generate_one_parse_fixture(
 
     try:
         tree = parse_example_file(dialect, sqlfile)
-    except SQLParseError as err:
+    except Exception as err:
```
alanmcruickshank, line 101: This change is to help with more robust debugging.

pwildenhain (approved):
💯

PR #5232, opened 2023-09-25
Set type automatically within the lexer.

alanmcruickshank:
@pwildenhain brought up a good question here: https://github.com/sqlfluff/sqlfluff/pull/5229#pullrequestreview-1639883063

> For all of these, the "type" key in the kwargs dictionary is just the name of the segment. Is there a reason we need to double document this for each of these segments?

That _was_ necessary but I think it's easily resolvable. This PR does that.


pwildenhain (approved):
Love me some refactoring for cleaner code ❤️

PR #5234, opened 2023-09-25
Introduce "word" segment

alanmcruickshank:
**This is a follow-on from #5232.**

We've got a bit of a rats nest around the `CodeSegment` currently. It doesn't define a `type` but some segments look for it in the `TypedParser`. This raises some really gnarly bugs when trying to enable reparsing of segments cleanly (which have arisen as part of #5230).

This resolves that by introducing a `WordSegment` which _inherits from_ `CodeSegment` but allows us to distinguish better between things that look like words (such as keywords etc...) and things that don't (like symbols or quotes).

NOTE: There are some dependency changes we should make, in particular that `KeywordSegment` should inherit from `WordSegment` - but that's not supported by the current layout of `KeywordSegment` being part of `sqlfluff.core.parser` and `WordSegment` (and others) being withing `sqlfluff.dialects.ansi`. I'll address that as another PR because it makes it very hard to see the changes already in this PR.


pwildenhain (approved):
Clean change, well done 👏🏻

Issue #5233, opened 2023-09-25
Unparsable sections when creating Snowflake integration objects


Kylea650:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

When creating a [storage integration](https://docs.snowflake.com/en/sql-reference/sql/create-storage-integration) or [notification integration](https://docs.snowflake.com/en/sql-reference/sql/create-notification-integration) as per Snowflake documentation, sqlfluff is unable to parse certain sections. The issue seems to be parsing 'notification_prodifer =' and 'storage_provider ='.

**notification_integration.sql**
```
create or replace notification integration notification_integration
    enabled = true
    type = queue
    notification_provider = azure_storage_queue
    azure_storage_queue_primary_uri = 'https://azure.storage.url/'
    azure_tenant_id = 'azure-tenant-id';
```
```
L:   2 | P:   1 | LT02 | Line should not be indented.
                       | [layout.indent]
L:   3 | P:   1 | LT02 | Line should not be indented.
                       | [layout.indent]
L:   3 | P:  17 |  PRS | Line 3, Position 17: Found unparsable section: '\n
                       | notification_provider = azure_stora...'
```

**storage_integration.sql**
```
create or replace storage integration storage_intergration
    type = external_stage
    storage_provider = 'azure'
    enabled = true
    azure_tenant_id = 'azure-tenant-id'
    storage_allowed_locations = (
        'azure://container.blob.core.windows.net/location-1',
        'azure://container.blob.core.windows.net/location-2'
    );
```
```
L:   2 | P:   1 | LT02 | Line should not be indented.
                       | [layout.indent]
L:   2 | P:  26 |  PRS | Line 2, Position 26: Found unparsable section: "\n
                       | storage_provider = 'azure'\n    enab..."
```

Additionally, sqlfluff complains about the indenting which follows Snowflake's documentation.

### Expected Behaviour

sqlfluff should be able to correctly parse integration objects.

### Observed Behaviour

sqlfluff is unable to parse integration objects and disagrees with Snowflake's reccomended indentation.

### How to reproduce

Create two .sql files as per the examples above and run `sqlfluff lint` with dialect set to Snowflake.

### Dialect

Snowflake

### Version

```
sqlfluff, version 2.3.2
Python 3.10.9
```

### Configuration

No configuration files, sqlfluff ran as standard with dialect=snowflake.

[...]

PR #5236, opened 2023-09-25
add azure_storage_queue and quoted providers

Kylea650:
PR relating to bug from issue #5233 

### Notification integration
- As per Snowflake's documentation [create-notification-integration](https://docs.snowflake.com/en/sql-reference/sql/create-notification-integration), when creating an azure notification integration for data loading, this must be set to AZURE_STORAGE_QUEUE and not AZURE_EVENT_GRID. AZURE_EVENT_GRID is only for push notifications from Snowflake. Setting this to AZURE_STORAGE_QUEUE caused sqlfluff to fail due to it being un-parsable.
- Instead of adding a new test, I updated a previous one. When creating a notification integration, if you are setting a azure_storage_queue_primary_uri then you must set the notification provider to be azure_storage_queue so this test did not represent a real situation. 

### Storage integration
- As per Snowflake's documentation [create-storage-integration](https://docs.snowflake.com/en/sql-reference/sql/create-storage-integration), the storage_provider should be quoted, e.g. 'S3' instead of S3. Both work in Snowflake, but I think it makes sense to allow for both options to be parsable. Perhaps there's a nicer way to do this directly in ```dialect_snowflake.py``` instead of adding quoted versions in the keywords file.
- New test added for each of S3, Azure and GCP.


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3583,7 +3585,9 @@ class CreateStatementSegment(BaseSegment):
             Sequence("TYPE", Ref("EqualsSegment"), "EXTERNAL_STAGE"),
             Sequence("ENABLED", Ref("EqualsSegment"), Ref("BooleanLiteralGrammar")),
             Sequence(
-                "STORAGE_PROVIDER", Ref("EqualsSegment"), OneOf("S3", "AZURE", "GCS")
+                "STORAGE_PROVIDER",
+                Ref("EqualsSegment"),
+                OneOf("S3", "AZURE", "GCS", "'S3'", "'AZURE'", "'GCS'"),
```
alanmcruickshank, line 3590: For the quoted variants here, rather than adding them as keywords, can you change this to just be a `QuotedLiteralSegment` (keyword segments should not be quoted):

```suggestion
                OneOf("S3", "AZURE", "GCS", Ref("QuotedLiteralSegment")),
```

That will mean in our parse tests, it won't validate that the storage provider is a valid one - but I think that's likely beyond the scope of SQLFluff parse checking.

src/sqlfluff/dialects/dialect_snowflake_keywords.py
```diff
@@ -594,4 +595,7 @@
 XML
 ZONE
 ZSTD
+'AZURE'
+'GCS'
+'S3'
```
alanmcruickshank, line 600: See my note on `dialect_snowflake.py` - I think we should remove these quoted variants.

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -3496,7 +3496,9 @@ class CreateStatementSegment(BaseSegment):
             Sequence(
                 "NOTIFICATION_PROVIDER",
                 Ref("EqualsSegment"),
-                OneOf("AWS_SNS", "AZURE_EVENT_GRID", "GCP_PUBSUB"),
+                OneOf(
+                    "AWS_SNS", "AZURE_EVENT_GRID", "GCP_PUBSUB", "AZURE_STORAGE_QUEUE"
+                ),
```
alanmcruickshank, line 3501: Do you want to add a quoted option here too, as per my change request below?

alanmcruickshank (changes requested):
Thanks for putting this together! Can you change the handling of the quoted options here? Otherwise I think this looks good, thanks especially for adding some good test cases to stretch it.

src/sqlfluff/dialects/dialect_snowflake.py
Kylea650: Should now be resolved.

src/sqlfluff/dialects/dialect_snowflake.py
Kylea650: Should now be resolved.

src/sqlfluff/dialects/dialect_snowflake_keywords.py
Kylea650: Should now be resolved.

alanmcruickshank:
@Kylea650 - looks like you need to regenerate the yaml test cases since making those changes:
`python test\generate_parse_fixture_yml.py`

Kylea650:
> @Kylea650 - looks like you need to regenerate the yaml test cases since making those changes: `python test\generate_parse_fixture_yml.py`

Of course! Committing now.

alanmcruickshank (approved):
LGTM! Thanks for contributing 🏆

PR #5242, opened 2023-09-26
#5239 Added (basic) support for properly linted Snowflake scripting

### LINKED ISSUE #5239: Snowflake: Stored procedure are not parsable
### Search before asking

### What Happened

I am using SQLFluff with the Snowflake dialect, however it seems to be struggling with every single stored procedure we have (and we have quite a lot).

For example, in this stored procedure, it is already having an issue with the first line:

```
CREATE OR REPLACE PROCEDURE HEALTHCARE.INPUT.AS_GET_CSV_HEADER_PATTERN(SOURCE_DIR VARCHAR, PATTERN VARCHAR, MAX_COL NUMBER(38,0), IS_GZIP BOOLEAN)
RETURNS ARRAY
LANGUAGE SQL
AS
BEGIN
  -- Construct the statement
  -- 1. Start the SQL statement
  LET HEADER_STMT VARCHAR := 'SELECT ARRAY_CONSTRUCT_COMPACT(';
  -- 2. Iterate through the columns and add them to the statement
  FOR i IN 1 TO MAX_COL DO
    HEADER_STMT := HEADER_STMT || 't.$' || i;
    IF (i < MAX_COL) THEN
        HEADER_STMT := HEADER_STMT || ',';
    END IF;
  END FOR;
  -- 3. Finish the statement
  LET FILE_FORMAT VARCHAR := (SELECT CASE WHEN :IS_GZIP THEN 'HEALTHCARE.INPUT.DEFAULT_CSV_GZIP_NO_HEADER' ELSE 'HEALTHCARE.INPUT.DEFAULT_CSV_NO_HEADER' END);
  HEADER_STMT := HEADER_STMT || ') FROM \'@s3_stage/' || :SOURCE_DIR || '/\' (FILE_FORMAT => ' || :FILE_FORMAT || ', PATTERN => \'' || :PATTERN || '\') t LIMIT 1;';

  -- Execute the statement
  LET RESULT RESULTSET := (EXECUTE IMMEDIATE HEADER_STMT);

  -- Fetch the result and return it
  LET c CURSOR FOR RESULT;
  LET RESULT_ARRAY ARRAY;
  OPEN c;
  FETCH c INTO RESULT_ARRAY;

  -- Return
  RETURN RESULT_ARRAY;
END;
```

Error:
`Line 1, Position 1: Found unparsable section: 'CREATE OR REPLACE PROCEDURE HEALTHCARE.I...'sqlfluff(PRS)`

### Expected Behaviour

This should be parsable, as far I know

### Observed Behaviour

It just means that we can basically do no syntax testing whatsoever within Snowflake stored procedures (which is around 60% of our code basis).

### How to reproduce

Simply try to save the file above with the .sqlfluff configuration as described.

### Dialect

Snowflakie

### Version

sqlfluff, version 2.3.2
Python 3.11.4

### Configuration

[sqlfluff]

# Or run 'sqlfluff dialects'
dialect = snowflake

# Affect only .sql files in the current directory and below
sql_file_exts = .sql

templater = raw

exclude_rules = ambiguous.column_count, structure.column_order, references.keywords

# something a little longer.
max_line_length = 120

processes = 1 # default -1

[sqlfluff:templater:dbt]
project_dir = ./

[sqlfluff:indentation]
allow_implicit_indents = True
tab_space_size = 2

aliasing = explicit
aliasing = explicit
[sqlfluff:rules:aliasing.length]
min_alias_length = 1

capitalisation_policy = upper
capitalisation_policy = upper
capitalisation_policy = upper

### Code of Conduct

robin-alphasophia:
fixes #5239

So far the Snowflake dialect does not contain proper support for Snowflake scripting. As such, it:

1. Does not support `BEGIN <one or multiple statements> END` blocks as defined in the [Snowflake documentation](https://docs.snowflake.com/en/sql-reference/snowflake-scripting/begin)
2. Does not support *proper* Snowflake SQL scripting as the body of a CREATE PROCEDURE. I.e., you can attach a SQL code as its body using the `...AS $$` syntax, however this turns the entire function body into the equivalent of a comment that is fully ignored by the sqlfluff linter, taking away a lot of its utility. The`...AS BEGIN <multiple statements END` syntax, taking into account proper SQL, is not supported
3. Does not support `RETURN <expression>` syntax for returning variables

This pull request, introduces all these capabilities into sqlfluff. As you can see in the added textures, now BEGIN...END, BEGIN...END in CREATE STORED PROCEDURE as an alternative to unchecked `$$ code` and the RETURN statement are supported.

While there are many specific features that are still missing in support, this is a first step. 

Hopefully I will find some time in the near future to address the rest.

None as far as I know

### Pull Request checklist

  - Other.


alanmcruickshank (approved):
Looks like a good start. Thanks for putting this together 👍

PR #5243, opened 2023-09-26
Bring together the generic segments

alanmcruickshank:
This is a natural progression from #5234, where I mentioned that I'd do this as a separate step.

Segments like `LiteralSegment` and `WordSegment` currently reside in the ansi dialect, but they're treated like fundamental building blocks of dialects more generally. I think it makes sense to bring them together into `parser.segments` rather than in the ansi dialect. Much more the `KeywordSegment` which already lives centrally.

This also also allows `KeywordSegment` to inherit from `WordSegment` which is going to be important for #5230.

This _looks_ like a lot of changes, but really it's a copy paste of a chunk of the segment definitions from `dialect_ansi.py` into a new `segments/common.py` file, along with some of the more auxiliary segments from other parts of the exiting `segments` module.

As you'll also see - only one test implication (which adds `"word"` to the type inheritance of `KeywordSegment`), otherwise just code moving.


pwildenhain (approved):
Fall cleaning 🧹 🍂  of the codebase 😁

PR #5249, opened 2023-09-27
Unnest fixing and re-address validation triggers

alanmcruickshank:
Found this while working on #5230.

The `.apply_fixes()` method was very deeply nested, in a way that made the code hard to follow. This flips some `if` statments so that we can un-nest that code, make it easier to follow and cleaner.

In the process I've also slightly changed the conditions for validating things. The the segment _type_ is the same and we're just editing the content (e.g. capitalisation changes) then we don't do a full re-parse validation (and we trust that the rule is being sensible). For anything else more significant we do still trigger a validation though.

### DIFF
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -1302,8 +1302,8 @@ class BaseSegment(metaclass=SegmentMetaclass):
                     # segment on the end
                     seg_buffer.append(seg)
 
-            # Invalidate any caches
-            self.invalidate_caches()
+        # Invalidate any caches
+        self.invalidate_caches()
 
         # If any fixes applied, do an intermediate reposition. When applying
         # fixes to children and then trying to reposition them, that recursion
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -1219,79 +1219,90 @@ class BaseSegment(metaclass=SegmentMetaclass):
         after = []
         fixes_applied: List[LintFix] = []
         todo_buffer = list(self.segments)
-        while True:
-            if len(todo_buffer) == 0:
-                break
-            else:
-                seg = todo_buffer.pop(0)
+        requires_validate = False
 
-                # Look for uuid match.
-                # This handles potential positioning ambiguity.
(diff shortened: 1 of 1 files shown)


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1219,79 +1219,90 @@ def apply_fixes(
...
         fixes_applied: List[LintFix] = []
         todo_buffer = list(self.segments)
-        while True:
-            if len(todo_buffer) == 0:
-                break
-            else:
-                seg = todo_buffer.pop(0)
+        requires_validate = False
+
+        while todo_buffer:
```
alanmcruickshank, line 1224: rather than `while True: ... if todo_buffer...`, better to just have `while todo_buffer:`

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1219,79 +1219,90 @@ def apply_fixes(
...
-                # Look for uuid match.
-                # This handles potential positioning ambiguity.
-                anchor_info: Optional[AnchorEditInfo] = fixes.pop(seg.uuid, None)
-                if anchor_info is not None:
-                    seg_fixes = anchor_info.fixes
+            # Look for uuid match.
+            # This handles potential positioning ambiguity.
+            anchor_info: Optional[AnchorEditInfo] = fixes.pop(seg.uuid, None)
+
+            if anchor_info is None:
```
alanmcruickshank, line 1231: Flip this `if` to deal with the `is None` case first and then `continue`, followed by an un nested opposite case.

src/sqlfluff/core/parser/segments/base.py
```diff
@@ -1219,79 +1219,90 @@ def apply_fixes(
...
+                    requires_validate = True
+                    # NOTE: We don't add the segment in this case.
+                    continue
+
+                # Otherwise it must be a replace or a create.
+                assert f.edit_type in (
+                    "replace",
+                    "create_before",
+                    "create_after",
+                ), f"Unexpected edit_type: {f.edit_type!r} in {f!r}"
```
alanmcruickshank, line 1267: Changed this `if` to an `assert` and then removed the `raise` clause later.

greg-finley (approved):

PR #5252, opened 2023-09-28
Revise MatchableType -> Matchable

alanmcruickshank:
I've wanted to do this for *AAAAAAGES*.

For a long while the dependency path for the common matching interface has been quite complex. We define the interface fairly clearly with `Matchable`, but a similar interface needs to be defined on the _uninstantiated_ `BaseSegment` class.

This led to the slightly strange `MatchableType`, which was `Union[Type[BaseSegment], Matchable]`, but this PR adds `Matchable` to the inherited classes for `SegmentMetaclass`, which means we can get rid of `MatchableType` and use the a _single_ interface definition.

I've also removed `type` as a defined attribute on `Matchable` because it was poorly and inconsistently implemented (grammar classes didn't support it). There were a few places where we were using it, notably in bracket matching and in dialect replacements, but those were fairly easy to resolve.

### DIFF
--- a/src/sqlfluff/core/parser/match_algorithms.py
+++ b/src/sqlfluff/core/parser/match_algorithms.py
@@ -11,8 +11,8 @@ from sqlfluff.core.errors import SQLParseError
 from sqlfluff.core.parser.context import ParseContext
 from sqlfluff.core.parser.helpers import trim_non_code_segments
 from sqlfluff.core.parser.match_result import MatchResult
+from sqlfluff.core.parser.matchable import Matchable
 from sqlfluff.core.parser.segments import BaseSegment, BracketedSegment
-from sqlfluff.core.parser.types import MatchableType
 
 
 def first_trimmed_raw(seg: BaseSegment) -> str:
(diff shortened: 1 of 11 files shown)


src/sqlfluff/core/parser/match_algorithms.py
```diff
@@ -306,16 +309,8 @@ def bracket_sensitive_look_ahead_match(
...
-                        # For the start bracket we don't have the matcher
-                        # but we can work out the type, so we use that for
-                        # the lookup.
-                        start_index = [
-                            bracket.type for bracket in start_brackets
-                        ].index(bracket_stack[-1].bracket.get_type())
-                        # For the end index, we can just look for the matcher
-                        end_index = end_brackets.index(matcher)
-                        bracket_types_match = start_index == end_index
-                        if bracket_types_match:
```
alanmcruickshank, line 318: I'm glad I was able to clean up this logic, it's been a bit ugly for a while.

greg-finley (approved):

PR #5263, opened 2023-10-02
Finish the removal of `GreedyUntil`

alanmcruickshank:
After #5167 I managed to remove `StartsWith` but not `GreedyUntil`. #5186 added terminators to `Anything()`. The combination of those mean that `Anything` and `GreedyUntil` effectively wrap the same functionality. At this point there is also only _one_ mention of `GreedyUntil` left in the dialects.

This PR replaces that last mention with a terminated `Anything`, adds the logic for the `Anything` grammar to reset terminators and then totally strips out `GreedyUntil`. Where possible, I've migrated tests from `GreedyUntil` to `Anything` rather than totally removing them.


alanmcruickshank:
I think this may conflict slightly with #5230, but I suggest this one is merged first and then I'll resolve the conflict after that.

greg-finley (approved):

PR #5266, opened 2023-10-02
New Rule ST11: Detect unused tables in join

### LINKED ISSUE #5251: Rule: Unused table in join
### Search before asking

### Description

It would be nice with a rule that warns about unused tables in joins.
I realize that it might need to be opt-in, since there are legitimate reasons to do this, but it is often a bug.

There are already rules for unused aliases and unused CTEs, but first aliasing a CTE, and then joining it in without referring to any of its columns is an edge-case that falls between these rules. I really couldn't find any other issues that matched exactly this use case, but if I overlooked something, I'm sorry!

### Use case

If, during the development or the lifetime of a view, you have reduced which columns are selected to the point that a joined-in table is no longer referenced, it would be nice to be warned.

For example if you have had a view
```sql
with 
  widget as (
    select * from some_widget_dbt_table
  ),

  inventor as (
    select * from some_inventor_dbt_table
  )

select
  widget.id,
  widget.name,
  inventor.name
from
  widget
  left join inventor
    on widget.inventor_id = inventor.id
```

And then you decided at some point that the inventor name doesn't need to be in the output any more, and you get

```sql
select
  widget.id,
  widget.name,
from
  widget
  left join inventor
    on widget.inventor_id = inventor.id
```

Here, the join with inventor is no longer necessary, but in a big select-statement, you might easily overlook this.

### Dialect

I don't think this is specific to any dialects?
But it is particularly helpful in dbt, where convention often leads you to define your "inputs" as CTEs with aliases, and then use those aliases in the join, immediately making them "used".

### Code of Conduct

danparizher:
@alanmcruickshank edits:

Based on the original bones of this PR by @danparizher, I've fleshed out the logic so that it's a working rule. Fixes #5251.

NOTE: I've decided not to enable a _fix_ routine for this rule, because it's effectively ambiguous what the intent was - and just removing a join could change the granularity of a table if not done with care. I've also decided that this rule shouldn't return unless all references are qualified - meaning it's effectively dependent on RF02 to enforce that first.

Very open to feedback!


keraion (commented):
I am a fan of this rule. I know I've run into queries that have had those unneeded joins.

My concern is it might trigger a bit too often which might lead to people disabling it entirely.

```sql
-- getting a count of two joined tables triggers this rule.
SELECT COUNT(*) AS row_count -- could also be COUNT(1) or COUNT(0)
FROM taba AS a
INNER JOIN tabb AS b
    ON a.col1 = b.col1;

-- ANTI and SEMI joins will always trigger this as well, `b` cannot be referenced
SELECT a.col1, a.col2
FROM taba AS a
SEMI JOIN tabb AS b
    ON a.col1 = b.col1;

SELECT a.col1, a.col2
FROM taba AS a
ANTI JOIN tabb AS b
    ON a.col1 = b.col1;

-- an INNER JOIN might be used like an EXISTS/SEMI JOIN
SELECT
    a.key_column,
    a.col2,
    a.col3,
    a.col4
FROM all_records AS a
INNER JOIN targeted_records AS t
    ON a.key_column = t.key_column;
```

It might make more sense to only apply this to `LEFT`, `RIGHT`, and `FULL` joins and have an exception for joins that utilize `COUNT`.

Overall, I'm excited to see this rule in place in the near future!

src/sqlfluff/rules/structure/ST10.py
```diff
@@ -0,0 +1,235 @@
...
+    This rule will check if there are any tables that are referenced in the
+    ``FROM`` or ``JOIN`` clause of a ``SELECT`` statement, but where no
+    columns from that table are referenced in the any of the other clauses.
+
+    This rule relies on all of the column references in the ``SELECT``
+    statement being qualified with at least the table name, and so is
+    designed to work alongside :sqlfluff:ref:`references.qualification`
+    (:sqlfluff:ref:`RF02`).
+
+    This rule does not propose a fix, because it assumes that it an unused
```
barrywhart, line 27: Makes sense. I could imagine unusual cases where someone joins with a table simply for filtering purposes, and there's nothing "wrong" with that. (Although one could argue in that case `WHERE EXISTS` with a sub-select would be clearer about the _intent_.

src/sqlfluff/rules/structure/ST10.py
```diff
@@ -0,0 +1,235 @@
...
+            foo.a,
+            foo.b,
+            bar.c
+        FROM foo
+        LEFT JOIN bar ON foo.a = bar.a
+
+    In the (*very rare*) situations that it is logically necessary to include
+    a table in a join clause, but not otherwise refer to it (likely for
+    granularity reasons, or as a stepping stone to another table), we recommend
+    ignoring this rule for that specific line by using ``-- noqa: ST10`` at
```
barrywhart, line 63: Think it's worth mentioning `WHERE EXISTS` as another possible remedy?

test/fixtures/rules/std_rule_cases/ST10.yml
```diff
@@ -0,0 +1,65 @@
...
+    from (VALUES (1, 'one'), (2, 'two'), (3, 'three'))
+
+# If there are any unqualified references, we shouldn't raise
+# an issue until they're resolved.
+test_pass_unqualified_select:
+  pass_str: |
+    select
+        a
+    from b join c using(d)
+test_pass_unqualified_where:
```
barrywhart, line 33: Nit: Could you add blank lines between the test cases (for readability)?

test/fixtures/rules/std_rule_cases/ST10.yml
```diff
@@ -0,0 +1,65 @@
+rule: ST10
```
barrywhart, line 1: Have you considered the possibility that an "unused" table is actually referenced in a _subquery_? Below is a quick example of what I mean. I don't have a real-world situation in mind -- just threw together this query as an example.

```
SELECT a.col1
FROM a
JOIN b ON a.id = b.a_id
WHERE a.some_column IN (
    SELECT c.some_column
    FROM c
    WHERE c.other_column = b.col
)
```

src/sqlfluff/rules/structure/ST10.py
```diff
@@ -0,0 +1,235 @@
...
+            yield parts[-2].part.upper().strip("\"'`[]")
+
+    def _extract_references_from_select(
+        self, segment: BaseSegment
+    ) -> List[Tuple[str, BaseSegment]]:
+        assert segment.is_type("select_statement")
+        # Tables which exist in the query
+        joined_tables = []
+        # Tables which are referred to elsewhere.
+        # NOTE: We populate this here if a table is referred to in the the
```
barrywhart, line 125: Duplicate word: "the the"

barrywhart:
Could one or more of the maintainers run this rule on a large set of real-world queries? I do like the rule, but we may see surprise warnings from it for cases we hadn't considered.

src/sqlfluff/rules/structure/ST10.py
alanmcruickshank: To address this issue, I've added an `allowlist` of specific keywords, which includes `INNER` to prevent this situation. By making it configurable, people can make it more or less aggressive as they prefer.

alanmcruickshank:
> Could one or more of the maintainers run this rule on a large set of real-world queries? I do like the rule, but we may see surprise warnings from it for cases we hadn't considered.

I've run this on my current large project, and got a few flags from `INNER` joins being used as filters - but since adding the allowlist config - not issues. Seems to be running ok 👍

alanmcruickshank:
Update post maintainers call: I've removed the proposed config value, and now it only flags on explicit `LEFT`, `RIGHT` and `FULL` joins. I think this is otherwise ready for a review. @WittierDinosaur ?

test/fixtures/rules/std_rule_cases/ST10.yml
alanmcruickshank: @barrywhart - I've added your query (slightly adjusted) as a test case, and we're all good 👍 . Introspection into subqueries is covered.

src/sqlfluff/rules/structure/__init__.py
```diff
@@ -21,6 +21,17 @@ def get_configs_info() -> Dict[str, Any]:
                 "Defaults to ``earlier``."
             ),
         },
+        "allowlist_join_keywords": {
```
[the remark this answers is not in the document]
alanmcruickshank: GOOD CATCH. Yeah I can remove that 👍

src/sqlfluff/rules/structure/ST11.py
```diff
@@ -0,0 +1,252 @@
...
+    def _eval(self, context: RuleContext) -> List[LintResult]:
+        """Implement the logic to detect unused tables in joins.
+
+        First we fetch all the tables brought *into* the query via
+        either FROM or JOIN clauses. We then search for all the
+        tables referenced in all the other clauses and look for
+        mismatches.
+
+        NOTE: If *any* references aren't appropriately qualified,
+        this rule will abort (because it won't know how to resolve
```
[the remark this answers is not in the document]
alanmcruickshank: well... if all the tables are referenced by some of the columns, then there's no error - so this rule wouldn't fire anyway.

You're correct that we _could_ check if not all the columns are qualified, but if we'd still only kick up an error if we were _sure_ that a table was unused then we need *both* a table to be unreferenced *and* all columns to be qualified. Given that second condition is the easiest to check, it seems smart to do that first.

alanmcruickshank (approved):
Thanks @barrywhart for the review. I'll get this merged now those parts are addressed 👍

Issue #3479, opened 2022-06-22
Parsing error while using ALTER commands on Snowflake Objects


arumugam-pounraj:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Getting parsing error while trying to parse SQL statements which uses ALTER commands.

Tried parsing below ALTER statements with SQLFLUFF and got parsing error. The statements are from Snowflake docs
```
ALTER ACCOUNT SET TIMEZONE = 'UTC';
ALTER DATABASE DB_TEST RENAME TO DB_TEST_NEW;
ALTER RESOURCE MONITOR RM_TEST SET CREDIT_QUOTA=2000 TRIGGERS ON 80 PERCENT DO NOTIFY ON 100 PERCENT DO SUSPEND_IMMEDIATE;
ALTER STORAGE INTEGRATION TEST_MYINT SET ENABLED = TRUE;
ALTER NETWORK POLICY MYPOLICY1 SET ALLOWED_IP_LIST=('192.168.1.0/24', '192.168.255.100');
ALTER ROLE ROLE1 RENAME TO ROLE2;
ALTER SHARE SALES_S ADD ACCOUNTS = AB12345;
ALTER FILE FORMAT MY_FORMAT SET FIELD_DELIMITER = ',';
ALTER SEQUENCE SEQ_TEST SET INCREMENT = 2;
ALTER PIPE MYPIPE SET PIPE_EXECUTION_PAUSED = TRUE;
ALTER PROCEDURE IF EXISTS PROCEDURE1(float) RENAME TO PROCEDURE2;
ALTER MASKING POLICY EMAIL_MASK SET BODY ->
  CASE
    WHEN CURRENT_ROLE() IN ('ANALYST') THEN VAL
    ELSE SHA2(VAL, 512)
  END;
```

```
Error Output
L:  31 | P:   1 |  PRS | Line 31, Position 1: Found unparsable section: "ALTER ACCOUNT SET TIMEZONE = 'UTC'"
L:  32 | P:   1 |  PRS | Line 32, Position 1: Found unparsable section: 'ALTER DATABASE DB_TEST RENAME TO DB_TEST...'
L:  34 | P:   1 |  PRS | Line 34, Position 1: Found unparsable section: 'ALTER RESOURCE MONITOR RM_TEST SET CREDI...'
[...]

PR #5272, opened 2023-10-03
Snowflake: Add CreateResourceMonitorStatementSegment & AlterResourceMonitorStatementSegment

moreaupascal56:

- creates CreateResourceMonitorStatementSegment and AlterResourceMonitorStatementSegment

I had to add a lot of keywords and to create the TimestampSegment to check that the given date is a timestamp. 

makes progress on #3479

no

### Pull Request checklist

  - Other.


alanmcruickshank (changes requested):
I've got a few suggestions to simplify this one:
- I can see why you've made the `TimestapLiteral`, however the regex parser has some performance implications, and given the timestamps are always quoted I wonder whether `quoted_literal` would always be sufficient? Try `Ref("QuotedLiteralSegment")` instead. I realise that doesn't specifically indicate it as a timestamp, but that's what we've mostly done in other places.
- Second suggestion is that your section with an `AnySetOf` I _think_ is duplicated between the two grammars and I'm wondering whether it's worth wrapping it in a reusable section to reduce duplication? That would also make it easier to add new options in the future. If I've not noticed that there are some subtle differences between the two and the _do_ need to have separate definitions, feel free to keep your current implementation.

moreaupascal56:
> I've got a few suggestions to simplify this one:
> 
> * I can see why you've made the `TimestapLiteral`, however the regex parser has some performance implications, and given the timestamps are always quoted I wonder whether `quoted_literal` would always be sufficient? Try `Ref("QuotedLiteralSegment")` instead. I realise that doesn't specifically indicate it as a timestamp, but that's what we've mostly done in other places.
> * Second suggestion is that your section with an `AnySetOf` I _think_ is duplicated between the two grammars and I'm wondering whether it's worth wrapping it in a reusable section to reduce duplication? That would also make it easier to add new options in the future. If I've not noticed that there are some subtle differences between the two and the _do_ need to have separate definitions, feel free to keep your current implementation.

You are right! the TImestamp option was over engineered. I have pushed new commits to fix both of these issues :)

PR #5289, opened 2023-10-08
Use ruff rule I replace isort

zhongjiajie:

Use ruff rule I replace isort which can be less dependence
and fast runtime. I have format all the code, if we do not
want it I can revert the format

None

### Pull Request checklist

  - Other.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,0 +1,4 @@
+extend-select = ["I"]
+
+[isort]
+combine-as-imports = true
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,1 +1,10 @@
 extend-select = ["I"]
+
+[isort]
+# Mark sqlfluff, test and it's plugins as known first party
+known-first-party = [
+    "sqlfluff",
+    "sqlfluff_plugin_example",
+    "sqlfluff_templater_dbt",
+    "test",
+]
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,4 +1,1 @@
 extend-select = ["I"]
-
-[isort]
-combine-as-imports = true
--- a/plugins/sqlfluff-templater-dbt/test/linter_test.py
+++ b/plugins/sqlfluff-templater-dbt/test/linter_test.py
@@ -7,8 +7,8 @@ import sys
 
 import pytest
 
-from sqlfluff.core import Linter, FluffConfig
 from sqlfluff.cli.commands import lint
+from sqlfluff.core import FluffConfig, Linter
 from sqlfluff.utils.testing.cli import invoke_assert_code
 from test.fixtures.dbt.templater import DBT_FLUFF_CONFIG, project_dir  # noqa: F401
 
--- a/plugins/sqlfluff-templater-dbt/test/rules_test.py
+++ b/plugins/sqlfluff-templater-dbt/test/rules_test.py
@@ -1,17 +1,17 @@
 """Tests for the standard set of rules."""
-import pytest
 import os
 import os.path
 from pathlib import Path
 
+import pytest
+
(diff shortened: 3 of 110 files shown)


zhongjiajie:
PTAL if you have time @alanmcruickshank

zhongjiajie:
update and fix the conflict

plugins/sqlfluff-templater-dbt/test/rules_test.py
```diff
@@ -1,19 +1,18 @@
...
-import pytest
 import os
 import os.path
 from pathlib import Path
-
-from sqlfluff.core import Linter
-from sqlfluff.core.config import FluffConfig
-from sqlfluff.utils.testing.rules import assert_rule_raises_violations_in_file
-
 from test.fixtures.dbt.templater import (  # noqa
```
alanmcruickshank, line 5: `test` here is a first party library, but only in the test section of the codebase. There might need to be a nested config file to mark this as first party so that it's not grouped up here.

plugins/sqlfluff-templater-dbt/test/templater_test.py
```diff
@@ -1,27 +1,25 @@
...
+from test.fixtures.dbt.templater import (  # noqa: F401
+    DBT_FLUFF_CONFIG,
+    dbt_templater,
+    project_dir,
+)
 from unittest import mock
 
 import pytest
-
+from sqlfluff.cli.commands import lint
```
alanmcruickshank, line 18: `ruff` is grouping `pytest` and `sqlfluff` together here, which I think is also because first part library detection is missing `sqlfluff`.

.ruff.toml
```diff
@@ -0,0 +1,4 @@
+extend-select = ["I"]
+
+[isort]
+combine-as-imports = true
```
alanmcruickshank, line 4: I'd rather not set this value. We didn't set this config value before and I don't see a good reason to deviate from the default `ruff` settings.

tox.ini
```diff
@@ -77,9 +77,6 @@ commands = python {toxinidir}/test/generate_parse_fixture_yml.py {posargs}
 commands =
     flake8
     ruff check .
```
alanmcruickshank, line 79:
```suggestion
    # ruff is configured to check linting and imports.
    # see: .ruff.toml and https://docs.astral.sh/ruff/rules/
    ruff check .
```

plugins/sqlfluff-templater-dbt/test/linter_test.py
```diff
@@ -4,13 +4,12 @@
 import os.path
 import shutil
 import sys
+from test.fixtures.dbt.templater import DBT_FLUFF_CONFIG, project_dir  # noqa: F401
```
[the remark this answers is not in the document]
zhongjiajie: I will try to find some config to config sqlfluff and test as first party libs. But it seems there is no cons for ruff recognize it or not, do you want to place sqlfluff same as build-in library?

zhongjiajie:
> `test` here is a first party library, but only in the test section of the codebase. There might need to be a nested config file to mark this as the first party so that it's not grouped up here.

I have marked `sqlfluff`, `sqlfluff_plugin_example`, `sqlfluff_templater_dbt`, and `test` as the first-party library. I use the command

```sh
grep -E '(from test|import test)' **/*.py
```

and found all `test` are in the test module and created by us, so I added it to the ruff config file instead of adding some nested config.

the reason I add `sqlfluff_plugin_example`, `sqlfluff_templater_dbt` it is because I find them via `grep 'from sqlfluff_'  **/*.py`

alanmcruickshank (approved):
Nice. Thanks for taking the time to configure this. Looks great!

PR #5292, opened 2023-10-09
Resolve some strange whitespace indentation behaviour

### LINKED ISSUE #5277: Issue when formatting Jinja's control whitespace (-)
### Search before asking

### What Happened

Hi 👋
(cc @alanmcruickshank )
I am having some issue when using both Jinja's whitespace control's "minus" (-) and {{ this }}.
When fixing the code, it seems to remove part of the content located arround {{ this }} leading to an error.
(When trying without the "minus" whitespace control, the output has however no issue)

```
WITH dummy_cte AS (
    SELECT MIN(d_date) AS min_date
    FROM {{ ref('dummy_table') }}
    WHERE TRUE
    {% if is_incremental() -%}
        AND dt_datetime > (SELECT MAX(dt_source_datetime) FROM {{ this }})
    {%- endif %}
        AND dt_source_datetime > '{{ model_start_date }}'
),
```

### Expected Behaviour

```
WITH
    dummy_cte AS (
        SELECT MIN(d_date) AS min_date
        FROM {{ ref('dummy_table') }}
        WHERE TRUE
            {% if is_incremental() -%}
                AND dt_datetime
                > (SELECT MAX(dt_source_datetime) FROM {{ this }})
            {%- endif %}
            AND dt_source_datetime > '{{ model_start_date }}'
    ),
```

### Observed Behaviour

```
WITH
    dummy_cte AS (
        SELECT MIN(d_date) AS min_date
        FROM {{ ref('dummy_table') }}
        WHERE TRUE
            {% if is_incremental() -%}
                AND dt_datetime
                > (SELECT MAX(dt_source_datetime) FROM {{
            {%- endif %}
            AND dt_source_datetime > '{{ model_start_date }}'
    ),
```

### How to reproduce

We could reproduce this bug by using Jinja's control whitespace in either an if condition either a for loop. In this specific case we also use {{ this }} within the clause and this part was partly erased after lint/fixing the code.

### Dialect

dbt / redshift

### Version

sqlfluff, version 2.3.2

### Configuration

```
# Enable working with dbt built-in Jinja functions and macro (ref, source etc..)
[sqlfluff:templater:jinja]
apply_dbt_builtins = True

[sqlfluff:templater:dbt]
project_dir = 
profiles_dir = dbt/

[sqlfluff]
templater = dbt
dialect = redshift
# Reference to custom macro
load_macros_from_path = utils/dbt_macros
runaway_limit = 10
# Do not ignore inline overrides (ignore rules)
disable_noqa = False
# Ignoring errors 
# AM04 : mentionning an unknown number of columns (mostly triggered in jinja for loops)
# ST07 : forcing the use of ON only in JOIN clauses (not accepting USING)
# AM06 : inconsistent column references in GROUP BY/ ORDER BY clauses
exclude_rules = AM04, ST07, AM06
# Max line length is set by default to 80 chars
max_line_length = 80

# Setting those limitations to 0 (unlimited)
# since the default value for skipping linting is set as 20k bytes per file
large_file_skip_char_limit = 0

[sqlfluff:indentation]
# See https://docs.sqlfluff.com/en/stable/layout.html#configuring-indent-locations
# We choose to have 4 spaces indents
indent_unit = space
tab_space_size = 4
# We want our 'ON'/'AND' subsections to be aligned but indented after 'JOIN'
allow_implicit_indents = True
indented_on_contents = False
# We want our CTEs to be indented below the WITH clause
indented_ctes = True

[sqlfluff:layout:type:comma]
# Allowing trailing spaces after commas
spacing_before = touch
# Allowing commas to be set at end-of-line
line_position = trailing

# Forcing aliasing using AS on tables
aliasing = explicit

# Forcing aliasing using AS on columns
aliasing = explicit

# Forcing all keywords / functions and other items
# to be written UPPERCASE
capitalisation_policy = upper
capitalisation_policy = upper
capitalisation_policy = upper

[sqlfluff:layout:type:casting_operator]
spacing_before = single
spacing_after = single

[sqlfluff:rules:layout.set_operators]
# Return to line when having a UNION or UNION ALL
set_operator_on_new_line = ['UNION', 'UNION ALL']

[sqlfluff:layout:type:start_square_bracket]
# The first character after a bracket should stick to the bracket
spacing_after = touch

# The last character before the end of bracket should stick to the bracket
spacing_before = touch

[sqlfluff:layout:type:common_table_expression]
# Make a query within a CTE fit in 1 line when possible
spacing_within = single:inline

[sqlfluff:rules:convention.casting_style]
# Cast all convert&cast to ::
preferred_type_casting_style = shorthand

[sqlfluff:rules:jinja.padding]
# Add a space between { and } decorators
single_space = true

[sqlfluff:rules:structure.join_condition_order]
# Use the first table used in the JOIN clauses
# within ON condition
preferred_first_table_in_join_clause = earlier

[sqlfluff:layout:type:groupby_clause]
line_position = alone

[sqlfluff:layout:type:orderby_clause]
# Order by clauses appear in many places other than in a select
# clause. To avoid unexpected behaviour we use `leading` in this
# case rather than `alone`.
#spacing_after = touch
line_position = leading

[sqlfluff:layout:type:having_clause]
line_position = alone

[sqlfluff:layout:type:limit_clause]
line_position = alone
```

### Code of Conduct

alanmcruickshank:
This resolves #5277. Fixing this bug actually highlighted some bigger issues in the indentation codebase. This resolves both the surface level bug and also the deeper issue with slicing and deduping of source fixes. The root cause of the linked issue was applying overlapping source fixes, which I haven't _prevented_ but I've added a warning in for to catch in future.


src/sqlfluff/core/errors.py
```diff
@@ -285,11 +285,23 @@ def source_signature(self) -> Tuple[Any, ...]:
         fix_raws = tuple(
             tuple(e.raw for e in f.edit) if f.edit else None for f in self.fixes
         )
-        source_fixes = tuple(
-            tuple(tuple(e.source_fixes) for e in f.edit) if f.edit else None
-            for f in self.fixes
-        )
-        return (self.check_tuple(), self.description, fix_raws, source_fixes)
+        _source_fixes = []
```
greg-finley, line 288: Slightly prefer explicitly typing lists

greg-finley (approved):

src/sqlfluff/core/errors.py
alanmcruickshank: Good idea

PR #5302, opened 2023-10-10
Extend ruff checking to docstring rules

alanmcruickshank:
Following on from #5289, this adds the docstring rules to ruff, copying the same configuration as currently exists for `flake8`.


zhongjiajie (approved):
LGTM and it catch some things we do not get before

greg-finley (approved):
Nice! Could/should we have flake8 not do this check now?

alanmcruickshank:
> Nice! Could/should we have flake8 not do this check now?

Yeah good idea. I'll strip some of that out

alanmcruickshank:
> > Nice! Could/should we have flake8 not do this check now?
> 
> Yeah good idea. I'll strip some of that out

Actually... I wont remove it yet, but I'll definitely get `ruff` to run first in the linting step.

PR #5304, opened 2023-10-10
Simplify `_process_lint_result`

alanmcruickshank:
This is pure refactoring, no functionality changes.

This method was a little convoluted. It's the main performance bottleneck for reflow linting, for reasons which I'm only just getting into - but before I do much restructuring I thought I should merge the simplified version.

### DIFF
--- a/src/sqlfluff/core/rules/base.py
+++ b/src/sqlfluff/core/rules/base.py
@@ -962,47 +962,41 @@ class BaseRule(metaclass=RuleMetaclass):
         if not self.template_safe_fixes:
             self.discard_unsafe_fixes(res, templated_file)
         lerr = res.to_linting_error(rule=self)
-        ignored = False
-        if lerr:
-            # Check whether this should be filtered out for being unparsable.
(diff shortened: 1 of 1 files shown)


src/sqlfluff/core/rules/base.py
```diff
@@ -962,47 +962,41 @@ def _process_lint_result(
...
-                    )
-                    lerr = None
-                    ignored = True
-                    break
+        if not lerr:
+            return None
+        if ignore_mask:
+            filtered = ignore_mask.ignore_masked_violations([lerr])
+            if not filtered:
+                return None
```
greg-finley, line 970: slightly faster to not assign to a variable? could put in a comment instead if we need to note "filtered"

```suggestion
            if not ignore_mask.ignore_masked_violations([lerr]):
                return None
```

greg-finley (approved):

src/sqlfluff/core/rules/base.py
alanmcruickshank: Yeah much better 👍

PR #5320, opened 2023-10-13
Cache the `BaseSegment` hash in reflow

alanmcruickshank:
This is a very small PR, it's a performance improvement for reflow by streamlining the creating of the hash function of segments. Small change - but it's one of the bottlenecks of LT02 so should have a performance uplift.


greg-finley (approved):

tunetheweb (approved):
LGTM

PR #5332, opened 2023-10-16
UUID Comparisons

alanmcruickshank:
For reasons which are a little obscure it appears that `UUID` comparison is a little slow. By storing the `int` representation of the uuid rather than the `UUID` class itself we can speed up segment comparisons. I think this is also a slight simplification too, but it's mostly performance optimisation.


alanmcruickshank:
> Is it faster even considering the time to convert to int?

The `int` is already there on creation. We're just extracting it.

PR #5333, opened 2023-10-16
Two depth map performance improvements

alanmcruickshank:
Two very small performance improvements in `DepthMap`:
1. Rather than using `min()` and `max()` on the `code_idxs` we can rely on the tuple always being sorted so just taking the first and last values instead.
2. We're calculating the `hash()` of a segment twice in this code - let's just calculate it once and then reuse it.

### DIFF
--- a/src/sqlfluff/utils/reflow/depthmap.py
+++ b/src/sqlfluff/utils/reflow/depthmap.py
@@ -28,10 +28,12 @@ class StackPosition:
         # If there's only one code element, this must be it.
         elif len(path_step.code_idxs) == 1:
             return "solo"
-        # Check for whether first or last code element
-        elif path_step.idx == min(path_step.code_idxs):
+        # Check for whether first or last code element.
+        # NOTE: code_idxs is always sorted because of how it's constructed.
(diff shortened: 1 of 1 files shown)


src/sqlfluff/utils/reflow/depthmap.py
```diff
@@ -28,10 +28,12 @@ def _stack_pos_interpreter(path_step: PathStep) -> str:
         # If there's only one code element, this must be it.
         elif len(path_step.code_idxs) == 1:
             return "solo"
-        # Check for whether first or last code element
-        elif path_step.idx == min(path_step.code_idxs):
+        # Check for whether first or last code element.
+        # NOTE: code_idxs is always sorted because of how it's constructed.
+        # That means the lowest is always as the start and the highest at the end.
+        elif path_step.idx == path_step.code_idxs[0]:
```
greg-finley, line 34: Could this sortedness ever be broken in future refactoring? I'm not sure how to enforce this assumption forever.

greg-finley (approved):

PR #5334, opened 2023-10-16
`Set` to `FrozenSet` in segment class_types

alanmcruickshank:
This was triggered by the overhead of converting a `set` to `frozenset` in `depthmap.py`, but really the `class_types` attribute of a segment should already be a `frozenset`. The reasons we hadn't done that already were that I couldn't work out an elegant way of doing it. This makes the change so that we're always dealing with `frozenset` objects right from the start. This should have a double whammy performance improvement, first on segment creation, and then later on `DepthMap` construction.


greg-finley (approved):
Nice, never heard of this data structure before!

PR #5346, opened 2023-10-24
Add placeholder templating for single brackets

michal-latkowski-avaus:


michal-latkowski-avaus:
Sorry, mistakenly made a PR to the main repository :(

PR #5348, opened 2023-10-24
Unpick dependencies between modules in `sqlfluff.core`

alanmcruickshank:
This PR unpicks a few counter-intuitive dependency pathways in the core library and then enforces their segregation in the linting rules. Specifically:

1. `helpers` now has it's own module containing `string`, `slice` and `dict` helpers. These either came from the previous `string_helpers` or `slice_helpers`, but I've also added in a few methods from `config` which didn't have any dependencies. The `helpers` module has no dependencies on any other SQLFluff module.
2. The `noqa` module, has been moved from `linter` into `rules`, which allows us to enforce the order of dependencies, but also on reflection feels more intuitive.
3. The methods for applying fixes have been moved out of `segments` (and therefore are no longer a method on `BaseSegment`) and into a new module within `rules` called `fix`. This logically makes a lot more sense and further streamlines the `BaseSegment` class which has become so bloated. It's worth noting that the new `fix.py` file is still fairly hefty, but it's still shorter than the two files which most of the logic has been extracted from (mostly either `segments.base` or `rules.base`).

No functionality changes here, just moving (and some renaming) of code.


WittierDinosaur (approved):
Looks good - high confidence as tests aren't failing. Nice work, the dependency chain have become so bloated.

PR #5349, opened 2023-10-24
Add a public API for nesting config strings.

### LINKED ISSUE #5207: Public API to build nested configs
### Search before asking

### Description

I want to have a public API to be able to build `FluffConfig` instance from several config files which override one another, replicating the behaviour with nested `.sqlfluff` configs in the filesystem.

### Use case

I'm building a lint functionality which uses SQLFluff. I want the user to be able to provide nested `.sqlfluff` config files in the service's virtual filesystem, so I need to build the config in my own code.

I have tried using `nested_combine` like this:

```Python
class LintConfigFile:
    content: str

def build_config(config_files: Optional[List[LintConfigFile]]) -> FluffConfig:
    if not config_files:
        return None

    loader = ConfigLoader.get_global()
    config_dicts = [
        loader.load_config_string(config_file.content)
        for config_file in config_files
    ]
    config = nested_combine(*config_dicts)
    return FluffConfig.from_kwargs(config=config)
```

However, this fails to combine contents of two different config files with this simple unit test:

```
        def test_build_config_multiple(self):
            config = build_config(
                [
                    LintConfigFile(
                        content="""
    [sqlfluff]
    dialect = duckdb
    """
                    ),
                    LintConfigFile(
                        content="""
    [sqlfluff]
    max_line_length = 120
    """
                    ),
                ]
            )
>           assert config.get("dialect") == "duckdb"
E           AssertionError: assert None == 'duckdb'
```

So I ended up using this:

```Python
def build_config(
    config_files: Optional[List[LintConfigFile]],
) -> Optional[FluffConfig]:
    if not config_files:
        return None

    # We build a temporary directory with the config files to replicate sqlfluff
    # nesting config behavior
    with tempfile.TemporaryDirectory(dir=TMP_DIR) as temp_dir_path:
        cur_path = temp_dir_path
        for i, config_file in enumerate(config_files):
            cur_path = cur_path + f"/config-{i}"
            os.makedirs(cur_path)
            config_path = f"{cur_path}/.sqlfluff"
            with open(config_path, "w") as f:
                f.write(config_file.content)
        return FluffConfig.from_path(cur_path)
```

Which behaves exactly as I expect, but looks to be very unoptimal.

### Dialect

not related to particular dialect

### Code of Conduct

alanmcruickshank:
This PR resolves #5207. This mostly uses existing machinery to create nested configs, but exposes it in an easier way and includes an example in the API examples folder. Eventually we might want to deprecate `from_string` and only allow `from_strings`, but that turns this into a more breaking release than it needs to be.

### DIFF
--- a/src/sqlfluff/core/config.py
+++ b/src/sqlfluff/core/config.py
@@ -1074,7 +1074,7 @@ class FluffConfig:
         will take precedence over any earlier values.
         """
         loader = ConfigLoader.get_global()
-        config_state = {}
+        config_state: Dict[str, Any] = {}
         for config_string in config_strings:
             config_state = loader.load_config_string(
                 config_string, configs=config_state
--- a/src/sqlfluff/core/config.py
+++ b/src/sqlfluff/core/config.py
@@ -1076,7 +1076,9 @@ class FluffConfig:
         loader = ConfigLoader.get_global()
         config_state = {}
         for config_string in config_strings:
-            config_state = loader.load_config_string(config_string, configs=config_state)
+            config_state = loader.load_config_string(
+                config_string, configs=config_state
+            )
         return cls(
             configs=config_state,
             extra_config_path=extra_config_path,
--- a/src/sqlfluff/core/config.py
+++ b/src/sqlfluff/core/config.py
@@ -1047,11 +1047,38 @@ class FluffConfig:
         overrides: Optional[Dict[str, Any]] = None,
         plugin_manager: Optional[pluggy.PluginManager] = None,
     ) -> "FluffConfig":
-        """Loads a config object given a particular path."""
+        """Loads a config object from a single config string."""
         loader = ConfigLoader.get_global()
         c = loader.load_config_string(config_string)
         return cls(
             configs=c,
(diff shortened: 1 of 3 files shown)


src/sqlfluff/core/config.py
```diff
@@ -1058,6 +1058,35 @@ def from_string(
...
 
+    @classmethod
+    def from_strings(
+        cls,
+        *config_strings: str,
+        extra_config_path: Optional[str] = None,
+        ignore_local_config: bool = False,
+        overrides: Optional[Dict[str, Any]] = None,
+        plugin_manager: Optional[pluggy.PluginManager] = None,
+    ) -> "FluffConfig":
```
greg-finley, line 1069: I wonder if all of these could return `FluffConfig` not `"FluffConfig"`. Maybe with `from __future__ import annotations`?

greg-finley (approved):

src/sqlfluff/core/config.py
alanmcruickshank: ohh good idea 👍

Issue #5325, opened 2023-10-14
Unable to format dbt snowflake sql file


coleheflin-gt:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

I was attempting to run the fix command on a snowflake dbt sql file and was receiving the following error:

```
> WARNING    One fix for LT01 not applied, it would re-cause the same error.
> WARNING    One fix for LT01 not applied, it would re-cause the same error.
> WARNING    One fix for LT01 not applied, it would re-cause the same error.
> WARNING    One fix for LT01 not applied, it would re-cause the same error.
> WARNING    One fix for LT01 not applied, it would re-cause the same error.
> WARNING    One fix for LT01 not applied, it would re-cause the same error.
> WARNING    Loop limit on fixes reached [10].
> ==== no fixable linting violations found ====
```

### Expected Behaviour

I was expecting to have the file get formatted properly but instead it said it found 26 unfixable linting violations.

### Observed Behaviour

After I ran the command, `sqlfluff fix test.sql`, I received the below error:

```
==== finding fixable violations ====
=== [dbt templater] Sorting Nodes...
=== [dbt templater] Compiling dbt project...
=== [dbt templater] Project Compiled.
WARNING    One fix for LT01 not applied, it would re-cause the same error.
WARNING    One fix for LT01 not applied, it would re-cause the same error.
WARNING    One fix for LT01 not applied, it would re-cause the same error.
WARNING    One fix for LT01 not applied, it would re-cause the same error.
WARNING    One fix for LT01 not applied, it would re-cause the same error.
WARNING    One fix for LT01 not applied, it would re-cause the same error.
WARNING    Loop limit on fixes reached [10].
==== no fixable linting violations found ====
All Finished 📜 🎉!
  [26 unfixable linting violations found]
```

### How to reproduce

Here is the SQL file:
```
{{
    config(
        materialized='incremental',
        unique_key="id",
        schema="public",
    )
}}

WITH monthly_purchases AS (
    SELECT 
        YEAR("DATE") AS YEAR
        , MONTH("DATE") AS MONTH
        , PLATFORM
        , ID
        , EXAMPLE_ID
        , IS_MOBILE
        , SUM(CASE WHEN EVENT_NAME = 'purchase' THEN 1 ELSE 0 END) AS NUM_PURCHASES
[...]

PR #5351, opened 2023-10-24
Resolve SparkSQL re-parsing issue + test validation in test suite.

alanmcruickshank:
This resolves part of #5325. There are two queries in that post, this resolves the issue with the latter one. The first one I couldn't recreate the issue.

In short - reparsing for sparksql was failing because a `KeywordSegment` rather than `SymbolSegment` was being used.

To try and catch situations like this I've added a validation check to the parse suite so that we know that any un-edited files automatically pass the "re-parse" check in advance.

### DIFF
--- a/src/sqlfluff/dialects/dialect_sparksql.py
+++ b/src/sqlfluff/dialects/dialect_sparksql.py
@@ -516,8 +516,8 @@ sparksql_dialect.add(
     ),
     # Adding Hint related segments so they are not treated as generic comments
     # https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-hints.html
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_sparksql.py
```diff
@@ -743,14 +743,14 @@
 # https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-hints.html
 sparksql_dialect.insert_lexer_matchers(
     [
-        RegexLexer("start_hint", r"\/\*\+", CodeSegment),
+        StringLexer("start_hint", "/*+", CodeSegment),
```
WittierDinosaur, line 746: Is that a wayward `+` ?

WittierDinosaur (approved):
LGTM

PR #5355, opened 2023-10-25
One (very) small typing improvements

alanmcruickshank:
Two small typing improvements. I noticed both of them while working on some other things.

### DIFF
--- a/src/sqlfluff/core/config.py
+++ b/src/sqlfluff/core/config.py
@@ -195,26 +195,29 @@ REMOVED_CONFIGS = [
 ]
 
 
-def coerce_value(val: str) -> Any:
+def coerce_value(val: str) -> Union[int, float, bool, None, str]:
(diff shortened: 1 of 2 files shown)


src/sqlfluff/core/config.py
```diff
@@ -195,26 +195,29 @@ class _RemovedConfig:
 ]
 
 
-def coerce_value(val: str) -> Any:
+def coerce_value(val: str) -> Union[int, float, bool, None, str]:
```
greg-finley, line 198: Ha I almost prefer `Any` over a Union of 5 types, not sure

greg-finley (approved):

src/sqlfluff/core/config.py
alanmcruickshank: Yeah that's fair - I'll revert this bit. It will result in the smallest PR ever though 🤣

PR #5363, opened 2023-10-30
Allow ignoring of comments from indentation entirely #3311

### LINKED ISSUE #3311: L003: Allow comment lines to be ignored
### Search before asking

### Description

I notice that L003 reports different messages for comments and for code:

- "Comment not aligned with following line." *(Aside: Should really say "preceding line", no?)*
- "Expected X indentations, found Y [compared to line N]"

So presumably, it wouldn't be too difficult to allow it to ignore comments completely (i.e. never report messages of the first kind), perhaps via the existing [`ignore_comment_lines` parameter used by L016](https://docs.sqlfluff.com/en/stable/rules.html#sqlfluff.core.rules.Rule_L016).

See use case below.

### Use case

I frequently write schemas with inline documentation, e.g.

```sql
CREATE TABLE t (
    a INT UNSIGNED NOT NULL PRIMARY KEY AUTO_INCREMENT,
    b VARCHAR(63),
        -- Description of what `b` is for
    c BOOLEAN NOT NULL
        /* Long description of
        what `c` is for */
);
``` 

Currently, L003 complains about the comments as they are indented further than the line that precedes them, but I would like this not to happen, i.e. would like the above to be accepted without disabling L003 in its entirety.

### Dialect

Shouldn't be dialect-specific, but I use `mysql`.

### Code of Conduct

alanmcruickshank:
This resolves #3311 (and oldie).

I've kept the same config variable name as used before and used in the long line rule config, but this time in the `indentation` config. If set, any comment-only lines will be totally ignored from reindentation.

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -75,6 +75,8 @@ skip_indentation_in = script_content
 # that a comment precedes the line it describes. However if you prefer
 # comments moved _after_, this configuration setting can be set to "after".
 trailing_comments = before
+# To exclude comment lines from indentation entirely set this to "True".
+ignore_comment_lines = False
 
 # Layout configuration
 # See https://docs.sqlfluff.com/en/stable/layout.html#configuring-layout-and-spacing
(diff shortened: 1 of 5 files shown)


greg-finley (approved):
👍

PR #5364, opened 2023-10-30
Build out rule and fix serialisation

### LINKED ISSUE #4238: Warnings are not output in GitHub annotations
### Search before asking

### What Happened

Running `sqlfluff lint` with the `--format github-annotation` option removes warnings from the output. 

### Expected Behaviour

I would expect the results to be annotated variously as warnings and/or errors, in line with the standard human-readable output.

### Observed Behaviour

Interestingly, the `--annotation-level` option seems to change the output level for all errors (e.g. `--annotation-level warning` sets actual errors to warnings) and still silently drops all actual warnings. 

### How to reproduce

`sqlfluff lint --format github-annotation file-with-errors-and-warnings.sql`

`.sqlfluff` with some rules downgraded to warnings e.g. 
```
[sqlfluff]
dialect = postgres
exclude_rules = L009, L030, L031, L032, L051, L055
warnings = L001, L005, L006, L008, L010, L016, L023, L050
```

### Dialect

postres

### Version

sqlfluff 1.4.5
python 3.8

### Configuration

```
[sqlfluff]
dialect = postgres
exclude_rules = L009, L030, L031, L032, L051, L055
warnings = L001, L005, L006, L008, L010, L016, L023, L050

[sqlfluff:rules]
tab_space_size = 4
indent_unit = space
max_line_length = 100
aliasing = explicit
```

### Code of Conduct

### LINKED ISSUE #4651: Report the end/stop location / width of violations so that violations can be accurately highlighted in a text editor
### Search before asking

### Description

Currently, SQLFluff only supports showing the starting location of a violation.  This means that anyone using the output of linting is unable to know the length / end location of the violation.

This is a limitation that is workable for a CLI tool, although even there it places limitations on what could potentially be rendered on the output.  However, it becomes really bad looking for IDE plugins.  I will use the plugin from Anbora labs for IntelliJ as an example: https://github.com/anboralabs/sqlfluff-intellij-community

Because the plugin doesn't know the end location of the violation, it has to make some assumption.  In this case, it assumes it's just one character wide.  This results in really goofy looking inspections like this in my IDE text editor (note the rule is a custom one I am making):

Screenshot 2023-03-31 at 9 41 56 PM

Notice that only the `t` is yellow highlighted, because the plugin doesn't know how long the violation was.  Ideally, it would yellow highlight the entire `test_column_three` token in the IDE, since that is the portion of the statement my rule considers to be in error.  That is how the builtin IDE inspections from JetBrains work.

Other linters do provide this information.  For example, eslint provides a json output formatter that includes end positions:  https://eslint.org/docs/latest/use/formatters/#json

Example eslint output from the documentation, which includes `endLine` and `endColumn`:
```json
[
  {
    "filePath": "/var/lib/jenkins/workspace/Releases/eslint Release/eslint/fullOfProblems.js",
    "messages": [
      {
        "ruleId": "no-unused-vars",
        "severity": 2,
        "message": "'addOne' is defined but never used.",
        "line": 1,
        "column": 10,
        "nodeType": "Identifier",
        "messageId": "unusedVar",
        "endLine": 1,                    <------ note the end location included here!
        "endColumn": 16
      },
```

This is probably a major effort since all the rules would need to be updated to support this.  Perhaps it could be introduced in a backwards-compatible way, where SQLFluff itself just assumes the violation is 1 character wide (or alternatively the entire starting anchor) unless the rule has been updated to provide an end location.

I am only just now learning how to write my own rule, and have never done something similar with other linters.  So my perspective is limited and perhaps a different design would be better.  But to try to give some initial thought to the issue:  intuitively I think what would make sense is if the rule can provide an optional `end_anchor` in the `LintResult`, in addition to the `anchor` already given.  When provided, it would be the inclusive location of where the end of the violation was.  Possibly one sensible default value for rules that don't yet provide one might be to assume a default of `end_anchor` equal to the existing starting `anchor`.  It might even be the right thing for some rules: for the initial rules I wrote it coincidentally would work...

An open question might be which child nodes, if any, are also included when the `end_anchor` is a parent node.  I kind of think the answer might be "all of them" so that I don't need to traverse the tree to find the final leaf node if I just want to highlight the entire tree branch.  For example, in my case, I just want to highlight the entire `column_reference` and have no interest otherwise in traversing it further.  If someone wanted to only highlight up to a specific child node then they could just return that specific child node.  Alternatively, you could be requiring users to provide the final leaf node in the tree and do that traversal themselves.

Another question might be what if the user wants to only highlight up to a specific location within the raw text of an individual tree node.  However, already the starting anchor does not support returning a specific start location within the starting anchor, so I think it's orthogonal to this issue.

### Use case

Ultimately I would like SQLFluff to provide a way for text editor plugins to accurately highlight the entire violation, from starting location to ending location.

I think the best way to do that will be if the JSON output of SQLFluff can be updated to include end location information, similar to ESLint.

### Dialect

General improvements to SQLFluff rule API

### Code of Conduct

### LINKED ISSUE #5353: Public API to get individual fixes
### Search before asking

### Description

I want to have a public API to get a list of errors with fixes for them, which could be displayed in an IDE.

I've been writing code that would use internal, undocumented APIs to achieve this, but I've ran into some problems.

I can get a list of errors with fixes like this:

```Python
from sqlfluff.core.linter import Linter

# ...some many lines later...

linter = Linter(config=config)
result = linter.lint_string_wrapped(args.content)
violations = result.paths[0].files[0].violations
for error in violations:
    if isinstance(error, SQLLintError):
        fixes: List[SQLFluffLintFix] = error.fixes
        # TODO Convert fixes
    else:
        logger.trace("Not SQLLintError")
```

And I was trying to come up with a function that would convert a LintFix instance to a serialisable representation, which could then be used in a code editor:

```Python
from sqlfluff.core.rules.base import LintFix as SQLFluffLintFix


# These data classes map to Monaco or other potential editor's types

class Range:
    start_line_number: int
    start_column: int
    end_line_number: int
    end_column: int


class LintFixEdit:
    range: Range
    text: str


class LintFix:
    title: str
    edits: List[LintFixEdit]


def sqlfluff_to_editor_lint_fix(
    sqlfluff_lint_fix: SQLFluffLintFix,
) -> Optional[LintFix]:
    popsql_edits: List[LintFixEdit] = []

    # PROBLEM
    # Position marker doesn't always exist
    anchor_pos = sqlfluff_lint_fix.anchor.pos_marker
    if not anchor_pos:
        raise ValueError(
            f"Anchor position not found for fix: {sqlfluff_lint_fix}"
        )

    range_start_line = anchor_pos.line_no
    range_start_column = anchor_pos.line_pos

    edit_text = ""

    if sqlfluff_lint_fix.edit_type == "delete":
        # Nothing to do, text stays empty and range defines the segment to delete
        pass
    elif sqlfluff_lint_fix.edit_type in [
        "create_before",
        "create_after",
        "replace",
    ]:
        if not sqlfluff_lint_fix.edit:
            raise ValueError(
                f"Edit segments are missing for a non-delete operation: {sqlfluff_lint_fix}"
            )

        for edit_segment in sqlfluff_lint_fix.edit:
            edit_text += edit_segment.raw

            # PROBLEM
            # How to get a range end position?

```

The core of the problem with this approach is that `BaseSegment` class is not actually a position in source code, but some kind of very representation. This makes sense for a parser, but I didn't understand how to convert `BaseSegment` to a pair of start and end line and column numbers.

### Use case

I want to display sqlfluff fixes as "quick fixes" in a code editor, and be able to instantly apply those edits when user selects one (without running sqlfluff again).

Screenshot 2023-10-25 at 12 47 27

(In my particular case, I will use Monaco editor).

### Dialect

Not specific to any dialect

### Code of Conduct

alanmcruickshank:
This resolves #5353 and resolves #4651. It also resolves #4238.

This PR significantly increases the amount of detail returned by the serialisation routines for violations and fixes. This means a lot more info is included in the non human output versions of `sqlfluff lint` (including both start _and end_ points for errors, and the initial fixes proposed). As a result, tests cases which specify the details of a violation are now _also_ required to include that detail - so most of the increased lines of code from this PR are from increased detail in test cases (which I think leads to better test coverage in essence).

Specfically it renames `get_info_dict()` to just `to_dict()` and adds a parallel `to_dict()` method on the `LintFix` class so that proposed fixes are shown in the output when linting issues are serialised. Warnings also come through now too - always with the check type `notice`, so that the `--annotation-type` cli option only applies to violations. Given that warnings come through at that level, I've changed the default level to `warning`.

Based on feedback from @golergka and @james-johnston-thumbtack, I've kept the format fairly flat rather than nesting starts and ends into some kind of combined structure.


golergka:
Thank you so much for your work on this! I've only read the changes, haven't tried running this code yet.

One thing that seems a little bit confusing for me is a different formatting for positions. For lints, we get `line_no` and  `line_pos` properties (which is great for my particular use case, because they translate nicely into Monaco's line and column numbers), but for fixes it's just `source_position` with a tuple of two ints.

From your code for `delete` edit type, it seems that this source position is supposed to be all the information about the edit — so those two numbers must encode both start and end of the segment, which means it's just a Nth character in the whole input.

Would it be possible to get `start_line_no`, `start_line_pos`, `end_line_no` and `end_line_pos` for the fixes instead? IMO, it might be a better API design, because it would use the same semantics which is already used for lints themselves.

james-johnston-thumbtack:
I assume you're referring to the `source_position` in `simple_test.py`?  And these are (byte? character?) offsets into the input string?  In which case yeah it makes sense.

Here is where the plugin of @dalgarins comes up with a range of text for IntelliJ to highlight.... in this case it appears that IntelliJ needs a starting character position and a length: https://github.com/anboralabs/sqlfluff-intellij-community/blob/08354f69b81f3e5ce83e5cdf6245d10a47bc210b/src/main/kotlin/co/anbora/labs/sqlfluff/lint/Linter.kt#L111 --- the existing code is assuming length "0" (since it doesn't know it), and a starting position calculated by using the getLineStartOffset function (https://github.com/anboralabs/sqlfluff-intellij-community/blob/08354f69b81f3e5ce83e5cdf6245d10a47bc210b/src/main/kotlin/co/anbora/labs/sqlfluff/lint/Linter.kt#L105 ) since SQLFluff only provides the line number.

So it appears that maybe it's six one way and half a dozen the other in terms of whether it's most convenient to provide start/end line numbers & positions, as @golergka proposes, or overall offset values as your current commit does.  Certainly, if we were writing a console app to write these offsets, it's easy to directly print the offsets as-is if the API is providing line numbers like @golergka says.  At any rate, it appears that the IntelliJ SQLFluff plugin has a way to convert between the two as needed.

dalgarins:
I agree with @golergka to show this values `start_line_no`, `start_line_pos`, `end_line_no` and `end_line_pos`, will be helpful for other tools not only my plugin.

alanmcruickshank:
> I agree with @golergka to show this values `start_line_no`, `start_line_pos`, `end_line_no` and `end_line_pos`, will be helpful for other tools not only my plugin.

Perhaps a trivial question, but one I'd love to be fairly consistent on: Would you expect these to be four "flat" values, or would you expect perhaps a more "nested" structure? We could do something like:

**Option 1: Flat**
```json
{
    "rule": "ABC123",
    "other": "stuff...",
    "start_line_no": 1,
    "start_line_pos": 4,
    "end_line_no": 3,
    "end_line_pos": 7
}
```

**Option 2: Nested by start/end**
```json
{
    "rule": "ABC123",
    "other": "stuff...",
    "start": {
        "line_no": 1,
        "line_pos": 4,
        "file_pos": 14
    },
    "end": {
        "line_no": 3,
        "line_pos": 7,
        "file_pos": 23
    }
}
```

I'm erring toward the latter because we can include multiple options for how people want to locate things - but I realise it's both a more complicated structure, and also a bigger change from past behaviour. Doubly so in the fixes because it makes the structure much denser.

dalgarins:
@alanmcruickshank with flat version is ok.

golergka:
Second flat version just because it's how dicts for the errors already work

greg-finley (approved):
Very nice!

alanmcruickshank:
@golergka @james-johnston-thumbtack @dalgarins do you have any additional feedback on this version before I get it merged? When it's in, I'll push out another pre-release for people to beta-test.

src/sqlfluff/core/errors.py
```diff
@@ -83,17 +95,18 @@ def desc(self) -> str:
 
         return self.__class__.__name__  # pragma: no cover
 
-    def get_info_dict(self) -> Dict[str, Union[str, int]]:
+    def to_dict(self) -> Dict[str, Any]:
```
golergka, line 98:
```suggestion
    def to_dict(self) -> Dict[str, Union[str, int, bool]]:
```

golergka (approved):
Thanks so much for your work on this. Reading this code, I just got a couple of small comments, but most importantly, the API that you expose for fixes and end positions is great.

src/sqlfluff/core/errors.py
alanmcruickshank: For the dependent classes the return value is a bit trickier, but your intent here to narrow the typing is probably sensible. I can do something which achieves that but still works for the other classes.

src/sqlfluff/cli/commands.py
```diff
@@ -662,12 +664,23 @@ def lint(
...
+                        "end_line": violation.get(
+                            "end_line_no", violation["start_line_no"]
+                        ),
+                        "end_column": violation.get(
+                            "end_line_pos", violation["start_line_pos"]
+                        ),
                         "title": "SQLFluff",
                         "message": f"{violation['code']}: {violation['description']}",
-                        "annotation_level": annotation_level,
+                        # Warnings are always notices.
```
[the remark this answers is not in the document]
alanmcruickshank: What I'm trying to say is that SQLFluff rules which are downgraded to warnings using the `warnings` config value will always be reported using the `notice` annotation level in github. I'll make the comment more specific.

dalgarins:
Thank you, for your work, this PR will allow a better integration with other tools, I don't have any other comment.

PR #5365, opened 2023-10-30
Another extraction of fixing logic.

alanmcruickshank:
This is another step on from #5348.

The original PR extracted a lot of logic from `BaseSegment` and put in in `core.rules.fix`. This pulls some of that logic up _again_ into `core.linter.fix` and `core.linter.patch`, where it was more aligned with the _application_ of fixes, and not the generation of them. Additionally, some very similar logic was embedded into `core.linter.linted_file`, which I've also extracted into `core.linter.patch`.

This allowed a little bit of of renaming to make it clearer which methods are public and private, but this is almost entirely just code _moving_. A few places, things that were object methods are now just methods (i.e. not attached to a class).

This should leave us in a much cleaner place, with similar logic living together and less intertwined dependencies.


src/sqlfluff/core/linter/fix.py
```diff
@@ -0,0 +1,308 @@
...
+                requires_validate = True
+                # NOTE: We don't add the segment in this case.
+                continue
+
+            # Otherwise it must be a replace or a create.
+            assert f.edit_type in (
+                "replace",
+                "create_before",
+                "create_after",
+            ), f"Unexpected edit_type: {f.edit_type!r} in {f!r}"
```
greg-finley, line 182: Wherever it's created, could be created as union of literals instead of a plain string, which would avoid a runtime check?

src/sqlfluff/core/linter/fix.py
```diff
@@ -0,0 +1,308 @@
...
+
+            if f.edit_type == "create_after" and len(anchor_info.fixes) == 1:
+                # in the case of a creation after that is not part
+                # of a create_before/create_after pair, also add
+                # this segment before the edit.
+                seg_buffer.append(seg)
+
+            # We're doing a replacement (it could be a single
+            # segment or an iterable)
+            assert f.edit, f"Edit {f.edit_type!r} requires `edit`."
```
greg-finley, line 192: Nice if the type system makes it impossible to get in a bad state

src/sqlfluff/core/linter/fix.py
```diff
@@ -0,0 +1,308 @@
...
+    for seg in seg_queue:
+        s, pre, post, validated = apply_fixes(seg, dialect, rule_code, fixes)
+        # 'before' and 'after' will usually be empty. Only used when
+        # lower-level fixes left 'seg' with non-code (usually
+        # whitespace) segments as the first or last children. This is
+        # generally not allowed (see the can_start_end_non_code field),
+        # and these segments need to be "bubbled up" the tree.
+        seg_buffer.extend(pre)
+        seg_buffer.append(s)
+        seg_buffer.extend(post)
```
greg-finley, line 245: `seg_buffer += pre + [s] + post` ?

greg-finley (approved):

src/sqlfluff/core/linter/fix.py
alanmcruickshank: This one might be tricky 🤔

src/sqlfluff/core/linter/fix.py
alanmcruickshank: This is a _GREAT_ idea. Will see if I can get that to work.

alanmcruickshank: Hmmmm.... actually at this point in the code it's kind of useful as documentation. It's also not on the critical path so I'll keep it as is for now. Neat idea though 👍

PR #5288, opened 2023-10-08
Drop support of python 3.7

zhongjiajie:
### Brief summary of the change made

python 3.7 is EOL at 2023-06-27, which we can see
in https://devguide.python.org/versions/ so we
should drop support of it

BTW, drop support python 3.7 can also reduce code
complex, such as we do not need some package only
support in python 3.7, including `importlib_metadata` and `backports.cached-property`

and we have to use code like

```python
import sys

if sys.version_info >= (3, 8):
    from functools import cached_property
else:  # pragma: no cover
    from backports.cached_property import cached_property

__all__ = ("cached_property",)
```

### Are there any other side effects of this change that we should be aware of?

python 3.7 user can not install sqlfluff

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #5372, opened 2023-11-02
Remove root requirements.txt

alanmcruickshank:
I noticed this as part of #5288 - we don't actually need a `requirements.txt` file any more. All of the details are duplicated in the `setup.cfg` file - and all the places that reference the `/requirements.txt` file, also usually install from the `setup.cfg` file.

That means it's not really necessary and we can remove it! (I think)


greg-finley (approved):
Looking forward to using `poetry` next 😉

PR #5373, opened 2023-11-02
Migrate to `pyproject.toml` for the core project.

alanmcruickshank:
This migrates everything from `setup.cfg` to `pyproject.toml`. In doing that we can consolidate a bunch of the random config files into that same file because config support is much more widespread. In the process I removed some of the `mypy` exceptions for libraries like `dbt` and `pluggy` which now have typing hints.

The only config which I haven't moved are:
- `tox`, because it has strange requirements.
- `pytest`, because while I _can_ move it, I got some strange coverage results when I did.
- `flake8`, because it's in the tox file and doesn't support toml.
- `coverage`, because it's also in the tox file and while it supports toml for later python versions, it looked sketchy.
- `yamllint` config, because that has to be in... `yaml` 🤷 

I've check the `build-dist` and `check-dist` tasks which both work fine. Local editable installing works ok and I've run a `diff -r` of the `.tar.gz` packages produced in both scenarios and the only differences are expected ones.

I think this is good to go 👍 .

Once it's merged I'll kick of a dev release to check that we can still install from pypi etc...


greg-finley (approved):
Good luck!

PR #5316, opened 2023-10-12
Add support for begin atomic functions in Postgres

joaostorrer:
Closes #5310 

### Pull Request checklist
- [X] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

Issue #5310, opened 2023-10-12
Add support for BEGIN ATOMIC


AndLut:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

Currently SQLfluff throws an error when trying to format a statement with `BEGIN ATOMIC`, e.g.

~~~
CREATE OR REPLACE FUNCTION data_wrapper()
RETURNS SETOF data
STABLE PARALLEL SAFE LEAKPROOF
BEGIN ATOMIC
  SELECT *
  FROM data:
END;
~~~

### Use case

I try to format/lint my sql files, but when using `BEGIN ATOMIC` the parser throws the following error:

`Found unparsable section: 'BEGIN ATOMIC`.

It was tested with the command `sqlfluff lint<path-to-sql-file> --dialect postgres`

### Dialect

postgres

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

PR #5376, opened 2023-11-05
postgres: add support for bodies of "language sql" functions

65278:
Postgres's `create function` grammar doesn't only allow to define functions as a string literal, but also as a plain body of SQL, should the function be `language sql`: https://www.postgresql.org/docs/current/sql-createfunction.html

Limited support for this was introduced in: https://github.com/sqlfluff/sqlfluff/issues/5310
#5310 does have some rather serious issues though:
 - Matching a semicolon after `END`  [here](https://github.com/sqlfluff/sqlfluff/pull/5316/files#diff-84df248c7619e658dbf10bea3cad532f9c0d17e06230f791f4cdd419721d2573R1318) means that the lexer doesn't close the function, as it's looking for a semicolon that has already been matched. The generated YAML testcase actually showcases this issue and only accidentally works because the added statement also happens to be the last statement in the file.
   I have fixed this issue.
 - BEGIN ATOMIC {body} END can actually have multiple statements, provided that they all can be executed in an atomic fashion.
 - Technically, these are not `CREATE FUNCTION [...] AS` statements, but their own category. SQLFluff will happily accept functions of `language plpgsql` type with a body of `language sql`. I have not addressed this issue.

Furthermore, I implemented the `CREATE FUNCTION [...] LANGUAGE SQL RETURN expression;` syntax. This has implications discussed below.

Unfortunately, the single statement `RETURN expression` syntax turns `RETURN` into a keyword (because now it is an actual keyword in this context). This may have repercussions I have not considered, but I see little choice in the matter.

### Pull Request checklist

### DIFF
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -1310,12 +1310,27 @@ class FunctionDefinitionGrammar(ansi.FunctionDefinitionGrammar):
                 ),
             ),
             Sequence(
+                "RETURN",
+                Ref("ExpressionSegment"),
+            ),
+            Sequence(
                 "BEGIN",
                 "ATOMIC",
-                Ref("SelectStatementSegment"),
-                Ref("SemicolonSegment"),
+                AnyNumberOf(
+                    Sequence(
+                        Ref("InsertStatementSegment"),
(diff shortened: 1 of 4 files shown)


WittierDinosaur (approved):
LGTM - great work!

PR #5390, opened 2023-11-08
Update publish actions and Dockerfile.

alanmcruickshank:
On publishing 3.0.0a1, several of the publish actions complained.

Mostly due to old version numbers (which I've updated here), but also because the update to `pyproject.toml` broke the docker image. This fixes that too.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,23 +103,6 @@ Changes = "https://github.com/sqlfluff/sqlfluff/blob/main/CHANGELOG.md"
 Twitter = "https://twitter.com/SQLFluff"
 Chat = "https://github.com/sqlfluff/sqlfluff#sqlfluff-on-slack"
 
-
-[tool.sqlfluff_docs]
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -103,23 +103,6 @@ Changes = "https://github.com/sqlfluff/sqlfluff/blob/main/CHANGELOG.md"
...
-# NOTE: Stable version is used by docs/conf.py
-stable_version = "2.3.5"
-
-
-[tool.setuptools]
-package-dir = {"" = "src"}
-
-[tool.setuptools.packages.find]
-where = ["src"]
-namespaces = false
```
alanmcruickshank, line 117: These two are removed because they crash `pip-compile` (which I use in the Dockerfile). Exactly _why_ they're a problem isn't clear, but they aren't necessary because they're effectively the defaults. I've tested building the package and it builds fine without them 👍

Issue #5418, opened 2023-11-17
Sqlfluff found unparsable section "URL" when creating a STAGE in Snowflake


k1drobot:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

I have a simple stage creation, 
```
CREATE OR REPLACE STAGE your_stage_name
  URL = 's3://your_s3_bucket/your_path_in_s3'
```
 and I get the following error:
 
 ```[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    statement:
[L:  1, P:  1]      |        create_stage_statement:
[L:  1, P:  1]      |            keyword:                                          'CREATE'
[L:  1, P:  7]      |            whitespace:                                       ' '
[L:  1, P:  8]      |            keyword:                                          'OR'
[L:  1, P: 10]      |            whitespace:                                       ' '
[L:  1, P: 11]      |            keyword:                                          'REPLACE'
[L:  1, P: 18]      |            whitespace:                                       ' '
[L:  1, P: 19]      |            keyword:                                          'STAGE'
[L:  1, P: 24]      |            whitespace:                                       ' '
[L:  1, P: 25]      |            object_reference:
[L:  1, P: 25]      |                naked_identifier:                             'your_stage_name'
[L:  1, P: 40]      |            [META] indent:
[L:  1, P: 40]      |            [META] dedent:
[L:  1, P: 40]      |    newline:                                                  '\n'
[L:  2, P:  1]      |    whitespace:                                               '  '
[L:  2, P:  3]      |    unparsable:                                               !! Expected: 'Nothing else in FileSegment.'
[L:  2, P:  3]      |        word:                                                 'URL'
[L:  2, P:  6]      |        whitespace:                                           ' '
[L:  2, P:  7]      |        equals:                                               '='
[L:  2, P:  8]      |        whitespace:                                           ' '
[L:  2, P:  9]      |        single_quote:                                         "'s3://your_s3_bucket/your_path_in_s3'"
[L:  2, P: 46]      |    newline:                                                  '\n'
[L:  3, P:  1]      |    [META] end_of_file:

==== parsing violations ====
L:   2 | P:   3 |  PRS | Line 2, Position 3: Found unparsable section: "URL =
                       | 's3://your_s3_bucket/your_path_in_..."
```

[...]

PR #5419, opened 2023-11-17
Simplify Snowflake regexes

greg-finley:

Closes https://github.com/sqlfluff/sqlfluff/issues/5418

The [link](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html) in the comment above the regex says:

```
Note
Before March 1, 2018, buckets created in the US East (N. Virginia) Region could have names that were up to 255 characters long and included uppercase letters and underscores. Beginning March 1, 2018, new buckets in US East (N. Virginia) must conform to the same rules applied in all other Regions.
```

But like, why bother maintaining these regexes? We just need to find that they start with the right thing to know which one it is.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_snowflake.py
+++ b/src/sqlfluff/dialects/dialect_snowflake.py
@@ -336,20 +336,19 @@ snowflake_dialect.add(
     ),
     S3Path=RegexParser(
         # https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
-        r"'s3://[a-zA-Z0-9_][a-zA-Z0-9_\.]{0,255}[a-zA-Z0-9_](?:/.*)?'",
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -5467,97 +5466,97 @@ class CreateStageSegment(BaseSegment):
                     optional=True,
                 ),
             ),
-            # External S3 stage
```
greg-finley, line 5470: Everything here and below is just a reindenting. I pulled

```
                "URL",
                Ref("EqualsSegment"),
```

up a level instead of repeating it in each case

alanmcruickshank (approved):
Looks sensible 👍

PR #5431, opened 2023-11-23
adds few fixes for databricks/sparksql

markbaas:
Following sql was causing parse errors:

```
CREATE TABLE alembic_version (
    version_num STRING NOT NULL,
    CONSTRAINT alembic_version_pkc PRIMARY KEY (version_num)
) USING DELTA;
```

This was fixed by setting sparksql to use the unused createtable implementation. There was a slight error there, that I also fixed. Now it works great.

Another parse error was with an update statement without where clause:

```
UPDATE table1
SET col1= 0
```

This was fixed by making the where clause optional.

Let me know what you think of it.


greg-finley (approved):
Thanks!

PR #5447, opened 2023-11-30
Update black linting

alanmcruickshank:
I think `black` has had some updates. This updates formatting accordingly.


mi-hol:
> I think `black` has had some updates. 

Isn't this actually a regression in `black`, because from a proper py syntax it makes no sense  to add trailing commas. 
Trailing comma could lead to further issues see https://stackoverflow.com/questions/3750632/why-does-adding-a-trailing-comma-after-an-expression-create-a-tuple

PR #5513, opened 2023-12-27
Re-enable subdir test for windows

alanmcruickshank:
This attempts to re-enable a test which has been disabled for failing on windows. It passes locally - so feels very viable to get it working again remotely.


alanmcruickshank:
This is basically superseded by #5579. There's some additional logic in here which should probably be rescued, but I'll do that as a different PR.

PR #5539, opened 2024-01-12
CV01: Add options for ANSI and consistent style.

### LINKED ISSUE #5529: CV01 `convention.not_equal`: Enable option to use `<>` in order to conform to ANSI SQL standard
### Search before asking

### Description

[Rule CV01](https://docs.sqlfluff.com/en/stable/rules.html#rule-convention.not_equal) `convention.not_equal` enforces consistent usage of the inequality operator `!=`:

> Use `!=` instead of `<>` for “not equal to” comparisons.

The reason given is:

> `<>` means not equal but doesn’t sound like this when we say it out loud.
> Use `!=` instead because its sounds more natural and is more common in other programming languages.

However, usage of `!=` does not comply to [ANSI 92 standard](https://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt), which specifies only `<>` for the `not equals operator`.

> ```        <not equals operator> ::= <>```

Some dialects only support `<>` and not `!=`:

- IBM DB2 UDB 9.5
- Microsoft Access 2010

It also seems counter-intuitive that a tool designed to enforce standards for SQL would only enforce an option that is "more common in _other_ programming languages" and was allegedly added to please C programmers.

Right now there is no way to enforce consistent usage of `<>` via sqfluff configuration.

I think there should be an option to enforce consistent usage of `<>` instead of `!=` given it's ANSI SQL compliant.

### Use case

If the user is trying to conform to ANSI SQL standards, they should have the option to enforce consistent usage of `<>` for the `not equals operator`.

We could add a `not_equals_policy` with options `<>`, `!=`, and `consistent`.

### Dialect

All dialects

### Code of Conduct

keraion:

fixes #5529 
This adds an ANSI (`<>`) style and consistent configuration option to CV01.

This changes the default setting from the C (`!=`) style to consistent.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -325,6 +325,10 @@ max_alias_length = None
 # We suggest instead using aliasing.length (AL06) in most cases.
 force_enable = False
 
+[sqlfluff:rules:convention.not_equal]
+# Consistent usage of preferred "not equal to" comparison
+preferred_not_equal_style = consistent
+
 [sqlfluff:rules:convention.select_trailing_comma]
 # Trailing commas
 select_clause_trailing_comma = forbid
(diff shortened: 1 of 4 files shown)


alanmcruickshank (approved):
Good addition. Well executed 👍

PR #5543, opened 2024-01-13
Improve support for Jinja templater plugins with custom tags

### LINKED ISSUE #4624: JinjaAnalyzer: Provide better support for custom Jinja tags / statements provided by custom Jinja extensions

james-johnston-thumbtack:
This would resolve #4624 by providing an easy way for plugins to provide templaters that have custom Jinja tags/extensions loaded in their environments.  The biggest problem is that JinjaAnalyzer has a lot of hard-coded tag names in it.  This commit would allow a plugin author to more easily support custom tags:

1.  The JinjaAnalyzer class now has a _get_tag_configuration function that returns a tag configuration for a given tag name.  Derived classes can override this function to provide additional behaviors for tag names.  The initial configuration provides a mapping of tag name to block type (e.g. block_start, block_mid, block_end). Plugins can also opt into the block tracing functionality.  For example, this can help ensure that a final tag at the end of the file without a trailing newline is preserved.

2.  The JinjaTemplater class did not provide an easy way for users to provide a derived JinjaAnalyzer class.  This class is refactored to add a new _get_jinja_analyzer function; this allows users to return a custom analyzer if they want.  This is similar in idea to the _get_jinja_env function, which allows returning a custom Jinja environment.

Some tests have also been included that test the behavior of a derived JinjaAnalyzer and JinjaTemplater.  This also effectively provides a small example of how to use it.

The provided test case also provides an example of why this commit is useful.  Without it, the sample input string will contain two "block_start" slices, and a "block_end": this is mismatched and causes all kinds of problems.  With this commit, it's easy to configure the analyzer so that the input is parsed with a single "block_start", "block_mid", and "block_end".  That makes things work much better!

Improve support for Jinja templater plugins with custom tags

Fixes #4624 

The code has been written so that any existing unknown Jinja tags will still parse exactly the same way they did before.  For example, the fallback behavior of assuming strings beginning with `el` are `block_mid` and strings beginning with `end` are `block_end` is preserved.  However, this commit will provide much better mechanisms for users to control the behavior than relying on that.

### Pull Request checklist

  - Other.


james-johnston-thumbtack:
This PR is ready for review!  I'll continue keeping it up-to-date with the main branch.

barrywhart:
This looks good to me. You may be the first person to delve into `JinjaAnalyzer` and `JinjaTracer`. That's some tricky but very vital code. 😅 Nice work! 🙌🏽🎉

I'll hold off approving -- @alanmcruickshank may want to be aware of this since he's working on some related functionality to the `_handle_unreached_code()` function, which is slightly modified by this PR.

src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -253,6 +275,118 @@ def __init__(self, raw_str: str, env: Environment) -> None:
...
+            block_type="block_end",
+        ),
+        "filter": JinjaTagConfiguration(
+            block_type="block_start",
+        ),
+        "endfilter": JinjaTagConfiguration(
+            block_type="block_end",
+        ),
+        # Common extensions
+        # Expression statement (like {{ ... }} but doesn't åctually print anything)
```
alanmcruickshank, line 356:
```suggestion
        # Expression statement (like {{ ... }} but doesn't actually print anything)
```

Scandinavian?

alanmcruickshank (commented):
One typo in there, but otherwise looks good. I'm happy to merge this in - I don't think it clashes with anything I've been doing (and I haven't been very successful in that anyway!). 👍 

Well done for diving into this code

james-johnston-thumbtack:
Thank for the review!  Good catch on the typo.  The proposed correction looks good to me - it looks like you've already fixed it. :)

alanmcruickshank (approved):

PR #5562, opened 2024-01-25
[Docs] Use extended policy for identifier capitalisation in starter config

j-svensmark:
Update docs starter config to use `extended_capitalisation_policy` for `capitalisation.identifiers`, since `capitalisation_policy` seems to not work there.

https://docs.sqlfluff.com/en/stable/rules.html#rule-capitalisation.identifiers


alanmcruickshank (approved):
Good catch! Thanks. 👍

PR #5580, opened 2024-01-29
Stricter recommended config for not_equals convention

alanmcruickshank:
Following on from #5539 (which relaxes the default configuration), this updates the _recommended_ starter config to reflect the previous default behaviour.


WittierDinosaur (approved):
LGTM

PR #5585, opened 2024-01-30
Memory overhead optimisations during linting

### LINKED ISSUE #5566: Linting a large dbt project consumes a lot of RAM
### Search before asking

### What Happened

When linting a dbt project with 4300+ models with `sqlfluff lint` sqlfluff's ram usage grows until all ram in the system is consumed.

### Expected Behaviour

sqlfluff linting should not consume more memory as a dbt project grows.

### Observed Behaviour

sqlfluff will be killed by the OS when all ram is consumed

### How to reproduce

given a large dbt project, run `sqlfluff lint` monitor ram usage

### Dialect

snowflake

### Version

2.3.2

### Configuration

N/A

### Code of Conduct

alanmcruickshank:
Resolves #5566 .

It adds a setting which doesn't keep the full parsed and linted file during linting operations. Only the metadata would then be kept - in theory lowering the memory overhead during long linting operations.

@noel - this is the feature I was talking about. UPDATE: Tested by noel on their big project and seems to reduce OOM killing events.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,7 +90,7 @@ dependencies = [
     "toml; python_version < '3.11'",
     # For handling progress bars
     "tqdm",
-    # better type hints for older python versions
+    # Literals and TypedDicts
     "typing_extensions",
 ]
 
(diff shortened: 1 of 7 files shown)


WittierDinosaur (approved):
LGTM

PR #5615, opened 2024-02-08
Add Support for Databricks `CREATE FUNCTION` Syntax in SparkSQL Parser

mitchellvanrijkom:
This PR introduces support for the Databricks-specific `CREATE FUNCTION` syntax within the SparkSQL parser, addressing the current limitation where only standard SparkSQL dialect for creating functions is supported. The need for this enhancement arises from the differences in syntax between the standard SparkSQL `CREATE FUNCTION` command and the Databricks variant, which can lead to parsing errors when working with Databricks SQL scripts.

More documenantation can be found [here](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-sql-function.html).

No. This PR only affects the parsing of `CREATE FUNCTION` implementation for Databricks SQL variant.

### Pull Request checklist

  - Other.


WittierDinosaur:
Why are you adding this to the sparksql dialect, and not the databricks dialect?

mitchellvanrijkom:
Thank you for bringing this up. Since the SparkSQL dialect already includes Databricks-specific features like `CreateWidgetStatementSegment` and `ApplyChangesIntoStatementSegment`, it seemed consistent to also implement Databricks' `CREATE FUNCTION` functionality within the SparkSQL dialect.

WittierDinosaur:
I think my concern here is that, while yes we do have databricks extensions in the sparksql dialect (though this is actually for legacy reasons), things like `CREATE FUNCTION` are core sql statements, not dialect specific extensions. I'm worried here that you may be overwriting spark's syntax with databricks, making breaking changes for spark users that arent databricks users

mitchellvanrijkom:
I will update the code so that the `CREATE FUNCTION` is available in the Databricks dialect, as I understand it can impact Spark dialect users.

phant0mw0lf:
@mitchellvanrijkom thanks for implementing this. does this also works for RETURNS TABLE? e.g.

```sql
CREATE OR REPLACE FUNCTION test_function (param BIGINT DEFAULT 0)
RETURNS TABLE
RETURN
SELECT 1 AS Test;

SELECT * FROM test_function()
```

mitchellvanrijkom:
@phant0mw0lf This implementation is only focused on SQL scalar functions.

WittierDinosaur (approved):
LGTM

WittierDinosaur:
@mitchellvanrijkom why did you close this? I was ready to merge?

mitchellvanrijkom:
@WittierDinosaur my mistake. I thought I had to close it myself.

PR #5649, opened 2024-03-05
Remove `DeprecatedOption`

alanmcruickshank:
As we approach 3.x, it seems like a good opportunity to remove the previously deprecated `--disable_progress_bar` option which was deprecated about 2 yrs ago.

This also means we can remove a whole module that isn't therefore required.


WittierDinosaur (approved):
LGTM

PR #5650, opened 2024-03-05
Swap fix `--force` for `--check`

alanmcruickshank:
As discussed in the maintainers call this drafts a PR for reversing the default behaviour of `sqlfluff fix` to use what used to be `--force` mode by default (because of the improved memory overhead).

NOTE: Doesn't actually add the memory optimisation for `fix` yet, because I've got a failing test case but that will come shortly.


WittierDinosaur (approved):
LGTM

PR #5651, opened 2024-03-05
Resolve broken loop limit test

alanmcruickshank:
I think I found an issue.

We have a test case `test__cli__fix_loop_limit_behavior` which designed to check error behaviour around the loop limit on fixes. We have one test case designed to see what happens when the limit is reached - and look for an exit code of 1 if it does.

What is _actually_ happening here however is that we're getting an exception because `BaseRule._eval()` raises a `NotImplementedError` when run rather than actually because we're hitting the loop limit. That means this test case is misleading.

If I resolve that first issue - the test case then doesn't work anymore because the error code is different. @barrywhart - you're the last person in that chunk of code, what's your view on what we do here?

### DIFF
--- a/test/cli/commands_test.py
+++ b/test/cli/commands_test.py
@@ -34,7 +34,7 @@ from sqlfluff.cli.commands import (
     version,
 )
 from sqlfluff.core.parser import CommentSegment
-from sqlfluff.core.rules import BaseRule, LintFix, LintResult
+from sqlfluff.core.rules import LintFix, LintResult
 from sqlfluff.utils.testing.cli import invoke_assert_code
 
 # tomllib is only in the stdlib from 3.11+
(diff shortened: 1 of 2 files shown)


barrywhart:
My thoughts and advice:
- My _guess_ is that this test got quietly broken when we reworked the crawling behavior a couple years ago, to crawl the desired segment types efficiently rather than exhaustively walking over every single segment.
- This test is intended to cause an infinite loop in the linter by generating an unending sequence of non-repeating fixes.
- It does so in a somewhat hacky way, by monkeypatching `BaseRule`. We should rewrite this test to use a new, test-specific rule rather than monkeypatching -- I think that'll be much cleaner. See these [test-specific rules](https://github.com/sqlfluff/sqlfluff/blob/main/test/core/rules/rules_test.py#L28-L109) for an example.

alanmcruickshank:
@barrywhart I agree.

What I've done instead is totally remove that old cli test - I think it's misleading. There's a test in the `rules_test.py` which already tests explicitly hitting the runaway limit - and I've extended that test to also test the ignore behaviour which was tested in the original.

I think that's a simpler solution which gets us where we need to be.

PR #5653, opened 2024-03-05
Complete the memory overhead work on cli fix

### LINKED ISSUE #5588: Progressing fixing consumes much RAM
### Search before asking

### What Happened

This is related to #5566. That describes excessive memory usage during a `lint` operation - but the same is true during a `fix` operation. This is hard to solve when waiting for user input at the end (although possible by only keeping references to files which need fixing) - but easy to fix when in `--force` mode because fixes are made progressively and the data can be discarded once the fixes are persisted.

### Expected Behaviour

Don't keep all the data for files we don't need.

### Observed Behaviour

All data kept during the fixing operation and eventually all RAM is consumed on large projects

### How to reproduce

see #5566 

### Dialect

any

### Version

main

### Configuration

default

### Code of Conduct

alanmcruickshank:
This resolves #5588 .

Enabling the `fix` cli command to use the serialised violation objects rather than the native classes took a bit of coaxing. I think this achieves the end that we want. It required a fair bit of refactoring of several methods, including pulling some mutation logic out of the formatter!

There's likely another round of refactoring which could be done here, but it's already become a fairly large PR.


alanmcruickshank:
Urgh - ok not so fast.

This is going to need an overhaul of `persist_timing_records` so that can also work without the original file objects. That will take a bit more time. Watch this space.

src/sqlfluff/cli/commands.py
```diff
@@ -769,6 +769,35 @@ def do_fixes(
     return False  # pragma: no cover
 
 
+def _handle_unparsable(
+    fix_even_unparsable: bool,
+    initial_exit_code: int,
+    linting_result: LintingResult,
+    formatter,
```
WittierDinosaur, line 776: can we add a type here?

alanmcruickshank: Hmmmmm good question. We haven't typically typed that before - but it's a good shout that we probably should. I'll see if I can make that work 🤔

alanmcruickshank: @WittierDinosaur - yes I can! Sorted now

WittierDinosaur (approved):
LGTM

PR #5654, opened 2024-03-06
BQ PK and FK

OTooleMichael:
Adds support for PrimaryKey an ForeignKey which are now possible in BQ

### Pull Request checklist

  -[x]`.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  -[x] `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Other.


src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -1487,6 +1485,50 @@ class OptionsSegment(BaseSegment):
...
 
 
+class NotEnforcedSegment(BaseSegment):
+    """NOT ENFORCED clause for a primary key."""
+
+    type = "not_enforced_segment"
+    match_grammar = Sequence(
+        "NOT",
+        "ENFORCED",
+    )
```
tunetheweb, line 1495: Does this need to be a Segment? Or could it just be a simpler Grammar?

Segments are typically used when you need to reference this (but the `type`) in a rule but I doubt that's the case here.

Or alternatively, since this is so simple, should this just be hardcoded in the two places it's needed rather than being a separate class/grammar?

OTooleMichael: I was just mirroring the PrimaryKey and Foreign Key Segments. These two things together do constitute an expression block.

I'm happy either way

tunetheweb: Yeah let's remove this one if you don't mind. Seems needless. And there's lots of examples of simple expression blocks being hardcoded.

OTooleMichael: Cool

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -1487,6 +1485,42 @@ class OptionsSegment(BaseSegment):
...
+    """A foreign key reference segment."""
+
+    type = "foreign_key_reference_segment"
+    match_grammar = Sequence(
+        "REFERENCES",
+        Ref("TableReferenceSegment"),
+        Ref("BracketedColumnReferenceListGrammar"),
+        "NOT",
+        "ENFORCED",
+    )
```
tunetheweb, line 1499: Actually now I'm wondering if it's weird that the Foreign key has a special segment type and the primary key doesn't when it's only slightly simpler.

Given the other dialects don't pull this out into it's own class I think we should remove this, and just add the grammar to the `TableConstraintSegment`.

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -1487,6 +1485,42 @@ class OptionsSegment(BaseSegment):
...
+
+
+class TableConstraintSegment(BaseSegment):
+    """A table constraint segment.
+
+    https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#table_constraints
+    """
+
+    type = "table_constraint"
```
tunetheweb, line 1508: It seems we're inconsistent here and some dialects do what you do and redefine this, while others inherit it from ANSI and override the `match_grammar`.

I've no idea which is better to be honest, or if they are basically equivalent for this simple example? I mean it does save you explicitly setting the `type` (since that's in the base class it inherits from), and I guess it does make it more obvious it's overridden something from ANSI which seems good, but probably not a big deal either way (especially as some dialects do it one way and some another). WDYT?

```suggestion
class TableConstraintSegment(ansi.TableConstraintSegment):
    """A table constraint segment.

    https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#table_constraints
    """

```

tunetheweb (changes requested):

src/sqlfluff/dialects/dialect_bigquery.py
OTooleMichael: But its an almost completely different syntax - to my eye at least.

src/sqlfluff/dialects/dialect_bigquery.py
OTooleMichael: I dont mind :)

src/sqlfluff/dialects/dialect_bigquery.py
tunetheweb: I don't mean combine them into one section for both. I mean like this:

https://github.com/sqlfluff/sqlfluff/blob/ee8e5365a223ac1047462345d4b4eed816467120/src/sqlfluff/dialects/dialect_ansi.py#L3068-L3089

src/sqlfluff/dialects/dialect_bigquery.py
tunetheweb: Let's go for the override option. Seems like the better way of doing it.

src/sqlfluff/dialects/dialect_bigquery.py
OTooleMichael: So I inlined it.
I looked at using `Ref(ReferenceDefinitionGrammar)` but the definition is different.

tunetheweb (approved):
LGTM!

Thanks for looking at all the feedback!

PR #5661, opened 2024-03-08
Progressively load plugins

alanmcruickshank:
As called out by @noel on slack. A single failed plugin can nuke running SQLFluff, this adds a bit more security.

If a plugin fails to load, we carry on and try the others, with an error message which includes the plugin name and version, the current SQLFluff version and a reference to the exception which caused the issue (but no long trace).


src/sqlfluff/core/plugin/host.py
```diff
@@ -42,7 +60,24 @@ def get_plugin_manager() -> pluggy.PluginManager:
     # points, this function gets called again - and we only
     # want to load the entry points once!
     _plugin_manager.set(plugin_manager)
-    plugin_manager.load_setuptools_entrypoints(project_name)
+
+    # Discover available plugins an load them individually.
```
barrywhart, line 64:
```suggestion
    # Discover available plugins and load them individually.
```

src/sqlfluff/core/plugin/host.py
```diff
@@ -42,7 +60,24 @@ def get_plugin_manager() -> pluggy.PluginManager:
...
+    # If any fail, log the issue and carry on.
+    for plugin_name, plugin_version in _discover_plugins():
+        plugin_logger.info(f"Loading plugin {plugin_name} version {plugin_version}.")
+        ep = importlib.metadata.entry_points(group=project_name, name=plugin_name)
+        assert ep
```
barrywhart, line 69: Is the assertion necessary? Do we think it may fail? I wonder if we'd want to catch and report this as an error, too.

src/sqlfluff/core/plugin/host.py
```diff
@@ -25,6 +29,56 @@
...
+    """Loads a single plugin with a bit of error handling."""
+    # NOTE: If the plugin is already loaded, then .register() will fail,
+    # so it's important that we check whether it's loaded at this point.
+    if plugin_manager.get_plugin(plugin_name):  # pragma: no cover
+        plugin_logger.info("...already loaded")
+        return None
+    try:
+        plugin = entry_point.load()
+    except Exception as err:
+        plugin_logger.error(
```
barrywhart, line 70: This will log the error message but no stack backtrace. Could a stack backtrace be useful here?

src/sqlfluff/core/plugin/host.py
```diff
@@ -25,6 +29,56 @@
...
+    return importlib.metadata.version("sqlfluff")
+
+
+def _discover_plugins() -> Iterator[Tuple[importlib.metadata.EntryPoint, str, str]]:
+    """Uses the same mechanism as pluggy to introspect available plugins.
+
+    This method is then intended to allow loading of plugins individually,
+    for better error handling.
+    """
+    for dist in list(importlib.metadata.distributions()):
```
barrywhart, line 47: Nit: Is it necessary to wrap `list()` around this? No big deal, just curious.

barrywhart (approved):
Approved. A couple of minor questions.

PR #5662, opened 2024-03-09
Revert Ruff Changes

alanmcruickshank:
As brought up on slack, it appears these ruff config changes are causing issues with the pre-commit hook.


WittierDinosaur (approved):
LGTM

PR #2764, opened 2022-03-02
Snowflake: Add CHANGES clause

jpy-git:
### Brief summary of the change made

Adds the `CHANGES` clause to the Snowflake dialect:
https://docs.snowflake.com/en/sql-reference/constructs/changes.html

### Are there any other side effects of this change that we should be aware of?
No

### Pull Request checklist
- [X] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #5698, opened 2024-03-19
Allow use of STREAM in snowflake CHANGES clause

alanmcruickshank:
Does what it says on the tin. Looks like we missed one keyword during the initial implementation https://github.com/sqlfluff/sqlfluff/pull/2764.

This resolves an issue raised on slack.

### DIFF
--- a/src/sqlfluff/dialects/dialect_snowflake.py
+++ b/src/sqlfluff/dialects/dialect_snowflake.py
@@ -1523,14 +1523,16 @@ class ChangesClauseSegment(BaseSegment):
             "INFORMATION",
             Ref("ParameterAssignerSegment"),
             OneOf("DEFAULT", "APPEND_ONLY"),
+            parse_mode=ParseMode.GREEDY,
         ),
         OneOf(
             Sequence(
                 "AT",
                 Bracketed(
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -1523,14 +1523,16 @@ class ChangesClauseSegment(BaseSegment):
             "INFORMATION",
             Ref("ParameterAssignerSegment"),
             OneOf("DEFAULT", "APPEND_ONLY"),
+            parse_mode=ParseMode.GREEDY,
```
tunetheweb, line 1526: Why are these needed now out of interest?

tunetheweb (approved):

src/sqlfluff/dialects/dialect_snowflake.py
alanmcruickshank: Not needed at all, then just mean slightly more specific parsing errors if snowflake change the syntax again. I'm guessing some of those keywords might change and if they do - only the bracketed section that has changed will fail and we'll get a faster report on the issue.

PR #5715, opened 2024-03-25
added refresh mode init on create table statement

IliyanKostov9:
- Added `REFRESH_MODE` & `INITIALIZE` snowflake sql clause support on `CREATE TABLE` statement

- added snowflake dialect sets for both refresh mode & initialize available values
- added `REFRESH_MODE` & `INITIALIZE` keywords to unreserved snowflake words
- added TC for the proposed change

- No side effects have been introduced.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -193,6 +193,16 @@
...
 
+snowflake_dialect.sets("refreshmode_types").clear()
+snowflake_dialect.sets("refreshmode_types").update(
+    ["AUTO", "FULL", "INCREMENTAL"],
+)
+
+snowflake_dialect.sets("initialize_types").clear()
+snowflake_dialect.sets("initialize_types").update(
+    ["ON_CREATE", "ON_SCHEDULE"],
+)
```
tunetheweb, line 204: Does this need to be a set rather than just including directly in the grammar?

src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -272,6 +282,20 @@
             type="warehouse_size",
         ),
     ),
+    RefreshModeType=OneOf(
+        MultiStringParser(
+            snowflake_dialect.sets("refreshmode_types"),
+            KeywordSegment,
+            type="refreshmode_type",
+        )
+    ),
```
tunetheweb, line 291: Why not just this?:

```suggestion
    RefreshModeType=OneOf(
        "AUTO",
        "FULL",
        "INCREMENTAL"
    ),
```

The main difference is you cannot set `type="refreshmode_type"` in this way, but unless you explicitly need to refer to this elsewhere (e.g. you have a rule to pick out this particular syntax) that's usually not needed.

I say this has been done for a lot of other Snowflake syntax (`compression_types`, `files_types`...etc.) but not entirely sure it's necessary, or that it's something that should be continued....

src/sqlfluff/dialects/dialect_snowflake.py
IliyanKostov9: Well initially I actually have tried to include it directly as you suggested, but then my test cases start failing - therefore I had to do some additional grammar modifications on the segment at question. Even with that it then started giving me false positives on the TCs (e.g the tests were passing even if my value segment of REFRESH_MODE had incorrect value).

Nevertheless, I think we should at least stick with this kind of separation between the dialect values and the segment definition themself (maybe even think to move it to different file ?)

src/sqlfluff/dialects/dialect_snowflake.py
IliyanKostov9: I have also tried your second suggestion, and it also fails, particularly on the `Match Pattern Received` stage `generate-fixture-yml`.  It looks like without the MultiStringParser argument, it's interpreting in some other way (maybe this one can be described in more detail somewhere in the doc).

Onto your second suggestion for removing the `type` part, I have removed it and no new issues have arisen, the only reference is in the TC `create_table.yml` content, it had switched the `AUTO` object key `refreshmode_type` to only `keyword`. 

In regard to the question if it should be continued, it really depends if it's going to affect the normal behaviour of the parser from the snowflake side, I'm not 100% sure even if the TCs are covering this aspect at all, but still we can put it on the backlog as part of refactoring the code :)

tunetheweb (approved):
OK I see the other benefit of doing it this way instead of via Grammars is that you don't need to add the keywords as reserved or unreserved keywords. Of course that has pluses and minuses but maybe that's why this was done.

Anyway happy to approve this since it seems there's some reasons for doing this.

Issue #5356, opened 2023-10-26
Linting error when using aliases in backticks


brianstoop:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

When using backticks in table aliases and using the alias in the SELECT statement gives an error.

### Expected Behaviour

I expect the linting to succeed

### Observed Behaviour

The sqlfluff lint failed with the following error:
```
L:   1 | P:   8 | RF01 | Reference '`nih`.`userID`' refers to table/view not
                       | found in the FROM clause or found in ancestor statement.
                       | [references.from]
L:   2 | P:  44 | AL05 | Alias '`nih`' is never used in SELECT statement.
                       | [aliasing.unused]
```

### How to reproduce

```
SELECT `nih`.`userID`
FROM `flight_notification_item_history` AS `nih`
```

### Dialect

mysql

### Version

sqlfluff, version 2.3.4

### Configuration

[sqlfluff]
templater     = jinja
sql_file_exts = .sql
large_file_skip_byte_limit = 0

[sqlfluff:rules]
capitalisation_policy = upper

[sqlfluff:rules:references.quoting]
prefer_quoted_identifiers = True

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

pprkut:
@greg-finley The original issue here was for the mysql dialect. I'm not sure that is covered by the linked pull request

greg-finley:
OK, yeah, you're right. Thanks.

PR #5718, opened 2024-03-26
MySql: Fix backticked identifier wrongly being reported as not found

pprkut:

Fixes the MySQL part of #5356

### Pull Request checklist

  - Other.


pprkut:
I added tests to verify the issue, but I'm not sure they are the correct tests, as they pass both with and without my change. Pointers appreciated on where I could add more appropriate tests :slightly_smiling_face:

test/fixtures/dialects/mysql/back_quoted_identifier.sql
```diff
@@ -0,0 +1,9 @@
+SELECT * FROM `foo`.`bar`;
```
WittierDinosaur, line 1: These are parser tests, to test if your change behaves as expected, you need to test the rules. See here: https://github.com/sqlfluff/sqlfluff/tree/main/test/fixtures/rules/std_rule_cases

pprkut:
I've added a couple more tests now and came to the conclusion that removing `trim_chars=("`",)` doesn't actually fix it.

The problem is that the aliases that get checked in https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/rules/references/RF01.py#L216 do not have the backticks trimmed. I think I traced this back all the way to `get_eventual_alias()` in https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/dialects/dialect_ansi.py#L1554, but I'm not sure where to take it from here.

Does this mean `dialect_mysql.py` would need a custom `get_eventual_alias()` that removes the backticks? Or am I on the wrong track here?

pprkut:
I missed that #5726 seems to already have fixed this. The test case I had no longer fails with the latest version of sqlfluff

PR #5728, opened 2024-03-28
Added support for 'greater/less than or equal' on ANSI CASE statement

IliyanKostov9:
- Added suppot for greater/less or equal operator on CASE statement for ANSI dialect

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2000,7 +2000,14 @@ class CaseExpressionSegment(BaseSegment):
             Dedent,
             "END",
         ),
-        terminators=[Ref("CommaSegment"), Ref("BinaryOperatorGrammar")],
+        terminators=[
+            OneOf(
```
WittierDinosaur, line 2004: I don't think this OneOf is needed, as the OneOf is implicit in the terminators handling

IliyanKostov9: Yep, you are right 👍 
I didn't try without `OneOf`, it's still retaining the same feature result.
Removed it

IliyanKostov9:
Hi @WittierDinosaur, I'm finished with my changes, can you please have another look ?

src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -2000,7 +2000,12 @@ class CaseExpressionSegment(BaseSegment):
             Dedent,
             "END",
         ),
-        terminators=[Ref("CommaSegment"), Ref("BinaryOperatorGrammar")],
+        terminators=[
+            Ref("GreaterThanOrEqualToSegment"),
+            Ref("LessThanOrEqualToSegment"),
+            Ref("CommaSegment"),
```
WittierDinosaur, line 2006: Sorry, one more thing. Does this support equal to? Can you add a test case for `CASE WHEN ... END == 1` ?

IliyanKostov9: No, this part doesn't cover the `==` part, due to the fact that currently ansi dialect doesn't support the use case for `==` segment (e.g. even if I add `Ref(EqualsSegment)` it still won't work.

So, I'm thinking of adding a new segment to the ansi_dialect and creating a corresponding class that handles `==`. But I'm kinda surprised no one had added this yet. Any particular reason, or it was just missed until now ?

WittierDinosaur: Oh hang on, `==` is python, it should just be equals. We also need not equals. Let me know whether you'd rather add to this PR, or for me to merge it. Happy to go either way

IliyanKostov9: lol, I also got confused for some magic reason, my bad.

Sure, I would like to add this part to the PR, since we already opened a discussion about it, might as well add it here.
I actually found a cleaner solution for adding the comparison segment by Alan - `ComparisonOperatorGrammar` does it all-in-one.
Checked it from my side, and it works as expected.

WittierDinosaur (approved):
LGTM

PR #5729, opened 2024-03-29
Fix spelling

mdeweerd:
# Fix spelling

Fix multiple typos with codespell.

PR #5730, opened 2024-03-29
Qual: Add pre-commit to CI

mdeweerd:
# Qual: Add pre-commit to CI

This adds a pre-commit workflow to CI which runs the default pre-commit hooks.
I also added a hook for codespell.
I corrected the exclusion filter for line-endings (i.e., '.*' was missing, but I made the exceptions compliant with .editorconfig).

FYI: The mdeweerd/logToCheckstyle github action can also convert sqlfluff messages to github annotations.

You'll notice that his new workflow:
- reveals a lot of files with unexpected line-endings - some in the documentation, but most in the SQL files - you have to decide if these need to be excludes;
- Reports spelling errors (fixed in #5729).


mdeweerd:
Codespell is confirmed ok - now it's only the end-of-file  and  end-of-line whitespaces that result in a failing pre-commit workflow.
Options:
1. Update the pre-commit configuration to exclude these files;
2. Fix the files (run pre-commit locally, commit and merge);
3. Exclude these steps from the workflow run.

I think that the 2nd option is most appropriate as long as these whitespaces are not needed for the testcases at hand.

WittierDinosaur:
Yep, I think this is a good change - files without newlines annoy me, let's add the fix of the files into this PR

mdeweerd:
All fixed:
- I committed each file separately to be able to do git bisects;
- I added exceptions to the end-of-* fixing;
- I fixed a typo to an exception in the original pre-commit configuration;
- I updated .editorconfig to match the exceptions added to the pre-commit configuration.

WittierDinosaur (approved):
Fantastic effort, thanks for this!

Issue #5732, opened 2024-03-30
SQLite statements with ALTER TABLE are not parsed


caffeinatedgaze:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

I used sqlfluff to format my sqlite code and it failed to parse `ALTER` statements.

### Expected Behaviour

Parsing `ALTER` statements.

### Observed Behaviour

Regardless of the fact that `sqlite` is derived from `ansi` dialect code, parsing of `alter` statements fails. 

### How to reproduce

Sample SQL code that causes trouble.

```sql
ALTER TABLE message RENAME TO nachrichten;
```

Command that fails.

```bash
$ sqlfluff --dialect sqlite trouble.sql
RuntimeError: Grammar refers to the 'Modify' keyword which was not found in the sqlite dialect. 

```

### Dialect

sqlite and ansi

### Version

```
$ sqlfluff --version
sqlfluff, version 3.0.3

$ python3 --version
Python 3.8.18
```

### Configuration

Default. I only pass the dialect as a keyword flag.

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

caffeinatedgaze:
This parsing error cannot be ignored by means other than `.sqlfluffignore`, which is an overkill for me. Steps to reproduce:

```sh
echo 'ALTER TABLE messages ADD author_id INTEGER NULL REFERENCES users (id); -- noqa: PRS' | sqlfluff format --dialect sqlite -
```

keraion:
Closed by #5888

PR #4765, opened 2023-04-16
SQLite: Remove refs to RESPECT and QUALIFY

WittierDinosaur:
### Brief summary of the change made
Fixes #4600 
Fixes #4759 

### Are there any other side effects of this change that we should be aware of?
No

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #5733, opened 2024-03-30
Fix `sqlite` dialect implementation breaking original `ansi` rules

caffeinatedgaze:
Responding to https://github.com/sqlfluff/sqlfluff/issues/5732, I figured that `sqlite` dialect implementation somehow breaks the original `ansi` rules for `ALTER` statements. Trial and error showed that keeping the original unreserved keywords rectifies it.

My fix is fairly rough, that is why I would want to request a second opinion from the sqlruff maintainers with deeper knowledge of the code. Could you share your thoughts?

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_sqlite.py
+++ b/src/sqlfluff/dialects/dialect_sqlite.py
@@ -36,7 +36,7 @@ sqlite_dialect = ansi_dialect.copy_as("sqlite")
 
 sqlite_dialect.sets("reserved_keywords").clear()
 sqlite_dialect.sets("reserved_keywords").update(RESERVED_KEYWORDS)
-sqlite_dialect.sets("unreserved_keywords").clear()
+# sqlite_dialect.sets("unreserved_keywords").clear()
 sqlite_dialect.sets("unreserved_keywords").update(UNRESERVED_KEYWORDS)
 
 sqlite_dialect.patch_lexer_matchers(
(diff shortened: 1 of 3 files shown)


WittierDinosaur:
Thanks for taking a look at this bug. We intentionally reset the keywords for sqlite, as it is a small dialect, and supports very limited sql functionality. What you've come across here is a part of the alter table statement which, in ansi, contains a keyword that doesn't exist in sqlite. In other examples of this, we've removed that part of the statement from sqlite. Here is an example PR where I did this for `RESPECT` and `QUALIFY` https://github.com/sqlfluff/sqlfluff/pull/4765

caffeinatedgaze:
> alter table statement which, in ansi, contains a keyword that doesn't exist in sqlite

Could you clarify this one for me? I am not sure which steps to take next. It looks to me that the statements I provided are valid for SQLite.

WittierDinosaur:
Okay, so we need to get this working without changing how the keywords are built. 

So, to investigate what's happening, you need to follow the hierarchy in ansi. 

Here is the alter table statement in ANSI https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/dialects/dialect_ansi.py#L3262C47-L3263C1

Modify is not there, so we need to go to there AlterTableOptionsGrammar
https://github.com/sqlfluff/sqlfluff/blob/7197f3fedf2aa5bbe4e2c035366a50a0179a0a77/src/sqlfluff/dialects/dialect_ansi.py#L663

As you can see, Modify is in there. So what we need to do is overwrite this grammar in sqlite, to only contain valid sqlite options. Modify is invalid, and there may be some other invalid stuff too, sqlite's docs are very good on what is allowed

greg-finley:
Thanks for getting this started, but this looks to be supported also in https://github.com/sqlfluff/sqlfluff/pull/5733

PR #5781, opened 2024-04-19
MySQL: Added SET TRANSACTION parsing

Xemptuous:
Added various ways of parsing SET TRANSACTION statements in MySQL according to [the docs](https://dev.mysql.com/doc/refman/8.0/en/set-transaction.html)

None that I am aware of.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -2721,6 +2722,37 @@ class ReturnStatementSegment(BaseSegment):
...
+    As specified in https://dev.mysql.com/doc/refman/8.0/en/set-transaction.html
+    """
+
+    type = "set_transaction_statement"
+    match_grammar: Matchable = Sequence(
+        "SET",
+        OneOf("GLOBAL", "SESSION", optional=True),
+        "TRANSACTION",
+        Delimited(
+            AnySetOf(
```
WittierDinosaur, line 2737: Chaining Delimited and AnySetOf doesn't really work, as Delimited inherits from OneOf - can you remove the AnySetOf here?

Xemptuous: My intention was to only allow a max of 2 of these options, as you can't have more than 2 of these options at once. e.g., this won't work:
```sql
SET TRANSACTION ISOLATION LEVEL SERIALIZABLE, READ ONLY, READ WRITE;
```
I didn't see any args in `Delimiter()` to allow a max # of items.
Is there a better way to do this?

WittierDinosaur: Nah, there's no way to get what you're going for here cleanly. We'd need some kind of DelimitedSet operator, which we don't have at the mo

alanmcruickshank (approved):
Resolved PR feedback

PR #5458, opened 2023-12-06
Change Color.lightgrey to have a white background - dark theme friendly

ryaminal:
### Brief summary of the change made
To allow a better experience for dark-background users, adding a background of WHITE seems to be a good experience.

Tested on about 10 themes in MacOS' Terminal.app and alacritty for both dark and light backgrounds.

Related to #1170 

Explored replacing the output formatter with `rich`, but that seems like too much work when I think this simple fix will do.
### Are there any other side effects of this change that we should be aware of?
No known side effects

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

Issue #1170, opened 2021-06-24
Colored output doesn't play nicely with solarized dark theme


louis-vines:
I can't see the filenames on the output of sql fluff when using solarized dark theme

## Expected Behaviour
Output using dark terminal theme:
<img width="709" alt="image" src="https://user-images.githubusercontent.com/10738810/123290735-6ae8dd80-d509-11eb-8006-e91d7b27271b.png">

## Observed Behaviour
Output using solarized dark terminal theme:
<img width="721" alt="image" src="https://user-images.githubusercontent.com/10738810/123291073-b00d0f80-d509-11eb-888c-cf3bc89f1116.png">

## Steps to Reproduce
Run sqlfluff with solarized dark themed console

## Version
```
❯ sqlfluff --version
sqlfluff, version 0.6.0
```

## Configuration
Not applicable

barrywhart:
Note that SQLFluff can't detect the color pallet of the terminal. We could consider adding a way to specify the colors (or color theme) you want. Note that the `--nocolor` option suppresses all colors from the output, so that's the best option for now.

alanmcruickshank:
Given the suggestion above of using `--nocolor` and the lack of other developments on this issue. I'm going to close it as stale.

PR #5784, opened 2024-04-19
Improve light colour highlight

alanmcruickshank:
Ok, so this might be a little controversial.

I understand why #5458 was merged - the previous situation didn't work well with dark mode terminals (see #1659 and #1170). _However_ I'm finding that the solution of changing the background is _also_ quite hard to read on some other themes (in particular on powershell), so I've played with a few options, and I think this might be an improvement on both.

This changes the `lightgrey` colour to just `light` and maps it to the ANSI _yellow_ code. It does change the overall feel quite a bit, but I think it's a good compromise. Here's some shots of what it looks like on a few different colour schemes:

Ubuntu
image

Monokai
image

Solarized
image

Unfortunately I don't have a mac and so can't confirm exactly what it looks like on MacOS dark solarized as per the original request.

@ryaminal , @louis-vines,  @puetsch - as the original posters of the linked issues, are you able to try this variant and confirm whether it's readable on MacOS dark mode? @WittierDinosaur if you're on a Mac, you might also be able to confirm?

### DIFF
--- a/src/sqlfluff/cli/formatters.py
+++ b/src/sqlfluff/cli/formatters.py
@@ -516,9 +516,7 @@ class OutputStreamFormatter:
             val = "" if v is None else str(v)
             text_buffer.write(
                 ("    " * i)
-                + self.colorize(
-                    pad_line(str(k) + ":", 20, "left"), color=Color.light
-                )
+                + self.colorize(pad_line(str(k) + ":", 20, "left"), color=Color.light)
                 + pad_line(val, 20, "left")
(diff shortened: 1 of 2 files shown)


ryaminal:
@alanmcruickshank, looks good to me. thanks for actually fixing the problem instead of bandaging it like i did. :)

WittierDinosaur:
MacOS
image

WittierDinosaur (approved):
LGTM

Issue #4061, opened 2022-11-11
Add the ability to lint and fix areas of templated code that are not actually executed


barrywhart:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

Several users have asked if SQLFluff does (or could) support this feature. Another SQL formatting tool, [sqlfmt](https://docs.sqlfmt.com/integrations/sqlfluff) supports this (although it's less sophisticated and doesn't actually parse the SQL code.)

Alan and Barry discussed how that might be done in SQLFluff:
this pr means we got fairly accurate links between closing and starting jinja tags: https://github.com/sqlfluff/sqlfluff/pull/3936
and the rework of placeholder segments (currently bound up in the reindent pr), means that we know where code is being rendered and where not.
...so we could detect if statements and find cases where code is never rendered.
the only thing we'd need to do would be to allow the templater to optionally return multiple versions of a query, parse them all and then lint them all - deduplicating the results.
if one of the versions fails, we'd probably hide the parsing errors because it could be use incorrectly interpreting the loops, because we wouldn't really understand the loops, we're just replacing code.
....thinking about it - the other thing we could do would be preprocess the templates, replacing any if statements with True or False to force clauses to be evaluated. :shrug:
all of this feels quite hacky though. Maybe just allowing users to add a ` --sqlfluff` directive which tells the templater what other options to insert for some context values to generate multiple versions of a query. Whatever we do it all boils down to linting multiple versions of the same query - which has a performance overhead.
that being said. templating is fast. we could pretty quickly evaluate for any query if any alternative variations exist. For most queries, they won't - so performance is the same. We could also evaluate how many exist, and limit ourselves to only evaluating the most significant n (e.g. the variations with the most different code). Altering the templater to optionally return an iterable of TemplatedFile wouldn't be too hard, and deduplicating errors based on their source position already exists (https://github.com/sqlfluff/sqlfluff/pull/4041) and the same method could easily be used here, so that the user doesn't know the file got linted more than once.
I'm worried about the potential for extreme cases. Just as we have people trying to lint 100,000 line files, someone will try to line a file with 10,000 if statements, which means 2^10,000 possible paths.
Do you have thoughts on how to determine and limit to the most significant variations?
limiting is easy once we know how many variations.
determining the variations is the hard bit.
I'm tempted initially to just start generating them and then just stop iteration after generating 5.
then we filter from those files.
[...]

PR #5785, opened 2024-04-19
Jinja Variant Configuration

alanmcruickshank:
This makes progress on #4061 . I got a bit burnt out before so I'm trying to keep the steps small.

This introduces the basic configuration for linting variants, and crosses the boundary into the linter. That also enables the `sqlfluff render` cli command to show rendered variants which is now supported.

image

It defaults to using `process_with_variants` for all templaters, and so should trigger a minor release because it will be a breaking change for anyone using custom templaters. Quite a few type hints were inaccurate - which is one of the reasons this has been a little convoluted. I've straightened out a bunch of them in the process of this work.

Unless someone _explicitly_ sets a configuration value to enable this - nothing is enabled. The default value for `sqlfluff:render_variant_limit` is 1 which effectively disables it unless changed.


src/sqlfluff/core/linter/linter.py
```diff
@@ -714,23 +723,34 @@ def render_string(
...
+
         try:
-            templated_file, templater_violations = self.templater.process(
+            for variant, vs in self.templater.process_with_variants(
                 in_str=in_str, fname=fname, config=config, formatter=self.formatter
-            )
+            ):
+                if variant:
+                    templated_variants.append(variant)
+                templater_violations += vs
```
WittierDinosaur, line 737: does this need to be a set of violations? Could we get repeats if we're doing all variants?

src/sqlfluff/core/linter/linter.py
```diff
@@ -714,23 +723,34 @@ def render_string(
...
                 )
             )
+
+        variant_limit = config.get("render_variant_limit")
+        templated_variants: List[TemplatedFile] = []
+        templater_violations: List[SQLTemplaterError] = []
+
         try:
-            templated_file, templater_violations = self.templater.process(
+            for variant, vs in self.templater.process_with_variants(
```
WittierDinosaur, line 732: can we have a more descriptive variable name than `vs`?

alanmcruickshank: yeah - good point

src/sqlfluff/core/linter/linter.py
alanmcruickshank: yes, we could. I haven't approached the problem of deduping between variants _at all_ yet. I'm going to suggest we don't try and solve that yet - and I'll pick it up as a whole separate challenge at the end, when we can do it for linting too.

I'll add a comment in to make a note on that though, it's a really good point.

src/sqlfluff/core/linter/linter.py
```diff
@@ -303,12 +307,17 @@ def parse_rendered(
...
         tokens: Optional[Sequence[BaseSegment]]
-        if rendered.templated_file is not None:
-            tokens, lvs = cls._lex_templated_file(
-                rendered.templated_file, rendered.config
-            )
+        # TODO: We're limiting ourselves to only the first variant for now.
+        # We'll eventually parse more variants here.
+        if rendered.templated_variants:
+            _root_variant = rendered.templated_variants[0]
+            tokens, lvs = cls._lex_templated_file(_root_variant, rendered.config)
```
WittierDinosaur, line 314: `lvs` is also pretty ambiguous

alanmcruickshank: lol - yeah 🤔 . Will fix that

PR #5786, opened 2024-04-19
Small refactor in jinja templater

alanmcruickshank:
Two small method extractions to lighten up the `.process()` method before I do a bigger structural rework here. Logic is unchanged - just pulled out into two smaller functions.


src/sqlfluff/core/templaters/jinja.py
```diff
@@ -509,6 +520,70 @@ def render_func(in_str: str) -> str:
...
+        if undefined_variables:
+            # Lets go through and find out where they are:
+            for template_err_val in self._crawl_tree(
+                syntax_tree, undefined_variables, in_str
+            ):
+                violations.append(template_err_val)
+        return violations
+
+    @staticmethod
+    def _init_undefined_tracking(
```
WittierDinosaur, line 540: Broadly this refactor looks fine. Any reason to declare the UndefinedRecorder class in a function?

alanmcruickshank: It needs to depend on a local variable, but there might be a tidier way. At the very least I'll add a better comment, but I'll see if I can find a better way.

PR #5788, opened 2024-04-20
Simplify one of the lexer methods

alanmcruickshank:
Small PR:
- Removes an uncovered patch of code which handled empty token lists (because that should now never happen).
- Renames variable from `tokens` to `segments` because that's what's actually in it.
- Removes the `config` return value and comment saying it was important - because it's not.


WittierDinosaur (approved):
LGTM

PR #5794, opened 2024-04-21
Bring multiple jinja variants through to the parser.

alanmcruickshank:
This makes progress of #4061. This extends the work of #5785 by bringing multiple jinja variants through into the parser, and exposing them through to `sqlfluff parse` (albeit only the human readable version for now). It extends the `ParsedString` class to hold multiple variants and updates all the associated tests and docstrings for that to work. I've tried to keep the overall API fairly similar even though there's more going on in the background.

The results are visible in the CLI similar to `sqlfluff render`:
image
image

Right now this doesn't do any deduplication of any violations found, it also intentionally doesn't change the results for the simple API, lint any of the additional variants or support multiple variants in the non-human-format parse CLI yet. All of those can come at a later date as extensions of this work. None of this work should get in the way of those.

Next step after this will be to tackle some initial linting of the different variants.


src/sqlfluff/cli/commands.py
```diff
@@ -1364,19 +1364,21 @@ def parse(
...
+        for parsed_string in parsed_strings:
+            # TODO: Multiple variants aren't yet supported here in the non-human
+            # output of the parse command.
+            root_variant = parsed_string.root_variant()
+            if root_variant:
+                assert root_variant.tree
+                segments = root_variant.tree.as_record(
+                    code_only=code_only, show_raw=True, include_meta=include_meta
+                )
+            else:  # pragma: no cover
```
WittierDinosaur, line 1377: here too - WDYT?

WittierDinosaur: unless this is the case of an empty file - but wouldn't that still have an empty file segment?

src/sqlfluff/core/linter/common.py
```diff
@@ -34,25 +39,86 @@ class RenderedFile(NamedTuple):
...
+            raised during the lexing phase.
+    """
+
+    templated_file: Optional[TemplatedFile]
+    tree: Optional[BaseSegment]
+    lexing_violations: List[SQLLexError]
+    parsing_violations: List[SQLParseError]
+
+    def violations(self) -> List[Union[SQLLexError, SQLParseError]]:
+        """Returns the combined set of violations for this variant."""
```
WittierDinosaur, line 63:
```suggestion
        """Returns the combined lexing and parsing violations for this variant."""
```
Small edit - I know lexing and parsing are necessarily separate - but it's not a set, and at the time of writing, parsing violations are not deduped between variants

src/sqlfluff/core/linter/common.py
```diff
@@ -34,25 +39,86 @@ class RenderedFile(NamedTuple):
...
+    templating_violations: List[SQLTemplaterError]
     time_dict: Dict[str, Any]
-    templated_file: Optional[TemplatedFile]
     config: FluffConfig
     fname: str
     source_str: str
+
+    @property
+    def violations(self) -> List[SQLBaseError]:
+        """Returns the combined set of violations for this variant.
```
WittierDinosaur, line 95:
```suggestion
        """Returns the combination of violations for this variant.
```
As above

src/sqlfluff/core/linter/common.py
```diff
@@ -34,25 +39,86 @@ class RenderedFile(NamedTuple):
...
+            *self.templating_violations,
+            *(v for variant in self.parsed_variants for v in variant.violations()),
+        ]
+
+    def root_variant(self) -> Optional[ParsedVariant]:
+        """Returns the root variant if successfully parsed, otherwise None."""
+        if not self.parsed_variants:  # pragma: no cover
+            return None
+        root_variant = self.parsed_variants[0]
+        if not root_variant.tree:  # pragma: no cover
```
WittierDinosaur, line 109: Same as a couple of files up - I think these should be asserts

src/sqlfluff/api/simple.py
```diff
@@ -180,11 +185,15 @@ def parse(
...
-    if parsed.violations:
-        raise APIParsingError(parsed.violations)
+    violations = parsed.violations
+    if violations:
+        raise APIParsingError(violations)
     # Return a JSON representation of the parse tree.
-    if parsed.tree is None:  # pragma: no cover
+    # NOTE: For the simple API - only a single variant is returned.
+    root_variant = parsed.root_variant()
+    if not root_variant:  # pragma: no cover
```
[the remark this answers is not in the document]
alanmcruickshank: Good question. I think I'm going to try and build out some test cases, to either cover these situations - or _yes_ they should be asserts.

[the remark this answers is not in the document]
alanmcruickshank: Here, you're right it's not possible. I've update the code.

src/sqlfluff/cli/commands.py
alanmcruickshank: This _is_ possible. It's failing templating and failing parsing that triggers this clause. I've added better coverage.

src/sqlfluff/core/linter/common.py
alanmcruickshank: This one is possible with failed tempalting and failed parsing. I've added test cases and commands to cover both cases.

WittierDinosaur (approved):
Looks good!!

PR #5810, opened 2024-04-24
Tighten up the return from .process()

alanmcruickshank:
Related to #5794. We've been fairly inconsistent in whether we raise issues or return `None` from the templater `.process()` methods. This PR tightens that up a bit.

Previously the methods could optionally return `None`, this PR prevents that - and instead for fatal templating errors requires the process method to explicitly raise a `SQLTemplaterError` which is then caught in the linter.

While it's not a massive change - it does modify the interface for the templater slightly. The upside of all of this is that it reduces the number of different scenarios that we need to handle outside the templater so simplify processing.


WittierDinosaur (approved):
Much neater

Issue #5803, opened 2024-04-22
Placeholder parsing seems to miss some blocks


alanmcruickshank:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

While putting together #5794 I parsed:
```jinja
-- sqlfluff:render_variant_limit:10
select {% if 1 > 2 %}1{% else %}2{% endif %}
```

And got the following results:
![image](https://github.com/sqlfluff/sqlfluff/assets/4670904/0a94a668-f49a-4681-b9ea-ab6c7cea4e51)

It appears that some of the placeholder blocks are missing in both variants. I don't think this is causing too many problems, but may need investigation.

### Expected Behaviour

All the templated sections should be present

### Observed Behaviour

In the first variants, the `block_mid` section appears to be missing, and in the second variant, the `block_end` section seems to have swallowed the final placeholder.

### How to reproduce

See above

### Dialect

ansi - but likely all of them would have the same impact

### Version

On the feature branch for #5794 at commit 871701c0e9dd25bd72220e06e089e169ce121a58

### Configuration

default

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

PR #5813, opened 2024-04-25
Fix placeholder labelling

alanmcruickshank:
This fixes half of #5803, but not all of it (and it only addresses the easy bit 🤷).

Good opportunity to simplify the code a little too at the same time.


greg-finley:
Seems ok but check CI output

greg-finley (approved):

Issue #5202, opened 2023-09-13
SparkSQL dialect parses spark config names as naked identifeirs


horaceyoung:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

![image](https://github.com/sqlfluff/sqlfluff/assets/32321396/47893889-e9fb-4e12-a079-d04f1768a2a5)
![image](https://github.com/sqlfluff/sqlfluff/assets/32321396/7e675774-d319-44fa-a213-347609e43928)

sqlfluff will parse the spark config name in the following set statement as identifiers and I'm not sure why.

SET spark.sql.adaptive.maxNumPostShufflePartitions=1500; 

### Expected Behaviour

Spark config names are parsed correctly

### Observed Behaviour

Spark config names are parsed as identifiers

### How to reproduce

1. Use this as a one-line input: SET hive.mapred.mode=nonstrict;
2. Run sqlfluff fix --dialect sparksql scripts/dwd/test.sql
3. Get the following result
   ![image](https://github.com/sqlfluff/sqlfluff/assets/32321396/a62fa558-90c2-47ea-8029-aecfa241fbb4)

### Dialect

sparksql

### Version

2.3.2

### Configuration

[sqlfluff]
max_line_length = 120
dialect = sparksql
processes = 0

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

peterbud:
With the latest `sqlfluff` the above statement parser properly as `set_statement` and `property_name_identifier`

PR #5814, opened 2024-04-25
Fix Jinja variant location correction

### LINKED ISSUE #5803: Placeholder parsing seems to miss some blocks

alanmcruickshank:
This fixes #5803 . It actually corrects something from #5339, which was originally taken from #5202.

The code to correct the source slices of variant `TemplatedFileSlices` wasn't working as expected. This PR takes a totally different approach to that correction which I think is more robust.

Rather than trying to use the old locations in a lookup (which I don't think would work), this instead keeps track of the lengths of adjustments and uses them to shift the locations of the slices.

Added more test coverage to make sure it works too.

### DIFF
--- a/src/sqlfluff/core/templaters/jinja.py
+++ b/src/sqlfluff/core/templaters/jinja.py
@@ -745,7 +745,7 @@ class JinjaTemplater(PythonTemplater):
 
         max_variants_generated = 10
         max_variants_returned = 5
-        variants: Dict[str, Tuple[int, JinjaTrace]] = {}
+        variants: Dict[str, Tuple[int, JinjaTrace, Dict[int, int]]] = {}
 
         # Create a mapping of the original source slices before modification so
         # we can adjust the positions post-modification.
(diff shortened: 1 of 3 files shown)


src/sqlfluff/core/templaters/jinja.py
```diff
@@ -829,26 +837,53 @@ def _handle_unreached_code(
...
+            # :TRICKY: We want to yield variants that _look like_ they were
+            # rendered from the original template. However, they were actually
+            # rendered from a modified template, which means they have source
+            # indices which won't line up with the source files. We correct that
+            # here by using the length deltas generated earlier from the
+            # modifications.
+            # This should ensure that lint issues and fixes for the variants are
+            # handled correctly and can be combined with those from the original
+            # template.
+            delta_stack = sorted(deltas.items(), key=lambda t: t[0])
```
barrywhart, line 856: Could you add a comment explaining the purpose of the sorting, and what you're sorting on?

Given the amount of code being added here, I suggest moving this bit to a new function for readability.

barrywhart (approved):
A couple of small suggestions. TBH, this code is as opaque to me as it probably is to everyone else, so it's difficult to review and confirm it's good without getting into it super deeply. Tests are the way to go. 😅

Do you think the new test cases are sufficient? I have often found in the past that it's worth adding a more complex test, especially if the issue was discovered in "real" SQL code -- I'll add a test that's more similar to the real SQL. It may not matter _today_, but the more complex tests have often paid off in the future, detecting other issues, if that makes sense.

src/sqlfluff/core/templaters/jinja.py
alanmcruickshank: Good idea

PR #5815, opened 2024-04-25
Test cases for linter fails.

alanmcruickshank:
This is based on #5794 and #5810, and fleshes out what should happen in some of the failure cases.

This is entirely just additional coverage and clarity on exactly what should happen in each case. No logic changes.


WittierDinosaur (approved):
LGTM

PR #5856, opened 2024-05-07
Feature/MariaDB dialect

### LINKED ISSUE #5840: MariaDB into existing MySQL, or new dialect for CREATE/DROP [OR REPLACE] ?
### Search before asking

### Description

MariaDB has some small syntactic differences compared to MySQL, but for the most part should be very similar. I have run into issues where the linter throws `unparsable section` errors because of the presence of `CREATE OR REPLACE TABLE` statements, because MySQL doesn't support the OR REPLACE syntax. 

There are many more instances where parsing errors happen due to small differences between these dialects. 

Would it be preferrable to add to the existing MySQL syntax, or to create a whole new dialect specifically for the small MariaDB changes?i

### Use case

`mariadb` dialect added that borrows heavily from `mysql` but with syntax differences accounted for.

### Dialect

MySQL

### Code of Conduct

Xemptuous:

fixes #5840 by adding new MariaDB dialect based off mysql.

This PR does a few things:

- Adds core functionality for MariaDB
- Adds partitions/subpartitions and partition_option/subpartition_option to table DDL for MySQL
- Adds `REGEXP` into `LikeGrammar` for MySQL

Potentially more complex tests would be required to ensure the added sections for partitions do not affect other code in larger scripts.

### Pull Request checklist

  - Other.


WittierDinosaur:
is this meant to be in draft?

Xemptuous:
> is this meant to be in draft?

Yea I set it as a draft so I can go through other areas that MariaDB's dialect potentially differs from MySQL in a critical way, while still providing visibility.

WittierDinosaur:
The PR is already pretty massive - feel free to add to this incrementally later. The first drop doesn't need to be a full implementation

test/fixtures/dialects/mariadb/.sqlfluff
```diff
@@ -0,0 +1,2 @@
+[sqlfluff]
+dialect = mysql
```
alanmcruickshank, line 2: Shouldn't this be mariadb?

Xemptuous: Yes, I just copied all the mysql tests over to mariadb and kept them as is. Will change this as well

src/sqlfluff/dialects/dialect_mysql.py
```diff
@@ -380,14 +381,209 @@ class CreateTableStatementSegment(ansi.CreateTableStatementSegment):
     https://dev.mysql.com/doc/refman/8.0/en/create-table.html
     """
 
+    ES = Ref("ExpressionSegment")
```
WittierDinosaur, line 384: I don't love this aliasing here. There's nothing inherently wrong with the pattern. But we don't use it elsewhere, so I think it overall makes it less readable.

Xemptuous: I'm not a big fan either; it was the only way I was able to get the linter to accept the code due to column lengths without it.

Xemptuous: I agree, but in any of my limited attempts to simply break the sequences out into their own segments, I end up getting test failures. I'm pretty limited on time right now, but I'm not opposed to getting rid of the partition/subpartition options. It was just a lacking feature I thought I'd implement while I was at it, but I don't really use it personally, so I don't mind stripping it to get this PR pushed through. Lmk your thoughts

src/sqlfluff/dialects/dialect_mariadb_keywords.py
```diff
@@ -0,0 +1,706 @@
+"""A List of MariaDB SQL keywords.
+
+https://mariadb.com/kb/en/reserved-words/
+"""
```
pprkut, line 4: I added slightly more documentation to my list of reserved keywords, especially where to get the unreserved keywords from. Feel free to steal it :)

https://github.com/pprkut/sqlfluff/blob/mariadb-dialect/src/sqlfluff/dialects/dialect_mariadb_keywords.py#L1

src/sqlfluff/dialects/dialect_mariadb_keywords.py
```diff
@@ -0,0 +1,706 @@
...
+WEIGHT_STRING
+WITHIN
+WITHOUT
+WORK
+WRAPPER
+X509
+XA
+XML
+YEAR
+"""
```
pprkut, line 706: The `mysql` dialect has a couple extra "keywords" at the end: https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/dialects/dialect_mysql_keywords.py#L784

Are you sure those aren't needed, given that this extends the mysql dialect?

pprkut (commented):
I don't know what the policy is for test fixtures, but is there a need to copy everything from the mysql dialect? IMHO, just adding fixtures for the features that differ from the mysql dialect should be enough, no?

WittierDinosaur (approved):
This is functional, and working. LGTM. I don't love the shorthand in the mysql dialect, but that's cosmetic, so I won't hold this up. We can always break this down more in the future.

WittierDinosaur:
@Xemptuous ready for this to go, but you've got some yml that needs regenerating after a recent change that was merged ahead of yours

src/sqlfluff/dialects/dialect_mariadb_keywords.py
Xemptuous: Thanks! Adding in upcoming commit :)

src/sqlfluff/dialects/dialect_mariadb_keywords.py
Xemptuous: They weren't listed in the mariadb repo or docs as keywords, so I don't know if they're supported or not. Might be worth exploring, as I've never seen or used some of these (e.g., `Zone` or `XID`)

PR #5872, opened 2024-05-13
Support variables with dot in python templater

### LINKED ISSUE #5871: Allow variables with dot in sqlfluff:templater:python:context
### Search before asking

### Description

**SQLFluff python templater does not currently support variables that contain a dot - e.g. `{foo.bar}`**

This SQL:
```sql
SELECT * FROM {foo.bar}
```

..with this config:
```cfg
[sqlfluff:templater:python:context]
foo.bar = foobar
```

..raises this error:
```python
sqlfluff.core.errors.SQLTemplaterError: Failure in Python templating: 'foo'. Have you configured your variables? https://docs.sqlfluff.com/en/stable/configuration.html#python-templater
```

Related tickets (but for Jinja templater):
* https://github.com/sqlfluff/sqlfluff/issues/4153
* https://github.com/sqlfluff/sqlfluff/issues/4019

### Use case

Python format strings support variables that use dot notation for object attribute access.  This is a valid Python example that is not yet supported by SQLFluff:

```python
class MyObject:
    bar = "blah"


sql = "SELECT * FROM {foo.bar}"
context = {"foo": MyObject()}
print(sql.format(**context))
```

### Dialect

This is not dialect specific.

### Code of Conduct

timchurch:

fixes #5871 

Adds support for python format strings containing variables with dot notation by temporarily converting these variables into a dictionary lookup using the fixed context key `sqlfluff`.  

I believe this is the simplest, most straightforward solution to this issue.  Adding this "magic" context key into configuration might not be intuitive at first, but I think it can be easily understood through the documentation.

No

### Pull Request checklist

  - Other.


docs/source/configuration.rst
```diff
@@ -993,6 +993,34 @@ If passed the following *.sql* file:
...
 
+Complex Python Variable Templating
+""""""""""""""""""""""""""""""""""""
+
+`Python string formatting`_ supports accessing object attributes
+via dot notation (e.g. :code:`{foo.bar}`).  However, since we cannot create Python
+objects within configuration files, we need a workaround in order to provide
+dummy values to render templates containing these values.  The SQLFluff
+python templater will temporarily convert any template variable
+containing a "." with a dictionary lookup using the fixed context key :code:`sqlfluff`:
```
alanmcruickshank, line 1004: I think the use of `temporarily` here is misleading. Perhaps this might be clearer:

```suggestion
dummy values to render templates containing these values.  The SQLFluff
python templater will interpret any variable containing a "." as a
dictionary lookup on the *magic* fixed context key :code:`sqlfluff`.
```

src/sqlfluff/core/templaters/python.py
```diff
@@ -249,9 +250,23 @@ def process(
...
+                raw_str_with_dot_notation_hack = re.sub(
+                    r"{([^:}]*\.[^:}]*)(:\S*)?}", r"{sqlfluff[\1]\2}", raw_str
+                )
+                templater_logger.debug(
+                    "    Raw String with Dot Notation Hack: %r",
+                    raw_str_with_dot_notation_hack,
+                )
+                rendered_str = raw_str_with_dot_notation_hack.format(**live_context)
             except KeyError as err:
                 raise SQLTemplaterError(
```
alanmcruickshank, line 271: @timchurch - can you check how your solution _fails_ if the user _has_ a variable containing a dot (e.g. `{foo.bar}`), but _hasn't defined the sqlfluff dict object at all_?

I think we need to account for that failure mechanism here and provide a hint to the user on how to resolve it. Especially given this will be a new feature, there's a chance that people might not know it's an option unless it's in an error message.

We might want to consider adding an _empty_ dictionary to the default config if that changes the behaviour, so that there's _always_ a `sqlfluff` value present even if the user hasn't set one. That also provides an opportunity to link to the docs in the default config. I'm less confident on this last point though, I think it depends more how the code fails.

alanmcruickshank (changes requested):
Initially I didn't like this hack, but I think I'm warming to it.

It's definitely a hack, but it's a convenient one, and I think a pragmatic one. Good work 👍 .

I've got a couple of notes on how to make this clearer and easier for the user to understand. Likewise, making it clear when it _fails_ so that the user knows what's happening.

alanmcruickshank:
Oh and on the merge conflict, I split out a lot of the very large config files as part of #6052 .

The file that your edits should now go in is `/docs/source/configuration/templating/python.rst`. The actual documentation is almost identical to before (and the python section is just a copy paste of that section of the old `configuration.rst` file) - just in a different place.

src/sqlfluff/core/templaters/python.py
timchurch: This is the current error message if sqlfluff dict object is missing from context:

```python
sqlfluff.core.errors.SQLTemplaterError: Failure in Python templating: 'sqlfluff'. Have you configured your variables? https://docs.sqlfluff.com/en/stable/configuration.html#templating-configuration
```

But I agree that this is confusing for users and worth improving.  I'll add a custom error message here.

timchurch: I've updated the error messages but still open to suggestions.

Here is the initial error message if `sqlfluff` key does not exist:
```
sqlfluff.core.errors.SQLTemplaterError: Failure in Python templating: magic key 'sqlfluff' missing from context.  This key is required for template variables containing '.'. https://docs.sqlfluff.com/en/stable/perma/python_templating.html
```

And here is the error message if `sqlfluff` key does exist but is missing an individual key:
```
sqlfluff.core.errors.SQLTemplaterError: Failure in Python templating: 'foo.bar' key missing from 'sqlfluff' dict in context. Template variables containing '.' are required to use the 'sqlfluff' magic fixed context key. https://docs.sqlfluff.com/en/stable/perma/python_templating.html
```

timchurch: > We might want to consider adding an empty dictionary to the default config if that changes the behaviour, so that there's always a sqlfluff value present even if the user hasn't set one. That also provides an opportunity to link to the docs in the default config. I'm less confident on this last point though, I think it depends more how the code fails.

I'm also open to this.  I briefly tested and seems easy to implement.  However, I'm not sure if it actually helps very much because end user will still need to define this key in their own context anyway for templating to work.

alanmcruickshank (approved):
Nice good work. I've added a test case to cover your new exception messages 👍 .

Sorry for the delayed review here.

PR #5880, opened 2024-05-14
Support quoted psql parameters with `placeholder` templater

### LINKED ISSUE #5140: psql variable support in Postgres dialect
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues. (open issues at least)

### What Happened

There is a previous issue, #1534, but it was closed by PR #1709 which seems to add the desired behavior. However, I am still having issues, hence labeling this a bug rather than enhancement

psql expects variables to be interpolated in sql files in one of three formats
- `:var_name`
- `:'var_name'`
- `:"var_name"`

Per [psql docs](https://www.postgresql.org/docs/current/app-psql.html#APP-PSQL-INTERPOLATION). This seems to be what was implemented in #1709, but when linting a file with these variables, I get errors. The following should be valid postgresql, if we're allowing psql variables, then the query
```sql
SELECT :"column" FROM :table WHERE bla = :'my_name'
```
should be valid.

It seems that there might be some overlap between a solution here in the postgres dialect and a solution using "placeholder" template. However, no default `param_style` allows for these types of variables with quotes, only unquoted. I've implemented a solution in [my fork](https://github.com/fvankrieken/sqlfluff/commit/c66b841d0c6bca7016487aab9e157455c05fc7fc) to add a `param_style` case, while locally for now I'm just using the "placeholder" template with the regex in that commit. I'd be happy to open a PR from my fork if it's decided this should be handled via `templater`s rather than as part of the postgres dialect, but upon digging realized that maybe this specific case should already be part of the postgres dialect. Though looking at #3343, it seems like last time something like this came up, it was fixed through a `templater`. I'm still just curious as to why #1709 did not implement this by default for postgres dialect.

Or maybe it's working fine for the dialect, and I'm just missing how to enable it in my `.sqlfluff` file! If it's that, please let me know

### Expected Behaviour

I would expect the above example to be linted with no error

### Observed Behaviour

`sqlfluff lint test.sql --dialect=postgres` outputs
```
== [test.sql] FAIL                                                                                                                                                                                          
L:   1 | P:   7 | LT01 | Unexpected whitespace before colon ':'.
                       | [layout.spacing]
L:   1 | P:   9 | LT01 | Expected single whitespace between colon ':' and
                       | parameter. [layout.spacing]
L:   1 | P:  17 |  PRS | Line 1, Position 17: Found unparsable section: " FROM
                       | :table WHERE bla = :'my_name'"
WARNING: Parsing errors found and dialect is set to 'postgres'. Have you configured your dialect correctly?
All Finished 📜 🎉!
```

### How to reproduce

Lint the above `SELECT :"column" FROM :table WHERE bla = :'my_name'` with the  command `sqlfluff lint test.sql --dialect=postgres`

### Dialect

postgres

### Version

2.3.0

### Configuration

No `.sqlfluff` used for above example, just specifying postgres as dialect via command line.

### Code of Conduct

fvankrieken:
Closes #5140, though that issue has a bit of maybe a misunderstanding of the problem

psql sql interpolation supports colon followed by variable name, unquoted, single-quoted, or double-quoted, such as
```sql
 SELECT :"column" FROM :table WHERE user = :'name'
```

When invoked with `psql -c "SELECT ..." -v column=columnWithUppercase -v table=my_table -v user=my_name` renders to
```sql
SELECT "columnWithUppercase" FROM my_table WHERE user = 'my_name'
```

This implements a new `param_style` for `placeholder` templater to allow for this. If a name more specific to its use case (rather than a description of the style) would be useful, like `psql_colon`, let me know

It does add a new capture group to the regex, and looks for this capture group in the generic placeholder code. Shouldn't cause issue now, but it is a bit of a deviation in code style and maybe deserves a more abstracted implementation

### Pull Request checklist
**I believe so but let me know if more documentation or testing is needed**

  - Other.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -896,6 +896,9 @@ A few common styles are supported:
     -- (use with caution as more prone to false positives)
     WHERE bla = table:my_name
 
+    -- colon_optional_quotes
+    SELECT :"column" FROM :table WHERE bla = :'my_name'
+
     -- numeric_colon
     WHERE bla = :2
 
--- a/src/sqlfluff/core/templaters/placeholder.py
+++ b/src/sqlfluff/core/templaters/placeholder.py
@@ -21,6 +21,10 @@ templater_logger = logging.getLogger("sqlfluff.templater")
 KNOWN_STYLES = {
     # e.g. WHERE bla = :name
     "colon": regex.compile(r"(?<![:\w\x5c]):(?P<param_name>\w+)(?!:)", regex.UNICODE),
+    # e.g. SELECT :"column" FROM :table WHERE bla = :'name'
+    "colon_optional_quotes": regex.compile(
+        r"(?<!:):(?P<quotation>['\"]?)(?P<param_name>[\w_]+)\1", regex.UNICODE
+    ),
     # e.g. WHERE bla = table:name - use with caution as more prone to false positives
     "colon_nospaces": regex.compile(r"(?<!:):(?P<param_name>\w+)", regex.UNICODE),
     # e.g. WHERE bla = :2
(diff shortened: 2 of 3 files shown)


WittierDinosaur (approved):
LGTM

PR #5924, opened 2024-05-30
Clarify docs around subdir handling when loading macros

james-johnston-thumbtack:

The documentation erroneously states that SQLFluff will not look in subfolders when loading macros using the load_macros_from_path config. This is not actually the case: the JinjaTemplater._extract_macros_from_path function will recursively search subdirectories, and load those macros into the Jinja global namespace as well:

https://github.com/sqlfluff/sqlfluff/blob/06b22f56a89043683915886a63d19aa34a4f63ec/src/sqlfluff/core/templaters/jinja.py#L169-L177

This commit does the following:

1.  Amends one of the existing test cases to demonstrate that macros can be loaded from subdirectories.  No obvious existing test case seemed to meaningfully cover this scenario.

2.  Updates the documentation to state that subdirectories will be searched as well.

There is no actual change in behavior of the product.  To the end user, this is a doc correction.

No

### Pull Request checklist

  - Other.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -595,31 +595,31 @@ loaded from files or folders. This is specified in the config file:
 .. code-block:: cfg
 
     [sqlfluff:templater:jinja]
-    load_macros_from_path = my_macros
+    load_macros_from_path = my_macros,other_macros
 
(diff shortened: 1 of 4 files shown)


WittierDinosaur (approved):
LGTM

PR #5930, opened 2024-05-30
Add loader_search_path setting to Jinja templater

### LINKED ISSUE #4699: Add special option to set only the Jinja search paths
### Search before asking

### Description

The `load_macros_from_path` Jinja templater configuration option controls two things:

1. The Jinja search path for the `FileSystemLoader`: https://github.com/sqlfluff/sqlfluff/blob/7f53c8b8eb0432a9404aaa65e6ec9c0dc2152ff9/src/sqlfluff/core/templaters/jinja.py#L269
2. Parsing every single SQL file in the directory to find any macros and include them in the context, so users don't have to include them themselves: https://github.com/sqlfluff/sqlfluff/blob/7f53c8b8eb0432a9404aaa65e6ec9c0dc2152ff9/src/sqlfluff/core/templaters/jinja.py#L323-L329

Both of these behaviors are documented at https://docs.sqlfluff.com/en/stable/configuration.html#macro-templating - so this is no complete surprise.

However, for users who _want_ every import to be explicitly declared, I think the behavior of (2) is undesirable:

- It hides any missing import directives.
- It is extremely slow when the search path has many files.

Proposal: I would like to decouple these behaviors by having separate configuration options.  For example, we could have a new `loader_search_paths` option that is used _only_ for the Jinja searchpath, and is not otherwise searched for macros.  if both `loader_search_paths` and `load_macros_from_path` are present, then the former could override the latter for purposes of the searchpath, while the latter is still used for loading macros into the context.

I would be willing to try my hand at a PR for this, if we can reach alignment on what the behavior changes / configuration options should be.

### Use case

For example, suppose I set up a simple SQL project organization like this:

- When I load Jinja templates in my application, I set FileSystemLoader to the root of the project's SQL files.
- When I need to use a macro in another SQL file, I will always use an `import` statement.
- In order to lint this project, I set the options in the root directory as follows:

```
[sqlfluff:templater:jinja]
load_macros_from_path = .
```
and run the linter with `sqlfluff lint .` from that root directory to lint all SQL files in the project.

This works, but it will unfortunately lead to really slow performance: O(N^2) because effectively the looping looks like this:
```
for each file in the project:
    context_macros = []
    for each file in the project:
        context_macros.append(load_macros_from_file(file))
    file.render(context_macros)
```
It also makes us more exposed to errors from unrelated files: even if another file is not imported, it can still cause a failure when this loop runs.

### Dialect

Jinja templater

### Code of Conduct

james-johnston-thumbtack:

Currently, in order to include a folder in the Jinja loader search path, one must specify it using the load_macros_from_path setting.  However, this setting carries a side effect, which may be unwanted in some scenarios: it will also load every single file in that directory (which is slow), and loads those macros into the global namespace (which may not be an accurate representation of the end-user application behavior). There is currently no option to avoid this additional behavior.

This commit adds a new loader_search_path setting, which is a simple way of specifying the Jinja loader's search path.  The behavior of the existing load_macros_from_path setting remains unchanged.  If both options are set, all directories are merged and included in the Jinja loader search path.

Further details of the change can be found in the updated documentation.

Fixes #4699 

None, this should be completely backwards compatible.

### Pull Request checklist

  - Other.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -497,6 +497,7 @@ options:
     [sqlfluff:templater:jinja]
     apply_dbt_builtins = True
     load_macros_from_path = my_macros
+    loader_search_path = included_templates
     library_path = sqlfluff_libs
 
     [sqlfluff:templater:jinja:context]
@@ -620,6 +621,10 @@ file :code:`my_macros/subdir/my_file.sql`, you can do:
 .. code-block:: jinja
 
    {% include 'subdir/my_file.sql' %}
+
+If you would like to define the Jinja search path without also loading the
+macros into the global namespace, use the :code:`loader_search_path` setting
+instead.
 
 .. note::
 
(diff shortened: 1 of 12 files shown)


alanmcruickshank (approved):
Very nice! Cool feature. And thanks for the excellent docs.

PR #5937, opened 2024-06-03
Add support for custom JinjaTracer implementations

james-johnston-thumbtack:

Currently, the JinjaAnalyzer implementation will directly construct an instance of JinjaTracer, making it difficult for users providing a derived JinjaAnalyzer to also have it return a derived JinjaTracer.

By moving the construction of JinjaTracer to a separate function, it is easy for plugin authors to override this function and provide their own tracer.

This approach is very similar to the one taken for allowing custom JinjaAnalyzer implementations in pull request #5543.

No.  This is a small refactor that makes it easier to write derived classes.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/templaters/slicers/tracer.py
+++ b/src/sqlfluff/core/templaters/slicers/tracer.py
@@ -14,7 +14,6 @@ import regex
 from jinja2 import Environment
 from jinja2.exceptions import TemplateSyntaxError
 
-from sqlfluff.core.config import FluffConfig
 from sqlfluff.core.templaters.base import RawFileSlice, TemplatedFileSlice
 
 # Instantiate the templater logger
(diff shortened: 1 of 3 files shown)


src/sqlfluff/core/templaters/slicers/tracer.py
```diff
@@ -51,6 +52,7 @@ def __init__(
         raw_slice_info: Dict[RawFileSlice, RawSliceInfo],
         sliced_file: List[TemplatedFileSlice],
         render_func: Callable[[str], str],
+        config: Optional[FluffConfig] = None,
```
alanmcruickshank, line 55: Do we need to pass the whole config in here? I'm not necessarily against it, but it seems unnecessary at the moment. Are you doing something interesting with it in the derived case?

james-johnston-thumbtack: Actually no.... not currently using it in the derived case (and don't currently expect to need to).  This was more of a future-proofing move.  I can remove it.

james-johnston-thumbtack: Added on a commit that removes the config parameters, including from the _get_jinja_analyzer that I added earlier this year but don't really need it there either.

WittierDinosaur (approved):
LGTM

PR #5938, opened 2024-06-03
Don't indent invisible template slices

james-johnston-thumbtack:

Some custom templater plugins might emit special zero-length marker template slices with a special slice type, as previously described in #4708.  These are transformed into a meta TemplateSegment with an empty source string, which mostly works great.  However, the LT02 indention rule will attempt to indent these meta nodes, which is impossible because there is no `source_str` that the indention would affect, and thus it is impossible to resolve this lint violation.

For example, suppose the parse tree looks like this:

```
[META] placeholder: [Type: 'block_start', Raw: "{% up 'Do nothing' %}", Block: 'f142fe']
[META] indent:      [Block: 'f142fe']
[META] placeholder: [Type: 'migration_up', Raw: '']
newline:            '\n'
whitespace:         '    '
```

The indention rule will perceive a violation and try to fix it by inserting a newline and whitespace prior to the second placeholder:

```
[META] placeholder: [Type: 'block_start', Raw: "{% up 'Do nothing' %}", Block: 'f142fe']
[META] indent:      [Block: 'f142fe']
newline:            '\n'
whitespace:         '    '
[META] placeholder: [Type: 'migration_up', Raw: '']
newline:            '\n'
whitespace:         '    '
```

However, this second placeholder doesn't have any source string.  When the file is reparsed, the templater will again result in a parse tree as seen in the original snippet above, and thus the violation is effectively unfixable.

This commit fixes the issue by treating these cases identically to trailing comments at the end of a line.  For example, the following template has a trailing comment that delays the indention requirement:

```
{% if True %} -- I don't need to put a trailing comment on the next line
    SELECT 1;
```

By treating the zero-length marker placeholder the same as a trailing comment, we can avoid trying to indent them.

Note that this commit also adds test cases to the `reindent_test` for trailing comments.  Previously, there was no direct test coverage for that branch of code.

No, there shouldn't be.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/utils/reflow/reindent.py
+++ b/src/sqlfluff/utils/reflow/reindent.py
@@ -825,7 +825,17 @@ def _crawl_indent_points(
 
             # Is the next element a comment? If so - delay the decision until we've
             # got any indents from after the comment too.
-            if "comment" in elements[idx + 1].class_types:
+            #
+            # Also, some templaters might insert custom marker slices that are of zero
+            # source string length as a way of marking locations in the middle of
+            # templated output.  These don't correspond to real source code, so we
+            # can't meaningfully indent before them.  We can safely handle them similar
+            # to the comment case.
+            if "comment" in elements[idx + 1].class_types or (
+                "placeholder" in elements[idx + 1].class_types
+                and cast(TemplateSegment, elements[idx + 1].segments[0]).source_str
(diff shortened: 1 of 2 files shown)


WittierDinosaur (approved):
LGTM

PR #5958, opened 2024-06-09
Parent dir config search

### LINKED ISSUE #5957: Support project configuration
### Search before asking

### Description

I propose to let sqlfluff search parent directories of the current working directory recursively to find a configuration, like other linting tools such as flake8 does.
Currently when invoking the sqlfluff cli it only looks for configuration files between the current working dir and the directory the target file is located (in addition to standard user locations).
It is common to have linting configuration at the root directory of a given project, but if working inside a sub directory sqlfluff does not pick up this configuration.
Putting the configuration in global locations such as `~/.config/sqlfluff` does not work when different projects use different config. For instance if one has DBT in one project, but writes raw queries in another, the same configuration would not work for both.

### Use case

I would like to be able to use the sqlfluff cli when working inside sub directories, rather than only being able to invoke it from the root directory of a project. This will also make it easier to work with linting tools in editors, which typically just executes the linter in the same directory as the sql file.

### Dialect

This is for the cli, so it is not dialect specific.

### Code of Conduct

j-svensmark:
When searching for configuration files, this PR adds a step that looks for configuration files between the users home directory and the current working directory. This makes it possible to execute sqlfluff in subdirectories of a project where the sqlfluff config is in the root of the project, making it simpler to configure linting in editors.
Fixes #5957 (at least its a possible solution)

## Open questions
- Would it be better to but this feature behind a feature or cli flag (i.e. make it not default)?
- In this PR, if there are several directories between home and cwd that contains a config, it will only take the config closest to cwd, and ignore all other configs. Wondering if it should just include all of them?

### Pull Request checklist

- Included test cases to demonstrate any code changes in `test/core/config_test.py`.

### DIFF
--- a/src/sqlfluff/core/config.py
+++ b/src/sqlfluff/core/config.py
@@ -685,6 +685,17 @@ class ConfigLoader:
             self.load_user_appdir_config() if not ignore_local_config else {}
         )
         user_config = self.load_user_config() if not ignore_local_config else {}
+        parent_config_paths = (
+            iter_intermediate_paths(Path(path).absolute(), Path(os.path.expanduser("~")))
+            if not ignore_local_config
+            else {}
+        )
+        parent_config_stack = (
+            [self.load_config_at_path(p) for p in reversed(list(parent_config_paths))]
+            if not ignore_local_config
+            else []
+        )
+        parent_config_stack = [config for config in parent_config_stack if config][:1]
         config_paths = (
             iter_intermediate_paths(Path(path).absolute(), Path.cwd())
             if not ignore_local_config
@@ -699,7 +710,7 @@ class ConfigLoader:
             self.load_extra_config(extra_config_path) if extra_config_path else {}
         )
         return nested_combine(
-            user_appdir_config, user_config, *config_stack, extra_config
+            user_appdir_config, user_config, *parent_config_stack, *config_stack, extra_config
         )
 
     @classmethod
--- a/src/sqlfluff/core/config.py
+++ b/src/sqlfluff/core/config.py
@@ -685,27 +685,33 @@ class ConfigLoader:
             self.load_user_appdir_config() if not ignore_local_config else {}
         )
         user_config = self.load_user_config() if not ignore_local_config else {}
-        parent_config_paths = (
-            iter_intermediate_paths(Path(path).absolute(), Path(os.path.expanduser("~")))
-            if not ignore_local_config
-            else {}
-        )
-        parent_config_stack = (
-            [self.load_config_at_path(p) for p in reversed(list(parent_config_paths))]
-            if not ignore_local_config
-            else []
-        )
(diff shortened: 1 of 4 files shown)


src/sqlfluff/core/config.py
```diff
@@ -685,21 +685,38 @@ def load_config_up_to_path(
...
+            parent_config_paths = list(
+                iter_intermediate_paths(Path(path).absolute(),
+                                        Path(os.path.expanduser("~")))
+            )
+            # Stripping off the home directory and the current working
+            # directory, since they are both covered by other code
+            # here
+            parent_config_paths = parent_config_paths[1:-1]
+            parent_config_stack = (
+                [self.load_config_at_path(p) for p in reversed(list(parent_config_paths))]
```
j-svensmark, line 704: This actually loads all configs it finds, but in the next step we only keep the first one. Maybe this can be more efficient by having an explicit for loop with a break statement? 🤔

WittierDinosaur (changes requested):
I'm generally in favour of this change, though I think we'll want a minor version bump for it. The only thing I'm concerned about is that it should produce the same result regardless of the depth of the current working directory. As such, I think we'd want to load configs in the same way. Lower dir overrides higher dir, but not just the first config. So we'd probably want all to load in reverse order

j-svensmark:
> I'm generally in favour of this change, though I think we'll want a minor version bump for it. The only thing I'm concerned about is that it should produce the same result regardless of the depth of the current working directory. As such, I think we'd want to load configs in the same way. Lower dir overrides higher dir, but not just the first config. So we'd probably want all to load in reverse order

Let me just try to clarify with an example.
Suppose we have the following folder structure
```
~ (home)
└── A
    ├── .sqlfluff
    └── B
        ├── .sqlfluff
        └── C (cwd)
            ├── .sqlfluff
            └── D
                ├── .sqlfluff
                └── E
                    ├── .sqlfluff
                    └── query.sql
```
Lets say the folder C is the current working directory (cwd). In this case, we have the following possible behaviours.
1. Before this PR. Configuration is determined by `.sqlfluff` in `[C, D, E]` dirs, configurations to the right overvwiriting those to the left.
2. The current content of this PR: `[B, C, D, E]`
3. Including all dirs up to home (~): `[A, B, C, D, E]`

Now suppose we change the cwd to `D`. We then get
1. `[D, E]`
2. `[C, D, E]`
3. `[A, B, C, D, E]`

Your argument above is that its best if the total configuration  is unchanged by such a cwd change. I agree, I can try to modify the PR along the lines of 3. later.

alanmcruickshank (approved):
Looks good to me. I think the config loading logic is going to need an overhaul at some point but this moves us in a positive direction. 👍

Thanks for taking the time to make the edits.

WittierDinosaur (approved):

PR #5965, opened 2024-06-15
Postgres: Fix multiline concat for special literals

### LINKED ISSUE #5929: False positive for comment with escape character E''
### Search before asking

### What Happened

Obtained false positive for parsing error on a comment using escape character E''

### Expected Behaviour

Parse as valid SQL.

### Observed Behaviour

Parsing error reported.

### How to reproduce

sqlfluff parse test.sql --dialect postgres

where test.sql is:
COMMENT ON TABLE "some_table" IS E''
	'This is a valid comment style'
	'\n\n'
	'but it is flagged as parsing error';

### Dialect

postgres

### Version

3.0.5

### Configuration

None.

### Code of Conduct

keraion:

Fixed Postgres multiline concatenation for escaped, bit, and unicode strings
- Fixes #5929

None, but also unsure if `dollar_quote` multiline concatenation is valid and should be removed.

### Pull Request checklist

  - Other.


WittierDinosaur (changes requested):
I think the proper way to handle this is to change the regex to allow for these literals to be multiline, WDYT?

keraion:
I initially agreed with the general thought of that approach, but I believe it has a few issues that are handled by the current implementation. I'm not sure if there is a clever way of using the subdivider to split out the newlines, comments, and whitespace or something else to handle these.

First issue is inline comments. Below is a completely valid way to write a string. Maybe we want to keep the comment embedded, but maybe not.

```sql
SELECT 
e'da' --this is a comment
'ta';
```

The second issue is determining the difference between string's value of newlines and whitespace or the multi-line's newlines and whitespace. One of these we'd like to be able to format and handle layout changes, the other is the value of the data.

```sql
SELECT 
e'value of newline here:
'
    'space '
'no'
'space';
```

That being said, there are also a number of issues with the current implementation that will need a second look once we choose a direction.
- `unicode_single_quote` and `unicode_double_quote` should support `u` or `U`, same for `uescape`'s case
- the multiline `unicode_single_quote` doesn't allow for `uescape`
- `escaped_single_quote` should support `[Ee]`

alanmcruickshank:
I think this approach makes sense. I agree it might have been cleaner to merge them into a single string in the parsing phase, but that might actually make layout rules harder - this way we can handle the layout rules appropriately.

@WittierDinosaur - what do you reckon?

WittierDinosaur (approved):
I'm not majorly keen on these being captured as multiple elements, but Alan seems to be. Either way, this is a positive change, and if it causes issues in the future, we can always built on it. So LGTM

PR #6010, opened 2024-07-15
Fix ignored inline rule overrides (#5697)

### LINKED ISSUE #5697: exclude_rules override in files are not used
### Search before asking

### What Happened

When setting `sqlfluff:exclude_rules` in a file override, the override is not honored, for example, putting:

```sql
-- sqlfluff:exclude_rules:CP02
```

### Expected Behaviour

The rule is excluded during the file. According to the docs, this is the preferred way to do file-specific overrides for rules.

> When considering configuration inheritance, each of rules and exclude_rules will totally overwrite any values in parent config files if they are set in a child file. While the subtraction operation between both of them is calculated “per file”, there is no combination operation between two definitions of rules (just one overwrites the other).

The exclude_rules is overridden for that file. 

### Observed Behaviour

The rule is not excluded for the file. This is in contrast to other in-file overrides, which take effect for only the current file.

### How to reproduce

Using the following example:

```toml
# config.toml
[sqlfluff]
dialect = ansi

capitalisation_policy = lower
```

```sql
-- query.sql
-- sqlfluff:exclude_rules:CP02
-- sqlfluff:rules:capitalisation.keywords:capitalisation_policy:upper

SELECT A FROM B.C;
```

We expect that running `sqlfluff lint --config config.toml query.sql` would result in no errors, as the file overrides:

- CP02 (capitalisation.identifiers) is disabled (identifiers can be any case)
- keywords capitalization is set to upper

In actuality, linting fails, and only part of the overrides are honored:

```bash
== [query.sql] FAIL                                                            
L:   5 | P:   8 | CP02 | Unquoted identifiers must be lower case.
                       | [capitalisation.identifiers]
L:   5 | P:  15 | CP02 | Unquoted identifiers must be lower case.
                       | [capitalisation.identifiers]
L:   5 | P:  17 | CP02 | Unquoted identifiers must be lower case.
                       | [capitalisation.identifiers]
All Finished 📜 🎉!
```

This shows that just the rules exclusion was skipped.

### Dialect

ansi

### Version

sqlfluff, version 3.0.2
Python 3.8.10

### Configuration

[sqlfluff]
dialect = ansi

capitalisation_policy = lower

### Code of Conduct

alesbukovsky:

Fixes #5697

- The code parsing comma-separated input in the `FluffConfig` class constructor taken out into a separate internal method. This method is called from the `FluffConfig` class constructor to preserve the original use. In addition, it is now also called when inline directive processing on a file is done to ensure the comma-separated lists are exploded into internal variables.

- While this fixes the encountered issue, it does so within the existing facilities for the inline configuration directive parsing. These are however implemented differently from the rest of the configuration loading - the former as a set of method on `FluffConfig`, the latter via a dedicated `ConfigLoader` functions. It is possible that the correct way to fix the problem is to re-implement the inline directive handling via the `ConfigLoader`. However, I am not familiar enough with the code base to make this call.

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/config.py
+++ b/src/sqlfluff/core/config.py
@@ -796,22 +796,7 @@ class FluffConfig:
             False if self._configs["core"].get("nocolor", False) else None
         )
         # Handle inputs which are potentially comma separated strings
-        for in_key, out_key in [
-            # Deal with potential ignore & warning parameters
-            ("ignore", "ignore"),
-            ("warnings", "warnings"),
-            ("rules", "rule_allowlist"),
-            # Allowlists and denylistsignore_words
-            ("exclude_rules", "rule_denylist"),
-        ]:
-            if self._configs["core"].get(in_key, None):
-                # Checking if key is string as can potentially be a list to
-                self._configs["core"][out_key] = split_comma_separated_string(
-                    self._configs["core"][in_key]
-                )
-            else:
-                self._configs["core"][out_key] = []
-
+        self._handle_comma_separated_values()
         # Dialect and Template selection.
         dialect: Optional[str] = self._configs["core"]["dialect"]
         self._initialise_dialect(dialect, require_dialect)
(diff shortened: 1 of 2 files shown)


alanmcruickshank (approved):
NICE! Good work. Thanks for contributing this 👍

PR #6031, opened 2024-07-29
Exclude Macros from a path provided

### LINKED ISSUE #6020: Add ability to exclude macros from a path
### Search before asking

### Description

It might be worth it to add a new config `exclude_macros_from_path` within `[sqlfluff:templater:jinja]`. 
This would allow a user to pass in a path of macros that the user wants sqlfluff to ignore.

Similar to issue[ #381](https://github.com/sqlfluff/sqlfluff/issues/381)

### Use case

I am currently using the jinja templater but have additional jinja tags from our dbt project but I do not want to use the dbt templater because of having it reach out to the database to compile the queries.

### Dialect

snowflake and jinja templater

### Code of Conduct

culpgrant:
fixes #6020 
This allows a user to pass in a list of paths in the new config parameter called `exclude_macros_from_path`. Sqlfluff will then ignore the macros in those paths. This helps resolve the issue if a user has macros with custom tags in them.

This is my first contribution so if I did anything wrong please let me know.

### Pull Request checklist

  - Other.


test/fixtures/templater/jinja_exclude_macro_path/macros/macros_exclude/macro_exclude.sql
```diff
@@ -0,0 +1,3 @@
+{% materialization my_materialization_name, default %}
+ -- materialization...
+{% endmaterialization %}
```
alanmcruickshank, line 3: I understand why you're adding this file, to check that the jinja templater doesn't error with the custom tag `materialization`. Can you add a comment in this file to document that?

test/fixtures/templater/jinja_exclude_macro_path/macros/macros_exclude/macro_exclude_2.sql
```diff
@@ -0,0 +1,3 @@
+{% materialization my_materialization_name_2, default %}
+ -- materialization...
+{% endmaterialization %}
```
alanmcruickshank, line 3: Rather than a second `materialization` example here, could you change the second file in this folder to be something which overrides `foo1` and `foo2`, for example something like:

```jinja
{%- macro foo1() -%}105{%- endmacro -%}
{%- macro foo2() -%}106{%- endmacro -%}
```

...to further demonstrate that this folder is not overriding the others?

test/fixtures/templater/jinja_exclude_macro_path/macros/macro_include.sql
```diff
@@ -0,0 +1,3 @@
+{%- macro foo1() -%}101{%- endmacro -%}
+
+{%- macro foo2() -%}102{%- endmacro -%}
```
alanmcruickshank, line 3: Can you change these macros to be different from the ones in the `.sqlfluff` to demonstrate the precedence between competing definitions? Potentially consider removing the definitions from the `.sqlfluff` file to make this a cleaner test of just the overrides.

test/core/parser/parse_test.py
```diff
@@ -48,3 +49,15 @@ def test__parser__parse_error():
         "after <WordSegment: ([L:  1, P:  1]) 'SELECT'>. "
         "Found nothing."
     ) in parsed.tree.stringify()
+
+
+def test_parse_jinja_macro_exclude():
+    """Test parsing when excluding macros with unknown tags."""
```
alanmcruickshank, line 55: Can we be a little more explicit here, just so it's really obvious what this test is about?

```suggestion
    """Test parsing when excluding macros with unknown tags.

    This test case has a file which defines the unknown tag `materialization` which would
    cause a templating error if not excluded. By ignoring that folder we can ensure there
    are no errors.
    """
```

test/core/linter/linter_test.py
```diff
@@ -655,3 +655,13 @@ def test_normalise_newlines():
...
+
+
+def test_lint_jinja_macro_exclude():
+    """Test linting when excluding macros with unknown tags."""
+    config_path = "test/fixtures/templater/jinja_exclude_macro_path/.sqlfluff"
+    config = FluffConfig.from_path(config_path)
+    linter = Linter(config=config)
+    sql_file_path = "test/fixtures/templater/jinja_exclude_macro_path/jinja.sql"
+    lint = linter.lint_path(sql_file_path)
+    assert lint.num_violations() == 0
```
alanmcruickshank, line 667: What's your thinking on having this test in addition to the _parse_ one? If the error would normally have been raised at the templating/parsing phase, I'm tempted to suggest removing this one as redundant.

Open to being persuaded though if I've missed something.

src/sqlfluff/core/templaters/jinja.py
```diff
@@ -420,6 +433,35 @@ def _get_macros_path(self, config: FluffConfig) -> Optional[List[str]]:
...
+            exclude_macros_path = config.get_section(
+                (self.templater_selector, self.name, "exclude_macros_from_path")
+            )
+            if exclude_macros_path:
+                result = [
+                    s.strip() for s in exclude_macros_path.split(",") if s.strip()
+                ]
+                if result:
+                    return result
+        return None
```
alanmcruickshank, line 463: This method duplicates a lot of `_get_macros_path`. Looks like everything is the same apart from the key (either `load_macros_from_path` or `exclude_macros_from_path`). To avoid so much duplication, could you consolidate them?

Open to other options, but perhaps by making the key one of the arguments, and then making this whole method a more generic `@staticmethod`?

alanmcruickshank (changes requested):
Thanks for putting this together. I think the overall approach is sound. Makes total sense 👍 .

There are a few places where I think we can streamline what you've done - most of them are about clarifying the test cases. My expectation is that none of them are controversial, but do let me know if I've missed something in the suggestions.

test/core/linter/linter_test.py
culpgrant: Yes, you are right - I will remove it. I wanted to ensure it was being tested properly but it is a duplicate.

src/sqlfluff/core/templaters/jinja.py
culpgrant: Good callout - I have simplified these into one method that takes in a key argument.

alanmcruickshank (approved):
Lovely 💯 .

Thanks for the contribution and the edits post-review. 👍

culpgrant:
Thanks for the edits @alanmcruickshank, that was helpful. I was on a plane when I wrote this so I probably should have caught some of those edits in the first place.

Issue #3495, opened 2022-06-27
.sqlfluffignore improvement


jonyscathe:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Links or command line

https://docs.sqlfluff.com/en/stable/configuration.html?highlight=ignore#id1

### Issue/Suggested Improvement

I am guessing what I want to do is possible somehow, and if so should be added into the sqlfluffignore section (or at least a link from this section).

My situation is that I am using sqlfluff-lint and sqlfluff-fix pre-commit hooks.
I have a file that is unparsable, due to #3494 
So I figure for now, I will put the unparsable file into a .sqlfluffignore file.
However, my pre-commit hook can never pass as I get a warning causing the pre-commit hook to fail:

```
sqlfluff-lint................................................................Failed
- hook id: sqlfluff-lint
- files were modified by this hook

WARNING    Exact file path src/..../table.sql was given but it was ignored by a .sqlfluffignore pattern in /home/..., re-run with `--disregard-sqlfluffignores` to skip .sqlfluffignore
```

I obviously don't want to run with `--disregard-sqlfluffignore` as I want the ignore to apply. But I want to be able to change that file, and still run the pre-commit hooks.
I am guessing there must be some way to do this as I can't be the first to have this issue. But it isn't obvious from the documentation how to suppress this warning so that the pre-commit hook can pass.
Or if there isn't already some way for the pre-commit hook to pass with an exact file path in the .sqlfluffignore file then I will change this to an enhancement request.

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

alanmcruickshank:
This is the same issue as #6003, and having looked at this I think the fix is the same:

> I've had a good dig into this one - and I don't think there's an easy fix because the warnings you're seeing are there intentionally for CLI users. Because of the way that pre-commit calls SQLFluff, we can't tell whether it's being called by a human directly or by pre-commit.

> 
> I do however have a workaround which I've documented in https://github.com/sqlfluff/sqlfluff/pull/6052:

```rst
Ignoring files while using :code:`pre-commit`
---------------------------------------------

Under the hood, `pre-commit`_ works by passing specific files to *SQLFluff*.
[...]

PR #6052, opened 2024-08-08
Big docs refactor.

### LINKED ISSUE #6003: Reduce/Remove log warnings for every ignored file
### Search before asking

### Description

When a file is skipped due to the .sqlfluffignore file a warning is added to the log. This happens for every file that is skipped.
Normally these warnings aren't shown, but when running sqlfluff through pre-commit, if there is an sqlfluff 'FAIL', the entire log output is written out.
When there are a lot of files which were skipped this can spam out the entire output so you can't really see where the the FAIL was that you need to fix.

### Use case

Basically would be good to keep the log output as clean as possible.
Ideas could be to have an option to not show these warnings or just show one warning instead of one warning for each file.

### Dialect

bigquery
dbt-templater

### Code of Conduct

alanmcruickshank:
Ok this is a pretty big PR.

This does a few key things:
- It breaks up some of the bigger docs files into smaller ones. In particular `configuration` and `templating`. This allows a more effective categorisation of some sections of docs where we can add more.
- Because the above change will change the url of several sections of docs which are referred to in the codebase, I've added both _redirects_ for the old locations and _permalinks_. **ALL** links top the docs in the main codebase now point to a `/perma` link, which is then redirected by the docs config to the right place. This means if we move config again in the future - we don't need to change the links in the main codebase.
- Because modularising the docs meant more subfolders, it seemed sensible to rename `docs/partials` to `docs/_partials` to distinguish it from the other folders.
- I've changed the title of the `layout` section from `Let's talk about whitespace` to `Layout & Whitespace Configuration`, which is less fun but more descriptive.
- I've added a section (lifted from the wiki) on how best to run tests when contributing into the development section.
- I've added a section titled `Ignoring files while using pre-commit` which resolves #6003 and #3495. Having checked exactly what's possible - I don't think we can/should resolve that issue in the way the OP suggests, but there is a feasible workaround.


WittierDinosaur (approved):
Massive line delta, all looks pretty reasonable

WittierDinosaur:
Merge conflicts :(

Issue #6036, opened 2024-08-01
Add `.sqlfluffignore` Configuration to `pyproject.toml`


nh916:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

Currently, `.sqlfluffignore` is used to specify files or directories that `sqlfluff` should skip over. However, it would be convenient to manage these ignore patterns directly within `pyproject.toml`, allowing all `sqlfluff` configurations to be centralized in a single file.

### Use case

Add support for a `[tool.sqlfluff.ignore]` section in `pyproject.toml` where users can specify the same ignore patterns that would typically go into a `.sqlfluffignore` file.

Example:

```toml
[tool.sqlfluff]
dialect = "ansi"
paths = ["./models"]

[tool.sqlfluff.ignore]
patterns = ["./venv", "./target"]
```

## Benefits

- **Centralized Configuration:** All sqlfluff settings are in one place.
- **Improved Workflow:** Avoids the need to manage separate .sqlfluffignore files.
- **Consistency:** Aligns with other tools using pyproject.toml for configuration.

I believe this enhancement would significantly improve sqlfluff usability and promote better project organization.

Thank you for considering this feature request.

### Dialect

All

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

alanmcruickshank:
This is a duplicate of #5596

Issue #5596, opened 2024-02-02
Support `.sqlfluffignore` configuration in `pyproject.toml`


aaaaahaaaaa:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

In the spirit of reducing the number of config files required, it would be great if the `sqlfluffignore` configuration could be supported in `pyproject.toml`.

### Dialect

`pyproject.toml` configuration

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

aaaaahaaaaa:
Sounds perfect

frehoy:
> This is definitely possible, and would also open up the option of specifying ignore options in the `.sqlfluff` file too. By setting up one we would do almost all of the work to set up the other.
> 
> My initial suggestion is that this would be it's actually just a new key in the top line config, something like:
> 
> ```ini
> [sqlfluff]
> ignore_files =
>     foo.sql
>     bar.sql
> ```
> 
> or
> 
> ```toml
> [tool.sqlfluff.core]
> ignore_files = [
>     "foo.sql",
>     "bar.sql",
> ]
> ```
> 
> How does that sound?

In our project we ignore entire directories like
```
#.sqlfluffignore
dbt_modules/
dbt_packages/
```
so calling it `ignore_files` files could be a bit confusing. How about `ignore_paths`?

PR #6057, opened 2024-08-09
Extract path discovery routines from config and linter.

alanmcruickshank:
This supports development on #5596 and #6036.

The methods to do path discovery and manage `.sqlfluffignore` files are in very strange places right now. There's a `@classmethod` on the `ConfigLoader` for finding the files, and then some methods in the `Linter` class for doing path discovery. None of these really need to be there, and would make more sense isolated out.

This PR does that. Pulls code from `linter.py` and from `config.py` to make a new module `discovery.py` which handles all the path discovery methods. That also means we can segregate the _tests_ for those methods out of the (already bloated) `linter_test.py`.

No new functionality here - just refactoring the existing code.


keraion (approved):
This is a nice, much needed split. LGTM.

PR #6060, opened 2024-08-11
fix(clickhouse): add support for DateTime64(precision, tz) and Tuples()

ogirardot:
The changes introduce support on the Clickhouse dialect for types and type casts related to `DateTime64` and `Tuple` (structs in clickhouse), the following syntaxes were unparseable using sqlfluff 3.1.0 (and dbt-clickhouse 1.8.1, with `sqlfluff-templater-dbt==3.1.0` as example :
```
SELECT (1, 'two', '2024-01-01')::Tuple(id Int64, name String, created_at Date32) as struct;
-- and 
SELECT '2024-01-01'::DateTime64(3, 'Europe/Paris') as datetime;
```

Nope

### Pull Request checklist


test/fixtures/dialects/clickhouse/create_table.yml
```diff
@@ -507,11 +507,10 @@ file:
...
-            bracketed_arguments:
-              bracketed:
-                start_bracket: (
-                data_type_identifier: String
-                end_bracket: )
+          bracketed:
+            start_bracket: (
+            word: String
+            end_bracket: )
       - end_bracket: )
```
alanmcruickshank, line 514: This change looks like a regression. While it won't be a `bracketed_arguments` anymore, it should still be part of the `data_type` segment. Can you make sure your change supports `Nullable` types appropriately?

alanmcruickshank (changes requested):
Thanks for contributing this!

Changes look mostly good. Thanks for following the same pattern used in the bigquery dialect.

One thing which isn't critical, but would be nice is whether we can get `DateTime64` to parse as a `data_type_identifier` rather than a `keyword` in the test cases 🤔 . That way they'll get handled appropriately by the capitalisation rules.

I _think_ the best way to achieve that would be instead of just `"DATETIME64"`, you use the slightly longer `StringParser("DATETIME64", DatatypeIdentifierSegment)`. It's a big more verbose but I think it's more robust on how things get treated in the rules.

test/fixtures/dialects/clickhouse/create_table.yml
ogirardot: oh you're right, I'll look into it

src/sqlfluff/dialects/dialect_clickhouse.py
```diff
@@ -328,15 +328,66 @@ class BracketedArguments(ansi.BracketedArguments):
...
+    the target type in casts.
+    """
+
+    type = "data_type"
+    match_grammar = OneOf(
+        Ref("TupleTypeSegment"),
+        Ref("DatatypeIdentifierSegment"),
+        Ref("NumericLiteralSegment"),
+        Sequence(
+            OneOf("DATETIME64"),
```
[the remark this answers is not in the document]
ogirardot: Got it using StringParser and CodeSegment

test/fixtures/dialects/clickhouse/create_table.yml
ogirardot: I decided to handle the nullable case more explicitely than before

ogirardot:
@alanmcruickshank it should be good to go, I fixed the last violations of the pre-commit. Thank you for your time and review

alanmcruickshank (approved):
LGTM! 👍

Thanks for your contribution, and especially for taking the time to make the edits.

PR #6062, opened 2024-08-12
ST05: Handle set statement's subsequent queries

### LINKED ISSUE #4347: L042 does not flag subqueries after UNION
### Search before asking

### What Happened

I have L042 set to forbid_subquery_in = both.

SQLFluff will not raise a violation about a subquery in the FROM clause when the statement occurs after a `UNION ALL` or `UNION` in SQL.
When I swapped the sides of the UNION around so that the statement containing the subquery was first, then SQLFluff correctly raised a L042 rule violation. 

### Expected Behaviour

I expect SQLFluff's rule 042  to be able to flag a subquery in the FROM clause, even if the SELECT statement is after a UNION or UNION ALL keyword.

### Observed Behaviour

SQLFluff will not raise a violation about a subquery in the FROM clause when the statement occurs after a `UNION ALL` or `UNION` in SQL.
When I swapped the sides of the UNION around so that the statement containing the subquery was first, then SQLFluff correctly raised a L042 rule violation. 

My workspace .sqlfluff 

image

### How to reproduce

1. Make a SQL file with the contents:
```
SELECT 1 AS value_name
UNION
SELECT value
FROM (SELECT 2 AS value_name)
```
1. Configure your .sqlfluff file to include
```
# Do not allow any subqueries. Not in from clauses, nor in join clauses.
[sqlfluff:rules:L042]
forbid_subquery_in = both
```
1. Lint the SQL file create in step 1.

### Dialect

databricks

### Version

sqlfluff version 1.4.5
python version 3.11.1

### Configuration
```
####################################################################################################
# Set dialect to databricks, line length rule to warning and verbose to '2' (so it shows stats)
[sqlfluff]
dialect = databricks
warnings = L062
verbose = 1
recurse = 1
output_line_length = 100
large_file_skip_byte_limit = 30000

# CTEs need to be indented, and ON should be indented from the JOIN clause
[sqlfluff:indentation]
indented_ctes = True
indented_using_on = True

# Change default to trailing commas (from 'consistent')
# Additionally, commas are to touch the element before
[sqlfluff:layout:type:comma]
spacing_before = touch
line_position = trailing

# Align aliased column names. Note that you will still need to add the word 'AS' if it is not there
[sqlfluff:layout:type:alias_expression]
# We want non-default spacing _before_ the alias expressions.
spacing_before = align
# We want to align them within the next outer select clause.
# This means for example that alias expressions within the FROM
# or JOIN clause would _not_ be aligned with them.
align_within = select_clause
# The point at which to stop searching outward for siblings, which
# in this example would likely be the boundary of a CTE. Stopping
# when we hit brackets is usually a good rule of thumb for this
# configuration.
align_scope = bracketed

# Set general rules for SQLFluff
[sqlfluff:rules]
# We want to only indent 2 spaces instead of 4
tab_space_size = 2
# Max line length should be 100, not 80 since that is too short
max_line_length = 100
# We use Databricks
dialect = databricks
# We want to indent our CTEs
indented_ctes = True

# Change Keywords to uppercase only. The default for this is 'consistent' which means if there are
# more lowercase Keywords it will make them all lowercase.
[sqlfluff:rules:L010]
capitalisation_policy = upper

# Implicit/explicit aliasing of columns. Explicit meaning we require an 'AS' to alias columns
[sqlfluff:rules:L012]
aliasing = explicit

[sqlfluff:rules:L013]
allow_scalar = False

# Change Unquoted Identifiers (aliased names) to lowercase only. The default for this is
# 'consistent' which means if there are more uppercase Unquoted Identifiers it will make
# them all uppercase.
[sqlfluff:rules:L014]

# Change Functions to uppercase only. The default for this is 'consistent' which means if there are
# more lowercase Functions it will make them all lowercase.
[sqlfluff:rules:L030]

# Avoid table aliases in from clauses and join conditions. Putting this code here runs this rule
# even for dialects where this rule is disabled by default.
[sqlfluff:rules:L031]
force_enable = True

# Change Null & Boolean Literals to uppercase only. The default for this is 'consistent' which means
# if there are more lowercase Null & Boolean Literals it will make them all lowercase.
[sqlfluff:rules:L040]
capitalisation_policy = upper

# Do not allow any subqueries. Not in from clauses, nor in join clauses.
[sqlfluff:rules:L042]
forbid_subquery_in = both

# Fully qualify JOIN clause. This means having the INNER on an INNER JOIN and the OUTER on the 
# LEFT or RIGHT OUTER JOINS
[sqlfluff:rules:L051]
fully_qualify_join_types = both

# GROUP BY/ORDER BY column references. Implicit meaning we should use numbers that represent
# each columns instead of explicitly stating the column name in the GROUP or ORDER BY.
[sqlfluff:rules:L054]
group_by_and_order_by_style = implicit

[sqlfluff:rules:L062]
# Comma separated list of blocked words that should not be used (BOOLEAN instead of BOOL, INT
# instead of INT8, LIKE instead of ILIKE, TIMESTAMP instead of TIME, IFNULL instead of NVL)
blocked_words = BOOL, INT8, ILIKE, TIME, NVL

# Change Data Types to UPPER case only. The default for this is 'consistent' which means if
# there are more lowercase Data Types it will make them all lowercase
[sqlfluff:rules:L063]

# Consistent usage of single quotes for quoted literals. Use double quotes only when needed
[sqlfluff:rules:L064]
preferred_quoted_literal_style = single_quotes

# SQL type casting. Use shorthand (::) instead of CAST() or CONVERT()
[sqlfluff:rules:L067]
preferred_type_casting_style = shorthand
```

### Code of Conduct

keraion:

This handles cases when a set statement such as a union has a from/join subquery in the second or later query.
- fixes #4347 

None

### Pull Request checklist

  - Other.


alanmcruickshank (approved):
VERY NICE.

Impressively clean implementation 💯

PR #6063, opened 2024-08-12
feat(clickhouse): add support for decimal(x,y), decimal32(x) and match

ogirardot:
Add support for clickhouse special decimals and match(regexp) functions, it includes : 
```
Numeric(x, y)
Decimal(x, y)
Decimal32(x)
Decimal64(x)
Decimal128(x)
Decimal256(x)
match(field, 'regexp')
```

Nope

### Pull Request checklist


src/sqlfluff/dialects/dialect_clickhouse.py
```diff
@@ -350,6 +350,32 @@ class DatatypeSegment(BaseSegment):
...
+                    StringParser("DECIMAL32", CodeSegment, type="data_type_identifier"),
+                    StringParser("DECIMAL64", CodeSegment, type="data_type_identifier"),
+                    StringParser(
+                        "DECIMAL128", CodeSegment, type="data_type_identifier"
+                    ),
+                    StringParser(
+                        "DECIMAL256", CodeSegment, type="data_type_identifier"
+                    ),
+                ),
+                OptionallyBracketed(Ref("NumericLiteralSegment")),  # precision
```
WittierDinosaur, line 375: Should this be `Optional(Bracketed(...))` rather than `OptionallyBracketed`?

ogirardot: I'm not sure about the difference, there are already quite a few OptionallyBracketed in the file, should it be refactored ?

alanmcruickshank: `OptionallyBracketed` implies that something can be bracketed or not bracketed - but one of them must be there. e.g. `OptionallyBracketed("FOO")` would match `FOO` and `(FOO)`.

`Bracketed("FOO", optional=True)` requires brackets around `FOO`, but also allows nothing to be present. i.e. matches `(FOO)` and ``, but _wouldn't_ match an unbracketed `FOO`.

src/sqlfluff/dialects/dialect_clickhouse.py
```diff
@@ -350,6 +350,32 @@ class DatatypeSegment(BaseSegment):
...
+                optional=True,
+            ),
+        ),
```
alanmcruickshank, line 378:
```suggestion
        # double args
        Sequence(
            OneOf(
                StringParser("DECIMAL", CodeSegment, type="data_type_identifier"),
                StringParser("NUMERIC", CodeSegment, type="data_type_identifier"),
            ),
            Ref("BracketedArguments", optional=True),
        ),
        # single args
        Sequence(
            OneOf(
                StringParser("DECIMAL32", CodeSegment, type="data_type_identifier"),
                StringParser("DECIMAL64", CodeSegment, type="data_type_identifier"),
                StringParser(
                    "DECIMAL128", CodeSegment, type="data_type_identifier"
                ),
                StringParser(
                    "DECIMAL256", CodeSegment, type="data_type_identifier"
                ),
            ),
            Bracketed(Ref("NumericLiteralSegment"), optional=True),  # precision
        ),
```

src/sqlfluff/dialects/dialect_clickhouse.py
ogirardot: OK then yes, and taking a second look at this it's not even right : 
```
>> SELECT 1::Decimal32

SELECT CAST('1', 'Decimal32')

Query id: 72b396fa-10d6-44fd-9771-229fb886c7a3


Elapsed: 0.000 sec. 

Received exception:
Code: 42. DB::Exception: Decimal32 | Decimal64 | Decimal128 | Decimal256 data type family must have exactly one arguments: scale: In scope SELECT CAST('1', 'Decimal32'). (NUMBER_OF_ARGUMENTS_DOESNT_MATCH)
```

So Decimal32() or Decimal32 is incorrect so the optional part is not even true.
Let me fix this

src/sqlfluff/dialects/dialect_clickhouse.py
ogirardot: Thanks, I'll restart from this because taking a second look Decimal32 is not even allowed

PR #6066, opened 2024-08-13
Add perma-links for rules

alanmcruickshank:
Unfortunately, #6052 will break the VSCode docs link, by moving the rules again. However I now have a better solution so that we never need to break them again!

**RULE PERMALINKS**.

This PR adds permanent links to the rules which can then internally redirect. e.g. the VSCode plugin would be able to point at `/perma/rule/RF01.html` and be internally redirected by the docs to the appropriate location - in this case `/reference/rules.html#sqlfluff.rules.sphinx.Rule_RF01`.

I'll make a PR for the VSCode plugin shortly.

### DIFF
--- a/docs/generate-rule-docs.py
+++ b/docs/generate-rule-docs.py
@@ -1,5 +1,6 @@
 """Generate rule documentation automatically."""
 
+import json
 from collections import defaultdict
 from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


keraion (approved):
LGTM. The rules might be able to use [target placeholders](https://documatt.com/sphinx-reredirects/usage.html#target-placeholders), but that's a matter of preference.

WittierDinosaur (approved):
LGTM

PR #6080, opened 2024-08-15
Rationalise config discovery routines.

alanmcruickshank:
This follows on from #6057. It supports development on #5596 and #6036.

Apart from removing tests for a method that doesn't exist anymore, this doesn't change the test cases - it's just a refactoring. I've also significantly improved the docs in this module which were inaccurate at best.

In particular here:
- I've simplified the path walk algorithm, doing better pruning, and ensuring we only need to walk the file tree _once_.
- I've modularised and extracted many smaller methods to reduce duplication.


WittierDinosaur (approved):
LGTM

Issue #5852, opened 2024-05-06
MySQL - `CREATE TABLE foo SELECT * FROM bar` statement causes `Found unparsable section` error 


gdubicki:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

I am trying to validate some SQL scripts that create syntax like:

```sql
CREATE TABLE foo SELECT * FROM bar;
```

### Expected Behaviour

This syntax should work as [MySQL supports it](https://dev.mysql.com/doc/refman/5.7/en/create-table-select.html).

### Observed Behaviour

```
$ sqlfluff lint --dialect mysql my_script.sql 
== [my_script.sql] FAIL                                                                                                                                          
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'CREATE
                       | TABLE foo SELECT * FROM bar;'
WARNING: Parsing errors found and dialect is set to 'mysql'. Have you configured your dialect correctly?
All Finished 📜 🎉!
```

### How to reproduce

1. Create a `my_script.sql` file:
```sql
CREATE TABLE foo SELECT * FROM bar;
```
2. Try to lint it with:
```bash
sqlfluff lint --dialect mysql my_script.sql 
```

### Dialect

MySQL

### Version

```
$ sqlfluff --version
sqlfluff, version 3.0.5
```
...installed using pipx, Python 3.11.8.

### Configuration

Default

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

MarkMarine:
I'm seeing the same parsing error when using CREATE FUNCTION ...

KulykDmytro:
same in `ansi` and `athena`

PR #6109, opened 2024-08-25
MYSQL: Create Table Optional AS

WittierDinosaur:
Fxies #5852

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_mysql.py
+++ b/src/sqlfluff/dialects/dialect_mysql.py
@@ -415,193 +415,219 @@ class CreateTableStatementSegment(ansi.CreateTableStatementSegment):
     TRS = Ref("TableReferenceSegment")
     SQIS = Ref("SingleQuotedIdentifierSegment")
 
-    match_grammar = ansi.CreateTableStatementSegment.match_grammar.copy(
-        insert=[
-            AnyNumberOf(
-                Sequence(
-                    Ref.keyword("DEFAULT", optional=True),
-                    OneOf(
-                        Ref("ParameterNameSegment"),
-                        Sequence("CHARACTER", "SET"),
-                        Sequence(OneOf("DATA", "INDEX"), "DIRECTORY"),
-                        Sequence("WITH", "SYSTEM"),
+    match_grammar = Sequence(
+        "CREATE",
+        Ref("OrReplaceGrammar", optional=True),
+        Ref("TemporaryTransientGrammar", optional=True),
+        "TABLE",
+        Ref("IfNotExistsGrammar", optional=True),
+        Ref("TableReferenceSegment"),
+        OneOf(
+            # Columns and comment syntax:
+            Sequence(
+                Bracketed(
+                    Delimited(
+                        OneOf(
+                            Ref("TableConstraintSegment"),
+                            Ref("ColumnDefinitionSegment"),
+                        ),
+                    )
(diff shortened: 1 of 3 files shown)


alanmcruickshank:
@WittierDinosaur, there's a lot of new dialect here, but not a lot of new test coverage. Reckon we can reference more of the ANSI dialect in some places and so not need the extra tests, or think enough of this is MySQL/MariaDB specific that it deserves it's own test cases?

WittierDinosaur:
@alanmcruickshank this is actually functionally identical to how it is now - the only change I'm actually making in this PR is changing `"AS"` to `Ref.keyword("AS", optional=True)`. In order to do that I needed to replicate the ANSI version in MySQL, and again in MariaDB, where before they were using `match_grammar.copy(...)`

alanmcruickshank:
@WittierDinosaur Ahh ok, I've properly reviewed what was in the ANSI dialect and what was in MySQL and Maria before your change.

- On your choice to replace the MySQL dialect create statement using `.copy()` to one which totally just overwrites the grammar - I think this makes a lot of sense 👍 . The ANSI one is very simple, and this approach is cleaner.
- Given that, I think my point about insufficient test coverage of this is still true, but given you're not the one adding the complexity, I guess that's a little moot at this stage.
- On the changes to MariaDB, I understand them less - what's the thinking behind making that changes there too? I would have thought that you shouldn't need to change that too - if the fix in MySQL already works?

I might add some test coverage to this PR if you don't mind?

WittierDinosaur:
Maria breaks when i changed the mysql grammar to not use copy. Can't exactly remember why. Feel free to add test coverage if you fancy

WittierDinosaur:
@alanmcruickshank any update on this one?

alanmcruickshank:
@WittierDinosaur - sorry for missing this one. Just taken it for a spin. I tried totally removing the definition of `CreateTableStatementSegment` from MariaDB so that it falls back to MySQL, and the test case which fails is `create_temporary_table_ddl.sql`:

```sql
CREATE TEMPORARY TABLE tbl_name (
    id INT PRIMARY KEY AUTO_INCREMENT,
    col VARCHAR(255) DEFAULT '' NOT NULL,
    INDEX(col)
) AS SELECT id, col FROM table_name;

CREATE TEMPORARY TABLE tbl_name (
    id INT PRIMARY KEY AUTO_INCREMENT,
    col VARCHAR(255) DEFAULT '' NOT NULL,
    INDEX(col)
) SELECT id, col FROM table_name;

CREATE TEMPORARY TABLE tbl_name (INDEX(col)) AS
    SELECT id, col FROM table_name;

CREATE TEMPORARY TABLE tbl_name (INDEX(col))
    SELECT id, col FROM table_name;
```

The reason it's failing is that the definition from this PR in the MySQL dialect doesn't support having any table definition *and* an `AS` clause (whether it uses the keyword or not). The problem I see is that this test case _should_ pass in the MySQL dialect too based on the docs (https://dev.mysql.com/doc/refman/8.4/en/create-table.html).

By my understanding that means:
1) we shouldn't need to redefine the statement in MariaDB (which is good for avoiding duplication).
2) the MySQL definition needs a tweek. It looks like the `AS` clause needs to move right to the end rather than being up front as an alternative to the `LIKE` clause.

Does that line up with your understanding?

PR #6112, opened 2024-08-27
Fix support of INTERVAL in ClickHouse

Pavel-Strybuk:
Fixes support of INTERVAL in ClickHouse as it differs from ANSI SQL. Also adds datetime units based on ClickHouse source code instead of (official but brief) documentation.

No

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_clickhouse.py
```diff
@@ -1664,3 +1724,18 @@ class LimitClauseSegment(ansi.LimitClauseSegment):
...
+    https://clickhouse.com/docs/en/sql-reference/data-types/special-data-types/interval
+    https://clickhouse.com/docs/en/sql-reference/operators#operator-interval
+    """
+
+    type = "interval_expression"
+    match_grammar = Sequence(
+        "INTERVAL",
+        Ref("ExpressionSegment"),
+        Ref("DatetimeUnitSegment"),
+    )
```
keraion, line 1741: According to the docs linked here, something along the lines of `INTERVAL '1 DAY'` should also be valid. The ANSI implementation is very close to what the documentation shows.

Pavel-Strybuk: @keraion Thank you for pointing me to the issue! I did the change similar to ansi dialect. Can you please check again when have time?
I have a question, is it possible to `parse` somehow '1 days' string literal, so it will not allow non-supported units, like '1 abcde'? Currently I use generic 'QuotedLiteralSegment'  as in other dialects but I suspect there is a better option...

keraion: Sorry for the delay on this, I've been offline for the past few days. Yes, it is technically possible to introduce the unit parsing, but it might not be worth adding the additional logic. Usually, it is a matter of letting the parser be more forgiving and let the DBMS reject the invalid query. If we start specifying every type of unit, that places more maintenance on the dialect with little benefit.

keraion (approved):
LGTM. Thanks!

PR #6115, opened 2024-08-28
Add option for allowing only a subset of rules for noqa

### LINKED ISSUE #6097: New setting: `allowed_noqa`
### Search before asking

### Description

`allowed_noqa` would allow the users to pass specific rules for which `noqa` statements will be respected by the linter/fixer, while all others will be ignored.

### Use case

Currently, the configuration admits for a property called `disable_noqa` which will ignore `-- noqa` statements and still run checks with all rules in given files. The problem with this rule is that it's binary. Either all `noqa`s are accepted, or all `noqa`s are ignored. But the reality of a code base is more nuanced as certain rules might be allowed to be `noqa`ed, while others might not.

Take for example line length, and let's suppose we have set it to 120 chars. Sometimes we run into the problem that a table name might be longer than this. Especially when working with systems like Google BQ where `FROM` and `JOIN` targets usually include project IDs and dataset IDs. There's no way of breaking this into multiple lines.

```sql
-- Example
SELECT
  mt.id
  , mt.created_date
  , CASE
      WHEN mt.status = "Done" THEN 1
      WHEN mt.status = "Not done" THEN 2
      ELSE 3
    END AS status_value
FROM `my_looooooooooooog_project_name.my_looooooooooooooog_dataset_name.my_very_very_very_loooooooooooooooooo_table_name` AS mt
```

Currently, the only way of solving this issue is to set `disable_noqa = False` and use a `--noqa: LT05` in that line. But this leaves the project exposed to anyone else adding any other `-- noqa:` statement in any other line, and for any other rule.

### Dialect

I would say all of them, but personally, I've seen this be more of an issue in BigQuery.

### Code of Conduct

keraion:

Adds an option to allow only a subset of rules when a line is marked with `--noqa: rule_id`
- fixes #6097 

None

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -304,6 +304,11 @@ def core_options(f: Callable) -> Callable:
         help="Set this flag to ignore inline noqa comments.",
     )(f)
     f = click.option(
+        "--allowed-noqa",
+        default=None,
+        help="Ignore all but the listed rules to ignore inline noqa comments.",
+    )(f)
+    f = click.option(
         "--library-path",
         default=None,
         help=(
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -304,9 +304,9 @@ def core_options(f: Callable) -> Callable:
         help="Set this flag to ignore inline noqa comments.",
     )(f)
     f = click.option(
-        "--allowed-noqa",
+        "--disable-noqa-except",
         default=None,
-        help="Ignore all but the listed rules to ignore inline noqa comments.",
+        help="Ignore all but the listed rules inline noqa comments.",
     )(f)
     f = click.option(
         "--library-path",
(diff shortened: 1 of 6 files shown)


alanmcruickshank:
This definitely resolves the issue originally raised, but in reviewing it I'm finding it very difficult to wrap my head around because of all the double negatives.

In this implementation: `allow_noqa` allows a `noqa` which disallows a rule which is otherwise enabled. Setting `allow_noqa` disables all other `noqa` statements, except for those which are listed in the cli argument.

I think the interaction between `allow_noqa` and `noqa` statements which _aren't_ specified feels very confusing to users, i.e. it has `allow` in the name, but it's actually a way of _disabling_ everything else. I wonder if we need a better name, any suggestions? @keraion @daczarne

I'd probably suggest `disable_noqa_except`, which makes it clearer that it's related to the other option, but totally up for other options which are more explicit.

keraion:
@alanmcruickshank I agree that the naming is a bit difficult to understand. I think the suggestion of `disable_noqa_except` is probably the easiest to understand.

daczarne:
I like this new name a lot better. I think `disable_noqa_except` reads much clearer and is more consistent with the current nomenclature of `disable_noqa`.

Issue #2848, opened 2022-03-12
diff-quality and dbt-templater interact in a way which skips some files.


noel:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

When I run `diff-quality --violations sqlfluff` I don't see the same violations that I see with `sqlfluff lint`
I also tried staging and committing the modified file, but that didn't change the results.

```
~/workspace/transform $ diff-quality --violations sqlfluff
-------------
Diff Quality
Quality Report: sqlfluff
Diff: origin/main...HEAD, staged and unstaged changes
-------------
transform/models/bays/bay_covid/base_cases.sql (100%)
-------------
Total:   1 line
Violations: 0 lines
% Quality: 100%
-------------
~/workspace/transform $ git commit -am "test commit"
[main c7624d2] test commit
 1 file changed, 1 insertion(+), 1 deletion(-)
~/workspace/transform $ diff-quality --violations sqlfluff
-------------
Diff Quality
Quality Report: sqlfluff
Diff: origin/main...HEAD, staged and unstaged changes
-------------
transform/models/bays/bay_covid/base_cases.sql (100%)
-------------
Total:   1 line
Violations: 0 lines
% Quality: 100%
-------------
~/workspace/transform $ sqlfluff lint
=== [dbt templater] Sorting Nodes...                                                  
=== [dbt templater] Compiling dbt project...                                          
=== [dbt templater] Project Compiled.                                                 
== [/config/workspace/transform/models/bays/bay_covid/base_cases.sql] FAIL            
L:   9 | P:   5 | L010 | Keywords must be lower case.                                 
All Finished 📜 🎉!  
```

### Expected Behaviour

Expected the same violation found with `sqlfluff lint`
```
L:   9 | P:   5 | L010 | Keywords must be lower case.                                 
```

### Observed Behaviour

No violation found
```
-------------
Diff Quality
Quality Report: sqlfluff
Diff: origin/main...HEAD, staged and unstaged changes
-------------
transform/models/bays/bay_covid/base_cases.sql (100%)
-------------
Total:   1 line
Violations: 0 lines
% Quality: 100%
-------------
```

### How to reproduce

Modify a file to make L010 fail and execute `diff-quality --violations sqlfluff`

### Dialect

`dialect = snowflake`

### Version

```
$ sqlfluff --version
sqlfluff, version 0.11.0

[...]

PR #6128, opened 2024-08-30
Split apart config module

alanmcruickshank:
As part of ongoing work on better handling of ignore files, I'm going to do some long overdue work on how config files are loaded and cached. It's code that hasn't been touched in a long time.

Before I do much significant surgery, I want to split apart the current `config.py` file, so that it's later clearer what I have and haven't changed in following PRs. I'll follow up later with some significant changes to the `loader` portion of the new module.

This PR does just that. No functionality changes, just splitting one module into three smaller files (and the tests into two smaller files).

This relates to work on #6036, #5596 & #2848


WittierDinosaur (approved):
LGTM

PR #6130, opened 2024-08-31
Allow ignore patterns in other config files.

### LINKED ISSUE #5596: Support `.sqlfluffignore` configuration in `pyproject.toml`

### LINKED ISSUE #6036: Add `.sqlfluffignore` Configuration to `pyproject.toml`

alanmcruickshank:
This resolves #6036 and resolves #5596. This builds on the restructuring of the `config` module (see #6128 & #6177). This also uses a little of the code from #6206.

This allows ignore patterns to be defined in `.sqlfluff` and `pyproject.toml` files. For this to work, we effectively end up loading config files in two phases, once while processing any ignore patterns and then later when sequencing files for parsing. That's why it's so important for the caching to be in place so we don't just duplicate a bunch of filesystem operations.

Under the hood, this is basically a re-write for the `discovery` module, but I think it's much better for it even without the new functionality.


src/sqlfluff/core/linter/discovery.py
```diff
@@ -54,13 +50,72 @@ def _check_ignore_specs(
...
+    # Use normalised path to ensure reliable caching.
+    config_dict = load_config_file_as_dict(Path(filepath).resolve())
+    ignore_section = config_dict.get("core", {})
+    if not isinstance(ignore_section, dict):
+        return None  # pragma: no cover
+    patterns = ignore_section.get("ignore_paths", [])
+    if isinstance(patterns, str):
+        patterns = patterns.split(",")
+    elif not patterns or not isinstance(patterns, list):
+        return None
```
WittierDinosaur, line 96: this probably wants an `else` to catch edge cases and throw an exception

alanmcruickshank: No need in this case, not being caught by either of these clauses means it's already a list and so doesn't need any edits. I'll add a comment to that explain that though 👍

WittierDinosaur (approved):
LGTM

keraion (approved):
LGTM

PR #6135, opened 2024-09-02
Method extraction and more robust typing in config.

alanmcruickshank:
This is a section of #6133 (which got too big).

It follows on from #6128.

The `ConfigLoader` class is a bit of a mess. Lots can be done to clean it up, and this starts that process by extracting out a few of the helper methods into the dict helpers module (`_config_elems_to_dict` and `_iter_config_elems_from_dict` - both of which are named as private methods but called all over the place 🙃 ), and removes another (`_incorporate_vals` which we can do without).

In doing that I've also introduced some much better typing across this module. In particular there were a lot of `Dict[str, Any]` types, which I've replaced with a much better `NestedStringDict` generic which is used to create a `ConfigMappingType` which avoids the use of `Any`.

As an aside, I learnt a ton more about typing Generics as part of putting this together! Very smug about some of my new mypy-fu.

There are a couple of other smaller bits in here - but that's the bulk of it.


WittierDinosaur (approved):
LGTM

PR #6138, opened 2024-09-02
Handle multi-processing dbt exceptions much better.

### LINKED ISSUE #6037: SQLFluff will hang on compilation error in dbt when using multiple processes
### Search before asking

Running sqlfluff with dbt templater on dbt project that fails compiling leads to a serious bug. When using more than one process, sqlfluff will hang after showing an exception and must be manually killed. This is esp. problematic when using it as a hook from within `pre-commit`. Then, `pre-commit` will hang endlessly without outputting or logging any error message. When using a single process, there is correct behavior: the error is properly displayed and sqlfluff will not hang but terminate. 

To reproduce, you need a a simple dbt project with at least two `.sql` files in it (only one file doesn't trigger this, because then only one process will be spawned). One file can be just a comment. The other file an obvious wrong model or something else that won't compile, e.g.:

```
select * from {{ref("abc")}}
```
Where `abc` doesn't exist. When running `sqlfluff lint -p 1` over this, it will show:

> 
> $ sqlfluff lint -p 1
> === [dbt templater] Sorting Nodes...                                                                                                                                                         
> file /home/mc/dev/open-source-contributions/pre-commit/dbt/models/model1.sql:   0%|                                                                                    | 0/2 [00:00<?, ?it/s]Traceback (most recent call last):
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/bin/sqlfluff", line 8, in <module>
>     sys.exit(cli())
>              ^^^^^
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
>     return self.main(*args, **kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/click/core.py", line 1078, in main
>     rv = self.invoke(ctx)
>          ^^^^^^^^^^^^^^^^
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
>     return _process_result(sub_ctx.command.invoke(sub_ctx))
>                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
>     return ctx.invoke(self.callback, **ctx.params)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/click/core.py", line 783, in invoke
>     return __callback(*args, **kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/sqlfluff/cli/commands.py", line 643, in lint
>     result = lnt.lint_paths(
>              ^^^^^^^^^^^^^^^
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/sqlfluff/core/linter/linter.py", line 1168, in lint_paths
>     for i, linted_file in enumerate(runner.run(expanded_paths, fix), start=1):
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/sqlfluff/core/linter/runner.py", line 108, in run
>     for fname, partial in self.iter_partials(fnames, fix=fix):
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/sqlfluff/core/linter/runner.py", line 61, in iter_partials
>     for fname, rendered in self.iter_rendered(fnames):
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/sqlfluff/core/linter/runner.py", line 44, in iter_rendered
>     for fname in self.linter.templater.sequence_files(
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/sqlfluff_templater_dbt/templater.py", line 407, in sequence_files
>     for key, node in self.dbt_manifest.nodes.items():
>                      ^^^^^^^^^^^^^^^^^
>   File "/home/mc/.rye/py/cpython@3.12.3/lib/python3.12/functools.py", line 995, in __get__
>     val = self.func(instance)
>           ^^^^^^^^^^^^^^^^^^^
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/sqlfluff_templater_dbt/templater.py", line 256, in dbt_manifest
>     _dbt_manifest = ManifestLoader.get_full_manifest(self.dbt_config)
>                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/dbt/parser/manifest.py", line 330, in get_full_manifest
>     manifest = loader.load()
>                ^^^^^^^^^^^^^
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/dbt/parser/manifest.py", line 486, in load
>     self.process_refs(self.root_project.project_name, self.root_project.dependencies)
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/dbt/parser/manifest.py", line 1132, in process_refs
>     _process_refs(self.manifest, current_project, node, dependencies)
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/dbt/parser/manifest.py", line 1591, in _process_refs
>     invalid_target_fail_unless_test(
>   File "/home/mc/dev/open-source-contributions/pre-commit/venv/lib/python3.12/site-packages/dbt/parser/manifest.py", line 1404, in invalid_target_fail_unless_test
>     raise TargetNotFoundError(
> dbt.exceptions.TargetNotFoundError: Compilation Error
>   Model 'model.my_dbt_model.model2' (models/model2.sql) depends on a node named 'abc' which was not found
> 

However, when running `sqlfluff lint -p 2`, we encounter the bug:

> 
> $ sqlfluff lint -p 2
> effective configured processes:  2
> === [dbt templater] Sorting Nodes...                                                                                                                                                         
> file /home/mc/dev/open-source-contributions/pre-commit/dbt/models/model1.sql:   0%|                                                                                    | 0/2 [00:00<?, ?it/s]Process ForkPoolWorker-1:
>   File "/home/mc/.rye/py/cpython@3.12.3/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
>     self.run()
>   File "/home/mc/.rye/py/cpython@3.12.3/lib/python3.12/multiprocessing/process.py", line 108, in run
>     self._target(*self._args, **self._kwargs)
>   File "/home/mc/.rye/py/cpython@3.12.3/lib/python3.12/multiprocessing/pool.py", line 114, in worker
>     task = get()
>            ^^^^^
>   File "/home/mc/.rye/py/cpython@3.12.3/lib/python3.12/multiprocessing/queues.py", line 389, in get
>     return _ForkingPickler.loads(res)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^
> TypeError: TargetNotFoundError.__init__() missing 3 required positional arguments: 'node', 'target_name', and 'target_kind'
> 

This must be aborted by `ctrl+c` to end the process. 

My config using most recent packages of everything involved:

```
==== sqlfluff ====
sqlfluff: 3.1.0 python: 3.12.3
implementation: cpython verbosity: 1
dialect: redshift templater: dbt
dbt: =1.8.4
rules: all
==== readout ====
```

Here's the `.sqlfluff` config:
```
[sqlfluff]
templater = dbt
dialect = redshift

[sqlfluff:templater:dbt]
project_dir = ./dbt
profiles_dir = ./dbt
```

### Code of Conduct

/e: After some research, I updated the original issue with a clearer description. The bug can in fact be attributed solely to sqlfluff. However, the bug itself becomes way more problematic when encountered from within pre-commit, because this fails / hangs silently then.

alanmcruickshank:
This resolves #6037. I hit this bug myself while working on a different project.

~The core problem was that even when a file wasn't successfully rendered, we still for a `partial` for it (although that partial didn't need any extra work). That partial was then pickled and sent across to the workers in a way which was unhelpful.~

~However, our use of `imap()` made it tricky to work around an iterator which sometimes should be farmed to a worker and sometimes not. After having a look at the other options available in `multiprocessing`, I opted to go a little more homegrown by using `apply_async` directly. This means there's a totally custom loop in there now, but one that handles successful rendering differently to unsuccessful and balances feeding a queue and processing completed work.~

The problem was with exceptions, but specifically that even when we catch the dbt exceptions, and raise a SQLFluff native one, the original exceptions were being stored in the __context__ and __cause__ attributes of the new exception (see [these docs](https://docs.python.org/3/library/exceptions.html#inheriting-from-built-in-exceptions)). Somewhat annoyingly, it's really hard to stop python doing this, so I had to do something a little hacky (although not too bad), to re-raise exceptions outside that context so we get "clean" exceptions without any reference to the initial dbt ones. I've put all this complexity into a decorator so that it's fairly isolated from everything else, and updated the test cases to make sure we don't have any dbt exceptions lingering in those attributes of the exceptions.

Thanks for all your help on this one, I was definitely going about this the wrong way initially and I think I now understand a lot better what the cause was of the original issue. I've confirmed that this solution resolves the issue in the project where I was experiencing the issue.


src/sqlfluff/core/linter/runner.py
```diff
@@ -127,54 +133,128 @@ def __init__(self, linter, config, processes) -> None:
...
+        the AsyncResult of each, which is deleted upon receipt of a
+        result from that job.
         """
         with self._create_pool(
             self.processes,
             self._init_global,
             (self.config,),
         ) as pool:
             try:
-                for lint_result in self._map(
```
barrywhart, line 144: I have an idea for a possible simpler approach (debatably, ha !-- at least less custom code?). Pseudocode:

```
for lint_result in itertools.chain(
    failed_filenames_empty_lint_result(),  # How to implement this? Similar logic as the loop below, I think.)
    self._map(
        pool,
        self._apply,
        self.iter_partials(fnames, fix=fix),  # Need to modify `Iter_partials()` so it skips failed files
    )
):
    # Loop body remains the same as previously
```

Thoughts? I like using `map()` if we can.

alanmcruickshank: The problem with `map` is that we don't get _any_ of the results until _all_ of the tasks have completed. The only implementations where we can get the results "as we go" appear to be `imap` and `imap_unordered`, but they have this strange hanging behaviour with exceptions.

I'm all for simple implementations though which take up less space!

alanmcruickshank: Oh - and on the skipping failed files - we do still need to return them, they just shouldn't be passed to the workers, so we do kind of need this "two path" handling which I've introduced. I admit that there's some code duplication in there which we can reduce though.

src/sqlfluff/core/linter/runner.py
```diff
@@ -53,16 +57,19 @@ def iter_partials(
         self,
         fnames: List[str],
         fix: bool = False,
-    ) -> Iterator[Tuple[str, Callable]]:
+    ) -> Iterator[Tuple[str, bool, Callable[[], LintedFile]]]:
```
barrywhart, line 60: Now that this tuple has 4 fields, should we change it to something more readable, like a `NamedTuple`?

src/sqlfluff/core/linter/runner.py
```diff
@@ -127,54 +133,128 @@ def __init__(self, linter, config, processes) -> None:
...
+                # And a dict to hold the async references
+                async_map: Dict[str, multiprocessing.pool.AsyncResult] = {}
+                # Target queue size should be twice the the number of processes.
+                target_queue_depth = self.processes * 2
+                partials_iterable = self.iter_partials(fnames, fix=fix)
+
+                while True:
+                    # If there's something to process, do so.
+                    try:
+                        lint_result = queue.get(block=False)
```
barrywhart, line 172: Should we at least sleep for a brief period? Otherwise, it looks like this loop would peg one of the CPUs at 100%.

barrywhart:
The changes look reasonable, but complex enough that I'd feel better with more testing. Is this something where we could ask some maintainers and/or power users to test it, especially in cases where we expect rendering to fail?

src/sqlfluff/core/linter/runner.py
alanmcruickshank: I'm wondering whether to get rid of the whole `iter_partials` method entirely 🤔. I wonder if there's a significant opportunity for simplification here.

alanmcruickshank: Yep, simpler without. I've done a few rounds of simplification. I'm going to come back to it tomorrow and see if I can go further.

alanmcruickshank:
@barrywhart @keraion I've done some extra testing and simplifaction on this PR and I thought I should report back with the results:
- The changes I've made to the dbt templater, _don't fix the bug alone_. Think of them as quality of life improvements and better logging, but if I just apply those changes - the hang still occurs.
- I've done some testing with this `apply` based setup, and it does seem to be slower than the old setup. Which is frustrating. I'm having a look now at some other options.

I do see the point of using as many of the `multiprocessing` functions as possible if we can.

barrywhart:
@alanmcruickshank: Have you tried the experiment I suggested in my earlier [comment](https://github.com/sqlfluff/sqlfluff/pull/6138#discussion_r1742639859)? It's possible that will help narrow down the reason it's hanging.

keraion:
@alanmcruickshank I did play around with this earlier and was able to get `imap_unordered` working. That being said, it does need some clean up first.

alanmcruickshank:
> @alanmcruickshank: Have you tried the experiment I suggested in my earlier [comment](https://github.com/sqlfluff/sqlfluff/pull/6138#discussion_r1742639859)? It's possible that will help narrow down the reason it's hanging.

I did have a try of some of it, but no luck. I think there's more I can try though.

mc51:
Have you considered the comment on `tblib` in the original issue [here](https://github.com/sqlfluff/sqlfluff/issues/6037#issuecomment-2272251950)? I wonder if it is possible to fix this without having to make those involved changes. I'm thinking one might selectively install tblib and leave out the relevant dbt exceptions following [this approach](https://github.com/ionelmc/python-tblib/tree/master?tab=readme-ov-file#pickling-exceptions-together-with-their-traceback-and-chain-python-3-only):

> Selectively install tblib for Exception instances just before they are pickled

Since you're [catching the dbt exceptions](https://github.com/sqlfluff/sqlfluff/pull/6144) more nicely now, this might be all it takes.

alanmcruickshank:
> Have you considered the comment on `tblib` in the original issue [here](https://github.com/sqlfluff/sqlfluff/issues/6037#issuecomment-2272251950)? I wonder if it is possible to fix this without having to make those involved changes. I'm thinking one might selectively install tblib and leave out the relevant dbt exceptions following [this approach](https://github.com/ionelmc/python-tblib/tree/master?tab=readme-ov-file#pickling-exceptions-together-with-their-traceback-and-chain-python-3-only):
> 
> > Selectively install tblib for Exception instances just before they are pickled
> 
> Since you're [handling the dbt exceptions](https://github.com/sqlfluff/sqlfluff/pull/6144) more nicely now, this might be all it takes.

Oh interesting! I just tried this approach, and it definitely gets us one step further. Thanks for the pointer. 💯

alanmcruickshank:
@barrywhart @keraion @mc51 I think I've solved this. And it didn't require re-writing the loop!

The problem was with exceptions, but specifically that even when we catch the dbt exceptions, and raise a SQLFluff native one, the original exceptions were being stored in the `__context__` and `__cause__` attributes of the new exception (see [these docs](https://docs.python.org/3/library/exceptions.html#inheriting-from-built-in-exceptions)). Somewhat annoyingly, it's _really hard_ to stop python doing this, so I had to do something a little hacky (although not too bad), to re-raise exceptions outside that context so we get "clean" exceptions without any reference to the initial dbt ones. I've put all this complexity into a decorator so that it's fairly isolated from everything else, and updated the test cases to make sure we don't have any dbt exceptions lingering in those attributes of the exceptions.

Thanks for all your help on this one, I was definitely going about this the wrong way initially and I think I now understand a lot better what the cause was of the original issue. I've confirmed that this solution resolves the issue in the project where I was experiencing the issue.

alanmcruickshank:
Incidentally, this refactor highlighted that the `sequential_fails` code was never being used (must only be relevant for older dbt versions). I've removed it to make the code even simpler.

keraion (commented):
This is much simpler than the original rework. Nice job!

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -70,6 +72,103 @@ class DbtConfigArgs:
...
+    unwanted dbt exceptions so that we don't cause issues when in
+    multithreaded/multiprocess operation.
+
+    https://docs.python.org/3/library/exceptions.html#inheriting-from-built-in-exceptions
+    https://github.com/sqlfluff/sqlfluff/issues/6037
+    """  # noqa E501
+
+    def decorator(func: Callable[..., T]) -> Callable[..., T]:
+        def wrapped_method(*args, **kwargs) -> T:
+            # NOTE: `_detail` also acts as a semaphore to indicate whether an exception
```
barrywhart, line 122: Nit: Can we say "flag" rather than "semaphore" here? In Python, there's an actual class, [Semaphore](https://docs.python.org/3/library/threading.html#semaphore-objects), that is used in parallel programming, so the term could be confusing here, given that this decorator helps with dbt exception passing in parallel programming.

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -70,6 +72,103 @@ class DbtConfigArgs:
...
+                    # Any other errors are re-raised but only after stripping any
+                    # dbt context errors they may have acquired. This includes any
+                    # native SQLFluff errors.
+                    if is_dbt_exception(err.__context__):
+                        err.__context__ = None
+                    if is_dbt_exception(err.__cause__):  # pragma: no cover
+                        # This one seems to be less of an issue in testing, but I'm
+                        # keeping it in for completeness.
+                        err.__cause__ = None
+                    raise err
```
barrywhart, line 145: I notice that we're always doing `raise err`, even if the `__context__` or `__cause__` were not dbt exceptions. Doing it this way loses the original stack backtrace, which can be valuable for debugging. Could we simply `raise` in the case where `__context__` and `__cause__` were not dbt exceptions?

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
```diff
@@ -442,47 +545,11 @@ def process(
         self.profiles_dir = self._get_profiles_dir()
         fname_absolute_path = os.path.abspath(fname) if fname != "stdin" else fname
 
+        # NOTE: dbt exceptions are caught and handled safely for pickling by the outer
+        # `handle_dbt_errors` decorator.
         try:
             os.chdir(self.project_dir)
-            processed_result = self._unsafe_process(fname_absolute_path, in_str, config)
-            # Reset the fail counter
-            self._sequential_fails = 0
```
barrywhart, line 449: It looks like the logic for detecting multiple (sequential) failures was removed in this PR. Was that a deliberate change?

plugins/sqlfluff-templater-dbt/test/templater_test.py
```diff
@@ -490,8 +494,24 @@ def test__templater_dbt_handle_exceptions(
...
+    if exception_class is SQLTemplaterError:
+        _msg = excinfo.value.desc().replace("\\", "/")
+    else:
+        _msg = str(excinfo.value).replace("\\", "/")
+    assert exception_msg in _msg
+    # Ensure that there's no context parent exception, because they don't pickle well.
+    # https://github.com/sqlfluff/sqlfluff/issues/6037
+    # We *should* be stripping any inherited exceptions from anything returned here.
+    # Any residual dbt exceptions are a risk for pickling errors.
+    assert not excinfo.value.__context__
```
barrywhart, line 513: To make the test more robust and direct, could we have the test actually pickle and unpickle the exception to verify it doesn't crash?

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
alanmcruickshank: Yes - it didn't have test coverage, and all of the times that it used to be used don't seem to apply anymore.

plugins/sqlfluff-templater-dbt/test/templater_test.py
alanmcruickshank: Ohh that's a great idea 💯

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
alanmcruickshank: Good idea 👍

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
alanmcruickshank: I don't think this is true, I've done some testing, and it appears that `raise err` and `raise` (in the context of handling `err`) behave the same way and don't remove the stacktrace. That also seems in line with these docs: https://docs.python.org/3/reference/simple_stmts.html#grammar-token-python-grammar-raise_stmt

Am I misunderstanding what you mean?

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
alanmcruickshank: Yeah, looking back at when those sequential fails were introduced, it was to solve #1088 (with PR #1106, which built on #1044). That related to the early days of parallel linting for dbt, which I think we now have more elegant and robust mechanisms for.

plugins/sqlfluff-templater-dbt/sqlfluff_templater_dbt/templater.py
barrywhart: Here's an example of the difference. (I'm using Python 3.8 below.)

Script #1 just uses `raise`.
```
def bah():
    1 / 0


def foo():
    try:
        bah()
    except Exception as e:
        raise


if __name__ == "__main__":
    foo()
```

Output:
```
  File "raise_example.py", line 13, in <module>
    foo()
  File "raise_example.py", line 7, in foo
    bah()
  File "raise_example.py", line 2, in bah
    1 / 0
ZeroDivisionError: division by zero
```

Script #2 uses `raise e`.
```
def bah():
    1 / 0


def foo():
    try:
        bah()
    except Exception as e:
        raise e


if __name__ == "__main__":
    foo()
```

Output:
```
  File "raise_example.py", line 13, in <module>
    foo()
  File "raise_example.py", line 9, in foo
    raise e
  File "raise_example.py", line 7, in foo
    bah()
  File "raise_example.py", line 2, in bah
    1 / 0
ZeroDivisionError: division by zero
```

I see you're right -- in both cases, line 2 appears in the stack backtrace. There _is_ a difference, i.e. whether we see line 9 in the backtrace. `raise e` includes it, `raise` does not. Not a huge difference.

I'm pretty sure in past Python versions (maybe years ago), `raise e` would've lost line 2 from the stack trace. Likely the reason we see it now is related to the things you're dealing with here -- the additional exception fields such as `__cause__` and `__context__`. In theory, it would be cleaner to use `raise` if it's not a dbt exception, but there seems to be no practical difference anymore.

alanmcruickshank: iiiinteresting! good sleuthing. Agreed that the difference is minor at this point, but worth knowing the difference. 👍

barrywhart: I haven't checked to confirm this, but Perplexity (AI search engine) provides a lot more context:

You're remembering correctly about the past behavior, but there have been some changes over Python versions regarding exception chaining and reraising. Let's break this down:

### Past Behavior (Python 2 and early Python 3)

In earlier versions of Python, there was indeed a distinction between:

1. `raise` (without arguments)
2. `raise e` (where e is the caught exception)

The first form (raise without arguments) would reraise the exception while preserving the original traceback and context. The second form (raise e) would create a new exception instance, losing the original traceback.

### Change in Python 3

Python 3 introduced significant changes to exception handling, including the concept of exception chaining.

In Python 3.0-3.2, `raise` and `raise e` behaved differently:
1. `raise` preserved the traceback and context
1. `raise e` created a new exception, but set `__context__` to the original exception

### Python 3.3 and Later (including 3.8)

From Python 3.3 onwards, the behavior was unified:
* Both `raise` and `raise e` preserve the exception context
* Both forms maintain the traceback of the original exception

This change was part of PEP 409, which aimed to simplify exception chaining and make it more consistent.

### Conclusion

Your observation about Python 3.8 is correct - both raise and raise e preserve the underlying context. This behavior has been consistent since Python 3.3.

The key change occurred in Python 3.3, released in 2012. Since then, the distinction between these two forms of reraising exceptions has been largely eliminated in terms of preserving context and traceback.

It's worth noting that while both forms preserve context, you can still explicitly control exception chaining using `raise ... from ... syntax` introduced in Python 3.

PR #6144, opened 2024-09-04
More robust exception handling for dbt.

alanmcruickshank:
This is part of #6138, but the uncontroversial part. This effectively contains only quality of life improvements around dbt exceptions.

This is loosely related to #6037, but doesn't resolve it on it's own.


WittierDinosaur (approved):
LGTM

PR #6146, opened 2024-09-05
ST08: Ignore `DISTINCT`s with subqueries

### LINKED ISSUE #6142: Necessary brackets for unnest function in BigQuery are removed
### Search before asking

### What Happened

I’ve encountered another issue that seems specific to BigQuery involving unnesting within a function. BigQuery requires brackets in this scenario, but SQLFLUFF removes them

### Expected Behaviour

The bracket in this case should not be removed in order not to introduce an error

### Observed Behaviour

I get the attached error when the brackets are removed
Screenshot 2024-09-04 at 8 45 45 AM

### How to reproduce

I ran the sqlfluff fix command. Below is the query before and after running the command

Before query:
```
select
    COUNT(
        DISTINCT
        (
            SELECT ANY_VALUE(id)
            FROM UNNEST(tag) t
        )
    )
from
    dataset_name.table_name;
```
After query:
```
SELECT
    COUNT(
        DISTINCT
        SELECT ANY_VALUE(id)
        FROM UNNEST(tag) t

    )
FROM
    dataset_name.table_name;
```

### Dialect

BigQuery

### Version

3.1.1

### Configuration

```cfg
[sqlfluff]

# Or run 'sqlfluff dialects'
dialect = bigquery

templater = jinja
ignore = templating

exclude_rules = ambiguous.column_count, structure.column_order, structure.using, capitalisation.identifiers, structure.subquery, structure.join_condition_order

# something a little longer.
max_line_length = 180

processes = 0

[sqlfluff:indentation]
indented_joins = True
indented_using_on = True
allow_implicit_indents = True
indented_on_contents = True
indented_ctes = True
indented_then = True

[sqlfluff:rules:layout.set_operators]
set_operator_on_new_line = ['UNION', 'UNION ALL']

[sqlfluff:rules:convention.statement_brackets]
convention.statement_brackets = True
[sqlfluff:convention.quoted_literals]
force_enable = True
preferred_quoted_literal_style = double_quote

aliasing = implicit
aliasing = explicit
# [sqlfluff:rules:aliasing.length]
# min_alias_length = 10

capitalisation_policy = upper
capitalisation_policy = upper
```

### Code of Conduct

keraion:

This ignores `DISTINCT`s with brackets which contain subquery expressions for ST08.
- fixes #6142

None, performance hit _should_ be minimal.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/structure/ST08.py
+++ b/src/sqlfluff/rules/structure/ST08.py
@@ -82,6 +82,8 @@ class Rule_ST08(BaseRule):
                 not function_name
                 or function_name[0].raw_upper != "DISTINCT"
                 or not bracketed
+                # If the DISTINCT has a subquery, don't remove the brackets
+                or bracketed.recursive_crawl("select_statement", recurse_into=False)
             ):
                 return None
             # Using ReflowSequence here creates an unneeded space between CONCAT
--- a/test/fixtures/rules/std_rule_cases/ST08.yml
+++ b/test/fixtures/rules/std_rule_cases/ST08.yml
@@ -56,3 +56,15 @@ test_fail_distinct_concat_inside_count:
     SELECT COUNT(DISTINCT(CONCAT(col1, '-', col2, '-', col3)))
   fix_str: |
     SELECT COUNT(DISTINCT CONCAT(col1, '-', col2, '-', col3))
+
+test_pass_distinct_subquery_inside_count:
+  pass_str: |
+    SELECT
+        COUNT(
+            DISTINCT(
+                SELECT ANY_VALUE(id)
+                FROM UNNEST(tag) t
+            )
+        )
+    FROM
+        dataset_name.table_name;


alanmcruickshank:
> Hi Keraion, Thanks for addressing this bug. I need to highlight that this issue is not related to distinct, it's the same case without the distinct as well, meaning that the brackets are required even if there is no distinct keyword, not sure if that makes a difference on your solution.

I think this should be ok, because the rule we're addressing is only triggered when there's a `DISTINCT` present.

However... I have a different question here and that's whether this should be a BigQuery specific exception. I've just tested in snowflake and this query compiles fine:

```sql
select count(distinct select any_value($1) from values (1, 2), (3, 4))
```

keraion:
I just tested in postgres and db2, and they both require the parentheses. Seems like snowflake might be exception.

alanmcruickshank:
> I just tested in postgres and db2, and they both require the parentheses. Seems like snowflake might be exception.

Fair enough. Thanks for checking 👍

alanmcruickshank (approved):
LGTM 👍

PR #6153, opened 2024-09-07
Auto generate dialect docs

alanmcruickshank:
This takes what we did for rules and applies it to dialects.

I've been doing a fair bit of docs development recently and I realised that we were maintaining a list of dialects in `dialects.rst` which didn't necessarily map to the actual dialects themselves. This removes that manual list, and instead adds an auto-generated docs section powered by the dialects themselves. That removes a duplication, but it also has the upshot that all the config is now in the dialects, and that we can use the nicely formatted names in the CLI too.

Here's a preview of the docs it generates now:
image

...compared to the _current_ docs:
image

...and here's the new CLI view with this PR:
image

...compared to the old CLI view (I admit, not a massive change, but a little nicer):
image


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -44,7 +44,16 @@
...
-postgres_dialect = ansi_dialect.copy_as("postgres")
+postgres_dialect = ansi_dialect.copy_as(
+    "postgres",
+    formatted_name="PostgreSQL",
+    docstring="""This is based around the `PostgreSQL spec`_. Many other SQL
+dialects are often based on the PostreSQL syntax. If you're running an unsupported
+dialect, then this is often the dialect to use (until someone makes a specific
+dialect).
+
+.. _`PostgreSQL spec`: https://www.postgresql.org/docs/9.6/reference.html""",
```
keraion, line 55: This should probably reference the most current docs.
```suggestion
.. _`PostgreSQL spec`: https://www.postgresql.org/docs/current/reference.html""",
```

alanmcruickshank: Good idea 👍

keraion (approved):
LGTM

PR #6162, opened 2024-09-08
Better documentation on how to cross reference rules and fix a few.

alanmcruickshank:
The docs for AL06 and AL07 have (in theory) links to other rule docs in them. Those links are currently broken, but not surprisingly because the details on how to cross reference rules is poorly documented.

This PR fixes the links and improves the documentation.


keraion (approved):
LGTM

PR #6163, opened 2024-09-08
Update Docs and tests for CP02

alanmcruickshank:
As part of resolving the conflict around RF06 and AL09, there's some influence from CP02. The docs for this rule were fairly poor, but the functionality is good I think.

This PR significantly updates the docs and adds/improves a few of the test cases.


keraion (approved):
LGTM

PR #6164, opened 2024-09-08
Cached property in RF06 rather than DIY

alanmcruickshank:
This is a very small method to simplify a section of RF06. There's a section here which effectively re-implements `functools.cached_property`. This PR updates that to just use `functools.cached_property`, and remove the complexity.


keraion (approved):
LGTM

PR #6165, opened 2024-09-08
More dialect documentation

alanmcruickshank:
After #6153, it's more obvious that some of the dialects have more extensive documentation than others. This starts to flesh out that documentation. I originally wanted to be more complete here, but some dialects were much easier to test and find than others.

I've added details for as many dialects as I'm either familiar with or have direct access to for testing. I've also added links where I can, especially where dialects do interesting things.

I've also started a bit of a convention (which I hope other dialects will follow) on what kinds of things to document.

(I also fixed a bug in the Postgres dialect which I found on the way through)


keraion (approved):
LGTM

PR #6044, opened 2024-08-06
RF06: Allow case sensitive quoted aliasing

aayushr7:
### Brief summary of the change made

Allow quoting of columns when the column name needs to be case sensitive.

- fixes #6030 

### Are there any other side effects of this change that we should be aware of?
No

### Pull Request checklist
- [X] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #3260, opened 2022-05-04
Bug fix: Disable L059 in snowflake dialect

barrywhart:
### Brief summary of the change made
Fixes #3173 

Implements the solution discussed in the issue -- disabling this rule by default in Snowflake (and adding support for `force_enable`).

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

Issue #3173, opened 2022-04-25
Snowflake - L059 corrupts SQL on column names created with quotes


WittierDinosaur:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

sqlfluff fix creates unparseable sql if "date" is quoted

```
SELECT d."date" AS date_key
FROM foo.dates d
;
```
becomes
```
SELECT d.date AS date_key
FROM foo.dates d
;
```

Snowflake docs explicitly state date is not a reserved keyword, but it appears it actually is

### Expected Behaviour

-

### Observed Behaviour

-

### How to reproduce

-

### Dialect

Snowflake

### Version

0.13.0

### Configuration

-

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

barrywhart:
@tunetheweb: Do you have any ideas about this? It looks like you've done some work on this rule, and this seems like an issue that may also relate to dialect things such as identifier types and keywords.

tunetheweb:
@WittierDinosaur already has the reason as he stated above:

> Snowflake docs explicitly state date is not a reserved keyword, but it appears it actually is

L059 will remove unnecessary quotes and quotes are unnecessary if it’s not a reserved keyword.

I agree [the Snowflake docs](https://docs.snowflake.com/en/sql-reference/reserved-keywords.html) appear to be incorrect here since they state:

> **Reserved & Limited Keywords**
> 
> Snowflake SQL reserves all ANSI keywords (with the exception of type keywords such as CHAR, DATE, DECIMAL, etc.),

So, easiest answer is to make `DATE` a reserved word - as it appears to be.

WittierDinosaur:
This is actually wrong - my bad. I did some more digging on this, and my initial assumption was wrong. It turns out `date` is not a reserved keyword. What's actually happening is that quoted columnar identifiers are different from their unquoted equivalents:
```
CREATE TEMP TABLE foo
("hello" string);

SELECT hello FROM foo; -- This fails
SELECT "hello" FROM foo; -- This succeeds
```
I think the solution here may be to turn off L059 for Snowflake

tunetheweb:
[...]

PR #6173, opened 2024-09-09
RF06: Case Sensitivity

### LINKED ISSUE #6030: RF06: Unnecessary Quoted Identifier error when column name needs to be case-sensitive
### Search before asking

### What Happened

When the column name is quoted so that it's case sensitive, the RF06 rule marks the quotes as unnecessary. 

Example Query:
```
SELECT casechange AS "CaseChange"
FROM foo
```

Fix Output:
```
SELECT casechange AS CaseChange
FROM foo
```

### Expected Behaviour

The rule should not fail.

### Observed Behaviour

```
L:   1 | P:  22 | RF06 | Unnecessary quoted identifier "CaseChange".
                       | [references.quoting]
```

### How to reproduce

<details> <summary> test.sql </summary>

```
SELECT casechange AS "CaseChange"
FROM foo
```

</details>

`sqlfluff lint test.sql --dialect ansi`
`sqlfluff fix test_alias.sql --dialect ansi --rules RF06`

### Dialect

ansi

### Version

```
sqlfluff, version 3.1.0
Python 3.11.7
```

### Configuration

NA

### Code of Conduct

alanmcruickshank:
This PR resolves #6030, supersedes #6044 and should also fulfil the original intent of #3173 (by effectively reverting #3260).

As per the conversation on slack, there were lots of issues with RF06:
- The docs weren't particularly clear, and made it look like it was only for aliases.
- It wasn't case sensitive (an issue raised by #3173, but which we "fixed" by disabling the rule for snowflake!). The root cause here was a fundamental misunderstanding of the case sensitivity of identifiers in databases.

This PR changes quite a lot:
- The docs are now much more extensive.
- I've removed the `force_enable` config value (and enabled this for snowflake and postres again).
- I've added a now `case_sensitive` config which defaults to `True` (users can override that to `False` to get the old behaviour).
- I've added a bunch more test cases.
- I've also done a small refactor of the way that we can access the `casefold` attribute of parsers, and how we instantiate segments. Should be minor though.

Would appreciate a review from @keraion and @aayushr7 who've been involved with the discussion to get to this level of clarity.

**NOTE:** This PR should trigger a minor release because it changes rule behaviour and config.


alanmcruickshank:
BTW: Who knew that DuckDB and Spark were both totally case insensitive!?

test/fixtures/rules/std_rule_cases/RF06.yml
```diff
@@ -1,104 +1,106 @@
...
+  # as long as the identifier is *valid* without quotes, we should
+  # remove the quotes.
   fail_str: |
-    SELECT foo
-    FROM [bar].[baz];
+    SELECT
+      "lowercase_reference",
+      "UPPERCASE_REFERENCE",
+      foo AS "column_alias",
+      "Case_Sensitive_Reference",
```
aayushr7, line 62: I guess this should be `Case_Insensitive_Reference`

alanmcruickshank: I see that it's ambiguous 🤔 . My intent here was a reference where the case matters, and therefore can't be unquoted. I'll change it to `Case_Specific_Reference`.

aayushr7: Oh! But the fix is removing the quotes since DuckDB is insensitive, which seems correct.

aayushr7 (approved):

keraion (approved):
LGTM. The documentation now definitely clears up the intent of this rule.

PR #6177, opened 2024-09-09
Deprecate the `ConfigLoader`

alanmcruickshank:
The `ConfigLoader` class has been a weird oddity for a while. It had many methods for loading config, none of which really needed to be in a class, it implemented a caching layer which we could just outsource to other libraries, and it used a `global` in a slightly hacky way that I was never really sure whether worked.

In any case, I think now is the time to break it up (partly because it makes it simpler to finish off #6130. Critically we migrate to using `functools.cache` for caching the results which totally gets rid of the need to juggle a `global` class to try and achieve the same thing.

This looks like a big PR (albeit turning a few very large modules into several much smaller ones), but it does a few key things:
- Most of the methods that were available on the `ConfigLoader` are now available as functions of the same name which can be imported directly. The functionality of each of them is exactly the same.
- The `ConfigLoader` class itself is still present, with a selection of deprecation warnings on it, for any plugins which still import it. It was present in our example plugin for a long while, so there's a chance that other users are still using it.
- It pulls out the `toml` and `ini` config loading routines into their own modules so they can be used more widely.
- It pulls out the validation of removed configs into the `removed` module which feels like a better home.
- Switches from using the tuple-based representation of configs so much (and a lot of the switching back and forth between records and dicts). There's still a bit of that, but most of the methods (and importantly now most of the validation ones), operate on dicts now.

Because of the deprecation in here, this should trigger a minor release.

This also supersedes #6133.


src/sqlfluff/core/config/__init__.py
```diff
@@ -5,16 +5,47 @@
...
 )
 
 
+def clear_config_caches():
+    """Clear any of the cached config methods.
+
+    This is primarily used during testing where the cache may be be rendered unreliable
+    by using moving around files while setting up tests. Some of the cached methods
+    rely on *filename* caching, and so we may break one of the assumptions of the
+    caching routines (that files aren't modified while SQLFluff is running) during
```
WittierDinosaur, line 41: is the final sentence unfinished?

src/sqlfluff/core/config/file.py
```diff
@@ -0,0 +1,133 @@
...
+
+from sqlfluff.core.config.ini import load_ini_string
+from sqlfluff.core.config.toml import load_toml_file_config
+from sqlfluff.core.config.types import ConfigMappingType
+from sqlfluff.core.config.validate import validate_config_dict
+from sqlfluff.core.helpers.string import (
+    split_comma_separated_string,
+)
+
+if sys.version_info >= (3, 9):
```
WittierDinosaur, line 26: happy to keep this for now, but 3.8 hits EOL in October, we should consider dropping support soon after

src/sqlfluff/core/config/file.py
```diff
@@ -0,0 +1,133 @@
...
+    cache = lru_cache(maxsize=None)
+
+
+COMMA_SEPARATED_PATH_KEYS = ("load_macros_from_path", "loader_search_path")
+RESOLVE_PATH_SUFFIXES = ("_path", "_dir")
+
+
+def _load_raw_file_as_dict(filepath: str) -> ConfigMappingType:
+    """Loads the raw dict object from file without interpolation."""
+    _, file_extension = os.path.splitext(filepath)
```
WittierDinosaur, line 46: unused variable?

src/sqlfluff/core/config/ini.py
```diff
@@ -0,0 +1,86 @@
...
+    # Try to coerce it to a more specific type,
+    # otherwise just make it a string.
+    try:
+        v: Any = int(val)
+    except ValueError:
+        try:
+            v = float(val)
+        except ValueError:
+            cleaned_val = val.strip().lower()
+            if cleaned_val in ["true"]:
```
WittierDinosaur, line 24:
```suggestion
            if cleaned_val == "true":
```

src/sqlfluff/core/config/ini.py
```diff
@@ -0,0 +1,86 @@
...
+                v = True
+            elif cleaned_val in ["false"]:
```
WittierDinosaur, line 26:
```suggestion
            elif cleaned_val == "false":
```

src/sqlfluff/core/config/ini.py
```diff
@@ -0,0 +1,86 @@
...
+                v = False
+            elif cleaned_val in ["none"]:
```
WittierDinosaur, line 28:
```suggestion
            elif cleaned_val == "none":
```

src/sqlfluff/core/config/__init__.py
alanmcruickshank: ...yeah. Good catch.

src/sqlfluff/core/config/file.py
alanmcruickshank: Oh - excellent catch

PR #6179, opened 2024-09-10
Add volume syntax support for Databricks

### LINKED ISSUE #6178: Databricks dialect does not support VOLUME syntaxes
### Search before asking

### What Happened

When linting sql in the Databricks dialect that has commands involving VOLUMEs, sqlfluff gives an "unparsable section" error.

### Expected Behaviour

sqlfluff will correctly lint commands involving VOLUMEs.

See: https://docs.databricks.com/en/sql/language-manual/sql-ref-volumes.html

### Observed Behaviour

An error like the following is generated:
`Found unparsable section: 'CREATE VOLUME IF NOT EXISTS bronze.track...'

### How to reproduce

Attempt to lint the following. It is known valid and works in Databricks

create_alter_volume.sql
```sql
CREATE VOLUME IF NOT EXISTS bronze.tracking;
ALTER VOLUME bronze.tracking SET OWNER TO `admins`;
```

### Dialect

Databricks

### Version

sqlfluff, version 3.1.1
Python 3.10.13

### Configuration

.sqlfluff
```toml
[sqlfluff]
verbose = 2
dialect = databricks
templater = placeholder
max_line_length = 120
exclude_rules = CP02, ST06
warnings = RF03

[sqlfluff:indentation]
tab_space_size = 2
indented_joins = True
indented_ctes = True
indented_using_on = True
indented_on_contents = True
# indented_then = True
# indented_then_contents = True
allow_implicit_indents = True
template_blocks_indent = True

capitalisation_policy = upper

```

### Code of Conduct

TheCleric:

Add support for VOLUME[S] syntax in Databricks. Fixes #6178 

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_databricks.py
+++ b/src/sqlfluff/dialects/dialect_databricks.py
@@ -319,6 +319,125 @@ class AlterDatabaseStatementSegment(sparksql.AlterDatabaseStatementSegment):
     )
 
 
+class VolumeReferenceSegment(ansi.ObjectReferenceSegment):
+    """Volume reference."""
+
+    type = "volume_reference"
+
+
+class AlterVolumeStatementSegment(BaseSegment):
+    """Alter Volume Statement.
+
+    https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-alter-volume.html
+    """
+
+    type = "alter_volume_statement"
+
+    match_grammar = Sequence(
+        "ALTER",
+        "VOLUME",
+        Ref("VolumeReferenceSegment"),
+        OneOf(
+            Sequence(
+                "RENAME",
+                "TO",
+                Ref("VolumeReferenceSegment"),
+            ),
+            Ref("SetOwnerGrammar"),
+            Ref("SetTagsGrammar"),
+            Ref("UnsetTagsGrammar"),
+        ),
+    )
+
+
+class CreateVolumeStatementSegment(BaseSegment):
+    """Create Volume Statement.
+
+    https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-volume.html
+    """
+
+    type = "create_volume_statement"
+
+    match_grammar = OneOf(
+        # You can create a non-external volume without a location
+        Sequence(
+            "CREATE",
+            "VOLUME",
+            Ref("IfNotExistsGrammar", optional=True),
+            Ref("VolumeReferenceSegment"),
+            Ref("CommentGrammar", optional=True),
+        ),
+        # Or you can create an external volume that must have a location
+        Sequence(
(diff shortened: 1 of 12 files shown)


src/sqlfluff/dialects/dialect_databricks.py
```diff
@@ -319,6 +319,125 @@ class AlterDatabaseStatementSegment(sparksql.AlterDatabaseStatementSegment):
...
+            "VOLUME",
+            Ref("IfNotExistsGrammar", optional=True),
+            Ref("VolumeReferenceSegment"),
+            Ref("LocationGrammar"),
+            Ref("CommentGrammar", optional=True),
+        ),
+    )
+
+
+class DescribeVolumeStatementSegment(BaseSegment):
```
WittierDinosaur, line 383: We generally don't go this granular on statement separation. We have a `DescribeStatementSegment` and `ShowStatementSegment`, which covers all of the describes and shows. You'll want to add to this, there is one of each defined in Spark

TheCleric: Should I copy each of those into the Databricks dialect and add these?

Only reason I didn't was so I wasn't duplicating the existing sparksql show/describe statements, but rather do a pseudo-extension of them. But if we'd prefer to copy over + modify I can do that as well.

TheCleric: Also this may be an item for future cleanup, but looking at the dialect_sparksql.py files, it looks like the describe behavior is broken out a few different ways (e.g., `DESCRIBE HISTORY` and `DESCRIBE DETAIL`).

This isn't directly relevant to how we want to handle this case, but may want to be standardized in the future to avoid confusion.

alanmcruickshank: There's a syntax for "copy and modify" which should be what you're looking for. It can't do everything, but it can do a certain amount of insertions and removals. Do a `crtl+f` for `.copy(` in a few of the dialect files and you'll see how it's used in various ways.

TheCleric: Okay I think I've done it in a more standardized way. I had to convert a few things to grammars as you can copy and insert new options into those, whereas segment's copy does not support inserts. So hopefully I've done it correctly.

TheCleric: I also found some of the show commands in Databricks are SLIGHTLY different from their Spark lineage, so I've updated those while there and documented as well as I can.

WittierDinosaur (approved):
LGTM - thanks for the refactor

TheCleric:
@WittierDinosaur can you try to remerge this when you can now that I've resolved conflicts.

Issue #5782, opened 2024-04-19
MySQL CRITICAL Exception AL09 


Xemptuous:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Wanting to integrate this into my repo, and it threw many errors. Ran it on single files too, and even after running fix on them all, I still get this error.

### Expected Behaviour

I wanted to be able to get proper fixes and lints going, but this error looked tragic and concerning enough to stop me.

### Observed Behaviour

```
CRITICAL   [AL09] Applying rule AL09 to 'declined_accounts.sql' threw an Exception:
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/sqlfluff/core/rules/base.py", line 507, in crawl
    res = self._eval(context=context)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/sqlfluff/rules/aliasing/AL09.py", line 99, in _eval
    assert whitespace and column_identifier and alias_identifier
```
Same error happens when I use a pypi venv version instead of my locally installed one (arch).

### How to reproduce

`sqlfluff lint .` with the specified config, or using `sqlfluff lint --dialect mysql`. Error happened on most sql files in the repo (hundreds)

### Dialect

MySQL (for mariadb)

### Version

sqlfluff, version 3.0.4
Python 3.11.8

### Configuration

```toml
[tool.sqlfluff.core]
verbose = 0
nocolor = false
dialect = "mysql"
runaway_limit = 10
max_line_length = 120
large_file_skip_char_limit = 0
large_file_skip_byte_limit = 50000
processes = -1
sql_file_exts = ".sql,.sql.j2,.dml,.ddl"

[tool.sqlfluff.rules]
allow_scalar = true
single_table_references = "consistent"
unquoted_identifiers_policy = "all"

[tool.sqlfluff.indentation]
indent_unit = "space"
tab_space_size = 4
indented_joins = true
indented_ctes = true
indented_on_contents = true
indented_using_on = true
template_blocks_indent = false
allow_implicit_indents = true

[tool.sqlfluff.rules.capitalisation.keywords]
capitalisation_policy = "upper"

[tool.sqlfluff.rules.capitalisation.identifiers]
extended_capitalisation_policy = "snake"
unquoted_identifiers_policy = "all"

[tool.sqlfluff.rules.capitalisation.functions]
[...]

PR #6186, opened 2024-09-12
Handle errors better in AL09

alanmcruickshank:
Bug reports like #5782 suggest that we're making assumptions in AL09 which aren't always true. Without better logging information we can't _fix_ them, but we can definitely make the user experience better by not raising an exception, and also by producing better logging and debug information.

This sorts that. No coverage because I don't actually know what will trigger this yet - just that it is definitely being triggered.


WittierDinosaur (approved):
LGTM

PR #6193, opened 2024-09-13
DuckDB: Support functions with walrus operators

### LINKED ISSUE #6189: "fix" breaks SQL / "lint" false positive on struct_pack := operator DuckDB
### Search before asking

### What Happened

Given DuckDB struct_pack notation with the `:=` operator

* ` sqlfluff lint --dialect duckdb false-positive-and-broken-fix.sql` will emit false positive diagnostics RF03
* ` sqlfluff fix --dialect duckdb false-positive-and-broken-fix.sql` will break the SQL

See https://duckdb.org/docs/sql/data_types/struct.html#creating-structs for docs:

>Structs can be created using the [struct_pack(name := expr, ...)](https://duckdb.org/docs/sql/functions/nested.html#struct-functions) function, the equivalent array notation {'name': expr, ...}, using a row variable, or using the row function.
>
> Create a struct using the struct_pack function. Note the lack of single quotes around the keys and the use of the := operator:

### Expected Behaviour

No lint output; fix not breaking the SQL.

### Observed Behaviour

On the `lint` part
```
L:  10 | P:   9 | RF03 | Unqualified reference 'val' found in single table
                       | select. [references.consistent]
L:  10 | P:   9 | RF03 | Unqualified reference 'val' found in single table select
                       | which is inconsistent with previous references.
                       | [references.consistent]
```

On the `fix` part, sqlfluff will generate
```
    struct_pack(
        t.val := t.val      -- <--- bad LHS
    ) as s
```
instead of the expected
```
    struct_pack(
        val := t.val
    ) as s
```
The LHS of the `:=` operator is not a reference, but a target 

### How to reproduce

Given the SQL below as `false-positive-and-broken-fix.sql`
```
create table t (
    id int,
    val int
);

create view v as
select
    t.id,
    struct_pack(
        val := t.val
    ) as s
from
    t;
```

* ` sqlfluff lint --dialect duckdb false-positive-and-broken-fix.sql`
* ` sqlfluff fix --dialect duckdb false-positive-and-broken-fix.sql`

### Dialect

duckdb

### Version

sqlfluff, version 3.1.1 - which is not correct; this is in fact `pip install --upgrade https://github.com/sqlfluff/sqlfluff/tarball/master` from 2024-09-10 11:51:57.063596941 +00:00:00  (i.e. a relatively recent HEAD of master)

### Configuration

none

### Code of Conduct

keraion:

This adds support for functions that use a `walrus operator` in DuckDB.
- fixes #6189

None

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_duckdb.py
+++ b/src/sqlfluff/dialects/dialect_duckdb.py
@@ -340,6 +340,21 @@ class ColumnsExpressionFunctionContentsSegment(
     )
 
 
+class NamedArgumentSegment(postgres.NamedArgumentSegment):
+    """Named argument to a function.
+
+    Some functions may use a `walrus operator`.
+    e.g. https://duckdb.org/docs/sql/functions/struct#struct_packname--any-
+    """
+
+    type = "named_argument"
+    match_grammar = Sequence(
+        Ref("NakedIdentifierSegment"),
+        OneOf(Ref("RightArrowSegment"), Ref("WalrusOperatorSegment")),
+        Ref("ExpressionSegment"),
+    )
+
+
 class LambdaExpressionSegment(BaseSegment):
     """Lambda function used in a function or columns expression.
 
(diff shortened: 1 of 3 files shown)


WittierDinosaur (approved):
LGTM

PR #6206, opened 2024-09-16
Strict typing in `config` and `helpers`.

alanmcruickshank:
This brings the `core.helpers` and `core.config` modules up to string mypy typing. Also helpfully it does a little bit of type narrowing so that any lists in config mappings can only contain strings. Practically this doesn't change downstream functionality (as they should only be lists anyway), but it makes type handling downstream simpler.


keraion (approved):
LGTM

PR #6213, opened 2024-09-18
New Rule LT14: Keyword line positioning

### LINKED ISSUE #1247: add rule to enforce `FROM`, `JOIN` and `ON` statements to be on a separate line
Currently none of the rules enforce any behavior with `FROM`, `JOIN` and `ON` statements. For example, the below are currently acceptable:
```
FROM a
JOIN b ON a.id = b.id
```
```
FROM a JOIN
 b ON a.id = b.id
```

It would be nice to have a rule that enforces `FROM` and `JOIN` to be on new lines, and `ON` to be indented on a new line:
```
FROM a
JOIN b
  ON a.id = b.id
```

Similarly, this rule (or a separate rule) could enforce `AND`s in a join to be indented on a new line:
```
FROM a
JOIN b
  ON a.id = b.id
  AND a.id2 = b.id2
```
instead of 
```
FROM a
JOIN b
  ON a.id = b.id AND a.id2 = b.id2
```

keraion:

This is the first steps on building out keyword positioning rules. ~~Currently, this sets no defaults in the default configuration. We may look at changing that in the future.~~ The following clauses now have a default `leading` configuration:  `WHERE`, `ORDER BY`, `GROUP BY`, and `HAVING`. 

I wanted to get some buy in from some others before going too far with this implementation.

This also reuses a number of the rebreak functions, but were distinct enough that it didn't make sense to combine at this time, also something to consider for the future.

Additionally, as part of this PR, documentation ~~will need to be~~ has been fleshed out, so users are able to add additional elements/types as needed.

- fixes #1247

This now currently sets a default for the following clauses: `WHERE`, `ORDER BY`, `GROUP BY`, and `HAVING`. Documentation to disable this default behavior has been included in the layout config section.

### Pull Request checklist

  - Other.


alanmcruickshank:
@keraion - well done for taking this one on! I don't think you need quite so much custom code, as all the basic operations are there in the reflow code, but they probably need extracting as separate methods.

In essence, all the logic is already there, it's just that at the moment it's only _triggered_ by long lines. For example if you lint this file:
```sql
-- sqlfluff:max_line_length:12
select * from a where b group by c order by d;
```

...then LT05 will reformat it to this:
```sql
-- sqlfluff:max_line_length:12
select *
from a
where b
group by c
order by d;
```

What I think we'd want to do with this rule is somehow force the reflow code to do that _even when the line isn't too long_. I'll have a look to see if I can find what the easiest way to do that might be...

alanmcruickshank:
I think we basically want to do what `lint_line_length` does, but simpler, and eventually doing something very like `_fix_long_line_with_fractional_targets` does, although that's a kind of helper method in the background at the moment.

alanmcruickshank:
hoooooold up. Just re-read your test cases properly. VERY INTERESTING. You've actually implemented something much more advanced that what I originally expected. Very cool.

alanmcruickshank:
@keraion - this is looking good! I really like the functionality here. Anything I can help with to help you finish it off?

keraion:
@alanmcruickshank I think I'm about done with this. I have been trying to find any other edge cases on how some items were being split, but most of those original issues have since been resolved. I have a feeling that we'll run into a few keywords that will have issues, but as this must be currently configured for any rules to apply the risk at the moment is very low.

Other than that, I don't know if we want to include more clarifications in the documentation and/or an example configuration file to go with this.

alanmcruickshank:
@keraion - For launch, can I suggest that we add the config to force them main select keywords (`SELECT`, `WHERE`, `ORDER BY`, `GROUP BY`, `HAVING`) to be leading on newlines? That seems to be what most people ask for.

Don't set a default config for `FROM` because that will conflict with `LT09`. I'm tempted to also not set it for `SELECT` too - but let's see what happens when we do.

In the [layout config docs](https://github.com/sqlfluff/sqlfluff/blob/main/docs/source/configuration/layout.rst) we should explain the new features, and I'd suggest using your config for `PARTITION BY` (which you put together in the tests) as a great example to use. I don't think that should be part of the default config, but I'm sure some users will want it. I think mirroring the level of detail used for the existing layout configuration options is a good guide for how much documentation we should add for your new options.

In the meantime, I think it's time I took this for a spin on a project!

src/sqlfluff/utils/reflow/rebreak.py
```diff
@@ -577,3 +677,206 @@ def rebreak_sequence(
...
+            # Generate the text for any issues.
+            pretty_name = loc.pretty_target_name()
+            if loc.strict:  # pragma: no cover
+                # TODO: The 'strict' option isn't widely tested yet.
+                desc = f"{pretty_name.capitalize()} should always start a new line."
+            else:
+                desc = (
+                    f"Found trailing {pretty_name}. Expected only leading "
+                    "near line breaks."
+                )
```
alanmcruickshank, line 766: See my note below about this part.

alanmcruickshank (commented):
Ok - I've taken this for a spin with the default settings on my current big project, and I like it!

In general, it works as expected. Interestingly, by putting in a default for the `ORDER BY` clause, it automatically splits partition statement too (because that's also an `ORDER BY` clause). Initially I thought this would be bad, but I quite like it!

image

My main feedback is that I think we need a better `description` on the `LintingResult`. This is the current output on a messy project:

image

I think it's currently really unclear what the user should do about it. A better output would be something like:

> Expected line break before `ORDER` keyword.

...or similar.

One thing to think through, is that it appears we're in strict mode by default. I don't think that's necessarily a bad thing - but it does mean we should probably take out the `if loc.strict` clause. I'll add a note to the logic.

alanmcruickshank:
@Jaxsss, @yannik20, @avijit3660, @eetu-h, @claytonjhamilton, @danny-dailypay, @puetsch, @jessie-carta

As people that have expressed interest in #1247. Would any of you be up for taking this branch for a spin on your projects? You can install SQLFluff from a git branch using `pip install git+https://github.com/sqlfluff/sqlfluff.git@keraion/LT14#egg=sqlfluff`. Would appreciate reviews from any of you about the functionality here and whether it achieves what you were looking for.

I think feedback on the rule defaults would also be very helpful.

WittierDinosaur:
As discussed in the meeting, here's an example of a statement which I'd like the ability to bind to one line
```
SET increment_key = (SELECT MAX(COALESCE(date_key, '2021-01-01') - 1) AS increment_key FROM foo.bar);
```

jayctran:
This is fantastic functionality - have started using this and it's linting exactly how we want it

alanmcruickshank:
Update on this PR from maintainers call: We'll target getting this rule into the next release. Last action is currently on me to add the exception for single line queries. Otherwise good to go.

alanmcruickshank:
@WittierDinosaur - I've added your example query as a test case - and it's actually fine! I'm pretty sure that because there's only a SELECT and FROM (and none of the rest), the rule doesn't try and force it onto more than one line. Does that mean we don't need to do any more work on this?

If you're able to take it for a spin on your codebase that would be a great check to see if we're good to go.

src/sqlfluff/utils/reflow/rebreak.py
```diff
@@ -263,8 +263,107 @@ def identify_rebreak_spans(
...
+            "keyword"
+        )
+    else:  # pragma: no cover
+        # TODO: get working coverage test case
+        return (
+            element_buffer[idx].segments
+            and element_buffer[idx].segments[0].is_type("keyword")
+            and element_buffer[idx - 1].segments
+            and element_buffer[idx - 1].segments[-1].is_type("indent")
+        )
```
alanmcruickshank, line 282: I'm going to spend a little time on trying to get a test case to cover this.

alanmcruickshank (approved):
I think this is mostly good to go 👍 . I'm going to try at find a way to get test coverage on the part marked with a TODO (or at least confirm how possible that would be), but otherwise I'll get this merged today.

Have spoken to @WittierDinosaur and he's cool for this to merge and then take it for a spin once released.

PR #6221, opened 2024-09-19
Exclude Macros - Allow multiple paths.

culpgrant:
There is an issue with the `exclude_macros_from_path` config that does not exclude macros if you are running `sqlfluff` from inside of the route directory where `.sqlfluff` file is.
Uses the absolute path for macros and the relative path of excluding_path to compare if the macro is in the excluding path
Fixes: https://github.com/sqlfluff/sqlfluff/issues/6235

### Pull Request checklist

  - Other.


alanmcruickshank:
@culpgrant - Looks very sensible. could you include a test case?

culpgrant:
> @culpgrant - Looks very sensible. could you include a test case?

Yeah - I seem to not be able to fully re-create the issue within this repo. Even when I cd into this `test/fixtures/templater/jinja_exclude_macro_path/model_directory` directory I cannot get the error to reproduce with the original code.

But it is an issue when I have sqlfluff installed in my project's virtual environment. 

@alanmcruickshank Would you happen to have any idea?

culpgrant:
> @culpgrant - Looks very sensible. could you include a test case?

I figured out the root cause that led me a simpler solution. What was really happening is if you had specified multiple paths to `exclude_macros_from_path` it was loading the paths with just the relative directory. While the paths from `load_macros_from_path` are always loaded with the absolute paths. This fix now loads the paths from `exclude_macros_from_path` the same way as `load_macros_from_path`.

This fix now loads the `exclude_macros_from_path` the same way as `load_macros_from_path`. I adjusted the existing test case to use two paths in `exclude_macros_from_path` and to run sqlfluff from inside a directory.

Let me know what you think of this @alanmcruickshank 

I think the failed checks are from an intermediate error with github runners.

alanmcruickshank (approved):
LGTM. Thanks for the edits. Thanks also for the tidy up that I can see you've done here 👍

PR #6229, opened 2024-09-20
Remove duplicate timing columns from the timing records

Tenzer:

The "templating", "lexing", "parsing" and "linting" columns appear twice in the timing records output, because their ordering both is configured explicitly, and because they still are included in the `rule_codes` set.

This fixes that by removing them from `rule_codes` so they only appear once in the resulting CSV file.

None

### Pull Request checklist

  - Other.

This part of the code doesn't seem to have any tests, and I guess it's intended for SQLFluff developers only anyways, so probably isn't important enough for that.

### DIFF
--- a/src/sqlfluff/core/linter/linting_result.py
+++ b/src/sqlfluff/core/linter/linting_result.py
@@ -155,6 +155,8 @@ class LintingResult:
                     continue
                 rule_codes.update(record["timings"].keys())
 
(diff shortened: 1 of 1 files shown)


alanmcruickshank (approved):
I'm not sure exactly _why_ they're getting pulled through in that way, but this seems like the most efficient way to remove them 👍

Issue #5213, opened 2023-09-15
Check and avoid circular `JOIN ON` conditions


d33bs:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

Great project, thanks so much for the effort put into things here! This issue highlights the possible enhancement of having SQLFluff check for circular join on condition which need not be in a SQL statement. I'm not certain of the exact labeling of the way this occurs, other words or phrases might include "self-referential, cyclical, or reflective". Please don't hesitate to let me know if there's an existing issue which covers this or if another tool or approach would be better.

### Use case

For example, a circular join on condition:
```sql
SELECT * FROM TABLE_A
JOIN TABLE_B ON
TABLE_B.COLUMN = TABLE_B.COLUMN
```

When checking this statement with SQLFluff, I'd expect a message stating something like:
`Circular join on condition discovered for JOIN TABLE_B for TABLE_B.COLUMN = TABLE_B.COLUMN`

I imagine there are occasions where there's a preference to use this behavior without SQLFluff raising an issue but I can't think of them off-hand. I'd welcome discussion on this front and would love to learn more. 

### Dialect

I believe this would apply to most or any SQL dialect.

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

barrywhart:
Could the scope of this potentially be expanded to **any** "boolean" condition which will always be true? E.g not only in a `JOIN`, but also in a `WHERE` clause, a `CASE` statement, etc.

d33bs:
Thanks @barrywhart - I think that's a great idea! When you added this did you have any examples in mind? I can think of some too, but mostly wondering about potential edge cases.

barrywhart:
Offhand, I was thinking of the two cases that came up in the PR:
* `1 = 1` or other "literal equals itself"
* `1 = 2` or other "literal equals a different literal"
* column equals itself

The warning could be something like "Replace constant-valued boolean expression with boolean constant TRUE or FALSE."

PR #6239, opened 2024-09-22
New Rules AM08 + CV12: Detect implicit cross joins

rogalski:
This is intended to avoid implicit cross-joins, which happened more than once in my commercial projects.
It can be considered as pre-work for #5213 as well.
If this is too opinionated, maybe we can consider this check to be opt-in instead of enabled by default.

Most likely we need more test cases to cover all unusual edge cases.

### Pull Request checklist

  - Other.


src/sqlfluff/rules/ambiguous/AM08.py
```diff
@@ -0,0 +1,95 @@
...
+
+    If dialect allows, alternatively use CROSS JOIN.
+
+    .. code-block:: sql
+       :force:
+
+        SELECT
+            foo
+        FROM bar
+        CROSS JOIN baz;
```
alanmcruickshank, line 53: I like this. If someone _wants_ a `cross join` they should explicitly ask for it.

src/sqlfluff/rules/ambiguous/AM08.py
```diff
@@ -0,0 +1,95 @@
...
+    LiteralSegment,
+    WhitespaceSegment,
+)
+from sqlfluff.core.rules import BaseRule, LintFix, LintResult, RuleContext
+from sqlfluff.core.rules.crawlers import SegmentSeekerCrawler
+
+
+class Rule_AM08(BaseRule):
+    """Join clauses should be present.
+
```
alanmcruickshank, line 17: I think there should be a more more preamble in the documentation to explain _why_ this rule exists. Something like:
> cross joins are valid, but rare in the wild - and more often created _by mistake_ than on purpose. This rule catches situations where a cross join has been specified, but not explicitly and so the risk of a mistaken cross join is highly likely.

alanmcruickshank: That will obviously need rewording, but I think you get the idea of what I mean.

src/sqlfluff/rules/ambiguous/AM08.py
```diff
@@ -0,0 +1,95 @@
...
+    Use const condition instead.
+
+    .. code-block:: sql
+       :force:
+
+        SELECT
+            foo
+        FROM bar
+        JOIN baz
+        ON 1=1;
```
alanmcruickshank, line 43: Maybe it's just me, but I don't know if I like this as best practice. As per #5213, having an equality statement that's always true (whether `1=1`, `a=a` or `TABLE_B.COLUMN = TABLE_B.COLUMN`) - feels more like an antipattern to me. Especially if there are many tables involved. e.g.

```sql
select *
from a
join b
join c
join d
where b.x = c.x
and d.x = a.x
and 1=1
```
Technically this query would pass this rule, but I don't think it's any clearer whether the cross join was intentional or not.

..._however_ I do also understand the desire to give an option to dialects which don't support an explicit `CROSS` join 🤔 . I have two other suggestions:
1. We disable this rule be default for any dialects which don't support an explicit `CROSS JOIN`. Similar to how `AM02` is only enabled for dialects which support `UNION DISTINCT`.
2. We say that best practice in dialects which don't have a `CROSS JOIN` is to explicitly disable the rule to signal that it's intentional, e.g.
   ```sql
   select *
   from a
   join b  -- noqa: AM08
   ```

Do either of those options (or any other options you can think of) feel like a better solution to this problem?

src/sqlfluff/rules/ambiguous/AM08.py
```diff
@@ -0,0 +1,95 @@
...
+                        edit_segments=[
+                            WhitespaceSegment(),
+                            KeywordSegment(on_kw),
+                            WhitespaceSegment(),
+                            LiteralSegment("1", type="numeric_literal"),
+                            ComparisonOperatorSegment("="),
+                            LiteralSegment("1", type="numeric_literal"),
+                        ],
+                    ),
+                ],
```
alanmcruickshank, line 92: I do wonder whether this rule should intentionally _not_ fix the code. This rule detects something ambiguous, which we think the user should be more specific about. If we auto-fix, we make that decision for them, rather than forcing them to decide whether it was intentional or not. Especially given my comment above about what best practice is.

Thoughts?

alanmcruickshank (changes requested):
I like this rule, but I'm less sure about what we're suggesting as best practice for dialects without an explicit `CROSS JOIN`. Ditto I'm not sure whether we should propose fixes in either case, given the thing we're detecting is ambiguous by definition.

Very happy to be persuaded otherwise however, given I'm sure you've thought about this more deeply than me and I might have missed something in the reasoning.

src/sqlfluff/rules/ambiguous/AM08.py
alanmcruickshank: ... interestingly, if we said that an "always true" condition like `1=1`, `a=a` or `TABLE_B.COLUMN = TABLE_B.COLUMN` doesn't count, then we wouldn't need separate rules for _implicit_ and _explicit_ cross joins, and you could resolve the original issue with one rule. 😄 

I'm still open to having two different rules for implicit/explicit, but could be a neat solution.

src/sqlfluff/rules/ambiguous/AM08.py
rogalski: Sounds good to me, I'll fix.

src/sqlfluff/rules/ambiguous/AM08.py
rogalski: @alanmcruickshank 

Please note that as of today, this fix only involves JOIN statements, not select from many targets, as in `SELECT * FROM a, b, c`.

In principle, I'd prefer to keep the default behavior to _refrain_ from emitting warnings instead of forcing users to inline disable.

I'd like to suggest additional test cases and then we can have another iteration on reviewing this.

src/sqlfluff/rules/ambiguous/AM08.py
rogalski: A very defensive solution would be to warn AND fix if and only if dialect supports explicit CROSS JOIN.

Which sounds sensible to me.

We do not have _exhaustive_ logic with the initial iteration. Rule logic can always be expanded after feedback from initial release.

src/sqlfluff/rules/ambiguous/AM08.py
```diff
@@ -0,0 +1,107 @@
...
+    Use CROSS JOIN.
+
+    .. code-block:: sql
+       :force:
+
+        SELECT
+            foo
+        FROM bar
+        CROSS JOIN baz;
+    """
```
alanmcruickshank, line 43:
```suggestion
    .. note::
       For SQL dialects which do not support an explicit :code:`CROSS JOIN`
       syntax we recommend using an inline ignore comments to mark a cross
       join as intentional:
       
       .. code-block:: sql

          SELECT
              foo
          FROM bar
          JOIN baz  -- noqa: ambiguous.join_condition    
    """
```

src/sqlfluff/rules/ambiguous/AM08.py
```diff
@@ -0,0 +1,107 @@
...
+
+class Rule_AM08(BaseRule):
+    """Join clauses should be present.
+
+    **Anti-pattern**
+
+    Cross joins are valid, but rare in the wild - and more often created by mistake than on purpose.
+    This rule catches situations where a cross join has been specified,
+    but not explicitly and so the risk of a mistaken cross join is highly likely.
+
```
alanmcruickshank, line 23: I think we should probably address people using the `WHERE` clause a join condition here. I think it's an anti-pattern, so it should still be a failure, but there's a question on what to do with it:

```sql
SELECT *
FROM bar
JOIN baz
WHERE bar.x = baz.x
AND bar.z > 5
```

This is functionally like having an `ON` clause, but the condition has been put in a `WHERE` clause instead. This would be what someone does in the case of "comma style" join too e.g.

```sql
SELECT *
FROM bar, baz
WHERE bar.x = baz.x
AND bar.z > 5
```

(I agree with you to ignore queries like this for now - and good to see you've added a test case for it 💖)

What I'm not sure of here is the fix, if we bluntly turn it into a `CROSS JOIN`, then it's still functionally the same, but that's not really the _intent_ of the rule. A better fix would be to extract the relevant conditions from the `WHERE` and put them in an `ON` clause instead, so that the first query would get fixed to:

```sql
SELECT *
FROM bar
JOIN baz
ON bar.x = baz.x
WHERE bar.z > 5
```

How hard do you think that would be to add up front?

alanmcruickshank (commented):
Nice. Good progress. Thanks for all the test cases too.

I've got one more situation that I think we should address - either in the fix, or at least in the docs, which is where someone has used a `WHERE` clause instead of an `ON` clause. In that situation, I think it's definitely an anti-pattern (so we should still fail), but I'm less sure that fixing directly to a `CROSS JOIN` is the right path.

What do you think?

src/sqlfluff/rules/ambiguous/AM08.py
rogalski: Simplistic where clauses (single condition or multiple conditions with AND statements only) should be relatively easy.

Complex conditions (AND/OR combinations, bracketed expressions, subqueries) will be super tricky to always get right.

Let me try to come up with some test cases.
Best case - autofixer will be able to solve _some_ of cases.
Worst case - we will reuse these test cases for 'violation without auto fix'.

src/sqlfluff/rules/ambiguous/AM08.py
rogalski: @alanmcruickshank 

IMO we recommend comma syntax instead.
```sql
SELECT foo FROM bar, baz
```

Are you aware of any syntax dialects which does not support CROSS JOIN? I failed to find one in codebase.

alanmcruickshank: I'm not aware of one either, so it's a fairly academic discussion. It's probably ok to not add this note then 👍

test/fixtures/rules/std_rule_cases/AM08.yml
```diff
@@ -24,6 +24,22 @@ test_fail_missing_multiple_joins_not_last:
...
+  # Implementation of fixes is not yet done.
+  # fix_str: "SELECT foo.a, bar.b FROM foo JOIN bar OR foo.a = bar.b;\n"
+
+test_fail_cross_join_with_complex_where_clause:
+  fail_str: "SELECT foo.a, bar.b FROM foo JOIN bar WHERE foo.a = bar.a AND foo.x = 3;\n"
+  # Implementation of fixes is not yet done.
+  # fix_str: "SELECT foo.a, bar.b FROM foo JOIN bar OR foo.a = bar.b WHERE foo.x = 3;\n"
+
+test_pass_cross_join_with_complex_where_clause_not_simplifiable:
+  pass_str: "SELECT foo.a, bar.b FROM foo JOIN bar WHERE foo.a = bar.a OR foo.x = 3;\n"
```
alanmcruickshank, line 38: In cases like this (where it's not a simple `AND` where clause, but the join is unqualified) - it's definitely simpler right now to just let it slide. I agree with not correcting it to a `CROSS JOIN`, but perhaps we make this rule a bit less useful by letting these slide... 🤔 

This specific case is probably more accurately a `LEFT JOIN` with some filtering - but detecting that could be quite convoluted. What's your thinking?

alanmcruickshank (commented):
Nice. Good progress.

Other than my more specific comments in a few places, I think "risky" parts of this rule are complex cases:
- Where `_is_where_clause_simplifable` is false.
- Where `_is_where_clause_simplifable` is true, but the conditions are complex enough functions that we don't feel comfortable just whacking them in an `ON` clause.

I'm totally onboard with not trying to exhaustively fix all of these cases, but I think we should be explicit in what the rule does in those kinds of cases. Whether that's just leaving it, or whether it's flagging an error and not a fix.

test/fixtures/rules/std_rule_cases/AM08.yml
rogalski: I don't believe that I'll be able to build exhaustive test case list for autofixer. Avoiding crashes / unintended behavior changes for autofixed code should be the ultimate priority IMO.

Check is still valuable as it is.

I hope this is not a blocker for PR being merged.

rogalski:
@alanmcruickshank For now, I've tuned the code to avoid any warning being emitted on non-select statements. This is to prevent unintended regression for cases where I have no idea of the best course of action.

This one is close to being mergable, but I look forward to any feedback.

alanmcruickshank (approved):
Yep - I think this is good to go! Well done. I won't merge straight away because I think it will trigger a minor release, and I might see what other maintainers think too.

Thanks for all your efforts to get this this far.

test/fixtures/rules/std_rule_cases/AM08.yml
```diff
@@ -0,0 +1,140 @@
...
+
+test_pass_cross_join_with_complex_where_clause_and_subqueries:
+  fail_str: "SELECT foo.a, bar.b FROM (SELECT * FROM abc) foo JOIN (SELECT * FROM def WHERE x = 3) bar WHERE foo.a = bar.a;\n"
+
+test_pass_inner_join_default:
+  pass_str: "SELECT foo.a, bar.b FROM foo INNER JOIN bar ON 1=1;\n"
+
+test_pass_left_join_default:
+  pass_str: "SELECT foo.a, bar.b FROM foo LEFT JOIN bar ON 1=1;\n"
```
OTooleMichael, line 67: Personally I think this is yet another different rule maybe.
`fixed_bool_condition_style`: `TRUE/FALSE, 1 = 1/ =0, never`

Personally I hate this style cause its redundant code, and i find it funny that users think they need to evaluate something to get a condition to be TRUE/FALSE. However sadly its also not possible to remove cause people often use it so that all other condition are `AND $cond`

OTooleMichael (changes requested):
I think this is a super great rule/work. My only questions/points would be:

1. I think the fix is dangerous, I assume the thing that drove you to do this was accidental cross joins (happened to me last week). If one doesn't review the code "fixing" here can turn an accidental cross join into one that look like its on purpose.
2. I think these are actually 3 rules, which compose together nicely
- Cross joins must be explicit
- Prefer join's with "ON" to where clause (fixable?)
- constant_boolean_style (enum): `1=1/0, TRUE/FALSE, never`

Those to me seem to add up to more composability for the technical user setting it up and more utility for the end user meeting the SQLFluff messages.

test/fixtures/rules/std_rule_cases/AM08.yml
```diff
@@ -0,0 +1,140 @@
...
+  fail_str: "SELECT foo.a, bar.b FROM foo left join bar on 1=2 left join baz;\n"
+  fix_str: "SELECT foo.a, bar.b FROM foo left join bar on 1=2 cross join baz;\n"
+
+test_fail_missing_multiple_joins_mixed_types:
+  fail_str: "SELECT foo.a, bar.b FROM foo left join bar on foo.x = bar.y left join baz;\n"
+  fix_str: "SELECT foo.a, bar.b FROM foo left join bar on foo.x = bar.y cross join baz;\n"
+
+test_fail_missing_multiple_joins_not_last:
+  fail_str: "SELECT foo.a, bar.b FROM foo left join bar left join baz on foo.x = bar.y;\n"
+  fix_str: "SELECT foo.a, bar.b FROM foo cross join bar left join baz on foo.x = bar.y;\n"
```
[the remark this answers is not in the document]
rogalski: I assumed that explicit cross join will scream error to code reviewer, but not fixing automatically also has clear merits.

I am comfortable with both approaches. I guess we just need majority consensus from maintainers.

test/fixtures/rules/std_rule_cases/AM08.yml
rogalski: I am happy to create backlog issue but I'd prefer to avoid feature creep in this PR.

test/fixtures/rules/std_rule_cases/AM08.yml
```diff
@@ -0,0 +1,140 @@
...
+      description: WHERE clause used for join condition. Use explicit ON instead.
+      end_file_pos: 57
+      end_line_no: 1
+      end_line_pos: 58
+      fixes: []
+      name: ambiguous.join_condition
+      start_file_pos: 38
+      start_line_no: 1
+      start_line_pos: 39
+      warning: false
```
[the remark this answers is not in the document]
rogalski: Where clause can contain combination of effective on clause and other where conditions. It can also be arbitrary nested expressions.

Decomposition is certainly possible for subset of cases, but we need to be careful not to introduce behavior change via autofix.

If we deem that autofix is critical, I'd be grateful for support in defining test case list.

src/sqlfluff/rules/ambiguous/AM08.py
barrywhart: FWIW, I found a Stack Overflow post which says that `CROSS JOIN` is part of ANSI-92 SQL. _Hopefully_, that means everybody has it by now. 😅

test/fixtures/rules/std_rule_cases/AM08.yml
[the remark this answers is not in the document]
barrywhart: It's an interesting point. We have some existing rules which "correct" the code by making the implicit behavior _explicit_, which hopefully makes the user realize they made a mistake (e.g. AM01, AM02, AM03).

In this case, I think I agree with Michael that the user may not realize what a `CROSS JOIN` means, because they are rarely used in practice. For that reason, I think I prefer a lint error with no automatic "fix" here.

src/sqlfluff/rules/ambiguous/AM08.py
```diff
@@ -0,0 +1,154 @@
...
+
+class Rule_AM08(BaseRule):
+    """Implicit cross join detected.
+
+    **Anti-pattern**
+
+    Cross joins are valid, but rare in the wild - and more often created by mistake
+    than on purpose. This rule catches situations where a cross join has been specified,
+    but not explicitly and so the risk of a mistaken cross join is highly likely.
```
barrywhart, line 17: What would we think of also saying something like:
>Cross joins can even cause query failures with large datasets, because they generate _all possible combinations_ of the two input tables, which can be a **huge** number of records.

I suggest this because cross join is a pretty technical term that some users won't be familiar with (for the very reason that they are rare), so this is an opportunity to provide some education and context for the existence of the rule.

test/fixtures/rules/std_rule_cases/AM08.yml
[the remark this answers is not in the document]
barrywhart: Related question: In what cases do we think an automatic fix is appropriate (or not)? My initial intuition is that maybe we _never_ want an automatic fix for this.

test/fixtures/rules/std_rule_cases/AM08.yml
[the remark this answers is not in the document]
barrywhart: I agree about making this a separate rule. When rules have a fairly narrow scope, they are easier to document, understand, and maintain. If it would be helpful, the two rules could potentially share/reuse code -- a number of the existing rules do that.

test/fixtures/rules/std_rule_cases/AM08.yml
barrywhart: Leaving it as is seems fine to me. We could always improve the rule later to handle trickier cases.

test/fixtures/rules/std_rule_cases/AM08.yml
rogalski: https://github.com/sqlfluff/sqlfluff/issues/6350

rogalski:
Just FYI, I am still planning to wrap up this work, but it will get slightly delayed

rogalski:
@alanmcruickshank @OTooleMichael @barrywhart 

This one got a little bit out of hand, but effectively we have two rules:
- AM08 detects implicit cross joins (no autofix at all)
- CV12 rewrites simple WHERE expressions into JOIN ON clauses (autofix)

I am somewhat unhappy with the implementation (really odd handling of whitespace, somewhat long rule code).

Any comments are **much appreciated**.

rogalski:
@alanmcruickshank @OTooleMichael @barrywhart
Kind ping on this one.

rogalski:
@alanmcruickshank @OTooleMichael @barrywhart

Quick summary on what was done since the last review.
Hopefully, this will be useful.

rogalski:
@alanmcruickshank @OTooleMichael @barrywhart

Quick summary on what was done since the last review.
Hopefully, this will be useful.

AM08:
- detects condition-less + non-explicit cross joins
- does not generate autofix, as we agreed that autofix is too confusing

CV12:
- Detects WHERE clauses which should be JOIN ON clauses
- Applies autofix if possible (if WHERE condition is simple AND expression)
- If expression is not simple, reports violation without autofix

Two checks in tandem fulfill all logical issues discussed in the review.

AM06: https://github.com/sqlfluff/sqlfluff/pull/6239#pullrequestreview-2363834896
CV12: https://github.com/sqlfluff/sqlfluff/pull/6239#discussion_r1798432199
Complementary PR #6392: https://github.com/sqlfluff/sqlfluff/pull/6239#discussion_r1797637959

One possible improvement area is documentation.

alanmcruickshank (approved):
I've made a slight tweak to the rule description/message to make CV12 a little more explicit - but otherwise I think this is good to go 👍 

I removed the note about maintainers checking the condition. I think it's reasonable.

Issue #6220, opened 2024-09-19
Using mypyc to speed up SQLFluff


Tenzer:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

@alanmcruickshank mentions in https://designingoverload.com/sqlfluff-performance/ that using `mypyc` as a way of speeding up SQLFluff wasn't possible in the end, but don't mention any further details on it. I've also found some PRs in #5012, #5013 and #5107 from Alan which mention `mypyc`.

I'm creating this ticket to try and get a better idea about what might be stopping us from making use of `mypyc` and, relatedly, if there's some benchmarking in place for SQLFluff to make it easier to test performance changes locally, and figure out where the most time is spent.

I'm hoping it could be a relatively low-barrier way of unlocking faster performance.

### Dialect

N/A

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

Tenzer:
I should probably mention that I managed to get the `sqlfluff.utils.reflow` module compiled with `mypyc` with only some minor changes required:
```diff
diff --git src/sqlfluff/utils/reflow/elements.py src/sqlfluff/utils/reflow/elements.py
index 1a386e189..4501cb81e 100644
--- src/sqlfluff/utils/reflow/elements.py
+++ src/sqlfluff/utils/reflow/elements.py
@@ -279,7 +279,7 @@ class ReflowPoint(ReflowElement):
         NOTE: This only returns _untemplated_ indents. If templated
         newline or whitespace segments are found they are skipped.
         """
-        indent = None
+        indent: Optional[RawSegment] = None
         for seg in reversed(self.segments):
             if seg.pos_marker and not seg.pos_marker.is_literal():
                 # Skip any templated elements.
diff --git src/sqlfluff/utils/reflow/reindent.py src/sqlfluff/utils/reflow/reindent.py
index f834062c4..a06d6609d 100644
--- src/sqlfluff/utils/reflow/reindent.py
+++ src/sqlfluff/utils/reflow/reindent.py
@@ -695,7 +695,7 @@ def _crawl_indent_points(
     TODO: Once this function *works*, there's definitely headroom
     for simplification and optimisation. We should do that.
     """
[...]

PR #6240, opened 2024-09-22
Strict typing for `sqlfluff.core.linter`

alanmcruickshank:
This brings the `sqlfluff.core.linter` module up to strict mypy typing (and enforces that in CI).

A lot of this is just type annotations, however there are a few more interesting bits:
- The `linter` module now defines a `FormatterInterface` for the `cli.OutputStreamFormatter` to inherit from. This means that all of the `Linter` methods which allow a `formatter` argument have an interface to use for typing without importing the `cli` module.
- The `check_tuples` method on `LintedFile`, `LintedDir` and `LintingResult` no longer supports a `by_path` argument. Instead there's now a seperate method called `check_tuples_by_path` to replace that functionality. That makes the typing much cleaner (and removes a bunch of overloads).
- For the `runner` module, there's more use of `abstractmethod` for the stub methods, and rather than using `MAP_FUNCTION_NAME`, each of the subclasses now just redefine the `_map()` method directly to make the call. It's a little more verbose, but much clearer and more explicit for typing.

Aside from the `templater` module, this is most of the `core` module done. Supports #6220

### DIFF
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -759,7 +759,9 @@ def lint(
     if not nofail:
         if not non_human_output:
             formatter.completion_message()
-        sys.exit(result.stats(EXIT_FAIL, EXIT_SUCCESS)["exit code"])
+        exit_code = result.stats(EXIT_FAIL, EXIT_SUCCESS)["exit code"]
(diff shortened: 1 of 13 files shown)


WittierDinosaur (approved):
Nice!

PR #6246, opened 2024-09-23
Strict mypy on `sqlfluff.core`

alanmcruickshank:
This follows on from #6240 and should support #6220.

This completes the job of making the whole of `sqlfluff.core` compliant with `mypy --strict`.

The majority of this is fairly uncontroversial and just fiddly type annotations. The only two more major changes are
* ~A slight hack for the `hookspec`, which involved moving the `types` module up a level, and introducing a `RuleType` class as a base class for rules, but which only acts as a return type for the base `get_rules()` method.~ UPDATE: Ignore this, I think it's gross so just added an exception in the import linter to allow it. There's already two other exceptions in there for type checking.
* A refactoring of `get_context` for the templaters, including a change for the jinja templater which use to _override_ this method, and now implements it's own custom method which _calls_ this method to allow it to change the type signature.

I've also added a bit more documentation around the `py.typed` file.


src/sqlfluff/core/plugin/hookspecs.py
```diff
@@ -14,15 +16,24 @@ class PluginSpec:
...
-    def get_rules(self):
-        """Get plugin rules."""
+    def get_rules(self) -> List[Type[RuleType]]:
+        """Get plugin rules.
+
+        NOTE: While the type annotation for this method returns `Type[RuleType]`
+        all plugin implementations should instead import `sqlfluff.core.rules` and
+        return the more specific `Type[BaseRule]`. This base definition of the
+        PluginSpec cannot import that yet as it would cause a circular import.
+        """
```
alanmcruickshank, line 26: @WittierDinosaur - this is the only bit that feels a bit gross, but this is the least bad of a few choices. Those options are:
1. (what I've done, which is) Have the plugin module describe something "rule-like" which `get_rules()` returns, and that be a parent class of `BaseRule` so that mypy is happy.
2. We could have the type annotation here reference `BaseRule` directly, but that violates the `lint-imports` rule which says that the `linter` can't depend on `rules` (it must be the other way around only) - and at the moment, `linter` depends on `config` which depends on `plugin`. `linter` is likely always going to depend on `config`, and refactoring `config` so that it doesn't need to depend on `plugin` seems hard or even impossible.
3. We could avoid the new `RuleType` class and just make this type hint really vague (i.e. `List[Any]` or `List[Type[Any]]`). You could argue that the current `RuleType` implementation isn't much better than this because the class doesn't actually define anything other than being a marker for inheritance, but it feels a little more descriptive than this.

What do you reckon?

src/sqlfluff/core/plugin/hookspecs.py
```diff
@@ -14,15 +16,24 @@ class PluginSpec:
...
+    def load_default_config(self) -> Dict[str, Any]:
         """Loads the default configuration for the plugin."""
 
     @hookspec
     @abstractmethod
-    def get_configs_info(self) -> dict:
+    # TODO: This type annotation could probably be more specific but that would
+    # require making the config info object something more like a namedTuple rather
+    # than a dict.
+    def get_configs_info(self) -> Dict[str, Dict[str, Any]]:
```
alanmcruickshank, line 38: I'm not proposing that I resolve this TODO in this PR. It's more of a marker for later improvements. What I've done here is already better than what was here before.

src/sqlfluff/core/plugin/hookspecs.py
alanmcruickshank: Ignore all of this. I think it's gross. I've just added an exception in the import linter instead.

PR #6262, opened 2024-09-26
Move most of the config validation settings out into rule bundles.

alanmcruickshank:
I noticed that while most of the rules are defined in plugin bundles now (and therefore outside of the core package), and that there's a plugin hook for importing config validation values - none of the bundles use it, and instead all the config validation settings are still centrally defined.

This moves most of them out into the rule bundles, for better isolation, and not better mimic how external plugin maintainers should structure their projects.

I also removed the validation for any config values which are no longer valid to have on rules.


test/core/rules/rules_test.py
```diff
@@ -270,10 +270,10 @@ def test_rules_cannot_be_instantiated_without_declared_configs():
...
         """Testing Rule."""
 
-        config_keywords = ["tab_space_size"]
+        config_keywords = ["case_sensitive"]
 
-    new_rule = Rule_NewRule_ZZ99(code="L000", description="", tab_space_size=6)
-    assert new_rule.tab_space_size == 6
+    new_rule = Rule_NewRule_ZZ99(code="L000", description="", case_sensitive=False)
+    assert new_rule.case_sensitive is False
     # Error is thrown since "tab_space_size" is defined in class,
```
keraion, line 277: Comment here referencing `tab_space_size` should be updated.

PR #6263, opened 2024-09-27
Almost all of `util` up to strict typing

alanmcruickshank:
There are a few corners of `util` which I couldn't find a tidier solution for - but this gets _most_ of `sqlfluff.utils` up to strict mypy standard. I might change some of this later, but thought I should make the existing typing clearer first.


src/sqlfluff/rules/layout/LT05.py
```diff
@@ -46,6 +46,7 @@ def _eval(self, context: RuleContext) -> List[LintResult]:
                 # of the line is a comment).
                 assert res.anchor
                 assert res.anchor.pos_marker
+                assert isinstance(res.anchor, RawSegment)
```
WittierDinosaur, line 49: could this not be properly typed in LintResult?

WittierDinosaur (approved):
LGTM

PR #6269, opened 2024-09-27
Python 3.13 support and make it default for test coverage

alanmcruickshank:
This is a little experimental, but I think we can now add pre-release testing of python 3.13. It's released properly next week, but the python setup github action does support pre-releases.

If this passes, I'm tempted to merge it in before next week.


alanmcruickshank:
Full release expected 7th October at the moment: https://peps.python.org/pep-0719/

keraion:
I think the failure here is from the change in how docstrings now strip whitespaces. See: https://docs.python.org/3.13/whatsnew/3.13.html#other-language-changes

keraion (approved):
LGTM. 🚀

PR #6284, opened 2024-10-02
Remove dependency on coveralls.

alanmcruickshank:
@greg-finley has identified that the coveralls jobs are failing, and that the repo doesn't seem to be updated anymore. All of our coverage jobs have already been migrated to github actions anyway, so this finishes the job and removes coveralls entirely.


greg-finley (approved):
❤️

PR #6292, opened 2024-10-04
Bugfix for LT12 with jinja whitespace consumption

alanmcruickshank:
I found an issue on a project I'm working on where LT12 misbehaves with trailing jinja tags. This resolves that issue, and adds test cases to cover it.


keraion (approved):
LGTM

PR #6293, opened 2024-10-04
Bugfix for JJ01 in parallel mode

alanmcruickshank:
I noticed a _really wierd bug_ with JJ01. It basically doesn't work in parallel mode at the moment.

On debugging, the cause seems to be that when it checks for the presence of a jinja templater (or subclass), and it's using `context.config.get("templater_obj")` which intentionally isn't rehydrated when the `FluffConfig` object is pickled/unpickled. Instantiating a new templater object seems overkill, but we do still need a way of checking the class.

This PR allows a new method `get_templater_class()` which rules can use to find out if they're in a jinja-like environment. That one does work in parallel mode.

In doing that, I've changed the type signature of the original `get_templater()` method to use the config value rather than the argument. I think it's cleaner and safer.

### DIFF
--- a/src/sqlfluff/core/config/fluffconfig.py
+++ b/src/sqlfluff/core/config/fluffconfig.py
@@ -97,9 +97,7 @@ class FluffConfig:
         dialect: Optional[str] = self._configs["core"]["dialect"]
         self._initialise_dialect(dialect, require_dialect)
 
-        self._configs["core"]["templater_obj"] = self.get_templater(
-            self._configs["core"]["templater"]
-        )
+        self._configs["core"]["templater_obj"] = self.get_templater()
 
     def _handle_comma_separated_values(self) -> None:
         for in_key, out_key in [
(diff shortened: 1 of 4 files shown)


src/sqlfluff/core/config/fluffconfig.py
```diff
@@ -298,20 +299,27 @@ def from_kwargs(
...
                 self._plugin_manager.hook.get_templaters()
             )
         }
+        # Fetch the config value.
+        templater_name = self._configs["core"].get("templater", "<no value set>")
         try:
             cls = templater_lookup[templater_name]
             # Instantiate here, optionally with kwargs
-            return cls(**kwargs)
+            return cls
```
keraion, line 322: We aren't actually instantiating the class here any longer. This comment should be updated to reflect that.

alanmcruickshank: Ooh good point 👍

keraion (approved):
LGTM

PR #6294, opened 2024-10-04
Bugfix for LT07 with consumed newlines.

alanmcruickshank:
I noticed that LT07 is throwing false positives if a CTE ends with a whitespace-consuming jinja tag. this adds logic to handle that situation and react accordingly.


keraion (approved):
LGTM

PR #6296, opened 2024-10-04
Dededuplicate rule ignore docs

alanmcruickshank:
Currently there are two, almost identical sections on how to ignore specific rules:
- https://docs.sqlfluff.com/en/stable/reference/rules.html#inline-ignoring-errors
- https://docs.sqlfluff.com/en/stable/configuration/ignoring_configuration.html#ignoring-errors-files

This is unhelpful and confusing, especially given the information in each is slightly different.

This PR sets the one in `configuration` as the main one, and then just links to that from the rules references section.

### DIFF
--- a/docs/source/configuration/ignoring_configuration.rst
+++ b/docs/source/configuration/ignoring_configuration.rst
@@ -3,6 +3,8 @@
 Ignoring Errors & Files
 -----------------------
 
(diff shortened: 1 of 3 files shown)


keraion (approved):
LGTM. This helps clean up and removes some of the contributor level information that was previously present.

PR #6298, opened 2024-10-05
ST06 - Fix union of CTE/Subquery

### LINKED ISSUE #5956: structure.column_order breaks subsequent UNION ALL.
### Search before asking

### What Happened

SQLFluff fix produces SQL with changed behavior.

### Expected Behaviour

SQL before and after fix is functionally equivalent.

### Observed Behaviour

T2 SELECT has reordered columns, which breaks UNION.

### How to reproduce

```sql
-- noqa: disable=structure.column_order
CREATE TABLE X AS
WITH T1 AS (
    SELECT
        'a'::varchar AS A,
        1::bigint AS B

),

T2 AS (
    SELECT
        CASE WHEN COL > 1 THEN 'x' ELSE 'y' END AS A,
        COL AS B
    FROM T
)

SELECT * FROM T1
UNION ALL
SELECT * FROM T2;
```

Run sqlfluff fix with disable statement, all good.
Run sqlfluff fix without disable statement, output columns in T2 are reordered, which breaks subsequent union.

### Dialect

snowflake

### Version

3.0.7

### Configuration

default

### Code of Conduct

rogalski:
Add failing test cases, no fix yet.

Fixes #5956

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/structure/ST06.py
+++ b/src/sqlfluff/rules/structure/ST06.py
@@ -87,15 +87,33 @@ class Rule_ST06(BaseRule):
 
         assert context.segment.is_type("select_clause")
 
+        # insert, merge, create table, union are order-sensitive
         for seg in reversed(context.parent_stack):
             if seg.is_type(
                 "insert_statement",
                 "set_expression",
-                "with_compound_statement",
                 "create_table_statement",
                 "merge_statement",
             ):
                 return None
+
+        # CTE is order-sensitive only if CTE is referenced in set expression
+        for seg in reversed(context.parent_stack):
+            if seg.is_type("common_table_expression"):
+                cte_identifier = seg.get_child("identifier")
+                assert cte_identifier is not None
+                maybe_with_compound_statement = seg.get_parent()
+                if maybe_with_compound_statement is None:
+                    break
+                with_compound_statement, _ = maybe_with_compound_statement
+                for ref in with_compound_statement.recursive_crawl("table_reference"):
+                    if ref.raw_upper == cte_identifier.raw_upper:
+                        path = with_compound_statement.path_to(ref)
+                        if any(
+                            path_step.segment.is_type("set_expression")
+                            for path_step in path
+                        ):
+                            return None
 
         select_clause_segment = context.segment
         select_target_elements = context.segment.get_children("select_clause_element")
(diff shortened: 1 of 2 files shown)


rogalski:
Not sure if CI will be 100% green, but this looks fairly good in my local environment.

src/sqlfluff/rules/structure/ST06.py
```diff
@@ -86,35 +86,34 @@ def _eval(self, context: RuleContext) -> EvalResultType:
...
+
+        # CTE is order-sensitive only if CTE is referenced in set expression
+        for seg in reversed(context.parent_stack):
+            if seg.is_type("common_table_expression"):
+                cte_identifier = seg.get_child("identifier")
+                assert cte_identifier is not None
+                maybe_with_compound_statement = seg.get_parent()
+                if maybe_with_compound_statement is None:
+                    break
+                with_compound_statement, _ = maybe_with_compound_statement
```
alanmcruickshank, line 108: The coverage checks are failing because of this line, but it's _very_ unlikely that would ever get used. If you add a comment here saying that clause is mostly a safety measure, then I think it's also ok to mark it as `# pragma: no cover`.

It would only be `None` if a fix has modified it in some way and we haven't yet re-populated the parent. Even in that scenario it's very unlikely.

src/sqlfluff/rules/structure/ST06.py
```diff
@@ -86,35 +86,34 @@ def _eval(self, context: RuleContext) -> EvalResultType:
...
+        for seg in reversed(context.parent_stack):
+            if seg.is_type(
+                "insert_statement",
+                "set_expression",
+                "create_table_statement",
+                "merge_statement",
+            ):
+                return None
+
+        # CTE is order-sensitive only if CTE is referenced in set expression
```
alanmcruickshank, line 100: Perhaps a small thing, which you can choose whether to address: The order is only sensitive _if a wildcard (*) is used_. In the case of a set expression where all the columns are re-specified, it would be ok to reorder them here - it's only when the set expression uses a `*` that it becomes a problem.

For bonus points, taking that into account would be awesome.

alanmcruickshank (dismissed):
Nice. I think there's an optional way of making this even better but I love the approach.

alanmcruickshank (approved):
Looks great! Thanks for the edits 🚀

PR #6312, opened 2024-10-08
Add groups to the `github-annotation-native` format option.

### LINKED ISSUE #6258: Include filename in visible GHA output
### Search before asking

### Description

When using `github-annotation-native` format, the raw output contains the filename
```
::warning title=SQLFluff,file=migrations/manual/20240912220714_foo.sql,line=2,col=1,endLine=2,endColumn=3::LT02: Line should not be indented. [layout.indent]
```
and the actual annotations work properly, however the output visible in the GH console doesn't show the filename, which makes it a little tricky from the console output. It might be a little redundant, but maybe the messagee should be prefixed with the filename (or else put the filename on top of each section the way we do in other output formats, if that won't confuse GH)

image

### Use case

_No response_

### Dialect

postgres

### Code of Conduct

alanmcruickshank:
I think this resolves #6258.

@wyardley, does this solution work for you? I've taken a lot of inspiration from how yamllint have approached this problem.

### DIFF
--- a/src/sqlfluff/cli/commands.py
+++ b/src/sqlfluff/cli/commands.py
@@ -715,6 +715,11 @@ def lint(
         github_result_native = []
         for record in result.as_records():
             filepath = record["filepath"]
+
+            # Add a group, titled with the filename
+            if record["violations"]:
+                github_result_native.append(f"::group::{filepath}")
+
             for violation in record["violations"]:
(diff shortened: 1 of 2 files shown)


wyardley:
Looks great if I’m reading that right. Thanks!

greg-finley (approved):

PR #6316, opened 2024-10-08
Allow double quotes to be escaped by writing twice

alanmcruickshank:
I found this issue on snowflake, but I checked the postgres docs, and I think it's fairly widespread, as such I'm adding it to the `ansi` dialect.

It turns out that this is valid SQL (don't ask how I found out 🫨 ):

```sql
select
    1 as foo,
    2 as "foo",
    3 as """foo""",
    4 as """""foo""""",
    bar,
    "bar",
    """bar""",
    """""bar"""""
from """""a"""""."""""b"""""."""""c"""""
```

This is because double quotes can be escaped by doubling them. I've confirmed this on snowflake live, and it's in the postgres docs:

> To include the escape character in the identifier literally, write it twice.

### DIFF
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -122,7 +122,7 @@ ansi_dialect.set_lexer_matchers(
         ),
         RegexLexer(
             "double_quote",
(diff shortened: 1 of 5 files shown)


greg-finley (approved):
most amusing test I've seen in a while 😅

PR #6320, opened 2024-10-08
Exclude templated casts from CV11

alanmcruickshank:
I found this bug earlier today. Consider the following query (although it would be better to imagine that the macro is defined somewhere else in the codebase entirely):

```sql
{% macro cast_macro(col) %}
    CAST(col as int)
{% endmacro %}

select
    a,
    {{ cast_macro('b') }},
    c::date as d
from tbl
```

The casting style is definitely inconsistent (there's an explicit `CAST()` and there's also an implicit `::date` operation). However for the user writing the final query they may now know (or have control over) the macro they're calling (over `cast_macro`). As such I don't think it should be an error if `::date` is inconsistent with the casting inside the `cast_macro`. This PR excludes any templated sections from being considered for `CV11`.

### DIFF
--- a/src/sqlfluff/rules/convention/CV11.py
+++ b/src/sqlfluff/rules/convention/CV11.py
@@ -205,6 +205,14 @@ class Rule_CV11(BaseRule):
         if context.dialect.name == "teradata":
             return None
 
+        # If we're in a templated section, don't consider the current location.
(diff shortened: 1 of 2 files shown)


keraion (approved):
LGTM!

PR #6324, opened 2024-10-09
Resolve some more edge cases in LT02

alanmcruickshank:
This follows on from #6306 where I mentioned there were some trickier bugs I couldn't resolve then. This resolves those bugs.

The logic in the indentation algorithms is pretty convoluted (which I'd kind of forgotten about). I've done a little bit of simplification here, but I'm sure there's more that could be done here.


keraion (approved):
LGTM.

PR #6332, opened 2024-10-11
Handle deprecation warning of "fork"

alanmcruickshank:
As discussed on slack. Python 3.13 deprecates the "fork" threading option. This explicitly sets the option to "spawn" for all platforms.

https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods


keraion (approved):
LGTM

PR #6026, opened 2024-07-26
AL09 (Self-alias): Quote alias when case changed

aayushr7:
### Brief summary of the change made

This change changes the fix scenario when the self-alias is done for case change. For case change, users often do not quote the alias, example: `id AS ID`. The existing logic will fix this to `id`. However, users expect this as case change.

If AL09 passes the rule, this can conflict with capitalisation policy rules. For example: if `capitalisation.identifiers` is set to `lower`, then the final fix would be `id AS id` which is not expected.

Therefore, the alias should not be unquoted when there is case change, rather quoted. 
Correct fix: `id AS "ID"`. For BigQuery: ```id as `ID`.```
However, this fix has conflicts with RF06 which considers quotes unnecessary. Therefore, for such cases fix is not applied yet.

- fixes #5636
- fixes #6019

### Are there any other side effects of this change that we should be aware of?
None

### Pull Request checklist
- [X] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

Thanks @keraion for the initial PR.

PR #6022, opened 2024-07-23
AL09: Handle case sensitivity with aliases

keraion:
### Brief summary of the change made

This handles a set of situations:
- Retains the usage of an alias if the case of the alias is different from the column name referenced. (e.g. `"COL" as "col"`)
- If the column name is case-insensitive for access, and the alias is a different case than the column name, replace the column name with the case used by the alias. (e.g. `col1 as COL` would be replace with `COL1`) This is dialect specific.
- fixes #5636
- fixes #6019

### Are there any other side effects of this change that we should be aware of?
None, but additional dialects that support the result set with case-insensitive column names will need to be added.

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #6333, opened 2024-10-11
[DRAFT] Revise AL09 (self aliasing)

### LINKED ISSUE #5636: AL09 - Case sensitivity with unquoted identifiers: TSQL
### Search before asking

### What Happened

When using sql fluff fix it rewrites our SQL queries and removes "self aliasing" even though the casing is different and even though this is nescessary.

### Expected Behaviour

I do not want "ID" to be interpreted as equal to "id" or "Id".

### Observed Behaviour

The following is changed by sqlfluff fix:
```
SELECT
	id AS  ID,
	date AS Date,
	description AS Description,
FROM
[...]
```
into:
```
SELECT
	id,
	date,
	description,
FROM
[...]
```

### How to reproduce

Use the above SQL query and run sqlfluff lint or fix on it

### Dialect

tsql

### Version

  - repo: https://github.com/sqlfluff/sqlfluff
    rev: 3.0.0a5
    hooks:
      - id: sqlfluff-fix
      - id: sqlfluff-lint

### Configuration

[tool.sqlfluff.core]
large_file_skip_byte_limit = 100000
dialect = "tsql"
ignore = "templating"

[tool.sqlfluff.indentation]
tab_space_size = "4"
indent_unit = "tab"

### Code of Conduct

### LINKED ISSUE #5993: AL09 Exception. TSQL aliasing expression
### Search before asking

### What Happened

I'm linting a T-SQL script (for SQL Server) and received the following:

```
L:  13 | P:   1 | AL09 | Unexpected exception: ;
Could you open an issue at
                       | https://github.com/sqlfluff/sqlfluff/issues ?
```

The line in question is as follows:

```sql
SELECT @CURRENT_SCHEMA_VERSION = [Version] FROM [HangFire].[Schema];
```

### Expected Behaviour

This line should not have any linting issues given the current configuration and dialect.

### Observed Behaviour

An empty exception is raised.

### How to reproduce

`sqlfluff lint Script.sql`

### Dialect

`tsql`

### Version

3.0.7

### Configuration

```ini
[sqlfluff]
dialect = tsql
exclude_rules = LT01,LT05,LT12

[sqlfluff:rules:references.quoting]
prefer_quoted_identifiers = True

[sqlfluff:rules:references.special_chars]
additional_allowed_characters = .
```

### Code of Conduct

### LINKED ISSUE #6019: AL09 - Case sensitivity with unquoted identifiers: Clickhouse
### Search before asking

### What Happened

The fix command unexpectedly deletes aliases when ignoring capitalization rules for identifiers.

Submitted query:
SELECT
    myidcol AS col_id,
    my_expected_drop_alias AS my_expected_drop_alias,
    My_Wrong_Case_Col as my_wrong_case_col,-- noqa: CP02
    my_good_col,
FROM
    file('myfile.parquet', 'Parquet')

Reformatted query:
SELECT
    myidcol AS col_id,
    my_expected_drop_alias,
    My_Wrong_Case_Col,-- noqa: CP02
    my_good_col,
FROM
    file('myfile.parquet', 'Parquet')

### Expected Behaviour

Expected result:
SELECT
    myidcol AS col_id,
    my_expected_drop_alias,
    My_Wrong_Case_Col as my_wrong_case_col,-- noqa: CP02
    my_good_col,
FROM
    file('myfile.parquet', 'Parquet')

### Observed Behaviour

When AL09 is used, but CP02 is ignored, AL09 is incorrectly triggered.

### How to reproduce

Example queries provided

### Dialect

Clickhouse

### Version

sqlfluff, version 3.1.0

### Configuration

Default config + inline noqa flag.

### Code of Conduct

alanmcruickshank:
This is a draft for now while I work through some of the corners, also for people to review what direction I'm going in.

This should supersede #6022 and #6026.

It (will) resolve #5636, resolve #6019 & resolve #5993.

While there's going to be logic changes in here, there's also a significant step up on documentation because I think that's going to help a lot here.


alanmcruickshank:
@keraion & @aayushr7 - I finally managed to consolidate all the input people gave on AL09 and how it interacts with other rules, and this is where I think I landed, which is even more conservative than both of your suggestions. Would love both of your reviews of this approach when you have a moment.

src/sqlfluff/rules/aliasing/AL09.py
```diff
@@ -13,29 +13,74 @@ class Rule_AL09(BaseRule):
...
+       and casing of the alias and reference are the same.
+
+       If those two rules are enabled, the fixes applied may result in a
+       situation where this rule can kick in as a secondary effect. For
+       example this :ref:`snowflake_dialect_ref` query:
+
+       .. code-block:: sql
+
+          -- Original Query. AL09 will not trigger because casing and
+          -- quoting a different. RF06 will however fix the unnecessary
```
keraion, line 38:
```suggestion
          -- quoting are different. RF06 will however fix the unnecessary
```

keraion (commented):
Just a small typo. Everything else looks great!

src/sqlfluff/rules/aliasing/AL09.py
```diff
@@ -13,29 +13,74 @@ class Rule_AL09(BaseRule):
...
+    :code:`foo as foo` or :code:`"BAR" as "BAR"`, see note below on
+    more complex examples). Aliases which effectively change the casing of
+    an identifier are still allowed.
+
+    .. note::
+
+       This rule works in conjunction with :sqlfluff:ref:`references.quoting`
+       (:sqlfluff:ref:`RF06`) and :sqlfluff:ref:`capitalisation.identifiers`
+       (:sqlfluff:ref:`CP02`) to handle self aliases with mixed quoting
+       and casing. In the situation that these two roles are not enabled
```
aayushr7, line 27:
```suggestion
       and casing. In the situation that these two rules are not enabled
```

test/rules/std_AL09_CP02_RF06_combo_test.py
```diff
@@ -0,0 +1,194 @@
...
+            "ansi",
+            """
+select
+    a as A,
+    B as b,
+    "C" as C,
+    "d" as d,
+    "E" as e,
+    f as F,
+    g as "G",
```
aayushr7, line 43: If `CP02` is not enabled, what is the reason that `a as A` and `B as b` are not considered self-alias?

test/rules/std_AL09_CP02_RF06_combo_test.py
```diff
@@ -0,0 +1,194 @@
...
+import sqlfluff
+
+input_query = """
+select
+    a as A,
+    B as b,
+    "C" as C,
+    "d" as d,
+    "E" as e,
+    f as F,
```
aayushr7, line 19: I believe this is same as `a as A` and can be removed.

test/rules/std_AL09_CP02_RF06_combo_test.py
alanmcruickshank: Good question.

Here's my thinking:
- Someone has disabled `CP02`. They've decided that they don't want SQLFluff to complain if they use inconsistent casing for unquoted identifiers. That implies the _want_ to be specific in the casing of unquoted identifiers. Remember that this question is only relevant if someone has made an active choice to not enable `CP02`.
- If someone has decided they want to be specific in their identifier quoting, then if we consider `a as A` and `B as b` to be unnecessary, we're effectively disagreeing with that choice.

In some dialects this would be inappropriate all the time (clickhouse, which is always case sensitive), or in others it's _configurably inappropriate_ (TSQL, can be configured to always be case sensitive).

Do you think we should be more aggressive for at least some dialects even if CP02 is off?

test/rules/std_AL09_CP02_RF06_combo_test.py
alanmcruickshank: agreed

test/rules/std_AL09_CP02_RF06_combo_test.py
alanmcruickshank: I think we've got the data on enough dialects now that we _could_ do it if we wanted (exclude the ones we know are problematic, and otherwise correct to the alias). What I'm thinking more is whether we _should_.

aayushr7: Thanks for the explanation! I agree that it could be an intentional behaviour to disable `CP02` to have the specific casing. 

However, my thinking is that we would not know if this is intentional behaviour by the user to do `a as A` and `B as b`, or an error. Since `AL09` is enabled, I believe we should still mark these errors as _unfixable_ lint errors. Then the user has the option to quote the alias, or disable the rule `AL09` for these lines or entirely. This way a user can specify their intended behaviour.

What do you think?

alanmcruickshank: I like that suggestion 👍 . I think I was getting hung up on the right way to fix it, but you're right that it should still be an error, just one that the user will need to resolve.

src/sqlfluff/rules/aliasing/AL09.py
```diff
@@ -13,29 +13,74 @@ class Rule_AL09(BaseRule):
...
     .. code-block:: sql
 
         SELECT
-            col
+            col,
+            COL,
+            -- Re-casing aliasing is still allowed where necessary
+            col as "Col",
+            "col" as "COL"
         FROM table;
```
aayushr7, line 84:
```suggestion
            col,
            "Col"
            COL,
        FROM table;
        
        -- Re-casing aliasing is still allowed where necessary, i.e.
        SELECT
            col as "Col",
            "col" as "COL"
        FROM table;
```

aayushr7:
Looks good to me! Suggested one improvement in the doc so that the anti-pattern and best practice refer to equivalent SQL.

src/sqlfluff/rules/aliasing/AL09.py
alanmcruickshank: Good idea 👍

aayushr7 (approved):

keraion (approved):
I think this is a good middle ground on the linting without an auto-fix. LGTM.

PR #6344, opened 2024-10-12
Utilize a deepcopy of the config object when parsing files

### LINKED ISSUE #6343: Configs are Mutated by File-Level Overrides
### Search before asking

### What Happened

When using a `FluffConfig` object as a config when linting a SQL string, any file-level overrides will mutate the config object, causing subsequent lintings to use the same file-level override. 

### Expected Behaviour

```python
def test__api__immutable_config():
    config = FluffConfig.from_path("test/fixtures/api/config_path_test/extra_configs/.sqlfluff")
    assert config.get("dialect") == "ansi"
    sqlfluff.lint(
        "-- sqlfluff:dialect: postgres\n"
        "SELECT * FROM table1\n",
        config=config
    )
    assert config.get("dialect") == "ansi"  # dialect should be unchanged
```

### Observed Behaviour

The test above fails:
`AssertionError: assert 'postgres' == 'ansi'`
The file-level override mutates the config object.

### How to reproduce

This test reproduces the issue:
```python
def test__api__immutable_config():
    config = FluffConfig.from_path("test/fixtures/api/config_path_test/extra_configs/.sqlfluff")
    assert config.get("dialect") == "ansi"
    sqlfluff.lint(
        "-- sqlfluff:dialect: postgres\n"
        "SELECT * FROM table1\n",
        config=config
    )
    assert config.get("dialect") == "ansi"
```

### Dialect

Ansi & Postgres, but this has been observed with every dialect.

### Version

3.2.3

### Configuration

I am using the test config available at `test/fixtures/api/config_path_test/extra_configs/.sqlfluff`.

### Code of Conduct

brandonschabell:

Uses a deepcopy of the FluffConfig when parsing files. This prevents the config from being mutated by file-level config overrides.

fixes #6343 

Configs will no longer be mutated when parsing files, so file-level config overrides will now only apply to the file in which they are applied.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/config/fluffconfig.py
+++ b/src/sqlfluff/core/config/fluffconfig.py
@@ -156,7 +156,11 @@ class FluffConfig:
         # the type of a templater in their context, use
         # `get_templater_class()` instead, which avoids instantiating
         # a new templater instance.
-        state["_configs"]["core"].pop("templater_obj", None)
+        # NOTE: It's important that we do this on a copy so that we
+        # don't disturb the original object if it's still in use.
+        state["_configs"] = state["_configs"].copy()
+        state["_configs"]["core"] = state["_configs"]["core"].copy()
+        state["_configs"]["core"]["templater_obj"] = None
         return state
 
     def __setstate__(self, state: Dict[str, Any]) -> None:  # pragma: no cover
@@ -178,6 +182,13 @@ class FluffConfig:
         configs_attribute_copy = deepcopy(self._configs)
         config_copy = copy(self)
         config_copy._configs = configs_attribute_copy
+        # During the initial `.copy()`, we use the same `__reduce__()` method
+        # which is used during pickling. The `templater_obj` doesn't pickle
+        # well so is normally removed, but it's ok for us to just pass across
+        # the original object here as we're in the same process.
+        configs_attribute_copy["core"]["templater_obj"] = self._configs["core"][
+            "templater_obj"
+        ]
         return config_copy
 
     @classmethod
--- a/src/sqlfluff/core/config/fluffconfig.py
+++ b/src/sqlfluff/core/config/fluffconfig.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+from copy import copy, deepcopy
 from itertools import chain
 from typing import (
     TYPE_CHECKING,
@@ -167,6 +168,17 @@ class FluffConfig:
         # NOTE: Likewise we don't reinstate the "templater_obj" config value
         # which should also only be used in the main thread rather than child
         # processes.
+
+    def copy(self) -> FluffConfig:
+        """Returns a copy of the FluffConfig.
+
+        This creates a shallow copy of most of the config, but with a deep copy of the
+        `_configs` dictionary.
+        """
+        configs_attribute_copy = deepcopy(self._configs)
+        config_copy = copy(self)
+        config_copy._configs = configs_attribute_copy
+        return config_copy
 
     @classmethod
     def from_root(
--- a/test/core/config/fluffconfig_test.py
+++ b/test/core/config/fluffconfig_test.py
@@ -5,6 +5,7 @@ import os
 
 import pytest
 
+import sqlfluff
 from sqlfluff.core import FluffConfig, Linter
 from sqlfluff.core.errors import SQLFluffUserError
 from sqlfluff.core.templaters import (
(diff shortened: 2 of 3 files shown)


alanmcruickshank (commented):
Thanks for picking this up!

There are a few things in the `FluffConfig` object which I we don't _need_ to copy, and also which might be expensive to copy (you've already noticed that the templater doesn't copy nicely, but I don't think copying the dialect will be very helpful or efficient). Rather than using `deepcopy` directly and then defining a `__deepcopy__` override method, might I suggest a slightly different approach?

My preference would be to make a public method directly on the `FluffConfig` class, something like `.copy()` (which would mirror what we've done on the `BaseSegment` class. That method would do a _shallow_ copy of the `FluffConfig` (i.e. `copy.copy()`), but then do a `deepcopy` of only the `_config` attribute - which is the only part that actually _needs_ a deepcopy. Something like:

```python
def copy(self) -> FluffConfig:
    config_copy = copy.copy(self)
    # deep copy the _config dict so any further edits don't propogate
    config_copy._config = copy.deepcopy(config_copy._config)
    return config_copy
```

What do you think of that approach?

brandonschabell:
I think that makes perfect sense! I'll make that change and circle back (likely tomorrow morning).

test/core/config/fluffconfig_test.py
```diff
@@ -374,3 +375,14 @@ def test__process_raw_file_for_config(raw_sql):
...
+
+def test__api__immutable_config():
+    config = FluffConfig.from_path(
+        "test/fixtures/api/config_path_test/extra_configs/.sqlfluff"
+    )
+    assert config.get("dialect") == "ansi"
+    sqlfluff.parse(
+        "-- sqlfluff:dialect: postgres\n" "SELECT * FROM table1\n", config=config
+    )
+    assert config.get("dialect") == "ansi"
```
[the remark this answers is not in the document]
brandonschabell: Thanks 😄

brandonschabell:
@alanmcruickshank I added the public `copy` method to the `FluffConfig` class- this should be ready for review, assuming CI passes.

alanmcruickshank (dismissed):
Excellent 👍 .

This is what I had in mind. If the tests pass, I think this is good to go. Thanks for picking this up.

brandonschabell:
@alanmcruickshank Thank you! And sorry- I had missed the docstring in the test. This should be updated and ready for another pass through CI.

src/sqlfluff/core/config/fluffconfig.py
```diff
@@ -168,6 +169,17 @@ def __setstate__(self, state: Dict[str, Any]) -> None:  # pragma: no cover
...
 
+    def copy(self) -> FluffConfig:
+        """Returns a copy of the FluffConfig.
+
+        This creates a shallow copy of most of the config, but with a deep copy of the
+        `_configs` dictionary.
+        """
+        configs_attribute_copy = deepcopy(self._configs)
+        config_copy = copy(self)
+        config_copy._configs = configs_attribute_copy
```
alanmcruickshank, line 180: There's one failing test in the main suite, which I think will be solved by moving across a reference to the `templater_obj` like this:

```suggestion
        config_copy._configs = configs_attribute_copy
        # During the initial `.copy()`, we use the same `__reduce__()` method
        # which is used during pickling. The `templater_obj` doesn't pickle
        # well so is normally removed, but it's ok for us to just pass across
        # the original object here as we're in the same process.
        config_copy.templater_obj = self.templater_obj
```

brandonschabell: I believe there may be a problem with the `FluffConfig.__getstate__` method causing some breakage here as well. If that method is called, the original config will lose its `templater_obj`, which I assume is not intentional.
Example:
```python
config._configs["core"].get("tempalter_obj") # object exists
partial_config_copy = config.__getstate__()
config._configs["core"].get("templater_obj") # returns None
```
This occurs because of the `state["_configs"]["core"].pop("templater_obj", None)` line, which mutates the original config.

This is problematic when the config is held as an attribute within the linter and is used more than once (or if a user were to use the config in multiple parses even without maintaining a single `Linter` object.

alanmcruickshank: Hmmmm good point. 🤔 

I've got an idea. Hold up a sec...

alanmcruickshank:
@brandonschabell - what do you think of this approach?

brandonschabell:
> @brandonschabell - what do you think of this approach?

That looks great!! I was trying around with something very similar. Looks good to me!

alanmcruickshank (approved):

PR #6377, opened 2024-10-20
Support identifier clause for Databricks

PaulBurridge:
Provides support for IDENTIFIER clause.

Supports IDENTIFIER(string or string expression) in stead of database, schema, table or view names as per [documentation].(https://docs.databricks.com/en/sql/language-manual/sql-ref-names-identifier-clause.html)

Not that I am aware of apologies if I have missed something as this is my first PR.

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_databricks.py
```diff
@@ -348,7 +348,63 @@
...
+class TableReferenceSegment(ObjectReferenceSegment):
+    """A reference to an table, CTE, subquery or alias."""
+
+    type = "table_reference"
+
+
+class SchemaReferenceSegment(ObjectReferenceSegment):
+    """A reference to a schema."""
+
+    type = "schema_reference"
```
alanmcruickshank, line 391: I don't think these need redefining? I can't see any edits vs the classess created in the ansi dialect, so no need to redefine. Unless I've missed something?

alanmcruickshank (commented):
I think you've redefined a couple of objects unnecessarily - otherwise this looks good 👍 .

Thanks for the contribution

src/sqlfluff/dialects/dialect_databricks.py
PaulBurridge: @alanmcruickshank 

These were needed to inherit changes to `ObjectReferenceSegment` which has been updated, without the tests fail.

alanmcruickshank: iiiiiineresting. Good catch. Not super helpful from a development point of view, but I guess that's what it is right now.

Thanks for the contribution!

PR #3610, opened 2022-07-14
Option to show errors on fix

chaimt:
### Brief summary of the change made
When running fix with pre-commit, errors are not shown only 
unfixable linting violations found

Added option of show_errors, to display the errors when running fix.
This is to limit the need to run lint after fix

### Pull Request checklist
- [ X ] Please confirm you have completed any of the necessary steps below.

 Added 2 test for fix, one with show_errors and one without using sql that has errrors

PR #6382, opened 2024-10-22
Fix: Parse violations not being shown when run `fix` command with `--show-lint-violations`

joaopamaral:
When a parse error occurs, the `fix` command does not display any output error when the `--show-lint-violations` flag is enabled:

```
All Finished!
```

This PR change to displays all errors when the `--show-lint-violations` option is enabled:

```
== [test/fixtures/linter/parse_lex_error.sql] FAIL
L:   9 | P:  21 |  PRS | Couldn't find closing bracket for opening bracket.
L:   9 | P:  22 |  LXR | Unable to lex characters: '\uf8ffü§∑‚Äç‚ôÄ...'
All Finished!
```

No

### Pull Request checklist

  - Other.


alanmcruickshank (commented):
- ✅ I've gone back and checked against the PR that originally introduced this feature (#3610) and I think this is a good addition/update to that feature.
- ✅ I'm tempted to change the name of the CLI option to reflect that strictly it's showing more than `lint` errors, but in I think it's probably better to keep it as-is for now with the current name.
- 🤔 I wonder whether with this option set to true, whether the `[1 templating/parsing errors found]` is still necessary. I'm tempted to remove that, as it was only really there to summarise errors that we're not showing - when we show them, I don't know if they need to be summarised.

My bigger question here is how to pass the signal through to the formatter. I can see that you've had to add `show_lint_violations` as an extra argument on a bunch of `Linter` functions here, including passing it through the multiprocessing runner architecture. This totally _works_ but I wonder if if adds more complexity than necessary.

What if instead of passing this to the `Linter`, we actually pass the value to the formatter instead (i.e. we pass the value into `get_linter_and_formatter` and then it's one of the keywords we give `OutputStreamFormatter` on instantiation). We'd need to give it a more helpful name, but that could simplify the calls to `dispatch_file_violations` and `print_out_residual_error_counts` because the formatter has already been set up to know whether to print out various things.

Does that make sense? In essence, trying to keep the formatting logic in the formatter, and the linting logic in the Linter. I realise that distinction isn't particularly clear in the codebase at the moment, but I think there's an opportunity to make it a little clearer with this step rather than make it less clear.

joaopamaral:
Thanks for the review @alanmcruickshank!

> Does that make sense? 

Yes, makes total sense! Let me work on it.

alanmcruickshank (approved):
Nice. LGTM. Thanks for the edits

PR #6392, opened 2024-10-23
New Rule ST10: const expression checker

### LINKED ISSUE #5213: Check and avoid circular `JOIN ON` conditions

### LINKED ISSUE #6350: Const expression rule
Personally I think this is yet another different rule maybe.
`fixed_bool_condition_style`: `TRUE/FALSE, 1 = 1/ =0, never`

Personally I hate this style cause its redundant code, and i find it funny that users think they need to evaluate something to get a condition to be TRUE/FALSE. However sadly its also not possible to remove cause people often use it so that all other condition are `AND $cond`

_Originally posted by @OTooleMichael in https://github.com/sqlfluff/sqlfluff/pull/6239#discussion_r1797637959_

rogalski:

This PR introduces a rule that catches expressions that evaluates to a constant value (TRUE or FALSE).

Fixes #6350 
Fixes #5213

Should allow-list for expression be configurable?

### Pull Request checklist

  - Other.


src/sqlfluff/rules/structure/ST10.py
```diff
@@ -0,0 +1,71 @@
...
+    **Best practice**
+
+    .. code-block:: sql
+
+         -- TODO
+        SELECT 1;
+
+    """
+
+    name = "structure.const_expression"
```
alanmcruickshank, line 28: I think the name here isn't obvious. Maybe something like `constant_condition` might be clearer?

```suggestion
    name = "structure.constant_condition"
```

src/sqlfluff/rules/structure/ST10.py
```diff
@@ -0,0 +1,71 @@
...
+        SELECT 1;
+
+    **Best practice**
+
+    .. code-block:: sql
+
+         -- TODO
+        SELECT 1;
+
+    """
```
alanmcruickshank, line 26:
```suggestion
    """Redundant constant condition.
    
    Including a condition in a ``WHERE`` clause which always evaluates to
    either ``TRUE`` or ``FALSE`` regardless of the input columns is
    unnecessary and makes statements harder to read and understand.
    
    Constant conditions are sometimes mistakes (by mistyping the column
    name intended), and sometimes the result of incorrect information that
    they are necessary in some circumstances. Given this ambiguity, this
    rule does not suggest an automatic fix, and instead invites the user
    to resolve the problem manually.

    **Anti-pattern**

    .. code-block:: sql

        SELECT *
        FROM my_table
        -- This following WHERE clause is redundant.
        WHERE my_table.col = my_table.co1

    **Best practice**

    .. code-block:: sql

        SELECT *
        FROM my_table
        -- Replace with a condition that includes meaningful logic,
        -- or remove the condition entirely.
        WHERE my_table.col > 3 

    """
```

alanmcruickshank (commented):
Looks good @rogalski .

I think the logic isn't too ugly at all. I've recommended some blurb for the rule and potentially a new name.

Could you include a couple more failing test cases just so there's a bit more breadth to them? Otherwise, when you're ready, mark this as ready for review and we can get it merged 👍

src/sqlfluff/rules/structure/ST10.py
rogalski: It's any expression, not necesarily a condition.

```sql
SELECT col1 = col1 AS are_equal
FROM tbl
```

This will (and should) trigger warning.

I'll add that to test cases.

src/sqlfluff/rules/structure/ST10.py
rogalski: I'll apply these docs as starting point, but here as well - we have constant_expression, not constant condition.

rogalski:
@alanmcruickshank 

Docs are improved.
Special case for `literal = literal` was introduced.
Test cases are more exhaustive.

If we agree that check is valuable, the only pending question are:
- is this strictness for `literal = literal` justified?
- If so, should we make the allowlist configurable?

rogalski:
I've also merged main.
After the fact I've noticed that main is failing, meaning this PR will likely be failing as well.

OTooleMichael (approved):
LGTM. 
It think this could be fix compatible however. If we just replaced the expression with `TRUE/FALSE` 

the only exception is that `a.col = a.col` is probably just a typo and shouldn't be fixed.

If `WHERE 1 = 1 AND foo > 2` fixed to `WHERE TRUE AND foo > 2`

Something - maybe this rule, maybe another could later do
`WHERE foo > 2`

test/fixtures/rules/std_rule_cases/ST10.yml
```diff
@@ -0,0 +1,85 @@
...
+  # 'a'!='b' should be true (or match allow-list)
+  fail_str: |
+    select col from foo WHERE 'a'!='b' AND col = 'val'
+
+test_fail_disallowed_literal_false:
+  # 1 = 2 should be false (or match allow-list)
+  fail_str: |
+    select col from foo WHERE 1 = 2 OR col = 'val'
+
+test_pass_disallowed_literal_false2:
```
WittierDinosaur, line 59:
```suggestion
test_fail_disallowed_literal_false2:
```

src/sqlfluff/rules/structure/ST10.py
```diff
@@ -0,0 +1,93 @@
...
+                    None,
+                )
+                if not lhs or not rhs:
+                    # Should be unreachable with correctly parsed tree
+                    continue  # pragma: no cover
+
+                # literals need explicit handling (due to well-defined allow-list)
+                if lhs.is_type("literal") and rhs.is_type("literal"):
+                    expr_s = f"{lhs.raw_normalized()} {seg.raw} {rhs.raw_normalized()}"
+                    if expr_s in ("1 = 1", "1 = 0"):
```
WittierDinosaur, line 85: Apologies for the delay in getting this merged. One comment though. This allowlist should be it's own variable so that it's not hidden in the code. It should also be explicitly documented, so that it's obvious to the user what's going on

alanmcruickshank: I've addressed this one. Will get this PR merged shortly 👍

PR #6416, opened 2024-11-01
Fix docs for CP04 config and add test cases

alanmcruickshank:
Very minor bugfix. The current rule docs for CP04 don't show any configuration values. That's not because they're not there, but it's because they're totally inherited. That also means it's ambiguous as to which capitalisation policy that rule uses (as I found out on a project recently 🤦 ).

This fixes the docs, and add more test cases for CP04 to show it uses the root capitalisation policy.

No actual functionality changes here.


WittierDinosaur (approved):
LGTM

PR #6447, opened 2024-11-10
Standardise json operator spacing between dialects

### LINKED ISSUE #6404: Inconsistent default spacing rules for JSON operators between PostgreSQL and SQLite
### Search before asking

### What Happened

PostgreSQL, MySQL, and SQLite all support the JSON operators `->` and `->>`. However, SQLFluff applies different spacing rules around these operators depending on the dialect.

* For SQLite and MySQL, SQLFluff expects no whitespace around the operator ("touch" behavior).
    * `SELECT a -> 'b'` results in two LT01 failures: "Unexpected whitespace before column path operator '->'." and "Unexpected whitespace before quoted literal."
    * `SELECT a->'b'` results in no linting failures.
* For PostgreSQL, SQLFluff expects whitespace around the operator ("single" behavior).
    * `SELECT a -> 'b'` results in no linting failures.
    * `SELECT a->'b'` results in two LT01 failures: "Expected single whitespace between naked identifier and binary operator '->'." and "Expected single whitespace between binary operator '->' and quoted literal."

This difference in behavior appears to be due to the `postgres` dialect treating the operator as a binary operator, while the `mysql` and `sqlite` dialects are treating it as a "column path operator".

Now personally, I find the "single" style to be more readable in my queries than the "touch" style, but we know that "[If there is one part of building a linter that is going to be controversial it’s going to be whitespace](https://docs.sqlfluff.com/en/stable/configuration/layout.html)". Nevertheless, I tried following the documentation to create a `.sqlfluff` file to get `column_path_operator` to use the "single" style, but it had no effect.

```
[sqlfluff:layout:type:column_path_operator]
spacing_before = single
spacing_within = single
spacing_after = single
```

Therefore, there are probably two things that should be looked at:
1. Consistency in whitespace style around JSON operators between the various dialects
2. Verify that `column_path_operator` or the relevant layout type can have configurable spacing

Many thanks!

### Expected Behaviour

The query `SELECT a -> 'b' FROM foo` should have consistent whitespace lint rules among the `postgres`, `sqlite`, and `mysql` dialects.

### Observed Behaviour

The query `SELECT a -> 'b' FROM foo` fails linting on sqlite and mysql, while the opposite `SELECT a->'b' FROM foo` fails linting on postgres.

### How to reproduce

```bash
cat >test.sql <<EOF
SELECT a -> 'b' FROM foo;
SELECT a->'b' FROM foo;
EOF

# should produce the same output for all three dialects
for dialect in mysql sqlite postgres; do
    sqlfluff lint --verbose --dialect=$dialect test.sql; 
done  
```

### Dialect

sqlite, mysql, postgres

### Version

sqlfluff, version 3.2.5
Python 3.11.10

### Configuration

```
[sqlfluff:layout:type:column_path_operator]
spacing_before = single
spacing_within = single
spacing_after = single
```

### Code of Conduct

WittierDinosaur:
Fixes #6404

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -95,6 +95,10 @@ line_position = trailing
 spacing_within = touch
 line_position = leading
 
+[sqlfluff:layout:type:column_path_operator]
+spacing_within = touch
+line_position = leading
+
 [sqlfluff:layout:type:statement_terminator]
 spacing_before = touch
 line_position = trailing
--- a/src/sqlfluff/dialects/dialect_mysql.py
+++ b/src/sqlfluff/dialects/dialect_mysql.py
@@ -213,6 +213,12 @@ mysql_dialect.replace(
         ],
         at=0,
     ),
+    BinaryOperatorGrammar=ansi_dialect.get_grammar("BinaryOperatorGrammar").copy(
+        insert=[
+            Ref("ColumnPathOperatorSegment"),
+            Ref("InlinePathOperatorSegment"),
+        ]
+    ),
     ArithmeticBinaryOperatorGrammar=ansi_dialect.get_grammar(
         "ArithmeticBinaryOperatorGrammar"
     ).copy(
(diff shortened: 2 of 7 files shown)


alanmcruickshank:
> This is an interesting one. Although you've made it consistent across dialects, I bet most people use SQLFluff for a single dialect and this seems like a pseudo–breaking change for them.

Agreed - although I think we've chosen the correct default going forward. I'll tag this with the minor release tag.

Issue #6455, opened 2024-11-13
sqlfluff snowflake dialect unable to parse "create or alter table" statement


ljones9119:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

The sqlfluff **snowflake** dialect is unable to parse a file with a "create or alter table" statement.

Ran sqlfluff lint on a file with a "create or alter table" statement:

```
CREATE or alter TABLE TEST (COL1 INTEGER,
    col2 integer
    , col3 integer
);

```

Got this:

```
python -m sqlfluff lint DAPROD\TEMP\TABLES\test.sql
== [DAPROD\TEMP\TABLES\test.sql] FAIL
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'CREATE or
                       | alter TABLE TEST (COL1 INTEGER...'
WARNING: Parsing errors found and dialect is set to 'snowflake'. Have you configured your dialect correctly?
All Finished!

Process finished with exit code 1
```

My .sqlfluff file is configured like so:

```
[sqlfluff]
dialect = snowflake

max_line_length = 120
comma_style = leading
indent_unit = space
indent_width = 4

[sqlfluff:rules:capitalisation.keywords]
capitalisation_policy = upper
[sqlfluff:rules:capitalisation.identifiers]
extended_capitalisation_policy = upper
[sqlfluff:rules:capitalisation.functions]
extended_capitalisation_policy = upper
[sqlfluff:rules:capitalisation.literals]
capitalisation_policy = upper
[sqlfluff:rules:capitalisation.types]
extended_capitalisation_policy = upper
```

### Expected Behaviour

I would expect the snowflake dialect to recognise the "create or alter table" and parse the file. 

### Observed Behaviour

Parsing errors are found:

```
python -m sqlfluff lint DAPROD\TEMP\TABLES\test.sql
== [DAPROD\TEMP\TABLES\test.sql] FAIL
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'CREATE or
                       | alter TABLE TEST (COL1 INTEGER...'
WARNING: Parsing errors found and dialect is set to 'snowflake'. Have you configured your dialect correctly?
All Finished!

Process finished with exit code 1
```

### How to reproduce

Create a file containing a "create or alter table" statement:

```
CREATE or alter TABLE TEST (COL1 INTEGER,
[...]

PR #6497, opened 2024-12-11
Add support for Snowflake's `CREATE OR ALTER` syntax

korverdev:

Resolves https://github.com/sqlfluff/sqlfluff/issues/6455.

`N/A`

Relevant documentation for easier review of this change: https://docs.snowflake.com/en/sql-reference/sql/create-or-alter


alanmcruickshank (approved):
LGTM!. Thanks for the contribution 🚀

PR #6500, opened 2024-12-11
Update acronym ETL/ELT in README.md

galenseilis:
It is "ETL" not "ELT"

https://en.wikipedia.org/wiki/Extract,_transform,_load

Assuming I inferred the desired acronym. Feedback appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 [![Gurubase](https://img.shields.io/badge/Gurubase-Ask%20SQLFluff%20Guru-006BFF?style=flat-square)](https://gurubase.io/g/sqlfluff)
 
 **SQLFluff** is a dialect-flexible and configurable SQL linter. Designed
-with ELT applications in mind, **SQLFluff** also works with Jinja templating
+with ETL applications in mind, **SQLFluff** also works with Jinja templating
 and dbt. **SQLFluff** will auto-fix most linting errors, allowing you to focus
 your time on what matters.
 


alanmcruickshank:
The acronym is intentionally ELT, referring to the paradigm of transforming _after_ load rather than before - however we probably don't need to be that exclusive. I'd be up for including both if you think that's a good middle ground: `ETL/ELT`.

What do you think?

https://www.techtarget.com/searchdatamanagement/definition/Extract-Load-Transform-ELT

galenseilis:
> The acronym is intentionally ELT, referring to the paradigm of transforming _after_ load rather than before - however we probably don't need to be that exclusive. I'd be up for including both if you think that's a good middle ground: `ETL/ELT`.
> 
> What do you think?
> 
> https://www.techtarget.com/searchdatamanagement/definition/Extract-Load-Transform-ELT

Ah! Thank you for that link! That was helpful in making me aware of the context.

Actually, I think ELT is fine. 

I have added a commit to link to that explanation you provided. My rationale for that change is that it will help people like myself quickly understand what "ELT" is rather than "ETL". May I have some feedback on that?

alanmcruickshank (approved):
Nice. Sorry that this is a bit slow to merge 👍

galenseilis:
> # Coverage Results ✅
> ```
> Name    Stmts   Miss  Cover   Missing
> -------------------------------------
> TOTAL   19304      0   100%
> 
> 248 files skipped due to complete coverage.
> ```

Neat automation!

galenseilis:
> Nice. Sorry that this is a bit slow to merge 👍

All good! :) It is a low-value change so I expect it to be low priority.

Issue #6494, opened 2024-12-11
ALTER TABLE DROP COLUMN statements in TSQL fail if dropping multiple columns


ibeltran:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

I linted an `ALTER TABLE DROP COLUMN` statement with 2 columns and sqlfluff returned `Found unparsable section`

### Expected Behaviour

When running parsing or linting on `ALTER TABLE DROP COLUMN` statements, properly formatted/written statements that list multiple columns should pass linting and parsing

### Observed Behaviour

Parsing fails on the second column in the list, but passes if I only drop one column. However, this statement works in SQL Server without issue, and is valid according to [SQL Server documentation](https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-ver16)

### How to reproduce

**Used SQL (drop_columns.sql)**
```
ALTER TABLE table_name
DROP COLUMN column1, column2;
```
**Steps**
1. I used the latest docker image to test it, on a Windows machine
2. Execute `docker run --rm -v ${PWD}:/sql sqlfluff/sqlfluff:3.3.0 lint /sql/drop_columns.sql --dialect=tsql` 
3. This is the error I get
```== [drop_columns.sql] FAIL
L:   2 | P:  20 |  PRS | Line 2, Position 20: Found unparsable section: ',
                       | column2;'
WARNING: Parsing errors found and dialect is set to 'tsql'. Have you configured your dialect correctly?
```

### Dialect

tsql

### Version

Docker image `sqlfluff/sqlfluff:3.3.0`

### Configuration

None

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

PR #6501, opened 2024-12-11
TSQL dialect - Allow for multiple columns in ALTER TABLE DROP COLUMN statements

korverdev:
…statements

Fixes https://github.com/sqlfluff/sqlfluff/issues/6494.

`N/A`


alanmcruickshank (approved):
LGTM! Thanks for the contribution

Issue #6499, opened 2024-12-11
Gaps in Snowflake `GRANT` syntax parsing


korverdev:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

SQLFluff throws a parsing error when it encounters a valid Snowflake SQL statement.

### Expected Behaviour

This should correctly parse.

### Observed Behaviour

The following error is printed.
```
== [test.sql] FAIL                                                                                                                                                                       
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'GRANT
                       | OPERATE ON ALL DYNAMIC TABLES IN S...'
```

### How to reproduce

The following SQL produces the error when linted using `--dialect=snowflake`.
```sql
GRANT OPERATE ON ALL DYNAMIC TABLES IN SCHEMA my_database.my_schema TO ROLE my_role;
```

### Dialect

Snowflake

### Version

v3.3.0

### Configuration

`N/A`

### Are you willing to work on and submit a PR to address the issue?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

PR #6503, opened 2024-12-12
Snowflake - Fill in gaps in `GRANT` command parsing

korverdev:
Fixes https://github.com/sqlfluff/sqlfluff/issues/6499.

`N/A`


keraion (approved):
LGTM. Thanks!

PR #6505, opened 2024-12-13
ST10: Fix edge cases

### LINKED ISSUE #6492: ST10: False positive on Redundant constant expression
### Search before asking

### What Happened

I upgraded to SQLFluff `3.3.0` and now below part of one of my queries reports an error where there. is none:

```sql
        case
            when (rank() over (
                partition by category_id
                order by year_count, month_count
            ) - 1) % 12 = 0
```

The error is on the `=` after 12 and it returns:

```shell
L: 153 | P: 43 | ST10 | Redundant constant expression. [structure.constant_expression]
```

Given that the `rank() over (..)` does not return a static value the evaluation of that line is also not static.

### Expected Behaviour

Not have an error in this scenario

### Observed Behaviour

See What Happened.

### How to reproduce

I tried to make a minimal case:

```sql
select case when 1 % 12 = 0 then 1 end as test
```

Not really sure if the ST10 is supposed to pick up this statement or not given that now my input is indeed a constant and could be simplified. So what I think is maybe a better minimal test case is:

```sql
with
cte as (
    select 1 as num
    union all
    select 2 as num
    union all
    select 3 as num
    union all
    select 4 as num
)

select
    *,
    case
        when num % 2 = 0 then 1
        else 2
    end as test
from cte
```

### Dialect

dbt-postgres

### Version

```shell
sqlfluff --version           
sqlfluff, version 3.3.0
```

```shell
python --version               
Python 3.11.9
```

### Configuration

My `.sqlfluff`:

```
[sqlfluff]
# Or run 'sqlfluff dialects'
dialect = postgres
templater = dbt
exclude_rules = L031
output_line_length = 160
max_line_length = 160
# The more efficient check is the _byte_ limit check which
# is enabled by default. The previous _character_ limit check
# is still present for backward compatibility. This will be
# removed in a future version.
# Set either to 0 to disable.
large_file_skip_byte_limit = 50000

[sqlfluff:templater:dbt]
project_dir = .
profiles_dir = ~/.dbt

[sqlfluff:templater:jinja]
apply_dbt_builtins = True

[sqlfluff:rules]
single_table_references = unqualified

# Keywords
capitalisation_policy = lower

# Unquoted identifiers

# Function names

# Null & Boolean Literals
capitalisation_policy = lower

# Data Types

[sqlfluff:rules:layout.long_lines]
# Line length
ignore_comment_lines = True
ignore_comment_clauses = True

ignore_words = date,year_month,month,name,attribute,version,level,path,bit

[sqlfluff:rules:structure.subquery]
forbid_subquery_in = both

[sqlfluff:rules:ambiguous.join]
fully_qualify_join_types = both

[sqlfluff:rules:ambiguous.column_references]
group_by_and_order_by_style = explicit

[sqlfluff:rules:references.special_chars]

[sqlfluff:rules:convention.casting_style]
# SQL type casting
preferred_type_casting_style = cast

[sqlfluff:layout:type:select_clause]
line_position = strict

[sqlfluff:layout:type:where_clause]
line_position = strict

[sqlfluff:layout:type:from_clause]
line_position = strict

[sqlfluff:layout:type:join_clause]
line_position = strict

[sqlfluff:layout:type:groupby_clause]
line_position = strict

[sqlfluff:layout:type:having_clause]
line_position = strict

```

### Code of Conduct

### LINKED ISSUE #6493: ST10: Regression with 3.3.0, causes false positives with named parameters
### Search before asking

### What Happened

After upgrading from sqlfluff 3.2.0 to 3.3.0, existing queries are now reporting a redundant constant expression. However, it seems to caused by a placeholder parameter that is the same name as the variable on the left hand side of the equality. 

I can reproduce the error with a simple query. Create a file called `test.sql`

```sql
SELECT a FROM b
WHERE a = :a
```

Then run the command: `sqlfluff lint --dialect=postgres test.sql`

```
== [test.sql] FAIL                                                                                                                                                                                      
L:   2 | P:   9 | ST10 | Redundant constant expression.
                       | [structure.constant_expression]
```

If you change the name of the named parameter to something like `:ab`, the error goes away.

I don't know if this is related to https://github.com/sqlfluff/sqlfluff/issues/6492, but they are regressions related to the same version and same rule.

### Expected Behaviour

I don't think this should be considered a redundant constant expression, since the right hand side is not constant.

### Observed Behaviour

See above.

### How to reproduce

See the sql and command above, with the configuration below.

### Dialect

postgres

### Version

sqlfluff, version 3.3.0

### Configuration

[sqlfluff]
templater = placeholder

param_style = colon
a = 'a'

### Code of Conduct

rogalski:
Fixes #6493 
Fixes #6492

### Pull Request checklist

  - Other.


keraion:
I've also found one other edge case between array accessors and array literals that we may want to include in this PR. I'm also good letting this one through for the time being, but wanted to see if this changes how you want to handle it.

```sql
SELECT my_array[1] = [1] FROM t;
```

src/sqlfluff/rules/structure/ST10.py
```diff
@@ -84,13 +101,18 @@ def _eval_gen(self, context: RuleContext) -> Iterator[LintResult]:
...
+
                 # literals need explicit handling (due to well-defined allow-list)
                 if lhs.is_type("literal") and rhs.is_type("literal"):
                     expr_s = f"{lhs.raw_normalized()} {seg.raw} {rhs.raw_normalized()}"
                     if expr_s in allowable_literal_expressions:
                         # ignore based on allowlist
                         continue
                 else:
+                    if lhs.type != rhs.type:
+                        continue
```
rogalski, line 115: @keraion This effectively fixes your example and makes the overall checker somewhat stricter, which I assume is a good thing.

keraion: Yeah, I like this.

keraion (approved):
LGTM.

PR #6641, opened 2025-02-09
New Rule: LT15

### LINKED ISSUE #2634: New rule proposal: avoid several empty lines in a row.
I don't know if several empty lines is something that makes sense in some dialects, but if not, I think SQLFluff should have a rule to flag this.

**Anti-pattern**
```sql
SELECT * FROM my_table;


SELECT * FROM my_other_table;
```

**Best practice**
```sql
SELECT * FROM my_table;

SELECT * FROM my_other_table;
```

Removing all empty lines makes no sense at all as it adds clean space to separate a list of statements, like a new paragraph in a novel. But having too much of them in a row creates a too big space that can make your SQL hard to read.

This style rule is also present in [JavaScript](https://eslint.org/docs/rules/no-multiple-empty-lines) for instance, and probably from other languages.

Technically, I would go on looking for `newline` segment type, and if the (at least) 2 next segments type encountered are also `newline`s, the rule should flag. The fix to provide is to simply keep 2 newlines in a row when 3+ are found.

Thanks a lot!

WittierDinosaur:
This PR adds a long-requested rule. It allows the control of how many blank lines there can be in a row. It is configurable to different values inside and outside of a statement.

This changes the default rule set - and so should trigger a minor release.

Fixes #2634 

No

### Pull Request checklist

  - Other.

### DIFF
--- a/test/fixtures/rules/std_rule_cases/LT15.yml
+++ b/test/fixtures/rules/std_rule_cases/LT15.yml
@@ -53,3 +53,30 @@ test_fail_one_empty_line_between_statements:
     rules:
       layout.newlines:
         maximum_empty_lines_between_statements: 1
+
+test_fail_bad_edge_case:
+  fail_str: |
+    SELECT foo
+    FROM
+      bar
+
+
+    ;
+
+
+    SELECT foo
+    ;
+  fix_str: |
+    SELECT foo
+    FROM
+      bar
+
+    ;
+
+    SELECT foo
+    ;
+  configs:
+    rules:
+      layout.newlines:
+        maximum_empty_lines_between_statements: 1
+        maximum_empty_lines_within_statements: 0
--- a/test/fixtures/rules/std_rule_cases/LT15.yml
+++ b/test/fixtures/rules/std_rule_cases/LT15.yml
@@ -14,7 +14,7 @@ test_pass_one_empty_line:
   configs:
     rules:
       layout.newlines:
-          maximum_empty_lines_inside_statements: 1
+        maximum_empty_lines_inside_statements: 1
 
 test_fail_no_empty_lines:
   fail_str: |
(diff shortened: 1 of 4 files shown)


test/fixtures/rules/std_rule_cases/LT15.yml
```diff
@@ -0,0 +1,82 @@
...
+    ;
+
+    SELECT foo
+    ;
+  configs:
+    rules:
+      layout.newlines:
+        maximum_empty_lines_between_statements: 1
+
+test_fail_bad_edge_case:
```
WittierDinosaur, line 57: @alanmcruickshank here is the edge case that i'd like to avoid

alanmcruickshank: I see what you mean 🤔 . I'm really not sure this is a problem. Personally I feel like the semicolon "belongs" to the preceding statement, but I also think a comma "belongs" to the preceding line. Leaving this open so that we're a bit more lenient on whether the "between statements" gap is _before_ or _after_ the semicolon I think is ok.

I don't think hard coding the behaviour is worth the extra complexity unless we also give people the option to configure it in case they want to do.... _leading semicolons?!_.

alanmcruickshank: I think I' trying to say: I think this PR is ok as-is. We can always optionally tighten it later.

alanmcruickshank: I'll add a comment to that test case to explain

alanmcruickshank (approved):
Nice. I don't think the edge case is a problem at the moment. Yeah some people will disagree but I think that's small potatoes and can be handled at a later date. This already adds a bunch of value as-is.

PR #6660, opened 2025-02-17
`ON CONFLICT` indentation for Postgres

### LINKED ISSUE #6659: Postgres `ON` preventing formatting after occurrence.
### Search before asking

### What Happened

I note that any query that I have is not formatted correctly (in terms of indentation) after a `on conflict (..) do` statement.
Within the query, anything that is before the `on` is formatted absolutely perfectly, but after it fails to indent.

Example of SQL to format:
```
insert into foo (
id,
bar
)
values (
$1,
$2
) on conflict (id) do update
set
bar = $2;
```

### Expected Behaviour

(I think I should expect this - at least, just any indentation after the `do`...)
```
insert into foo (
    id,
    bar
)
values (
    $1,
    $2
)
on conflict (id) do update
    set
        bar = $2;
```

If I have just a `update` statement (without a `do`) it indents as follows just fine.
```
update foo
set
bar = $2;
```
is converted to
```
update foo
set
    bar = $2;
```

### Observed Behaviour

Any query that I have is indented absolutely fine until a `on conflict ... do update` is reached.
Running `fix` on the above yields:

```
insert into foo (
    id,
    bar
)
values (
    $1,
    $2
)
on conflict (id) do update
set
bar = $2;
```

### How to reproduce

Here is some example SQL that will fail to indent after `on conflict .. do update` 

```
insert into foo (
    id,
    bar
)
values (
    $1,
    $2
)
on conflict (id) do update
set
bar = $2;
```

### Dialect

Postgres

### Version

sqlfluff, version 3.3.1

### Configuration

[.sqlfluff](https://github.com/user-attachments/files/18827374/sqlfluff.txt)

### Code of Conduct

WillMatthews:
Hoping to close #6659 
I *think* this is the change that's needed to resolve the problem I'm seeing.

This is my first contribution here -- so I hope I'm doing everything right! Please do let me know if I'm not. :smile: 

Postgres `ON CONFLICT` statements had no indentation for the `SET` or inside the `SET` stmt. See #6659 for an example of the lacking indentation.

No known side effects. 

### Pull Request checklist
  -  `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.

### TODO
- Add appropriate documentation for the change.


src/sqlfluff/dialects/dialect_postgres.py
```diff
@@ -4996,7 +4996,9 @@ class ConflictActionSegment(BaseSegment):
             "NOTHING",
             Sequence(
                 "UPDATE",
+                Indent,
                 "SET",
+                Indent,
```
WittierDinosaur, line 5001: I think this one should be `ImplicitIndent`

PR #6662, opened 2025-02-17
Add support for Obevo annotations.

### LINKED ISSUE #4135: Add support for Obevo syntax (PostgreSQL dialect extension)
### Search before asking

### Description

We've been wanting to use SQLFluff as a linter for a project we are working on which uses PostgreSQL and [Obevo](https://goldmansachs.github.io/obevo/) to manage the migrations system. These migrations files are plain `.sql` files with a few differences which cause the linter to be unable to parse these files. 

In particular the difference is that Obevo forces you to wrap SQL statements into "changes" using the following syntax:

```
//// CHANGE name="<name>"

[... Your good old PostgreSQL syntax here ...]

GO
```

You can also write a `//// METADATA` line at te begining of the file which is also unparseable by SQLFLuff.

```
//// METADATA includeEnvs="local"
```

But that is pretty much it! These three statements `//// CHANGE`, `//// METADATA` and `GO` are the only three differences with the PostgreSQL syntax (there is actually just two, `//// <anything>` and `GO`).

As these files are just plain `.sql` files written using PostgreSQL syntax, SQLFluff should work just fine using the PostgreSQL dialect, however this few differences make the parser throw some errors which are a nuance:

```
sqlfluff lint centre.sql --dialect postgres
== [centre.sql] FAIL
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: '//// CHANGE name="create" dependencies="...'
L:  25 | P:   2 | L003 | Expected 0 indentations, found 0 [compared to line 24]
L:  27 | P:   1 | L003 | Expected 0 indentations, found 0 [compared to line 24]
```

I've been thinking about creating a **new dialect (extending the one for PostgreSQL)** which sets some rules in order to parse these statements and mark them as comments in a similar way to what is being done with the `meta commands`:

```
postgres_dialect.insert_lexer_matchers(
    [
        # Explanation for the regex
        # \\([^(\\\r\n)])+((\\\\)|(?=\n)|(?=\r\n))?
        # \\                                        Starts with backslash
        #   ([^\\\r\n])+                            Anything that is not a newline or a
        #                                           backslash
        #                 (
        #                  (\\\\)                   Double backslash
        #                        |                  OR
        #                         (?=\n)            The next character is a newline
        #                               |           OR
        #                                (?=\r\n)   The next 2 characters are a carriage
        #                                           return and a newline
        #                                        )
        #                                         ? The previous clause is optional
        RegexLexer(
            # For now we'll just treat meta syntax like comments and so just ignore
            # them. In future we may want to enhance this to actually parse them to
            # ensure they are valid meta commands.
            "meta_command",
            r"\\([^\\\r\n])+((\\\\)|(?=\n)|(?=\r\n))?",
            CommentSegment,
        )
    ],
    before="code",  # Final thing to search for - as psql specific
)
```

However, I'm quite new to SQLFluff and I don't know if this is the best alternative or there is a better way to handle something like this. It seems to me that adding a whole new dialect for something like this is a bit overkill, but I can't think of any other options.

PD: I've tried ignoring the parsing errors using the `--ignore parsing` flag, and everything works just fine! But i've read somewhere that this is not recommended and I can see why.

### Use case

I want Obevo files to be parsed correctly:

```
//// METADATA includeEnvs="local"

//// CHANGE name="createTable"

CREATE TABLE user (
    id INT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
    name TEXT NOT NULL
);

GO
```

### Dialect

The proposed solution is to extend the **PostgreSQL** dialect using a new dialect, but nothing would change in the current **PostgreSQL** dialect.

### Code of Conduct

joaostorrer:
Fixes #4135.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -172,6 +172,11 @@ ansi_dialect.set_lexer_matchers(
             r"(?>\d+\.\d+|\d+\.(?![\.\w])|\.\d+|\d+)(\.?[eE][+-]?\d+)?((?<=\.)|(?=\b))",
             LiteralSegment,
         ),
+        RegexLexer(
+            "obevo_annotation",
+            r"////\s*(CHANGE|BODY|METADATA)[^\n]*",
+            CommentSegment,
+        ),
         RegexLexer("like_operator", r"!?~~?\*?", ComparisonOperatorSegment),
         RegexLexer("newline", r"\r\n|\n", NewlineSegment),
         StringLexer("casting_operator", "::", CodeSegment),
--- a/test/fixtures/dialects/ansi/obevo.sql
+++ b/test/fixtures/dialects/ansi/obevo.sql
@@ -1,0 +1,13 @@
+//// CHANGE name=alter1
+
+ALTER TABLE table1 ADD COLUMN colY NUMBER;
+
+ALTER TABLE table1 ADD COLUMN colZ NUMBER;
+
+//// METADATA test
+//// CHANGE name="init"
+CREATE TABLE OrigTable (
+    Field1 int,
+    Field2 int
+)
+//// CHANGE name="dropOld" DROP_TABLE dependencies="OldToNewTableMigration.migration"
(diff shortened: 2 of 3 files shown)


WittierDinosaur (approved):
LGTM

Issue #5678, opened 2024-03-14
[TSQL] AL09 throws an AssertionError where assigning variables in a select statement


paysni:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

When having sqlfluff run fix on query where the variables are assigned in a select statement, AL09 will throw an error.

sqlfluff\rules\aliasing\AL09.py", line 99, in _eval
    assert whitespace and column_identifier and alias_identifier
AssertionError

### Expected Behaviour

It should be able to handle this valid sql statement.

### Observed Behaviour

It will throw an exception during the fix

sqlfluff\rules\aliasing\AL09.py", line 99, in _eval
    assert whitespace and column_identifier and alias_identifier
AssertionError

if I run lint it will give this error:
L:   4 | P:   1 | AL09 | Unexpected exception: ;
Could you open an issue at
                       | https://github.com/sqlfluff/sqlfluff/issues ?
You can
                       | ignore this exception for now, by adding '-- noqa: AL09'
                       | at the end
of line 4

                       | [aliasing.self_alias.column]

### How to reproduce

created simple query:
declare @variable nvarchar(50)

select @variable = Column1
from Table1
where Column2 = 'test'

and ran sqlfluff fix against that query.

### Dialect

tsql

### Version

3.0.1

### Configuration

[sqlfluff]
dialect = tsql
max_line_length = 120
 #RF06 Exclude forcing Quoted Identifier 
 #CV01 Exclude using != over <>
 #ST06 Exclude column reordering
exclude_rules = RF06, CV01, ST06
large_file_skip_byte_limit = 0
processes = -1

[sqlfluff:rules:layout.long_lines]
ignore_comment_lines = True
ignore_comment_clauses = True

[sqlfluff:layout:type:comma]
spacing_after = touch
line_position = leading

[sqlfluff:rules:capitalisation.identifiers]
ignore_words = dbo
#don't mess with Kronos tables that start with lower case t
ignore_words_regex = ^t[A-z][A-z]*$
extended_capitalisation_policy = pascal

[sqlfluff:indentation]
indent_joins = False
indent_unit = tab
indented_using_on = True
allow_implicit_indents = True
indented_on_contents = False

[sqlfluff:rules:references.special_chars]
ignore_words_regex = ([$(])

[sqlfluff:rules:capitalisation.types]
extended_capitalisation_policy = lower

[...]

PR #6732, opened 2025-03-15
T-SQL: modify parsing variable assignment

### LINKED ISSUE #6000: regression 3.1.0: tsql: variable assignment is incorrectly fixed and breaks code. rule  CV05
### Search before asking

### What Happened

I have updated to sqlfluff 3.1.0.

I wanted to format:
```sql
DECLARE @job_notify_netsend_operator_name bigint;
SELECT @job_notify_netsend_operator_name = NULL;
```

### Expected Behaviour

code is left untouched.

### Observed Behaviour

code is fixed to

```sql
DECLARE @job_notify_netsend_operator_name bigint;
SELECT @job_notify_netsend_operator_name IS NULL;
```

and thus totally broken.

### How to reproduce

see above

```
sqlfluff fix .\query1.sql
== [query1.sql] FAIL
L:   2 | P:  42 | CV05 | Comparisons with NULL should use "IS" or "IS NOT".
                       | [convention.is_null]
== [query1.sql] FIXED
```

### Dialect

tsql

### Version

3.1.0

### Configuration

none

### Code of Conduct

drjwelch:

Fixes #6000 

T-SQL allows a widely-used "legacy" variable assignment syntax thus: `SELECT @variable = <expression>`

Microsoft recommend replacing this syntax with `SET @variable = <expression>` (see [MS docs](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/select-local-variable-transact-sql?view=sql-server-ver16))

Prior to sqlfluff 3.1.0 rule AL09 threw assertion error because this legacy parameter assignment in SELECT was parsed incorrectly as AltAliasSegment.  This was issue #5678.

PR #5934 resolved this by restricting what segment types could be classified as type 'alias_expression'.

This resolution inadvertently caused issue #6000 : in this issue, rule ST05 makes the following **invalid** fix:
- from: `SELECT @variable = NULL`
- to: `SELECT @variable IS NULL`

This did not occur previously because the projection was incorrectly parsed as an alias expression and rule ST05 did not run on the segment; the above fix corrected this but did not correct the behaviour of ST05.  This new incorrect behaviour was a result of the '=' still being parsed as a comparison operator, rather than assignment.

The current change:
- abstracts a new ParameterAssignmentSegment (matching `@myVariable = <expression>`) out of SetStatementSegment and re-uses it in SelectClauseElementSegment;
- adds a new test TQ02 to lint this syntax as an error;
- adds several more test cases;
- adds a new exception to rule RF01 (for assignment to sequences).

TODO: be able to lint **fix** for TQ02 - until this is done many users with legacy codebases (most of us!) might want to disable this rule.  Should we disable it as default?

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_tsql.py
+++ b/src/sqlfluff/dialects/dialect_tsql.py
@@ -774,6 +774,7 @@ class SelectClauseElementSegment(ansi.SelectClauseElementSegment):
     match_grammar = OneOf(
         # *, blah.*, blah.blah.*, etc.
         Ref("WildcardExpressionSegment"),
+        Ref("ParameterAssignmentSegment"),
         Sequence(
             Ref("AltAliasExpressionSegment"),
             Ref("BaseExpressionElementGrammar"),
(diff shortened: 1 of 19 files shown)


drjwelch:
@fmms this should solve the issue.  @greg-finley who is the right person to request PR review, sorry I don't know the process

alanmcruickshank (approved):
Nice. Good Rule. Thanks for taking the time to contribute. We're going to do a minor release soon, and this will be bundled in that 👍

drjwelch:
Thanks @alanmcruickshank - I see the py313 and yml fixture tests failed before you approved - do I need to resolve these?

alanmcruickshank:
@drjwelch - yes, tests will need to pass before I can merge it. It would be excellent if you could resolve those.

drjwelch:
@alanmcruickshank sorry for delay in this; the only conflicts are the test fixture hashes which I think means something else changed since I ran them - can you point me in the right direction for how to resolve that, please?

saulotoledo:
@alanmcruickshank

tboggiano:
Could anyone help resolve this, this would greatly help with my implementation of the SQLFluff which I'm getting ready to do? Our error handling in our TRY...CATCH block uses this type of variable assignment is going cause my devs to grumble to have to put in ignores everytime they submit a PR.

PR #6743, opened 2025-03-22
Drop support for Python 3.8

WittierDinosaur:
### Brief summary of the change made
Python 3.8 hit EOL in October 2024. 5 months on is probably a good time to drop support for it. It allows us to drop a dependency, there's probably some other code clean up we can do without it. This'll trigger a minor release.

### Are there any other side effects of this change that we should be aware of?
About 2% of our downloads are still on Python 3.8. But this still seems like a reasonable time to stop developing new features for it.

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #6745, opened 2025-03-23
Python 3.9+ Type simplifications

alanmcruickshank:
Following on from #6743, this tidies up all the type annotations that used to use py3.8 compatible types. Much cleaner and much wider use of generics.

This PR is mostly find & replace, with only a few manual tweeks. No intended functionality changes.


WittierDinosaur (approved):
LGTM - spot checked a few files, tests passing, low risk despite being a huge PR

PR #6749, opened 2025-03-26
Add new check for ambiguous`LIMIT` or `OFFSET` without `ORDER BY` (non-deterministic results)

### LINKED ISSUE #5940: Warn on missing `ORDER BY` when using `LIMIT` and `OFFSET`
### Search before asking

### Description

Oftentimes it's crucial to use `ORDER BY` with `LIMIT` and `OFFSET` to help guarantee determinism with results (especially when using multiple queries with different offsets). If one forgets to use `ORDER BY` there is a chance the results provided by the query could be duplicative or miss values entirely. This appears in SQL-based database documentation but there often isn't a strict requirement on this.

SQLFluff could lint for the inclusion of `ORDER BY` on detection of `LIMIT` and `OFFSET`. I feel this should be a warning and not an error, perhaps with an option to set it as an error, in order to avoid issues with cases where we truly do want to use `LIMIT` and `OFFSET` without sorting (such as quick samples of data).

### Use case

From the [Postgres docs](https://www.postgresql.org/docs/current/queries-limit.html):
> When using `LIMIT`, it is important to use an `ORDER BY` clause that constrains the result rows into a unique order. Otherwise you will get an unpredictable subset of the query's rows. You might be asking for the tenth through twentieth rows, but tenth through twentieth in what ordering? The ordering is unknown, unless you specified `ORDER BY`.

### Dialect

I imagine there are small variations in dialect here, but I'm not absolutely certain.

### Code of Conduct

d33bs:

This PR adds a new check for explicit `ORDER BY` when using `LIMIT` or `OFFSET`. When `LIMIT` or `OFFSET` are used without `ORDER BY` the results are non-deterministic in many databases, sometimes resulting in unexpected results.

Closes #5940 

This change would push the package towards having an opinion when it comes to ambiguous use of `LIMIT` or `OFFSET`. I didn't receive much feedback on the issue so please don't hesitate to let me know if you think we should ditch this or if there's a better way forward.

I noticed during development that there could be deviation from dialect to dialect in how `LIMIT` or `OFFSET` appear. I don't have as much experience in the realm of the dialects used within the package so I wasn't sure how the new check might impact dialects where the sequencing might differ.

If there's a location I may add additional documentation please don't hesitate to let me know (I tried to include code-based documentation but wasn't as sure about docsite work).

### Pull Request checklist

  - Other.


WittierDinosaur:
A couple of things:
- is this meant to still be in draft?
- Can you remove the coverage.lcov file?

d33bs:
Hi @WittierDinosaur - thanks for the outreach and help here! I've just removed the `lcov` file (apologies about this - it was added by accident when troubleshooting the coverage messages from GH Actions). I'll now convert this to a non-draft PR, I believe it may be ready for review and comments when appropriate.

WittierDinosaur (approved):
LGTM

WittierDinosaur:
Hey, not sure what happened here. I've queued to merge again (sorry for the delay, had limited availability for a while)

WittierDinosaur:
You've got one line of code coverage missing

d33bs:
Thanks @WittierDinosaur and no worries! I've made a change in an attempt to cover the 1 line.

PR #6750, opened 2025-03-26
Tighter typing of Rule configs

alanmcruickshank:
We've had some `Any` types in the Rule configs for a while which I've been wanting to get rid of. This introduces more explicit typing for them and banishes some of some of the remaining `Any` types.

### DIFF
--- a/src/sqlfluff/core/rules/config_info.py
+++ b/src/sqlfluff/core/rules/config_info.py
@@ -13,11 +13,32 @@ of more general wider use - please define it in the specific plugin
 rather than here.
 """
 
-from typing import Any, Dict
+from typing import Dict, List, NotRequired, TypedDict
 
 from sqlfluff.core.plugin.host import get_plugin_manager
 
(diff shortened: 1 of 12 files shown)


src/sqlfluff/core/rules/config_info.py
```diff
@@ -13,11 +13,32 @@
...
+        definition: Required string containing a detailed description of the config
+            option and its purpose. This should be clear enough for users to
+            understand when and how to use the config.
+        validation: Optional list or range of valid values for the config option.
+            Can contain boolean, string, or integer values. If not provided,
+            the config option accepts any value of its expected type.
+    """
+
+    definition: str
+    validation: NotRequired[List[bool | str | int] | range]
```
WittierDinosaur, line 38: NotRequired is Python 3.11+

alanmcruickshank: balls. Back to the drawing board...

PR #6758, opened 2025-03-29
Additional cleanup for python 3.9

keraion:

I found some additional opportunities with dropping python 3.8 and updating things that would likely need to be touched in October 2025.
- Migrate away from deprecated `typing` collections for `collections.abc`
- fix missed PEP 585 typing hints
- cleanup some `__new__` function `super()` calls
- remove pre-3.9 `functools.lru_cache` usage
- `LintingRecord` converted to a `TypedDict` class
- simplified some comprehensions

None, this should functionally be the same.

### Pull Request checklist

  - Other.


alanmcruickshank (changes requested):
Nice. I think we can further simplify the `cache` thing, but not re-exporting, and just importing directly from `functools` in the places that need it.

Otherwise this looks great 💯

alanmcruickshank (approved):

PR #6827, opened 2025-04-18
Refactor Bracketed

WittierDinosaur:
This PR aims to remove recursion from the Bracketed grammar. It matches start bracket, contents, then end bracket, without recursing.

Yes, it rooted out a missing keyword in Oracle. Other than that, every case in the test suite parses exactly the same.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/parser/grammar/sequence.py
+++ b/src/sqlfluff/core/parser/grammar/sequence.py
@@ -6,6 +6,7 @@ from collections.abc import Sequence as SequenceType
 from os import getenv
 from typing import Optional, Union, cast
 
+from sqlfluff.core.errors import SQLParseError
 from sqlfluff.core.helpers.slice import is_zero_slice
 from sqlfluff.core.parser.context import ParseContext
 from sqlfluff.core.parser.grammar.base import (
@@ -14,15 +15,15 @@ from sqlfluff.core.parser.grammar.base import (
 )
 from sqlfluff.core.parser.grammar.conditional import Conditional
 from sqlfluff.core.parser.match_algorithms import (
-    resolve_bracket,
     skip_start_index_forward_to_code,
     skip_stop_index_backward_to_code,
-    trim_to_terminator,
 )
 from sqlfluff.core.parser.match_result import MatchResult
 from sqlfluff.core.parser.matchable import Matchable
 from sqlfluff.core.parser.segments import (
     BaseSegment,
+    BracketedSegment,
+    Dedent,
     Indent,
     MetaSegment,
     TemplateSegment,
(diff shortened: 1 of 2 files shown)


test/core/parser/grammar/grammar_sequence_test.py
```diff
@@ -492,17 +474,25 @@ def test__parser__grammar_bracketed_modes(
...
+        # Unclosed greedy brackets always raise errors.
+        (["(", "a"], ParseMode.GREEDY, ["a"], {}),
+        (
+            ["(", " ", ")"],
+            ParseMode.GREEDY,
+            [],
+            # Greedy matching, without allowing gaps, should raise a parsing error.
+            # NOTE: This functionality doesn't get used much.
+            {"allow_gaps": False},
+        ),
```
alanmcruickshank, line 488: 👍 I think I'm good with this behaviour change. It's not used, and as long as the other unparsable cases work, then I think we're good.

src/sqlfluff/core/parser/grammar/sequence.py
```diff
@@ -466,15 +468,11 @@ def match(
     ) -> MatchResult:
         """Match a bracketed sequence of elements.
 
-        Once we've confirmed the existence of the initial opening bracket,
-        this grammar delegates to `resolve_bracket()` to recursively close
-        any brackets we fund until the initial opening bracket has been
-        closed.
+        This implementation matches a bracketed expression without recursion
+         or seeking ahead.
```
alanmcruickshank, line 472: Stray indentation 😉 
```suggestion
        or seeking ahead.
```

src/sqlfluff/core/parser/grammar/sequence.py
```diff
@@ -484,109 +482,107 @@ def match(
...
+                child_matches += content_match.child_matches
+        if end_bracket_idx > gap_start:
+            gap_match = MatchResult(
+                matched_slice=slice(gap_start, end_bracket_idx),
+                matched_class=None,
+                segment_kwargs={},
+                insert_segments=(),
+                child_matches=(),
+            )
+            child_matches += (gap_match,)
```
alanmcruickshank, line 539: We add a gap match at the end, and not at the start. That feels inconsistent, but I'm also not sure about the implications of that given the test suite still passes. Can you review what happens if we remove this, or what happens if we also put one at the start?

I'm assuming there _is_ a reason, and if you work out what that is, I think we should add a comment to explain that.

src/sqlfluff/core/parser/grammar/sequence.py
```diff
@@ -14,7 +15,6 @@
 )
 from sqlfluff.core.parser.grammar.conditional import Conditional
 from sqlfluff.core.parser.match_algorithms import (
-    resolve_bracket,
```
alanmcruickshank, line 17: With the removal of `resolve_bracket` here - I think this is the only real use of the method. We can probably strip it out entirely if it isn't used here anymore.

alanmcruickshank (approved):
In general I think this is great 👍 
- Much cleaner.
- Faster.
- Removes the hacky bits and all my old `TODO` and `NOTE` comments.
- Keeps mostly the same functionality.

Well done 💯 

I've got a few more specific bits of feedback, all as comments in the code, but the main one is:
- Can you check the gap matching logic? On that face of it, it seems inconsistent that we match at the end but not at the start. I have a feeling there's a good reason for this, but I can't remember what that is. If it's not necessary, then that's another simplification we can make, but otherwise it probably just needs an explanatory comment.

src/sqlfluff/core/parser/grammar/sequence.py
alanmcruickshank: ...actually ignore me here. I've checked and it's still used in the greedy matching algorithm.

WittierDinosaur: I want to get to that point - but it'll take a refactor on Anything()

Issue #6847, opened 2025-04-28
large_file_skip_byte_limit not working: '>' not supported between instances of 'int' and 'str'


M-van-alten-BW:
### Search before asking

- [x] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

I wanted to test some files slightly larger than the default limit so set the option `large_file_skip_byte_limit = 0` then ran a `!sqlfluff parse` on the files. 

### Expected Behaviour

I would expect it to then properly parse those files, perhaps taking a bit long. 

### Observed Behaviour

Instead I got an error message of `TypeError: '>' not supported between instances of 'int' and 'str'`, located on the file https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/core/linter/linter.py, line 133. 

I checked this file in the repository and it shows that the OS immediately gets both values. In my current configuration I cannot edit this code for myself, but I suspect that line 129, `limit = file_config.get("large_file_skip_byte_limit")`, returns a string which should be cast to int before being compared. A quick look at the getter code in https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/core/config/fluffconfig.py line 520 confirms that this code doesn't do any type casting. A simple solution to this issue would be to change line 133 to `if file_size > int(limit):`

### How to reproduce

To reproduce, simply set this option, large_file_skip_byte_limit, to anything, then do any parse command.

### Dialect

I used the databricks dialect

### Version

I used sqlfluff version 3.4.0 and python version 3.11.10

### Configuration

A configuration sufficient for reproduction is this:
```
[sqlfluff]
large_file_skip_byte_limit = 0 #increased from 20000 because a few files were a bit larger than the limit
```

### Are you willing to work on and submit a PR to address the issue?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

M-van-alten-BW:
Closed by https://github.com/sqlfluff/sqlfluff/pull/6848

PR #6848, opened 2025-04-28
converted large_file_skip_byte_limit value to integer before comparison, fixes Issue #6847

M-van-alten-BW:
### In the linter, converted the value for large_file_skip_byte_limit to an integer to prevent the error in the comparison
This should fix Issue #6847 with no side effects.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -126,7 +126,7 @@ class Linter:
         config_encoding: str = file_config.get("encoding", default="autodetect")
         encoding = get_encoding(fname=fname, config_encoding=config_encoding)
         # Check file size before loading.
-        limit = file_config.get("large_file_skip_byte_limit")
+        limit = int(file_config.get("large_file_skip_byte_limit"))
         if limit:
             # Get the file size
             file_size = os.path.getsize(fname)


src/sqlfluff/core/linter/linter.py
```diff
@@ -126,7 +126,7 @@ def load_raw_file_and_config(
         config_encoding: str = file_config.get("encoding", default="autodetect")
         encoding = get_encoding(fname=fname, config_encoding=config_encoding)
         # Check file size before loading.
-        limit = file_config.get("large_file_skip_byte_limit")
+        limit = int(file_config.get("large_file_skip_byte_limit"))
```
alanmcruickshank, line 129: I like the approach, but at the moment this will probably just introduce another hard to ready error if someone puts in the wrong value for the config.

In particular, if the value isn't set then we'll get an error because `int(None)` is a `ValueError` think.

alanmcruickshank (dismissed):
You're on the right lines. Can I ask you to do a couple of extra things:

1. Make a test case which covers the change you're making. I can give you some pointers if you're stuck, but if you have a look through the current test suite you'll probably be able to work out something. It should ideally cover setting this value to `None`, `100`, `0` and `"this_is_not_an_int"`.
2. Add a better error message here if the conversion fails, so users know what to do about it if it happens?

WittierDinosaur (approved):
LGTM

WittierDinosaur:
Still got some linting errors

M-van-alten-BW:
@WittierDinosaur Thanks for rerunning. I think I've fixed the linting errors now so I hope it can pass now. Is there a way for me to test this myself btw? Then I wouldn't have to ask a contributor every time.

keraion:
Linting here is still an issue. If you have tox installed, you can run `tox -e linting` to run the linting steps. You can also add the pre-commit hook with `tox -e pre-commit -- install` to check things before committing.

More info on setting things up can be found in the [contributing docs](https://github.com/sqlfluff/sqlfluff/blob/main/CONTRIBUTING.md).

M-van-alten-BW:
I once again believe I have fixed all linting issues with this PR, against all previous experience

PR #6852, opened 2025-04-28
dbup style variables into placeholder templater

dreik:
fixes-6851

not aware of any

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/templaters/placeholder.py
+++ b/src/sqlfluff/core/templaters/placeholder.py
@@ -41,6 +41,10 @@ KNOWN_STYLES = {
     "dollar": regex.compile(
         r"(?<![:\w\x5c])\${?(?P<param_name>[\w_]+)}?", regex.UNICODE
     ),
+    # e.g. WHERE bla = $name or WHERE bla = ${name} (DbUp compatible)
+    "dollar_surround": regex.compile(
+        r"(?<![:\w\x5c])\$(?P<param_name>[\w_]+)\$?", regex.UNICODE
+    ),
     # e.g. USE ${flyway:database}.schema_name;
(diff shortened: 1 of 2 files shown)


src/sqlfluff/core/templaters/placeholder.py
```diff
@@ -41,6 +41,10 @@
     "dollar": regex.compile(
         r"(?<![:\w\x5c])\${?(?P<param_name>[\w_]+)}?", regex.UNICODE
     ),
+    # e.g. WHERE bla = $name or WHERE bla = ${name} (DbUp compatible)
```
keraion, line 44: I think this might be what you've intended here.
```suggestion
    # e.g. WHERE bla = $name$ (DbUp compatible)
```

src/sqlfluff/core/templaters/placeholder.py
```diff
@@ -41,6 +41,10 @@
...
+    "dollar_surround": regex.compile(
+        r"(?<![:\w\x5c])\$(?P<param_name>[\w_]+)\$?", regex.UNICODE
```
keraion, line 46: The documentation suggests that a `-` is valid inside the dollar quotes (`\w` should include `_`). Also, that second `$` I don't think is supposed to be optional.
```suggestion
        r"(?<![:\w\x5c])\$(?P<param_name>[-\w]+)\$", regex.UNICODE
```

keraion (changes requested):
Overall, a good start but I think I see a few issues. Also, I'd like to see some tests. Those should fit in [test/core/templaters/placeholder_test.py](https://github.com/sqlfluff/sqlfluff/blob/a75d0de6b4f4a35deb97e350f571acfe0962dd44/test/core/templaters/placeholder_test.py#L355)

keraion (approved):
LGTM. Thanks for adding the test!

PR #6948, opened 2025-06-16
TSQL: Add support for CREATE TABLE GRAPH statements

peterbud:

This PR is adding support for graph tables in TSQL. These use special syntax at table creation and a specific constraint for such tables.

* Add required keywords
* Add graph table syntax variation with AS NODE and AS EDGE reusing as much as possible
* Add related constraints

PS: My first contribution to this repo, happy to take advice on further improvements in this PR.

No

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/dialects/dialect_tsql.py
+++ b/src/sqlfluff/dialects/dialect_tsql.py
@@ -681,6 +681,7 @@ class StatementSegment(ansi.StatementSegment):
             Ref("SetStatementSegment"),
             Ref("AlterTableSwitchStatementSegment"),
             Ref("PrintStatementSegment"),
+            Ref("CreateTableGraphStatementSegment"),
             Ref(
                 "CreateTableAsSelectStatementSegment"
             ),  # Azure Synapse Analytics specific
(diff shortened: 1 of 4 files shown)


src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -3635,6 +3703,35 @@ class TableConstraintSegment(BaseSegment):
...
+            ),
+            Sequence(  # FOREIGN KEY ( column_name [, ... ] )
+                # REFERENCES reftable [ ( refcolumn [, ... ] ) ]
+                Ref("ForeignKeyGrammar"),
+                # Local columns making up FOREIGN KEY constraint
+                Ref("BracketedColumnReferenceListGrammar"),
+                # REFERENCES reftable [ ( refcolumn) ] + ON DELETE/ON UPDATE
+                Ref("ReferencesConstraintGrammar"),
+            ),
+            Ref("ConnectionConstraintGrammar", optional=True),
```
keraion, line 3719: I wonder if it makes more sense to add this to `TableConstraintSegment` instead. If not, there are a number of references where `TableConstraintSegment` is used, which likely would need this graph version added to those locations as well. I think keeping the parallel logic in a single location would be a bit more maintainable.

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -3459,6 +3487,46 @@ class CreateTableStatementSegment(BaseSegment):
...
+                            Ref("ColumnDefinitionSegment"),
+                            Ref("TableIndexSegment"),
+                            Ref("PeriodSegment"),
+                        ),
+                        allow_trailing=True,
+                    )
+                ),
+            ),
+            optional=True,
+        ),
```
keraion: The outer `OneOf` and `Sequence` each only have a single element inside of it so we can remove those.

The `OneOf` inside of `Delimited` is redundant because `Delimited` is a type of `OneOf`.

```suggestion
        # Columns and comment syntax:
        Bracketed(
            Delimited(
                Ref("GraphTableConstraintSegment"),
                Ref("ComputedColumnDefinitionSegment"),
                Ref("ColumnDefinitionSegment"),
                Ref("TableIndexSegment"),
                Ref("PeriodSegment"),
                allow_trailing=True,
            ),
            optional=True,
        ),
```

peterbud: Thanks, I have had that structure form the original `CreateTableStatementSegment`, I assumed that is efficient - no problem to simplify both classes.

src/sqlfluff/dialects/dialect_tsql.py
peterbud: These constraint sets are different. the `ConnectionConstraintGrammar` is **only** applicable when one creates a graph table  (with the `AS NODE` and `AS EDGE` pieces). I wanted to reuse the base `TableConstraintSegment` and sublass ir and then `.copy` and `insert` the additional element, but as it is a more complex structure I could not make that work. Maybe you have a good idea how we can at least re-use the base `TableConstraintSegment` somehow?

keraion: That makes sense. I'm good with this the way it is.

keraion (approved):
Nice work. LGTM. Thanks for updating the `CreateTableStatementSegment` too!

PR #7025, opened 2025-07-18
In User Mapping ensure that PUBLIC doesn't get parsed as an identifier

Singletoned:

Changed the precedence of `PUBLIC` so that it gets chosen first.  Otherwise it was getting parsed as an identifier, and I was getting warnings saying that it should be lower case.

I have a couple of unit tests that I used to check that it solved the problem, but it seemed overkill to push them for such a small change.  Happy to add them if desired.

None that I can think of.

### Pull Request checklist


keraion (approved):
The yaml lets me know that this change had the intended effect, no need for any additional tests. Thanks for the contribution!

PR #7059, opened 2025-07-28
fix confusion between "toml" and "tomli"

a-detiste:
Hi,

It was realy hard for me too to wrap my head around this, so I documented here: https://wiki.debian.org/Python/Backports

Here's the full evolution story of this library: pytoml -> toml -> tomli -> tomllib

tomli & tomllib should be 99% the same thing


keraion (approved):
Makes sense. Thanks for the contribution!

PR #7100, opened 2025-08-19
Fix parse error with bracketed case statement

### LINKED ISSUE #7099: Valid CASE statement causes parse error in Oracle dialect if WHEN condition uses brackets
### Search before asking

### What Happened

A valid CASE statement had a parsing error when the WHEN clause used brackets.

This statement fails:

```sql
SELECT 
    CASE
        WHEN abc = 1 THEN NULL
        WHEN (
            defg = 2
            AND hijk = 3
        ) THEN NULL
    END AS result
FROM abc;
```

### Expected Behaviour

It should not fail as it is valid

### Observed Behaviour

Fails with a parse error "Could not parse"

### How to reproduce

Using the SQL above with `sqlfluff parse test.sql --dialect oracle`

### Dialect

Oracle

### Version

3.4.2

### Configuration

No config needed apart from `--dialect oracle`

### Code of Conduct

adam-carruthers:

Fixes #7099 

Currently, valid CASE statements with a bracket in a WHERE clause will fail, such as

```sql
SELECT 
    CASE
        WHEN abc = 1 THEN NULL
        WHEN (
            defg = 2
            AND hijk = 3
        ) THEN NULL
    END AS result
FROM abc;
```

This PR fixes that.

No, side effect free.

### Pull Request checklist

### DIFF
--- a/test/fixtures/dialects/oracle/case_expressions.sql
+++ b/test/fixtures/dialects/oracle/case_expressions.sql
@@ -290,4 +290,4 @@ SELECT
             AND hijk = 3
         ) THEN NULL
     END AS result
-FROM abc
+FROM abc;
--- a/test/fixtures/dialects/oracle/case_expressions.yml
+++ b/test/fixtures/dialects/oracle/case_expressions.yml
@@ -3,7 +3,7 @@
 # computed by SQLFluff when running the tests. Please run
 # `python test/generate_parse_fixture_yml.py`  to generate them after adding or
 # altering SQL files.
-_hash: 8aa984698afe8e05dcfab94ac8ad94976382dc195ad14316b7739101591c26bc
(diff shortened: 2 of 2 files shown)


joaostorrer:
Actually the error is not with bracketed case statement, if you try to parse only:
```sql
SELECT 
    CASE
        WHEN (
            defg = 2
            AND hijk = 3
        ) THEN NULL
    END AS result
FROM abc;
```
It will parse without error.

The problem is multiple `WHEN` clauses.

You will have to add `WHEN` keyword as a terminator in `CaseExpressionSegment` to fix this.

adam-carruthers:
Closed in favour of https://github.com/sqlfluff/sqlfluff/pull/7122

Issue #6058, opened 2024-08-10
Redshift: Parsing error while using Shorthand casting with double quote


mrcool4:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

We're attempting to format a large amount of legacy code using SQLFluff, but have encountered an issue with casted columns during testing. The code contains double quotes around the datatype, causing parsing errors and breaking the formatting.
Although this syntax is correct and works in Redshift without issues, it seems the parser is treating the double-quoted string as an identifier rather than a datatype. Is it possible to resolve this issue?
`select null::"varchar" as col1, null::varchar as col2 from table1`

### Expected Behaviour

well formatted code
```
select
    null::"varchar" as col1
    , null::varchar as col2
from table1
```

### Observed Behaviour

`select null::"varchar" as col1, null::varchar as col2 from table1`
[L:  1, P:  8]      |                select_clause_element:
[L:  1, P:  8]      |                    null_literal:                             'null'
[L:  1, P: 12]      |                [META] dedent:
[L:  1, P: 12]      |                unparsable:                                   !! Expected: 'Nothing here.'
[L:  1, P: 12]      |                    casting_operator:                         '::'
[L:  1, P: 14]      |                    double_quote:                             '"varchar"'

### How to reproduce

paste the below code into a .sql file and use the configuration provided.
`select null::varchar as col1, null::varchar as col2 from table1`

### Dialect

Redshift

### Version

sqlfluff:  3.0.3 
python: 3.10.11

### Configuration

```ini
[sqlfluff]

# Supported dialects https://docs.sqlfluff.com/en/stable/dialects.html
# Or run 'sqlfluff dialects'
dialect = redshift

# One of [raw|jinja|python|placeholder]
templater = placeholder

# Comma separated list of rules to exclude, or None
# See https://docs.sqlfluff.com/en/stable/configuration.html#enabling-and-disabling-rules
# AM04 (ambiguous.column_count) and ST06 (structure.column_order) are
# two of the more controversial rules included to illustrate usage.
exclude_rules = ambiguous.column_count, structure.column_order

# The standard max_line_length is 80 in line with the convention of
# other tools and several style guides. Many projects however prefer
# something a little longer.
# Set to zero or negative to disable checks.
max_line_length = 120

# CPU processes to use while linting.
# The default is "single threaded" to allow easy debugging, but this
# is often undesirable at scale.
[...]

PR #7124, opened 2025-09-12
Redshift: Allows quoted identifiers as datatypes

franloza:

Fixes https://github.com/sqlfluff/sqlfluff/issues/6058.

Fixes the issue where double-quoted datatypes in cast expressions were not parsed correctly in the Redshift dialect.

#### Root Cause

The issue was in the Redshift `DatatypeSegment` implementation. While the ANSI dialect supports quoted identifiers as datatypes through the `DatatypeIdentifierSegment` (which includes `SingleIdentifierGrammar` that contains both naked and quoted identifiers), the Redshift dialect completely overrode this behavior with a hardcoded list of datatype keywords only.

When parsing `null::"varchar"`, the parser would:
1. Successfully parse `null` as a null literal
2. Successfully parse `::` as a casting operator  
3. Fail to parse `"varchar"` as a datatype because the Redshift `DatatypeSegment` only accepted hardcoded string literals, not quoted identifiers

#### Solution

This PR adds support for quoted datatypes in cast expressions specifically by adding `QuotedIdentifierSegment` to the list of accepted references in `DatatypeSegment`

This targeted approach allows quoted identifiers specifically in `::` cast expressions while maintaining all existing functionality in other contexts like procedure definitions, table creation, etc.

#### Verification

**Before this fix**, the original issue example would fail with a parsing error:

```sql
select null::"varchar" as col1, null::varchar as col2 from table1
```

```
==== parsing violations ====
L:   1 | P:  12 |  PRS | Line 1, Position 12: Found unparsable section: '::"varchar" as
                       | col1, null::varchar as co...'
```

**After this fix**, the same SQL is parsed correctly.

There are no breaking changes. This is a purely additive change that extends the existing datatype parsing to support quoted identifiers, which is consistent with the ANSI standard and other SQL dialects.

### Pull Request checklist

  - [ ] Other.
- [ ] Added appropriate documentation for the change.

### DIFF
--- a/src/sqlfluff/dialects/dialect_redshift.py
+++ b/src/sqlfluff/dialects/dialect_redshift.py
@@ -644,7 +644,7 @@ class DatatypeSegment(BaseSegment):
         ),
         "ANYELEMENT",
         # quoted simple datatypes
-        Ref("DatatypeIdentifierSegment"),
+        Ref("QuotedIdentifierSegment"),
         Sequence(
             Ref("SingleIdentifierGrammar"),
             Ref("DotSegment"),


WittierDinosaur:
Looks like your ymlchecks are failing - can you regenerate the yml?

franloza:
@WittierDinosaur It looks the solution breaks the `CREATE PROCEDURE` tests. I'll take a look and try to come up with a better solution.

franloza:
Closing for now as the only working alternative I found (Adding `QuotedIdentifierSegment` to `DatatypeSegment`) has a side-effect on `CREATE PROCEDURE` parsing logic. Not sure how to fix this without modifying the logic in `ProcedureParameterListSegment`. I'll re-open if I find a solution

PR #7132, opened 2025-09-17
Write an implementation of the lexer in Rust

keraion:

This is the first step to having a rust-based backend by having an optional rust based lexer. This is very much a work in progress but is an optional install.

None

### Pull Request checklist

  - Other.

### DIFF
--- a/examples/02_timing_api_steps.py
+++ b/examples/02_timing_api_steps.py
@@ -25,7 +25,7 @@ def time_function(func, name, iterations=20):
 
 # Set up some classes to process the data
 kwargs = dict(dialect="ansi")
-lexer = Lexer.build(**kwargs)
+lexer = Lexer(**kwargs)
 parser = Parser(**kwargs)
 linter = Linter(**kwargs)
 
--- a/plugins/sqlfluff-templater-dbt/test/templater_test.py
+++ b/plugins/sqlfluff-templater-dbt/test/templater_test.py
@@ -147,7 +147,7 @@ def _run_templater_and_verify_result(
     assert str(templated_file) == fixture_path.read_text()
     # Check we can lex the output too.
     # https://github.com/sqlfluff/sqlfluff/issues/4013
-    lexer = Lexer.build(config=config)
+    lexer = Lexer(config=config)
     _, lexing_violations = lexer.lex(templated_file)
     assert not lexing_violations
 
@@ -297,7 +297,7 @@ def test__templater_dbt_templating_test_lex(
         config=config,
     )
 
-    lexer = Lexer.build(config=config)
+    lexer = Lexer(config=config)
     # Test that it successfully lexes.
     _, _ = lexer.lex(templated_file)
 
--- a/test/api/classes_test.py
+++ b/test/api/classes_test.py
@@ -7,7 +7,7 @@ test_query = "SELECt 1"
 
 def test__api__lexer():
     """Basic checking of lexing functionality."""
-    tokens, violations = Lexer.build(dialect="ansi").lex(test_query)
+    tokens, violations = Lexer(dialect="ansi").lex(test_query)
     assert violations == []
     assert isinstance(tokens, tuple)
     # The last element is the file end marker.
@@ -16,14 +16,14 @@ def test__api__lexer():
 
 def test__api__parser():
     """Basic checking of parsing functionality."""
-    tokens, _ = Lexer.build(dialect="ansi").lex(test_query)
+    tokens, _ = Lexer(dialect="ansi").lex(test_query)
     parsed = Parser(dialect="ansi").parse(tokens)
     assert parsed.raw == test_query
 
 
 def test__api__linter_lint():
     """Basic checking of parsing functionality."""
-    tokens, _ = Lexer.build(dialect="ansi").lex(test_query)
+    tokens, _ = Lexer(dialect="ansi").lex(test_query)
     parsed = Parser(dialect="ansi").parse(tokens)
     violations = Linter(dialect="ansi").lint(parsed)
     assert [v.rule.code for v in violations] == ["CP01", "LT12"]
@@ -31,7 +31,7 @@ def test__api__linter_lint():
 
 def test__api__linter_fix():
     """Basic checking of parsing functionality."""
-    tokens, _ = Lexer.build(dialect="ansi").lex(test_query)
+    tokens, _ = Lexer(dialect="ansi").lex(test_query)
     parsed = Parser(dialect="ansi").parse(tokens)
     fixed, _ = Linter(dialect="ansi").fix(parsed)
     assert fixed.raw == "SELECT 1\n"
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -119,7 +119,7 @@ def parse_example_file(dialect: str, sqlfile: str):
     # Load the SQL
     raw = load_file(dialect, sqlfile)
     # Lex and parse the file
-    tokens, _ = Lexer.build(config=config).lex(raw)
+    tokens, _ = Lexer(config=config).lex(raw)
     tree = Parser(config=config).parse(tokens, fname=dialect + "/" + sqlfile)
     return tree
 
--- a/test/dialects/ansi_test.py
+++ b/test/dialects/ansi_test.py
@@ -23,7 +23,7 @@ from sqlfluff.core.parser import Lexer
 def test__dialect__ansi__file_lex(raw, res, caplog):
     """Test we don't drop bits on simple examples."""
     config = FluffConfig(overrides=dict(dialect="ansi"))
-    lexer = Lexer.build(config=config)
+    lexer = Lexer(config=config)
     with caplog.at_level(logging.DEBUG):
         tokens, _ = lexer.lex(raw)
     # From just the initial parse, check we're all there
--- a/test/dialects/bigquery_test.py
+++ b/test/dialects/bigquery_test.py
@@ -37,7 +37,7 @@ def test_bigquery_relational_operator_parsing(data):
     note(f"query: {raw}")
     # Load the right dialect
     config = FluffConfig(overrides=dict(dialect="bigquery"))
-    tokens, lex_vs = Lexer.build(config=config).lex(raw)
+    tokens, lex_vs = Lexer(config=config).lex(raw)
     # From just the initial parse, check we're all there
     assert "".join(token.raw for token in tokens) == raw
     # Check we don't have lexing issues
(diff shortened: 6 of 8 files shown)


src/sqlfluff/core/parser/lexer.py
```diff
@@ -887,3 +891,87 @@ def map_template_slices(
                     f"{template.templated_str[template_slice]!r}"
                 )
         return templated_buff
+
+    @classmethod
+    def build(
```
alanmcruickshank: This `Lexer.build()` method feels wierd. I get that we need something that will select and instantiate the correct `Lexer` given the available options. I feel like a simple method like `get_lexer()` (or even just `lexer()`) might be more pythonic. It also makes the type inheritance simpler.

Views?

src/sqlfluff/core/parser/lexer.py
```diff
@@ -887,3 +891,87 @@ def map_template_slices(
...
+            """
+            return cls(
+                config=config,
+                last_resort_lexer=last_resort_lexer,
+                dialect=dialect,
+            )
+
+    lexer_logger.info("Using sqlfluffrs lexer.")
+except ImportError:
+    ...
```
alanmcruickshank: Worth also logging here? 
```suggestion
except ImportError:
    lexer_logger.info("sqlfluffrs lexer not present or failed to load.")
```

src/sqlfluff/core/parser/lexer.py
```diff
@@ -887,3 +891,87 @@ def map_template_slices(
...
+            if (
+                hasattr(subclass, "type") and subclass.type
+            ):  # Ensure the subclass has a type
+                segment_map[subclass.type] = subclass
+            # Recursively add subclasses of subclasses
+            segment_map.update(get_segment_type_map(subclass))
+        return segment_map
+
+    # Dynamically generate the segment_types map
+    segment_types = get_segment_type_map(RawSegment)
```
alanmcruickshank, line 914: This looks like we might need to do it on the specific dialect? If this does what I think it does, and iterates through all the subclasses of `RawSegment`, then those segments might behave differently in different dialects.

src/sqlfluff/core/parser/lexer.py
```diff
@@ -887,3 +891,87 @@ def map_template_slices(
...
+                template.raw_sliced,  # type: ignore
+            )
+
+            return (
+                tuple(
+                    segment_types.get(token.type, RawSegment).from_rstoken(
+                        token, py_template
+                    )
+                    for token in tokens
+                ),
```
alanmcruickshank: This part should probably be a `staticmethod`. Would love to get some tests around it.

sqlfluffrs/sqlfluffrs.pyi
```diff
@@ -0,0 +1,143 @@
...
+    ) -> "RsToken": ...
+    def to_tuple(
+        self,
+        code_only: Optional[bool] = None,
+        show_raw: Optional[bool] = None,
+        include_meta: Optional[bool] = None,
+    ) -> TupleSerialisedSegment: ...
+    def __repr__(self) -> str: ...
+    @property
+    def instance_types(self) -> List[str]: ...
```
alanmcruickshank, line 107: It looks like we convert `RsToken` into `RawSegment` pretty soon after getting back into python land. Do we need all these methods on the segment? It's very possible we do, and I've just misundertsood.

I think this class (and probably the others too) should have dosctrings in this file even if just for IDE context.

sqlfluffrs/src/dialect/ansi/matcher.rs
```diff
@@ -0,0 +1,699 @@
+/* This is a generated file! */
+use once_cell::sync::Lazy;
+use uuid::Uuid;
+use crate::matcher::{LexMatcher, extract_nested_block_comment};
+use crate::token::Token;
+use crate::regex::RegexModeGroup;
+use crate::dialect::Dialect;
```
alanmcruickshank, line 7: So..... 🤔 

If the dialect files are all auto generated (which I think is a good thing), should we commit them to git? I'm tempted to say we _don't_, and that they're `.gitignored`, so that they have to be built directly from the python files on each build cycle.

What do you think?

src/sqlfluff/core/errors.py
```diff
@@ -17,9 +17,17 @@
...
+    try:
+        from sqlfluffrs import RsSQLLexerError
+    except ImportError:
+        ...
+
+
 CheckTuple = tuple[str, int, int]
 SerializedObject = dict[str, Union[str, int, bool, list["SerializedObject"]]]
 
+SQLLexErrorType = Union["SQLLexError", "RsSQLLexerError"]
```
alanmcruickshank: Perhaps I'm missing something. Why do we need to do this, if we have a method below to convert `RsSQLLexerError` into `SQLLexError`? My assumption would be that the python bits would still only get a `SQLLexError` (even if it was a translated `RsSQLLexerError`).

It would avoid changing so many type signatures if we did that too.

sqlfluffrs/README.md
```diff
@@ -0,0 +1,20 @@
...
+This package is automatically handled when installing SQLFluff with the appropriate optional dependencies. Direct installation or standalone usage is not supported.
+
+To install from pip:
+```sh
+pip install sqlfluff[rs]
+```
+
+## Development Status
+
+This is a supplementary component and is not meant to replace or function independently of the main SQLFluff package. For SQL linting, please use the main [SQLFluff](https://github.com/sqlfluff/sqlfluff) package.
```
alanmcruickshank, line 20: 💯

src/sqlfluff/core/parser/lexer.py
keraion: I'll all up for ideas around this one. I think I was fighting mypy on it initially; this is what landed. I agree it is a bit more rust-like. I'm absolutely up for changing it around.

src/sqlfluff/core/parser/lexer.py
keraion: Good point.

sqlfluffrs/src/dialect/ansi/matcher.rs
keraion: I've been up in the air on this one. On one hand, having it generated and in git shows changes to the rust specific things, such as what elements are being captured from the segments. Kinda like the yaml files for the sql fixtures. On the other hand, not having it here would prevent some potentially out of sync issues.

I'm not sold on one way or the other, but it made sense to include them as part of the review.

src/sqlfluff/core/errors.py
keraion: This was me fighting with exceptions and passing them from rust to python. I didn't get a clean solution that I liked and was between limitations of pyo3 and mypy. I may see if this has improved since I initially wrote it.

sqlfluffrs/sqlfluffrs.pyi
keraion: This file was initially created to appease mypy issues. They are the definitions on the rust structs that are available in python. I'm not opposed to docstrings on them.

src/sqlfluff/core/parser/lexer.py
keraion: Rewrote this into `get_lexer_class` which sets `Lexer` to either one of the lexers depending on the import results. Now using `Lexer` is the same as before and mypy is happy.

src/sqlfluff/core/parser/lexer.py
keraion: makes sense. I threw some mocks for unit testing and an integration test. I might need to revisit with your note on dialects once I get to it.

src/sqlfluff/core/errors.py
keraion: Ok, removed without complaints from mypy. This still "exists" in the tests because it's doing comparisons to the types.

src/sqlfluff/core/parser/lexer.py
keraion: I think we might be ok here. I don't think any dialects are creating a new RawSegment based class that is passed to the lexer right now. 

This is the output as of this moment
```
{'binary_operator': <class 'sqlfluff.core.parser.segments.common.BinaryOperatorSegment'>,
 'comment': <class 'sqlfluff.core.parser.segments.common.CommentSegment'>,
 'comparison_operator': <class 'sqlfluff.core.parser.segments.common.ComparisonOperatorSegment'>,
 'dedent': <class 'sqlfluff.core.parser.segments.meta.Dedent'>,
 'end_of_file': <class 'sqlfluff.core.parser.segments.meta.EndOfFile'>,
 'identifier': <class 'sqlfluff.core.parser.segments.common.IdentifierSegment'>,
 'indent': <class 'sqlfluff.core.parser.segments.meta.Indent'>,
 'keyword': <class 'sqlfluff.core.parser.segments.keyword.KeywordSegment'>,
 'literal': <class 'sqlfluff.core.parser.segments.common.LiteralSegment'>,
 'meta': <class 'sqlfluff.core.parser.segments.meta.MetaSegment'>,
 'newline': <class 'sqlfluff.core.parser.segments.common.NewlineSegment'>,
 'placeholder': <class 'sqlfluff.core.parser.segments.meta.TemplateSegment'>,
 'raw': <class 'sqlfluff.core.parser.segments.common.CodeSegment'>,
 'symbol': <class 'sqlfluff.core.parser.segments.common.SymbolSegment'>,
 'template_loop': <class 'sqlfluff.core.parser.segments.meta.TemplateLoop'>,
 'unlexable': <class 'sqlfluff.core.parser.segments.common.UnlexableSegment'>,
 'whitespace': <class 'sqlfluff.core.parser.segments.common.WhitespaceSegment'>,
 'word': <class 'sqlfluff.core.parser.segments.common.WordSegment'>}
```

alanmcruickshank: interesting! Not what I expected - but I checked and you're right 🤔 . That's lucky!

sqlfluffrs/src/dialect/ansi/matcher.rs
alanmcruickshank: That's fair. I think you've sold me - let's keep them committed for now, at least while we're rolling out.

alanmcruickshank:
@keraion - when you're ready, mark this as ready for review and let's get it merged! 🚀

keraion:
@alanmcruickshank I think I'm good with the progress on this one so far. I've made a bit of progress on the parser and I think I've run into most issues that I would have with the lexer at this point. There may be some additional follow up clean up that may happen once the parser is together completely, but that's down the road for now.

alanmcruickshank (approved):
Let's do this 🚀

PR #7198, opened 2025-10-19
SparkSQL: allow accessing nested array fields for RF01

### LINKED ISSUE #7162: RF01 is throw when dot access syntax is found and dialect=sparksql
### Search before asking

### What Happened

RF01 is applied on SparkSQL dialect when there is a dot access syntax. Based on this [issue](https://github.com/sqlfluff/sqlfluff/pull/6400) I am assuming the RF01 to not show for query with dot syntax operator. 

### Expected Behaviour

RF01 should not show up for dot access syntax

### Observed Behaviour

```
L:  3 | P:  17 | RF01 | Reference 'tbl.b.c' refers to table/view not found in
                      | [references.from]
```

### How to reproduce

Create sql file with following content
```
SELECT
    tbl.a AS a_new,
    EXPLODE(tbl.b.c) AS a_b_new
FROM test AS tbl;
```

Run the following command from within the same directory:
`run sqlfluff lint test.sql`

### Dialect

sparksql

### Version

3.4.0

### Configuration

[sqlfluff]
dialect = sparksql

### Code of Conduct

Ruward:

Fixes #7162

The docs state that the rule RF01 is disabled by default for SparkSQL and other dialects. An IF statement for this rule handles these exceptions. This statement enforces len(distinct_targets) == 1 or dialect='bigquery'.

In the case of directly accessing a nested array field in SparkSQL, the len(distinct_targets) is not equal to 1 due to the 'double dot' notation (e.g. tbl.b.c., which returns field c in array b column in table tbl). Because of this rule RF01 is triggered for this SparkSQL case even though this syntax is valid in SparkSQL and the rule should be disabled by default. 

To fix this the 'sparksql' dialect was added to the IF statement filter, next to 'bigquery'. 

SparkSQL will now be excluded in any future testcase for RF01 by default. I have added the same test in a 'force_enable: true' version to make sure this fails on purpose, but I have not been able to test other SparkSQL cases that should pass or fail for this rule, or other dialects that may need to be added to this exclusion logic. 

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/rules/references/RF01.py
+++ b/src/sqlfluff/rules/references/RF01.py
@@ -236,7 +236,10 @@ class Rule_RF01(BaseRule):
         if self._dialect_supports_dot_access(query.dialect):
             # BigQuery supports having multiple aliases in the FROM statement
             # SparkSQL supports directly accessing values in nested array columns
-            if len(distinct_targets) == 1 or query.dialect.name in ["bigquery", "sparksql"]:
+            if len(distinct_targets) == 1 or query.dialect.name in [
+                "bigquery",
+                "sparksql",
+            ]:
                 self.force_enable: bool
                 if self.force_enable:
                     # Backwards compatibility.


WittierDinosaur:
Small linting error

PR #7200, opened 2025-10-20
TSQL: Rework DatatypeSegment

peterbud:
Reworked the `DatatypeSegment` class in the T-SQL dialect to comprehensively support all T-SQL data types following the PostgreSQL dialect pattern and based on the official Microsoft T-SQL data types documentation.

- Comprehensive coverage of all T-SQL data type categories:
  - **Exact numeric types**: `TINYINT`, `SMALLINT`, `INT`, `BIGINT`, `BIT`, `DECIMAL`, `NUMERIC`, `DEC`, `MONEY`, `SMALLMONEY`
  - **Approximate numeric types**: `FLOAT`, `REAL`
  - **Date and time types**: `DATE`, `TIME`, `DATETIME`, `DATETIME2`, `DATETIMEOFFSET`, `SMALLDATETIME`
  - **Character string types**: `CHAR`, `VARCHAR`, `TEXT`, `CHARACTER`, `CHARACTER VARYING`
  - **Unicode character string types**: `NCHAR`, `NVARCHAR`, `NTEXT`, `NATIONAL CHARACTER`, `NATIONAL CHARACTER VARYING`
  - **Binary string types**: `BINARY`, `VARBINARY`, `IMAGE`
  - **Other data types**: `CURSOR`, `SQL_VARIANT`, `TABLE`, `TIMESTAMP`, `ROWVERSION`, `UNIQUEIDENTIFIER`, `XML`, `JSON`, `GEOGRAPHY`, `GEOMETRY`, `HIERARCHYID`, `VECTOR`

Also added tests for these data types.

The change shifts T-SQL's AST structure to match PostgreSQL's convention, where built-in data types (like INT, VARCHAR) now appear as `keyword` segments instead of `data_type_identifier` segments, making the dialect more consistent with other SQL dialects. The `data_type_identifier` segment type is now reserved exclusively for user-defined types and schema-qualified types (like dbo.custom_type or [sys].[sysname]), providing a clearer semantic distinction in the parse tree between built-in and custom data types.

### Pull Request checklist

  - Other.


alanmcruickshank:
@peterbud - With this change, can you check that the new parsing of types is still handled by CP05 rather than CP01. I'm a little worried that this this change, because they're being classed as `keyword` within the wider `data_type` segment, they might get wrongly handled by CP01, or worse - by both!

For context, a raw string (e.g. `"INT"`) in a dialect will get handled as a keyword, unless we explicitly type it as something else. There are examples in the other dialects of how to explicitly type as a `data_type`.

peterbud:
Hi @alanmcruickshank , thanks for the feedback. 

I have actually looked into that before submitting the PR, and I have chacked it once again.

TLDR: The new TSQL `DatatypeSegment` parsing is still correctly handled by CP05, NOT CP01. There is NO risk of double-handling.

**CP01 (Keywords Rule)**

* Has `_exclude_parent_types = ("data_type", "datetime_type_identifier", "primitive_type")`
* When it encounters a keyword segment, it checks the parent type
* If the parent is `data_type`, it excludes the keyword from processing
* This prevents CP01 from linting data type keywords

**CP05 (Data Types Rule)**

* Crawls for `data_type`, `primitive_type`, `datetime_type_identifier`, and `data_type_identifier` segments
* When it finds a `data_type segment`, it iterates through children that are `is_type("raw")`
* Keyword segments ARE raw segments (KeywordSegment → WordSegment → CodeSegment → RawSegment)
* So CP05 correctly processes data type keywords

The new approach uses keyword segments as direct children of `data_type` segments. This is safe because:

* Keywords under `data_type` parent → CP01 excludes them (via `_exclude_parent_types`)
* Raw segments under `data_type` → CP05 includes them (via iteration over children)
* No overlap or double-handling

PR #7226, opened 2025-10-28
Slightly more obvious logging when rust extensions are used

alanmcruickshank:
When using the packaged version of `sqlfluff`, the logging indications that the rust routines are being used are all swallowed, because they're emitted before the loggers are properly initialised.

Rather than logging the availability of the rust routines on _module load_, this moves the logging to two places:
1. On lexer selection (i.e. on `get_lexer_class()`)
2. On lexer use (i.e. on `.lex()`)

This should make it easier for testing users to know whether the rust libraries are being used or not during testing.


keraion (approved):
This makes a lot of sense. LGTM

PR #7229, opened 2025-10-30
refactor: Move Rust parts into their own crates

keraion:

This moves the Rust parts into their own crates. This helps with recompile times when typically the dialects themselves have not changed.

None

### Pull Request checklist

  - Other.


sqlfluffrs/sqlfluffrs_dialects/src/dialect/bigquery/matcher.rs
```diff
@@ -1081,3 +1033,42 @@ pub static BIGQUERY_LEXERS: Lazy<Vec<LexMatcher>> = Lazy::new(|| { vec![
...
+
+            if depth == 0 {
+                return Some(&input[..comment.len()]);
+            }
+        }
+    }
+
+    // If we reach here, the comment wasn't properly closed
+    match dialect {
+        "sqlite" => Some(&input[..comment.len()]),
```
WittierDinosaur: why do we have this in every dialect?

sqlfluffrs/src/lexer.rs
```diff
@@ -1,1335 +0,0 @@
-use std::sync::Arc;
```
WittierDinosaur, line 1: is this delete right?

WittierDinosaur: oh, ignore, missed the addition

sqlfluffrs/sqlfluffrs_dialects/src/dialect/bigquery/matcher.rs
keraion: I do need to fix this but it was to split each dialect into a new crate going forward, I just didn't get there yet. I've started pulling everything in for the parser and I can tell the parallelism we will get will be a nice thing to have. 

At this time I just need to think on how to structurally do this.

WittierDinosaur: okay, but why do they all have sqlite? Is there some quirk of the sqlite implementation that requires this?

WittierDinosaur:
^ conflicts

sqlfluffrs/sqlfluffrs_dialects/src/dialect/bigquery/matcher.rs
keraion: I've moved this to a common location that each dialect can pass itself (ready for future state). Sqlite basically supports unterminated block comments so it can reach the end of the file and it still will lex correctly. This is a helper function that can run much faster than the regex can.

We may want to look at supporting the lexing of unterminated blocks in the future for potential rules, but this matches the current behavior.

WittierDinosaur: ah okay, makes sense

PR #7288, opened 2025-11-16
fix: handle case sensitivity at RF06 for quoted aliases in TSQL

### LINKED ISSUE #6918: AL09 misses some redundant aliases
### Search before asking

### What Happened

File `test.sql`:

```sql
SELECT
    City AS "City",
    Citycode AS "CityCode"
FROM Foo
```

Command:

`sqlfluff lint --dialect=tsql --rules=AL09,RF06,CP02 test.sql`

No errors are found.

This is correct for `"CityCode"` where the alias changes the case, but not correct for `"City"` which repeats the column name with no case change and should therefore be flagged by AL09 as a redundant alias.

Note that setting the `case_sensitive` option of RF06 to `False` does not solve the problem. The output of the above command is then

```
== [test.sql] FAIL
L:   2 | P:  13 | RF06 | Unnecessary quoted identifier "City".
                       | [references.quoting]
L:   3 | P:  17 | RF06 | Unnecessary quoted identifier "CityCode".
                       | [references.quoting]
All Finished!
```

This is correct for `"City"` (and after removing its unnecessary quotes AL09 correctly flags it as a redundant alias), but not correct for `"CityCode"` which should be quoted according to AL09 to make the case change explicit.

I guess this is a conflict between the two rules.

### Expected Behaviour

See above

### Observed Behaviour

See above

### How to reproduce

See above

### Dialect

tsql

### Version

`sqlfluff, version 3.4.0`
`Python 3.11.2`

### Configuration

None

### Code of Conduct

peterbud:

`RF06` (references.quoting rule) with `case_sensitive=False` was removing quotes from all identifiers that could be unquoted, without considering whether those quotes were preserving an intentional case change in alias expressions.

This PR modified `RF06` to add a check when `case_sensitive=False`: if a quoted identifier is in an alias expression, compare the case of the alias with the case of the column it references. If they differ, the quotes are necessary to preserve the case change and should not be removed.

Fixes #6918

### Pull Request checklist

  - Other.

### DIFF
--- a/test/fixtures/rules/std_rule_cases/RF06.yml
+++ b/test/fixtures/rules/std_rule_cases/RF06.yml
@@ -753,3 +753,21 @@ test_fail_alias_default_case_tsql_case_sensitive:
     rules:
       references.quoting:
         case_sensitive: true
+
+test_pass_alias_quoted_column_case_change_tsql_case_insensitive:
+  # When case_sensitive=False and the column reference is quoted,
+  # this test covers line 294 where we strip quotes from the column
+  # identifier to compare with the alias.
+  # Column: "lowercase" (quoted, ignored via ignore_words so not flagged)
+  # Alias: "Mixedcase" (quoted)
+  pass_str: |
+    SELECT
+        "lowercase" AS "Mixedcase"
+    FROM Foo
+  configs:
+    core:
+      dialect: tsql
+    rules:
+      references.quoting:
+        case_sensitive: false
+        ignore_words: lowercase


peterbud:
And this has been open for 2 months, maybe a quick review, @WittierDinosaur ?

test/fixtures/rules/std_rule_cases/RF06.yml
```diff
@@ -648,3 +648,126 @@ test_pass_exasol_password_literals:
...
+        case_sensitive: false
+
+test_fail_alias_no_case_change_tsql_case_insensitive:
+  # When case_sensitive=False, quoted aliases that do NOT change
+  # the case should have quotes removed.
+  # City (naked -> CITY) and "City" (quoted -> City) are different,
+  # but when unquoted both become CITY, so the alias is redundant.
+  fail_str: |
+    SELECT
+        City AS "City"
```
alanmcruickshank, line 675: @peterbud are you sure about this one? Either the comment is inaccurate or I don't understand this test, especially in comparison to the one above.

alanmcruickshank: My understanding is that TSQL is an uppercase dialect unless configured otherwise, so aliasing to `"City"` and to `City` are not equivalent.

alanmcruickshank (commented):
I realise this is a fairly old review but I'd love to get this one merged. Could you take another look at the test cases?

test/fixtures/rules/std_rule_cases/RF06.yml
peterbud: Hi @alanmcruickshank : TSQL is lowercase dialect according to our docs, but can be configured for case sensitive or case-insensitive way. 
It's been a while since I made this PR, but it is meant to fix the case-insensitive mode. The comment indeed might be misleading, as it does not play any role whether the alias is redundant or not, but whether we need quotation around the alias or should we remove the quotation: in this particular case we do NOT need the quotation around the alias, as it is not making any difference. Maybe check 6918, the original issue, there is a good explanation there.

peterbud:
I tend to believe that there is very little added value this PR might bring, so closing.

PR #7294, opened 2025-11-22
feat(parse): export position info in JSON/YAML output

peterbud:

This PR adds a feature to export position information (line numbers and character positions) in JSON/YAML output when using the `--include-meta` flag with the `sqlfluff parse` command.

The long term goal with this change is to enable symbol parsing and outlining of SQL files in the VS Code extension.

**Changes Made**
- Enhanced the `BaseSegment` and `TemplateSegment` classes to include position information in their output when requested.
- Updated the `parse` command to support the new position information feature. Updated `--include-meta` flag documentation to mention position information is included in JSON/YAML output
- Added tests to verify that position information is included when the `--include-meta` flag is used and is accurate for specific segments.

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -857,6 +857,7 @@ class BaseSegment(metaclass=SegmentMetaclass):
         code_only: bool = False,
         show_raw: bool = False,
         include_meta: bool = False,
+        include_position: bool = False,
     ) -> TupleSerialisedSegment:
         """Return a tuple structure from this segment."""
         # works for both base and raw
@@ -876,6 +877,7 @@ class BaseSegment(metaclass=SegmentMetaclass):
                         code_only=code_only,
                         show_raw=show_raw,
                         include_meta=include_meta,
+                        include_position=include_position,
                     )
                     for seg in self.segments
                     if seg.is_code and not seg.is_meta
@@ -889,6 +891,7 @@ class BaseSegment(metaclass=SegmentMetaclass):
                         code_only=code_only,
                         show_raw=show_raw,
                         include_meta=include_meta,
+                        include_position=include_position,
                     )
                     for seg in self.segments
                     if include_meta or not seg.is_meta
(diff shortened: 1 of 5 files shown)


src/sqlfluff/core/parser/segments/base.py
```diff
@@ -824,39 +853,55 @@ def to_tuple(
...
         elif code_only:
-            return (
+            base_tuple = (
                 self.get_type(),
                 tuple(
                     seg.to_tuple(
                         code_only=code_only,
                         show_raw=show_raw,
                         include_meta=include_meta,
+                        include_position=include_position,
```
WittierDinosaur, line 880: wha'ts the reasoning behind passing this as an extra argument if it'll always be set as include_meta in the cli input?

peterbud: I think originally I wanted to have a separate CLI switch to include positions in the output, but then I have realized that the existing `include_meta` setting is probably a better switch for this purpose. 

If you want I can remove that argument, as it has no purpose at the moment.

WittierDinosaur: Yeah, let's remove it

peterbud: @WittierDinosaur ,  well, I forgot, that there was a purpose to keep `include_position` separate: all the The YAML fixtures expect now 2-element tuples (no position info) when `include_meta=True`. 

So if we want to use ONLY the `include_meta`, then ALL the YAML fixtures should be updated which are using 'include_meta' to have them position info. If I'm not mistaken all such YAML files are related to the jinja templater (around a dozen or so).

Is that what you want, or should we still keep the `include_position` in order to avoid changing the fixtures?

peterbud: Ping @WittierDinosaur : can you let me know your POV on the above question:
> Is that what you want, or should we still keep the include_position in order to avoid changing the fixtures?

peterbud: I have tried to modify the test files and fixtures to have all the position info there, but it would be a massive change, dozens of files, with a large "blast radius". I think we are better to keep this `include_position` flag, which enables us to keep the existing test files intact.

I have reverted back to the original PR, and now all tests are passing.

PR #6892, opened 2025-05-13
Pin to Click < 8.3.0

wyardley:
### Brief summary of the change made
While #6676 has a fix for Click 8.2.0, since Click can have breaking changes in minor releases, it is probably a good idea to explicitly pin the Click version in `pyproject.toml` to prevent issues like #6888 from affecting older releases in the future. Then the upper bound can be increased along with any changes needed.

https://github.com/pallets/click/issues/2909#issuecomment-2874389113

Part of #6888
(while this won't retroactively fix issues with older releases, it should help avoid similar problems in the future)

### Are there any other side effects of this change that we should be aware of?
When updating for newer Click versions, the upper bound will need to be updated along with any changes needed to code / tests.

IMO, the project might want to consider at least some basic version constraints in other dependent packages as well.

### Pull Request checklist
- [ ] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #7297, opened 2025-11-24
Fix click version constraint, bumping to 8.3.x

serl:

Fix version constraint for `click`, while bumping it to include all 8.3.x versions.

In #6892, we tried to pin click to keep it under 8.3, but the pin as defined _includes_ 8.3.0.
[That version has been out for about a month](https://pypi.org/project/click/#history) and I don't see anyone complaining, so I'm assuming that we're indeed compatible with 8.3.x by chance.

Now, the pin as it's defined keep click from updating to 8.3 patches, which I guess it wasn't the original plan.

With this PR, I'd avance the pin to include all 8.3.x, but excluding 8.4+.

Don't think so

### Pull Request checklist

  - Other.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ dependencies = [
     "chardet",
     # Click can include breaking changes in minor releases. Make sure to test
     # well before updating upper bound.
-    "click<=8.3.0",
+    "click<8.4.0",
     "colorama>=0.3",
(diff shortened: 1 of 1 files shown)


serl:
@wyardley @WittierDinosaur sorry for the noise, I'm mentioning you since you were involved in #6892 and this is pretty related

Issue #6968, opened 2025-06-24
Documentation: show all possible values for configuration and examples


tiffanygwilson:
### Search before asking

- [x] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Links or command line

- https://docs.sqlfluff.com/en/stable/reference/rules.html
- https://docs.sqlfluff.com/en/stable/configuration/layout.html
- https://docs.sqlfluff.com/en/stable/configuration/default_configuration.html

### Issue/Suggested Improvement

This is similar to, but builds on, Issue #6601. First of all, the work that has already gone into the documentation has been great for someone just getting into SQLFluff, so THANK YOU! That said, especially as a newbie, in general the documentation feels incomplete, and I often find myself going in circles trying to track things down. Here are some examples:
 
## Core rules
The [rules reference](https://docs.sqlfluff.com/en/stable/reference/rules.html#core-rules) states that there are core rules. Which rules are core rules? The link in that section goes to the [general configuration section](https://docs.sqlfluff.com/en/stable/configuration/index.html#config), where nothing mentions core rules explicitly. Taking a guess and going to [Rule Configuration](https://docs.sqlfluff.com/en/stable/configuration/rule_configuration.html). On that page it only says that some rules are core rules, and the link goes right back to the rules reference where we started, so I still don't know what the core rules are. 

![Image](https://github.com/user-attachments/assets/faf93dfc-b631-42b5-be3a-69441378c7b7) 
_Figure 1_

Additionally there is some inconsistency regarding rule groups. The Enabling and Disabling Rules section mentions that `layout` and `capitalisation` are rule groups: 

![Image](https://github.com/user-attachments/assets/e74bfc00-6f35-41cc-9fe5-002210566848)
_Figure 2_

but that is in direct conflict with `core` being the only rule group (Figure 1). I'm guessing this is just an old reference, but it's confusing.
- Rule _bundle_ and rule _group_ are confusing since they are not clearly defined. It seems that every bundle is a group, but there are groups (`all`, `core`) that are not bundles and groups that are elements of bundles (`aliasing.unique`). It should be made explicit in the Rules Reference what these different groupings/labels are.  
[...]

PR #7337, opened 2026-01-02
docs: replace aliases in rule table with core rule specification

annebelleo:

Makes progress on #6968.
Currently, there is no easy way to see which rules are core rules other than checking which groups a certain rule belongs to.  
The rule codes were replaced in sqlfluff version 2.0, but this was released in March 2023, and we are coming up on version 4. Since the rule aliases can still be found under their respective rule names, it would make more sense to visually display which rules belong to core rules. This PR replaces the aliases column in the rule table with a checkmark denoting whether a rule is a core rule.

### Pull Request checklist

  - Other.


alanmcruickshank (approved):
I think on balance this is a good idea. Thanks for proposing.

The old rule codes are all still present further down the docs, but core rules feels like more useful information for the table. 👍 

Thanks for the contribution.

PR #7354, opened 2026-01-12
Optimise rust implementation by using sparse representations

alanmcruickshank:
For `CASEFOLD` and `TRIM_CHARS`, almost _all_ grammars use the default, so rather than generating values for all of them, I think it's more efficient to have a sparse rather than dense representation.

@keraion what do you reckon?

### DIFF
--- a/docs/source/gettingstarted.rst
+++ b/docs/source/gettingstarted.rst
@@ -3,23 +3,22 @@
 Getting Started
 ===============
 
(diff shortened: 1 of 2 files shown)


keraion:
@alanmcruickshank __HUGE__ fan of this. Absolutely makes sense to make this a sparse lookup.

It does appear that the initial Rust PR did introduce a little formatting issue with the docs which is failing pre-commit.

PR #7383, opened 2026-01-20
fix(config): Support in-file configuration directives for `dialect` and `templater` settings.

### LINKED ISSUE #7114: Cannot set core config (eg, dialect or templater) inline
### Search before asking

### What Happened

I have the following .sql file:

```sql
-- sqlfluff:dialect:tsql
-- sqlfluff:templater:placeholder
-- sqlfluff:templater:placeholder:param_style:question_mark
-- sqlfluff:templater:placeholder:1:'2000-01-01'
-- sqlfluff:templater:placeholder:2:'2000-01-01'

select *
from dbo.my_table
where date_column between ? and ?;
```

Running a `sqlfluff parse` (or `fix`, etc) command I get a "no dialect was specified" message.

If I create a `.sqlfluff` file and move the dialect in there:

```ini
[sqlfluff]
dialect = tsql
```

...and remove the `sqlfluff:dialect:tsql` inline configuration from the .sql file and run the `sqlfluff parse` command again I get a "found unparsable section" message.

If I then move the `templater = placeholder` config from the .sql file to the .sqlfluff file and run the `sqlfluff parse` command, things work as expected.

This leads me to think that core config items (usually found under the [sqlfluff] section of a .sqlfluff file) are not being parsed and set properly.

### Expected Behaviour

It should be possible to parse/fix/etc the .sql file with inline configuration without the need for a .sqlfluff file.

### Observed Behaviour

The inline configuration would only work once the "core" configuration items were moved from inline to a .sqlfluff file.

### How to reproduce

Follow the steps in the What Happened section.

### Dialect

tsql

### Version

sqlfluff, version 3.4.2
Python 3.11.9

### Configuration

See What Happened.

### Code of Conduct

peterbud:

This PR fixes SqlFluff to properly respect in-file configuration directives (`-- sqlfluff:...`) for both `dialect` and `templater` settings.

Fixes #7114 

#### Problem

Previously, in-file config comments like:
```sql
-- sqlfluff:dialect:tsql
-- sqlfluff:templater:placeholder
```

were being ignored because:

1. **Dialect verification** happened during config initialization, before in-file directives were processed
2. **Templater selection** was fixed at linter initialization and couldn't be changed per-file

This resulted in parse errors when files used dialect-specific syntax or required different templaters.

#### Changes Made

* Fix Dialect Configuration (`src/sqlfluff/core/linter/linter.py`): The config is created before `process_raw_file_for_config()` is called. By passing `require_dialect=False`, we defer dialect verification until later (it still happens in `render_string()`), allowing in-file dialect directives to be processed first.

* Fix Templater Configuration (`src/sqlfluff/core/linter/linter.py`): Each file can now use its own templater based on its config, rather than being locked to the linter-level templater. I hope that's in line with other SqlFluff principles.

* Added tests

### Pull Request checklist

  - Other.


src/sqlfluff/core/linter/linter.py
```diff
@@ -868,22 +871,13 @@ def render_string(
...
         config.verify_dialect_specified()
-        if not config.get("templater_obj") == self.templater:
-            linter_logger.warning(
-                f"Attempt to set templater to {config.get('templater_obj').name} "
-                f"failed. Using {self.templater.name} templater. Templater cannot "
-                "be set in a .sqlfluff file in a subdirectory of the current "
-                "working directory. It can be set in a .sqlfluff in the current "
-                "working directory. See Nesting section of the docs for more "
-                "details."
-            )
```
alanmcruickshank, line 946: @peterbud - I don't think it's necessarily safe to remove this restriction.

We introduced it because people were getting odd effects with the dbt templater if some of the project used it and some of it didn't. Have you done any testing on that scenario?

peterbud: @alanmcruickshank thanks for the feedback. 

I think there are some nuances here:

Two Different Scenarios:
1. Subdirectory `.sqlfluff` files - Original restriction's concern

    * Affects multiple files implicitly
    * Can cause project-wide inconsistencies
    * This is where dbt issues likely occur

2. In-file `-- sqlfluff:templater:X` comments - What we enabled

    * Affects only that specific file explicitly
    * User has direct, per-file control
    * Less likely to cause project-wide issues
  
Potential Issues:
* If file A uses dbt templater and references file B, but file B switches to placeholder templater via in-file config, dbt might not process B correctly
* The `sequence_files()` call would still use the linter-level dbt templater, but `process_with_variants()` would use placeholder

Safer Approaches:
* **Add a warning for dbt specifically**: Allow templater changes but warn when switching away from dbt
* **Only allow "safe" templater changes**: Allow switching between jinja/placeholder/python, but not to/from dbt
* **Distinguish config sources**: Keep restriction for subdirectory configs, but allow in-file comments (with warning)
* **Revert the change**: Go back to the original behavior

What would be your preference? LMK!

peterbud: ping @alanmcruickshank

PR #7393, opened 2026-01-23
tsql alter table rebuild and drop without constraint keyword

### LINKED ISSUE #7392: [tsql] sqlfluff fails to parse alter table rebuild and drop without constraint keyword
### Search before asking

### What Happened

Parser fails to to parse the following:

ALTER TABLE dbo.SomeTable DROP df_SomeConstraint;
ALTER TABLE dbo.SomeTable REBUILD;

### Expected Behaviour

Parser should handle ALTER TABLE syntax above for REBUILD and DROP options.

### Observed Behaviour

SQLFluff parser error (PRS)

### How to reproduce

Parse the above statements in the tsql dialect.

### Dialect

tsql

### Version

Version 4.0.0

### Configuration

[sqlfluff]
dialect = tsql

### Code of Conduct

dpurfield:

fixes #7392

Added support for the tsql dialect for the following statements:
- ALTER TABLE dbo.SomeTable DROP SomeContraint;
- ALTER TABLE dbo.SomeTable REBUILD;
 
The rebuild syntax has  many optional parts including: DATA_COMPRESSION, XML_COMPRESSION and/or PARTITION options as part of this check in.

I noticed and corrected a couple of missing commas in the temporal_table.sql file.

The only side effect is the addition of "on_partitions_clause" in the existing create table and alter index.  It was requested by the AI to reuse the "ON PARTITIONS" segments between the three statements.

### Pull Request checklist

  - Other.

### DIFF
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/table_driven/anynumberof.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/table_driven/anynumberof.rs
@@ -2,6 +2,7 @@ use crate::parser::{
     table_driven::frame::{TableFrameResult, TableParseFrame, TableParseFrameStack},
     FrameContext, FrameState, MatchResult, ParseError, Parser,
 };
+#[cfg(feature = "verbose-debug")]
 use crate::vdebug;
 use smallvec::SmallVec;
 use sqlfluffrs_types::GrammarId;
@@ -21,53 +22,51 @@ impl Parser<'_> {
     ) -> Result<TableFrameResult, ParseError> {
         self.pos = frame.pos;
         let grammar_id = frame.grammar_id;
-
-        #[cfg(feature = "verbose-debug")]
-        let child_count = self.grammar_ctx.inst(grammar_id).child_count;
         let reset_terminators = self.grammar_ctx.inst(grammar_id).flags.reset_terminators();
         let parse_mode = self.grammar_ctx.inst(grammar_id).parse_mode;
         let start_pos = frame.pos;
-
-        vdebug!(
-            "AnyNumberOf[table] Initial: frame_id={}, pos={}, grammar_id={}, children={}",
-            frame.frame_id,
-            start_pos,
-            grammar_id.0,
-            child_count
-        );
-
-        // Extra debug: build a readable grammar name (works for Ref/String/Regex variants)
-        #[cfg(feature = "verbose-debug")]
-        let grammar_name = self.grammar_ctx.grammar_id_name(grammar_id);
+        let (_min_times, _max_times, _max_times_per_element, has_exclude) =
+            self.grammar_ctx.anynumberof_config(grammar_id);
 
         #[cfg(feature = "verbose-debug")]
-        if start_pos < self.tokens.len() {
-            let tok = &self.tokens[start_pos];
+        {
+            let child_count = self.grammar_ctx.inst(grammar_id).child_count;
+
             vdebug!(
+                "AnyNumberOf[table] Initial: frame_id={}, pos={}, grammar_id={}, children={}",
+                frame.frame_id,
+                start_pos,
+                grammar_id.0,
+                child_count
+            );
+
+            // Extra debug: build a readable grammar name (works for Ref/String/Regex variants)
+            let grammar_name = self.grammar_ctx.grammar_id_name(grammar_id);
+
+            if start_pos < self.tokens.len() {
+                let tok = &self.tokens[start_pos];
+                vdebug!(
                 "AnyNumberOf[table] Initial: grammar='{}' start_token='{}' token_type='{}' start_pos={}",
                 grammar_name,
                 tok.raw,
                 tok.get_type(),
                 start_pos
             );
-        } else {
+            } else {
+                vdebug!(
+                    "AnyNumberOf[table] Initial: grammar='{}' start_token=<EOF>",
+                    grammar_name
+                );
+            }
+
             vdebug!(
-                "AnyNumberOf[table] Initial: grammar='{}' start_token=<EOF>",
-                grammar_name
+                "AnyNumberOf[table]: min_times={}, max_times={:?}, max_times_per_element={:?}, has_exclude={}",
+                _min_times,
+                _max_times,
+                _max_times_per_element,
+                has_exclude
             );
         }
-
-        // Get config from aux_data
-        let (_min_times, _max_times, _max_times_per_element, has_exclude) =
-            self.grammar_ctx.anynumberof_config(grammar_id);
-
-        vdebug!(
-            "AnyNumberOf[table]: min_times={}, max_times={:?}, max_times_per_element={:?}, has_exclude={}",
-            _min_times,
-            _max_times,
-            _max_times_per_element,
-            has_exclude
-        );
 
         // Check exclude grammar if present
         if has_exclude {
@@ -78,10 +77,7 @@ impl Parser<'_> {
                 {
                     if !exclude_result.is_empty() {
                         vdebug!("AnyNumberOf[table]: Exclude grammar matched, returning Empty");
-                        stack.results.insert(
-                            frame.frame_id,
-                            (Arc::new(MatchResult::empty_at(start_pos)), start_pos, None),
-                        );
+                        stack.insert_empty_result(frame.frame_id, start_pos);
                         return Ok(TableFrameResult::Done);
                     }
                 }
@@ -92,29 +88,27 @@ impl Parser<'_> {
         // Get all element children (excludes exclude grammar via element_children)
         let element_ids: Vec<GrammarId> = self.grammar_ctx.element_children(grammar_id).collect();
         let pruned_children = self.prune_options_table_driven(&element_ids);
-        // Debug element names for easier tracing
         #[cfg(feature = "verbose-debug")]
-        let element_names: Vec<String> = pruned_children
-            .iter()
-            .map(|gid| self.grammar_ctx.grammar_id_name(*gid))
-            .collect();
-        vdebug!(
-            "AnyNumberOf[table] element_ids_count={} names={:?}",
-            pruned_children.len(),
-            element_names
-        );
-        // (will log terminators after they are combined below)
+        {
+            // Debug element names for easier tracing
+            let element_names: Vec<String> = pruned_children
+                .iter()
+                .map(|gid| self.grammar_ctx.grammar_id_name(*gid))
+                .collect();
+            vdebug!(
+                "AnyNumberOf[table] element_ids_count={} names={:?}",
+                pruned_children.len(),
+                element_names
+            );
+        }
+
         if pruned_children.is_empty() {
             vdebug!("AnyNumberOf[table]: No elements to match after filtering");
-            stack.results.insert(
-                frame.frame_id,
-                (Arc::new(MatchResult::empty_at(start_pos)), start_pos, None),
-            );
+            stack.insert_empty_result(frame.frame_id, start_pos);
             return Ok(TableFrameResult::Done);
         }
 
         // Initialize option counter for max_times_per_element tracking
-        // TODO: actually prune element_ids
         let pruned_children_count = pruned_children.len();
         let first_element = pruned_children[0];
         let option_counter: hashbrown::HashMap<u64, usize> =
@@ -122,25 +116,26 @@ impl Parser<'_> {
 
         // Combine terminators
         let local_terminators: Vec<GrammarId> = self.grammar_ctx.terminators(grammar_id).collect();
-        let all_terminators = crate::parser::core::Parser::combine_terminators_table_driven(
+        let all_terminators = Parser::combine_terminators_table_driven(
             &local_terminators,
             parent_terminators,
             reset_terminators,
         );
-        let grammar_parse_mode = parse_mode;
 
-        // Debug: log pruned_children ids and the combined terminators
         #[cfg(feature = "verbose-debug")]
-        let term_names: Vec<String> = all_terminators
-            .iter()
-            .map(|gid| self.grammar_ctx.grammar_id_name(*gid))
-            .collect();
-        vdebug!(
-            "AnyNumberOf[table] pruned_children_ids={:?} all_terminators_count={} names={:?}",
-            pruned_children,
-            all_terminators.len(),
-            term_names
-        );
+        {
+            // Debug: log pruned_children ids and the combined terminators
+            let term_names: Vec<String> = all_terminators
+                .iter()
+                .map(|gid| self.grammar_ctx.grammar_id_name(*gid))
+                .collect();
+            vdebug!(
+                "AnyNumberOf[table] pruned_children_ids={:?} all_terminators_count={} names={:?}",
+                pruned_children,
+                all_terminators.len(),
+                term_names
+            );
+        }
 
         // Calculate max_idx using terminators only (match Python behavior)
         // Python's AnyNumberOf uses a simple trim-to-terminator rather than
@@ -151,7 +146,7 @@ impl Parser<'_> {
         let max_idx = self.calculate_max_idx_table_driven(
             start_pos,
             &all_terminators,
-            grammar_parse_mode,
+            parse_mode,
             frame.parent_max_idx,
         )?;
 
@@ -163,7 +158,7 @@ impl Parser<'_> {
             grammar_id.0,
             start_pos,
             max_idx,
-            grammar_parse_mode,
+            parse_mode,
             frame.parent_max_idx
         );
 
@@ -173,7 +168,7 @@ impl Parser<'_> {
         };
 
         // Store context with max_times config and pruned element list
-        frame.context = crate::parser::FrameContext::AnyNumberOfTableDriven {
+        frame.context = FrameContext::AnyNumberOfTableDriven {
             grammar_id,
             pruned_children,
             count: 0,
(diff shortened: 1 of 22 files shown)


test/fixtures/dialects/tsql/alter_table.yml
```diff
@@ -3,7 +3,11 @@
 # computed by SQLFluff when running the tests. Please run
 # `python test/generate_parse_fixture_yml.py`  to generate them after adding or
 # altering SQL files.
+<<<<<<< HEAD
+_hash: 486fa3596ddb907518e076390ad52ec74660807067e3c1c74969ee489e50cc5c
+=======
 _hash: 5eda5a07300cd682fb037f8410d65c6540a88b15c1dfe4ac51b2bf40dc257444
+>>>>>>> 0fac6efac83994f8da9bd66a69b7a9cddcaad3e3
```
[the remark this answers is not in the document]
dpurfield: Refreshed yaml after merge to address

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -4496,6 +4563,46 @@ class AlterTableStatementSegment(BaseSegment):
...
+                                Bracketed(
+                                    Delimited(
+                                        _single_partition_rebuild_table_option,
+                                    ),
+                                ),
+                                optional=True,
+                            ),
+                        ),
+                        optional=True,
+                    ),
```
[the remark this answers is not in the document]
dpurfield: This is how the ALTER INDEX handled the rebuild.  It needs to handle three cases:
REBUILD
REBUILD PARTITION=ALL
REBUILD PARTITION=`<number>`

Should I add an empty Sequence to replace the optional in the OneOf?

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -4429,6 +4429,70 @@ class AlterTableStatementSegment(BaseSegment):
...
+                Ref("NumericLiteralSegment"),
+            ),
+            Sequence(
+                "TO",
+                Ref("NumericLiteralSegment"),
+                optional=True,
+            ),
+        ),
+        optional=True,
+    )
```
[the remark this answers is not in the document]
dpurfield: There where already 2 others one for CREATE TABLE and another for ALTER INDEX.   Should I just make a OnPartitionsSegment and update all three to use the new segment?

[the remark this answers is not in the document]
dpurfield: Added OnPartitionsSegment like suggested replacing the three copies of _on_partitions

src/sqlfluff/dialects/dialect_tsql.py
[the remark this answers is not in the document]
dpurfield: An empty sequence is not valid but was able to remove the optional on the first of the one of items.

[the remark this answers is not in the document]
dpurfield: @greptileai

dpurfield:
@greptileai

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -4496,6 +4529,43 @@ class AlterTableStatementSegment(BaseSegment):
...
+                                Bracketed(
+                                    Delimited(
+                                        _single_partition_rebuild_table_option,
+                                    ),
+                                ),
+                                optional=True,
+                            ),
+                        ),
+                        optional=True,
+                    ),
```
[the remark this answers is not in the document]
dpurfield: This is incorrect.   My test cases fail without this additional optional.

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -1577,6 +1577,25 @@ class CursorDefinitionSegment(BaseSegment):
...
+            Delimited(
+                Ref("NumericLiteralSegment"),
+            ),
+            Sequence(
+                "TO",
+                Ref("NumericLiteralSegment"),
+                optional=True,
+            ),
+        ),
+    )
```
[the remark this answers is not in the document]
dpurfield: Resolved.

dpurfield:
@greptileai

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -4496,6 +4509,43 @@ class AlterTableStatementSegment(BaseSegment):
                     OneOf("AS", "TO", optional=True),
                     Ref("TableReferenceSegment"),
                 ),
+                Sequence(
+                    "REBUILD",
+                    OneOf(
+                        Sequence(
+                            "PARTITION",
+                            Ref("EqualsSegment"),
+                            "ALL",
```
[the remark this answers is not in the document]
dpurfield: @greptileai I believe the following analysis is incorrect:
Issues found:

The REBUILD grammar structure doesn't fully match the ALTER INDEX pattern. The first branch requires PARTITION=ALL (line 4516-4518) instead of making it optional, which prevents valid syntax like REBUILD WITH (DATA_COMPRESSION=PAGE ON PARTITIONS (1,2)) from parsing correctly. This edge case should work but currently routes through the second branch which uses _single_partition_rebuild_table_option that doesn't support ON PARTITIONS clauses.


Only when PARTITION=ALL is the on_partitions_caluse (ON PARITIONS (1,2)) valid.   When 2nd option PARTITION=1 it still allows for DATA_COMPRESSION and XML_COMPRESSION options but doesn't allow for another ON PARITIONS(1,2).

dpurfield:
@greptileai

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -1577,6 +1577,28 @@ class CursorDefinitionSegment(BaseSegment):
...
+    https://docs.microsoft.com/en-us/sql/t-sql/statements/create-index-transact-sql
+    """
+
+    type = "on_partitions_clause"
+    match_grammar = Sequence(
+        "ON",
+        "PARTITIONS",
+        Bracketed(
+            Delimited(
+                OneOf(
```
peterbud: Instead of `OneOf` `num` | `num TO num`
I'd go with `Sequence` `num` `Sequence ( TO, num, Optional=True)`

I know it was in the original, but maybe we can use this opportunity to make it a bit better

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -4429,6 +4393,52 @@ class AlterTableStatementSegment(BaseSegment):
     TODO: Flesh out TSQL-specific functionality
     """
 
+    _rebuild_table_option = AnyNumberOf(
```
peterbud: Why `AnyNumberOf`? As I read in the spec `rebuild_option` can have a `DATA_COMPRESSION` _or_ an `XML_COMPRESSION` part, and only ONCE.
This grammar enables both at he same time as I ebliev it should not, and also enables repeating them, which is I believe not in the specs.

Also, there are a bunch of other rebuild options here: https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-table-index-option-transact-sql?view=sql-server-ver17 - should we take the time to support them as well?

The 'DATA_COMPRESSION' and 'XML_COMPRESSION' is just a typical one and they are used in Synapse Analytics, but mainstream SQL Servers support more rebuild options.

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -4496,6 +4509,43 @@ class AlterTableStatementSegment(BaseSegment):
...
+                                ),
+                                optional=True,
+                            ),
+                        ),
+                        Sequence(
+                            Sequence(
+                                "PARTITION",
+                                Ref("EqualsSegment"),
+                                Ref("NumericLiteralSegment"),
+                                optional=True,
```
peterbud: Why would be this optional? I think that part is mandatory for single_partition_rebuild_option.

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -4496,6 +4509,43 @@ class AlterTableStatementSegment(BaseSegment):
                     OneOf("AS", "TO", optional=True),
                     Ref("TableReferenceSegment"),
                 ),
+                Sequence(
+                    "REBUILD",
+                    OneOf(
+                        Sequence(
+                            "PARTITION",
```
peterbud: In the specs 'PARTITION = ALL' seems to be optional

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -4429,6 +4393,52 @@ class AlterTableStatementSegment(BaseSegment):
...
+            Ref("EqualsSegment"),
+            OneOf(
+                "ON",
+                "OFF",
+            ),
+            Ref("OnPartitionsSegment", optional=True),
+        ),
+    )
+
+    _single_partition_rebuild_table_option = AnyNumberOf(
```
peterbud: Same here, I don't think `AnyNymberOf` is needed

src/sqlfluff/dialects/dialect_tsql.py
dpurfield: Initially I added a new one the way you are recommending it.  But after the greptileai pointed out the original (duplicate), I thought it was better so kept that version.   Just curious but what is the reasoning you prefer the later version?

src/sqlfluff/dialects/dialect_tsql.py
dpurfield: I originaly started by looking at the code for ALTER INDEX and it used AnyNumberOf.  So since I'm still learning seemed like the logical play to start.   The spec does allow for either one or both.   And I have examples with both, so that rules out using OneOf.   I had not thought of AnyNumberOf allowing multiple of the same sections which would not be valid.  So thinking the correct method should be sequence with both parts marked optional.

I don't mean to sound selfish but in my day job we use SQL 2019/2022 and we're switching over to supporting 2022/2025.  I'm hoping to switch our team over from another tool to SQLFluff.  And I actually have a long list of issues we've ran into.  We have a really large and old code base.  So just picked this one to try and keep it small knowing I would learn a lot my first PR.   Which I am.  So was focusing on the lesser syntax that we needed to get my feet wet and learn from.   Once I'm up to speed I have no problem on updating to support all syntax, but didn't want to take on something that big out of the gate.

src/sqlfluff/dialects/dialect_tsql.py
dpurfield: Actually it is optional any of the following are valid:
ALTER TABLE [TestTable] REBUILD;
ALTER TABLE [TestTable] REBUILD PARTITION=ALL;
ALTER TABLE [TestTable] REBUILD PARTITION=1;
ALTER TABLE [TestTable] REBUILD WITH (DATA_COMPRESSION=PAGE);
ALTER TABLE [TestTable] REBUILD PARTITION=1 WITH (DATA_COMPRESSION=ROW);
ALTER TABLE [TestTable] REBUILD PARTITION=ALL WITH (
  XML_COMPRESSION = ON ON PARTITIONS (2),
  DATA_COMPRESSION=COLUMNSTORE ON PARTITIONS (1, 3, 5 TO 6)
  );

Notice the 4th one has no mention of either of the two PARTITION options but still uses the single_parition_rebuild_option of "WITH (DATA_COMPRESSION=PAGE)". This is actually the most common as most tables are not partitioned.

src/sqlfluff/dialects/dialect_tsql.py
dpurfield: This was a bit confusing for me figuring out what the best option is. 
With "PARTITION=ALL" the <rebuild_option> is optional which includes the optional OnPartitionSegments, but this is the only way you can use the OnPartitionsSegments.
With both "PARTITION=num" or no PARTITION section you can use <single_partition_rebuild__option> which is the same as <rebuild_option> but without the ability to use the OnParitionSegments.

So I made PARTITION=ALL as not optional as only this one allows the optional <rebuild_option>.   Otherwise you go to the second path of OneOf where both the PARTITION=<num> and the more limited <single_parition_rebuild_option> are made optional.

src/sqlfluff/dialects/dialect_tsql.py
dpurfield: Agreed.   From above, I now think the correct pattern is to have a sequence of 2 optional sequences.

src/sqlfluff/dialects/dialect_tsql.py
dpurfield: I tried to go back to the original Delimited(num), Sequence (TO, num, optional=True) but there was an issue that was never discovered.  Because the later part was out of the Delimited it only supported one TO and it had to be last, which eliminates a lot of valid combinations.   I added a test case to cover this.   Was able to use this approach but it does require an additional Sequence for it to handle all valid cases.

src/sqlfluff/dialects/dialect_tsql.py
peterbud: I see you have switched from 'AnyNumberOf` to `Sequence` with each of them optional.

I don't think that's the proper grammar, yous should use IMHO 'OneOf`.

dpurfield: I can't use OneOf as they can both be used:

ALTER TABLE [TestTable] REBUILD PARTITION=ALL WITH (
  XML_COMPRESSION = ON ON PARTITIONS (2),
  DATA_COMPRESSION=COLUMNSTORE ON PARTITIONS (1, 3, 5 TO 6, 8, 20 TO 40)
  );

dpurfield: I was just retested what is actually valid syntax on SQL 2022 and this brings up a bit of a conundrum.   The documentation is actually not 100% correct.  

Both PARTITION=ALL and PARITION=num are using _rebuild_table_option (with ON PARTITONS (), ....)
  and they actually do allow duplicates.
Only REBUILD without "PARITION=" is using single_parition_rebuild_option

As the following are are valid and run without exception, even though the 2nd one doesn't make much sense:

ALTER TABLE dbo.test REBUILD PARTITION=ALL WITH (
  DATA_COMPRESSION=NONE ON PARTITIONS (1, 2),
  DATA_COMPRESSION=ROW ON PARTITIONS(3),
  XML_COMPREESSION=ON 
);

ALTER TABLE dbo.test REBUILD PARTITION=1 WITH (
  DATA_COMPRESSION=NONE ON PARTITIONS (1, 2),
  DATA_COMPRESSION=ROW ON PARTITIONS(3),
  XML_COMPREESSION=ON 
);

But the following is not valid and SQL Server throws an exception:
ALTER TABLE dbo.test REBUILD WITH (
  DATA_COMPRESSION=NONE ON PARTITIONS (1, 2),
  DATA_COMPRESSION=ROW ON PARTITIONS(3),
  XML_COMPREESSION=ON 
);
The exception states  the "ON PARTITIONS" is only valid with PARTITION=ALL which matches the documentation, but its not what it is actually excepting.

The following is still valid:
ALTER TABLE dbo.test REBUILD WITH (
  DATA_COMPRESSION=ROW,
  XML_COMPREESSION=ON 
);

So I need to go back to AnyNumberOf for _rebuild_table_option.
And keep Sequence(optional=True, optional=True) for _single_partition_rebuild_table_option

But the question is should I follow the documentation and restrict PARTITION=num to single_parition_rebuild_table_option like the documentation states, or go by what the server is actual allowing?
Note I'm testing SQL 2022.

dpurfield:
> Thanks for the updates. It's getting better.
> 
> Still some additional changes needed I believe:
> 
> * `_rebuild_table_option` and `_single_partition_rebuild_table_option` are now identical. I have tried to see what's the difference in the specs, but it's rather vague. I think there are some options which can be applied only to a single partition, and other which can be applied to many, but I believe we should not re-implement the whole logic.
> 
> So let's do the following:
> 
> * keep only the `_rebuild_table_option`
> * use 'OneOf`at the`_rebuild_table_option`
> * adjust the `REBUILD to use this option only.
> * unify the `PARTITION=ALL` and `PARTITION=<num>` branches to a single one

_rebuild_table_option and _single_partition_rebuild_table_option are NOT identical.
The difleenance is _rebuild_tale_options contains the optional "ON PARTITIONS" segment and allows for AnyNumber.
Where _single_partition_table_option doesn't allow for the "ON PARITIONS" segment and allows for each optional section (DATA_COMPRESION and/or XML_COMPRESION) only once.

But as stated above the documentation is not 100%.  So should I follow the documentation or what the server allows?

dpurfield:
I can't help but wonder if we're saying we're making the parser for complicated then we should.   It's not the actual linter or rules.  So should it really only be able to parse 100% valid SQL?   Shouldn't it be able to parse SQL that is close to valid and let the Server and/or rules validate the minutiae?   

How about this:
  - keep only the _reubild_table_option
  - use AnyNumberOf instead of OneOf or Sequence (Required to support example below)
  - adjust the REBUILD to use only this option
  - unify the `PARITION=ALL` and `PARTITION=<num>`
  - add some additional test cases so all valid syntax options are clear
  
Complex valid SQL:
ALTER TABLE [TestTable] REBUILD PARTITION=ALL WITH (
  DATA_COMPRESSION = NONE ON PARITIONS (4),
  DATA_COMPRESSION = COLUMNSTORE ON PARTITIONS (1, 5 TO 7, 10, 20 TO 40),
  XML_COMPRESSION = ON
);

peterbud:
Indeed, they are different - you are right, sorry. Nevertheless it might be better to unify them and just use ` Ref("OnPartitionsSegment", optional=True)` for the sake of simplicity.

I'm still convinced the `Sequence` or `AnyNumberOf` at the beginning of `_rebuild_table_option` and `_single_partition_rebuild_table_option` is definitely incorrect and not needed.

Your grammar ensures the ability to handle _multiple_ of such grammars, delimited by commas at 4527-4533:
```python
Sequence(
  "WITH",
  Bracketed(
    Delimited(
      _rebuild_table_option,
    )
  ),
  optional=True,
),
```
I have tested with `OneOf` and it just works fine, even with your most complex example above.

So my suggestion is:
* keep `_rebuild_table_option`, use the optional `Ref("OnPartitionsSegment")' to handle with and without partitions
* Use 'OneOf'
* While you are at it, the individual Sequences don't need `optional=True`
* adjust the REBUILD to use only this option
* unify the `PARTITON=ALL` and `PARTITION=<num>`

dpurfield:
Sorry, you are totally correct OneOf works perfectly fine.   It was not occurring to me the delimited section was addressing my concern, until you pointed it out.   I was being dense.   Still learning the patterns.   But I won't make that mistake again.  Thank you for being patient.

Issue #7397, opened 2026-01-27
L043 working incorrectly on alternate syntax of CASE/WHEN


tcalik:
### Search before asking

- [x] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

Running sqlfluff on the `CASE _expression_ WHEN _value_ THEN ...` variation of the CASE statement results in simplyfing to a `COALESCE` that skips the comparison in some edge cases

### Expected Behaviour

```sql
CASE status_code
  WHEN 100 THEN true
  ELSE false
END AS is_success
```
should compile to
```sql
COALESCE(status_code=100, false) AS is_success
```

### Observed Behaviour

L043 `structure.simple_case` ignores the _expression_ in the alternate CASE/WHEN form (available in Redshift)
``` sql
CASE expression
  WHEN value THEN result
  ELSE alternative
END
```
The output of `sqlfluff fix` is
```sql
coalesce(value, alternative)
```

Example: 
```sql
CASE status_code
  WHEN 100 THEN true
  ELSE false
END AS is_success
```
compiles to
```sql
COALESCE(100, false) AS is_success
```

### How to reproduce

Create file with
```sql
SELECT 
CASE status_code
  WHEN 100 THEN true
  ELSE false
END AS is_success
FROM a
```
Run
```sh
sqlfluff fix path/to/file
```
Result is
```sql
SELECT COALESCE(100, false) AS is_success
FROM a
```

### Dialect

redshift

### Version

sqlfluff, version 3.5.0
Python 3.11.14

### Configuration

```ini
[sqlfluff]
dialect = redshift
templater = dbt
max_line_length = 120
large_file_skip_byte_limit = 0

exclude_rules = L009, L014

[sqlfluff:templater:dbt]
project_dir = .
```

### Are you willing to work on and submit a PR to address the issue?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

Tanishq3031881:
Hi @tcalik , I noticed this issue is still open. Are you planning on working on this? If you're busy, I'd be happy to work on this as my first contribution! Doing a data engineering internship right now, this would be perfect.

tcalik:
Hey @Tanishq3031881, I haven't started work on it yet, feel free to contribute!

[...]

PR #7432, opened 2026-02-04
Fix ST02/L043 corruption of Simple CASE statements

Tanishq3031881:
Fixes issue #7397 
Rule ST02  (formerly 
Screenshot 2026-02-04 132230
L043) was incorrectly identifying Simple CASE statements (CASE x WHEN y THEN...) as Searched CASE statements (CASE WHEN x=y THEN...). Then it would attempt to simplify them into COALESCE(y, ...) which resulted in an invalid SQL.

Added a guard clause in the eval function sqlfluff\src\sqlfluff\rules\structure\ST02.py. The rule now checks for any expression segments between the `CASE` keyword and the first `WHEN`. If an expression is found the rule aborts and returns None, preserving the actual SQL.

No this only prevents the rule from triggering the the unsopported case statement, does not affect any other behaviour except thi.

### Pull Request checklist
- [x ] Please confirm you have completed any of the necessary steps below.
-Verified issue on local terminal
-Added proper documentation

  - Other.


peterbud (approved):
Thank you!

Tanishq3031881:
> Thank you!

All good! Was really good practise for my first open source contribution.

PR #7436, opened 2026-02-04
Snowflake: Enhance create dynamic table statement parsing

gmkr0:

Fixes parsing of column comments in Snowflake CREATE DYNAMIC TABLE statements when columns are specified without datatypes.

Previously, the parser failed on syntax like:
```sql
CREATE DYNAMIC TABLE product (
  product_id COMMENT 'product id',
  product_name
)
  TARGET_LAG = '20 minutes'
  WAREHOUSE = mywh
  AS SELECT product_id, product_name FROM staging_table;
```

**Root Cause:** The `COMMENT` keyword is unreserved in Snowflake, causing ambiguity when parsing `identifier COMMENT 'text'` patterns. The parser tried to match `comment` as another identifier in the delimited column list rather than recognizing it as the start of a `CommentClauseSegment`.

**Solution:** Added an explicit pattern `Sequence(Ref("ColumnReferenceSegment"), Ref("CommentClauseSegment"))` before the bare `Ref("SingleIdentifierGrammar")` in the `OneOf` block within `CreateTableStatementSegment`. This makes the grammar unambiguous and matches the pattern used in CREATE VIEW.

This aligns with Snowflake's documentation:
https://docs.snowflake.com/en/sql-reference/sql/create-dynamic-table
(look for column_list section)

None. The fix:
- Uses existing `ColumnReferenceSegment` and `CommentClauseSegment` segments
- Does not affect other CREATE TABLE patterns (with datatypes, AS expressions, etc.)
- All existing CREATE TABLE/VIEW regression tests pass
- The pattern already works in CREATE VIEW and now works consistently in CREATE TABLE/DYNAMIC TABLE

### Pull Request checklist
    - Added test case in `test/fixtures/dialects/snowflake/create_dynamic_table.sql`
    - Generated corresponding YAML in `test/fixtures/dialects/snowflake/create_dynamic_table.yml`
  - The change is straightforward. No documentation needed
- [x] Created GitHub issues for any relevant follow-up/future enhancements if appropriate.
  - None needed - this is a complete fix for the reported issue

**Changed Files:**
- `src/sqlfluff/dialects/dialect_snowflake.py` - Added explicit identifier+COMMENT pattern
- `test/fixtures/dialects/snowflake/create_dynamic_table.sql` - Added test case
- `test/fixtures/dialects/snowflake/create_dynamic_table.yml` - Regenerated fixture

### DIFF
--- a/test/fixtures/dialects/snowflake/create_table.sql
+++ b/test/fixtures/dialects/snowflake/create_table.sql
@@ -336,7 +336,7 @@ CREATE OR REPLACE HYBRID TABLE target_hybrid_table (
     col1 VARCHAR(32) PRIMARY KEY,
     col2 NUMBER(38,0) UNIQUE,
     col3 NUMBER(38,0),
-    INDEX index_col3 (col3)
+    INDEX index_col3 (col3) INCLUDE (col1, col2)
     )
   AS SELECT col1, col2, col3 FROM source_table;
 
--- a/test/fixtures/dialects/snowflake/create_table.yml
+++ b/test/fixtures/dialects/snowflake/create_table.yml
@@ -3,7 +3,7 @@
 # computed by SQLFluff when running the tests. Please run
 # `python test/generate_parse_fixture_yml.py`  to generate them after adding or
 # altering SQL files.
-_hash: 61fd35843693b9255b7accbb714b5777476e2e1b2c095b65ba08efaa81ed379d
+_hash: 5eb9b8fdbd02c3c803bae3ea9de4f8c57a2f9907316a4848ff461dfa6ace58a0
 file:
 - statement:
     create_table_statement:
@@ -2490,13 +2490,22 @@ file:
               - end_bracket: )
       - comma: ','
       - index_properties_segment:
-          keyword: INDEX
-          naked_identifier: index_col3
-          bracketed:
+        - keyword: INDEX
+        - naked_identifier: index_col3
+        - bracketed:
             start_bracket: (
             column_reference:
               naked_identifier: col3
             end_bracket: )
+        - keyword: INCLUDE
+        - bracketed:
+          - start_bracket: (
+          - column_reference:
+              naked_identifier: col1
+          - comma: ','
+          - column_reference:
+              naked_identifier: col2
+          - end_bracket: )
       - end_bracket: )
     - keyword: AS
     - select_statement:


src/sqlfluff/dialects/dialect_snowflake.py
```diff
@@ -4979,6 +4979,10 @@ class CreateTableStatementSegment(ansi.CreateTableStatementSegment):
                             OneOf(
                                 Ref("OutOfLineConstraintPropertiesSegment"),
                                 Ref("ColumnDefinitionSegment"),
+                                Sequence(
```
peterbud: I don't think that is an optimal solution for multiple reasons:
* COMMENT should be part of the ColumnDefinitionSegment
* COMMENT should be optional
* The current grammar would work for `Col1 COMMENT ....`, but I believe it would fail for `Col1 <col_type> COMMENT ....`

test/fixtures/dialects/snowflake/create_dynamic_table.sql
```diff
@@ -44,3 +44,18 @@ CREATE DYNAMIC TABLE product
   REQUIRE USER
   AS
     SELECT product_id, product_name FROM staging_table;
+
```
peterbud: You should add more examples for CREATE TABLE with / without COMMENT to se that it won't break

test/fixtures/dialects/snowflake/create_dynamic_table.sql
```diff
@@ -44,3 +44,18 @@ CREATE DYNAMIC TABLE product
...
+CREATE DYNAMIC TABLE product (
```
peterbud, line 57: add more example with/without col type, with/without comment etc

peterbud (changes requested):
Hi @gmkr0 - thanks for the contribution.

Pls. see the individual comments.
Also there are linting errors which should be fixed.

src/sqlfluff/dialects/dialect_snowflake.py
gmkr0: Thank you for the review! 

Let me clarify the intent here.

This syntax is Snowflake-specific and applies only to `create dynamic table`. The goal is to support a valid Snowflake use case that is currently not expressible: To add a comment to a column without specifying its type 

Regarding the points you raised:
* ColumnDefinitionSegment cannot represent `colum_name comment 'text'` at all. Extending ColumnDefinitionSegment does not seem appropriate, since it originates in the ANSI dialect
* Forms like  `colum_name type comment 'text'` or just `column_name type` are already  handled by existing segments in the OneOf clause, so this PR does not interfere with or replace those parts. That's why definitions without comments, or definitions with both types and comments, continue working as expected

In short, this PR fills a Snowflake-specific gap without changing the existing column definition handling.

I'll address the linting issues and add additional examples as requested. Happy to revisit the structure if you see a cleaner way to model this.

gmkr0: I took a closer look at how ColumnDefinitionSegment is used in th Snowflake dialect. 
The segment appears in various contexts, such as 
* `ALTER TABLE ADD COLUMN [column definition]` 
* `CREATE STORED PROCEDURE .... RETURNS TABLE ([column definition])`. 

In these contexts, allowing a column definition without a data type would be syntactically invalid

test/fixtures/dialects/snowflake/create_dynamic_table.sql
gmkr0: done

src/sqlfluff/dialects/dialect_snowflake.py
gmkr0: @peterbud I have revisited the approach and made some changes.

Instead of adding the new Sequence, I have decided to implement Snowflake-specific `DatatypeIdentifierSegment` with an explicit list of Snowflake-supported data types. This allows the parser to properly disambiguate between:
* `column_name TYPE COMMENT 'text'` - column with type and comment (handled by ColumnDefinitionSegment)
* `column_name COMMENT 'text'` - column with comment only 
* `column_name`

With an explicit list of data type identifiers, the parser can recognize that `COMMENT` is not a data type and therefore `col1 COMMENT 'text'` should not be parsed as a column with type `COMMENT`

Additionally, while working on this I noticed that `INDEX` clauses in `CREATE HYBRID TABLE` were being incorrectly parsed as column definitions. I had to add `OutOfLineIndexPropertiesSegment` to properly handle the INDEX syntax per the [documentation](https://docs.snowflake.com/en/sql-reference/sql/create-hybrid-table#syntax)

Let me know if that makes sense

Issue #7384, opened 2026-01-20
Improving CI pipeline efficiency


peterbud:
### Search before asking

- [x] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### Description

The current CI pipeline is really thorough, but not surprisingly it is slow, a normal CI cycle now takes 40+ minutes.

This issue is meant to serve as a platform to initiate a discussion on how we could improve on this.

I'm not an expert on those pipelines, but here are few items which _might_ help us to make the pipeline execution faster:

- [x] Transition to 'uv' from 'pip' (as we ourselves are using Rust based tool set, that might provide us with performance benefits)
- [x] Dependency caching: `ci-test-python.yml`, `ci-test-dbt.yml`: Add pip/uv cache for faster dependency installation, and/or cache tox environments
- [x] `pre-commit.yml`: Cache pre-commit environments
- [x] `ci-tests.yml`: If calling other workflows sequentially, we can maybe running them in parallel using matrix strategies or separate jobs with no dependencies
- [x] `publish-sqlfluff-docker-image-to-dockerhub.yam`: we can use Docker buildx with GitHub Actions cache
- [ ] Path filtering: Skip irrelevant jobs when only docs/non-code files change
- [ ]  `ci-test-dbt.yml`: Cache dbt project dependencies if applicable
- [x] Reduce Checkout Depth: we can use `fetch-depth: 1`
- [x] `add-to-release-notes.yml` and `ci-pr-comments.yml`: use `ubuntu-slim` runner

Let me know your POV!

### Dialect

N/A

### Are you willing to work on and submit a PR to address the issue?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

keraion:
The biggest increase in CI runtime is the Rust compilation specifically the `sqlfluffrs_dialect` crate (it's roughly 400k LOC).

Right now, each step in the matrix compiles for each version of Python. For example, the `rs-build-windows-x64` job compiles and produces a wheel for each version of python 3.10-3.14, where `sqlfluffrs` is compiled 5 times (each taking around 5 minutes each). 

[...]

PR #7438, opened 2026-02-05
build: Further workflow optimizations

peterbud:

This PR is aiming to add further optimization to the existing workflows.

The following changes were added:

* Using `ubuntu-slim` image when it is enough to tun the workflows
* Splitting `rs-build-linux` job into two part:
  * `rs-build-linux-x86_64`: that is the fastest, and that is needed for the subsequent python test.
  * `rs-build-linux-exotic`: all other linux architectures, which are typically 2-3 minutes slower than the x86_64, so decoupling brings some time benefits. This part is configured to run only on main, and not on every pull request.
* I have disabled `sccache` for the rust builds as it has generated excessive cache entries without too much chance for reusability
* But added `Swatinem/rust-cache@v2` which caches the rust build env, and brings around 20 -30 seconds benefits.

I have experimented with 4 test workers (github runners for public repos are having 4 vCPUs), but -n 4 actually even made tests significantly slower to my great surprise. So I have not changed the two worker config.

I have tested the CI build and the overall runtime is defined by two jobs:
* the length of the `rs-build-linux-x86_64` job: currently with the cache it is around ~6:20-6:40 minutes
* the length of the parser_suite test with coverage job: that is around 8:40-8:50 minutes

Here is a run log: https://github.com/peterbud/sqlfluff/actions/runs/21716386629

If any of these can be accelerated the overall pipeline would be faster. All other jobs are running parallel, and are shorter.

Makes progress on #7384

### Pull Request checklist

  - Other.

### DIFF
--- a/.github/workflows/add-to-release-notes.yml
+++ b/.github/workflows/add-to-release-notes.yml
@@ -8,7 +8,7 @@ on:
       - main
 jobs:
   draft-release:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     if: github.repository == 'sqlfluff/sqlfluff'
     steps:
     - name: Update release notes
--- a/.github/workflows/ci-pr-comments.yml
+++ b/.github/workflows/ci-pr-comments.yml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   comment-on-pr:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     if: >
       github.event.workflow_run.event == 'pull_request'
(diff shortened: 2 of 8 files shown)


peterbud:
@greptileai: see the available github runners here:
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

`ubuntu-slim` does exist. So much that it has even run successfully:
https://github.com/peterbud/sqlfluff/actions/runs/21709640385/job/62609614528

PR #7447, opened 2026-02-06
ST07: Qualify ambiguous columns when converting USING to ON

### LINKED ISSUE #7230: Rule structure.using (ST07) breaks some BigQuery statements
First of all, thanks for this amazing tool! ❤️

While using it, we ran into the following problem (minimal example):

When applying `sqlfluff fix --dialect=bigquery` (version 3.5.0) to the following query

```sql
WITH
table_a AS (
    SELECT
        1 AS field_1,
        "A" AS field_2
),

table_b AS (
    SELECT
        1 AS field_1,
        "X" AS field_3
)

SELECT
    field_1,
    field_2,
    field_3
FROM
    table_a
INNER JOIN
    table_b
    USING
        (field_1)
```

the result is

```sql
WITH
table_a AS (
    SELECT
        1 AS field_1,
        "A" AS field_2
),

table_b AS (
    SELECT
        1 AS field_1,
        "X" AS field_3
)

SELECT
    field_1,
    field_2,
    field_3
FROM
    table_a
INNER JOIN
    table_b
    ON table_a.field_1 = table_b.field_1
```

which is broken, i.e., BigQuery throws the following error:

`Column name field_1 is ambiguous at [15:5]`

### Code of Conduct

nadjichalal:
Fixes #7230

ST07 rule was not qualifying columns from USING clauses in JOINs, leading to ambiguous references in SELECT statements.
When a query uses a JOIN ... USING (col) clause, the referenced column in SELECT is unqualified. If both tables have a column with the same name, this creates ambiguity and can lead to errors or unexpected results.

Example: 
-- Before fix
-- Ambiguous (both tables have "id") 
SELECT id, name 
FROM table_a JOIN table_b USING (id);

-- After fix 
SELECT table_a.id, name 
FROM table_a JOIN table_b USING (id);

Solution: Automatically qualify USING columns in the SELECT clause with the appropriate table alias, preventing ambiguity. 

None. Only affects queries using USING clauses. 
Regular JOINs without USING or other SELECTs are unaffected.

### Pull Request checklist
- Please confirm you have completed any of the necessary steps below.
- [ ] Included test cases to demonstrate any code changes, which may be one or more of the following:
  - [ ] `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - [ ] Other.
- [ ] Added appropriate documentation for the change.

### DIFF
--- a/src/sqlfluff/rules/structure/ST07.py
+++ b/src/sqlfluff/rules/structure/ST07.py
@@ -37,6 +37,17 @@ class Rule_ST07(BaseRule):
        This rule is disabled for ClickHouse as it supports ``USING`` without
        brackets which this rule does not support.
 
+    .. note::
+
+       When ``USING`` is rewritten to ``ON``, any columns referenced
+       unqualified in the ``SELECT`` clause that were previously resolved
+       unambiguously by the ``USING`` clause may become ambiguous. For
+       example, in BigQuery ``SELECT field_1 FROM a JOIN b USING (field_1)``
+       is valid, but after rewriting to ``ON`` the unqualified ``field_1``
+       in ``SELECT`` can raise an ambiguity error in some dialects (e.g.
+       BigQuery). You may need to qualify those column references manually
+       after applying this fix.
+
     **Anti-pattern**
 
     .. code-block:: sql
@@ -142,15 +153,6 @@ class Rule_ST07(BaseRule):
             *[LintFix.delete(seg) for seg in to_delete],
         ]
 
-        # Find and fix unqualified column references in SELECT clause
-        # that match the USING columns (to prevent ambiguity)
-        select_clause = parent_select.get_child("select_clause")
-        if select_clause:
-            column_fixes = _generate_column_qualification_fixes(
-                select_clause, using_columns, table_a.ref_str
-            )
-            fixes.extend(column_fixes)
-
         return LintResult(
             anchor=anchor,
             description=description,
@@ -232,33 +234,3 @@ def _create_col_reference(table_ref: str, column_name: str) -> ColumnReferenceSe
         IdentifierSegment(raw=column_name, type="naked_identifier"),
     )
     return ColumnReferenceSegment(segments=segments, pos_marker=None)
-
-
-def _generate_column_qualification_fixes(
-    select_clause: BaseSegment, using_columns: list[str], table_ref: str
-) -> list[LintFix]:
-
-    fixes: list[LintFix] = []
-    using_columns_upper = [col.upper() for col in using_columns]
-
-    # Find all column_reference segments in the SELECT clause
-    for col_ref in select_clause.recursive_crawl(
-        "column_reference", no_recursive_seg_type="select_statement"
-    ):
-        # Check if this is an unqualified reference (no dot)
-        has_dot = any(seg.is_type("symbol", "dot") for seg in col_ref.segments)
-        if has_dot:
-            # Already qualified, skip
-            continue
-
-        # Get the column name
-        identifiers = [seg for seg in col_ref.segments if seg.is_type("identifier")]
-        col_name = identifiers[0].raw.upper()
-
-        # Check if this column is in the USING list
-        if col_name in using_columns_upper:
-            # Create a qualified column reference
-            qualified_col_ref = _create_col_reference(table_ref, identifiers[0].raw)
-            fixes.append(LintFix.replace(col_ref, [qualified_col_ref]))
-
-    return fixes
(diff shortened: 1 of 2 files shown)


nadjichalal:
Hi @peterbud 
Could you please review this PR ? 
Thank you !

peterbud:
Hi @nadjichalal - thanks for the PR!

Please rebase the branch to the main / merga main branch first. 
Also I think it would be worthwhile to check and resolve the comments from the AI.

nadjichalal:
Hi @peterbud 
Done ! 
Also took into account the bot's recommendations

src/sqlfluff/rules/structure/ST07.py
```diff
@@ -221,3 +232,33 @@ def _create_col_reference(table_ref: str, column_name: str) -> ColumnReferenceSe
...
     )
     return ColumnReferenceSegment(segments=segments, pos_marker=None)
+
+
+def _generate_column_qualification_fixes(
+    select_clause: BaseSegment, using_columns: list[str], table_ref: str
+) -> list[LintFix]:
+
+    fixes: list[LintFix] = []
+    using_columns_upper = [col.upper() for col in using_columns]
```
peterbud: Some dialects are default lower case, some by default upper case (https://docs.sqlfluff.com/en/stable/reference/dialects.html).
Would the `col.upper()` cause problems in other dialects? I'm not sure TBH. Can you add test cases or test them?

peterbud:
Hi @nadjichalal, are you planning to work on this PR, or should we close it?

nadjichalal:
Hi @peterbud, let’s keep it open for now if that’s alright with you. I’ll try to work on it over the next couple of days.

nadjichalal:
Hi @peterbud, 
I've updated this PR with a different approach. 

The previous version auto-qualified unqualified USING columns in the SELECT clause after rewriting USING to ON. While that fixed the ambiguity error, it was overstepping the rule's scope (ST07 is a JOIN rule, not a SELECT rule) and was silently changing query semantics by arbitrarily picking table_a to qualify columns, which isn't always correct.

Regarding the col.upper() casing concern you raised. That code has been removed entirely along with the SELECT modification logic, so the issue is no longer relevant.

The new approach:

- Keeps ST07 focused solely on rewriting USING → ON (no SELECT modifications)
- Adds a docstring note warning users that unqualified USING columns in SELECT may become ambiguous after the rewrite in some dialects (for example BigQuery) and that they should qualify them manually. The note is fairly detailed. I can trim it down if you feel it's too verbose for the terminal output.

This new approach stays within the scope of ST07. This keeps the rule focused and predictable. Happy to hear your thoughts.

If this seems good for you, could you approve the workflow run ? Thank you !

src/sqlfluff/rules/structure/ST07.py
```diff
@@ -121,13 +132,14 @@ def _eval(self, context: RuleContext) -> Optional[LintResult]:
...
         table_a, table_b = table_aliases[:2]
+        using_columns = _extract_cols_from_using(segment, using_anchor)
         edit_segments = [
             KeywordSegment(raw="ON"),
             WhitespaceSegment(raw=" "),
         ] + _generate_join_conditions(
             table_a.ref_str,
             table_b.ref_str,
-            _extract_cols_from_using(segment, using_anchor),
+            using_columns,
```
peterbud, line 142: What is the change here? It's functionally identical. The only difference is extracting `_extract_cols_from_using(segment, using_anchor)` into a local variable `using_columns` before passing it to `_generate_join_conditions`. It's a pure readability refactor — no change in behavior.

nadjichalal:
Hi @peterbud,

You're right that the behavior is unchanged.

While investigating #7230, I initially tried to automatically qualify unqualified columns in the SELECT clause after rewriting 'USING' to 'ON'. However, this turned out to be unsafe because it requires choosing which table to qualify from ('table_a' or 'table_b'), which can change the semantics of the query and is not reliable.

In my view, the needed changes would impact query semantics and make this issue go beyond the scope of what SQLFluff should handle. 

I updated my contribution to instead only add a note to the rule documentation, which should help users by warning them about this behavior so they can manually qualify columns if needed after applying the fix.

An alternative approach would be to skip the rule for certain dialects. The downside is that this introduces dialect-specific behavior. If you feel that adding dialect-specific behavior is acceptable, I can work on skipping this rewrite for the BigQuery dialect.

peterbud:
OK, in this case, let's close this PR - thanks for the efforts nevertheless!

PR #7473, opened 2026-02-10
Recursion check

alanmcruickshank:
Addressing a security vulnerability which has been identified (especially for SQLFluff online) where malicious queries could effectively bring the service down. This adds a cap on recursion within both the python and rust parsers.

### DIFF
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/python.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/python.rs
@@ -480,19 +480,19 @@ pub struct PyParser {
     indent_config: hashbrown::HashMap<&'static str, bool>,
     max_parser_iterations: usize,
     parser_warn_threshold: usize,
-    max_parse_depth: Option<usize>,
+    max_parse_depth: usize,
 }
 
 #[pymethods]
 impl PyParser {
     #[new]
-    #[pyo3(signature = (dialect=None, indent_config=None, max_parser_iterations=None, parser_warn_threshold=None, max_parse_depth=None))]
+    #[pyo3(signature = (dialect=None, indent_config=None, max_parser_iterations=None, parser_warn_threshold=None, max_parse_depth=0))]
     pub fn new(
         dialect: Option<&str>,
         indent_config: Option<HashMap<String, bool>>,
         max_parser_iterations: Option<usize>,
         parser_warn_threshold: Option<usize>,
-        max_parse_depth: Option<usize>,
+        max_parse_depth: usize,
     ) -> PyResult<Self> {
         let dialect = dialect
             .and_then(|d| Dialect::from_str(d).ok())
--- a/sqlfluffrs/sqlfluffrs.pyi
+++ b/sqlfluffrs/sqlfluffrs.pyi
@@ -193,7 +193,7 @@ class RsParser:
         indent_config: Optional[dict[str, bool]] = None,
         max_parser_iterations: Optional[int] = None,
         parser_warn_threshold: Optional[int] = None,
-        max_parse_depth: Optional[int] = None,
+        max_parse_depth: int = 0,
     ): ...
     def parse_match_result_from_tokens(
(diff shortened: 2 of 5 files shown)


alanmcruickshank:
@greptile, I've made quite a few changes since your initial review. Can you review the PR again and update your confidence score. For context, the "core" config section is the main [sqlfluff] section.

sqlfluffrs/sqlfluffrs_parser/src/parser/python.rs
```diff
@@ -480,17 +480,19 @@ pub struct PyParser {
...
 impl PyParser {
     #[new]
-    #[pyo3(signature = (dialect=None, indent_config=None, max_parser_iterations=None, parser_warn_threshold=None))]
+    #[pyo3(signature = (dialect=None, indent_config=None, max_parser_iterations=None, parser_warn_threshold=None, max_parse_depth=None))]
     pub fn new(
         dialect: Option<&str>,
         indent_config: Option<HashMap<String, bool>>,
         max_parser_iterations: Option<usize>,
         parser_warn_threshold: Option<usize>,
+        max_parse_depth: Option<u32>,
```
keraion: This could be defined as a `usize` instead of a `u32` which would make the 
```rust
    max_parse_depth: max_parse_depth.map(|v| v as usize),
```
below on 520 just 
```rust 
    max_parse_depth,
```

.gitignore
```diff
@@ -15,12 +15,14 @@ _build
 dist
 .pytest_cache
 /sqlfluff-*
+.tmp
```
peterbud: Yep. it has been already there, no need a second one

alanmcruickshank:
> We should probably add a test that targets the vulnerability itself to prevent future regressions.

@keraion - I agree in principle, but I'm not sure precisely _how_ aside from watching whether the pytest case gets OOM killed 🤔 . Any good ideas on that front?

keraion:
@alanmcruickshank The best way I can think of is handle it like a related attack, ReDoS. Push a vulnerable input, and the parser should error with the limit being triggered. This way if we introduce something that bypasses the stack limit as an optimization (just as an example) and it doesn't fail with the limit, it fails with oom or a crash, that means we've inadvertently reintroduced the issue.

Obviously we'll need to make sure we have a limit to the runner, maybe like `ulimit` on an Ubuntu runner to strictly hit this.

alanmcruickshank:
> @alanmcruickshank The best way I can think of is handle it like a related attack, ReDoS. Push a vulnerable input, and the parser should error with the limit being triggered. This way if we introduce something that bypasses the stack limit as an optimization (just as an example) and it doesn't fail with the limit, it fails with oom or a crash, that means we've inadvertently reintroduced the issue.
> 
> Obviously we'll need to make sure we have a limit to the runner, maybe like `ulimit` on an Ubuntu runner to strictly hit this.

Ah perfect, yes I think I've got one of these in already! an intentionally very large test.

test/core/config/validate_test.py
```diff
@@ -144,3 +145,39 @@ def test__validate_indentation_valid(config_dict):
...
     _validate_indentation_config(config_dict, "<test>")
+
+
+@pytest.mark.parametrize(
+    "config_dict,expected",
+    [
+        ({"core": {"max_parse_depth": None}}, -1),
+        ({"core": {"max_parse_depth": ""}}, -1),
+        ({"core": {"max_parse_depth": -1}}, -1),
+        ({"core": {"max_parse_depth": 0}}, 0),
```
keraion: Does it make sense to have `-1` and `0` do the same thing? If we default to `0` we get better alignment with Rust's interface using an unsigned value.

alanmcruickshank: That's very fair - I think it felt convoluted as I was writing it.

alanmcruickshank: @keraion - have another look. I think it's simpler now after that suggestion.

PR #7499, opened 2026-02-14
TSQL parsing issues: table var w index, create login, neg identity seed, alter drop [column/constraint]

dpurfield:

Parser support for table variables with indexes
Parser support for `CREATE LOGIN`
- `DEFAULT_DATABASE` non-quoted identifier
- `DEFAULT_LANUGE` non-quoted identifier 
- unicode (nString) password values
Parser support for column identity seed negative numbers
Parser support `ALTER TABLE`
- multiple delimited `DROP COLUMN`
- multiple delimited `DROP CONSTRAINT`

Removed invalid delimited section in alter table and invalid alter table example.

None

### Pull Request checklist

  - Other.


src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -4749,9 +4750,13 @@ class AlterTableStatementSegment(BaseSegment):
...
                 Sequence(
                     "DROP",
-                    "COLUMN",
-                    Ref("IfExistsGrammar", optional=True),
-                    Delimited(Ref("ColumnReferenceSegment")),
+                    Delimited(
+                        Sequence(
+                            "COLUMN",
+                            Ref("IfExistsGrammar", optional=True),
+                            Delimited(Ref("ColumnReferenceSegment")),
```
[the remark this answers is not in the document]
dpurfield: I tested and contrary to the documentation  DROP COLUMN col1, col2 is valid and executed  by the server and works as expected.  I did discover there was an outer Delimited section that was not needed and was allowing invalid and mixed alter table statements like:
ALTER TABLE dbo.SomeTable ADD COLUMN SomeCol int, DROP COLUMN SomeOtherColumn;  
Which is against the documentation and SQL considers invalid as well.

src/sqlfluff/dialects/dialect_tsql.py
```diff
@@ -5109,6 +5116,7 @@ class IdentityGrammar(BaseSegment):
         # optional (seed, increment) e.g. (1, 1)
         Bracketed(
             Sequence(
+                Ref("SignedSegmentGrammar", optional=True),
                 Ref("NumericLiteralSegment"),
                 Ref("CommaSegment"),
                 Ref("NumericLiteralSegment"),
```
peterbud, line 5123: If we are at it, the increment cannot be negative? TBH: I did not even know that seed can be negative

dpurfield: Never thought of it, but I tested and sure enough it can be.

dpurfield:
> Thanks @dpurfield - it looks good. I have one question on the seed/identity only.

@peterbud 

Fixed.  Nice catch!
Thanks

peterbud:
Hi @dpurfield  - sorry for coming back with another idea. but could you look at the `QualifiedLiteralSegment` from ANSI? I _believe_ that would be even a better solution for signed integers.

dpurfield:
> Hi @dpurfield - sorry for coming back with another idea. but could you look at the `QualifiedLiteralSegment` from ANSI? I _believe_ that would be even a better solution for signed integers.

@peterbud 

I don't think we can use QualifiedNumericLiteralSegment if I'm reading it correct as that would require + or -.   We want it to be optional.  Unless I changed it to optional in QualifiedNumericLiteralSegment.  I'm very leery of making changes to Ansi since that would impact all the other dialects.

peterbud:
OK, makes sense

PR #7505, opened 2026-02-16
Support for Snowflake DCM Projects `DEFINE` keyword

### LINKED ISSUE #7503: Support for Snowflake DCM Projects DEFINE keyword
### Search before asking

### Description

Snowflake DCM Projects are in private preview currently, but will hopefully be generally available soon. The gist of DCM Projects is that you define a set of Snowflake objects *declaratively* as code and Snowflake figures out the *declarative* instructions needed to make your database objects match what is in your code. The goal is to make CI/CD easier and prevent inconsistent state/schema drift.

Central to DCM projects is the `DEFINE` keyword. It is very similar to the `CREATE OR ALTER` syntax, but not quite interchangeable. It is specific to DCM projects. So to migrate an existing table definition to part of a DCM project, you would make the following change:

```diff
- CREATE OR ALTER TABLE mytable (
+ DEFINE TABLE mytable (
    col1 INT,
    col2 VARCHAR
);
```

Currently the `DEFINE` keyword results in an unparsable section. `DEFINE` should be more or less interchangeable with `CREATE`.

### Use case

Currently the `DEFINE` keyword results in an unparsable section. `DEFINE` should be more or less interchangeable with `CREATE`.

### Dialect

Snowflake

### Code of Conduct

robert-norberg:

Snowflake DCM Projects are in private preview currently, but will hopefully be generally available soon. The gist of DCM Projects is that you define a set of Snowflake objects declaratively as code and Snowflake figures out the declarative instructions needed to make your database objects match what is in your code. The goal is to make CI/CD easier and prevent inconsistent state/schema drift.

Central to DCM projects is the DEFINE keyword. It is very similar to the CREATE OR ALTER syntax, but not quite interchangeable. It is specific to DCM projects. So to migrate an existing table definition to part of a DCM project, you would make the following change:

```diff
- CREATE OR ALTER TABLE mytable (
+ DEFINE TABLE mytable (
    col1 INT,
    col2 VARCHAR
);
```

Currently DCM projects support `DEFINE`ing the following objects:

* Database
* Schema
* Table
* Dynamic Table
* View
* Internal stage
* Warehouse
* Role
* Grant
* Authentication Policy
* Task
* SQL Function
* Tag

Resolves #7503

No.

### Notes

In most cases, I just changed the existing classes to support beginning with either `CREATE` or `DEFINE`. `DEFINE` is mutually exclusive of (replaces) `CREATE OR REPLACE`/`CREATE OR ALTER` syntax, so I used `OneOf` to formalize that. Temporary objects are disallowed in DCM projects, and that is also reflected in the `DEFINE` portion of the `OneOf`.

In the case of tags, warehouses, and databases, the existing `CreateStatementSegment` covers these and several other `CREATE ...` statements that are *not* supported by DCM projects, so I created a separate `DefineStatementSegment` class rather than allowing `CREATE` or `DEFINE` at the beginning of `CreateStatementSegment`.

Similarly, DCM projects do not support creating external stages, so I created `DefineStageSegment` separate from `CreateStageSegment`.


peterbud (commented):
Hi @robert-norberg - thanks for this PR - it looks good.
I have only one question: semantically most of the changes will keep the `create_table/functions...` elements upon parsing, but the newly added pieces where no CREATE counterparts are in place will result in a semantic `define_...` elements. I'm not sure whether later any rules would need a homogeneous 'create_...' semantics or not. Honestly I think it's OK as is, just wanted to double check.

robert-norberg:
> Hi @robert-norberg - thanks for this PR - it looks good. I have only one question: semantically most of the changes will keep the `create_table/functions...` elements upon parsing, but the newly added pieces where no CREATE counterparts are in place will result in a semantic `define_...` elements. I'm not sure whether later any rules would need a homogeneous 'create_...' semantics or not. Honestly I think it's OK as is, just wanted to double check.

That's a good callout that I had not considered carefully. I searched the codebase for `create_function_statement`, `create_schema_statement`, and so on. I only found matches for two of the classes I modified/added:

* `create_table_statement` referenced in rules ST05, ST06
* `create_view_statement` referenced in rule ST06

I don't fully understand the use of these class `type`s in those rule files, but those rules applying identically to tables created via `CREATE` or `DEFINE` is desired behavior.

Importantly, I didn't find `create_statement` or `create_stage_statement` used anywhere other than `dialect_snowflake.py` and in `test/`. So my creation of alternative `define_statement` and `define_stage_statement` class `type`s does not result in any rules applying inconsistently to the creation of tags/wareshouses/databases/stages depending on the use of `CREATE` vs `DEFINE`.

However, `create_statement` or `create_stage_statement` could appear in a rule in the future, so if you would like, I can refactor to combine `CreateStatementSegment` + `DefineStatementSegment` and `CreateStageSegment` + `DefineStageSegment`. Or alternatively, I could change the `type` for `DefineStatementSegment` to `create_statement` (instead of the current `define_statement`) and change the `type` for `DefineStageSegment` to `create_stage_statement` (instead of the current `define_stage_statement`). Do you prefer either of these or no change?

PR #7556, opened 2026-02-28
ANSI: handle `IS [NOT] DISTINCT FROM` comparison statement

### LINKED ISSUE #3284: Postgres/Snowflake: `IS [NOT] DISTINCT FROM` is unparsable
### Search before asking

### What Happened

SQLFluff is not able to parse statement containing `IS [NOT] DISTINCT FROM` in Postgres/Snowflake.

Related Docs
* Postgres: https://wiki.postgresql.org/wiki/Is_distinct_from
* Snowflake: https://docs.snowflake.com/en/sql-reference/functions/is-distinct-from.html

### Expected Behaviour

SQLFluff is able to parse SQL statements containing `IS [NOT] DISTINCT FROM` for Postgres & Snowflake dialect.

### Observed Behaviour

SQLFluff reports a parsing violation:
```
==== parsing violations ====
L:   1 | P:  13 |  PRS | Line 1, Position 13: Found unparsable section: 'IS NOT DISTINCT'
```

### How to reproduce

1. Create a SQL file with the following content:
```
SELECT NULL IS DISTINCT FROM TRUE;
SELECT NULL IS NOT DISTINCT FROM TRUE;
```

2. Run SQLFluff using either Postgres / Snowflake dialect.
```
sqlfluff parse --dialect postgres sample.sql
```

### Dialect

Postgres, Snowflake

### Version

```
(.venv) ~/code/sqlfluff (main) $ sqlfluff --version
sqlfluff, version 0.13.1
(.venv) ~/code/sqlfluff (main) $ python --version
Python 3.9.12
```

### Configuration

None. Just the defaults.

### Code of Conduct

peterbud:

This PR is aiming to enable parsing the `IS [NOT] DISTINCT FROM` comparison statement in ANSI and enable this grammar in other supported dialects as well.

While the ANSI dialect already contained a grammar for this, unfortunately it was not working.

Summary of changes

* First of all: it's rather **unconventional** adding a new matcher, especially a lookbehind matcher, so please do a thorough review
* Added `PrecededByMatcher` class that looks backward from a FROM token to check if it's preceded by IS [NOT] DISTINCT. Used as an `exclude` on the FROM terminator in `SelectClauseTerminatorGrammar`, so FROM inside `IS [NOT] DISTINCT FROM` is not treated as the start of a FROM clause.

* 7 child dialects (bigquery, oracle, postgres, snowflake, sqlite, exasol, trino) — Updated their overridden `SelectClauseTerminatorGrammar` to use the same `Ref("FromKeywordSegment", exclude=_is_distinct_from_lookbehind)` pattern instead of bare "FROM".
* As it was changing a matcher, I had to update the rust layer as well

Fixes #3284

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/parser/grammar/lookbehind.py
+++ b/src/sqlfluff/core/parser/grammar/lookbehind.py
@@ -26,25 +26,18 @@ class PrecededByMatcher(Matchable):
     on the ``FROM`` keyword in ``SelectClauseTerminatorGrammar``.
 
     Args:
-        preceding: An ordered tuple of uppercase keyword strings that
-            must appear (in order, right-to-left) immediately before the
-            current position. Whitespace and meta segments between
-            keywords are skipped.
-        optional_preceding: An optional set of uppercase keyword strings.
-            Between each pair of required keywords, if the keyword at the
-            current lookbehind position matches any of these, it is consumed
-            (skipping whitespace/meta) before continuing to check the next
-            required keyword. This handles optional keywords like ``NOT``
-            in ``IS [NOT] DISTINCT FROM``.
+        preceding_sequences: Ordered tuples of uppercase keyword strings
+            that may appear immediately before the current position, where
+            matching is performed from right to left within each sequence
+            (i.e. the last item is checked first). Non-code and meta
+            segments between keywords are skipped.
     """
 
     def __init__(
         self,
-        preceding: tuple[str, ...],
-        optional_preceding: tuple[str, ...] = (),
+        preceding_sequences: tuple[tuple[str, ...], ...],
     ) -> None:
-        self.preceding = preceding
-        self.optional_preceding = optional_preceding
+        self.preceding_sequences = preceding_sequences
 
     def is_optional(self) -> bool:  # pragma: no cover
         """Return whether this element is optional.
@@ -61,18 +54,17 @@ class PrecededByMatcher(Matchable):
 
     def cache_key(self) -> str:  # pragma: no cover
         """Get the cache key for the matcher."""
-        return (
-            f"preceded-by-{'-'.join(self.preceding)}"
-            f"-opt-{'-'.join(self.optional_preceding)}"
+        return "preceded-by-" + "|".join(
+            "-".join(sequence) for sequence in self.preceding_sequences
         )
 
     @staticmethod
-    def _skip_whitespace_and_meta_backward(
+    def _skip_non_code_and_meta_backward(
         segments: Sequence["BaseSegment"],
         idx: int,
     ) -> int:
-        """Move ``idx`` backward past any whitespace or meta segments."""
-        while idx >= 0 and (segments[idx].is_whitespace or segments[idx].is_meta):
+        """Move ``idx`` backward past any non-code or meta segments."""
+        while idx >= 0 and (not segments[idx].is_code or segments[idx].is_meta):
             idx -= 1
         return idx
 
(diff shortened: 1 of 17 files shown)


peterbud:
@keraion I have added you as a reviewer, as the rust changes are largely AI generated, and I'm on thin ice with that, and need your expertise.

src/sqlfluff/core/parser/grammar/lookbehind.py
```diff
@@ -0,0 +1,114 @@
...
+            f"-opt-{'-'.join(self.optional_preceding)}"
+        )
+
+    @staticmethod
+    def _skip_whitespace_and_meta_backward(
+        segments: Sequence["BaseSegment"],
+        idx: int,
+    ) -> int:
+        """Move ``idx`` backward past any whitespace or meta segments."""
+        while idx >= 0 and (segments[idx].is_whitespace or segments[idx].is_meta):
```
keraion: Does this account for comments?

src/sqlfluff/dialects/dialect_bigquery.py
```diff
@@ -41,6 +41,7 @@
     TypedParser,
 )
 from sqlfluff.dialects import dialect_ansi as ansi
+from sqlfluff.dialects.dialect_ansi import _is_distinct_from_lookbehind
```
keraion: Importing this as it is internal doesn't feel right. Shouldn't we just add it to the dialect's grammar?

src/sqlfluff/core/parser/grammar/lookbehind.py
```diff
@@ -0,0 +1,114 @@
...
+from typing import Optional
+
+from sqlfluff.core.parser.context import ParseContext
+from sqlfluff.core.parser.match_result import MatchResult
+from sqlfluff.core.parser.matchable import Matchable
+from sqlfluff.core.parser.segments import BaseSegment
+from sqlfluff.core.parser.types import SimpleHintType
+
+
+class PrecededByMatcher(Matchable):
```
keraion, line 18: Should this inherit from `BaseGrammar` instead?

src/sqlfluff/core/parser/grammar/lookbehind.py
```diff
@@ -0,0 +1,114 @@
...
+        preceding: An ordered tuple of uppercase keyword strings that
+            must appear (in order, right-to-left) immediately before the
+            current position. Whitespace and meta segments between
+            keywords are skipped.
+        optional_preceding: An optional set of uppercase keyword strings.
+            Between each pair of required keywords, if the keyword at the
+            current lookbehind position matches any of these, it is consumed
+            (skipping whitespace/meta) before continuing to check the next
+            required keyword. This handles optional keywords like ``NOT``
+            in ``IS [NOT] DISTINCT FROM``.
```
keraion: Maybe this is just me, but this optional bit feels fragile. I think I'd prefer seeing the optional bit inline like the other Grammars do it `("IS", Ref.keyword("NOT", is_optional=True), "DISTINCT")` or similar so the order is explicit.

If we want to avoid `Matchable`s here, maybe it's a matter of checking against a `tuple[tuple[str]]` instead.

```python
(
    ("IS", "DISTINCT"),
    ("IS", "NOT", "DISTINCT"),
)
```

>uppercase keyword strings that must appear (in order, right-to-left) immediately before the current position

Wouldn't this be left-to-right for usage here, but evaluated right-to-left like noted in match below?

keraion (commented):
I know you were looking for the Rust review, but I think there are a few small issues with the approach. I think the idea is sound, but it breaks a few expectations we currently have. 

Ideally, if we can have it work with a dialect's grammar or allowing `Matchable`s, that would handle the importing issue, as well as the optional keywords, but I know that comes with additional implementation issues. If the `Matchable` is too much, I think the other alternative I commented may work.

src/sqlfluff/dialects/dialect_bigquery.py
peterbud: I have added a shared exclude pattern and all relevant dialects are importing `is_distinct_from_lookbehind`

src/sqlfluff/core/parser/grammar/lookbehind.py
peterbud: In theory yes, it would be possible, but probably not desirable. Inheriting from Matchable makes the intent clearer: this object is a lightweight matcher used for exclude/terminator-style checks, not a compositional grammar node.

BaseGrammar carries grammar composition machinery: _elements, allow_gaps, terminators, parse_mode, equality/copy behavior, and ref resolution. PrecededByMatcher does not use any of that. It is just a positional predicate.

src/sqlfluff/core/parser/grammar/lookbehind.py
peterbud: I have removed the optional_preceding, it looks much cleaner now!

PR #7558, opened 2026-03-01
feat(Clickhouse): Add QUALIFY keyword and grammar to Clickhouse

duvenagep:
## Add QUALIFY to Clickhouse

Currently SQLFluff fails to parse `qualify` when dialect is set to Clickhouse:

Screenshot 2026-03-01 at 11 44 57

<br>

This PR introduces the `QUALIFY` keyword along with the necessary grammar to correctly parse the `QualifyClauseSegment`. Test cases for all variants have also be added to the `test/fixtures/dialects/clickhouse/qualify.sql`

- no

### Pull Request checklist


src/sqlfluff/dialects/dialect_clickhouse.py
```diff
@@ -604,6 +604,16 @@ class SettingsClauseSegment(BaseSegment):
         optional=True,
     )
 
+class QualifyClauseSegment(BaseSegment):
+    """A `QUALIFY` clause like in `SELECT`."""
+
+    type = "qualify_clause"
+    match_grammar = Sequence(
+        "QUALIFY",
+        ImplicitIndent,
```
peterbud: Why ImplicitIndent? WHy not a simple Indent?

src/sqlfluff/dialects/dialect_clickhouse.py
```diff
@@ -613,10 +623,14 @@ class SelectStatementSegment(ansi.SelectStatementSegment):
         before=Ref("WhereClauseSegment", optional=True),
     ).copy(
         insert=[
+            Ref("QualifyClauseSegment", optional=True),
             Ref("FormatClauseSegment", optional=True),
             Ref("IntoOutfileClauseSegment", optional=True),
             Ref("SettingsClauseSegment", optional=True),
         ],
+    ).copy(
```
peterbud: Inserted twice. Also why is important to insert before OrdeByClause?

src/sqlfluff/dialects/dialect_clickhouse.py
duvenagep: @peterbud I looked at the Indent doc string and I think you are correct a simple `Indent` is better here.

src/sqlfluff/dialects/dialect_clickhouse.py
duvenagep: @peterbud based on the Clickhouse [SQL Reference](https://clickhouse.com/docs/sql-reference/statements/select/qualify) for the QUALIFY clause, it is before the `order by` clause. I have removed the double reference as well and all tests still pass.

peterbud: The order of the grammars are ONLY relevant where there are certain keywords which are overlapping, and it is important from the POV of parsing which one is getting parsed first. I'd suggest to include the `QualifeClauseSegment` just like the other segments, without specifying the before element. Test will reveal if that would not work (I believe it would work without problem)

duvenagep: @peterbud, I see what you mean! I tested the impl without specifying the `before` element, however now I do run into various errors related to the `order by` & `limit` clauses. Thus my understanding here was it is necessary to parse the `qualify` keyword before `order by` or `limit`? Unless I still understand the order of operation incorrectly?

Screenshot 2026-03-01 at 17 49 29
Screenshot 2026-03-01 at 17 47 20

peterbud (commented):
You are right, we need the `before` element.

peterbud:
Pls fix the linting errors and merge the latest main branch

duvenagep:
@peterbud Done! Should be ready now! Thanks for the help.

peterbud (approved):

PR #7577, opened 2026-03-04
BigQuery: support chained function calls like (x).UPPER()

ko-ya346:

BigQuery supports chained function calls where the result of an expression
is passed as the first argument to a function. For example:

```sql
SELECT (x).UPPER();
```

Currently sqlfluff parses this as:  
- an expression (x)
- followed by a semi_structured_expression .UPPER
- and the trailing () becomes unparsable

which results in PRS Found unparsable section.

This PR adds support for chained function calls in the BigQuery dialect by
introducing a ChainedFunctionCallSegment and extending AccessorGrammar
to recognize postfix calls of the form:  

```
.<function_name>(...)
```

This allows expressions such as:  
```
SELECT (x).UPPER();
```

to be parsed correctly.

Tests were added to ensure that chained function calls no longer produce
unparsable segments.  

---

This change only affects the BigQuery dialect and adds an additional accessor
pattern to AccessorGrammar.

It does not modify ANSI grammar or other dialects.

The existing behavior for expressions such as:  

```
SELECT x.UPPER();
```  
continues to work as before.   

### Pull Request checklist

  - Other.

This implementation follows BigQuery's documented chained function call syntax.

### DIFF
--- a/test/fixtures/dialects/bigquery/chained_function_call.sql
+++ b/test/fixtures/dialects/bigquery/chained_function_call.sql
@@ -1,2 +1,14 @@
 SELECT (x).UPPER() AS y;
 SELECT (x).SUBSTR(1, 3) AS y;
+SELECT
+  ('one two three four five')
+  .REPLACE('one', '1')
+  .REPLACE('two', '2')
+  .REPLACE('three', '3')
+  .REPLACE('four', '4')
+  .REPLACE('five', '5');
+SELECT (x).STRPOS("pattern string");
+SELECT (x).ARRAY_CONCAT(y);
+SELECT "Two birds and one mouse"
+  .REPLACE("bird", "dog")
+  .REPLACE("mouse", "cat") AS result;
--- a/test/fixtures/dialects/bigquery/chained_function_call.yml
+++ b/test/fixtures/dialects/bigquery/chained_function_call.yml
@@ -3,7 +3,7 @@
 # computed by SQLFluff when running the tests. Please run
 # `python test/generate_parse_fixture_yml.py`  to generate them after adding or
 # altering SQL files.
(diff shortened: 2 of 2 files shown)


test/dialects/bigquery_test.py
```diff
@@ -98,3 +98,23 @@ def test_cast_as_float_fails():
     parsed = Linter(dialect="bigquery").parse_string(sql)
     # Parsing produces a parse error for the unsupported `FLOAT` type.
     assert parsed.violations
+
+
+def test_bigquery_chained_function_calls_parse():
```
peterbud: Pls remove this. The designated parser tests should use the SQL fixtures and yml files.

test/fixtures/dialects/bigquery/chained_function_call.sql
```diff
@@ -0,0 +1,2 @@
+SELECT (x).UPPER() AS y;
```
peterbud, line 1: We need more tests here. You can get inspiration here: https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/functions-reference#chained-function-calls

test/dialects/bigquery_test.py
ko-ya346: Thanks for the feedback! I've removed the test.

test/fixtures/dialects/bigquery/chained_function_call.sql
ko-ya346: Thanks! I've added more examples based on the BigQuery documentation for chained function calls.

keraion (approved):
Seems sensible to me. Thanks for the contribution!

peterbud (approved):

PR #7593, opened 2026-03-11
Handle invalid pyproject.toml parse errors

alanmcruickshank:
## Summary
- catch TOML parse failures when loading pyproject.toml and raise a SQLFluffUserError
- include the config file path plus parse message and line/column details when available
- add a targeted hint for UTF-8 BOM-prefixed pyproject.toml files and cover the loader and CLI flows with tests

## Testing
- direct Python verification of invalid TOML and BOM-prefixed pyproject.toml handling
- python -m py_compile src/sqlfluff/core/config/toml.py test/core/config/loader_test.py test/cli/commands_test.py

## Notes
- pytest itself hit local temp-directory permission issues in this sandbox, so I verified the paths directly instead of relying on the test runner here.


alanmcruickshank:
Draft until we merge #7597

PR #7597, opened 2026-03-11
build: migrate linting and formatting checks to Ruff

alanmcruickshank:
tl;dr: We're getting annoying deviations between how `black` wants to format a file and how `ruff` does. I think it's time we went full in on ruff and just remove the old flake8 and black dependencies.

## Summary
- Remove Black and flake8-black from developer tooling and pre-commit, and standardize on Ruff for both linting and formatting.
- Update CI linting execution (`tox -e linting`) to run `ruff check .` and `ruff format --check .`, plus keep `lint-imports`.
- Align documentation/badges and apply one-time Ruff formatting churn (including one additional `test/cli/commands_test.py` formatting adjustment needed against current `main`).

## Rationale
- Current CI had formatter drift between Black and Ruff, causing checks to fight each other depending on job/tool versions.
- Using a single formatter authority removes non-deterministic formatting outcomes and reduces contributor friction.
- Keeping Ruff as both linter + formatter simplifies local setup and CI maintenance while preserving existing lint/import safety gates.

## Verification
- Verified locally on this branch: `python -m tox -e linting` passes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -234,6 +234,12 @@ ignore_missing_imports = true
 module = "tblib.*"
 ignore_missing_imports = true
 
+[tool.ruff]
+line-length = 88
+
+[tool.ruff.format]
+quote-style = "double"
+
 [tool.ruff.lint]
 extend-select = ["I", "D"]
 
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -84,9 +75,10 @@ repos:
         args: [-c=.yamllint]
   - repo: https://github.com/charliermarsh/ruff-pre-commit
```
keraion: We should go ahead and update this to the updated repo
```suggestion
  - repo: https://github.com/astral-sh/ruff-pre-commit
```

.pre-commit-config.yaml
```diff
@@ -84,9 +75,10 @@ repos:
...
     # Ruff version.
-    rev: "v0.14.13"
+    rev: "v0.15.5"
     hooks:
       - id: ruff
```
keraion: And update this from the legacy alias
```suggestion
      - id: ruff-check
```

PR #7624, opened 2026-03-17
lint: Handle ignored PRS lines without suppressing later diagnostics

### LINKED ISSUE #6199: Ignoring a single line (parse error) makes sqlfluff ignore the remainder of the file
### Search before asking

### What Happened

I need to ignore a few lines that give me parse errors. Appending `-- noqa` to those line(s) effectively fixes the parse error but also ignores any other true errors in the remainder of the file.

### Expected Behaviour

Any line not having the `--noqa` directive should be linted and any rule violation should be marked.

### Observed Behaviour

All lines coming after the line having `--noqa` never gets marked with errors, even if they clearly are erroneous.

### How to reproduce

Below lintresult is OK even though there are multiple errors.
```bash
sqlfluff lint --dialect sparksql - <<EOF
select 1 from tbl;
echo "this breaks sqlfluff" -- noqa
select from tbl; -- should fail
EOF
All Finished 📜 🎉!
```

Commenting the middle line now results in the 2 expected errors.
```bash
sqlfluff lint --dialect sparksql - <<EOF
select 1 from tbl;
-- echo "this breaks sqlfluff" -- noqa
select from tbl; -- should fail
EOF
== [stdin] FAIL
L:   3 | P:   1 |  PRS | Line 3, Position 1: Found unparsable section: 'select'
WARNING: Parsing errors found and dialect is set to 'sparksql'. Have you configured your dialect correctly?
All Finished 📜 🎉!
```

### Dialect

sparksql

### Version

sqlfluff, version 3.1.1
Python 3.11.9

### Configuration

no custom config

### Code of Conduct

peterbud:

This change fixes a parser recovery edge case where an inline `-- noqa` on a parse error line could cause the resulting unparsable section to swallow later statements, preventing their lint or parse violations from being reported.

For non-templated files, when a parse error is already ignored inline, SQLFluff now reparses using a masked version of just the ignored PRS line so later statements can still be analyzed normally.

The test coverage was expanded to include:

* ignored PRS followed by later lint and parse failures
* ignored PRS at end of file
* unsuppressed PRS before or after an ignored PRS
* recovery helper edge cases and early-return fallbacks

Fixes #6199

**Important** I'm a bit hesitant wit this PR TBH:
* This fixes only the `lint` command and `fix` won't work. 
* Also it only helps when the parse error is anchored on the same line as the `-- noqa`. For multiline broken statements, that is often not true. The parser usually reports the PRS at the start of the unparsable region, which may be an earlier line than the comment. So I think the applicability of this is rather limited.

Feedback is welcome!

### Pull Request checklist

  - Other.

### DIFF
--- a/test/core/rules/noqa_test.py
+++ b/test/core/rules/noqa_test.py
@@ -535,6 +535,160 @@ select from tbl; -- should fail PRS
     ]
 
 
+def test_linter_noqa_prs_at_end_of_file():
+    """Ignored PRS at EOF should not trigger recovery-only lint noise."""
+    lntr = Linter(
+        config=FluffConfig(
+            overrides={
+                "dialect": "sparksql",
+            }
+        )
+    )
+    sql = """select 1 from tbl;
+echo "this breaks sqlfluff" -- noqa
+"""
+
+    violations = lntr.lint_string(sql).get_violations(filter_warning=False)
+
+    assert violations == []
+
+
+def test_mask_ignored_parse_error_lines_empty_line_set():
+    """No masking should occur if no line numbers are supplied."""
+    source_str = 'echo "this breaks sqlfluff" -- noqa\n'
+    dialect = Linter(config=FluffConfig(overrides={"dialect": "sparksql"})).dialect
+
+    assert (
+        Linter._mask_ignored_parse_error_lines(source_str, dialect, set()) == source_str
+    )
+
+
+def test_mask_ignored_parse_error_lines_out_of_range_without_comment_matcher():
+    """Out-of-range lines are skipped and lines can be blanked without comments."""
+    source_str = 'echo "this breaks sqlfluff"\n'
+
+    class _DialectWithoutInlineComment:
+        lexer_matchers = []
+
+    masked = Linter._mask_ignored_parse_error_lines(
+        source_str,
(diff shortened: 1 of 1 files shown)


alanmcruickshank:
I'm on the fence too about this one. It feels like a lot of complexity to add for not much upside.

I think a strategy where we take an ignored PRS line and then allow linting outside of that statement or other greedy block would be good - but I think that already happens.

Anyone else views?

keraion:
I'm in agreement with Alan here. The recovery mechanism feels very fragile and it's making `noqa` have two distinct jobs. I think we should let `noqa` continue to do its thing, but we could introduce a couple of mechanics to help users when we don't support a specific dialect statement.

- 1) add a parser sync comment. something like `-- sqlfluff prs:sync` where it can be like a sync point to go back up to the root segment and continue to the next statement.
- 2) not exclusive of the first option but adding an `UnparsableStatementSegment` or `UnparsableBatchSegment` as a last resort matcher for the non-greedy blocks to progress. Something along the lines of `Anything(terminators=[Ref("DelimiterGrammar")])`

Just some thoughts

peterbud:
OK, I think we agree on this. Let's close this PR, and close the issue as not planned ATM. Long term we can think on other alternatives.

PR #7642, opened 2026-03-18
Oracle: parse TABLESPACE/STORAGE and other table/index options

### LINKED ISSUE #7641: Oracle: parser missing multiple physical/storage & segment-attribute keywords in CREATE TABLE/INDEX
### Search before asking

### What Happened

The Oracle dialect grammar referenced several physical/storage tokens that were not present in the dialect implementation, causing parse failures and preventing correct parsing/formatting of Oracle CREATE TABLE/INDEX. The following example will fail to parse on the current sqlfluff version:

```sql
CREATE TABLE myspace.tablename (
    id NUMBER(18) PRIMARY KEY,
    payload VARCHAR2(4000)
)
TABLESPACE MYSPACE_DATA_01
LOGGING
PCTFREE 10
PCTUSED 70
INITRANS 1
MAXTRANS 255
STORAGE (
    INITIAL 655K
    NEXT 655K
    MINEXTENTS 1
    MAXEXTENTS UNLIMITED
    PCTINCREASE 1
    FREELISTS 1
    FREELIST GROUPS 1
    BUFFER_POOL DEFAULT
)
PARALLEL 4
NOCACHE;
```

### Expected Behaviour

No parse errors.

### Observed Behaviour

The example fails to parse.

### How to reproduce

- Create a file with the example from above
- Parse it with sqlfluff using the oracle dialect

### Dialect

oracle

### Version

4.0.4

### Configuration

Default.

### Code of Conduct

saulotoledo:

- Add missing Oracle keywords
- Implement `StorageClauseSegment`, `OraclePhysicalAttributesSegment` and `CreateIndexStatementSegment`
- Add comprehensive `CREATE TABLE` and `CREATE INDEX` fixtures exercising the new clauses
- Update indentation tests at `LT02-indent-oracle` to cover some of the new introduced structures.
- Align sqlfluffrs (Rust parser) with Python by removing premature advancement of `matched_idx` into whitespace in `AnyNumberOf/AnySetOf`, so early-return paths use the pre-gap boundary and trailing whitespace is excluded from matches.

Fixes #7641.

No.

### Pull Request checklist

  - Other.

### DIFF
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/table_driven/anynumberof.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/table_driven/anynumberof.rs
@@ -369,11 +369,6 @@ impl Parser<'_> {
             return Ok(stack.transition_to_combining(frame, None));
         }
 
-        // We no longer "collect tokens", it's all lazy evaluation now.
-        if allow_gaps && *ctx.matched_idx < *ctx.working_idx {
-            *ctx.matched_idx = *ctx.working_idx;
-        }
-
         // Check max_times constraint
         if let Some(max) = max_times {
             if *ctx.count >= max {
--- a/src/sqlfluff/dialects/dialect_oracle.py
+++ b/src/sqlfluff/dialects/dialect_oracle.py
@@ -6,6 +6,7 @@ This inherits from the ansi dialect.
 from sqlfluff.core.dialects import load_raw_dialect
 from sqlfluff.core.parser import (
     AnyNumberOf,
+    AnySetOf,
     Anything,
     BaseFileSegment,
     BaseSegment,
@@ -87,8 +88,8 @@ oracle_dialect.sets("reserved_keywords").update(
         "ELSE",
         "ENABLE",
         "EXCLUSIVE",
-        "EXISTS",
         "EXECUTE",
+        "EXISTS",
         "FILE",
         "FLOAT",
         "FOR",
@@ -152,8 +153,8 @@ oracle_dialect.sets("reserved_keywords").update(
         "REBUILD",
         "RENAME",
         "RESOURCE",
-        "REVOKE",
         "REVERSE",
+        "REVOKE",
         "ROW",
         "ROWID",
         "ROWNUM",
@@ -200,27 +201,36 @@ oracle_dialect.sets("unreserved_keywords").update(
         "ACCESSIBLE",
         "ACTIVE",
         "ADMINISTER",
+        "ADVANCED",
         "ADVISE",
         "ADVISOR",
         "ANALYTIC",
-        "ARCHIVE",
         "ARCHIVAL",
+        "ARCHIVE",
         "AUTHENTICATED",
         "AUTHID",
+        "AUTO",
+        "BASIC",
         "BECOME",
+        "BITMAP",
         "BODY",
+        "BUFFER_POOL",
         "BULK",
-        "COMMITTED",
-        "CONSTRAINTS",
         "BULK_EXCEPTIONS",
         "BULK_ROWCOUNT",
         "BYTE",
+        "CAPACITY",
+        "CELL_FLASH_CACHE",
         "COLLECT",
+        "COMMITTED",
         "COMPILE",
         "COMPOUND",
         "CONSTANT",
+        "CONSTRAINTS",
         "CONTAINER",
         "CONTEXT",
+        "CREATION",
+        "CRITICAL",
         "CROSSEDITION",
         "CURSOR",
         "DBA_RECYCLEBIN",
@@ -234,7 +244,9 @@ oracle_dialect.sets("unreserved_keywords").update(
         "DIRECTIVE",
         "DIRECTORIES",
         "DIRECTORY",
+        "DISTRIBUTE",
         "DML",
+        "DUPLICATE",
         "EDITION",
         "EDITIONABLE",
         "EDITIONING",
@@ -248,14 +260,21 @@ oracle_dialect.sets("unreserved_keywords").update(
         "EXTERNALLY",
         "FINE",
         "FLASHBACK",
+        "FLASH_CACHE",
         "FOLLOWS",
         "FORALL",
+        "FREELIST",
+        "FREELISTS",
         "GLOBALLY",
+        "GROUPS",
         "GUARD",
         "HIERARCHY",
+        "HIGH",
         "HTTP",
         "INDICES",
         "INHERITANY",
+        "INITRANS",
+        "INMEMORY",
         "ISOLATION_LEVEL",
         "ISOPEN",
         "JAVA",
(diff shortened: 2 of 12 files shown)


saulotoledo:
@greptile-apps

saulotoledo:
@greptile-apps

saulotoledo:
For reviewers: I will fix the remaining issues tomorrow.

saulotoledo:
@greptile-apps

saulotoledo:
@greptile-apps

saulotoledo:
@greptile-apps

saulotoledo:
@greptile-apps

saulotoledo:
@greptile-apps

saulotoledo:
@greptile-apps

saulotoledo:
@greptile-apps

saulotoledo:
@greptile-apps

saulotoledo:
@peterbud, I believe this is ready to review.
The following concern pointed out by the Greptile Bot is real:

> One grammar accuracy gap: NOSORT and REVERSE are documented as mutually exclusive in Oracle but the AnyNumberOf grammar allows both to appear in sequence without error

However, I do not see how to solve it directly in the grammar at this point. Note that the problem also happens to `LOGGING/NOLOGGING`, `VISIBLE/INVISIBLE`, and we can also have repeated attributes, which should not be allowed. We cannot use `Sequence`, because that would enforce order. Maybe this should be addressed by a lint rule instead, but before trying to implement it, I would like to hear from you, and maybe this could be addressed in a separate issue.

I also added the examples suggested by @123Haynes in the linked issue.

Let me know how do you prefer to proceed with this PR.

peterbud:
> However, I do not see how to solve it directly in the grammar at this point. Note that the problem also happens to `LOGGING/NOLOGGING`, `VISIBLE/INVISIBLE`, and we can also have repeated attributes, which should not be allowed. We cannot use `Sequence`, because that would enforce order. Maybe this should be addressed by a lint rule instead, but before trying to implement it, I would like to hear from you, and maybe this could be addressed in a separate issue.

I think that's fine. Sqlfluff is not doing semantic analysis, it's a humble parser / linter with it's own limitation, and we are OK to have some cases where we accept the statements as syntactically OK, but the given engine would not allow such combinations based on semantic rules.

peterbud:
@saulotoledo the other option you have is to use `AnySetOf` instead of 'AnyNumberOf`:

```python
    match_grammar: Matchable = AnySetOf(
        # Tablespace placement
        Sequence("TABLESPACE", Ref("ObjectReferenceSegment")),
        ....
```

That would NOT allow to have `NOSORT REVERSE`.

The key point is that `AnySetOf` deduplicates by matched top-level element, not by the literal token chosen inside that element. So that whole `OneOf` group is treated as one slot.

Practical caveat:
* If Oracle really allows some of the other physical attributes to repeat, then `AnySetOf` would be too strict for the whole construct. I have tried the `AnySetOf` and it works with the current fixture set , but they have repetitions.
* If they are all intended to be singleton options, `AnySetOf` is the right model and is probably better than the current `AnyNumberOf` anyway.

saulotoledo:
Hi @peterbud, thanks for your comments! I think `AnySetOf` is what we need here. I will review the grammar to be sure and do this change today later.

saulotoledo:
@greptile-apps

saulotoledo:
@greptile-apps

saulotoledo:
@greptile-apps

saulotoledo:
@greptile-apps

saulotoledo:
@greptile-apps

saulotoledo:
@greptileai Oracle docs show `NOREVERSE` is valid for `ALTER INDEX ... REBUILD` (the rebuild clause), but not a `CREATE INDEX` physical-attribute: `CREATE` supports `REVERSE` only. Reevaluate the gap mentioned and propose valid changes, if required.

saulotoledo:
@greptileai This is already implemented; recheck the two locations below and update your comments:

- *CREATE INDEX*: `OracleIndexPhysicalAttributesSegment` allows only `NOSORT` or `REVERSE` (no `NOREVERSE`). `REVERSE` and `NOSORT` are mutually exclusive and appear in a `OneOf`.
- *ALTER INDEX REBUILD*: `AlterIndexStatementSegment` accepts `REVERSE` or `NOREVERSE`.

saulotoledo:
Hi @peterbud. I think this is ready for another review.

A few comments regarding the AI review:

a) https://docs.oracle.com/en/database/oracle/oracle-database/26/sqlrf/CREATE-INDEX.html states:
- **SORT | NOSORT**: "...You cannot specify `REVERSE` with this clause...."
- **REVERSE**: There is no mention to `NOREVERSE` in this page
- Therefore, I believe the AI request to add `NOREVERSE` is not correct. Could you help me confirm this?

b) Regarging the new file `oracle_test.py` that "uses Python-based rejection tests instead of SQL/YAML fixtures": I am not sure these tests can be implemented using the YAML fixtures. And I believe they are important to make sure we do not break the grammar with later changes.

When implementing the solution, I found the parser error failure message quite generic ("Found unparsable section...") for indicating what happened, and maybe a lint would be a better alternative. However, this would require relaxing the grammar first (not failing on the duplications there -> back to `AnyNumberOf`, for example), and then implementing a lint rule for the duplications, and we could give better error messages like "Duplicate attribute 'PCTFREE' with values...". I still like the idea of a stronger grammar (and thus prefer the `AnySetOf` approach). But I also see that failures in complex queries do not give enough information on what failed, and a precise error message would be nice as well. Maybe this should be added as a comment somewhere, so the project could decide on what approach to follow in the future... What do you think about it?

Let me know if you require more changes, and we take it from there. :)

peterbud:
Hi @saulotoledo  - the CI pipeline shows failing tests - can you look into those and fix them?

saulotoledo:
@greptileai, analyze the PR and report findings on:

- Rust parser parity: Review the sqlfluffrs fix for `AnyNumberOf`/`AnySetOf`. Confirm whether it matches the Python parser's behavior and is correct.

- Oracle `REVERSE`/`NOREVERSE`: Verify grammar against Oracle docs (https://docs.oracle.com/en/database/oracle/oracle-database/26/sqlrf/CREATE-INDEX.html). Confirm `CREATE INDEX` only allows `SORT | NOSORT` and `REVERSE` (no `NOREVERSE`), and that `NOREVERSE` is handled at `ALTER INDEX` via `AlterIndexStatementSegment`. Flag mismatches and suggest concrete fixes if needed.

saulotoledo:
@greptileai

saulotoledo:
@peterbud I believe the last issue was in the Rust parser, not in the grammar. I commited a change there that both fixed the tests and was accepted by the AI review, but the Rust code here is new for me and I am not yet confident that the solution is correct. Could you please review that change and let me know your impressions?

sqlfluffrs/sqlfluffrs_parser/src/parser/table_driven/anynumberof.rs
keraion: First off, good catch here.

I think the fix might be much simpler. If we delete the `if allow_gaps && ...` block here the return paths have the correct `matched_idx` and old line 404 is unconditionally moving that index afterwards. We don't need to adjust the `matched_idx`, it was only making it incorrect in these two paths.

saulotoledo: Thanks for your review, @keraion! The proposed change worked.
Let me know if you require more changes on the PR.

Issue #4080, opened 2022-11-17
Leaked Semaphore objects


samLozier:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

When running sqlfluff with multiprocessing, I frequently get leaked semaphore warnings when sqlfluff finishes. I haven't identified what causes this exactly, but I'm wondering if it happens when I kill a sqlfluff run and sqlfluff doesn't shut down all the processes it started (eg ctrl+c). 

### Expected Behaviour

No leakage of seamphore objects.

### Observed Behaviour

```/usr/local/Cellar/python@3.9/3.9.13_4/Frameworks/Python.framework/Versions/3.9/lib/python3.9/multiprocessing/resource_tracker.py:216: UserWarning: resource_tracker: There appear to be 104 leaked semaphore objects to clean up at shutdown
```

### How to reproduce

As I mentioned above, I'm not exactly sure what's been triggering this, but my best guess is that I get the warning after this sequence: 
1. start sqlfulff
2. kill sqlfluff with ctrl+c
3. start sqlfluff again
4. sqlfluff finishes successfully and produces this leaked semaphore warning

What i'm not sure of is if the warning is coming from the second successful run, or is a leftover error from the first run. 

### Dialect

Snowflake 

### Version

sqlfluff, version 1.4.1
sqlfluff-templater-dbt 1.4.1

### Configuration

[sqlfluff]
exclude_rules = L031, L033, L034, L023
select_clause_trailing_comma="forbid"
verbose = 0
nocolor = False
dialect = snowflake
templater = dbt
rules = None
recurse = 0
output_line_length = 120
runaway_limit = 10
ignore_templated_areas = True
encoding = autodetect
processes = 0
# Comma separated list of file extensions to lint.

# NB: This config will only apply in the root folder.
sql_file_exts = .sql,.sql.j2,.dml,.ddl

[sqlfluff:templater:dbt]
# project_dir = .
# profiles_dir = ./profiles/
# profile = snowflake-db

[sqlfluff:indentation]
indented_joins = True
indented_using_on = True
template_blocks_indent = True

[sqlfluff:templater]
unwrap_wrapped_queries = True

[sqlfluff:templater:jinja]
apply_dbt_builtins = True
load_macros_from_path= macros/

[sqlfluff:templater:jinja:macros]
# Macros provided as builtins for dbt projects
[...]

PR #7686, opened 2026-03-29
fix: Ensure `pool.join()` is called in ParallelRunner

peterbud:
It's another attempt to eliminate "leaked semaphore" messages  and to prevent resource leaks.

Makes progress on #4080

`Pool.__exit__()` (used by the `with pool:` context manager) calls `pool.terminate()` but not `pool.join()`. Without `join()`, worker processes may still be running when Python's resource_tracker checks for leaked resources at shutdown. The pool's internal `SimpleQueue` objects use named POSIX semaphores for their locks, and these (might) get reported as "leaked semaphore objects" because the workers haven't fully exited yet.

Changes:
* Replaced `with self._create_pool(...) as pool:` with explicit `pool = self._create_pool(...)` + `try/finally`. 
* The finally block always calls `pool.terminate()` followed by `pool.join()`, ensuring all worker processes have fully exited and their semaphore references are released before the method returns
* This covers all exit paths: normal completion, GeneratorExit (early close from `lint_paths`, and `KeyboardInterrupt`
* Extended mock ErrorPool in tests, and added tests to verify the method calls

### Pull Request checklist

  - Other.

### DIFF
--- a/src/sqlfluff/core/linter/runner.py
+++ b/src/sqlfluff/core/linter/runner.py
@@ -179,46 +179,53 @@ class ParallelRunner(BaseRunner):
         the main thread can do the IO work while passing the parsing
         and linting work out to the threads.
         """
-        with self._create_pool(
+        # NOTE: We avoid using `with pool:` here because Pool.__exit__
+        # calls pool.terminate() but NOT pool.join(). Without join(), worker
+        # processes may still be alive when Python's resource_tracker runs at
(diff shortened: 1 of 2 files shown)


alanmcruickshank (approved):
Looks sensible. I kind of hit a dead end on trying to resolving this issue, so hopefully this does the trick!

PR #6720, opened 2025-03-11
Leading comma alignment

vinceyyy:
### Brief summary of the change made
Closes #4901. 

Add a modifier `:align-following` to `line_position = leading`. When enable, the indentation would be calculated based on the first non-whitespace following
- When leading comma is enabled, allow the indentation to be aligned the select target instead of the leading comma
- When leading binary operator is enabled, allow the indentation to be aligned on the join on conditions instead of the leading binary operator

Example:
```sql
/* 
  ...
  [sqlfluff:layout:type:comma]
  line_position = trailing

  [sqlfluff:layout:type:binary_operator]
  line_position = trailing
*/
SELECT 
    foo.a
    , foo.b
FROM
    foo
    INNER JOIN bar
        ON 
            foo.a = bar.a
            AND foo.a = bar.a;

/* 
  assuming all other configs are the same except the following:

  [sqlfluff:layout:type:comma]
  line_position = trailing:align-following

  [sqlfluff:layout:type:binary_operator]
  line_position = trailing:align-following
*/
SELECT 
    foo.a
  , foo.b
FROM
    foo
    INNER JOIN bar
        ON 
            foo.a = bar.a
        AND foo.a = bar.a;
```

### Are there any other side effects of this change that we should be aware of?
Not that I am aware of.

### Pull Request checklist
- [ ] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

PR #7702, opened 2026-04-01
fix: align following line position handling

mauro-lanza:

Rules LT03, LT04, and LT08 crash or misbehave when `line_position` is set to `leading:align-following` (introduced in #6720) These rules compared the config value with equality (`== "leading"` / `== "trailing"`), which fails when modifier suffixes like `:align-following` are added.

**LT08** crashes with `NotImplementedError` when linting multi cte queries:
```
CRITICAL [LT08] Applying rule LT08 threw an Exception:
  Unexpected global comma style 'leading:align-following'
```

**LT03/LT04** silently skip their performance shortcut, falling through to the full reflow path. No crash, but unnecessary overhead (not sure if there are other rules affected as well).

**Reindent engine** logs a (distracting) `WARNING` ("Not enough space to compensate indentation") when `leading:align-following` commas appear at indent level 0 (e.g. between top-level ctes) where there is simply no indent to compensate.

**Fixes:**
- LT08: strip modifier suffix with `.split(":")[0]` before comparison
- LT03: strip modifier in `_check_trail_lead_shortcut()` (also fixes inherited LT04)
- `_calculate_desired_starting_indent`: early-return with `debug` log at indent level 0 (expected no-op), preserve `warning` for genuinely insufficient indent space

> [!NOTE]
> This is my first contributing pr in the project so feel free to raise any issue with it or provide guidance

The indentation compensation warning at indent level 0 is downgraded from `WARNING` to `DEBUG`. This only affects the case where compensation is impossible (zero indent), which was always a no-op returning the correct result. The `WARNING` is kept for cases where indent space is actually insufficient (e.g. `tab_space_size=2` with a 4-char compensation).

### Pull Request checklist

  - Other.

### DIFF
--- a/test/utils/reflow/reindent_test.py
+++ b/test/utils/reflow/reindent_test.py
@@ -1004,3 +1004,41 @@ def test_reflow__indent_compensation_success_path(default_config, caplog):
     assert result is not None
     # The warning is logged (visible in test output) but we don't assert on it
     # due to pytest-xdist capture limitations
+
+
+def test_reflow__indent_compensation_skipped_at_level_zero(default_config):
+    """Test that indent compensation is skipped at indent level 0.
+
+    When a leading comma appears at indent level 0 (e.g. between top-level
+    CTEs), there is no indent to compensate. This should be a silent no-op
+    rather than a warning.
+    """
+    from sqlfluff.utils.reflow.config import ReflowConfig
+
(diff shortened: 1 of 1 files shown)


peterbud:
@mauro-lanza - thanks for opening this. The CI failure is due to a CI drift. Just rebase / merge to the main branch first pls.

PR #7743, opened 2026-04-10
Add aliasing.required rule AL10: derived tables must have an alias

### LINKED ISSUE #3650: Refine L025: flag missing aliases if configured to do so
### Search before asking

### Description

( Discussed in this Slack thread: https://sqlfluff.slack.com/archives/C01JS7H598E/p1658480978527069 )

changing L025 from:
```
Tables should not be aliased if that alias is not used.
```

To something like:
```Require table aliases if database requires them or if configured to do so. Remove unused table aliases if database does not require them unless configured to keep them.
```

Removing aliases is an "autofix". Missing aliases are a lint error, presumably with no autofix (I'm guessing people don't want SQLFluff generating alias names for them.)

### Use case

Allow to enforce use of table aliases (for example for subqueries)

### Dialect

All dialects AFAIK

### Code of Conduct

cherylschaefer:

Adds a new rule **AL10** (`aliasing.required`) that flags derived tables (subqueries in `FROM` clauses) missing an alias. This is a runtime error in every major SQL dialect (MySQL, PostgreSQL, T-SQL) but was not previously caught by sqlfluff.

We discovered this gap when an unaliased derived table passed sqlfluff linting but caused a production error in MySQL. ("Every derived table must have its own alias")

Fixes: #3650. Rather than extending AL05 with a config option, this is implemented as a separate rule because the concerns are distinct: AL05 flags *unused* aliases, while AL10 flags *missing required* aliases. The fix strategy differs between the two. An extra alias could be easily removed (or configured to be left in), while a missing required alias causes a production error.

Existing SQL that has unaliased derived tables will now receive a lint violation. This is intentional, the SQL would fail at runtime in MySQL, PostgreSQL, and T-SQL. The rule is not auto-fixable (`is_fix_compatible = False`) since meaningful alias names cannot be generated automatically.

This rule applies to all dialects, which matches the behavior described in AL05's `_is_alias_required()`. Derived tables require aliases in every dialect I checked. I didn't extract a shared utility because this is only one other usage, also that would mean modifying AL05 which increases the scope and risk of the PR.

### Pull Request checklist

  - Other.

AI disclosure: Claude (Anthropic) helped me out with this one. I'm less familiar with Python and would be happy to make corrections. The result was verified against a real-world production SQL file that triggered this specific error.

### DIFF
--- a/src/sqlfluff/rules/aliasing/AL10.py
+++ b/src/sqlfluff/rules/aliasing/AL10.py
@@ -1,0 +1,80 @@
+"""Implementation of Rule AL10."""
+
+from sqlfluff.core.parser import BaseSegment
+from sqlfluff.core.rules import BaseRule, EvalResultType, LintResult, RuleContext
+from sqlfluff.core.rules.crawlers import SegmentSeekerCrawler
+
+
+class Rule_AL10(BaseRule):
+    """Derived tables must have an alias.
+
+    A derived table (subquery in a ``FROM`` clause) without an alias will
+    cause a runtime error in most SQL dialects including MySQL, PostgreSQL,
+    and T-SQL.
+
+    **Anti-pattern**
+
+    A subquery in a ``FROM`` clause without an alias.
+
+    .. code-block:: sql
+
+        SELECT *
+        FROM (
+            SELECT 1 AS a
+        )
+
+    **Best practice**
+
+    Add an alias to the derived table.
+
+    .. code-block:: sql
+
+        SELECT *
+        FROM (
+            SELECT 1 AS a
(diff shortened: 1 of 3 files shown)


peterbud (changes requested):
Hi @cherylschaefer - thanks for opening this PR. Overall looks good, pls see the feedback related to the consistency of the docstrings.

cherylschaefer:
Good catch, thanks! Updated

peterbud (approved):
Thanks @cherylschaefer 👌

cherylschaefer:
Thank you @peterbud !!

PR #7781, opened 2026-04-22
fix(tsql): keep required OPENROWSET BULK aliases

### LINKED ISSUE #7779: [tsql] OPENROWSET BULK correlation name treated as alias
### Search before asking

### What Happened

sqlfluff treats `OPENROWSET` correlation names as an alias.  However, where `OPENROWSET` is used with the `BULK` option, a correlation name is required in the FROM clause, regardless of whether the alias is used or not.

Syntax varies slightly but the requirement exists across Synapse, SQL Server & (I believe) Fabric.  

https://learn.microsoft.com/en-us/sql/t-sql/functions/openrowset-bulk-transact-sql?view=sql-server-ver17&viewFallbackFrom=azure-sqldw-latest

### Expected Behaviour

When AS <name> follows `OPENROWSET(BULK...` alias checks are not applied.

### Observed Behaviour

`sqlfluff fix <filename>` removes `OPENROWSET BULK` correlation names.  Presumably these are identified as an alias, then if not referenced elsewhere, removed as "unused". 

### How to reproduce

Running `sqlfluff fix <filename>` over the code below will result in `AS result` being detected as an (unused) alias and subsequently removed.  

```
SELECT jsoncontent
FROM
    OPENROWSET (
        BULK (
            '/lake-raw/activity*.json'
            ,'/lake-raw/this_file_doesnt_exist.json'
        ), 
        DATA_SOURCE = 'datasource_lake'
    )
    WITH (
        jsoncontent varchar(MAX)
    ) AS result
```
Second, similar pattern - `AS name` will be removed.
```
SELECT jsoncontent
FROM
    OPENROWSET (
        BULK (
            '/lake-raw/activity*.json'
            ,'/lake-raw/this_file_doesnt_exist.json'
        )
        ,DATA_SOURCE = 'datasource_lake'
    ) AS name
```

### Dialect

tsql

### Version

sqlfluff: 4.1.0
python: 3.12.11

### Configuration

n/a

### Code of Conduct

cloudyun888:
## Summary
- treat T-SQL `OPENROWSET(BULK ...)` correlation names as required instead of unused aliases
- keep `sqlfluff fix` from removing valid `AS result` / `AS name` correlation names
- add regression fixtures covering OPENROWSET BULK forms with and without a WITH clause

## Testing
- `.venv/bin/python -m pytest test/rules/yaml_test_cases_test.py -k openrowset_bulk_requires_correlation_name -q`
- `.venv/bin/python -m py_compile src/sqlfluff/rules/aliasing/AL05.py`
- `.venv/bin/sqlfluff fix /tmp/sqlfluff_7779.sql --dialect tsql` (verified the alias is preserved)

Fixes #7779

### DIFF
--- a/src/sqlfluff/rules/aliasing/AL05.py
+++ b/src/sqlfluff/rules/aliasing/AL05.py
@@ -230,6 +230,13 @@ class Rule_AL05(BaseRule):
         # is actually *required* in order for SQL Server to parse it.
         for segment in from_expression_element.iter_segments(expanding=("bracketed",)):
             if segment.is_type("table_expression"):
+                if dialect_name == "tsql":
+                    openrowset = segment.get_child("openrowset_segment")
+                    if openrowset and any(
+                        seg.raw_upper == "BULK"
+                        for seg in openrowset.recursive_crawl("keyword")
+                    ):
+                        return True
                 # Found a table expression. Does it have a VALUES clause?
                 if segment.get_child("values_clause"):
                     # Found a VALUES clause. Is this a dialect that requires
--- a/test/fixtures/rules/std_rule_cases/AL05.yml
+++ b/test/fixtures/rules/std_rule_cases/AL05.yml
@@ -241,6 +241,39 @@ test_pass_snowflake_values:
     core:
       dialect: snowflake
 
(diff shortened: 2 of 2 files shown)


peterbud (approved):
Thanks!

PR #7783, opened 2026-04-22
fix(duckdb): prevent factorial operator from consuming ! in != comparisons

armorbreak001:
## Summary

In the DuckDB dialect, `!` serves dual purposes: as a postfix factorial operator (`5!`) and as part of the not-equal comparison operator (`!=`). The current grammar greedily matches `!` as a factorial operator via `FactorialOperatorSegment` in `Expression_C_Grammar`, which prevents the comparison parser from recognizing `!=` as a single not-equal operator.

**Before (incorrect):**
```
factorial_operator:        !
comparison_operator:
    raw_comparison_operator: =   # only =, ! was consumed by factorial
```

**After (correct):**
```
comparison_operator:
    raw_comparison_operator: !   # both tokens under comparison_operator
    raw_comparison_operator: =
```

This caused false-positive **LT01** (spacing) violations when using `preferred_not_equal_style = c_style` with DuckDB, since the linter saw two separate operators (factorial + equals) that appeared to need whitespace between them.

## Fix

Introduced a lightweight `_FactorialTypedParser` subclass of `TypedParser` that checks the next raw token before matching. When `!` is immediately followed by `=`, it yields (returns no match), allowing the comparison grammar to handle `!=` correctly. Normal factorial usage like `5!` is unaffected since the next token is not `=`.

## Testing

- All 131 existing DuckDB dialect tests pass
- Verified `select 5!` still parses with `factorial_operator`
- Verified `where a != b` now parses with `comparison_operator` containing both `!` and `=` as `raw_comparison_operator`

### DIFF
--- a/src/sqlfluff/dialects/dialect_duckdb.py
+++ b/src/sqlfluff/dialects/dialect_duckdb.py
@@ -30,6 +30,28 @@ from sqlfluff.core.parser import (
     SymbolSegment,
     TypedParser,
 )
+from sqlfluff.core.parser.match_result import MatchResult
+from sqlfluff.core.parser.parsers import TypedParser as _TypedParser
+
+
+class _FactorialTypedParser(_TypedParser):
+    """TypedParser for factorial ``!`` that yields when followed by ``=``.
+
+    In DuckDB, ``!`` serves as both a postfix factorial operator and as
+    part of the not-equal comparison operator (``!=``).  To avoid
+    greedily consuming the ``!`` before the comparison parser can
+    match ``!=``, we skip matching when the next raw token is ``=``.
+    """
+
+    def match(
+        self, segments, idx, parse_context
+    ):  # pragma: no cover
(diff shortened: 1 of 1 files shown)


peterbud:
Hi @armorbreak001 - thanks for opening this PR.

Have you checked the other open PR #7742 ? Is this aiming to fix the same?

armorbreak001:
Hi @peterbud, thanks for the heads-up! I saw #7742 — it takes a different approach (tokenizing `!=` as a single token at the lexer level), while this PR works at the grammar level by preventing the factorial operator from consuming `!` when followed by `=`.

Both target the same root cause (DuckDB's dual use of `!`) but from different angles. If #7742 gets merged first, happy to close this one — just wanted to offer an alternative fix that doesn't touch the tokenizer.

peterbud:
Closing as #7742 has been merged. Thanks nevertheless!

PR #7788, opened 2026-04-24
feat: Support function calls in MERGE USING clause

sebastien-irco:
## Description

The `MERGE ... USING` clause currently only accepts table references and
subqueries. Function calls like `UNNEST()` are not supported, causing parse
failures.

This is valid syntax in BigQuery (and other dialects) for merging against
array data:

```sql
MERGE dataset.target AS t
USING UNNEST(arr) AS s
ON t.id = s.id
WHEN MATCHED THEN
  UPDATE SET val = s.val
```

The workaround is wrapping in a subquery (`USING (SELECT * FROM UNNEST(...)) AS s`),
but the direct form should parse.

## Fix

Add `FunctionSegment` (with optional alias) as an additional option in the
USING clause's `OneOf` in `MergeStatementSegment`. This is consistent with
how `TableExpressionSegment` already accepts `FunctionSegment` for FROM
clauses.

The change is in `dialect_ansi.py` because function calls in table positions
are an ANSI-level concept, already present in `TableExpressionSegment`. If
maintainers prefer a BigQuery-specific override, happy to move it.

## Test cases

Added a `MERGE ... USING UNNEST(arr) AS alias` test case to the existing
BigQuery `merge_into` fixture.

All 27 MERGE tests pass across all dialects. Full BigQuery suite: 426 passed.

---
AI disclosure: Claude Code was used to assist with this contribution.

### DIFF
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -3114,6 +3114,10 @@ class MergeStatementSegment(BaseSegment):
                 ),
                 Ref("AliasExpressionSegment", optional=True),
             ),
+            Sequence(
+                Ref("FunctionSegment"),
+                Ref("AliasExpressionSegment", optional=True),
+            ),
         ),
         Dedent,
         Conditional(Indent, indented_using_on=True),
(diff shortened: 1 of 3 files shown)


keraion (approved):
LGTM. Thanks!

PR #7789, opened 2026-04-24
fix: Support compound statements in BigQuery WHILE body

sebastien-irco:
## Description

`WHILE` loops in BigQuery fail to parse when the body contains compound
statements like `BEGIN...END` or `IF...END IF`. The parser's `END WHILE`
terminator matches the inner `END` keyword prematurely.

```sql
-- Fails to parse
WHILE x > 0 DO
  BEGIN
    SET x = x - 1;
  EXCEPTION WHEN ERROR THEN
    SET x = 0;
  END;
END WHILE;
```

## Root cause

`WhileStatementsSegment` only accepts `StatementSegment` in its body,
while `ForInStatementsSegment` and `RepeatStatementsSegment` accept
`OneOf(StatementSegment, MultiStatementSegment)`.
`MultiStatementSegment` handles compound statements (`BEGIN...END`,
`IF...END IF`, `FOR...END FOR`, etc.), so omitting it causes any `END`
keyword inside the WHILE body to be misinterpreted as `END WHILE`.

## Fix

Add `MultiStatementSegment` as an alternative in `WhileStatementsSegment`,
matching the pattern already used by `ForInStatementsSegment` and
`RepeatStatementsSegment`.

## Test cases

Two test cases added to the existing `procedural_statements` fixture:
- `WHILE` with `BEGIN...EXCEPTION...END` block
- `WHILE` with `IF...END IF` block

All 426 BigQuery dialect tests pass.

---
AI disclosure: Claude Code was used to assist with this contribution.

### DIFF
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -967,7 +967,10 @@ class WhileStatementsSegment(BaseSegment):
     type = "while_statements"
     match_grammar = AnyNumberOf(
         Sequence(
-            Ref("StatementSegment"),
+            OneOf(
+                Ref("StatementSegment"),
+                Ref("MultiStatementSegment"),
+            ),
             Ref("DelimiterGrammar"),
         ),
         terminators=[Sequence("END", "WHILE")],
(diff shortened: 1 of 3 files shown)


keraion (approved):
LGTM. Thanks again!

PR #7810, opened 2026-05-02
fix(tsql): allow Unicode letters in @parameter names

### LINKED ISSUE #7806: tsql: Unicode (e.g. Cyrillic) characters in @parameter names break the parser
### Search before asking

### What Happened

In the `tsql` dialect, `@parameter` names that contain Unicode letters
(Cyrillic, Greek, accented Latin, CJK, …) are not parsed. SQLFluff fails
the entire enclosing batch as one big unparsable section starting at line
1, even though the parameter is a perfectly valid T-SQL identifier in
SQL Server.

T-SQL allows Unicode letters in regular identifiers (and therefore in
parameter names), since SQL Server uses Unicode identifier rules:

> A regular identifier ... must comply with rules from the Unicode 3.2
> standard. ... Letters, as defined in the Unicode Standard 3.2.

Reference (identifier rules):
https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers

Reference (parameter name rules — same as identifier rules):
https://learn.microsoft.com/en-us/sql/t-sql/statements/declare-local-variable-transact-sql

### Root cause (from reading the source) by claude code

In `src/sqlfluff/dialects/dialect_tsql.py` (line ~578):

```python
    r"@(?!@)[A-Za-z0-9_@$#]+", CodeSegment, type="parameter"
),
```

The character class `[A-Za-z0-9_@$#]+` is ASCII-only and so does not match
any Unicode letter that T-SQL identifier rules permit.

Suggested fix: broaden the regex to allow Unicode letters/digits/marks. A
conservative replacement that still excludes whitespace and operators
would be something like:

```python
    r"@(?!@)[\w@$#]+",         # \w is Unicode-aware in Python re
    CodeSegment, type="parameter"
),
```

…or, if a stricter character set is preferred, an explicit Unicode class
matching SQL Server's identifier rules (`L`, `Mn`, `Mc`, `Nd`, plus `_`,
`@`, `$`, `#`):

```python
    r"@(?!@)[\p{L}\p{Mn}\p{Mc}\p{Nd}_@$#]+",
    CodeSegment, type="parameter"
),
```

(`regex` package or `re` with the appropriate flag, depending on the
project's preferred dependency).

A YAML test fixture under
`test/fixtures/dialects/tsql/unicode_parameter_names.{sql,yml}` would
cover regression.

The same issue likely affects `NakedIdentifierSegment` and other
identifier-matching grammars in this dialect — worth a once-over while
fixing parameters.

### Expected Behaviour

Parameter names containing Unicode letters should parse cleanly. For
example:

```sql
DECLARE @Дата DATETIME2
```

…should parse

### Observed Behaviour

```
$ cat repro.sql
DECLARE @Дата DATETIME2

$ sqlfluff parse --dialect tsql repro.sql
==== parsing violations ====
L:   1 | P:   1 |  PRS | Line 1, Position 1: Found unparsable section: 'CREATE FUNCTION
                       | [dbo].[fn_test]\n(\n...'
```

### How to reproduce

1. Save the snippet above as `repro.sql` (UTF-8 encoding).
2. Run `sqlfluff parse --dialect tsql repro.sql`.
3. Observe the unparsable section at line 1.
4. Replace `@Дата` with `@Date` — the file parses with no PRS.

### Dialect

tsql

### Version

4.1.0
Python 3.14.4 (Windows)

### Configuration

```
[sqlfluff]
dialect = tsql
```

### Code of Conduct

Booyaka101:

Closes #7806.

T-SQL identifiers (and therefore `@parameter` names) [allow any Unicode 3.2 letter](https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers), but two regexes in the tsql dialect were still ASCII-only:

| Location | Pattern | Used for |
| --- | --- | --- |
| `dialect_tsql.py:217` (`atsign` lexer) | `r"[@][a-zA-Z0-9_@$#]+"` | tokenizing `@variable` |
| `dialect_tsql.py:579` (`ParameterNameSegment` parser) | `r"@(?!@)[A-Za-z0-9_@$#]+"` | matching the parameter token |

When `@Дата` arrived, the `atsign` lexer failed (no ASCII char after `@`), the patched `word` lexer (which already had `\p{L}` since #7262) caught `@Дата` as a generic `word`, and parsing then failed for the entire enclosing batch:

```sql
DECLARE @Дата DATETIME2;
-- Found unparsable section: 'DECLARE @Дата DATETIME2;...'
```

Fix is two characters: add `\p{L}` to both classes. Same idiom is already used elsewhere in this file (line 333 `word` lexer, line 524 `NakedIdentifierSegment`) and in vertica's `ParameterNameSegment` (line 521 — comment there literally says *"need to cover cases where non-ascii word is parameter"*).

Background: PR #7262 added Unicode support for the `word`/`NakedIdentifierSegment` path so unquoted Unicode column/table names work, but didn't carry the change over to the `@parameter` path. This PR finishes that work.

### Test fixture

Added `test/fixtures/dialects/tsql/unicode_parameter_names.sql` (+ auto-generated `.yml`) covering Cyrillic, CJK, German umlaut, Greek, and mixed Unicode+ASCII parameter names in `DECLARE`, `SELECT`, and `CREATE PROCEDURE` contexts. Pre-fix all of these are unparsable; post-fix they round-trip through `parse`/`broad_fix`/`parse_struct`.

### Verification

- `pytest -k tsql` → 510 passed, 0 failed (Windows / Python 3.11.9)
- `pytest -k "tsql and unicode_parameter"` → 3 passed (parse, broad_fix, parse_struct)
- The reporter's exact repro (`DECLARE @Дата DATETIME2`) now produces `parameter: '@Дата'` inside a `declare_segment` instead of an unparsable `word` token

### Out of scope

- The `hash_prefix` lexer (`#temp_table`, `##global_temp`) at line 241 has the same ASCII-only shape but is a different concern — happy to send a follow-up if you'd like it tracked separately.
- The `system_variable` parser (`@@SERVERNAME` at line 378) is intentionally ASCII-only by SQL Server spec.

No. The change strictly broadens what is accepted: every existing ASCII parameter name still matches, and the regex remains anchored at `@` and is still bounded by the same `[A-Za-z0-9_@$#…]+` shape (just with `\p{L}` added to that class).

### Pull Request checklist

- [x] Confirmed checks have run
- [x] Added/updated documentation if relevant — no doc change needed (regex is internal)
- [x] Added/updated tests if relevant — `unicode_parameter_names.{sql,yml}` fixture
- [x] Added new/updated rules pass linting — N/A (no new rules)
- [x] Added a clear one-line description in the [changelog](CHANGELOG.md) — handled by release-drafter PR template

Credit to @msergein for the report and root-cause pointer.

---
## Summary by cubic
Allow Unicode 3.2 letters in T‑SQL `@parameter` names by updating lexer and parser regexes. Fixes parsing failures for non-ASCII parameters like `DECLARE @Дата DATETIME2`.

- **Bug Fixes**
  - Added `\p{L}` to `atsign` lexer and `ParameterNameSegment` to accept Unicode letters.
  - Parameters now tokenize as `parameter` and parse correctly in DECLARE, SELECT, and PROCEDURE contexts.
  - Added `unicode_parameter_names` fixture covering Cyrillic, CJK, umlauts, Greek, and mixed names.

<sup>Written for commit 5600d3770c40afc8246d8c77d8a98bed2d9dc7bf. Summary will update on new commits.</sup>

### DIFF
--- a/src/sqlfluff/dialects/dialect_tsql.py
+++ b/src/sqlfluff/dialects/dialect_tsql.py
@@ -210,11 +210,11 @@ tsql_dialect.sets("currency_symbols").update(
 tsql_dialect.insert_lexer_matchers(
     [
         # According to Microsoft spec, subsequent characters in identifiers can include
-        # @, $, #, _ in addition to letters and numbers
+        # @, $, #, _ in addition to letters (Unicode 3.2) and numbers
         # https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers
         RegexLexer(
             "atsign",
-            r"[@][a-zA-Z0-9_@$#]+",
+            r"[@][a-zA-Z0-9_@$#\p{L}]+",
             CodeSegment,
         ),
         # Note: $ can only appear in subsequent positions of identifiers, not as prefix
@@ -576,7 +576,7 @@ tsql_dialect.replace(
         ],
     ),
     ParameterNameSegment=RegexParser(
-        r"@(?!@)[A-Za-z0-9_@$#]+", CodeSegment, type="parameter"
+        r"@(?!@)[A-Za-z0-9_@$#\p{L}]+", CodeSegment, type="parameter"
     ),
     FunctionParameterGrammar=Sequence(
         Ref("ParameterNameSegment", optional=True),
--- a/test/fixtures/dialects/tsql/unicode_parameter_names.sql
+++ b/test/fixtures/dialects/tsql/unicode_parameter_names.sql
@@ -1,0 +1,32 @@
+-- Per Microsoft's identifier rules, T-SQL identifier (and parameter) names
+-- may contain any Unicode 3.2 letter, not just ASCII A-Z.
+-- https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers
+
+-- Cyrillic
+DECLARE @Дата DATETIME2 = GETDATE();
+
+-- CJK
+DECLARE @正常 INT = 0;
+
+-- Accented Latin (German umlaut + sharp s)
+DECLARE @Größe NVARCHAR(50);
+
+-- Greek
+DECLARE @Παράμετρος INT;
+
+-- Mixed Unicode + ASCII
+DECLARE @user_Имя NVARCHAR(100);
+
+-- Used in SELECT and as procedure parameters
+SELECT
+    @Дата AS [report_date],
+    @正常 AS [status],
+    @Größe AS [size];
+
+CREATE PROCEDURE dbo.GetByCriteria
+    @Дата DATETIME2,
+    @user NVARCHAR(50)
+AS
+BEGIN
+    SELECT @Дата, @user;
+END;
--- a/test/fixtures/dialects/tsql/unicode_parameter_names.yml
+++ b/test/fixtures/dialects/tsql/unicode_parameter_names.yml
@@ -1,0 +1,133 @@
+# YML test files are auto-generated from SQL files and should not be edited by
+# hand. To help enforce this, the "hash" field in the file must match a hash
+# computed by SQLFluff when running the tests. Please run
+# `python test/generate_parse_fixture_yml.py`  to generate them after adding or
+# altering SQL files.
+_hash: c803dee1172dcccd14abdcfa6e77ff098a91d11536b621fbecaf0156a69f59b8
+file:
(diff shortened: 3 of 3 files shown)


peterbud (approved):
Thanks @Booyaka101

PR #7813, opened 2026-05-02
fix(parser): refine Python parser depth accounting

### LINKED ISSUE #7805: tsql: `Maximum parse depth exceeded (limit 255)` triggers on modest nested function calls
### Search before asking

### What Happened

Several short, perfectly ordinary T-SQL scalar UDFs in our codebase fail to
parse with `PRS | Maximum parse depth exceeded (limit 255)`, even though the
human-readable nesting depth is small (5–7 levels of `REPLACE` / `LTRIM` /
`RTRIM` / `CHAR`).

Because the limit fails the entire batch as a single top-level unparsable
section starting at L1,P1, this also masks any other diagnostic the user
might want from such files. There is no `--noqa` workaround for this
particular error since it is emitted before noqa-directives are applied.

Setting `max_parse_depth = 0` in `.sqlfluff` resolves all of them with no observed
side effects

### Expected Behaviour

A scalar function with ~6 nested function calls should parse without
hitting the parse-depth limit. The default limit of 255 should be reachable
only for genuinely pathological / DoS-style input, not normal
`REPLACE(REPLACE(...))` chains

### Observed Behaviour

```
$ cat repro.sql
CREATE FUNCTION [dbo].[fn_StringWithoutSpace]
(
    @string NVARCHAR(MAX)
)
RETURNS NVARCHAR(MAX)
WITH INLINE = OFF
AS
BEGIN
    RETURN (
        SELECT LTRIM(RTRIM(REPLACE(REPLACE(REPLACE(REPLACE(
            @string,
            CHAR(160), CHAR(32)),
            CHAR(32),  '()'),
            ')(',      ''),
            '()',      CHAR(32))))
    )
END

$ sqlfluff parse --dialect tsql repro.sql
==== parsing violations ====
L:   1 | P:   1 |  PRS | Maximum parse depth exceeded (limit 255). This may indicate deeply
                       | nested SQL or a malicious input.
```

Visible nesting:
`LTRIM( RTRIM( REPLACE( REPLACE( REPLACE( REPLACE( ... ) ) ) ) ) )` — 6
levels — produces a parse-depth value of 256+.

Setting `max_parse_depth = 0` (disable check) makes the file parse cleanly
with no further violations, confirming that the grammar itself can handle
the input — only the depth counter trips.

### How to reproduce

1. Save the snippet above as `repro.sql`.
2. Run `sqlfluff parse --dialect tsql repro.sql`.
3. Observe the `Maximum parse depth exceeded (limit 255)` PRS at L1,P1.
4. Add `max_parse_depth = 0` to `.sqlfluff`
   under `[sqlfluff]`. Re-run — the file parses with no PRS.

### Dialect

tsql

### Version

4.1.0
Python 3.14.4 (Windows)

### Configuration

[sqlfluff]
dialect = tsql

### Code of Conduct

peterbud:
Fix a false positive `max_parse_depth` failure in the Python parser by excluding helper-only match wrappers from depth accounting.

Fixes #7805

**What changed**
* Added `track_parse_depth` to `ParseContext.deeper_match()`
* Stopped charging depth for helper wrappers like `Ref`, `Sequence`, `AnyNumberOf`, `Delimited`, and greedy terminator matching
* Added test fixtures for the new behavior and the reported T-SQL case

**Note**: Python parser only. Rust parser unchanged.

### Pull Request checklist

  - Other.

---
## Summary by cubic
Fixes false positive `max_parse_depth` errors in the Python parser by not counting helper-only match wrappers toward depth. Prevents unexpected depth-limit failures (e.g., in T‑SQL). Rust parser remains unchanged.

- **Bug Fixes**
  - Added `track_parse_depth` to `ParseContext.deeper_match()` to control depth accounting.
  - Stopped charging depth for helper wrappers (refs, sequences, delimited groups) and greedy terminator matching.
  - Added tests for the new behavior, including the reported T‑SQL case.

<sup>Written for commit 0b67b7daa7fb150fcdc5c6954dfa38a362f636f1. Summary will update on new commits.</sup>

### DIFF
--- a/src/sqlfluff/core/parser/context.py
+++ b/src/sqlfluff/core/parser/context.py
@@ -216,6 +216,7 @@ class ParseContext:
         clear_terminators: bool = False,
         push_terminators: Optional[Sequence["Matchable"]] = None,
         track_progress: Optional[bool] = None,
+        track_parse_depth: bool = True,
     ) -> Iterator["ParseContext"]:
         """Increment match depth.
 
@@ -234,10 +235,15 @@ class ParseContext:
                 tracking for deeper matches. This avoids having the linting
                 progress bar jump forward when performing greedy matches on
                 terminators.
+            track_parse_depth (:obj:`bool`, optional): Whether this context
+                should contribute to the `max_parse_depth` guard. Helper
+                wrappers used only for logging or dispatch can disable this
+                so the depth limit tracks real recursive descent instead.
         """
         self._match_stack.append(self.match_segment)
         self.match_segment = name
-        self.match_depth += 1
+        depth_delta = 1 if track_parse_depth else 0
+        self.match_depth += depth_delta
         if self.max_parse_depth > 0 and self.match_depth > self.max_parse_depth:
             raise SQLParseError(
                 f"Maximum parse depth exceeded (limit {self.max_parse_depth}). "
@@ -256,7 +262,7 @@ class ParseContext:
             self._reset_terminators(
                 _append, _terms, clear_terminators=clear_terminators
             )
-            self.match_depth -= 1
+            self.match_depth -= depth_delta
             # Reset back to old name
             self.match_segment = self._match_stack.pop()
             # Reset back to old progress tracking.
--- a/src/sqlfluff/core/parser/grammar/anyof.py
+++ b/src/sqlfluff/core/parser/grammar/anyof.py
@@ -136,7 +136,8 @@ class AnyNumberOf(BaseGrammar):
         """
         if self.exclude:
             with parse_context.deeper_match(
-                name=self.__class__.__name__ + "-Exclude"
+                name=self.__class__.__name__ + "-Exclude",
+                track_parse_depth=False,
             ) as ctx:
                 if self.exclude.match(segments, idx, ctx):
                     return MatchResult.empty_at(idx)
@@ -192,6 +193,7 @@ class AnyNumberOf(BaseGrammar):
                 name=self.__class__.__name__,
                 clear_terminators=self.reset_terminators,
                 push_terminators=self.terminators,
+                track_parse_depth=False,
             ) as ctx:
                 match, matched_option = longest_match(
                     # TODO: Resolve re-slice limit hack
--- a/src/sqlfluff/core/parser/grammar/base.py
+++ b/src/sqlfluff/core/parser/grammar/base.py
@@ -413,6 +413,7 @@ class Ref(BaseGrammar):
                 name=self._ref + "-Exclude",
                 clear_terminators=self.reset_terminators,
                 push_terminators=self.terminators,
+                track_parse_depth=False,
             ) as ctx:
                 if self.exclude.match(segments, idx, ctx):
                     return MatchResult.empty_at(idx)
@@ -423,6 +424,7 @@ class Ref(BaseGrammar):
             name=self._ref,
             clear_terminators=self.reset_terminators,
             push_terminators=self.terminators,
+            track_parse_depth=False,
         ) as ctx:
             return elem.match(segments, idx, parse_context)
 
--- a/src/sqlfluff/core/parser/grammar/delimited.py
+++ b/src/sqlfluff/core/parser/grammar/delimited.py
@@ -128,7 +128,9 @@ class Delimited(OneOf):
                 break
 
             # Check whether there is a terminator before checking for content
-            with parse_context.deeper_match(name="Delimited-Term") as ctx:
+            with parse_context.deeper_match(
+                name="Delimited-Term", track_parse_depth=False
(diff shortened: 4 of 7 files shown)


alanmcruickshank:
@copilot can you compare and contrast this with how depth accounting is handled in rust?

peterbud:
> @copilot can you compare and contrast this with how depth accounting is handled in rust?

Hi @alanmcruickshank , let me try to contrast it

In Python, the PR changes `context.py:219-265` so a deeper-match scope can opt out of consuming parse-depth budget. That matters because the Python parser uses one shared mutable counter, and before this PR it charged that counter for helper-only wrappers as well as real descent. The new call sites in `base.py:416-427`, `anyof.py:140-196`, `delimited.py:132-150`, and `sequence.py:236-558` explicitly mark wrappers like Ref, Sequence, Delimited, bracket helpers, and greedy terminator checks as non-depth-consuming. Real descent still goes through normal counted scopes such as segment matching in `base.py:696`.

Rust is different in two important ways. First, it enforces max depth from the explicit frame stack inside the iterative parser, not from a shared context-manager counter: see `iterative.rs:58-124`. Real recursive descent is represented by child frames pushed onto that stack. Second, some helper probes in Rust do not add onto the current frame depth at all because they launch a separate subparse with a fresh local stack, such as exclude checks.

So the practical contrast is this: old Python overcounted because every wrapper scope incremented one shared depth counter; this PR fixes that selectively. Rust already avoids that exact problem because depth is tied to active parse frames, not logging or dispatch scopes. But Rust is not identical to the post-PR Python behavior either, because some helper work is isolated into separate subparses, which means Rust can undercount relative to a fully cumulative notion of nested helper activity. I would not expect the exact false-positive fixed by this PR to reproduce in Rust. The Rust side already passes the same configured limit through from `rust_parser.py:90-102`, but it applies that limit at the frame-stack level rather than at every helper wrapper.

Does that help?

keraion:
I have minimal concerns that this undoes some of protection against CVE-2026-46373.

That being said, I probably don't fully understand the attack vector of the CVE. This change basically allows us to hit a `RecursionError` at around 47 nested brackets instead of returning a parser error. I might be missing something about the CVE which makes this beyond a `RecursionError` and there being an actual stack overflow, but without a reproduction it's hard to say.

@alanmcruickshank do you have any additional information on the CVE that we can make an informed decision?

alanmcruickshank:
Thanks @keraion . MY main view is that depth and breath are counted, but I don't have strong views on exactly _how_ they are counted. If python and rust count in equivalent ways, then that's definitely a bonus, but not a strong requirement. I care more that big (potentially malicious) queries trigger the inbuilt limits before they trigger an OOM kill.

Have forwarded the steps to reproduce the vulnerabilities, could you take a look and make a sensible decision here.

keraion:
I've done some rough math on the depth counting. By removing the depth count at the wrapper-grammar levels, this PR introduces a lot of variation in how much native stack each counted level actually represents — which is what makes "did we retrigger the CVE?" hard to reason about. With the current accounting on main, match_depth tracks the native stack at a fairly uniform ~1:1.3 ratio (each unit of depth ≈ 1.3 native frames). Once you stop counting the wrappers, that ratio swings widely by construct, and there's no longer a single limit that reliably trips before the stack overflows.

The option that makes the most sense is to keep the existing accounting and just raise the default max_parse_depth — somewhere around 600. At the ~1.3× ratio, a 600 limit puts the native peak during matching at roughly 780 frames, leaving ~220 under Python's default recursion limit of 1000. That margin covers measurements plus the CLI/API/templater frames that sit below the parser in a real invocation. (Tree construction and crawlers recurse too, but they run as separate, shallower phases after matching returns. They aren't stacked on top of the parse peak, so they're not the binding constraint here.)

We should apply the same 600 limit to Rust. Rust counts depth against its own iterative frame stack and doesn't over-count function nesting the way the Python parser does and it has no #7805 false positive (the UDF parses clean even at 255), and its matcher being iterative means it can't overflow during matching. The one residual risk is applying the results back on the Python side, which recurses; that path's ceiling is ~20 nested subqueries at the default recursion limit, so anything up to ~1000 max depth is safe and 600 has comfortable margin. (For reference, it's around 2000 where the guard goes too lax and a RecursionError starts leaking through.) Keeping the two parsers aligned at 600 for now makes the most sense.

As far as resource limits (OOM in particular), this PR doesn't directly touch those, but I'll likely continue down that path to make sure we're handling memory limits appropriately.

peterbud:
Thanks @keraion  - so you suggestion is to raise the default limit to 255 -> 600, but keep the depth counting as is? I'm fine with that.

keraion:
@peterbud Yep, that's exactly what I'm suggesting.

PR #7818, opened 2026-05-07
fix: prevent placeholder templater from suppressing lint violations

### LINKED ISSUE #7817: bug: placeholder templater silently suppresses lint violations for rules with broad anchors
## Summary

When using the `placeholder` templater, lint rules with broad anchors (e.g. statement-level `SegmentSeekerCrawler`) silently produce **zero violations** for SQL that should fail. The same SQL linted with literal values correctly reports violations.

## Expected Behaviour

`sqlfluff lint` should report the same violations regardless of whether a value is written as a literal or substituted via a placeholder template variable.

## Observed Behaviour

Rules such as AM04 (ambiguous column count in `UNION`), AM07 (`UNION` vs `UNION ALL`), and ST11 (set operation without brackets) return **no violations** when the SQL contains placeholder substitutions like `{{schema}}`, even though the substituted values have no effect on the rule logic.

## Steps to Reproduce

```bash
# Create test SQL and config
echo "SELECT a, b FROM {{schema}}.t1 UNION SELECT c FROM {{schema}}.t2" > /tmp/test.sql
echo -e "[sqlfluff]\ndialect = ansi\ntemplater = placeholder\n[sqlfluff:templater:placeholder]\nparam_regex = \$\{[a-zA-Z_]+\}\nparam_style = dollar" > /tmp/.sqlfluff

# Lint - expect AM04/AM07 violations but get NONE
sqlfluff lint /tmp/test.sql --config /tmp/.sqlfluff
```

Compare with literal SQL (no placeholders):
```bash
echo "SELECT a, b FROM myschema.t1 UNION SELECT c FROM myschema.t2" > /tmp/test_literal.sql
sqlfluff lint /tmp/test_literal.sql --dialect ansi
# Correctly reports AM04, AM07 violations
```

## Root Cause

`Linter.remove_templated_errors()` discards any violation whose anchor position marker returns `is_literal() == False`. The placeholder templater creates `RawFileSlice` objects with `slice_type="templated"` for substitutions, causing `TemplatedFile.is_source_slice_literal()` to return `False` for any source span overlapping a placeholder. Rules with broad anchors (spanning an entire statement) inevitably overlap placeholder regions and get silently dropped.

## Affected Rules

Any rule using `SegmentSeekerCrawler` with broad segment types is affected. Confirmed affected rules include:

| Rule | Description | Confirmed |
|------|-------------|-----------|
| AL04 | Unique column aliases | Yes |
| AL05 | Unused CTE aliases | Yes |
| AL06 | Short alias length | Yes |
| AL07 | Avoid right joins | Yes |
| AL08 | Distinct used with parentheses | Yes |
| AL09 | Self-aliasing | Yes |
| AM04 | Ambiguous column count in set operations | Yes |
| AM07 | UNION vs UNION ALL | Yes |
| AM09 | ORDER BY in set operations | Yes |
| CV03 | Trailing commas | Yes |
| LT09 | SELECT targets on separate lines | Yes |
| LT10 | Inconsistent indent of SELECT targets | Yes |
| ST11 | Set operations without brackets | Yes |

## Dialect

All dialects using the placeholder templater are affected.

## Version

Tested against `main` branch (commit 39b5a60).

ben-duivenvoorden:
## Description

Fixes #7817.

The `placeholder` templater silently suppresses lint violations for any rule whose anchor spans a placeholder substitution region. This is because `remove_templated_errors()` discards violations where `is_literal()` returns `False`, and placeholder substitutions create `slice_type='templated'` raw slices that cause this check to fail.

## Fix

Two minimal changes:

1. **`placeholder.py`**: Tag placeholder substitution `RawFileSlice` objects with `tag='literal'` to mark them as semantically literal SQL (the substituted values are simple literals, not control flow).

2. **`linter.py`**: Add `_is_semantically_literal_templated()` static method that checks whether all non-literal raw slices overlapping a violation anchor have `tag='literal'`. If so, the violation is preserved rather than discarded.

This approach is centralised (no per-rule changes needed) and extensible to other templaters that may have similar semantics.

## Affected Rules

13 rules with broad anchors confirmed working after fix: AL04, AL05, AL06, AL07, AL08, AL09, AM04, AM07, AM09, CV03, LT09, LT10, ST11.

## Test Cases

Added YAML regression tests for AM04 (3 cases), AM07 (1 case), and ST11 (1 case) using placeholder templater config.

## Checklist

- [x] Tests pass (2,285 tests, 0 failures)
- [x] `ruff check` clean
- [x] `ruff format` clean
- [x] `mypy` clean
- [x] No user-facing API/config changes (internal fix only)

*This PR was developed with AI assistance (GitHub Copilot).*

---
## Summary by cubic
Fixes a bug where the `placeholder` templater hid lint violations when a rule’s anchor overlapped placeholder substitutions. Also fixes a crash with the Rust parser by exposing required slice methods.

- **Bug Fixes**
  - Tag placeholder substitution slices with `tag='literal'` in the `placeholder` templater to mark them as semantically literal SQL.
  - Update `remove_templated_errors()` to keep violations that only overlap templated slices tagged as literal via `_is_semantically_literal_templated()`.
  - Add regression tests for AM04, AM07, and ST11 using the `placeholder` templater.
  - Expose `end_source_idx()` and `source_slice()` on Rust `PyRawFileSlice` bindings to prevent `AttributeError` when the Rust parser is enabled.

<sup>Written for commit dbf61e937cc1f6a9358df2000dfa03c4ef471d51. Summary will update on new commits.</sup>

### DIFF
--- a/sqlfluffrs/sqlfluffrs_python/src/templater/fileslice.rs
+++ b/sqlfluffrs/sqlfluffrs_python/src/templater/fileslice.rs
@@ -69,6 +69,14 @@ impl PyRawFileSlice {
     pub fn tag(&self) -> Option<String> {
         self.0.tag.clone()
     }
+
+    pub fn end_source_idx(&self) -> usize {
+        self.0.end_source_idx()
+    }
+
+    pub fn source_slice(&self) -> PySlice {
+        PySlice(self.0.source_slice())
+    }
 }
 
 impl From<PyRawFileSlice> for RawFileSlice {
--- a/src/sqlfluff/core/linter/linter.py
+++ b/src/sqlfluff/core/linter/linter.py
@@ -333,6 +333,9 @@ class Linter:
                 if (
                     # Is it in a literal section?
                     e.segment.pos_marker.is_literal()
+                    # Is it in placeholder-tagged template slices that are
+                    # semantically literal SQL (e.g. ${catalog})?
+                    or Linter._is_semantically_literal_templated(e)
                     # Is it a rule that is designed to work on templated sections?
                     or e.rule.targets_templated
                 ):
@@ -342,6 +345,23 @@ class Linter:
                 # malformed "noqa" comment).
                 result.append(e)
         return result
+
+    @staticmethod
+    def _is_semantically_literal_templated(error: SQLLintError) -> bool:
+        """Whether all non-literal raw slices in the anchor are semantically literal."""
+        assert error.segment.pos_marker
+        source_slice = error.segment.pos_marker.source_slice
+        non_literal_slices = []
+        for raw_slice in error.segment.pos_marker.templated_file.raw_sliced:
+            if raw_slice.end_source_idx() <= source_slice.start:
+                continue
+            if raw_slice.source_idx >= source_slice.stop:
+                break
+            if raw_slice.slice_type != "literal":
+                non_literal_slices.append(raw_slice)
+        return bool(non_literal_slices) and all(
+            raw_slice.tag == "literal" for raw_slice in non_literal_slices
+        )
 
     @staticmethod
     def _report_conflicting_fixes_same_anchor(message: str) -> None:  # pragma: no cover
--- a/src/sqlfluff/core/templaters/placeholder.py
+++ b/src/sqlfluff/core/templaters/placeholder.py
@@ -200,6 +200,9 @@ class PlaceholderTemplater(RawTemplater):
                     raw=in_str[span[0] : span[1]],
                     slice_type="templated",
                     source_idx=span[0],
+                    # Tagged as semantically literal, allowing lint filtering
+                    # to keep violations that only overlap placeholder params.
+                    tag="literal",
                 )
             )
             out_str += replacement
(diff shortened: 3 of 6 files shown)


ben-duivenvoorden:
## CI Fix (commit dbf61e9)

The previous CI failures in the `-rust` test jobs were caused by a missing method on the Rust Python bindings.

**Root cause:** The Python code in this PR calls `end_source_idx()` on `RawFileSlice` objects. The pure-Python `RawFileSlice` dataclass has this method, but the Rust `RsRawFileSlice` (PyO3 bindings) did not expose it — resulting in `AttributeError` in all Rust-enabled CI jobs.

**Fix:** Added `end_source_idx()` and `source_slice()` methods to `PyRawFileSlice` in `sqlfluffrs/sqlfluffrs_python/src/templater/fileslice.rs`, delegating to the existing Rust implementations.

**Verified locally** with the Rust extension compiled via `maturin develop` — 325 templater + linter tests pass.

alanmcruickshank (approved):
Good catch. Looks good to me, I'll check again once CI has run.

PR #7832, opened 2026-05-11
Add Rule CV13: CREATE INDEX should use CONCURRENTLY (Postgres)

OmerDital:
## Summary

Adds a new convention rule `CV13` that flags `CREATE INDEX` statements without `CONCURRENTLY` in PostgreSQL.

`CREATE INDEX` without `CONCURRENTLY` acquires a lock that blocks writes for the duration of the index build. On large tables this can cause significant downtime.

## Rule details

- **Code**: `CV13`
- **Name**: `convention.create_index_concurrently`
- **Groups**: `all`, `convention`
- **Dialect**: PostgreSQL only (skips other dialects)
- **Exclusions**: `CREATE UNIQUE INDEX` is skipped (unique constraints often need transactional guarantees)

## Anti-pattern

```sql
CREATE INDEX idx_foo ON bar (tenant_id);
```

## Best practice

```sql
CREATE INDEX CONCURRENTLY idx_foo ON bar (tenant_id);
```

## Motivation

We had a production incident where a `CREATE INDEX` on a large table locked writes for several minutes during the index build. This rule helps catch this pattern in CI before it reaches production.

## Test plan

8 YAML test cases covering:
- `CREATE INDEX CONCURRENTLY` (pass)
- `CREATE INDEX CONCURRENTLY IF NOT EXISTS` (pass)
- `CREATE UNIQUE INDEX` without CONCURRENTLY (pass — excluded)
- `CREATE INDEX` without CONCURRENTLY (fail)
- `CREATE INDEX IF NOT EXISTS` without CONCURRENTLY (fail)
- `CREATE INDEX` on JSONB expression without CONCURRENTLY (fail)
- `CREATE INDEX CONCURRENTLY` on JSONB expression (pass)
- Non-postgres dialect (pass — skipped)

---
## Summary by cubic
Adds `PG01` (`postgres.excessive_locks`) as a Postgres-only rule to flag heavy-lock DDL and suggest `CONCURRENTLY`/`NOT VALID`. Moves the rule from `CV13`, registers the `sqlfluff_rules_postgres` plugin entry point, adds 19 tests (incl. ALTER TABLE early-return cases), and tidies ruff/coverage.

- **New Features**
  - Flags: `CREATE INDEX` → `CONCURRENTLY` (skips `UNIQUE`), `DROP INDEX` → `CONCURRENTLY`, `REINDEX` → `CONCURRENTLY`, `REFRESH MATERIALIZED VIEW` → `CONCURRENTLY`, `ALTER TABLE ... ADD CONSTRAINT FOREIGN KEY` → `NOT VALID`.
  - Handles `IF [NOT] EXISTS` and expression indexes; ignores non-Postgres dialects.

<sup>Written for commit 270d453338e21baf12dc7454db2f04a40e2784a2. Summary will update on new commits. <a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7832?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,6 +130,7 @@ sqlfluff_rules_structure = "sqlfluff.rules.structure"
 sqlfluff_rules_convention = "sqlfluff.rules.convention"
 sqlfluff_rules_jinja = "sqlfluff.rules.jinja"
 sqlfluff_rules_tsql = "sqlfluff.rules.tsql"
+sqlfluff_rules_postgres = "sqlfluff.rules.postgres"
 sqlfluff_rules_oracle = "sqlfluff.rules.oracle"
 
 
--- a/src/sqlfluff/rules/convention/CV13.py
+++ b/src/sqlfluff/rules/convention/CV13.py
@@ -122,7 +122,7 @@ class Rule_CV13(BaseRule):
 
     def _check_add_foreign_key(self, segment) -> Optional[LintResult]:
         action = segment.get_child("alter_table_action_segment")
-        if not action:
+        if not action:  # pragma: no cover
             return None
         constraint = action.get_child("table_constraint")
         if not constraint:
--- a/src/sqlfluff/rules/convention/CV13.py
+++ b/src/sqlfluff/rules/convention/CV13.py
@@ -86,16 +86,15 @@ class Rule_CV13(BaseRule):
 
         if seg_type == "create_index_statement":
             return self._check_create_index(segment)
-        elif seg_type in (
+        if seg_type in (
             "drop_index_statement",
             "reindex_statement_segment",
             "refresh_materialized_view_statement",
         ):
             return self._check_concurrently(segment)
-        elif seg_type == "alter_table_statement":
+        if seg_type == "alter_table_statement":
             return self._check_add_foreign_key(segment)
-
-        return None
+        return None  # pragma: no cover
 
     def _check_create_index(self, segment) -> Optional[LintResult]:
         if _has_keyword(segment.segments, "CONCURRENTLY"):
--- a/src/sqlfluff/rules/convention/CV13.py
+++ b/src/sqlfluff/rules/convention/CV13.py
@@ -1,0 +1,66 @@
+"""Implementation of Rule CV13."""
+
+from sqlfluff.core.rules import BaseRule, LintResult, RuleContext
+from sqlfluff.core.rules.crawlers import SegmentSeekerCrawler
+
+
+class Rule_CV13(BaseRule):
+    """``CREATE INDEX`` should use ``CONCURRENTLY`` to avoid locking the table.
+
+    ``CREATE INDEX`` without ``CONCURRENTLY`` acquires a lock that blocks
+    writes for the duration of the index build. On large tables this can
+    cause significant downtime.
+
+    This rule is disabled by default and must be explicitly enabled.
+    It only applies to the ``postgres`` dialect.
+
+    .. note::
+        ``CREATE UNIQUE INDEX`` statements are excluded because unique
+        constraints often need transactional guarantees.
+
+    **Anti-pattern**
+
+    Creating an index without ``CONCURRENTLY``.
+
+    .. code-block:: sql
+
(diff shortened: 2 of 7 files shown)


alanmcruickshank:
@OmerDital - thanks for putting this together. In general we don't accept rules which have a very narrow focus (see our [contributing file](https://github.com/sqlfluff/sqlfluff/blob/main/CONTRIBUTING.md#pull-request-acceptance-principles)), which this rule currently is - however I wonder if it could be expanded to something more widely useful. Are there other CREATE ... statements (beyond indices) where CONCURRENTLY (or similar options) would also be much safer additions?

@peterbud @WittierDinosaur - either of you two have views on this front?

peterbud:
> @peterbud @WittierDinosaur - either of you two have views on this front?

I don't have a view, as I don't use postgres and I never thought to create a rule for the more or less similar `ONLINE = ON`. I think it's a personal/architectural choice rather than a broadly applicable rule TBH.

OmerDital:
Thanks for the feedback @alanmcruickshank and @peterbud.

I understand the principle about broad applicability across dialects. The challenge here is that `CONCURRENTLY` is postgres-specific syntax — there's no direct equivalent in MySQL, SQLite, or ANSI SQL. T-SQL has `ONLINE = ON` but it's completely different syntax, so a single rule can't naturally span multiple dialects.

That said, within postgres this is a very common footgun — any `CREATE INDEX` on a production table without `CONCURRENTLY` blocks writes for the entire duration of the build. It's not an organizational preference, it's a well-documented operational risk that affects every postgres user running migrations.

I noticed you have a `tsql/` rule group (`TQ01`–`TQ03`) for T-SQL-specific rules. Would a `postgres/` rule group be acceptable here? That would let us scope this cleanly as `PG01` and leave room for other postgres-specific rules (e.g. `DROP INDEX CONCURRENTLY`, `REINDEX CONCURRENTLY`).

If not, happy to keep this as a plugin — just wanted to check since the dialect-specific precedent exists.

peterbud:
I'm leaning to see the positive impact of such rule specific to Postgres. I have looked it up and maybe it would make sense to make it a bit broader and include other excessive locks:
* `DROP INDEX CONCURRENTLY`
* `REFRESH MATERIALIZED VIEW CONCURRENTLY`
* `REINDEX [INDEX | TABLE | SCHEMA] CONCURRENTLY`
* Maybe also add `NOT VALID` for ADD CONSTRAINT ? Like `ADD CONSTRAINT <> FOREIGN KEY (...) REFERENCES .. NOT VALID;`

OmerDital:
@peterbud great suggestions! I've expanded the rule to cover all the patterns you mentioned. CV13 now flags:

- `CREATE INDEX` without `CONCURRENTLY` (still excludes `CREATE UNIQUE INDEX`)
- `DROP INDEX` without `CONCURRENTLY`
- `REINDEX` without `CONCURRENTLY`
- `REFRESH MATERIALIZED VIEW` without `CONCURRENTLY`
- `ALTER TABLE ... ADD CONSTRAINT ... FOREIGN KEY` without `NOT VALID`

Renamed the rule from `convention.create_index_concurrently` to `convention.excessive_locks` to reflect the broader scope. 19 test cases covering all patterns, all passing locally.

peterbud:
Hi @OmerDital - nice job so far.
Considering that all these are postgres specific rule, I'd suggest to move this to a specific postgres rule group, similar to t-sql, as you mentioned earlier.

OmerDital:
@peterbud done! Moved the rule to a dedicated `postgres/` rule group as `PG01`, mirroring the `tsql/` pattern:

- `src/sqlfluff/rules/postgres/__init__.py` — plugin bundle registration
- `src/sqlfluff/rules/postgres/PG01.py` — rule renamed from `Rule_CV13` to `Rule_PG01`, name changed to `postgres.excessive_locks`, groups changed to `("all", "postgres")`
- `test/fixtures/rules/std_rule_cases/PG01.yml` — test fixture renamed
- `pyproject.toml` — registered `sqlfluff_rules_postgres` entry point
- Removed CV13 from `convention/__init__.py`

peterbud (approved):
Thanks @OmerDital

PR #7840, opened 2026-05-13
fix(snowflake): accept pyformat placeholders in CREATE ICEBERG TABLE options

maxnoahb:

`CREATE ICEBERG TABLE` statements in Snowflake can be authored to be executed via Python DB drivers that perform pyformat-style parameter substitution (e.g. `%(external_volume)s`). Today these tokens are lexed as separate symbols (`%`, `(`, identifier, `)`, `s`) and the Snowflake parser then errors when they appear as values for options like `EXTERNAL_VOLUME` and `METADATA_FILE_PATH`.

Example that previously failed to parse:

```sql
CREATE ICEBERG TABLE IF NOT EXISTS schema.table
EXTERNAL_VOLUME=%(external_volume)s
CATALOG='catalog'
METADATA_FILE_PATH=%(metadata_file_path)s
;
```

This PR:

- Adds a `pyformat_placeholder` lexer matcher in the Snowflake dialect that captures `%(name)s` as a single token.
- Adds a corresponding `PyformatPlaceholderSegment` grammar segment.
- Updates `IcebergTableOptionsSegment` so `EXTERNAL_VOLUME`, `CATALOG`, `CATALOG_TABLE_NAME`, `CATALOG_NAMESPACE`, `METADATA_FILE_PATH`, and `BASE_LOCATION` accept either a quoted literal or a pyformat placeholder.
- Adds a fixture in `test/fixtures/dialects/snowflake/create_table.sql` covering the new syntax (with the regenerated `.yml`).

The new lexer matcher is scoped to the Snowflake dialect and only fires on the very specific `%(<identifier>)s` shape, so it should not affect parsing of other constructs. Pyformat placeholders are still only accepted in `IcebergTableOptionsSegment`; other contexts that previously rejected them continue to do so. Users who prefer to resolve placeholders before parsing can still configure the `placeholder` templater as before — this change simply makes the parser tolerant when that hasn't been done (e.g. when the configured templater is `jinja`, which leaves pyformat placeholders untouched).

### Pull Request checklist

- [x] Added appropriate documentation for the change. (Inline docstring/comments in the dialect file describe the new lexer matcher and segment.)

---
## Summary by cubic
Snowflake now accepts Python pyformat placeholders (e.g. %(external_volume)s) in CREATE ICEBERG TABLE option values. This fixes parse errors when SQL is executed via Python DB drivers.

- **Bug Fixes**
  - Added Snowflake-only `pyformat_placeholder` and `PyformatPlaceholderSegment`; options `EXTERNAL_VOLUME`, `CATALOG`, `CATALOG_TABLE_NAME`, `CATALOG_NAMESPACE`, `METADATA_FILE_PATH`, `BASE_LOCATION` now accept placeholders or quoted strings.
  - Added a fixture covering the new syntax; other contexts remain unchanged.

<sup>Written for commit 935240cc3813bd1ad35b4aa3773a78a916caaa17. Summary will update on new commits.</sup>


alanmcruickshank:
@maxnoahb - I'm not sure this should be handled at the parser layer, but instead at the templater layer, by using the placeholder templater: https://docs.sqlfluff.com/en/stable/configuration/templating/placeholder.html

That allows the use of any arbitrary style of placeholder (python or otherwise) across the whole dialect and not just in iceberg table creation. Does that fix your use case?

peterbud:
Closing due to inactivity

PR #7866, opened 2026-05-16
fix(snowflake): support SELECT * ILIKE '<pattern>' syntax

pif:

Snowflake supports `SELECT * ILIKE '<pattern>'` clause ([Snowflake docs](https://docs.snowflake.com/en/sql-reference/sql/select)).

This PR adds `IlikeClauseSegment` and updates `WildcardExpressionSegment` to support all possible
`SELECT` extensions as documented by Snowflake.

The order of operations in Snowflake according to the docs is:

```
[ ILIKE ... | EXCLUDE ... ] [ REPLACE ... ] [ RENAME ... ]
```

Notice, how `ILIKE` and `EXCLUDE` are [mutually exclusive](https://docs.snowflake.com/en/sql-reference/sql/select#selecting-all-columns).

None expected.

### Pull Request checklist

- [x] Parser test cases added as `test/fixtures/dialects/snowflake/select_ilike.{sql,yml}` (YML regenerated via `tox -e generate-fixture-yml`).

### DIFF
--- a/src/sqlfluff/dialects/dialect_snowflake.py
+++ b/src/sqlfluff/dialects/dialect_snowflake.py
@@ -2062,11 +2062,29 @@ class WildcardExpressionSegment(ansi.WildcardExpressionSegment):
 
     match_grammar = ansi.WildcardExpressionSegment.match_grammar.copy(
         insert=[
-            # Optional Exclude or Rename clause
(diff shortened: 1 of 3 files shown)


peterbud (approved):
Thanks and wecome @pif !

PR #7869, opened 2026-05-18
fix: add bash to Docker image for GitLab CI compatibility

### LINKED ISSUE #7868: Latest docker images incompatible with gitlab CI pipelines
### Search before asking

### What Happened

As of version 4.2.0 our gitlab pipelines using sqlfuff no longer function.
The issue is due to bash being removed from the sqfluff docker container here: https://github.com/sqlfluff/sqlfluff/pull/7752

Unfortunately gitlab expects any stage running in a docker container to be running `sh` and execute a provided script. Their instructions on how to do this are to overwrite any existing entrypoint and provide your own script.

Something like:
```
SQLFluff:
  stage: test
  image:
    name: sqlfluff/sqlfluff:4.2.1@sha256:0b1f131e3e9b4ac10bf45f1b2cb8680e67f74e63f8e6db9f0f0207dbbe3c71d2
    entrypoint:
      - ''
  script:
    - sqlfluff lint --processes 0 --disable-progress-bar
```

This works fine for 4.1.0, but as of 4.2.0 starts failing with:
```docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "sh": executable file not found in $PATH```

Is it possible for the base container image to change from a prod dhi to a dev dhi image?

I understand that including bash within the container does make it less secure. But I'm not sure that too many people would be using a linting tool in a production environment as opposed to a dev or CI environment so is that extra security actually required?

### Expected Behaviour

Our gitlab pipline to still work

### Observed Behaviour

```docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "sh": executable file not found in $PATH```

### How to reproduce

Run `docker run -it --entrypoint sh  sqlfluff/sqlfluff:4.1.0`  - runs fine
Run `docker run -it --entrypoint sh  sqlfluff/sqlfluff:4.2.1`  - complains about sh missing

### Dialect

Not applicable

### Version

4.2.0 and 4.2.1

### Configuration

Not applicable

### Code of Conduct

### PR DESCRIPTION
**Pull request by algojogacor**
## Summary

The removal of bash from the Docker image in #7752 broke GitLab CI pipelines which require `/bin/sh` to be available in the container to execute provided scripts. This PR adds bash back to the final runtime stage of the Docker image.

## Root Cause

PR #7752 switched to a minimal `dhi.io/python:3.14-debian13` base image that does not include bash. GitLab CI (and other CI platforms) execute stages inside containers using the shell (`/bin/sh`). Without a shell available, CI pipelines fail with errors like:

```
Unable to perform this action because the container has no shell
```

## What Was Fixed

Added a single `apt-get install bash` step to the final stage of the Dockerfile:
- Uses `--no-install-recommends` to keep the image minimal
- Cleans up apt cache after installation to avoid layer bloat
- Preserves existing ENTRYPOINT and CMD behavior

## Testing Done

- ✅ Dockerfile syntax is valid
- ✅ Change is minimal — only adding 1 RUN instruction
- ✅ No impact on existing ENTRYPOINT/CMD behavior
- ✅ bash is installed via apt (Debian 13 base)

## Related

Closes #7868

---
## Summary by cubic
Adds `bash` to the final Docker image so GitLab CI can run stages with `/bin/sh`. Restores failing pipelines caused by the minimal Debian base image. Closes #7868.

- **Bug Fixes**
  - Install `bash` in the runtime stage with `--no-install-recommends`.
  - Clean apt lists to keep the image small.
  - No changes to ENTRYPOINT/CMD.

<sup>Written for commit d7938382d013e713cdfc14d2d72acfb5a77f3a54. Summary will update on new commits. <a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7869?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,12 @@ LABEL org.opencontainers.image.title="sqlfluff" \
       org.opencontainers.image.source="https://github.com/sqlfluff/sqlfluff" \
       org.opencontainers.image.documentation="https://docs.sqlfluff.com/en/stable/"
 
+# Install bash for CI/CD compatibility.
+# GitLab CI and other CI platforms require /bin/sh to be available
+# in the container to execute provided scripts.
+RUN apt-get update && apt-get install -y --no-install-recommends bash \
+    && rm -rf /var/lib/apt/lists/*
+
 COPY --from=build /app/.venv /app/.venv
 ENV VIRTUAL_ENV=/app/.venv PATH=/app/.venv/bin:$PATH
 


peterbud:
Hi @algojogacor - thanks for opening this PR

Because we are using a hardened Debian base, `sh` is intentionally omitted to reduce the attack surface. If we absolutely need shell execution, we should install `dash` to provide a minimal `/bin/sh` rather than full `bash`, but I'm not an expert.

@alanmcruickshank @123Haynes - what do you think?

123Haynes:
@peterbud usually projects with hardened images would publish a second image with a shell and "-dev" in the tag.  
In sqlfluff we already use this in the builder image. "dhi.io/python:3.14-debian13-dev".  
So imho we should create a second Dockerfile with that as a baseimage and publish that as a Dockerimage with the "-dev" suffix.
 That way we keep the minmal attack surface by default while still providing a slightly less secure option with a shell that is clearly marked.

peterbud:
Closing in favor of #7872

PR #7906, opened 2026-06-02
fix: AL05 false positive on whole-row alias references (postgres)

### LINKED ISSUE #7039: Incorrect aliasing.unused AL05 error when aliasing is used inside a subquery
### Search before asking

### What Happened

I have a dbt model looking like the one below which on sqlfluff linting produces an aliasing.unused error for the alias t, while it is definitely being used inside a subquery.

### Expected Behaviour

aliasing.unused errors should not be detected

### Observed Behaviour

L:  16 | P:  56 | AL05 | Alias 't' is never used in SELECT statement.
                       | [aliasing.unused]

### How to reproduce

templater = dbt
```
{{ config(schema='stg_bth') }}

SELECT
    team,
    member,
    planfeld AS plan_field,
    TO_DATE((x).key, 'mon_YY') AS month,
    (x).value::float AS plan_value
FROM
    (
        SELECT
            team,
            member,
            planfeld,
            EACH(HSTORE(t)) AS x
        FROM member_plan AS t
    ) AS q
WHERE
    (x).key ~ '^[a-z]{3}_[0-9]{2}$'
    )
```

### Dialect

postgres

### Version

sqlfluff==3.4.2
sqlfluff-templater-dbt==3.4.2

### Configuration
```ini
[sqlfluff]
templater = dbt
dialect = postgres

# See https://docs.sqlfluff.com/en/stable/perma/rule_disabling.html
exclude_rules =
    ambiguous.column_count,
    ambiguous.column_references,
    ambiguous.distinct,
    layout.long_lines,
    layout.operators,
    references.consistent,
    references.from,
    references.keywords,
    references.qualification,
    references.quoting,
    references.special_chars,
    structure.column_order,
    structure.else_null,
    structure.simple_case,
    structure.subquery,
    structure.using,
    structure.unused_cte,
    structure.unused_join
# aliasing.unused (table aliases not used in the query) is a common rule
# aliasing.self_alias.column (prohibits aliases of form of "a AS a")
# structure.join_condition_order (enforces join conditions to be in a specific order)
# ambiguous.join (ensures that joins are not ambiguous)
# aliasing.expression (enforces that expressions are aliased with AS)

processes = -1

warnings = ST09

[sqlfluff:indentation]
indented_joins = False
indented_using_on = True
template_blocks_indent = False

[sqlfluff:templater]
unwrap_wrapped_queries = True

[sqlfluff:templater:jinja]
apply_dbt_builtins = True

[sqlfluff:templater:dbt]
project_dir = .
profiles_dir = profile/
profile = dwh-postgres
target = postgres
dbt_skip_compilation_error = True 

[sqlfluff.rules.capitalisation.keywords]

# Consistent usage of preferred "not equal to" comparison
preferred_not_equal_style = consistent
```

.sqlfluffignore
```ini
*target/
# dbt <1.0.0
dbt_modules/
# dbt >=1.0.0
dbt_packages/
dbt_internal_packages/
edr_target/
macros/
.dbt/
.github/
venv/
logs/
```

### Code of Conduct

Bortlesboat:
## Problem

Fixes #7039.

AL05 (`aliasing.unused`) reports a table alias as unused when it is referenced as a **bare, single-part identifier** rather than a qualified `alias.column`:

```sql
-- both flag `t` as "never used", incorrectly:
SELECT each(hstore(t)) AS x FROM member_plan AS t   -- whole-row value into a function
SELECT t                     FROM foo         AS t   -- whole-row select
```

In Postgres a table alias can be used as a composite/whole-row value (passed to a function, or selected directly). These are genuine uses of the alias, but they parse as a single-segment `column_reference`, so `extract_possible_references(level=TABLE)` returns nothing and `_analyze_table_aliases` never records the reference — the alias looks unused and AL05 even offers to delete it.

## Fix

In `_analyze_table_aliases`, when a reference yields no table-level parts, fall back to treating a lone identifier as a possible alias reference.

This is **gated by dialect** (`postgres`, `redshift`) via a new `_dialects_with_row_references` list and an `allow_row_references` flag, so standard-SQL behaviour is unchanged. In particular ANSI still treats `func(t)` as *not* a table use — preserving the existing `test_ansi_function_not_table_parameter` expectation. Keeping it opt-in avoids masking genuinely unused aliases in dialects without whole-row references.

## Behaviour

| query | dialect | before | after |
|-------|---------|--------|-------|
| `SELECT each(hstore(t)) ... FROM member_plan AS t` | postgres | ❌ flagged | ✅ ok |
| `SELECT t FROM foo AS t` | postgres | ❌ flagged | ✅ ok |
| `SELECT a FROM foo AS t` (truly unused) | postgres | flagged | flagged (unchanged) |
| `SELECT TO_JSON_STRING(t) FROM my_table AS t` | ansi | flagged | flagged (unchanged) |
| `SELECT zoo.a FROM foo AS zoo` | any | ok | ok (unchanged) |

## Tests

- Added `test_postgres_row_reference_in_function` and `test_postgres_whole_row_reference` to `AL05.yml`.
- Full AL05 / aliasing suite passes (135 yaml cases), including the unchanged ANSI case that distinguishes this behaviour.
- `ruff check` clean.

---
## Summary by cubic
Fixes AL05 false positives in Postgres/Redshift when a table alias is used as a whole-row value, including in subqueries. Bare identifiers like t in SELECT each(hstore(t)) or SELECT t FROM foo AS t are now treated as valid alias uses.

- **Bug Fixes**
  - When no table-level parts are found, treat a lone identifier as a possible alias reference in dialects with row references (postgres, redshift); ANSI unchanged.
  - Propagates this behavior into nested queries.
  - Adds tests for function parameter, whole-row select, subquery, and Redshift cases.

<sup>Written for commit c5ceac8a8349ef8c68a4de8f8921aa820d02eff7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7906?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/rules/aliasing/AL05.py
+++ b/src/sqlfluff/rules/aliasing/AL05.py
@@ -5,7 +5,7 @@ from dataclasses import dataclass, field
 from typing import cast
 
 from sqlfluff.core.dialects.common import AliasInfo
-from sqlfluff.core.parser.segments import BaseSegment, RawSegment
+from sqlfluff.core.parser.segments import BaseSegment
 from sqlfluff.core.rules import (
     BaseRule,
     EvalResultType,
@@ -318,7 +318,7 @@ class Rule_AL05(BaseRule):
                 cast(AL05Query, child), allow_row_references=allow_row_references
             )
 
-    def _resolve_and_mark_reference(self, query: AL05Query, ref: RawSegment) -> None:
+    def _resolve_and_mark_reference(self, query: AL05Query, ref: BaseSegment) -> None:
         # Does this query define the referenced alias?
         _ref = self._cs_str_id(ref)
         if any(_ref == self._cs_str_id(a.segment) for a in query.aliases if a.segment):
--- a/src/sqlfluff/rules/aliasing/AL05.py
+++ b/src/sqlfluff/rules/aliasing/AL05.py
@@ -69,6 +69,16 @@ class Rule_AL05(BaseRule):
         "postgres",
         "mariadb",
     ]
+    # Dialects where a bare table alias can be used as a whole-row/composite
+    # value (e.g. `SELECT to_json(t) FROM tbl AS t`). In these dialects such a
+    # reference is a legitimate use of the alias, so AL05 must not flag it as
+    # unused (issue #7039). In standard SQL this is not generally valid, so the
+    # behaviour stays opt-in per dialect to avoid masking genuinely unused
+    # aliases (see test_ansi_function_not_table_parameter).
+    _dialects_with_row_references = [
+        "postgres",
+        "redshift",
+    ]
     is_fix_compatible = True
 
     # config
@@ -87,7 +97,11 @@ class Rule_AL05(BaseRule):
         query = cast(
             AL05Query, AL05Query.from_segment(context.segment, dialect=context.dialect)
         )
-        self._analyze_table_aliases(query)
+        self._analyze_table_aliases(
+            query,
+            allow_row_references=context.dialect.name
+            in self._dialects_with_row_references,
+        )
 
         if context.dialect.name in ("redshift", "bigquery"):
             # Redshift supports un-nesting using aliases.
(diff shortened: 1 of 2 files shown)


peterbud:
Hi @Bortlesboat - thanks for opening this PR and welcome.

Pls fix the linting errors, and check out also the review suggestion.

peterbud (approved):
Thanks!

PR #7918, opened 2026-06-06
fix: ST10 false positive when RHS operand is part of a binary expression

### LINKED ISSUE #7910: tsql: False-positive for ST10 Redundant Constant Expression when using bitwise operators
### Search before asking

### What Happened

Rule ST10 checks if a constant comparison is taking place. However, it does not detect a bitwise operator when evaluating the value on both sides of the comparison.

### Expected Behaviour

sqlfluff should be able to detect that a bitwise operation is happening on one side of the comparison.

### Observed Behaviour

❯ sqlfluff lint --dialect tsql --rules ST10 test.sql
== [test.sql] FAIL
L:   1 | P:  58 | ST10 | Redundant constant expression.
                       | [structure.constant_expression]

### How to reproduce

`echo "SELECT BitFlagValue FROM Dbo.BitFlags WHERE BitFlagValue = BitFlagValue & @Mask" > test.sql && sqlfluff lint --dialect tsql --rules ST10 test.sql`

### Dialect

tsql

### Version

sqlfluff, version 4.2.1

### Configuration

N/A

### Code of Conduct

Booyaka101:

Fixes #7910.

ST10 (`structure.constant_expression`) treats the segments immediately left and right of the comparison operator as its operands. It already skips the case where the **left** operand is part of a larger binary expression (e.g. `num % 2 = 0`), but had no symmetric check for the **right** operand. So:

```sql
SELECT BitFlagValue FROM Dbo.BitFlags WHERE BitFlagValue = BitFlagValue & @Mask
```

was reported as a redundant constant expression — the rule compared `BitFlagValue` against `BitFlagValue` and ignored the trailing `& @Mask`, even though the two sides differ.

The fix adds the symmetric guard: skip when the segment following the RHS operand is a binary operator other than `AND`/`OR`, mirroring the existing left-hand check.

No. Genuine redundancy is still caught (`col = col`, `col = col OR x = y` both still flag), the existing left-hand exception (`num % 2 = 0`) is unchanged, and `AND`/`OR` after the RHS is still treated as a boundary so `... OR a = a` continues to be flagged. The full ST10 test suite passes (23 cases including the two added below).

### Pull Request checklist
- Included test cases: two `.yml` rule cases in `test/fixtures/rules/std_rule_cases/ST10.yml` (`test_other_operators_on_rhs`, `test_bitwise_operator_on_rhs_tsql`).
- Verified locally: `ruff check`, `ruff format --check`, and `mypy` all clean on the changed rule; ST10 yaml cases pass.

### AI assistance disclosure

Per the AI-assisted contributions policy: this change was drafted with AI assistance (Claude Code). The root cause was identified from the parse tree, and the fix, the regression tests, and the full verification (reproduction of the false positive, confirming no over-suppression, running the ST10 suite + ruff + mypy) were reviewed and validated manually by me.

---
## Summary by cubic
Fixes a false positive in ST10 where comparisons like `flags = flags & @Mask` were flagged as redundant when the RHS was part of a larger binary expression. Preserves existing behavior for `AND`/`OR` and left-hand exceptions. Fixes #7910.

- **Bug Fixes**
  - Add symmetric RHS guard: if a binary operator (other than `AND`/`OR`) follows the RHS, skip the rule, mirroring the existing LHS check.
  - Add tests: `flags = flags & 4` and T-SQL `BitFlagValue = BitFlagValue & @Mask`.

<sup>Written for commit 9d89e904c62e06541884e0937fb89f31a9b598cc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7918?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/rules/structure/ST10.py
+++ b/src/sqlfluff/rules/structure/ST10.py
@@ -74,15 +74,15 @@ class Rule_ST10(BaseRule):
                     (None, None),
                 )
 
-                rhs = next(
+                rhs_idx, rhs = next(
                     (
-                        subsegments[i]
+                        (i, subsegments[i])
                         for i in range(idx + 1, count_subsegments, 1)
                         if not subsegments[i].is_whitespace
                     ),
-                    None,
+                    (None, None),
                 )
-                if lhs is None or rhs is None or lhs_idx is None:
+                if lhs is None or rhs is None or lhs_idx is None or rhs_idx is None:
                     # Should be unreachable with correctly parsed tree
                     continue  # pragma: no cover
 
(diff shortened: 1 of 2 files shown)


keraion (approved):
Nice. Looks good and thanks for the PR!

PR #7927, opened 2026-06-09
Support Databricks ?:: try_cast operator

### LINKED ISSUE #7926: Databricks: Missing Syntax Support For `?::` as a Shorthand For `TRY_CAST`
### Search before asking

### What Happened

Unable to parse statements containing the "question double colon" (`?::`) shorthand syntax for `try_cast` using databricks dialect.

### Expected Behaviour

Section is parsed and no error is raised.

### Observed Behaviour

When parsing 

```
SELECT 1?::INT
```

as dialect databricks

We get 

L:   1 | P:   9 |  PRS | Line 1, Position 9: Found unparsable section: '?::INT'

Even though this is valid syntax in databricks as of version 15.3 and above

https://docs.databricks.com/aws/en/sql/language-manual/functions/questiondoublecolonsign

### How to reproduce

Create a file called "test.sql" containing this text:

```
SELECT 1?::INT
```

run `sqlfluff lint --dialect databricks`

observe 
L:   1 | P:   9 |  PRS | Line 1, Position 9: Found unparsable section: '?::INT'

### Dialect

Databricks 

### Version

sqlfluff, version 4.2.2

### Configuration

Empty

### Code of Conduct

Booyaka101:

Fixes #7926.

`?::` is the [try_cast shorthand](https://docs.databricks.com/aws/en/sql/language-manual/functions/questiondoublecolonsign) in Databricks (Databricks Runtime 15.3 and above). The docs describe it as casting "the value `expr` to the target data type `type` with error toleration", and a synonym for `try_cast` — i.e. the error-tolerating counterpart of `::`. It previously failed to parse:

```sql
SELECT 1?::INT
-- L: 1 | P: 9 | PRS | Found unparsable section: '?::INT'
```

The change mirrors the existing Vertica `::!` (`null_casting_operator`) handling:

- A `?::` lexer matcher, inserted before the `question` (`?`) matcher so `?::` lexes as a single token rather than `?` followed by `::`.
- A `TryCastOperatorSegment` parser and a `ShorthandCastSegment` override accepting `OneOf(CastOperatorSegment, TryCastOperatorSegment)`.
- A `[sqlfluff:layout:type:try_casting_operator]` config entry with the same `touch` spacing as `casting_operator`, so the layout rules treat `?::` like `::`.

No. Plain `::` casts and a standalone `?` are unaffected (the latter still does not parse as an expression on its own — pre-existing behaviour). Without the layout entry, LT01 flags `1?::INT` and `fix` rewrites it to `1 ?:: INT`; with it, `fix` leaves `1?::INT` unchanged.

Note: Vertica's `::!` (`null_casting_operator`) has the same missing layout entry — `fix` currently turns `'1'::!INT` into `'1' ::! INT`. That's out of scope here; happy to address it in a follow-up if useful.

### Pull Request checklist

- Included test cases:
  - `.sql`/`.yml` parser fixture in `test/fixtures/dialects/databricks/try_cast_operator.sql` (literal/string/identifier/alias casts, and a row mixing `::` with `?::`).
  - `.yml` rule cases in `test/fixtures/rules/std_rule_cases/LT01-excessive.yml` (`test_try_casting_operator_fix`, `test_try_casting_operator_pass`).
- Verified locally: reproduced the parse failure on the unpatched tree; `pytest test/dialects -k databricks` (165) and the full LT01 suite (198) pass; `ruff check`, `ruff format --check`, and `mypy` are clean on the changed dialect.

### AI assistance disclosure

Per the AI-assisted contributions policy: this change was drafted with AI assistance (Claude Code). The approach (mirroring the Vertica `::!` operator), the lexer ordering, the layout config, and the tests were reviewed and validated manually by me — including reproducing the parse failure on the unpatched tree, confirming `::`/standalone-`?` are unchanged, checking that `lint`/`fix` treat `?::` like `::`, and running the databricks dialect suite, the LT01 suite, `ruff`, `ruff format`, and `mypy` locally.

---
## Summary by cubic
Adds Databricks `?::` try_cast shorthand so it parses and formats like `::`. Also updates CV11 to avoid rewriting dialect-specific shorthand casts.

- **New Features**
  - Lexes `?::` as a single `try_casting_operator` (before `?`).
  - Parses via `TryCastOperatorSegment` and an overridden `ShorthandCastSegment` that accepts `::` or `?::`.
  - Adds `[sqlfluff:layout:type:try_casting_operator]` with touch spacing so LT01 treats `?::` like `::`.
  - Adds Databricks parser and LT01 fixtures, including mixed `::`/`?::`.

- **Bug Fixes**
  - CV11 only rewrites shorthand casts using `::`. Skips Databricks `?::` and Vertica `::!` to avoid semantic changes; adds CV11 tests.

<sup>Written for commit 260178bb86143445e25d86106db240791a795869. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7927?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/core/default_config.cfg
+++ b/src/sqlfluff/core/default_config.cfg
@@ -162,6 +162,10 @@ spacing_before = touch
 spacing_before = touch
 spacing_after = touch:inline
 
+[sqlfluff:layout:type:try_casting_operator]
+spacing_before = touch
+spacing_after = touch:inline
+
 [sqlfluff:layout:type:slice]
 spacing_before = touch
 spacing_after = touch
--- a/src/sqlfluff/dialects/dialect_databricks.py
+++ b/src/sqlfluff/dialects/dialect_databricks.py
@@ -66,6 +66,18 @@ databricks_dialect.insert_lexer_matchers(
 
 
 databricks_dialect.insert_lexer_matchers(
+    # ?:: is the try_cast shorthand (an error-tolerating cast). It must be matched
+    # before the "?" (question) matcher so that "?::" lexes as a single token
+    # rather than as "?" followed by "::".
+    # https://docs.databricks.com/aws/en/sql/language-manual/functions/questiondoublecolonsign
+    [
+        StringLexer("try_casting_operator", "?::", CodeSegment),
+    ],
+    before="question",
+)
+
+
+databricks_dialect.insert_lexer_matchers(
     # Databricks Pipeline Parameters:
     # https://docs.databricks.com/en/delta-live-tables/parameters.html
     # Must come before dollar_quote since both start with $
@@ -136,6 +148,9 @@ databricks_dialect.add(
         type="pipeline_parameter",
     ),
     RightArrowSegment=StringParser("=>", SymbolSegment, type="right_arrow"),
+    TryCastOperatorSegment=StringParser(
+        "?::", SymbolSegment, type="try_casting_operator"
+    ),
     # https://docs.databricks.com/en/sql/language-manual/sql-ref-principal.html
     PrincipalIdentifierSegment=OneOf(
         Ref("NakedIdentifierSegment"),
@@ -490,6 +505,30 @@ class WithinGroupClauseSegment(BaseSegment):
         "WITHIN",
         "GROUP",
         Bracketed(Ref("OrderByClauseSegment")),
+    )
+
+
+class ShorthandCastSegment(ansi.ShorthandCastSegment):
+    """A casting operation using '::' or '?::'.
+
+    '?::' is shorthand for try_cast (an error-tolerating cast).
+    https://docs.databricks.com/aws/en/sql/language-manual/functions/questiondoublecolonsign
+    """
+
+    match_grammar: Matchable = Sequence(
+        OneOf(
+            Ref("Expression_D_Grammar"),
+            Ref("CaseExpressionSegment"),
+        ),
+        AnyNumberOf(
+            Sequence(
+                OneOf(Ref("CastOperatorSegment"), Ref("TryCastOperatorSegment")),
+                Ref("DatatypeSegment"),
+                Ref("TimeZoneGrammar", optional=True),
+                allow_gaps=True,
+            ),
+            min_times=1,
+        ),
     )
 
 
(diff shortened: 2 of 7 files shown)


peterbud:
Hi @Booyaka101 - thanks for opening this PR.

While the parser improvements are clean, I think introducing a new operator would lead to a regression with rule CV11, if it is set to `preferred_type_casting_style = cast`. Eg. `SELECT CAST('20' AS INTEGER), '20'?::INTEGER` raises CV11, then SQLFluff warns that fixes were not applied because the generated fix would be unparsable. That means the PR adds parse support but leaves existing cast-aware behavior inconsistent and partially broken.

So as a minumum, CV11 should be updated and a Databricks-specific regression case should be added to the CV11 fixtures. 

The cleaner longer-term design would be to introduce a shared shorthand-cast operator abstraction so rules consume one semantic category instead of hard-coding operator token names.

Booyaka101:
Fixed in 260178b. Rather than special-casing `?::`, CV11 now only rewrites the standard `::` shorthand and skips any `cast_expression` whose operator isn't `casting_operator`. That covers `?::`, and also the same class of bug on Vertica's `::!` (null cast), which CV11 was silently rewriting to `cast('20'::! as INTEGER)`.

Added CV11 fixtures: Databricks `?::` (`preferred_type_casting_style = cast` and the `consistent` default), Vertica `::!`, and a control asserting plain `::` is still rewritten to CAST.

On the shared shorthand-cast abstraction — agree that's the cleaner long-term shape. This guard at least stops hard-coding the operator token names; it keys off the standard `::` only. Happy to follow up with the abstraction as a separate PR if you want to scope it, but kept it out of here to keep this change minimal.

peterbud (approved):
Thanks @Booyaka101  👌

Issue #7933, opened 2026-06-10
Rust CI: Fix clippy issues raised


keraion:
We run `cargo clippy` as informational-only in CI. We'll work to get these numbers down.

Issue #7931, opened 2026-06-10
Rust CI: Tests


keraion:
We should run cargo tests in CI

davidfierro:
Apologies for the stray commit references above — those came from a branch on my
fork while I was measuring this. There's no PR yet, and the commit no longer
references this issue, so no more should appear.

I'd like to pick this up, if it's still wanted.

## Current state

The `rust-lint` job runs `cargo test --workspace --lib`, so only the crates'
unit tests execute. The integration targets under `sqlfluffrs/tests/` — the
fixture and YAML parity suites — do get *compiled* in CI (the clippy step uses
`--all-targets`), but they are never *run*. The job's own comment says this is
deliberate, and that parity is covered by the python `-rust` suites.

## Evidence

I ran the full set against `main` (39a9d059d) on a 4-core Linux x86_64 box,
with the dialect tables generated via `utils/rustify.py build`:

```
cargo test --workspace --tests
→ 162 passed; 0 failed; 0 ignored
```

| target | tests |
|---|---:|
| `fixture_tests.rs` (every dialect, fixture + YAML-vs-Python) | 60 |
| `table_driven_small.rs` | 27 |
| `yaml_compare.rs` | 2 |
| 8 smaller regression targets | 12 |
| unit tests across the workspace | 61 |
| **total** | **162** |

So flipping the step to `--tests` today is green — it wouldn't land a backlog of
failures on anyone.

## Cost

Measured from a clean target dir:

| step | time |
|---|---:|
| `cargo test --workspace --lib` (what the job does today) | 68s |
| + compile/link the 11 integration binaries | 5s |
| + run the added tests | 68s |
| **added to the job** | **74s** |

The extra compile is small because every workspace crate — including the ~410k
generated lines of dialect tables — is already built as an rlib by the `--lib`
step; the integration targets only link against them. Almost all the added time
is execution, dominated by `fixture_tests.rs` sweeping every dialect.

Two deliberate choices: keep the step feature-free so it reuses that same
no-feature build, and use `--tests` rather than `--all-targets` (the latter
would also build and run the criterion bench targets). `rust-lint` has no
dependent jobs and isn't in `coverage_check`'s `needs`, so this doesn't extend
the pipeline's critical path.

Caveat on the numbers: they're from my machine, not a GitHub runner — same vCPU
count, but different hardware and no `Swatinem/rust-cache` in the picture. The
order of magnitude should hold; the exact figures won't.

## The honest caveat

[...]

PR #7942, opened 2026-06-11
ci: Add Rust fmt/clippy/unit-test job for sqlfluffrs

### LINKED ISSUE #7932: Rust CI: Linting
We should run linting for Rust in CI

keraion:

Adds a rust-lint CI job covering the sqlfluffrs workspace, which previously had no CI coverage.

The job:

- generates the Rust dialect tables (`utils/rustify.py build`, we need to actually build it, not just use the pre-built `.so`)
- enforces `cargo fmt --all --check`,
- runs `cargo clippy --workspace --all-targets` as advisory (continue-on-error, to be flipped to -D warnings once the existing warning surface is cleaned up),
- runs `cargo test --workspace --lib`.
- makes progress on #7933 
- closes #7932
- makes progress on #7931

Note:
Integration tests (sqlfluffrs/tests/*.rs) are currently excluded; cross-parser parity is already covered by the Python `pyXXX-rust` suites.

This introduces a new CI run.

### Pull Request checklist

  - Other.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -81,6 +81,66 @@ jobs:
     - name: Run the tests
       run: tox -e ${{ matrix.job }}
 
+  rust-lint:
+    # Rust formatting, lints and unit tests for the sqlfluffrs workspace.
+    # Integration tests (sqlfluffrs/tests/*.rs) are not run here; parity is
+    # covered by the python -rust suites below.
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Checkout code
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 1
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.14'
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+          cache-dependency-glob: |
+            requirements_dev.txt
+            pyproject.toml
+            constraints/*.txt
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: rustfmt, clippy
+
+      - name: Cache Rust dependencies
+        uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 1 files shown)


peterbud:
Hi @keraion  - few questions:
* would it make sense to separate this new job, like: adding the linting part to the existing `linting` job as a new matrix element, and keep the unit tests as a standalone job?
* I'm not sure how much time the rust unit tests take, but maybe we should run them only when the build is successful, similar to that we run only the python dialect tests only when the build is successful?
* Not in this PR, but later on we might need to introduce some filtering to run certain jobs only when the python/rust code has changed?

keraion:
@peterbud All fair questions
> would it make sense to separate this new job, like: adding the linting part to the existing linting job as a new matrix element, and keep the unit tests as a standalone job?

I think that makes sense. I would like to utilize caching (test build is different from release), so we'd want to make sure we run the build as part of main.

> I'm not sure how much time the rust unit tests take, but maybe we should run them only when the build is successful, similar to that we run only the python dialect tests only when the build is successful?

Unit tests: if from cache, ~20s to compile the sqlfluffrs_* bits and run. ~45s from nothing.
Clippy is a bit slower: ~60s from cache, ~75s from nothing.

> Not in this PR, but later on we might need to introduce some filtering to run certain jobs only when the python/rust code has changed?

I was already thinking about that. Python tests should still run with Rust changes, but Rust changes are fair to be independent from Python changes. Right now, these aren't required tests, but if they ever do become that we'd potentially have issues with bypassed tests.

keraion:
@peterbud 

> would it make sense to separate this new job, like: adding the linting part to the existing linting job as a new matrix element, and keep the unit tests as a standalone job?

Upon further review, these are a bit more connected that I'd would have hoped. We need to generate the dialect files to do any: `cargo fmt`, `cargo clippy`, and `cargo test`. Also, if we split out `clippy` and `test` we need to compile both of those independently.

Right now, I think we go ahead and merge as-is. As we add more unit tests we can reevaluate then.

peterbud (approved):

Issue #6733, opened 2025-03-15
TSQL false positive RF05 + AL05 on table-valued method alias; sqlfluff fix causes script corruption


exit81:
### Search before asking

- [x] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

The following select statement is valid TSQL using the table-valued method [nodes()](https://learn.microsoft.com/en-us/sql/t-sql/xml/nodes-method-xml-data-type?view=sql-server-ver16):
```
select r.c.value('column[1]', 'varchar(10)') as 'Value'
from @xml.nodes('/rows/row') as r(c);
```

 `sqlfluff lint` incorrectly raises RF05 (do not use special characters in identifiers) on the column expression and AL05 (alias is never used in SELECT statement) on the method expression. The identifier does not contain special characters, and the alias is used. `sqlfluff fix` will delete the alias, which results in invalid TSQL. Executing the "fixed" script will produce the following error message:
```
Msg 318, Level 15, State 0, Line 4
The table (and its columns) returned by a table-valued method need to be aliased.
```

Additionally, `sqlfluff lint` raises LT01 (expected single whitespace between naked identifier and start bracket) on the alias. Microsoft's docs do not have spaces in the aliases, e.g.: `Instructions.nodes('/root/Location') AS T2(Loc)`

### Expected Behaviour

`.` accessor is appropriately recognized and not considered a special character.
Usage of table-valued method alias is appropriately recognized.

### Observed Behaviour

`.` accessor is treated as a special character in the identifier.
Usage of table-valued method alias is not recognized.

### How to reproduce

1. Create `xml.sql`
```
-- xml.sql
declare @xml xml = '<rows><row><column>asdf</column></row></rows>';

select r.c.value('column[1]', 'varchar(10)') as 'Value'
from @xml.nodes('/rows/row') as r(c);
```

2. Perform `sqlfluff lint xml.sql` and see the spurious RF05, AL05 and LT01 violations.
```
== [xml.sql] FAIL
L:   4 | P:  49 | RF05 | Do not use special characters in identifiers.
                       | [references.special_chars]
L:   5 | P:  33 | AL05 | Alias 'r' is never used in SELECT statement.
                       | [aliasing.unused]
L:   5 | P:  34 | LT01 | Expected single whitespace between naked identifier and
                       | start bracket '('. [layout.spacing]
```

[...]

PR #7955, opened 2026-06-16
Don't flag required alias on T-SQL table-valued XML methods (AL05)

gaoflow:
## Summary

Fixes the AL05 (`aliasing.unused`) false positive and corrupting autofix reported in #6733.

In T-SQL, a table-valued XML method such as `@x.nodes('...')` must be given a correlation name when used in a `FROM` clause; without one SQL Server raises a syntax error. AL05 treated that alias as unused (it frequently is not referenced in the select list) and proposed deleting it, so `sqlfluff fix` turned valid SQL into invalid SQL:

```sql
-- before
SELECT 1 FROM @Xml.nodes('a') AS FN;

-- after `sqlfluff fix --rules AL05` (invalid: .nodes() now has no alias)
SELECT 1 FROM @Xml.nodes('a');
```

## Fix

`_is_alias_required` already enumerates the cases where a `FROM` alias must be kept (VALUES clauses in dialects that require them, derived tables, and T-SQL `OPENROWSET ... BULK`). This adds one more T-SQL case: when the `FROM` expression contains a `datatype_method` segment (the `.nodes()` style XML method), the alias is required and must not be reported as unused. Genuinely unused aliases on ordinary tables are still flagged, in T-SQL and every other dialect.

## Tests

Added `test_pass_tsql_nodes_method_requires_alias` to `test/fixtures/rules/std_rule_cases/AL05.yml`. It fails on `main` (AL05 flags and then deletes the required alias) and passes with this change. The full AL05 suite (138 cases) and the combined AL04/AL05 alias-rule tests (148 cases) pass locally.

## Note on the other symptoms in #6733

On current `main` the "Found unparsable section" parse error and the RF05 finding from the issue no longer reproduce for me (`@x.nodes('...') AS r(c)` parses cleanly), so this PR targets the remaining live problem: the AL05 corrupting autofix. Happy to adjust the scope or wording.

## AI assistance disclosure

I prepared this change with AI assistance under my direction. I reviewed and validated the root cause, the fix, and the test myself (confirming the red/green behaviour and that ordinary unused aliases are still reported), and I ran the rule test suite locally.

---
## Summary by cubic
Prevent AL05 (`aliasing.unused`) from deleting the required alias on T‑SQL XML rowset method `@x.nodes('...')`, which produced invalid SQL. Update `_is_alias_required` to require an alias only when the `FROM` expression contains a rowset `datatype_method` from an allowlist (currently `nodes`), and add T‑SQL tests for the official `nodes(...) AS Table(Column)` syntax and an unreferenced‑alias case.

<sup>Written for commit a7c9e34d7cdfac1ba068f94006ccfcae522ac517. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7955?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/rules/aliasing/AL05.py
+++ b/src/sqlfluff/rules/aliasing/AL05.py
@@ -69,6 +69,12 @@ class Rule_AL05(BaseRule):
         "postgres",
         "mariadb",
     ]
+    # T-SQL datatype methods that produce a rowset and therefore require a
+    # correlation name in the FROM clause. This is deliberately an allowlist:
+    # most datatype methods (`query`, `value`, spatial and `hierarchyid`
+    # methods, ...) are scalar-valued and are not table expressions, so only
+    # the rowset-producing `nodes` belongs here (issue #6733).
+    _tsql_rowset_datatype_methods = frozenset({"nodes"})
     # Dialects where a bare table alias can be used as a whole-row/composite
     # value (e.g. `SELECT to_json(t) FROM tbl AS t`). In these dialects such a
     # reference is a legitimate use of the alias, so AL05 must not flag it as
@@ -251,12 +257,18 @@ class Rule_AL05(BaseRule):
                         for seg in openrowset.recursive_crawl("keyword")
                     ):
                         return True
-                    # A table-valued XML method such as `@x.nodes('...')` must be
-                    # given a correlation name in T-SQL; removing the alias
-                    # produces invalid syntax, so the alias is required even when
-                    # it is not otherwise referenced (issue #6733).
-                    if next(segment.recursive_crawl("datatype_method"), None):
-                        return True
+                    # A rowset-producing datatype method such as
+                    # `@x.nodes('...')` must be given a correlation name in
+                    # T-SQL; removing the alias produces invalid syntax, so the
+                    # alias is required even when it is not otherwise referenced
+                    # (issue #6733).
+                    for method in segment.recursive_crawl("datatype_method"):
+                        name = method.get_child("datatype_method_name_identifier")
+                        if (
+                            name is not None
+                            and name.raw.lower() in self._tsql_rowset_datatype_methods
+                        ):
+                            return True
                 # Found a table expression. Does it have a VALUES clause?
                 if segment.get_child("values_clause"):
                     # Found a VALUES clause. Is this a dialect that requires
--- a/src/sqlfluff/rules/aliasing/AL05.py
+++ b/src/sqlfluff/rules/aliasing/AL05.py
@@ -251,6 +251,12 @@ class Rule_AL05(BaseRule):
                         for seg in openrowset.recursive_crawl("keyword")
                     ):
                         return True
+                    # A table-valued XML method such as `@x.nodes('...')` must be
+                    # given a correlation name in T-SQL; removing the alias
+                    # produces invalid syntax, so the alias is required even when
+                    # it is not otherwise referenced (issue #6733).
+                    if next(segment.recursive_crawl("datatype_method"), None):
+                        return True
                 # Found a table expression. Does it have a VALUES clause?
                 if segment.get_child("values_clause"):
                     # Found a VALUES clause. Is this a dialect that requires
--- a/test/fixtures/rules/std_rule_cases/AL05.yml
+++ b/test/fixtures/rules/std_rule_cases/AL05.yml
@@ -1314,10 +1314,22 @@ test_pass_postgres_wildcard_row_passed_to_function:
 test_pass_tsql_nodes_method_requires_alias:
   # https://github.com/sqlfluff/sqlfluff/issues/6733
   # A table-valued XML method (`.nodes()`) must be given a correlation name in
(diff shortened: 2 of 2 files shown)


peterbud:
Hi @gaoflow and thanks for opening this PR

The original issue with this SQL has no more AL05 violation, but it DOES still show RF05 violation (contrary to what the AI said):

```sql
select r.c.value('column[1]', 'varchar(10)') as 'Value'
from @xml.nodes('/rows/row') as r(c);
```


Somebody later added to that issue a second SQL, which indeed is generating AL05:

```sql
SELECT 1 FROM @Xml.nodes('a') AS FN;
```

Am I reading it that this PR won't fix #6733, just a small subset of it (AL05 for the second example), and leaves RF05 still present?

binggao1230:
Thanks @peterbud, your reading is correct.

This PR is deliberately scoped to the AL05 problem, which is the dangerous one: AL05 reports the `.nodes()` correlation name as unused and `sqlfluff fix` then deletes it, turning valid T-SQL into invalid SQL. The fix keys on the `datatype_method` segment, so it covers both the original `... as r(c)` example and the later `AS FN` one.

You are also right that RF05 is still present, and my note in the description was misleading. I checked a reduced snippet (`@x.nodes('...') AS r(c)`) that does not include the `as 'Value'` column alias, and that quoted alias is what still trips RF05 on the original statement. That is a separate rule with a separate cause, so I left it out to keep this change focused on the corrupting autofix.

So this PR does not fully close #6733 on its own; it removes the AL05 false positive and the script corruption. My suggestion would be to land this first, since it is the symptom that silently turns valid scripts invalid, and track RF05 (special characters on a single-quoted T-SQL column alias) separately. I am happy to open a follow-up PR for that RF05 case if you would like.

src/sqlfluff/rules/aliasing/AL05.py
```diff
@@ -251,6 +251,12 @@ def _is_alias_required(
                         for seg in openrowset.recursive_crawl("keyword")
                     ):
                         return True
+                    # A table-valued XML method such as `@x.nodes('...')` must be
+                    # given a correlation name in T-SQL; removing the alias
+                    # produces invalid syntax, so the alias is required even when
+                    # it is not otherwise referenced (issue #6733).
+                    if next(segment.recursive_crawl("datatype_method"), None):
```
peterbud: I think using `datatype_method` is too broad. This collection contains XML methods like `nodes`, `query`, `value`, plus `hierarchyid` and spatial methods like `GetAncestor` and `STArea` etc. 
So this PR would encode “alias is required” for methods that are not rowset-producing table expressions. I would narrow this to an explicit allowlist, currently just `nodes`, rather than using `datatype_method` as the trigger.

test/fixtures/rules/std_rule_cases/AL05.yml
```diff
@@ -1310,3 +1310,14 @@ test_pass_postgres_wildcard_row_passed_to_function:
...
     core:
       dialect: postgres
+
+test_pass_tsql_nodes_method_requires_alias:
+  # https://github.com/sqlfluff/sqlfluff/issues/6733
+  # A table-valued XML method (`.nodes()`) must be given a correlation name in
+  # T-SQL. The alias is syntactically required, so it must not be flagged as
+  # unused -- removing it (as the autofix did) produces invalid SQL.
+  pass_str: |
+    SELECT 1 FROM @Xml.nodes('a') AS FN;
```
peterbud: I think the official syntax for `nodes()` is containing the column as well, not just the table alias:
`nodes(XQuery) AS Table(Column) `

peterbud (changes requested):

binggao1230:
Thanks, both good points.

1. **Allowlist instead of `datatype_method`.** Agreed that the whole `datatype_methods` set is too broad. I replaced the trigger with an explicit allowlist, `_tsql_rowset_datatype_methods = frozenset({"nodes"})`, and the check now only fires when the method name in the table expression is in that set. So `query`, `value`, `hierarchyid`, the spatial methods, etc. are no longer treated as requiring an alias. It's a frozenset so adding another genuinely rowset-producing method later is a one-line change.

2. **`nodes(XQuery) AS Table(Column)` syntax.** Updated the fixture to the official form with a column name, using a realistic select that reads the column:

   ```sql
   SELECT Col.value('.', 'NVARCHAR(MAX)') AS V
   FROM @Xml.nodes('/a') AS Tbl(Col);
   ```

   and kept a second pass case `... AS Tbl(Col)` where neither the table nor the column is referenced, since that's the exact shape AL05's autofix used to corrupt.

All 139 AL05 yaml cases pass and ruff check/format are clean.

peterbud (approved):

PR #7972, opened 2026-06-18
Fix LT05 breaking SQL at literal/Jinja boundaries

### LINKED ISSUE #7971: LT05: _is_templated_safe_break incomplete — corrupts SQL with mixed literal+Jinja lines
## Description

The `_is_templated_safe_break` guard added in #7640 (released in 4.1.0) only catches the case where **both** adjacent segments at a break point are non-literal (within template expansion). It misses the case where **one side is literal SQL and the other is a rendered Jinja template** — which is the common pattern in dbt models.

## Environment

- sqlfluff: 4.2.2
- templater: dbt (dbt-bigquery 1.10)
- dialect: bigquery

## Observed behavior

When running `sqlfluff fix` on ~268 dbt model files in bulk (BigQuery dialect, dbt templater), we observed:

1. **Duplicated lines** — a SQL line containing a mix of literal SQL and Jinja macro calls was duplicated during LT05 line-breaking
2. **Truncated logic** — complex SQL expressions were split at a break point between a literal segment and a template segment, and trailing clauses were lost
3. **Silent semantic changes** — the fixed SQL compiled and ran without errors, but produced different (incorrect) results

All three files had long lines with mixed literal SQL + Jinja (e.g., `SELECT col1, {{ macro_call() }}, TIMESTAMP_MILLIS(context.time) AS ...`).

## Root cause analysis

The `_is_templated_safe_break` function at `sqlfluff/utils/reflow/reindent.py:2403` checks:

```python
# Unsafe only if BOTH sides are non-literal (within template expansion).
return before_literal or after_literal
```

This returns `True` (safe) when one side is literal and the other is templated. But inserting a line break at such a boundary can still cause corruption because the template expansion may span across what sqlfluff considers independent segments. The rendered template content can't always be cleanly split at arbitrary literal/non-literal boundaries.

## Expected behavior

`_is_templated_safe_break` should be more conservative — either:
1. Flag a break as unsafe if **either** adjacent segment is non-literal (not just both)
2. Or check whether the break point falls within a broader template span, not just the immediately adjacent segments

## Workaround

Exclude `layout.long_lines` from auto-fix when running on files that haven't been previously linted:
```
sqlfluff fix --exclude-rules layout.long_lines
```

## Related

- #7639 (original report, fixed by #7640 but fix is incomplete)
- #6098 (dbt templater producing incorrect SQL with duplicated macro lines)
- #5667 (related duplication issue)

ntantri:
## Description

The `_is_templated_safe_break` guard from #7640 only blocked line breaks where **both** adjacent segments were non-literal (within a template expansion). Breaks at literal/non-literal boundaries were still allowed, which can corrupt dbt models containing a mix of SQL and Jinja macros on long lines.

We observed three corruption patterns when running `sqlfluff fix` on ~268 dbt model files (BigQuery dialect, dbt templater, sqlfluff 4.2.2):

1. Duplicated lines
2. Truncated trailing clauses
3. Silent semantic changes (SQL compiles and runs but produces wrong results)

All affected files had long lines mixing literal SQL with Jinja macros (e.g., `SELECT col1, {{ macro() }}, TIMESTAMP_MILLIS(context.time)`).

## Fix

Tighten `_is_templated_safe_break` to block breaks where **either** side is non-literal, not just both. This is more conservative — some previously fixable long lines will now be reported as unfixable — but it prevents corruption at all template boundaries.

## Test

Added `test_fail_no_fix_mixed_literal_jinja` to verify that LT05 reports but does not fix lines with mixed literal SQL and Jinja content.

Fixes #7971

---
## Summary by cubic
Prevent `LT05` from inserting line breaks at SQL/Jinja boundaries to avoid corrupting dbt-templated files. Any break where either side is non-literal is now treated as unsafe, and mixed lines are reported but not auto-fixed.

- **Bug Fixes**
  - Tighten `_is_templated_safe_break` to block breaks when either adjacent segment is non-literal.
  - Prevents duplicated lines, truncated clauses, and silent semantic changes on long mixed SQL/Jinja lines.
  - Add `test_fail_no_fix_mixed_literal_jinja` to ensure `LT05` reports without fixing.
  - Fixes #7971.

<sup>Written for commit 53fe546abbe90c82d9766930d5451883efd9c060. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7972?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/utils/reflow/reindent.py
+++ b/src/sqlfluff/utils/reflow/reindent.py
@@ -2448,29 +2448,33 @@ def _fix_long_line_with_integer_targets(
 def _is_templated_safe_break(elements: ReflowSequenceType, e_idx: int) -> bool:
     """Check if inserting a line break at e_idx is safe for templated files.
 
-    A break point is unsafe if BOTH adjacent content segments are non-literal
-    (i.e., the break is within a template expansion). Inserting a line break
-    within rendered template content cannot be mapped back to the source file
-    correctly, which can lead to content duplication or corruption.
+    A break point is unsafe if EITHER adjacent content segment is non-literal
+    (i.e., at least one side is within a template expansion). Inserting a line
+    break at a boundary involving rendered template content cannot always be
+    mapped back to the source file correctly, which can lead to content
+    duplication or corruption.
+
+    The original guard (#7640) only blocked breaks where BOTH sides were
+    non-literal. However, breaks at literal/non-literal boundaries can also
+    corrupt dbt models with mixed SQL + Jinja (e.g. a long SELECT with
+    inline macro calls).
 
     See: https://github.com/sqlfluff/sqlfluff/issues/7639
+    See: https://github.com/sqlfluff/sqlfluff/issues/7971
     """
     # Check the segment before the break point.
-    before_literal = True
     if e_idx > 0 and elements[e_idx - 1].segments:
         seg_before = elements[e_idx - 1].segments[-1]
         if seg_before.pos_marker and not seg_before.pos_marker.is_literal():
-            before_literal = False
+            return False
 
     # Check the segment after the break point.
-    after_literal = True
     if e_idx + 1 < len(elements) and elements[e_idx + 1].segments:
         seg_after = elements[e_idx + 1].segments[0]
         if seg_after.pos_marker and not seg_after.pos_marker.is_literal():
-            after_literal = False
+            return False
 
-    # Unsafe only if BOTH sides are non-literal (within template expansion).
-    return before_literal or after_literal
+    return True
 
 
 def lint_line_length(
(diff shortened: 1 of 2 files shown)


keraion (changes requested):
Thanks for the PR. However, there is a problem with your test. The test will run under jinja and the `dbt_utils` macro will fail entirely.

I'm also unable to reproduce the issue from your test running under dbt. Can you please provide an MRE that can produce the problem you're seeing?

ntantri:
Closing this PR. The issue turned out to be a false negative on our side. The file corruption was caused by git merge conflict resolution (`-X ours` dropping new content after sqlfluff reformatting created line level divergence), not by LT05 or the `_is_templated_safe_break` guard. See the comment on #7971 for the full investigation. Sorry for the noise.

PR #7976, opened 2026-06-20
Add rule CV13: final SELECT of a CTE model should be `SELECT * FROM ...` (dbt convention)

### LINKED ISSUE #546: Last select statement in a file should be select * from
https://github.com/fishtown-analytics/corp/blob/master/dbt_coding_conventions.md#example-sql

Also from ['Why the Fishtown SQL style guide uses so many CTEs'](https://discourse.getdbt.com/t/why-the-fishtown-sql-style-guide-uses-so-many-ctes/1091):

> What’s the deal with the final CTE, isn’t that redundant?
> By now, hopefully you’ve bought into the fact that CTEs don’t affect query performance. But what’s with the last CTE, it is just so uneccesary!

> I thought so too when I started at Fishtown and had to adopt this style guide. Having this last CTE makes debugging so much easier. If something weird is happening in your query, it’s much easier to replace:

```
select * from final
```

> with

```
select * from customer_orders
```

youdie006:

Adds **CV13** (`convention.last_select_star`), a new rule in the dbt-coding-conventions bundle: when the last top-level statement of a file is a CTE-based model (`WITH ... SELECT`), its final passthrough should be `SELECT * FROM ...` rather than an explicit column list — the dbt community convention from #546.

The rule uses `RootOnlyCrawler` (like LT12) to inspect only the file's last statement, mirrors the `Rule_CV12` class structure, and follows @barrywhart's guidance to make it a **separate** rule (cf. AM04, the existing `SELECT *` rule) and @pwildenhain's low-impact scoping.

fixes #546

**Scope is intentionally narrowed to CTE-based models** (`WITH ... SELECT`). The broader reading of the issue title — flag *any* trailing non-`SELECT *` statement — also flagged unrelated ad-hoc queries already present in the test fixtures (e.g. `select column from table`), so I scoped it to the CTE/model case that matches the documented dbt convention. With this scoping the full rule/core/api suite passes unchanged (2567 passed, 0 failed). Happy to broaden if you'd prefer the wider interpretation. (TSQL wraps statements in a `batch` segment, so the rule safely no-ops there — noted limitation.)

### Pull Request checklist

- [x] Included test cases — `.yml` rule test cases in `test/fixtures/rules/std_rule_cases` (`CV13.yml`: 7 pass + 5 fail).
- [x] Added appropriate documentation — the rule docstring (anti-pattern + best-practice + SQL code blocks) feeds the auto-generated rules reference; docstring/registration tests pass.
- [x] Created GitHub issues for followups if appropriate — n/a.

---
## Summary by cubic
Adds rule CV13 (`convention.last_select_star`) to enforce the dbt convention: when a file ends with a CTE model, the final statement must be a pure passthrough `SELECT * FROM <cte>`. The rule is disabled by default to avoid surprising changes.

- Flags when the last top-level statement is `WITH ... SELECT` and the trailing select is not exactly `SELECT * FROM <cte>`; allows qualified `final.*`; rejects `DISTINCT`, `WHERE`, `GROUP BY`, `ORDER BY`, `LIMIT`, `QUALIFY`, and set expressions (e.g., `UNION`). Plain trailing selects and files ending in DML/DDL are out of scope.
- Opt-in via `[sqlfluff:rules:convention.last_select_star]` (`force_enable = True` to enable; default is `False`). Adds YAML tests (`CV13.yml`) and updates the rule docstring link to the current dbt style guide.

<sup>Written for commit 8a16d253691ee19877b4d0552bb956755e774fd2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7976?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


WittierDinosaur (dismissed):
Thanks for raising the PR. Personally I'm firmly against this convention, and makes the sql less easy to use, and grep through. So I'd want this rule to be in the tranche of rules that are disabled default - or at the very least have a switch so you can litigate for or against the wildcard

youdie006:
Thanks for the review @WittierDinosaur — that's fair. I've made CV13 **disabled by default**: it now uses the `force_enable` config flag (following the established `AL07`/`aliasing.forbid` pattern — `config_keywords` + `force_enable = False` in the default config), so it never runs unless a user explicitly opts in. That also keeps it out of the dbt plugin tests' default config (the failing checks were CV13 firing on their ephemeral CTE fixtures).

I also addressed the `cubic-dev-ai` finding: the rule now requires a *pure* passthrough — `SELECT * FROM final WHERE ...` / `GROUP BY` / `ORDER BY` / `LIMIT` / `SELECT DISTINCT *` are no longer treated as compliant.

So anyone who disagrees with the convention is unaffected by default, and teams that follow it can switch it on. Happy to adjust further.

src/sqlfluff/rules/convention/CV13.py
```diff
@@ -0,0 +1,178 @@
...
+from sqlfluff.core.parser import BaseSegment
+from sqlfluff.core.rules import BaseRule, LintResult, RuleContext
+from sqlfluff.core.rules.crawlers import RootOnlyCrawler
+
+
+class Rule_CV13(BaseRule):
+    """The final ``SELECT`` of a CTE model should be ``SELECT * FROM ...``.
+
+    This is a convention popularised by the dbt
+    `coding conventions <https://github.com/dbt-labs/corp/blob/main/dbt_style_guide.md>`_,
```
peterbud: This link seems to be not working - is there a published guide you can reference here?

youdie006:
Good catch @peterbud — the old `dbt-labs/corp/dbt_style_guide.md` link 404s. I've pointed it at the current dbt best-practices page (`how-we-style/0-how-we-style-our-dbt-projects`), which is the live home of that style guide. Pushed.

peterbud (approved):

PR #7982, opened 2026-06-21
perf(rust): add per-stage profiling to the Rust parse path

WittierDinosaur:
## Summary

`RustParser.parse()` is a fast Rust core wrapped in O(nodes) Python work on both sides, but `bench-compare` only ever measured the whole parse as one opaque number. This adds opt-in instrumentation that splits `parse()` into its four internal stages, so we can see where the time actually goes before investing in larger architectural changes (e.g. moving AST construction into Rust).

## Changes

**`src/sqlfluff/core/parser/rust_parser.py`**
- `set_profiling()` / `get_parse_profile()` + an `SQLFLUFF_RS_PROFILE` env var. When enabled, `parse()` records per-stage wall-clock time for:
  - `rust_core` — the Rust parse (`parse_match_result_from_tokens`)
  - `convert` — the recursive Python rebuild into a `MatchResult`
  - `apply` — building the `BaseSegment` tree (`MatchResult.apply`)
  - `apply_as_node` — building the (currently unused) `_rs_node` tree
- No-op / effectively zero overhead when disabled.

**`utils/benchmark_parsing.py`**
- New `--profile` flag (works with `--compare` and `--rust-only`). Collects the breakdown per timed iteration and prints an overall table plus small/medium/large buckets, with each stage's ms and % of parse, plus a `parse()` total row as a sanity check.

## Measured findings (ansi fixtures, release build)

`--compare`: **parsing is 64% faster with Rust** (1037ms → 371ms); lexing is near-parity (+4%).

Per-stage split of the Rust parse path (overall): `rust_core` ~42%, `convert` ~24%, `apply` ~30%, `apply_as_node` ~5%. So **the Python-side stages are the majority of parse() time**, with `apply` the single largest Python cost. There is also ~11% un-instrumented Python glue between stages.

This quantifies the prize for moving AST construction into Rust — the 64% speedup is achieved *despite* ~58% of the parse path still being Python.

Tooling only; no behavioural change when profiling is off.

---
## Summary by cubic
Adds opt-in per-stage profiling to the Rust parse path and a `--profile` benchmark mode to show where parse time goes. Timings now accumulate across rendered variants and are reset per top-level parse; disabled by default with zero overhead.

- **New Features**
  - `RustParser.parse()` can record stage timings when enabled via `SQLFLUFF_RS_PROFILE` or `set_profiling(True)`; read via `get_parse_profile()`. Profile accumulates across variants until `reset_parse_profile()`.
  - Stages: `rust_core`, `convert`, `apply`, `apply_as_node`.
  - `utils/benchmark_parsing.py --profile` reports per-stage breakdowns (works with `--compare` and `--rust-only`), prints overall and small/medium/large buckets with a `parse()` total sanity check; usage/docs updated in benchmark help and `sqlfluffrs/AGENTS.md`.

- **Bug Fixes**
  - Benchmark records every stage on every timed iteration (0.0 when absent) so per-stage means align with total parse time.
  - No-code/early-return parses don’t leak stale timings; profiles are scoped via `reset_parse_profile()`. Added tests and a missing type annotation for `_prof`.

<sup>Written for commit ad5ee0c42650377e689c0293e7820e99e2d2b24e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7982?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/core/parser/rust_parser.py
+++ b/src/sqlfluff/core/parser/rust_parser.py
@@ -58,11 +58,24 @@ def set_profiling(enabled: bool) -> None:
     _PROFILE_ENABLED = enabled
 
 
-def get_parse_profile() -> dict[str, float]:
-    """Return per-stage timings (seconds) from the most recent parse().
+def reset_parse_profile() -> None:
+    """Clear the accumulated per-stage profile.
 
-    Only populated when profiling is enabled (see set_profiling / the
-    SQLFLUFF_RS_PROFILE env var). Keys, in execution order:
+    parse() accumulates timings (across rendered variants), so callers reset
+    between top-level parses to scope the profile to a single source. The
+    benchmark calls this before each timed iteration.
+    """
+    _PARSE_PROFILE.clear()
+
+
+def get_parse_profile() -> dict[str, float]:
+    """Return per-stage timings (seconds) accumulated since the last reset.
+
+    parse() *accumulates* into this profile (summing across rendered variants,
+    e.g. branching Jinja), so it reflects all parse() calls since the last
+    reset_parse_profile() rather than just the final one. Only populated when
+    profiling is enabled (see set_profiling / the SQLFLUFF_RS_PROFILE env var).
+    Keys, in execution order:
         rust_core      - the Rust parse (parse_match_result_from_tokens)
         convert        - Python rebuild of the tree as a MatchResult
         apply          - building the BaseSegment tree (MatchResult.apply)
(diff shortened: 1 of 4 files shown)


src/sqlfluff/core/parser/rust_parser.py
```diff
@@ -140,6 +176,14 @@ def parse(
                 parse_context = ParseContext.from_config(config=self.config)
                 parse_context.seed_parse_nodes(len(segments))
 
+                # Per-stage profiling (no-op unless profiling is enabled).
+                # Clear up front so that an early return or an exception leaves
+                # an empty profile rather than stale data from a previous parse.
+                _prof: Optional[dict[str, float]] = {} if _PROFILE_ENABLED else None
+                _ts = 0.0
+                if _prof is not None:
+                    _PARSE_PROFILE.clear()
```
keraion: I think this clears per render variant. Not a problem for normal sql, but branching jinja only records the last variant.

WittierDinosaur:
Good catch — fixed in `ad5ee0c42`.

You're right: `parse()` runs once per rendered variant, so clearing `_PARSE_PROFILE` up front meant branching templates recorded only the last variant, which is also inconsistent with `time_dict["parsing"]` (the linter's total across variants).

Changed it to **accumulate** instead of clear-and-overwrite:
- `parse()` now sums each variant's per-stage timings into the profile.
- Added `reset_parse_profile()`; `benchmark_parsing.py` resets before each timed parse, so the profile is scoped to that source and summed across its variants — matching the iteration's parse time.
- A no-code parse (early return) contributes nothing, so there's no stale carry-over either.

Added a test covering accumulation across repeated parses + the reset/no-code semantics.

keraion (approved):
LGTM. Super helpful instrumentation.

keraion:
close/reopen because github is struggling with the PR dispatch

PR #7984, opened 2026-06-21
spike(rust): Rust-native lint rule detection (CP01 proof-of-concept)

WittierDinosaur:
> ⚠️ **Experimental.** Lays the foundation for Rust-accelerated lint rules and wires up the first one (CP01) end to end, **gated off by default** (`core.use_rust_rules=False`).

## What this is

The plumbing for running a lint rule's *detection* natively in Rust over the parser's arena, while keeping fixes (for now) on the Python side — with CP01 as the first worked example. The layering, bottom-up:

- **`sqlfluffrs_parser` — public arena read API.** `Arena`/`NodeId` and the read accessors are now `pub` (construction/mutation stay internal), re-exported from the crate root, plus `PyTree::with_arena()` so other crates can read the arena.
- **`sqlfluffrs_rules` — new crate (pure Rust).** `cp01::cp01_violations(&Arena, …)` runs CP01's whole detection loop over that read API and returns `(leaf_index, fixed_raw)`. Its PyO3 binding lives here too under an optional `python` feature (`register(m)`), so rule + binding are co-located and the detection stays unit-testable.
- **Root crate** depends on both and registers the binding (`sqlfluffrs.cp01_violations`).
- **Dispatch:** `BaseRule._eval_rust(context)` — a generic override hook (default `None`). `crawl()` consults it once when `core.use_rust_rules` is enabled *and* the parse produced an arena: a rule may return all its `LintResult`s from one Rust pass, post-processed identically to `_eval` results, then `crawl()` returns early. Rules without a Rust path, or parses with no arena, fall through to the Python crawl unchanged.
- **CP01** overrides `_eval_rust`: maps each `(leaf_index, fixed_raw)` to `raw_segments[idx]` and emits a standard `LintFix`. Falls back to Python for a non-keyword policy or a regex word-ignore.
- **Config guard:** `use_rust_rules` requires the Rust parser (it reads the parser's arena); enabling rules while disabling the parser raises `SQLFluffUserError` instead of silently doing nothing. `auto` degrades gracefully.

## Fix model (deliberate, for now)

Detection runs in Rust; the **fix is applied Python-side** — the rule returns data and the linter anchors a `LintFix` by leaf index (arena and Python `BaseSegment` uuids are *not* shared, so leaf index is the bridge). Rust-side (mutating) fixing waits on the arena's mutation milestone and is out of scope here.

## Status / scope

- Default **off**; nothing changes unless `use_rust_rules` is enabled.
- CP01 is the cheap guinea-pig that establishes the pattern. The next rule is `cpNN.rs` + an `_eval_rust` override — **no framework changes**.
- Performance: CP01 detection is ~9× the Python-over-BaseSegment path, but CP01 is sparse/trivial; the architecture is set by the expensive (reflow/layout) rules, where running the loop in Rust — and, later, fixing in Rust — is the real win.

## Parity & tests

- Identical fixes to stock CP01 across the dialect fixtures × all four policies (2,340 checks).
- Dispatch fix-parity through the real linter (`use_rust_rules` on == off) across policies/samples + `ignore_words`; a spy confirms the Rust path engages only when enabled; the regex-ignore case and CP03 (which inherits the hook but uses a different policy key) both fall back; the config guard is tested.
- No regression: existing CP01/CP02/CP03/LT01/AL01/RF01 yaml cases and the config tests pass with the default (Python) path unchanged.

## Open questions

- Rust-side fixing on the mutable arena (the next big lever for reflow-class rules).
- Whether/when `use_rust_rules` should default to `auto`.
- Porting a second, more representative rule (an AL/LT rule) to stress the arena read API.

---
## Summary by cubic
Enabled Rust-native CP01 (keyword capitalisation) detection behind `core.use_rust_rules`, running over the public Rust parse arena and returning `(leaf_index, fixed_raw)`; fixes still apply in Python with parity unchanged. Hardened dispatch so only CP01 uses the Rust path, errors surface as recoverable lint violations, and config is re-validated after inline overrides.

- **New Features**
  - Added `core.use_rust_rules` (auto/True/False; default False) and a `BaseRule._eval_rust()` hook; CP01 calls `sqlfluffrs.cp01_violations(...)`, maps `leaf_index -> raw_segments`, and emits standard `LintFix`es. Falls back when no arena or for configs not implemented natively (non-keyword policies, `ignore_words_regex`).
  - Exposed a public, read-only arena API in `sqlfluffrs_parser` (`Arena`, `NodeId`, accessors; `PyTree::with_arena()`), and moved CP01 detection into a new `sqlfluffrs_rules` crate with a thin binding `sqlfluffrs.cp01_violations(tree, policy, ignore_words=[], ignore_templated=False)` registered by the root. Added a guard in config that raises if `use_rust_rules=True` while `use_rust_parser=False` (`auto` degrades gracefully).

- **Bug Fixes**
  - Limited the Rust path to CP01 only (`self.name == "capitalisation.keywords"`) so CP02–CP05 are not hijacked; verified CP04 parity.
  - Wrapped `_eval_rust()` in the same exception handling as `_eval()`, turning Rust binding/mapping errors into recoverable `SQLLintError`s instead of aborting lint.
  - Re-ran the Rust config guard after inline config is processed to catch per-file overrides; kept mypy/ruff cleanups and CP03’s fallback to Python (different policy key).

<sup>Written for commit 52f0e6f109e06c4a3cec422139705bf870752d66. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7984?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/AGENTS.md
+++ b/sqlfluffrs/AGENTS.md
@@ -16,10 +16,16 @@ from 5.0.
 - **Lexer** — done; Rust tokenizes and returns `RsToken`s that Python wraps.
 - **Parser** — core done for **all dialects**, but **hybrid**: Rust returns a
   lightweight `MatchResult` (slices/classes/inserts); Python's
-  `MatchResult.apply()` still builds the `BaseSegment` AST. A parallel Rust node
-  tree (`_rs_node`) is built and cached for future Rust-side rules.
-- **Linting rules & fixing** — **not migrated**; 100% Python. The `_rs_node`
-  hook exists but currently has no consumers.
+  `MatchResult.apply()` still builds the `BaseSegment` AST. A parallel,
+  id-addressable Rust arena (`_rs_tree`, `RsTree`/`RsHandle`) is built and cached
+  as the substrate for Rust-side rules (read-only to Python today).
+- **Linting rules & fixing** — **not migrated**; 100% Python. One **experimental
+  prototype** exists: `RsTree.cp01_violations(...)` runs CP01's detection loop
+  natively over the arena (validated at parity with stock CP01), but it is **not
+  wired into rule dispatch** — the Python rule still runs. It currently lives in
+  the parser crate (`parser/rules_cp01.rs`) as scaffolding; a production home (a
+  `rules` module) and dispatch/gating are pending an arena public-API decision.
+  See `parser/rules_cp01.rs` and PR #7984.
 
 **Not a replacement**: the Rust components work alongside Python. When
 `sqlfluffrs` is unavailable, Python falls back transparently (auto mode).
--- a/sqlfluffrs/AGENTS.md
+++ b/sqlfluffrs/AGENTS.md
@@ -20,12 +20,13 @@ from 5.0.
   id-addressable Rust arena (`_rs_tree`, `RsTree`/`RsHandle`) is built and cached
   as the substrate for Rust-side rules (read-only to Python today).
 - **Linting rules & fixing** — **not migrated**; 100% Python. One **experimental
-  prototype** exists: `RsTree.cp01_violations(...)` runs CP01's detection loop
-  natively over the arena (validated at parity with stock CP01), but it is **not
-  wired into rule dispatch** — the Python rule still runs. It currently lives in
-  the parser crate (`parser/rules_cp01.rs`) as scaffolding; a production home (a
-  `rules` module) and dispatch/gating are pending an arena public-API decision.
-  See `parser/rules_cp01.rs` and PR #7984.
+  rule** exists: CP01 detection in the `sqlfluffrs_rules` crate, run over the
+  arena's public read API (`sqlfluffrs_parser::Arena`) and exposed as the
+  `sqlfluffrs.cp01_violations(tree, …)` function. Validated at parity with stock
+  CP01, but **not wired into rule dispatch** — the Python rule still runs, and
+  fixes are applied Python-side (anchored by leaf index). Dispatch/gating (e.g.
+  `use_rust_rules`) and Rust-side fixing are still to come. See
+  `sqlfluffrs_rules/` and PR #7984.
 
 **Not a replacement**: the Rust components work alongside Python. When
 `sqlfluffrs` is unavailable, Python falls back transparently (auto mode).
--- a/sqlfluffrs/AGENTS.md
+++ b/sqlfluffrs/AGENTS.md
@@ -19,14 +19,15 @@ from 5.0.
   `MatchResult.apply()` still builds the `BaseSegment` AST. A parallel,
   id-addressable Rust arena (`_rs_tree`, `RsTree`/`RsHandle`) is built and cached
   as the substrate for Rust-side rules (read-only to Python today).
-- **Linting rules & fixing** — **not migrated**; 100% Python. One **experimental
-  rule** exists: CP01 detection in the `sqlfluffrs_rules` crate, run over the
-  arena's public read API (`sqlfluffrs_parser::Arena`) and exposed as the
-  `sqlfluffrs.cp01_violations(tree, …)` function. Validated at parity with stock
-  CP01, but **not wired into rule dispatch** — the Python rule still runs, and
-  fixes are applied Python-side (anchored by leaf index). Dispatch/gating (e.g.
-  `use_rust_rules`) and Rust-side fixing are still to come. See
-  `sqlfluffrs_rules/` and PR #7984.
+- **Linting rules** — mostly Python; **one experimental rule is wired end to
+  end**. CP01 detection lives in the `sqlfluffrs_rules` crate, runs over the
+  arena's public read API (`sqlfluffrs_parser::Arena`), and is dispatched by the
+  linter via `BaseRule._eval_rust` when **`core.use_rust_rules`** is enabled
+  (default `False`; also requires the Rust parser's arena, else it falls back
+  per-rule to Python). Validated at parity with stock CP01. **Fixing is still
+  Python-side** — the Rust rule returns `(leaf_index, fixed_raw)` and the linter
+  anchors a `LintFix`; Rust-side (mutating) fixing waits on the arena mutation
+  milestone. See `sqlfluffrs_rules/`, `BaseRule._eval_rust`, and PR #7984.
 
 **Not a replacement**: the Rust components work alongside Python. When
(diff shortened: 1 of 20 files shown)


keraion:
So CP01 is a good starter rule that makes sense to port over. But, I think we need to stop for a minute here and think a little further out. The tree structure in Rust, the Node with children Nodes is immutable. When we eventually let Rust make state changes, we need a structure that's better at being modified, namely an arena.

I'm looking at dropping the `apply_as_node` from the Python side and having it build out an arena tree instead.

WittierDinosaur:
Fair point - I'm gonna leave this PR open for a while in case it's a useful reference point

WittierDinosaur:
Following up on your point above (immutable Node → arena) — the arena substrate from #7986 is exactly what this needed, thanks. I'd like to turn this spike into the first *production* Rust-accelerated rule on top of it, and want to align on the seam before building, since it's the "Rust-side linting" area you flagged.

**Proposed shape (stays strictly on the read side of the arena — no mutation, doesn't touch your fixing milestone):**

1. **Detection in Rust over `_rs_tree`.** A Rust entry point walks the arena (the whole CP01 detection loop — target selection, exclusions, `ignore_words`/templated skips via `pos_marker.is_literal()`, the consistent-policy inference, the casing) and returns a compact list of violations in **one** FFI crossing. This is the part the spike showed matters: detection-loop-in-Rust was ~17× vs. Python reading the tree node-by-node.
2. **Anchor + fix in Python.** Map each violation back to its Python `BaseSegment` and emit a normal `LintFix.replace`. Since #7997 unified the uuid counter across Python/Rust, I'd anchor by `uuid` (via `RsHandle.uuid` / `RsTree.node_by_uuid`) rather than leaf-index — robust to ordering. Fix *application* stays 100% Python; nothing structural/mutating, so no overlap with the arena-mutation milestone.
3. **Gated, with fallback.** A `use_rust_rules` gate mirroring `use_rust_parser` (auto/true/false), falling back to the stock Python rule. Parity proven by running the existing `CP01.yml` std-rule cases through both paths.

The one piece I think is genuinely yours to steer is the **dispatch/gating architecture** — how a Rust-accelerated rule is invoked and gated, and the Python-fallback contract — so I don't want to unilaterally bake that in. Two questions:

- Are you already mid-flight on rule dispatch? If so I'll slot into whatever you're building rather than propose a parallel design.
- Any preference on the gate (global `use_rust_rules` vs. per-rule) and on uuid- vs index-based anchoring?

Happy to do CP01 as the guinea pig to establish the pattern, then it generalises. (Keeping fixes Python-side for now; the structural/reflow rules that actually need Rust-side mutation wait for your milestone.)

WittierDinosaur:
Built the prototype on top of the arena to pressure-test the approach: `perf/rust-rule-cp01-arena` (commit d3f9ecf8d). CP01's whole detection loop now runs natively over `_rs_tree` via a read-only `RsTree.cp01_violations(policy, ignore_words, ignore_templated)` → `[(leaf_index, fixed_raw)]`, anchored Python-side. **Parity: identical fixes to stock CP01 across the dialect fixtures × all four policies.** It's explicitly marked a prototype (module docstring + AGENTS.md) and is **not** wired into dispatch — stock CP01 still runs. No arena mutation; nothing structural.

Two concrete findings for the design discussion, both of which I think are really yours to call since they're about the arena's contract:

**1. The rule wants a public arena read API.** The detection has to read `Arena`/`NodeId` + the accessors (`class_types`, `raw`, `pos_marker`, `children`, …), which are all `pub(crate)`. So today it can only live *inside* the parser crate (`parser/rules_cp01.rs`) — a rule in the parser crate, which is a layering smell. Moving rules to their proper home (a `rules` module/crate) needs the arena to expose a deliberate **public navigation API**. That's a substrate-interface decision I didn't want to make for you.

**2. uuid doesn't bridge arena ↔ Python.** I'd hoped to anchor fixes by `uuid` (since #7997 unified the counter, and `node_by_uuid` exists for exactly this). But empirically the arena and the Python `BaseSegment` tree carry **different uuids** for the same token — same *count*, different *values* — because `apply_as_tree` builds a fresh Node tree independently of the Python `MatchResult.apply()` construction. So I fell back to leaf-index anchoring (1:1 with `raw_segments`, incl. metas), which works. Flagging it because `node_by_uuid` is presumably meant to underpin fix-anchoring in your mutation milestone — if so, the arena would need to ingest the Python-segment uuids (or fixing standardizes on a different handle).

No rush — happy to take direction on the dispatch/gating shape and the arena API question whenever you've got bandwidth. The prototype's just there to prove the detection path and give us real parity numbers to build on.

WittierDinosaur:
Addressed the cubic findings in 52f0e6f10:

- **P1 (CP04 hijack):** valid and the important one — CP04 inherits `_eval_rust` and shares `capitalisation_policy`, so it would have run CP01's keyword detection and dropped literal violations. Guarded the Rust path to `self.name == "capitalisation.keywords"`; CP02–CP05 now fall back to Python. Added a CP04 parity test.
- **P2 (exception handling):** wrapped `_eval_rust()` with the same handler as `_eval()` — a Rust error is now a recoverable `SQLLintError`, not a whole-file abort. Test monkeypatches the binding to raise and asserts recovery.
- **P2 (inline config guard):** re-run `_verify_rust_config()` after `process_raw_file_for_config`, so an inline `-- sqlfluff:` change can't slip past it.
- **P3 ×2:** fixed the stale rules-crate doc and added `rust-version = "1.82"`.

**Deferring P2 (arena `NodeId` ownership):** real in principle — `NodeId` is an arena-local `u32` index with no cross-arena validation, so mixing IDs from two arenas would misread or panic. But the only consumer is the rules crate, which only ever uses IDs from the single arena it's handed inside one `with_arena` call — it never holds or mixes IDs across arenas. The robust fix is generational / arena-tagged handles, which `arena.rs` already notes as deferred to the mutation milestone; bolting on partial validation now wouldn't catch the same-index-wrong-arena case anyway. Leaving it for that milestone.

keraion (approved):
This looks good to me. We'll have some adjustments that will need happen to scale this (namely the leaf index work). But as an experiment this is a good start.

PR #7985, opened 2026-06-21
fix(rust): correct aux_end so node class_types aren't truncated

WittierDinosaur:
## Problem

The Rust parser's String/Typed/MultiString/Regex handlers derived the end of a grammar's `aux_data` block from `aux_data_offsets[grammar_id + 1]`. But `aux_data` blocks are **not stored in grammar-id order**, so the next grammar's offset isn't this block's end — and it's `0` when that grammar has no `aux_data`. When the bound came out smaller than the block, the **count-prefixed** `instance_types` / `class_types` reads were wrongly rejected, and the resulting node lost its class hierarchy.

## Symptom

A parse-tree fidelity gap on the Rust node tree (`_rs_node`): a quoted-string `MultiStringParser` match — snowflake's quoted compression value `'GZIP'` — got

```
class_types = ['compression_type']
```

instead of Python's

```
class_types = ['base', 'compression_type', 'keyword', 'raw', 'word']
```

so `is_type("keyword")` on the Rust node disagreed with Python. (The *unquoted* `GZIP`, whose next grammar happened to have a valid offset, was unaffected — which is why it only showed up for the quoted alternative.)

Root-caused with a probe:
```
gid=2568 (unquoted): aux_start=2152 aux_end=2174 → class_types read OK
gid=2569 (quoted):   aux_start=2174 aux_end=0    → class_types read rejected (empty)
```

## Fix

The `instance_types` / `class_types` reads are count-prefixed and self-delimiting, so `aux_end` is only a safety bound, never a delimiter. Fall back to the full `aux_data` length when the next grammar's offset doesn't sit after this block's start. Applied to all four parser handlers.

## Validation

- **All dialects' Rust-vs-Python YAML fixture parity tests pass (60/60).**
- New regression test asserts `_rs_node` leaf `class_types` mirror Python's for the quoted compression value.
- Verified the quoted and unquoted `GZIP` now both yield the full keyword hierarchy.

Found while prototyping Rust-native lint rules (#7984), which depend on `_rs_node` `class_types` matching Python's `is_type` semantics.

---
## Summary by cubic
Fix Rust parser `aux_data` bounds so `class_types` reads aren’t truncated; `_rs_node` now preserves the full class hierarchy and matches Python. Adds a note that codegen always emits new‑schema fields, so the `aux_end` check is a bounds guard.

- **Bug Fixes**
  - Compute `aux_end` safely via `aux_block_bounds`: use the next `aux_data_offsets` only if it’s after `aux_start`, otherwise fall back to the full `aux_data` length. Applied to String, Typed, MultiString, and Regex handlers.
  - Fix Snowflake case where quoted `'GZIP'` lost its keyword hierarchy; quoted and unquoted values now match Python.
  - Add a regression test comparing `_rs_node` leaf `class_types` to Python for the quoted compression value; all Rust-vs-Python fixture parity tests still pass (60/60).

<sup>Written for commit b47a1d80438fdb631925db80127a0be149ec07e8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7985?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


WittierDinosaur:
Thanks — flagged by the cubic review. I looked into it: the concern is valid in theory but cannot manifest with the current codegen, and I've added a clarifying comment in `33b8ce7d5`.

The `aux_end >= aux_start + N` check isn't actually an old- vs new-schema discriminator — it's a bounds guard. `build_parsers.py` **unconditionally** emits the new-schema fields (`instance_types` count + `class_types` count + ids) for every String/Typed/Multi/Regex parser (`build_parsers.py:919`, `:983`, `:1053`, `:1122`), so the grammars reaching these handlers are always new-schema. There are no old-schema blocks for the full-length fallback to mis-detect, and the `instance_types`/`class_types` reads are count-prefixed and self-delimiting, so they read exactly the right span regardless of where `aux_end` lands.

(Worth noting the heuristic was never a sound schema signal even before this change — a non-monotonic next-offset could already over- or under-bound it. The fix makes it a pure safety bound, which is what it effectively was.)

Validation: all 60 Rust-vs-Python YAML fixture parity tests pass, and the new regression test confirms the quoted/unquoted `GZIP` both yield the full keyword hierarchy.

sqlfluffrs/sqlfluffrs_parser/src/parser/core.rs
```diff
@@ -430,11 +430,28 @@ impl<'a> Parser<'a> {
...
+        let next_aux_off = if (grammar_id.get() as usize + 1) < tables.aux_data_offsets.len() {
             tables.aux_data_offsets[grammar_id.get() as usize + 1] as usize
         } else {
             tables.aux_data.len()
         };
+        let aux_end = if next_aux_off > aux_start {
+            next_aux_off
+        } else {
+            tables.aux_data.len()
+        };
```
keraion: I think this should probably roll into an inline function.

```rust
let (aux_start, aux_end) = aux_block_bounds(tables, grammar_id);
```

```rust
/// Returns the `[start, end)` bounds of `grammar_id`'s `aux_data` block.
///
/// The end is normally the next grammar's offset, but `aux_data` blocks are
/// **not** stored in grammar-id order, so the next grammar may have a smaller
/// offset (e.g. `0` when it has no `aux_data`). When that offset doesn't sit
/// after this block's start, fall back to the full `aux_data` length: the
/// inst/class_types reads in the callers are count-prefixed and self-delimiting,
/// so `aux_end` is only a safety bound, never a delimiter.
///
/// NOTE: codegen (build_parsers.py) unconditionally emits the new-schema fields
/// (instance_types + class_types) for every String/Typed/Multi/Regex parser, so
/// those fields are always present for the grammars that reach these handlers;
/// the `aux_end >= aux_start + N` checks in the callers are bounds guards, not a
/// real old- vs new-schema discriminator.
#[inline]
fn aux_block_bounds(
    tables: &sqlfluffrs_types::GrammarTables,
    grammar_id: GrammarId,
) -> (usize, usize) {
    let idx = grammar_id.get() as usize;
    let aux_start = tables.aux_data_offsets[idx] as usize;
    let next_aux_off = tables
        .aux_data_offsets
        .get(idx + 1)
        .map(|&off| off as usize)
        .unwrap_or_else(|| tables.aux_data.len());
    let aux_end = if next_aux_off > aux_start {
        next_aux_off
    } else {
        tables.aux_data.len()
    };
    (aux_start, aux_end)
}
```

keraion (commented):
This makes sense. We probably want to roll it into an inline function if performance isn't impacted.

WittierDinosaur:
Done — rolled the bounds computation into an inline `aux_block_bounds(tables, grammar_id)` exactly as suggested (`b47a1d804`), replacing the duplicated logic in all four parser handlers. The docstring carries the new-schema-invariant note, which also covers the earlier cubic question (the `aux_end >= aux_start + N` checks are bounds guards, not a schema discriminator). Also merged `main` to clear the conflict (kept both the new `class_types` regression test and the profiling tests that landed on main).

Verified after the refactor + merge:
- `cargo check` / `fmt` clean, `clippy` clean for the changed code.
- `test/core/parser/rust_parser_test.py`: 17 passed.
- Rust-vs-Python YAML fixture parity: 60/60 dialects pass.

keraion (approved):
LGTM

PR #7991, opened 2026-06-22
perf(rust): precompute raw_upper on token construction

rubytobi:
Precomputes `raw_upper` (the uppercased form of the raw token string) at token construction time and stores it on the `Token` struct, eliminating repeated `to_uppercase()` calls in the hot parse path. Previously every grammar rule that
performed a case-insensitive comparison recomputed the uppercase form on each call; with this change the cost is paid once at lex time and amortised across all subsequent uses during parsing.

Both phases improve. Lex time decreases modestly because the single upfront allocation in the constructor is cheaper than the scattered per-call allocations that previously occurred during lexing. Parse time decreases more substantially
because the repeated on-demand allocations in the hot grammar-matching loop are eliminated entirely.

| Benchmark | Before | After | Change |                                                                                                                                           
|---|---|---|---|
| athena/lex_athena | 3.394 s | 3.136 s | -7.6% |
| athena/parse_athena | 4.456 s | 4.125 s | -7.4% |
| tpch/lex_tpch_22 | 22.10 ms | 21.58 ms | -2.3% | 
| tpch/parse_tpch_22 | 148.9 ms | 138.6 ms | -6.9% |
| tpcds/lex_tpcds_99 | 188.0 ms | 185.3 ms | -1.4% |
| tpcds/parse_tpcds_99 | 1.536 s | 1.395 s | -9.2% |

No

### Pull Request checklist

  - Other.

---
## Summary by cubic
Precompute and store the uppercased token string (`raw_upper`) when constructing `Token` to remove repeated allocations in the hot parse path. Parsing is ~4–6% faster on TPC-H/TPC-DS.

- **Performance**
  - Added `raw_upper: String` to `Token` in `sqlfluffrs_types`; compute once and make `raw_upper()` return `&str`.
  - Updated pruning: `GrammarTables::hint_can_match` now takes `raw_upper`, `instance_types`, and `class_types`; `sqlfluffrs_parser` passes them directly, avoiding per-call set builds and checking instance types first.
  - `Token::edit` now recomputes `raw_upper` when `raw` changes to keep them in sync.
  - `sqlfluffrs_python::PyToken.raw_upper()` returns an owned `String` from the borrowed uppercase value.

<sup>Written for commit 2c1683d10e489bfb8b5e3d383740f50137976544. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7991?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs_types/src/token/mod.rs
+++ b/sqlfluffrs/sqlfluffrs_types/src/token/mod.rs
@@ -46,6 +46,7 @@ pub struct Token {
     pub allow_empty: bool,
     pub pos_marker: Option<PositionMarker>,
     pub raw: String,
+    pub raw_upper: String,
     is_whitespace: bool,
     is_code: bool,
     is_comment: bool,
@@ -128,8 +129,8 @@ impl Token {
         self.raw.clone()
     }
 
-    pub fn raw_upper(&self) -> String {
-        self.raw.to_uppercase()
+    pub fn raw_upper(&self) -> &str {
+        &self.raw_upper
     }
 
     /// Get the quoted_value pattern for this token (if any)
@@ -217,8 +218,9 @@ impl Token {
 
     pub fn first_non_whitespace_segment_raw_upper(&self) -> Option<String> {
         self.raw_segments().iter().find_map(|seg| {
-            if !seg.raw_upper().trim().is_empty() {
-                Some(seg.raw_upper().clone())
+            let upper = seg.raw_upper();
+            if !upper.trim().is_empty() {
+                Some(upper.to_owned())
             } else {
                 None
             }
--- a/sqlfluffrs/sqlfluffrs_types/src/token/mod.rs
+++ b/sqlfluffrs/sqlfluffrs_types/src/token/mod.rs
@@ -333,8 +333,10 @@ impl Token {
     }
 
     pub fn edit(&self, raw: Option<String>, source_fixes: Option<Vec<SourceFix>>) -> Self {
+        let new_raw = raw.unwrap_or(self.raw.clone());
         Self {
-            raw: raw.unwrap_or(self.raw.clone()),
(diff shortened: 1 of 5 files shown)


sqlfluffrs/sqlfluffrs_types/src/token/mod.rs
keraion, line 344: We need to update raw_upper here too.

keraion (commented):
Overall this makes sense, just the missed raw_upper site. We might want to introduce a raw + all derived struct to catch this, but this is beyond the scope of this PR.

```rust
pub struct RawStr {
    raw: String,
    upper: String,       // raw.to_uppercase()
    // normalized: String,  // fold in raw_value too, etc.
}

impl RawStr {
    pub fn new(raw: String) -> Self {
        let upper = raw.to_uppercase();
        Self { upper, raw }
    }
    pub fn as_str(&self) -> &str { &self.raw }
    pub fn upper(&self) -> &str { &self.upper }
}

impl std::ops::Deref for RawStr { 
    type Target = str;
    fn deref(&self) -> &str { &self.raw }
}
```

rubytobi:
> Overall this makes sense, just the missed raw_upper site. We might want to introduce a raw + all derived struct to catch this, but this is beyond the scope of this PR.

Good idea, I've gone ahead and drafted one up (https://github.com/rubytobi/sqlfluff/pull/9) to look at after this one has landed.

keraion (approved):
LGTM. Thanks again

Issue #7778, opened 2026-04-22
Rust optimization ideas


peterbud:
@keraion  I let one of the more complex AI models run a static performance analysis on the Rust codebase to look for optimization ideas. I'm sharing the ideas here to check them later on individually as time permits.

The strongest opportunities are mostly about allocation and string churn rather than missing parser caches, because the parser already has several good structural optimizations in place such as simple-hint pruning, terminator caches, precomputed bracket links, and `Arc::try_unwrap` on match results in core.rs, helpers.rs, lexer.rs, and iterative.rs.

The highest-value options are:

- [x] The lexer and parser copy token text too often. `TemplateElement` copies `raw` and clones the matcher in lexer.rs, and `Token::raw()` / `Token::raw_upper()` allocate new `String`s in mod.rs. That cost then leaks into hot loops like bracket pairing in lexer.rs and terminator scanning in match_algorithms.rs. A borrowed `&str` path or `Arc<str>`-style sharing would likely move the needle the most.

- [x] Token construction pays for random UUID generation per token in construction.rs and mod.rs, while equality can already fall back to structural checks in eq.rs. If full cryptographic uniqueness is not required, replacing this with a cheap monotonic ID would reduce lexer cost materially on large files.

- [x] Literal normalization recompiles regexes on demand. `Token::normalize()` calls `RegexMode::new()` repeatedly in mod.rs. If quoted-value and escape regexes were compiled once per matcher and reused, string-heavy SQL should get cheaper.

- [x] The table-driven parser still allocates a lot of transient `Vec`s and re-clones terminator lists in its frame handlers. Examples are in oneof.rs, delimited.rs, anynumberof.rs, and repeated `frame.table_terminators.to_vec()` calls such as sequence.rs. This looks like a good fit for more `SmallVec` usage and borrowed slices from `GrammarContext` instead of `collect()` + `to_vec()`.

- [ ] Sequence combination still clones nested match containers in sequence.rs. On deeply nested expressions that means extra vector copies even though the tree is already mostly `Arc`-based. Avoiding flatten-by-clone here is worth profiling.

- [ ] `GrammarContext::grammar_id_name()` allocates a `String` every time in grammar_api.rs, and delimited parsing uses it in non-debug control flow for delimiter filtering in delimited.rs. A borrowed-name API or comparison by stable string ID would remove avoidable hot-path allocations.

[...]

PR #7997, opened 2026-06-24
perf: replace per-token uuid4 with a tagged monotonic counter (Python + Rust)

keraion:

Replace per-instance `uuid4()` / `Uuid::new_v4()` segment-id generation with a process-global monotonic counter on both the Python and Rust sides.

Because `PyToken.uuid` surfaces the Rust `u128` to Python verbatim, a fixed tree mixes Rust-origin tokens with Python-created segments in one id space. To keep the two counters from colliding (both start at 1), each id carries a source tag in its high bits — Python `1 << 120`, Rust `2 << 120` — so a Rust id can never equal a Python id.

### Benchmarks
Per-call cost of the id generator (called once per token/segment). Isolated microbenchmarks — an explicit end-to-end parse delta is omitted because the Rust match-application path currently falls back to Python on this branch, which would muddy attribution.

Python (`timeit`, 2M calls, 3.13):

| method | ns/call | vs uuid4 |
|---|---|---|
| `uuid4().int` (before) | 1201 | 1.00x |
| `os.urandom(16)` int (#7988) | 809 | 0.67x |
| `get_next_id()` (after) | 35.3 | 0.03x (~34x faster) |

Rust (`criterion`, release):

| method | ns/call | vs uuid4 |
|---|---|---|
| `Uuid::new_v4().as_u128()` (before) | ~740 | 1.00x |
| `next_id()` (after) | ~1.6 | ~466x faster |

Segment/token `uuid`s are now small tagged sequential integers rather than random 128-bit values. They remain unique within a process (the only property the fix engine relies on) but are no longer globally/cross-process unique, and are no longer valid RFC-4122 UUIDs.

### Pull Request checklist

  - Other.

---
## Summary by cubic
Replaced per-token UUIDs with a fast, process-global monotonic counter in Python and Rust. IDs are source-tagged to keep Python (`1 << 120`) and Rust (`2 << 120`) spaces disjoint, cutting per-id cost by ~34x (Python) and ~466x (Rust).

- **Refactors**
  - Added `get_next_id()` in `sqlfluff.core.helpers.identity` and `identity::next_id()` in `sqlfluffrs_types`.
  - Replaced `uuid4()`/`os.urandom(16)` and `Uuid::new_v4()` for segment/token `uuid` fields with tagged counters; `block_uuid` remains an RFC-4122 UUID.
  - Updated segment class `_cache_key` to use `get_next_id()` (hex).

<sup>Written for commit a635170eaf11536fa0f336da2323ad7d647db592. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7997?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs_types/src/identity.rs
+++ b/sqlfluffrs/sqlfluffrs_types/src/identity.rs
@@ -1,0 +1,55 @@
+//! Process-global unique identifiers for tokens.
+//!
+//! Replaces per-token `Uuid::new_v4()`. A token's `uuid` is only ever used as
+//! an opaque identity key (compared by equality and used as a map key in the
+//! fix engine); it is never parsed back as an RFC-4122 UUID, so a monotonic
+//! counter is sufficient -- and far cheaper than drawing 16 bytes of CSPRNG
+//! entropy per token.
+//!
+//! The high bits carry a source tag so these ids stay disjoint from ids minted
+//! on the Python side (which uses `PYTHON_TAG = 1 << 120`). `PyToken.uuid`
+//! surfaces this `u128` to Python verbatim, where Python-created segments and
+//! Rust-origin tokens share one id space inside the fix engine. Without the
+//! tag, both counters would start at 1 and a Rust token would collide with an
+//! unrelated Python segment.
+//!
+//! NOTE: this is the token *identity* uuid only. Templated `block_uuid`s remain
+//! genuine `Uuid`s -- they are round-tripped through `Uuid::parse_str` at the
+//! Python boundary and must keep RFC-4122 structure.
+
+use std::sync::atomic::{AtomicU64, Ordering};
+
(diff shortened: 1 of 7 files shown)


peterbud:
Is this PR also contributes to #7778 ?

keraion:
> Is this PR also contributes to #7778 ?

~~It does, marked it as such~~ actually I don't want to close that one outright

peterbud (approved):
LGTM

Issue #6919, opened 2025-05-30
RF03: Databricks  - STRUCT function alias is identified as column reference


PhilippLange:
### Search before asking

- [x] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

I have an issue with the behavior of linting the STRUCT function in databricks dialect.
I'm willing to attempt a fix as I contributed in the past.
Someone has to point me in the right direction though.

The issue occurs with the following statement:
```sql
-- input
SELECT
    tbl.a,
    STRUCT(
        SUM(tbl.b) as MEASUREMENT_VALUE,
        'MeasurementName' as MEASUREMENT_NAME
    ) AS measurement
FROM tbl
```

### Expected Behaviour

I expect the AS to be capitalized and the Coumn alias (within the struct) not to be altered:
```sql
-- expected output
SELECT
    tbl.a,
    STRUCT(
        SUM(tbl.b) AS MEASUREMENT_VALUE,
        'MeasurementName' AS MEASUREMENT_NAME
    ) AS measurement
FROM tbl
```

### Observed Behaviour

Instead the column alias is identified as a reference and made explicit and lowerecased while the AS is not capitalized:
```sql
-- actual output
SELECT
    tbl.a,
    STRUCT(
        SUM(tbl.b) as tbl.measurement_value,
        'MeasurementName' as tbl.measurement_name
    ) AS measurement
FROM tbl
```

```sh
L:   4 | P:  23 | RF03 | Unqualified reference 'MEASUREMENT_VALUE' found in
                       | single table select. [references.consistent]
L:   5 | P:  30 | RF03 | Unqualified reference 'MEASUREMENT_NAME' found in single
                       | table select. [references.consistent]
```

### How to reproduce

Lint the following statement:
```sql
SELECT
    tbl.a,
    STRUCT(
        SUM(tbl.b) as MEASUREMENT_VALUE,
        'MeasurementName' as MEASUREMENT_NAME
    ) AS measurement
FROM tbl
```

### Dialect

databricks

### Version

3.4.0

### Configuration

[sqlfluff]
dialect = databricks

### Are you willing to work on and submit a PR to address the issue?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

PR #7999, opened 2026-06-25
Fix RF02 false positive on SELECT * EXCEPT (...) columns

### LINKED ISSUE #5764: RF02 falsely triggering for SELECT * EXCEPT (...) in Databricks
### Search before asking

### What Happened

For statements in databricks SQL which use "SELECT table.* EXCEPT (column_name)" RF02 is triggered even though it probably should not be, since the reference to the table is clear.

image
image

### Expected Behaviour

RF02 should not trigger if the unqualified Reference is part of the EXCEPT clause in the SELECT * context.

### Observed Behaviour

image
RF02 triggers.

### How to reproduce

Minimal SQL: 
```
SELECT 
    test.* EXCEPT (test_column),
    test_2.*
FROM test
    LEFT JOIN test_2 ON test.test_column = test_2.test_column
```
`
sqlfluff lint test.sql --dialect databricks`

### Dialect

Databricks

### Version

3.0.3

### Configuration

```
[sqlfluff]
dialect = databricks
```

### Code of Conduct

arpitjain099:
Fixes #5764.

RF02 was wrongly flagging the columns inside a `SELECT * EXCEPT (col1, col2)` clause as unqualified references. Those columns just name what to drop from the star expansion, so they don't need to be qualified and shouldn't be treated like ordinary column references.

The fix excludes object references that descend from a `select_except_clause` when building the select clause reference buffer, mirroring the existing STRUCT-alias exclusion (`exclude_ids`) that was added for #6919. Because it's a single change in the shared analysis layer, it covers every dialect that has an EXCEPT clause: Databricks, SparkSQL, ClickHouse, and BigQuery.

I kept the scope to the EXCEPT false positive only. The separate semi/anti-join request from the issue comments is left out.

Validation:
- added pass cases for databricks/sparksql/clickhouse plus a fail case proving a genuinely unqualified reference elsewhere in the same select still gets flagged
- `python -m pytest test/rules/yaml_test_cases_test.py -k "RF01 or RF02 or RF03 or AL05"` is green (337 passed)
- re-ran the issue's repro and the false positive is gone

I used AI assistance while working on this, but I wrote and reviewed the change and the tests myself and verified the behavior locally.

---
## Summary by cubic
Fixes RF02 falsely flagging columns listed in SELECT * EXCEPT (...) as unqualified. We now ignore those names during select analysis, removing the false positive across Databricks, SparkSQL, ClickHouse, and BigQuery.

- **Bug Fixes**
  - Exclude `object_reference` nodes under `select_except_clause` when building the select reference buffer, mirroring the STRUCT alias exclusion.
  - Added tests for Databricks, SparkSQL, and ClickHouse, plus a counterexample to ensure genuine unqualified references still fail.

<sup>Written for commit dd0212d63ae566fe5d7622127695ebfcb86d776d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/7999?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/utils/analysis/select.py
+++ b/src/sqlfluff/utils/analysis/select.py
@@ -70,6 +70,26 @@ def _get_struct_alias_refs(segment: BaseSegment) -> set[int]:
     return struct_alias_ids
 
 
+def _get_select_except_refs(segment: BaseSegment) -> set[int]:
+    """Get the ids of object references inside a SELECT * EXCEPT clause.
+
+    Dialects like Databricks, SparkSQL, and ClickHouse support
+    ``SELECT t.* EXCEPT (col1, col2)`` where the excluded columns are listed
+    inside an ``EXCEPT (...)`` clause. Those columns name what to drop from the
+    star expansion, so they don't need to be qualified and shouldn't be treated
+    as ordinary column references by rules like RF02.
+    See: https://github.com/sqlfluff/sqlfluff/issues/5764
+    """
+    except_ref_ids: set[int] = set()
+    for except_clause in segment.recursive_crawl(
+        "select_except_clause",
+        no_recursive_seg_type=["select_statement", "merge_statement"],
+    ):
+        for ref in except_clause.recursive_crawl("object_reference"):
+            except_ref_ids.add(id(ref))
+    return except_ref_ids
+
+
 def _get_object_references(
     segment: BaseSegment,
     exclude_ids: Optional[set[int]] = None,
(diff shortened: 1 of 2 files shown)


peterbud (approved):
Thanks @arpitjain099  and welcome.

PR #8011, opened 2026-06-26
[databricks]: support bare Databricks notebook magic cells

### LINKED ISSUE #7877: Databricks dialect to support magic commands without MAGIC
### Search before asking

### Description

Databricks SQL notebooks support code segment declarations without the MAGIC prefix:
```
-- COMMAND ----------
%python
print("Hello world")
-- COMMAND ----------
```
This is quite idiomatic when you maintain notebooks in a repository for example, as opposed to prefixing every line with `-- MAGIC `.

Essentially, if `%<name>` appears right after a command start, that cell is simply specified in that language, no need for any magic prefixes.

### Use case

Linting modern Databricks notebook style.

### Dialect

Databricks

### Code of Conduct

peterbud:

This change adds Databricks support for notebook cells that begin with a bare language marker like %python or %run, without requiring the `-- MAGIC`  prefix.

It introduces a lexer/parser path that treats these cells as opaque magic-cell content until the next `-- COMMAND ----------` delimiter, and adds parse fixtures covering both multi-line and single-line bare magic cells.

Fixes #7877 

### Pull Request checklist

  - Other.

---
## Summary by cubic
Adds support in the Databricks dialect for bare magic cells (`%python`, `%run`, etc.), including when they appear right after the notebook header; `%sql` headers are handled so the SQL body is parsed normally. Fixes #7877.

- **New Features**
  - Recognizes bare magic cells without `-- MAGIC` (including at notebook start), capturing content until the next command delimiter; `%sql` lines are treated as headers and the following SQL is parsed as normal SQL.
  - Only the first line of a cell can start a bare magic; preserves `%` as the SQL modulo operator even at line start or next to magic keywords.

- **Refactors**
  - Adopts regex-based lexer rules (`bare_magic_cell`, `bare_magic_sql`, `command`, and first-cell variants) with a typed grammar hook in `MagicCellStatementSegment`, replacing the earlier parser-only approach.

<sup>Written for commit 6d70ed516cddce4238930f5a7981d6700c9c0ba6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8011?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_databricks.py
+++ b/src/sqlfluff/dialects/dialect_databricks.py
@@ -118,11 +118,6 @@ databricks_dialect.insert_lexer_matchers(
         ),
         RegexLexer("magic_line", r"(-- MAGIC)( [^%]{1})([^\n]*)", CodeSegment),
         RegexLexer("magic_start", r"(-- MAGIC %)([^\n]{2,})(\r?\n)", CodeSegment),
-        RegexLexer(
-            "bare_magic_cell",
-            r"(?s)%[^\r\n]*(?:(?!(?:\r?\n){2}-- COMMAND ----------(?:\r?\n)).)*(?=(?:\r?\n){2}-- COMMAND ----------(?:\r?\n)|\Z)",
-            CodeSegment,
-        ),
     ],
     before="inline_comment",
 )
@@ -306,9 +301,6 @@ databricks_dialect.add(
     ),
     MagicLineGrammar=TypedParser("magic_line", CodeSegment, type="magic_line"),
     MagicStartGrammar=TypedParser("magic_start", CodeSegment, type="magic_start"),
-    BareMagicCellGrammar=TypedParser(
-        "bare_magic_cell", CodeSegment, type="bare_magic_cell"
-    ),
     VariableNameIdentifierSegment=OneOf(
         Ref("NakedIdentifierSegment"),
         Ref("BackQuotedIdentifierSegment"),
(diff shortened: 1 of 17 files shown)


src/sqlfluff/dialects/dialect_databricks.py
```diff
@@ -118,6 +118,11 @@
         ),
         RegexLexer("magic_line", r"(-- MAGIC)( [^%]{1})([^\n]*)", CodeSegment),
         RegexLexer("magic_start", r"(-- MAGIC %)([^\n]{2,})(\r?\n)", CodeSegment),
+        RegexLexer(
+            "bare_magic_cell",
+            r"(?s)%[^\r\n]*(?:(?!(?:\r?\n){2}-- COMMAND ----------(?:\r?\n)).)*(?=(?:\r?\n){2}-- COMMAND ----------(?:\r?\n)|\Z)",
```
keraion: I think this breaks the modulo operator.

peterbud: I will check - but that means we don't have module test for databricks dialect?

keraion: Yeah, I didn't see one explicitly for modulo in databricks.

peterbud: Hi @keraion I have tried to use a proper grammar, but I'm not sure that's the best approach - what do you think?

keraion: I don't think that your basis was off on the utilizing the lexer, it was just a bit overly greedy. Let me know what you think of https://github.com/sqlfluff/sqlfluff/commit/d7d2fc298b1e50415b1f2d78f5c80da61a25d0bd

peterbud: Hi @keraion - thanks for the feedback. I have also tried to make the regex smarter, and my agent also suggested a hard-coded list similar to ` r"%(?:python|scala|r|sh|md|run|fs|pip|conda)\b[^\r\n]*"`, but I felt this is rather brittle, and  not an exhaustive list. What if there will be a different starter word there?

keraion: We could make the list managed like we do with keywords and have it expanded out from that. If we don't do that, we at least need the top line `%` to anchor for those non-sql languages.

keraion (approved):
cubic noted a nice to have, but LGTM

PR #8020, opened 2026-06-28
perf(rust): borrow static segment type/class strings into the Node + arena tree

keraion:
### Brief summary of the change made

Use Cow for a number of static str cases so we aren't reallocating and cloning all over the place.

### Are there any other side effects of this change that we should be aware of?
None

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

---
## Summary by cubic
Borrowed static segment class/type strings across the Rust parser using Cow to avoid cloning, cutting allocations when building the parse/arena tree. No behavior changes.

- **Refactors**
  - Replaced owned `String` with `Cow<'static, str>` for `segment_class` and `segment_type` in `Node`, arena, and `MatchedClass`, and updated parser paths to pass borrowed values.
  - Borrow grammar-table names and token types into nodes; only allocate for runtime overrides (e.g., `instance_types`).
  - Updated getters and Python bindings to return owned strings at API boundaries (`get_type`, `segment_class`, `PyMatchResult.matched_class`).

<sup>Written for commit b5ee6c5f9d8be44ce991b6ec90d19b596f96ed53. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8020?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #8019, opened 2026-06-28
perf(rust): compute raw_normalized lazily with cached regex compilation

keraion:
### Brief summary of the change made

`Token::base_token` called `Token::normalize()` for every lexed token to populate a `raw_value` field that had no readers (`raw_normalized()` was `todo!()`). For most tokens that was a wasted string clone; for quoted/escaped tokens it recompiled a regex per token, which dominated lexing cost.

Rather than dropping the work (it's needed once `raw_normalized()` is implemented for linting), this makes it lazy and cached:
- Remove the eagerly-computed `raw_value` field; nothing read it.
- Implement `raw_normalized()` to compute on demand from fields already on the token, with a fast path that skips all regex work when there's no `quoted_value`/`escape_replacement`.
- Add a process-global `RegexMode` cache so each distinct dialect pattern compiles at most once instead of once per token.

Performance
Criterion, lexing the TPC-H (22) and TPC-DS (99) query suites, vs. main:

| Suite | Stage | main | branch | change |
|---|---|---|---|---|
| TPC-H | lex | 16.13 ms | 7.61 ms | **−52.9%** |
| | lex+parse | 114.3 ms | 106.4 ms | −7.0% |
| TPC-DS | lex | 147.72 ms | 65.25 ms | **−55.8%** |
| | lex+parse | 1158.3 ms | 1072.5 ms | −7.4% |
Lexing ~2.2× faster; parse unchanged (within noise), confirming the change is isolated to token construction.

**Now with Tokens 20% smaller!**

Supersedes #8005 

### Are there any other side effects of this change that we should be aware of?
None

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`.
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - Full autofix test cases in `test/fixtures/linter/autofix`.
  - Other.
- Added appropriate documentation for the change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate.

---
## Summary by cubic
Make token normalization lazy inside `RawString` with a process‑global regex cache, and shrink token memory. Removes the unused eager `raw_value`; boxes the transform spec (including `casefold`) and skips duplicate uppercase strings; lexing ~2× faster, parse unchanged.

- **Refactors**
[...]

PR #8021, opened 2026-06-28
perf(rust): borrow static Token strings as Cow<'static, str>

keraion:

`Token` held several `String` fields that are always set from a tiny fixed vocabulary of static literals, so every lexed token was heap-allocating them. This borrows them as `Cow<'static, str>` so the literal cases are `Cow::Borrowed` (no allocation), and only genuinely dynamic values allocate via `Cow::Owned`:

- `token_type` / `class_name` — the segment type/class (e.g. `"whitespace"`, `"BaseSegment"`). Matches the `Node` / `MatchedClass` change in #8020, so the borrowed value threads straight through and `token_to_node` drops its `Cow::Owned(...clone())` at the boundary.
- `_default_raw` (`""` / `" "` / `"\n"`), `preface_modifier` (`""` / `"[META] "`), and the common `suffix` (`""`) — now `Cow::Borrowed`; only the dynamic `suffix` cases (a raw token's quoted form, an indent's block uuid) allocate.

### Performance

Pure speed win — `Token` size is unchanged (`Cow<str>` is the same size as `String`); the gain is allocation avoidance at construction, which compounds over the millions of tokens lexing builds. Criterion, lexing the TPC-H/TPC-DS suites, vs. the stacked base (#8019 + #8020):

|---|---|---|---|
| `lex_tpch_22` | 7.57 ms | 6.50 ms | **−13.2%** |
| `lex_tpcds_99` | 66.45 ms | 56.98 ms | **−14.3%** |

Both `p = 0.00`. No behavior change.

### Stacking

> [!IMPORTANT]
> Stacked on **#8019** and **#8020** — this branch includes their commits and should merge after both. Once they land, this will be rebased onto `main` so the diff reduces to just the Cow-on-Token commits.

None. Public getters (`get_type`, `preface_modifier`, the Python bindings) still return owned `String` at the API boundary.

---
## Summary by cubic
Borrowed static token strings as `Cow<'static, str>` and added lazy normalization/regex caching to cut allocations and speed up lexing by ~13–14%, with no behavior change.

- **Refactors**
  - `Token` now holds `token_type`, `class_name`, `_default_raw`, `preface_modifier`, and `suffix` as `Cow<'static, str>` (literals borrow; dynamic values allocate). Parser and node construction use `as_ref()` and pass through borrowed values; public getters still return `String`.
  - `RawString` computes normalization lazily and skips redundant uppercase copies; added a process-global regex cache in `sqlfluffrs_types::regex` via `RegexMode::cached{,_with_flags}`.

<sup>Written for commit af3e02a6cc75093e49a7245efb8bd5782dfbb48f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8021?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/match_result.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/match_result.rs
@@ -793,7 +793,7 @@ fn node_pos_marker(node: &Node) -> Option<PositionMarker> {
 /// determine the variant.  All other tokens become `Node::Raw`.
 fn token_to_node(tok: &Token) -> Node {
     if tok.is_meta {
-        let meta_type = match tok.token_type.as_str() {
+        let meta_type = match tok.token_type.as_ref() {
             "end_of_file" => MetaType::EndOfFile,
             "indent" => MetaType::Indent { is_implicit: false },
             "dedent" => MetaType::Dedent { is_implicit: false },
@@ -821,9 +821,9 @@ fn token_to_node(tok: &Token) -> Node {
             .instance_types
             .first()
             .map(|s| Cow::Owned(s.clone()))
-            .unwrap_or_else(|| Cow::Owned(tok.token_type.clone()));
+            .unwrap_or_else(|| tok.token_type.clone());
         Node::new_raw_with_class_types(
-            Cow::Owned(tok.class_name.clone()),
+            tok.class_name.clone(),
             segment_type,
             tok.raw().to_owned(),
             tok.pos_marker.clone(),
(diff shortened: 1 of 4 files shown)


WittierDinosaur (approved):
LGTM

PR #8022, opened 2026-06-28
perf(rust): share class_types as a per-kind Arc instead of per-token HashSet

keraion:

Every token built its own `class_types` `HashSet` at construction: the lexer always supplies an empty set and the constructor chain clones+inserts `{base, raw, kind}`, so each token paid ~6 allocations (growing `HashSet` clones + `String` type names) to build a set that is **identical for every token of that kind** and **never mutated** afterward.

Replace the per-token build with a shared `Arc<HashSet<String>>` per kind, built once via a `Lazy` static. `class_types` becomes `Arc<HashSet<String>>`; each constructor assigns its kind's static (an `Arc` clone, no allocation).

### Performance

Criterion, lexing the TPC-H/TPC-DS suites, vs. the stacked base:

|---|---|---|---|
| `lex_tpch_22` | 6.60 ms | 4.91 ms | **−25.5%** |
| `lex_tpcds_99` | 57.72 ms | 42.36 ms | **−26.6%** |

Both `p = 0.00`. `Token` also shrinks 528 → 496 bytes, and the whole token stream now shares ~15 sets instead of one `HashSet` per token. No behavior change.

### Stacking

> [!IMPORTANT]
> Stacked on **#8021** (which is itself stacked on #8019 and #8020). Merge those first; this will be rebased onto `main` as they land.

None. `class_types` is read-only after construction (verified — all mutations are on `segment_kwargs`, never on a `Token`).

---
## Summary by cubic
Share the class‑type hierarchy across tokens by switching `Token.class_types` to an `Arc<HashSet<String>>` and using per‑kind statics via `once_cell::sync::Lazy` instead of building a new set per token. This removes redundant allocations and speeds up lexing with no behavior change.

- **Performance**
  - TPC‑H 22: 6.60 ms → 4.91 ms (−25.5%); TPC‑DS 99: 57.72 ms → 42.36 ms (−26.6%).
  - `Token` size 528 → 496 bytes.

<sup>Written for commit 1dcc8fed5c4491cf8e303a6b2d727974e535f557. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8022?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs_types/src/token/construction.rs
+++ b/sqlfluffrs/sqlfluffrs_types/src/token/construction.rs
@@ -5,6 +5,7 @@ use std::borrow::Cow;
 use std::sync::Arc;
 
 use hashbrown::HashSet;
+use once_cell::sync::Lazy;
 use uuid::Uuid;
 
 impl Token {
@@ -15,7 +16,9 @@ impl Token {
         segments: Vec<Token>,
     ) -> Self {
         let TokenConfig {
-            class_types,
+            // The lexer always supplies an empty class_types; the hierarchy is
+            // derived per-kind from a shared static instead (see *_CT below).
+            class_types: _,
             instance_types,
             trim_start,
             trim_chars,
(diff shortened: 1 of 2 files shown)


WittierDinosaur (approved):
LGTM

PR #8024, opened 2026-06-28
perf(rust): memoize Ref child-grammar resolution (~22% faster parsing)

keraion:

A symbolicated profile of the table-driven parser showed the single biggest hot spot was the generated `get_*_segment_grammar` lookup — a ~1,200-arm `match name { ... }` hit on every `Ref` to resolve its name to a grammar — at **~20% of parse self-time**.

Resolving a `Ref` to its child grammar (first element child if present, else the by-name dialect lookup) depends only on the `Ref`'s `grammar_id`, but the same `Ref` is hit thousands of times per parse. This memoizes `grammar_id -> child grammar_id` in a per-`Parser` `HashMap`, so the by-name match runs at most once per distinct `Ref`; subsequent hits are an integer lookup. The per-call `element_children` `Vec` allocation is removed too (`.collect()[0]` → `.next()`).

A second small commit pre-sizes the frame parse cache (`TableParseCache`) to `tokens.len() * 8` in `Parser::new` — it grows to several times the token count during a parse, and reserving up front avoids repeated rehashing.

### Performance

Criterion, parsing the TPC-H/TPC-DS suites:

|---|---|---|---|
| `parse_tpch_22` | 97.4 ms | 75.8 ms | **−22.2%** |
| `parse_tpcds_99` | 979.6 ms | 771.4 ms | **−21.3%** |

(Memo ≈ −22%, cache pre-sizing ≈ −1% more.) Both `p = 0.00`. A re-profile with the memo in place confirms `get_*_segment_grammar` drops from 19.5% → 0.8% self-time.

None — behavior is unchanged. The memoized value is a pure function of the `Ref`'s `grammar_id` (both `element_children` and the by-name `ref_name → get_segment_grammar` resolution are deterministic from it), and the cache is per-`Parser` (single dialect). All dialect fixture tests pass.

---
## Summary by cubic
Memoized Ref child-grammar resolution and pre-sized the frame parse cache to cut rehashing, resulting in ~22% faster parsing on TPC‑H/TPC‑DS. Behavior is unchanged.

- **Refactors**
  - Cache `grammar_id -> child grammar_id` per `Parser` to avoid the large by-name match; use `.next()` instead of allocating a `Vec`.
  - Add `TableParseCache::with_capacity` and pre-size to `tokens.len() * 8` in `Parser::new`.

<sup>Written for commit 6079b0c2900a742b8655fc3a7c305ee1131ac4b7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8024?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/cache.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/cache.rs
@@ -95,6 +95,16 @@ impl TableParseCache {
         }
     }
 
+    /// Pre-size the cache. The frame cache grows to several times the token
+    /// count during a parse; reserving up front avoids repeated rehashing.
+    pub fn with_capacity(cap: usize) -> Self {
+        TableParseCache {
+            cache: HashMap::with_capacity(cap),
+            hits: 0,
+            misses: 0,
+        }
+    }
+
     /// Check cache for a result
     pub fn get(&mut self, key: &TableCacheKey) -> Option<&TableCacheValue> {
         match self.cache.get(key) {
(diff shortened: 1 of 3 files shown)


WittierDinosaur (approved):
Nice!

PR #8045, opened 2026-07-01
Fix ClickHouse double equals parsing

### LINKED ISSUE #8033: ClickHouse: == equality operator is unparsable
### What happened

ClickHouse accepts `==` as a synonym for `=` (both map to `equals()`).
The `clickhouse` dialect reports `==` as an unparsable section.

### Expected behaviour

`a == b` parses as an equality comparison (equivalent to `a = b`).

### Observed behaviour

```
L1:P25 PRS: Found unparsable section: '== 1'
```

### How to reproduce

```sql
SELECT a FROM t WHERE a == 1
```

```bash
sqlfluff parse --dialect clickhouse -
```

Valid, verified on ClickHouse 26.4.2.
Docs: https://clickhouse.com/docs/sql-reference/operators (comparison operators).

### Dialect

clickhouse

### Version

sqlfluff 4.2.2 (latest release), Python 3.12; also applies to `main`.

### Configuration

None — default rules, `--dialect clickhouse`.

ishaanlabs-gg:
## Summary
- Add ClickHouse lexer/parser support for the double-equals equality operator.
- Add a ClickHouse parse fixture covering SELECT a FROM t WHERE a == 1.

Fixes #8033.

## Testing
- Reproduced the original PRS failure with SQLFluff parse for SELECT a FROM t WHERE a == 1, then reran the same command after the fix and it parsed successfully.
- Generated ClickHouse parse fixtures successfully: 48 fixtures built.
- Ran focused fixture test: 3 passed, 6498 deselected.
- Ran ClickHouse dialect slice: 144 passed, 6357 deselected.
- Ran py_compile for src/sqlfluff/dialects/dialect_clickhouse.py.
- Ran git diff --check for the changed parser and fixture files.

### DIFF
--- a/src/sqlfluff/dialects/dialect_clickhouse.py
+++ b/src/sqlfluff/dialects/dialect_clickhouse.py
@@ -11,6 +11,7 @@ from sqlfluff.core.parser import (
     BaseSegment,
     Bracketed,
     CodeSegment,
+    CompositeComparisonOperatorSegment,
     Conditional,
     Dedent,
     Delimited,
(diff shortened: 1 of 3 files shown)


peterbud (approved):
Thanks

PR #8050, opened 2026-07-01
fix(sparksql): treat LEFT/RIGHT as non-reserved so they can be lambda variables

### LINKED ISSUE #5004: Spark SQL anonymous functions don't parse
### Search before asking

### What Happened

`sqlfluff lint` complains about unparsable sections in a Spark SQL query that contains an anonymous function.

### Expected Behaviour

Anonymous functions should parse properly.

### Observed Behaviour

Spark SQL queries that contain anonymous functions fail to parse.

### How to reproduce

This example from the [array_sort() function's documentation](https://spark.apache.org/docs/latest/api/sql/index.html#array_sort) fails to parse:

```
SELECT array_sort(array('bc', 'ab', 'dc'), (left, right) -> case when left is null and right is null then 0 when left is null then -1 when right is null then 1 when left < right then 1 when left > right then -1 else 0 end)
```

Passing this SQL in a file named `test.sql` to sqlfluff results in the following output:

```
sqlfluff lint --dialect sparksql test.sql
== [test.sql] FAIL
L:   1 | P:   1 | LT05 | Line is too long (222 > 80).
                       | [layout.long_lines]
L:   1 | P:  45 |  PRS | Line 1, Position 45: Found unparsable section: 'left,
                       | right'
L:   1 | P:  58 |  PRS | Line 1, Position 58: Found unparsable section: '-> case
                       | when left is null and right is n...'
WARNING: Parsing errors found and dialect is set to 'sparksql'. Have you configured your dialect correctly?
All Finished 📜 🎉!
```

### Dialect

Spark SQL

### Version

sqlfluff, version 2.1.4
Python 3.10.6

### Configuration

This is reproducible in a vanilla sqlfluff configuration without `.sqlfluff` or `.sqlfluffignore`.

### Code of Conduct

anxkhn:

The Spark SQL function reference documents `array_sort` with the example
`array_sort(arr, (left, right) -> case when left < right then ... end)`, but it
fails to parse (`unparsable` / PRS) because `left` and `right` are treated as
reserved keywords and so are rejected as lambda variables, both in the parameter
list and in the lambda body.

`LEFT` and `RIGHT` are only reserved in Spark's [ANSI mode](https://spark.apache.org/docs/latest/sql-ref-ansi-compliance.html#sql-keywords).
In Spark's default (non-ANSI) mode they are strict-non-reserved and may be used
as identifiers. This moves `LEFT`/`RIGHT` from the reserved to the unreserved
keyword set, matching how the sibling strict-non-reserved join keywords
`ANTI`/`SEMI` are already handled. They are still matched as keywords by
`JoinTypeKeywordsGrammar`, so `LEFT`/`RIGHT [OUTER] JOIN` continue to parse and
they remain unusable as table aliases.

fixes #5004

- Lambdas still parse implicitly (bracketed list + binary `->`), so no existing
  sparksql fixtures change; the only new YAML is the added `select_lambda` test.
- `SELECT 1 AS left` is still rejected, but so are the other existing
  strict-non-reserved keywords (`ANTI`/`SEMI`/`NATURAL`); this is pre-existing
  behaviour brought to parity, not a new regression.
- Out of scope: the `databricks` dialect (which has its own keyword file) has a
  separate, pre-existing lambda issue and is intentionally left untouched here.
- The change was validated against the Python parser locally (the compiled Rust
  parser is not built in my environment). It edits only the dialect keyword
  sets, from which the Rust tables are regenerated on build, so it is expected to
  be a no-op for the Rust path.

### Pull Request checklist

  - [x] `.sql`/`.yml` parser test cases in `test/fixtures/dialects`
    (`test/fixtures/dialects/sparksql/select_lambda.sql` + auto-generated
    `select_lambda.yml`), covering the issue example, reserved-keyword lambda
    parameters, a bare reserved-keyword parameter, and `left`/`right` as ordinary
    column identifiers.
- [x] No documentation change needed (internal dialect grammar).
- [x] No follow-up issue needed for this fix.

---
## Summary by cubic
Make `LEFT`/`RIGHT` unreserved in the Spark SQL dialect so lambdas parse in non-ANSI mode, and make Spark keyword lists override inherited ANSI to avoid conflicts. Fixes #5004.

- **Bug Fixes**
  - Join parsing is unchanged; `LEFT`/`RIGHT` still work in joins and cannot be table aliases.
  - Added parser fixtures covering the documented example and use as identifiers.

<sup>Written for commit 5f6ac80ef3746b12bf84cf02a6cb599bb1196009. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8050?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/sqlfluff/dialects/dialect_sparksql.py
```diff
@@ -265,6 +265,16 @@
...
+# `LEFT` and `RIGHT` are only reserved in Spark's ANSI mode. In the default
+# (non-ANSI) mode they are strict-non-reserved, so they can be used as
+# identifiers, e.g. as lambda variables in higher-order functions like
+# `array_sort(arr, (left, right) -> ...)` (the documented example in #5004).
+# Treat them as unreserved (as already done for the sibling strict-non-reserved
+# join keywords `ANTI`/`SEMI`). They are still matched as keywords in
+# `JoinTypeKeywordsGrammar`, so `LEFT`/`RIGHT [OUTER] JOIN` continue to parse.
+# https://spark.apache.org/docs/latest/sql-ref-ansi-compliance.html#sql-keywords
+sparksql_dialect.sets("reserved_keywords").difference_update(["LEFT", "RIGHT"])
+sparksql_dialect.sets("unreserved_keywords").update(["LEFT", "RIGHT"])
```
keraion: The lists themselves should be changed under `dialect_sparksql_keywords.py` instead.

keraion (commented):
Thanks for the PR. The dialect's keyword changes should go under the `*_keywords.py` file.

src/sqlfluff/dialects/dialect_sparksql.py
anxkhn: Thanks, moved it. `LEFT`/`RIGHT` are now in `UNRESERVED_KEYWORDS` (and out of `RESERVED_KEYWORDS`) in dialect_sparksql_keywords.py.

One wrinkle worth flagging: the sparksql dialect inherits ANSI, which also reserves `LEFT`/`RIGHT`, so editing the lists alone leaves them reserved via inheritance (there are already 6 keywords like `ROWS`/`SET` double-listed as unreserved but effectively reserved for this reason). To make the Spark lists authoritative I replaced the old inline `difference_update(["LEFT", "RIGHT"])` with a single generic line that drops from the inherited reserved set anything Spark declares unreserved:

```python
sparksql_dialect.sets("reserved_keywords").difference_update(UNRESERVED_KEYWORDS)
```

This is the same reconciliation pattern used in dialect_vertica.py / dialect_postgres.py / dialect_db2.py, keeps `LEFT`/`RIGHT [OUTER] JOIN` working, and regenerating the sparksql fixtures produces zero diffs. Happy to scope it back to just `["LEFT", "RIGHT"]` if you'd prefer the narrower form.

keraion (approved):
Thanks for making the change! LGTM

PR #8055, opened 2026-07-02
Don't auto-fix ST07 for BigQuery hyphenated table names

### LINKED ISSUE #7230: Rule structure.using (ST07) breaks some BigQuery statements

CodeBlackwell:
Fixes #7230

When ST07 rewrites a `USING (...)` join into an explicit `ON` clause, it builds each side of the condition from the table's `ref_str`. For a multi-part BigQuery reference like `project-a.dataset-b.table-c`, `ref_str` is only the final component — `table-c`. That component contains a hyphen, so it isn't a valid naked identifier, and splicing it into the generated clause produced invalid SQL.

Before:

```sql
SELECT * FROM project-a.dataset-b.table-c JOIN dataset-c.table-d USING (a);
```

got "fixed" to `... ON table-c.a = table-d.a`, which no longer parses.

After: the rule still flags the `USING` clause, but if either side's `ref_str` isn't a valid naked identifier it returns the unfixable result instead of offering a broken fix — matching how ST07 already declines other cases it can't safely rewrite.

I added a regression test in `test/rules/std_ST07_test.py` that lints the hyphenated-name example under the BigQuery dialect and asserts the violation is reported with no fixes. The full ST07 suite passes.

---
## Summary by cubic
Prevent ST07 from proposing invalid ON rewrites for BigQuery joins with hyphenated table names, and avoid ambiguous rewrites when USING columns are referenced unqualified.

- **Bug Fixes**
  - Replace the regex guard with a parsed-segment check; only fix when the alias is a single identifier segment and segment.raw == ref_str. Blocks BigQuery hyphenated refs while keeping valid cases.
  - Decline autofix if any USING column is referenced unqualified. Added tests for hyphenated names (no fix), unqualified references (no fix), and qualified references (still fixed).

<sup>Written for commit 1d29df5cd3a7cc8656c14f956cb8bbd76b5061b7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8055?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/rules/structure/ST07.py
+++ b/src/sqlfluff/rules/structure/ST07.py
@@ -1,6 +1,5 @@
 """Implementation of Rule ST07."""
 
-import re
 from typing import Optional
 
 from sqlfluff.core.parser import (
@@ -122,14 +121,19 @@ class Rule_ST07(BaseRule):
         to_delete, insert_after_anchor = _extract_deletion_sequence_and_anchor(segment)
 
         table_a, table_b = table_aliases[:2]
-        # ref_str is only the last part of a multi-part reference. For BigQuery
-        # hyphenated names it is e.g. "table-c", which is not a valid naked
-        # identifier, so splicing it into a generated ON clause yields broken SQL.
-        if not (
-            _NAKED_IDENTIFIER.match(table_a.ref_str)
-            and _NAKED_IDENTIFIER.match(table_b.ref_str)
-        ):
-            return unfixable_result
+        # Only splice ref_str into the ON clause if it is the raw of a single
+        # parsed identifier segment (BigQuery hyphenated names are not).
+        for table in (table_a, table_b):
+            if not table.segment or table.segment.raw != table.ref_str:
+                return unfixable_result
+
+        using_cols = _extract_cols_from_using(segment, using_anchor)
+        # An ON rewrite makes unqualified references to USING columns
+        # ambiguous (issue #7230), so decline to fix.
+        using_cols_upper = {col.upper() for col in using_cols}
+        for ref in select_info.reference_buffer if select_info else []:
+            if ref.raw_upper in using_cols_upper:
+                return unfixable_result
 
         edit_segments = [
             KeywordSegment(raw="ON"),
@@ -137,7 +141,7 @@ class Rule_ST07(BaseRule):
         ] + _generate_join_conditions(
             table_a.ref_str,
             table_b.ref_str,
-            _extract_cols_from_using(segment, using_anchor),
+            using_cols,
         )
 
         assert table_a.segment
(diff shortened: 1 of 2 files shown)


peterbud (changes requested):
Hi @CodeBlackwell - thanks for this PR. 

I agree with the AI feedback: using a hardcoded regex for identifying safe / unsafe table references is not a safe solution for other dialects or more complex constructs. like TVF or similar.

CodeBlackwell:
Agreed, regex was the wrong tool. Reworked in 1d29df5e:

- guard now checks ref_str matches the raw of a single parsed identifier segment, no lexical rules. TSQL temp tables / quoted identifiers keep fixes, BigQuery hyphenated names (which span multiple segments) get declined
- also noticed #7230 is actually a different bug: USING -> ON leaves bare `field_1` in the select list ambiguous. Parses fine so the post-fix check misses it. Now unfixable if any USING column is referenced unqualified
- tests for both + the happy path

peterbud:
Hi @CodeBlackwell thanks for the update.

At ST07.py, the new ambiguity guard only checks whether `ref.raw_upper` exactly matches a bare `USING` column name. That avoids autofixing `SELECT field_1 ... USING (field_1)`, but it still misses quoted unqualified references like `SELECT "field_1" ... USING (field_1)` or BigQuery ``SELECT `field_1` ... USING (field_1)``. Those are still unqualified references, so rewriting `USING` to `ON` changes the semantics in the same way, but this code will currently offer a fix anyway because the reference buffer contains `"FIELD_1"` / `` `FIELD_1` `` rather than `FIELD_1`.

I think this means the new logic is still incomplete: it fixes one spelling of the ambiguous case, not the ambiguous case itself. Can we base the check on whether the reference is unqualified, rather than string equality against the raw text? Also, these new cases look like a better fit for the existing ST07 fixture coverage in ST07.yml than a separate Python-only test file in std_ST07_test.py, since the fixture harness already covers dialect-specific no-fix cases and stronger fix expectations.

peterbud:
I guess no more tokens left

PR #8056, opened 2026-07-02
Fix ST06 reordering SELECT columns in CREATE VIEW with an explicit column list

### LINKED ISSUE #6807: Inconsistent view columns re-ordering causing mismatch between view column lists and SELECT column order
### Search before asking

### What Happened

We have views defined with below syntax: 
`create or replace view {MY_VIEW} ( col1, col2, col3, col4) as SELECT col1, col2, case when condition then col3 else col5, col4`

When running sql-fluff it gets linted into: 
`create or replace view {MY_VIEW} (col1, col2, col3, col4) as SELECT col2, col1, col4, case when condition then col3 else col5`

This cause some columns misalignment between view definition and selection.

### Expected Behaviour

I would expect the linter to raise an alert for the view definition, and to remove the definition in case it would change the selected columns ordering. 

### Observed Behaviour

Running sql-fluff breaks the view columns ordering, results in the view's `col3` to expose actual values of `col4`

`create or replace view {MY_VIEW} (col1, col2, col3, col4) as SELECT col2, col1, col4, case when condition then col3 else col5`

### How to reproduce

Create a view with a case when statement in the first selected column, and run sql-fluff fix 

### Dialect

Snowflake

### Version

3.3.1

### Configuration

[sqlfluff]
dialect = snowflake
max_line_length = 110
processes = -1

capitalisation_policy = lower
capitalisation_policy = lower

preferred_not_equal_style = c_style
ignore_words = account,owner

### Code of Conduct

CodeBlackwell:
ST06 was reordering the SELECT columns of a CREATE VIEW that declares an explicit column list. For a view like

```sql
CREATE VIEW v (col1, col2, col3, col4)
AS SELECT col1, col2, CASE WHEN condition THEN col3 ELSE col5 END, col4 FROM t
```

the rule moves the CASE expression to the end, which silently misaligns the view's declared columns against the columns the SELECT actually returns — `col3` ends up exposing `col4`'s values. This is #6807.

The problem is in `_is_view_with_explicit_columns()`, which is supposed to make ST06 skip these views. Two things stopped it from working:

- A bracketed SELECT body (`AS (SELECT ...)`) is structurally the same as an explicit column list — both are bracketed segments containing column references — so the guard could be tripped by the parenthesised SELECT rather than a real column list. I split the view body out into its own segment type (`CreateViewSelectableSegment`) so a parenthesised SELECT is no longer a bare `bracketed` direct child of the create view statement, and the detection loop now skips any bracketed child that contains a select/with statement.
- Detection only looked for `column_reference` and `index_column_definition`. BigQuery represents an explicit column list with `column_definition` and ClickHouse with a plain `identifier`, so those dialects were never recognised. Both are handled now.

Existing YAML rule cases for CREATE VIEW with explicit columns continue to pass, and the parse fixtures for the affected dialects (ansi, athena, vertica) are updated to reflect the new view-body segment. Full test suite is green.

Fixes #6807

---
## Summary by cubic
Fixes ST06 reordering SELECT columns in CREATE VIEW statements that declare an explicit column list. Correctly ignores parenthesized SELECT bodies to prevent view column misalignment. Fixes #6807.

- **Bug Fixes**
  - Added `CreateViewSelectableSegment` so `AS (SELECT ...)` is not treated as a bracketed column list.
  - Extended explicit column-list detection to BigQuery (`column_definition`) and ClickHouse (`identifier`).
  - Updated parse fixtures (ansi, athena, vertica) and added ST06 rule cases for the new scenarios.

<sup>Written for commit 0efb1cb57ce2da946724d6609d753b401f2fa183. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8056?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_ansi.py
+++ b/src/sqlfluff/dialects/dialect_ansi.py
@@ -3421,6 +3421,25 @@ class AlterTableStatementSegment(BaseSegment):
     )
 
 
+class CreateViewSelectableSegment(BaseSegment):
+    """The SELECT body of a `CREATE VIEW` statement, optionally bracketed.
+
+    Kept as its own segment type (rather than inlining
+    ``OptionallyBracketed(Ref("SelectableGrammar"))`` directly into
+    ``CreateViewStatementSegment``) so that a parenthesised SELECT body,
+    e.g. ``CREATE VIEW v AS (SELECT col1, col2 FROM t)``, is never a bare
+    ``bracketed`` segment at the direct-child level of
+    ``create_view_statement``. Without this wrapper, that parenthesised
+    SELECT is structurally indistinguishable from an explicit bracketed
+    column list (``CREATE VIEW v (col1, col2) AS ...``) to any code that
+    checks direct children for a "bracketed" segment containing
+    column references.
+    """
+
+    type = "create_view_selectable"
+    match_grammar: Matchable = OptionallyBracketed(Ref("SelectableGrammar"))
+
+
 class CreateViewStatementSegment(BaseSegment):
     """A `CREATE VIEW` statement."""
 
@@ -3437,7 +3456,7 @@ class CreateViewStatementSegment(BaseSegment):
         # Optional list of column names
         Ref("BracketedColumnReferenceListGrammar", optional=True),
         "AS",
-        OptionallyBracketed(Ref("SelectableGrammar")),
+        Ref("CreateViewSelectableSegment"),
         Ref("WithNoSchemaBindingClauseSegment", optional=True),
     )
 
--- a/src/sqlfluff/rules/structure/ST06.py
+++ b/src/sqlfluff/rules/structure/ST06.py
@@ -66,6 +66,13 @@ class Rule_ST06(BaseRule):
                 # Look for a bracketed segment containing column references
                 for child in parent.segments:
                     if child.is_type("bracketed"):
+                        # A bracketed SELECT body (e.g. `AS (SELECT ...)`) is
+                        # not an explicit column list, even though it may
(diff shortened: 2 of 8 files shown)


src/sqlfluff/dialects/dialect_ansi.py
```diff
@@ -3421,6 +3421,25 @@ class AlterTableStatementSegment(BaseSegment):
     )
 
 
+class CreateViewSelectableSegment(BaseSegment):
```
peterbud, line 3424: Why is this needed?
The reported regression is covered by the new BigQuery and ClickHouse ST06 cases in `ST06.yml`, but BigQuery overrides CREATE VIEW locally and ClickHouse has its own CREATE VIEW grammar, so neither dialect relies on the new ANSI wrapper.

peterbud, line 3424: Why is this needed?
The reported regression is covered by the new BigQuery and ClickHouse ST06 cases in `ST06.yml`, but BigQuery overrides CREATE VIEW locally and ClickHouse has its own CREATE VIEW grammar, so neither dialect relies on the new ANSI wrapper.

peterbud:
Closing due to inactivity

PR #8057, opened 2026-07-02
Don't strip newlines that sit directly before a comment

### LINKED ISSUE #6801: Formatting `AND` followed by blank/comment line with `binary_operation` `spacing_after="any"` changes meaning of the code
### Search before asking

### What Happened

Formatting an expression that has an `AND` followed by a line that's empty or comment only and then by an expression (like the two following queries) places the `AND` and the expression together without any spacing (`ANDb`), when using `binary_operation` `spacing_after = "any"`.

```sql
SELECT a AND

b
```
```sql
SELECT a AND
-- some comment
b
```

These following files are equivalent SQL, and do not exhibit the problematic behaviour when formatted (there remains at least one space between the `AND` and `b` after formatting):

```sql
SELECT a AND b
```

```sql
SELECT a AND
b
```

(Thank you for sqlfluff!)

### Expected Behaviour

The `AND` to have a space after it

### Observed Behaviour

The output is: (note the `ANDb` with no spacing)

```sql 
SELECT
    a

    ANDb
```

This changes meaning to the original code: 
- original equivalent to `SELECT (a AND b)`  (i.e. and-ing two columns)
- formatted equivalent to `SELECT a AS ANDb` (i.e. renaming the `a` column, no mention of `b`).

### How to reproduce

Run `sqlfluff format` using these two files:

`pyproject.toml`:

```toml
[tool.sqlfluff.core]
dialect = "ansi"
[tool.sqlfluff.layout.type.binary_operator]
spacing_after = "any"
```

`example.sql`:

```sql
SELECT a AND

b
```

Executable reproducer: (save it and run like `bash script.sh`)

```shell
#!/bin/bash

cd $(mktemp -d)

# Prepare the inputs:
cat > pyproject.toml <<EOF
[tool.sqlfluff.core]
dialect = "ansi"
[tool.sqlfluff.layout.type.binary_operator]
spacing_after = "any"
EOF

cat > example.sql <<EOF
SELECT a AND

b
EOF

# Install sqlfluff 3.3.1
python3.13 -m venv venv
. venv/bin/activate

# (specific versions of dependencies for when I reproduced the bug)
pip install sqlfluff==3.3.1 chardet==5.2.0 click==8.1.8 colorama==0.4.6 diff_cover==9.2.4 exceptiongroup==1.2.2 iniconfig==2.1.0 Jinja2==3.1.6 MarkupSafe==3.0.2 packaging==24.2 pathspec==0.12.1 platformdirs==4.3.7 pluggy==1.5.0 Pygments==2.19.1 pytest==8.3.5 PyYAML==6.0.2 regex==2024.11.6 tblib==3.1.0 toml==0.10.2 tomli==2.2.1 tqdm==4.67.1


# BUG: results in the final SQL containing `ANDb` with no space
sqlfluff format
cat example.sql
# SELECT
#     a
#
#     ANDb
```

### Dialect

ansi and postgres

### Version

- `sqlfluff, version 3.3.1`
- Python 3.13

### Configuration

```toml
[tool.sqlfluff.core]
dialect = "ansi"
[tool.sqlfluff.layout.type.binary_operator]
spacing_after = "any"
```

### Code of Conduct

CodeBlackwell:
Fixes #6801

### What was happening

When a layout constraint asks for newlines to be stripped between two tokens (for example `spacing_within = single:inline` on an `expression`), the reflow code could delete a newline that sat right in front of a comment. The result was that the token before the comment got glued straight onto the `--` marker:

```sql
SELECT * FROM t WHERE a AND
-- comment
b
```

came back as

```sql
SELECT * FROM t WHERE a AND-- comment
b
```

which quietly changes what the statement means, since everything after `--` is now commented out on that line.

### Root cause

In `determine_constraints` (`src/sqlfluff/utils/reflow/respace.py`), the guard that keeps newlines around comments only looked at the *previous* block. If the previous block was a comment it kept the newline, but if the *next* block was the comment it did nothing, so an inherited `strip_newlines=True` went ahead and removed the newline in front of it.

### The fix

Check the next block for comment segments as well as the previous one. If either side of the point is a comment, we leave the newline alone. A newline that sits immediately before a comment now survives the respace.

### Before / after

- Before: `a AND\n-- comment\nb` -> `a AND-- comment\nb` (meaning changed)
- After: `a AND\n-- comment\nb` -> unchanged

### Tests

Added `test_reflow__respace_does_not_strip_newline_before_comment` in `test/utils/reflow/respace_test.py`, which sets the `single:inline` expression constraint and asserts the newline before the comment is preserved. The existing reflow suite (`test/utils/reflow/`) still passes.

---
## Summary by cubic
Stop stripping a newline that sits directly before a comment during respace. Prevents code being glued to the `--` marker and changing SQL meaning. Fixes #6801.

- **Bug Fixes**
  - In `determine_constraints`, also check the next block for comment segments; if previous or next is a comment, do not strip newlines.
  - Added regression test `test_reflow__respace_does_not_strip_newline_before_comment`; existing reflow tests still pass.

<sup>Written for commit 56e06b5378f0640c4c72fa162fdead8c0da854d1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8057?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/utils/reflow/respace.py
+++ b/src/sqlfluff/utils/reflow/respace.py
@@ -186,8 +186,12 @@ def determine_constraints(
             within_spacing, strip_newlines = _unpack_constraint(
                 within_constraint, strip_newlines
             )
-        # Prohibit stripping newlines after comment segments
-        if any(seg.is_type("comment") for seg in prev_block.segments):
+        # Prohibit stripping newlines adjacent to comment segments (either
+        # immediately before or immediately after this point), since doing
+        # so could glue code to a comment marker and change meaning.
+        if any(seg.is_type("comment") for seg in prev_block.segments) or any(
+            seg.is_type("comment") for seg in next_block.segments
+        ):
             strip_newlines = False
 
     # If segments are expected to be touch within. Then modify
--- a/test/utils/reflow/respace_test.py
+++ b/test/utils/reflow/respace_test.py
@@ -7,7 +7,7 @@ import logging
 
 import pytest
 
-from sqlfluff.core import Linter
+from sqlfluff.core import FluffConfig, Linter
 from sqlfluff.utils.reflow.elements import ReflowPoint
 from sqlfluff.utils.reflow.helpers import fixes_from_results
 from sqlfluff.utils.reflow.sequence import ReflowSequence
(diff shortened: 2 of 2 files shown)


peterbud (approved):
Thanks @CodeBlackwell !

PR #7759, opened 2026-04-17
fix(postgres): handle COPY FROM STDIN data blocks

RedZapdos123:
Issue #7697 reported that PostgreSQL COPY ... FROM STDIN blocks could fail parsing because raw data lines after the COPY statement were parsed as SQL, causing PRS errors.

This PR updates the Postgres dialect to treat COPY FROM STDIN data blocks (up to the \\. terminator) as a dedicated file-level statement, so raw data is not parsed as SQL.

It also adds a focused regression test that covers COPY FROM STDIN raw-data block parsing.

Closes #7697.

## Checklist:

- [x] I have reviewed all changes in this PR myself.
- [x] I have added relevant regression test cases for this fix.
- [x] I have run linting and formatting checks in WSL using:
      ruff check src/sqlfluff/dialects/dialect_postgres.py test/dialects/postgres_test.py
      ruff format --check src/sqlfluff/dialects/dialect_postgres.py test/dialects/postgres_test.py
- [x] I have run focused tests in WSL using:
      pytest -q test/dialects/postgres_test.py -k copy_from_stdin_data_block_parses
      pytest -q test/dialects/postgres_test.py

#### The validation screenshots of the tests run, locally on WSL:

image

PR #8059, opened 2026-07-03
fix(postgres): parse COPY ... FROM STDIN inline data blocks

### LINKED ISSUE #7697: Support for ignoring raw data blocks after COPY ... FROM stdin;
### Search before asking

### What Happened

When parsing PostgreSQL dump files or fixture scripts that use `COPY ... FROM stdin;` followed by raw tab-separated data, `sqlfluff` attempts to parse the raw data as SQL. Because it is raw data and not standard SQL syntax, it immediately throws a `PRS` error.

### Expected Behaviour

When `sqlfluff` encounters `COPY [table] FROM stdin;`, it should ideally ignore all subsequent lines until it reaches the standard `\.` terminator, recognizing that the intervening lines are raw data and not SQL syntax to be linted.

### Observed Behaviour

`sqlfluff` throws a `PRS` (Found unparsable section) error at the very first line of the raw data.

```bash
$ sqlfluff lint --dialect postgres scripts/dev-instance/patron_data.sql
== [scripts/dev-instance/patron_data.sql] FAIL
L:  65 | P:   1 |  PRS | Line 65, Position 1: Found unparsable section:
                       | 'openlibrary\t15298516\t1\t\\N\t\\N\t2025-01-02 ...'

### How to reproduce

1. Create a file named ex- test1.sql
   ```sql
   COPY public.bookshelves_books (username, work_id, bookshelf_id) FROM stdin;
   openlibrary	15298516	1
   openlibrary	45310	3
   \.
   ```
2. Run the following command in your terminal:
   ```bash
   sqlfluff lint --dialect postgres test1.sql
   ```

### Dialect

postgres

### Version

4.1.0

### Configuration

N/A

### Code of Conduct

truffle-dev:

`pg_dump` and `psql` emit `COPY ... FROM STDIN;` followed by the table data inlined as tab-separated rows, terminated by a line containing only `\.`. Those raw rows are not SQL, so sqlfluff surfaced the first data line as an unparsable section (`PRS`), as reported in #7697.

This lexes the whole construct as a single token and parses it as a dedicated file-level statement, so the raw data no longer reaches the parser. The lexer pattern is self-anchored on the `COPY ... FROM STDIN;` prefix and requires the trailing `\.` line, so it stays context-free: a `COPY ... FROM STDIN` without an inline data block still parses as a normal statement.

Fixture-based regression test added at `test/fixtures/dialects/postgres/copy_from_stdin_data_block.sql` (+ auto-generated `.yml`), covering two back-to-back data blocks followed by an ordinary `SELECT` to confirm parsing resumes after the terminator.

This re-opens the fix from #7759 by @RedZapdos123 in the fixture-based form the [dialect contribution guide](https://docs.sqlfluff.com/en/stable/guides/contributing/dialect.html) asks for, per @peterbud's note there. Original report by @Sanket17052006, MRE by @RedZapdos123.

Fixes #7697.

No. The new lexer matcher only fires on the full `COPY ... FROM STDIN; ... \.` shape; existing `COPY` and `\copy` handling is untouched.

### Pull Request checklist

  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated by running `python test/generate_parse_fixture_yml.py`).

---
## Summary by cubic
Parse Postgres COPY ... FROM STDIN inline data blocks, including multiline headers and quoted options with semicolons, so they no longer show as unparsable. Parsing resumes after the `\.` terminator. Fixes #7697.

- **Bug Fixes**
  - Added a lexer rule `postgres_copy_stdin_data_block` that captures `COPY ... FROM STDIN;` through the `\.` line and parses it as `postgres_copy_stdin_data_statement`.
  - Handles multiline headers and semicolons inside quoted literals; normal `COPY` and `\copy` remain unchanged.
  - Added fixtures for back-to-back data blocks, a multiline header, an option with `';'`, and a trailing `SELECT`.

<sup>Written for commit 8d37d938456fbb8aa102ef214dec7da7f8862071. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8059?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -219,13 +219,14 @@ postgres_dialect.insert_lexer_matchers(
             # so it stays context-free; a `COPY ... FROM STDIN` without a trailing
             # `\.` data block does not match and parses as a normal statement.
             # The header may span multiple lines (e.g. a column list or `WITH`
-            # options on their own lines). `[^;]` rather than `[^\r\n]` allows the
-            # newlines while still stopping at the statement terminator, so the
-            # match cannot absorb a preceding statement to reach a later
-            # `FROM STDIN`.
+            # options on their own lines). The scan toward the terminating `;`
+            # must also skip over quoted string literals so option values like
+            # `DELIMITER ';'` do not prematurely end the match. This still keeps
+            # the match bounded to one statement, so it cannot absorb a preceding
+            # statement to reach a later `FROM STDIN`.
             # Approach originally proposed in #7759 by @RedZapdos123.
             "postgres_copy_stdin_data_block",
-            r"(?i)COPY\b[^;]*?\bFROM\b\s+STDIN\b[^;]*?;[ \t]*\r?\n"
+            r"(?i)COPY\b(?:[^;']|'(?:[^']|'')*')*?\bFROM\b\s+STDIN\b(?:[^;']|'(?:[^']|'')*')*?;[ \t]*\r?\n"
             r"(?:[^\r\n]*\r?\n)*?\\\.[ \t]*(?=\r?\n|$)",
             CodeSegment,
         ),
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -211,6 +211,20 @@ postgres_dialect.insert_lexer_matchers(
             LiteralSegment,
         ),
         RegexLexer(
+            # Matches a `COPY ... FROM STDIN;` command followed by an inline data
+            # block terminated by a line containing only `\.` (the format emitted
+            # by pg_dump / psql). The raw data rows are not SQL, so the whole
+            # construct is lexed as a single token to keep it out of the parser.
+            # The pattern is self-anchored on the `COPY ... FROM STDIN;` prefix,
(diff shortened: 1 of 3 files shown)


truffle-dev:
Good catch on the multiline header. Fixed in `a5c9fe8`: the two header spans now use `[^;]` instead of `[^\r\n]`, so a column list or `WITH (...)` clause on its own line matches, while the header still stops at the statement terminator.

I went with `[^;]` rather than a plain `[\s\S]` on purpose: an unrestricted newline match would let a lazy `COPY[\s\S]*?FROM STDIN` absorb a preceding statement to reach a later `FROM STDIN` (e.g. `COPY a FROM src;\nCOPY b FROM STDIN;`). `[^;]` can't cross the terminating semicolon, so the block stays anchored to its own statement. Verified that case parses the first `COPY` as a separate section.

Added a multiline case to the existing fixture; all 495 postgres dialect tests pass, and the new case goes red if the pattern is reverted to `[^\r\n]`.

peterbud (approved):
Thanks!

PR #8066, opened 2026-07-03
feat(mysql): support GROUP BY ... WITH ROLLUP modifier

Synvoya:
## Brief summary of the change made

MySQL's [`GROUP BY ... WITH ROLLUP`](https://dev.mysql.com/doc/refman/8.0/en/group-by-modifiers.html) was reported as an unparsable section:

```sql
SELECT a, SUM(b) FROM t GROUP BY a WITH ROLLUP;
--                                  ^^^^^^^^^^^ Found unparsable section: 'WITH ROLLUP'
```

The `mysql` dialect inherited ansi's `GroupByClauseSegment`, which only supports the `ROLLUP(...)` function form, not MySQL's postfix `WITH ROLLUP` modifier.

This overrides `GroupByClauseSegment` in the `mysql` dialect, mirroring the ansi grammar and appending an optional trailing `WITH ROLLUP`.

## Are there any other side effects?

No. The override is scoped to the `mysql` dialect and existing `GROUP BY` forms parse unchanged (all pre-existing mysql fixtures still pass).

## Which tests were run / added?

- New dialect fixture `test/fixtures/dialects/mysql/group_by_with_rollup.sql` (+ generated `.yml`) covering single-column, multi-column, `HAVING`, and `ORDER BY` variants.
- `pytest test/dialects/dialects_test.py -k mysql` → **632 passed, 0 failed**.
- `ruff check` / `ruff format --check` clean.

## AI-assisted contribution disclosure

Per the CONTRIBUTING guide's *AI-Assisted Contributions* policy: this change was prepared with AI assistance. It was reviewed, tested, and verified locally before submission.

---
## Summary by cubic
Adds support for MySQL’s GROUP BY ... WITH ROLLUP so these queries parse correctly. Scoped to the `mysql` dialect.

- New Features
  - Overrode `GroupByClauseSegment` in `mysql`, extending `ansi.GroupByClauseSegment.match_grammar` via `.copy(insert=[...])` to allow optional trailing WITH ROLLUP.
  - Added fixtures covering single/multi column, HAVING, and ORDER BY variants.

<sup>Written for commit 9f2e711c9db6b1a90cb004c4f31f722320b85c07. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8066?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_mysql.py
+++ b/src/sqlfluff/dialects/dialect_mysql.py
@@ -372,6 +372,36 @@ mysql_dialect.add(
 )
 
 
+class GroupByClauseSegment(ansi.GroupByClauseSegment):
+    """A MySQL `GROUP BY` clause, adding the `WITH ROLLUP` modifier.
+
+    https://dev.mysql.com/doc/refman/8.0/en/group-by-modifiers.html
+    """
+
+    match_grammar: Matchable = Sequence(
+        "GROUP",
+        "BY",
+        Indent,
+        OneOf(
+            "ALL",
+            Ref("GroupingSetsClauseSegment"),
+            Ref("CubeRollupClauseSegment"),
+            Sequence(
+                Delimited(
+                    OneOf(
+                        Ref("ColumnReferenceSegment"),
+                        Ref("NumericLiteralSegment"),
+                        Ref("ExpressionSegment"),
+                    ),
+                    terminators=[Ref("GroupByClauseTerminatorGrammar")],
+                ),
+            ),
+        ),
+        Dedent,
+        Sequence("WITH", "ROLLUP", optional=True),
+    )
+
+
 class TableReferenceSegment(ansi.TableReferenceSegment):
     """A reference to a table.
 
--- a/src/sqlfluff/dialects/dialect_mysql.py
+++ b/src/sqlfluff/dialects/dialect_mysql.py
@@ -378,27 +378,8 @@ class GroupByClauseSegment(ansi.GroupByClauseSegment):
     https://dev.mysql.com/doc/refman/8.0/en/group-by-modifiers.html
     """
 
-    match_grammar: Matchable = Sequence(
-        "GROUP",
-        "BY",
-        Indent,
-        OneOf(
-            "ALL",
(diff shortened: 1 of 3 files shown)


Synvoya:
Thanks @cubic-dev-ai — good catch. Switched to `ansi.GroupByClauseSegment.match_grammar.copy(insert=[Sequence("WITH", "ROLLUP", optional=True)])`, matching the `FunctionSegment` pattern in this file. The parse tree is identical, so the fixtures are unchanged, and the mysql dialect tests + ruff stay green.

peterbud (changes requested):
Hi @Synvoya  - thanks for opening this PR.

Unfortunately this change (copying ANSI and then appending a ROLLUP segment) is not conform with other GROUP BY approaches, and poses a problem:

The new MySQL override appends WITH ROLLUP after ANSI’s existing `Dedent`, so it escapes the GROUP BY indentation block instead of staying inside it.

So for example this:
```sql
GROUP BY
    a,
    b WITH ROLLUP
```

will be fixed to this:
```sql
GROUP BY
    a,
    b
WITH ROLLUP
```

Which is not correct.

peterbud:
Closing due to inactivity

Synvoya:
Reopening this wasn't possible on my end, so I've addressed your review in #8130. `WITH ROLLUP` is now inserted before ANSI's trailing `Dedent` (`copy(insert=[...], before=Dedent)`), so it stays inside the `GROUP BY` indentation block — your `b WITH ROLLUP` example now lints clean under LT02 (it previously reported "Expected line break and no indent before 'WITH'"), and I added a regression case that fails without the change. Thanks for the detailed feedback, and sorry for the slow turnaround.

PR #8069, opened 2026-07-05
feat(rust): port CP03 and CP04 to Rust-native detection

WittierDinosaur:

Extends CP01's Rust-native detection proof-of-concept (#7984) to two more capitalisation rules — CP03 (function names) and CP04 (boolean/null literals) — gated the same way, behind `core.use_rust_rules` (default `False`).

Both fit CP01's existing leaf target/exclude/policy model directly, so this is mostly reuse:

- **Generalized CP01's walker** into a shared `sqlfluffrs_rules::capitalisation` module, parameterized by target/exclude segment types. CP01/CP03/CP04's own `.rs` files are now thin wrappers over one refute/resolve/apply state machine instead of three copies of it.
- **Added pascal/camel/snake policy support** to the Rust side (needed for CP03's `extended_capitalisation_policy`). Python's transforms use regex lookbehind/lookahead the Rust `regex` crate can't express, so these are hand-written char-iteration functions mirroring the exact Python behaviour (word-boundary casing for pascal/camel, underscore-insertion + `str.isupper()` special-case for snake).
- **New `cp03_violations`/`cp04_violations` PyO3 bindings**, registered alongside the existing `cp01_violations`.
- **CP03 and CP04 each get their own `_eval_rust` override** rather than sharing CP01's inherited dispatch — both delegate to a new `Rule_CP01._eval_rust_capitalisation(context, violations_fn, policy)` helper for the FFI-call + leaf-index-to-segment plumbing shared across the bundle. This is deliberate: CP01's original rollout needed a `self.name` guard to stop CP04 (which shares `capitalisation_policy` with CP01) from being silently hijacked into CP01's keyword detection. Giving each rule its own override removes that failure mode by construction — there's no shared dispatch method left for a future rule to leak through.
- Matching Rust detection unit tests (`test/core/parser/rust_cp03_arena_test.py`, `rust_cp04_arena_test.py`) and Python dispatch-parity tests (`cp03_rust_dispatch_test.py`, `cp04_rust_dispatch_test.py`), mirroring CP01's existing test files. Removed two tests from `cp01_rust_dispatch_test.py` whose docstrings assumed CP03/CP04 had no Rust path, since that's no longer true and is now covered more thoroughly by the dedicated files.

**Not covered here:** CP02 (identifiers) needs an ancestor-walk for `identifiers_policy_applicable` plus a Databricks/SparkSQL dialect special-case; CP05 (types) has a different crawl shape entirely (scans a container's children rather than a leaf target/exclude set). Both deferred to follow-up work.

### Performance

Detection-only timings for the two rules added in this PR (CP01's own numbers were already covered in #7984), measured the same way: `rule.crawl()` with the expensive shared fix-anchoring post-processing stubbed out, isolating the walk/eval cost that actually differs between the two paths — that post-processing is identical overhead on both sides otherwise and dominates a naive end-to-end comparison. Synthetic 400-statement file, median of 30 runs after 3 warmup iterations, `use_rust_rules` off vs on:

| Rule | Python (median) | Rust (median) | Speedup |
|---|---|---|---|
| CP03 (functions) | 42.2ms | 20.3ms | ~2.1x |
| CP04 (literals) | 57.5ms | 15.9ms | ~3.6x |

(Local run, not a committed benchmark — comparable in spirit to the informal timings quoted in the earlier CP01 arena-perf commits, e.g. `365a7d2e6`.)

None — fixing stays Python-side for all three rules, same as CP01 (leaf-index anchoring; arena and Python `BaseSegment` uuids aren't shared). No behavior change with the default `use_rust_rules=False`.

### Pull Request checklist

- [x] Included test cases: Rust detection parity tests + Python dispatch parity tests for both rules.
- [x] Ran the full CP01/CP03/CP04 YAML fixture suite (51 cases) — no regressions on the default Python path.
- [x] `cargo fmt`/`cargo clippy` clean on `sqlfluffrs_rules`; `ruff check`/`ruff format`/`mypy --strict` clean on the changed Python files.

---
## Summary by cubic
Ports CP03 (function names) and CP04 (boolean/null literals) to Rust-native detection via a shared `sqlfluffrs_rules::capitalisation` engine. Gated by `core.use_rust_rules`; default behavior is unchanged.

- **New Features**
  - Rust detection for CP03/CP04 via `sqlfluffrs.cp03_violations` and `sqlfluffrs.cp04_violations`, behind `core.use_rust_rules`.
  - Added `pascal`/`camel`/`snake` support for CP03 and a shared `_eval_rust_capitalisation` helper; new arena and dispatch parity tests.

- **Bug Fixes**
  - Match Python `str.capitalize()` by using Unicode titlecase for the first char (handles Dz/Dž/Lj/Nj digraphs).
  - Restore CP01 coverage via a CP02 fallback test and mark unreachable guards in CP03/CP04; minor perf tweak by avoiding a per-node Vec clone in the walker.

<sup>Written for commit 520dd578d51d0b7e7323346c1ff23019fb62bf05. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8069?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs.pyi
+++ b/sqlfluffrs/sqlfluffrs.pyi
@@ -266,3 +266,28 @@ def cp01_violations(
     parse tree's ``raw_segments`` order.
     """
     ...
+
+def cp03_violations(
+    tree: RsTree,
+    policy: str,
+    ignore_words: List[str] = ...,
+    ignore_templated: bool = ...,
+) -> List[Tuple[int, str]]:
+    """Detect CP03 (function-name capitalisation) violations natively over the arena.
+
+    Same contract as ``cp01_violations``; ``policy`` additionally accepts
+    ``pascal``/``camel``/``snake``.
+    """
+    ...
+
+def cp04_violations(
+    tree: RsTree,
+    policy: str,
+    ignore_words: List[str] = ...,
+    ignore_templated: bool = ...,
+) -> List[Tuple[int, str]]:
+    """Detect CP04 (boolean/null literal capitalisation) violations natively over the arena.
+
+    Same contract as ``cp01_violations``.
+    """
+    ...
--- a/sqlfluffrs/sqlfluffrs_rules/src/capitalisation.rs
+++ b/sqlfluffrs/sqlfluffrs_rules/src/capitalisation.rs
@@ -1,0 +1,295 @@
+//! Shared detection logic for the capitalisation rule bundle (CP01/CP03/CP04
+//! today; CP02/CP05 have extra logic — ancestor-based policy applicability and
+//! a container-scanning crawl shape, respectively — and are not yet ported).
+//!
+//! All three rules subclass `Rule_CP01` in Python and share its `_eval`/
+//! `_handle_segment` state machine: refute capitalisation cases inconsistent
+//! with what's been seen so far, resolve a concrete policy under `consistent`,
+//! apply it, and compare. This module holds that state machine once; each
+//! rule's arena walk differs only in which segment types it targets/excludes.
+
+use std::collections::HashSet;
+
+use sqlfluffrs_parser::{Arena, NodeId};
+
+/// `capitalisation_policy`/`extended_capitalisation_policy` options minus
+/// "consistent", in validation order. Only these three are ever inferred under
+/// `consistent` — pascal/camel/snake are refuted unconditionally in the Python
+/// rule (`refuted_cases.update(["camel", "pascal", "snake"])`) and so are only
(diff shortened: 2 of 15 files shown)


keraion (approved):
Nice work!

PR #8092, opened 2026-07-08
perf: convert PositionMarker from a frozen dataclass to a plain __slots__ class

rubytobi:
`PositionMarker` is constructed once per lexed token, so it sits on the lexer's hot path. This converts it from `@dataclass(frozen=True)` to a plain class using `__slots__` with a hand written `__init__`.

Frozen dataclasses enforce immutability by routing every field assignment in the generated `__init__` through `object.__setattr__()`, which is considerably slower to construct than direct attribute assignment on an ordinary class. All comparison operators, properties, and classmethods are unchanged.

Profiled with `cProfile`, isolating just the lex stage (`Lexer.lex` through `RsLexer._tokens_to_segments`) over a 991KB Athena SQL fixture. There were 710,095 `PositionMarker` constructions per run.

| Metric | Change |
| --- | --- |
| `PositionMarker.__init__` cumulative time | ↓ ~32% |
| `PositionMarker.__init__` self (tottime) | ↓ ~29% |

No

### Pull Request checklist

  - Other.

---
## Summary by cubic
Convert `PositionMarker` from a frozen dataclass to a lightweight `__slots__` class to speed up token lexing. Adds a manual `__init__` that infers working positions when unset; behavior is unchanged.

- **Refactors**
  - Replaced `@dataclass(frozen=True)` with a plain class using `__slots__` and a hand-written `__init__` that infers `working_line_no` and `working_line_pos` when unset.
  - Removed unused `__repr__`; comparisons and public behavior are unchanged.

- **Performance**
  - On a 991KB Athena SQL fixture (710,095 constructions): `__init__` cumulative time ↓ ~32%, self time ↓ ~29%.

<sup>Written for commit f9ddca9ba6f6255ae317d2e835c5f8ca717eecb8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8092?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/core/parser/markers.py
+++ b/src/sqlfluff/core/parser/markers.py
@@ -4,7 +4,6 @@ This class is a construct to keep track of positions within a file.
 """
 
 from collections.abc import Sequence
-from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Optional
 
 from sqlfluff.core.helpers.slice import zero_slice
@@ -14,7 +13,6 @@ if TYPE_CHECKING:  # pragma: no cover
     from sqlfluffrs import RsPositionMarker
 
 
-@dataclass(frozen=True)
 class PositionMarker:
     """A reference to a position in a file.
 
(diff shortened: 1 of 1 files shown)


peterbud (commented):
Hi @rubytobi  - thanks for this PR as well.

`PositionMarker` is publicly importable via sqlfluff.core.parser, so the “nothing in the codebase relies on this” argument may be a bit narrow here. This refactor intentionally changes externally observable behavior by dropping frozen-dataclass semantics. That may still be acceptable, but it seems worth either documenting `PositionMarker` as internal/unstable or explicitly noting the plugin compatibility risk in the PR.

I have limited experience with plugins, so I cannot really judge if this can be a real issue, or we can take this risk, so I'd rely on @alanmcruickshank

rubytobi:
Ooh, yeah! I've been totally ignoring the plugins, thanks for reminding me about them.

Thinking about it, I could reimplement most `dataclass(frozen=True)` functions, e.g hash, repr, replace, ... to get near but not identical functionality and still keep the performance gain.

rubytobi:
@peterbud, I'm happy to park this one. I've got some other ideas to pursue as well to improve performance. This here is not a must do and given the clear possible impact to the plugin surface, not worth the risk.

peterbud:
I'm afraid seeing the activity level of @alanmcruickshank this will be waiting for a while, apologies.

PR #8048, opened 2026-07-01
Fix ClickHouse GROUP BY modifiers

ishaanlabs-gg:
## Summary

- allow ClickHouse `GROUP BY ... WITH TOTALS`, `WITH ROLLUP`, and `WITH CUBE`
- add generated parser fixtures for the three modifier forms

Fixes #8037.

## Verification

- `./.venv-py313/bin/sqlfluff parse --dialect clickhouse test/fixtures/dialects/clickhouse/group_by_modifiers.sql`
- `./.venv-py313/bin/python -m pytest test/dialects/dialects_test.py -k 'clickhouse and group_by_modifiers'`
- `./.venv-py313/bin/python -m pytest test/dialects/dialects_test.py -k 'clickhouse'`
- `./.venv-py313/bin/python -m ruff check src/sqlfluff/dialects/dialect_clickhouse.py`
- `git diff --check`

---
## Summary by cubic
Enables ClickHouse GROUP BY modifiers (`WITH TOTALS`, `WITH ROLLUP`, `WITH CUBE`) with correct constraints. Supports valid single or combined forms and rejects duplicates or unsupported placements.

- **Bug Fixes**
  - Updated `GroupByClauseSegment` to parse modifiers and enforce order: optional `WITH ROLLUP | WITH CUBE` then optional `WITH TOTALS`.
  - Added `WithRollupOrCubeClauseSegment` and `WithTotalsClauseSegment`.
  - Added fixtures for valid forms and tests that reject invalid duplicates and unsupported combos (e.g., with `GROUPING SETS` or `ROLLUP()`/`CUBE()` forms).

<sup>Written for commit 1b7528a40d1fe7b96276ec87b6d0fb6a38b1ec77. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8048?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #8101, opened 2026-07-08
ClickHouse: parse GROUP BY WITH ROLLUP / CUBE / TOTALS (#8037)

### LINKED ISSUE #8037: ClickHouse: GROUP BY ... WITH TOTALS / ROLLUP / CUBE modifiers are unparsable
### What happened

ClickHouse's `GROUP BY` modifiers `WITH TOTALS`, `WITH ROLLUP` and `WITH CUBE`
produce an unparsable section in the `clickhouse` dialect.

### Expected behaviour

The statements parse; the modifier is recognised as part of the `GROUP BY`
clause.

### Observed behaviour

```
L1:P37 PRS: Found unparsable section: 'WITH TOTALS'
```

(equivalently `WITH ROLLUP` / `WITH CUBE`)

### How to reproduce

```sql
SELECT a, count() FROM t GROUP BY a WITH TOTALS
SELECT a, count() FROM t GROUP BY a WITH ROLLUP
SELECT a, count() FROM t GROUP BY a WITH CUBE
```

```bash
sqlfluff parse --dialect clickhouse -
```

All three are valid and were verified on ClickHouse 26.4.2.
Docs: https://clickhouse.com/docs/sql-reference/statements/select/group-by

### Dialect

clickhouse

### Version

sqlfluff 4.2.2 (latest release), Python 3.12. Also applies to `main` — the only
change to the ClickHouse dialect grammar since 4.2.2 is the ternary operator
(#8016).

### Configuration

None — default rules, `--dialect clickhouse`.

apoorvdarshan:

Fixes #8037. ClickHouse `GROUP BY` modifiers `WITH ROLLUP`, `WITH CUBE`, and `WITH TOTALS` produced an unparsable section error.

### How

Define a ClickHouse-local `GroupByClauseSegment` based on the ANSI grammar and place the ClickHouse modifiers before the clause's `Dedent`. This keeps `WITH ROLLUP` / `WITH CUBE` and optional `WITH TOTALS` within the `GROUP BY` indentation scope, so layout fixes don't incorrectly move them outside the clause.

No. The override is scoped to ClickHouse and plain `GROUP BY` remains unchanged.

Verified:

- Fixture coverage includes each modifier, `ROLLUP` / `CUBE` combined with `TOTALS`, and `WITH TOTALS` followed by `HAVING`.
- Added a regression test asserting `TOTALS` occurs between the group-by `Indent` and `Dedent` meta segments.
- ClickHouse-selected suite passes: **151 tests**.
- Fixture regeneration succeeds for all 50 ClickHouse fixture files.
- Ruff, mypy, doc8, yamllint, codespell, and all repository pre-commit hooks pass.

(Supersedes the stale #8048, which was closed for inactivity.)

### Pull request checklist

- [x] Added fixtures and regenerated YAML.
- [x] Added a grammar-shape regression test.
- [x] Lint and type checks are clean.

---

*Disclosure: prepared with AI assistance; reviewed and verified locally against the dialect test suite.*


peterbud (changes requested):
Hi @apoorvdarshan - thanks for the PR.

I understand your intention to keep code maintianable, and instead of redefining GroupByClausSegment, you opted for copy + insert. It's not a coincidence that other dialect don't do that, but rather redefine locally GroupBy, and the reason is Indent/Dedent. 

`ansi.GroupByClauseSegment.match_grammar.copy(insert=[...])` does not insert into the ANSI `GROUP BY` body; `BaseGrammar.copy()` appends at the end.

Parsing still succeeds, but the modifiers are outside the inherited `Indent`/`Dedent` scope, so they will be formatted badly. A concrete check with `SELECT a, count() FROM t GROUP BY a WITH TOTALS HAVING count() > 1;` shows the parse tree emits `[META] dedent` before `WITH TOTALS`, so it is a real grammar-shape bug. 

I had a similar recent comment: https://github.com/sqlfluff/sqlfluff/pull/8066#pullrequestreview-4637355154 

I'm afraid, you cannot avoid redefining / duplicating a local `GroupByClauseSegment`

peterbud (approved):
Thanks @apoorvdarshan !

PR #8104, opened 2026-07-08
Oracle: parse the SQL*Plus SHOW command (#8102)

### LINKED ISSUE #8102: [oracle] SHOW unparsable
### Search before asking

### What Happened

### Expected Behaviour

The example should be parsed correctly. Show doesnt seem to be supported at all currently
See: https://docs.oracle.com/en/database/oracle/oracle-database/26/sqpug/SHOW.html

### Observed Behaviour

Linting the example below gives the following error:
`Line 1, Position 1: Found unparsable section: 'show errors'`

SHOW is currently not supported at all.

### How to reproduce

Lint the following example:

```sql
show errors
``` 
Other options don't work as well.

### Dialect

oracle

### Version

4.2.1
python 3.11.3

### Configuration

[sqlfluff]
dialect = oracle

### Code of Conduct

apoorvdarshan:

Fixes #8102. The Oracle dialect didn't support the SQL*Plus `SHOW` command at all, so `show errors` (and every other `SHOW` option) produced a `Found unparsable section` error.

```sql
show errors;   -- was: PRS 'show errors'
```

### How

Add a `ShowStatementSegment` (registered in the Oracle `StatementSegment`) covering the documented [`SHOW`](https://docs.oracle.com/en/database/oracle/oracle-database/26/sqpug/SHOW.html) options:

- `SHOW ERRORS [ {FUNCTION | PROCEDURE | PACKAGE [BODY] | TRIGGER | VIEW | TYPE [BODY] | DIMENSION | JAVA SOURCE | JAVA CLASS} [schema.]name ]`
- `SHOW PARAMETER[S] [name]`
- the reserved-keyword options `ALL` and `USER`
- any other single-word option or SET system variable (`SGA`, `PDBS`, `RELEASE`, `RECYCLEBIN`, `SPOOL`, `LINESIZE`, `PAGESIZE`, …), matched via a naked identifier so **no new keywords are required**.

No. `SHOW` is a new statement; nothing else is affected. Verified:

- New fixture `oracle/show` (14 forms incl. `ERRORS`, `ERRORS FUNCTION ...`, `ERRORS PACKAGE BODY ...`, `PARAMETER[S]`, `RECYCLEBIN`, `ALL`, `USER`, `SGA`, `RELEASE`, `PDBS`, `CON_NAME`, `LINESIZE`); cases fail on `main` and pass here.
- Full Oracle dialect suite passes (291 tests).
- `ruff check` / `ruff format --check` clean; no keyword-set changes.

### Pull request checklist

- [x] Added a fixture and regenerated the YAML.
- [x] Reproduces/resolves the issue; new cases fail on `main`.
- [x] Lint clean; no regressions.

---

*Disclosure: prepared with AI assistance; reviewed and verified locally against the dialect test suite.*

---
## Summary by cubic
Adds parsing support for the Oracle SQL*Plus `SHOW` command and newline-terminated commands, removing “Found unparsable section” errors (e.g., `SHOW ERRORS`). Fixes #8102.

- Added `ShowStatementSegment` and registered it in the Oracle `StatementSegment`.
- Supports `SHOW ERRORS` with optional object type and `[schema.]name`, and `SHOW PARAMETER[S] [name]`.
- Accepts reserved options `ALL` and `USER`, plus any single-word option or SET variable (e.g., `SGA`, `PDBS`, `RELEASE`, `RECYCLEBIN`, `SPOOL`, `LINESIZE`, `LONG`, `NULL`).
- Excludes `SHOW`, `SET`, `PROMPT`, and `SELECT` from option names so newline-terminated `SHOW` statements don't consume a following `SELECT`.

<sup>Written for commit cfd786287b55c604a4066b6718bc5689e8dee975. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8104?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_oracle.py
+++ b/src/sqlfluff/dialects/dialect_oracle.py
@@ -1202,6 +1202,50 @@ class IndexTypeReferenceSegment(BaseSegment):
     match_grammar = ansi.ObjectReferenceSegment.match_grammar.copy()
 
 
+class ShowStatementSegment(BaseSegment):
+    """A SQL*Plus ``SHOW`` command.
+
+    https://docs.oracle.com/en/database/oracle/oracle-database/latest/sqpug/SHOW.html
+    """
+
+    type = "show_statement"
+
+    match_grammar = Sequence(
+        "SHOW",
+        OneOf(
+            # SHOW ERRORS [object_type [schema.]name]
+            Sequence(
+                "ERRORS",
+                Sequence(
+                    OneOf(
+                        "FUNCTION",
+                        "PROCEDURE",
+                        "TRIGGER",
+                        "VIEW",
+                        "DIMENSION",
+                        Sequence("PACKAGE", Ref.keyword("BODY", optional=True)),
+                        Sequence("TYPE", Ref.keyword("BODY", optional=True)),
+                        Sequence("JAVA", OneOf("SOURCE", "CLASS")),
+                    ),
+                    Ref("ObjectReferenceSegment"),
+                    optional=True,
+                ),
+            ),
+            # SHOW PARAMETER[S] [name]
+            Sequence(
+                OneOf("PARAMETERS", "PARAMETER"),
+                Ref("ParameterNameSegment", optional=True),
+            ),
+            # Reserved keywords that cannot be matched as a bare identifier.
+            "ALL",
+            "USER",
+            # Any other single-word option, or a SET system variable, e.g.
+            # SGA, PDBS, RELEASE, RECYCLEBIN, SPOOL, LINESIZE, PAGESIZE.
+            Ref("NakedIdentifierSegment"),
+        ),
+    )
+
+
 # Adding Oracle specific statements.
 class StatementSegment(ansi.StatementSegment):
     """A generic segment, to any of its child subsegments.
@@ -1255,6 +1299,7 @@ class StatementSegment(ansi.StatementSegment):
             Ref("AlterSynonymStatementSegment"),
             Ref("DropProfileStatementSegment"),
             Ref("DropClusterStatementSegment"),
+            Ref("ShowStatementSegment"),
         ],
     )
 
--- a/src/sqlfluff/dialects/dialect_oracle.py
+++ b/src/sqlfluff/dialects/dialect_oracle.py
@@ -189,6 +189,7 @@ oracle_dialect.add(
     ),
     PowerOperatorSegment=StringParser("**", SymbolSegment, type="binary_operator"),
     ModOperatorSegment=StringParser("MOD", WordSegment, type="binary_operator"),
+    ShowOptionSegment=TypedParser("word", CodeSegment, type="show_option"),
     OnCommitGrammar=Sequence(
         "ON",
         "COMMIT",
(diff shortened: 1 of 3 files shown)


devanshranjan10:
Ran into something while testing this branch that might be worth handling.

`SHOW` commands without a trailing semicolon don't delimit. SQL*Plus ends them at the newline, which the parser never sees, so the optional argument consumes the first word of the next command:

```sql
show parameters
show user
```

gives `Found unparsable section`. Same for `show parameters` followed by `set scan on`. `SHOW ERRORS` isn't affected, since its optional part needs an object type keyword first.

A newline terminator doesn't help, because newlines are non-code and get skipped. Excluding the command words from the argument does work, and matches the existing pattern in the ANSI dialect (`Ref("ExpressionSegment", optional=True, exclude=Ref.keyword("FROM"))`):

```python
Ref(
    "ParameterNameSegment",
    optional=True,
    exclude=OneOf("SHOW", "SET", "PROMPT"),
),
```

Every case in your fixture file ends in a semicolon, so this doesn't change CI either way. It only shows up in scripts that leave them off, which SQL*Plus allows.

Unrelated, but I noticed it nearby: the `PROMPT` lexer rule is case sensitive, so lowercase `prompt foo` is unparsable on `main` today. Separate issue, not yours.

rubytobi:
Hi @apoorvdarshan, thanks for looking into this. 

Can you please check the Oracle links? Both are `That page is not available.` for me.

Also, it looks like the below doesn't parse correctly after the dialect change. Based on my - limited - understanding this should parse tho:
```sql
show parameters
select 1 from dual;
```
```bash
L:   2 | P:   8 |  PRS | Line 2, Position 8: Found unparsable section: '1 from dual;'
```

rubytobi (changes requested):
See comment above.

apoorvdarshan:
Addressed the review feedback in `cfd78628`: newline-separated `SHOW PARAMETERS` no longer consumes a following `SELECT`, the Oracle reference link now targets the current `/26/` docs path, and a regression fixture covers the boundary. Verification: all 97 Oracle fixtures regenerated successfully and the focused dialect tests passed (3 passed).

rubytobi (changes requested):
Thanks for addressing the SELECT case. I think the fix is still too narrow though. The same swallowing happens with basically any other statement keyword after a bare SHOW PARAMETERS. Let's fix this at the root (e.g. exclude reserved/statement-starting keywords generally, or add a proper terminator) and not keep growing an exclude list.

Also, comparing against the SHOW syntax doc directly, a couple of gaps:

* SHOW ERRORS is missing ANALYTIC VIEW, ATTRIBUTE DIMENSION, and HIERARCHY as object types.

* RECYCLEBIN and SPPARAMETER[S] both take an optional name argument per the docs, same as PARAMETER[S], but that’s not connected yet. The name gets left unparsed.

rubytobi:
Closing due to lack of activity

PR #8106, opened 2026-07-09
Add regression tests for Lexer and Parser divergences

rubytobi:
I'd like to make the RustLexer and fused native-AST RustParser path the default parser given the superior performance. That switch is only safe once we have confidence that RustParser's output matches Python's across all dialects. This PR starts building the required confidence by making existing divergences visible and enforces parity going forward, so we can track progress toward the default switch.

`RustParser` should produce the same parse trees and errors as the pure-Python `Parser`, but no existing test parsed the same SQL with both and compared the results:
- `test__rust_parser__native_ast_parity` compared RustParser's two internal tree-building modes against each other, not against the Python parser.
- `test__dialect__base_file_parse` only checks round-trip raw-text fidelity, not tree structure or segment types.
- `test__dialect__base_parse_struct` compares against a checked-in YAML ground truth, but always parses with `Parser`, never `RustParser`.

Additionally, I've pointed Claude at the three parsers to find differences not already captured with the existing fixtures and create tests for it.

- Added `xfail(strict=True)` regressions to `test/core/parser/rust_parser_test.py` that run the same SQL through `Parser` and `RustParser` and compare parse tree tuples, or exception type/message for invalid SQL. Each documents one known divergence. Fixes are left to separate follow-up PRs, which also remove the corresponding `xfail`.
- Extended `test__rust_parser__native_ast_parity` to a three-way comparison (`Parser`, RustParser convert+apply, RustParser fused native-AST) across the full dialect fixture corpus, instead of just the two RustParser modes. Fixtures with an already-documented divergence are excluded via `_KNOWN_PYTHON_RUST_DIVERGENCES`; everything else must agree across all three paths.

I'd suggest keeping the `xfail` tests after they've been resolved to prevent regression, but we can also drop them/rework them into other fixtures.

CI will run a tick slower given more work is done, but negligible in my opinion.

### Pull Request checklist

  - Other.

---
## Summary by cubic
Adds strict regression tests and expands parity checks to ensure `RustParser` matches the Python `Parser` across all dialect fixtures and key error-recovery paths, making current divergences visible and preventing regressions. This builds confidence to switch the default to the `RustLexer` + fused native-AST `RustParser` path.

- New Features
  - Extended three-way parity across `Parser`, `RustParser` convert+apply, and `RustParser` native-AST for all dialect fixtures; known mismatches are excluded via `_KNOWN_PYTHON_RUST_DIVERGENCES`.
  - Added strict `xfail` regressions comparing trees/exceptions between `Parser` and `RustParser` for:
    - GREEDY/Sequence partial matches, stray/mismatched/unclosed brackets, and trailing-trivia handling in unparsable spans.
    - Real fixture divergences in `databricks`/`sparksql` PIVOT/UNPIVOT, Snowflake numeric literals, and T-SQL datatype methods and `sqlcmd` commands.
  - Added recursion-depth asymmetry test between `native_ast=True` and `False`.
  - Added lexer-layer regression for `PyLexer` vs `PyRsLexer` `SQLLexError` message formatting.
  - All regressions are `xfail(strict=True)` to document known gaps and enforce future parity; fixes will remove the corresponding `xfail`.

<sup>Written for commit 7ef3cacda9329a2f2a0f31d2aa6e56f27550b4eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8106?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test/core/parser/rust_parser_test.py
+++ b/test/core/parser/rust_parser_test.py
@@ -851,6 +851,156 @@ def test__rust_parser__vs_python_stray_bracket_swallows_union_arm():
     assert rust_result == python_result
 
 
+# ---------------------------------------------------------------------------
+# RustParser vs. pure-Python Parser divergences on well-formed, already-shipped
+# dialect fixtures.
+#
+# Unlike the malformed-SQL cases above, these reproduce on VALID SQL that's
+# already checked into the repo as a dialect fixture with a Python-generated
+# .yml ground truth - i.e. RustParser disagrees with the fixture's own
+# checked-in expected output, on input nobody had to invent. Found by running
+# Parser vs RustParser over every fixture in every dialect (not just ansi);
+# unlike ansi (0 mismatches across 208 fixtures), other dialects' fixtures
+# turned up 7 mismatches across 4 dialects.
+# ---------------------------------------------------------------------------
+
+
+def _read_fixture(dialect: str, filename: str) -> str:
+    return (_FIXTURE_DIR / dialect / filename).read_text(encoding="utf-8")
+
+
+@pytest.mark.skipif(not _HAS_RUST_PARSER, reason="Rust parser not available")
+@pytest.mark.xfail(
+    strict=True,
+    reason=(
+        "Regression: for a PIVOT clause whose aggregate expression is a "
+        "function call (e.g. SUM(sales)), RustParser inserts a spurious "
+        "extra Indent leaf inside PivotClauseSegment's Bracketed content "
+        "that Python's Parser doesn't produce, shifting every subsequent "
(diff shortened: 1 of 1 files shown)


keraion (approved):
Nice, I think I found a few of these. I'll get these tests added and double check with my PRs to knock down the list a bit.

PR #8023, opened 2026-06-28
fix(teradata): parse BTEQ FILE= command arguments

anxkhn:
### Brief summary of the change made

BTEQ meta-commands such as `.RUN FILE=POSTING` failed to parse in the Teradata dialect: the command keyword (`.RUN`) matched, but the `FILE=<path>` argument was left as an unparsable section.

This adds a command-specific `FILE=<path>` option to `BteqStatementSegment`, plus a dedicated `BteqFilePathSegment` that models unquoted file paths from existing tokens (`SingleIdentifierGrammar`, `DotSegment`, `SlashSegment`) rather than introducing a new lexer token. Both unquoted paths (`.RUN FILE=POSTING`, `.RUN FILE=reports/out.sql`) and quoted paths (`.RUN FILE='posting.sql'`) now parse.

This deliberately uses a narrow, command-specific tail rather than a broad "consume everything until newline" matcher, so invalid trailing tokens are still rejected (e.g. `.QUIT 4 unexpected` and `.RUN garbage` remain unparsable).

makes progress on #1673

### Are there any other side effects of this change that we should be aware of?

This change is narrowly scoped to the `.RUN FILE=<path>` syntax only. It does **not** add general BTEQ coverage: other BTEQ meta-commands (`.SET`, `.IMPORT`, `.EXPORT`, `.LOGON`, etc.) are still not parsed and remain out of scope here. They continue to parse via their existing grammar (where one exists) or remain unparsable (where one does not). This is intentional, so the PR does not imply broader BTEQ support than it delivers.

Beyond that:
- No changes to the lexer.
- No changes to grammar outside the Teradata dialect.
- No other dialects are affected.
- All existing Teradata fixtures continue to pass.

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - [x] `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (`test/fixtures/dialects/teradata/bteq.sql` + `bteq.yml`, YAML auto-generated with `python test/generate_parse_fixture_yml.py`).
- [x] Added appropriate documentation for the change. (No user-facing docs needed; this is an internal dialect grammar change. The new command is documented inline in the `BteqStatementSegment` docstring.)
- [x] Created GitHub issues for any relevant followup/future enhancements if appropriate. (No new issue created; broader BTEQ command support remains tracked by the existing #1673.)

---
## Summary by cubic
[...]

PR #8114, opened 2026-07-09
Add lexer support for Teradata BTEQ dot-prefixed meta-commands

### LINKED ISSUE #1673: Add support for BTEQ commands in Teradata parsing
BTEQ has a number of meta commands (docs at https://docs.teradata.com/r/jmAxXLdiDu6NiyjT6hhk7g/TvACxJd5BGW6uUDX_l2O4g). These are commands that generally start with a `.` character.

I see support for meta commands in postgres has been provided (https://github.com/sqlfluff/sqlfluff/pull/1423) - where we simply ignore the commands as long as they are confined to a single line. Doing the same thing here would be just fine.

## Expected Behaviour

We should simply ignore commands that start with a dot-command.

## Observed Behaviour

```
echo ".RUN FILE=POSTING" | sqlfluff parse --dialect teradata -
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |        bteq_statement:
[L:  1, P:  1]      |            dot:                                              '.'
[L:  1, P:  2]      |            bteq_key_word_segment:
[L:  1, P:  2]      |                keyword:                                      'RUN'
[L:  1, P:  5]      |            unparsable:                                       !! Expected: 'Nothing...'
[L:  1, P:  5]      |                whitespace:                                   ' '
[L:  1, P:  6]      |                raw:                                          'FILE'
[L:  1, P: 10]      |                raw:                                          '='
[L:  1, P: 11]      |                raw:                                          'POSTING'
[L:  1, P: 18]      |    newline:                                                  '\n'

==== parsing violations ====
L:   1 | P:   5 |  PRS | Line 1, Position 5: Found unparsable section: ' FILE=POSTING'
```

## Dialect

Teradata

## Version

sqlfluff==0.7.0

## Configuration

Default

DresdenGman:
Add a lexer matcher for Teradata BTEQ meta-commands (dot-prefixed commands like `.RUN FILE=POSTING`, `.QUIT`, `.IF`) so the parser can skip them as comments.\n\nThis follows the same pattern as the Postgres psql meta-command support added in #1423.\n\nFixes #1673

---
## Summary by cubic
Handle Teradata BTEQ dot-prefixed commands (`.RUN`, `.QUIT`, `.IF`) using the existing grammar so they’re treated as non-SQL and SQL on the same line is preserved. Remove the regex-based lexer matcher that swallowed trailing SQL; fixes #1673.

<sup>Written for commit c732f985f7aea4deb772308b1fd9fe3102741bbf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8114?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


keraion:
Hi @DresdenGman thanks for the PR. Can you please include tests covering what you've changed? Thanks!

DresdenGman:
Thanks for the prompt! I've added `bteq_lexer.sql` fixture covering BTEQ commands interleaved with SQL, including `table.column` syntax to verify no false matches. Also fixed `insert_lexer_matchers(before="code")` to `before="newline"` (caught by the fixture generator). Regenerated `bteq.yml` to match the new lexer behavior.

DresdenGman:
Added 3 edge-case fixtures covering inline BTEQ+SQL on same line, mixed-case keywords, and BTEQ without semicolons — all with auto-generated YML. The existing bteq_lexer fixture already covered basic BTEQ consumption + mixed valid SQL.

test/fixtures/dialects/teradata/bteq.yml
```diff
@@ -3,32 +3,5 @@
...
-      dot: .
-      bteq_key_word_segment:
-        keyword: RUN
-      keyword: FILE
-      comparison_operator:
-        raw_comparison_operator: '='
-      bteq_file_path:
-        naked_identifier: POSTING
+_hash: fdda373cd9cd649f82a9c5cf7ba9e290375c0ceae29477b0bad5a25f24a52ae3
+file: null
```
keraion: I'm not sure if this is the right answer here.

test/fixtures/dialects/teradata/bteq_inline.sql
```diff
@@ -0,0 +1,13 @@
+-- Edge cases: BTEQ commands mixed inline with SQL on same line.
+-- These verify the lexer correctly splits BTEQ comment from SQL.
+.IF errorcode > 0 then .quit 4; SELECT 1;
+.RUN FILE=POSTING SELECT a FROM t;
+.QUIT SELECT 1 + 2;
+.LOGOFF SELECT * FROM t;
+
+SELECT 1; .LOGOFF
```
keraion: The yaml suggests it doesn't handle the SQL found on the same line after because of the `[^\r\n]*`. It been a few years since I've used Teradata and BTEQs so I don't recall the behavior of a BTEQ like this.

keraion (changes requested):
I have some concerns with the change in approach from what was added in #8023. I think it might be good to have a second pass on what should be lexed here.

DresdenGman:
Thanks for the thorough review — you were right about the inline SQL issue. The  regex was consuming SQL on the same line after BTEQ commands.

I've reverted to the grammar-based approach — removed the RegexLexer entirely and restored the existing  grammar (which already parses BTEQ tokens precisely and stops at SQL boundaries, as introduced in #8023). The  fixture now shows the parse tree with  segments again instead of .

Let me know if there are other concerns.

keraion:
I'll go ahead and close this if there isn't anything else to do here. That last commit basically undid everything. Feel free to open a new PR if there is any additional work you feel is needed.

PR #8116, opened 2026-07-10
fix GREEDY sequence partial-match failure dropping matched children

rubytobi:
When a `Sequence` grammar in GREEDY mode fails partway through matching (after some elements already matched successfully), `RustParser` discarded all the already-matched children and returned an empty/default match result instead of keeping what had matched so far.

For SQL that partially matches a sequence grammar before failing, `RustParser` produced a different (and less informative) parse tree than the Python parser, which preserves the matched prefix. This affects error recovery and unparsable-segment reporting for malformed SQL.

```sql
SELECT CASE
```

In `sqlfluffrs_parser/src/parser/table_driven/sequence.rs`, the partial-match failure path was changed to keep the children matched so far instead of resetting to `..Default::default()`.

No

### Pull Request checklist

  - Other.

---
## Summary by cubic
Fixes GREEDY sequence partial-match handling in `RustParser` to preserve already-matched children, aligning with the Python parser and improving error recovery on malformed SQL (e.g. `SELECT CASE`).

- **Bug Fixes**
  - Keep matched children and inserts when a GREEDY sequence fails; only wrap the unmatched tail as an `UnparsableSegment`.
  - Pass `current_element_grammar_id` to `calculate_sequence_child_start_position` to mirror Python gap-skipping and keep trailing whitespace as a sibling.
  - Added a regression test to ensure Rust vs Python parity on `SELECT CASE` and removed the previous xfail.

<sup>Written for commit adda5533adc9084cf8a6851116f4f8f7de9afe5e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8116?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/table_driven/sequence.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/table_driven/sequence.rs
@@ -411,13 +411,18 @@ impl Parser<'_> {
         }
 
         // GREEDY modes with partial match - create UnparsableSegment
+        //
+        // NOTE: pass current_element_grammar_id here, the element that just
+        // failed to match, not the next element. Python's equivalent
+        // (skip_start_index_forward_to_code in grammar/sequence.py) always
+        // skips forward over any gap before the failed element, regardless of
+        // what comes after it, so mirroring that keeps trailing whitespace as
+        // a sibling gap even when the next element happens to be a Meta
+        // (e.g. Dedent).
         let child_start_pos = self.calculate_sequence_child_start_position(
             matched_idx,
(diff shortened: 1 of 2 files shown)


keraion (approved):
Nice

PR #8120, opened 2026-07-10
Fix nested bracket-type mismatch recovering instead of raising, matching Python's resolve_bracket

rubytobi:
`RustParser`'s bracket-matching (`compute_bracket_pairs`) did not follow the same last-in-first-out discipline as the Python parser when a closing bracket didn't match the innermost open bracket. Depending on the nesting, this could
cause `RustParser` to either silently recover from a bracket-type mismatch that Python treats as fatal, or leave brackets in an inconsistent state after one opener could not be resolved (a "stranded" opener). The error message raised also didn't always name the specific nested opener involved.

For SQL with nested brackets where an inner bracket is closed with the wrong type, `RustParser` could produce a different parse result or a different (or missing) error than the Python parser, instead of consistently raising with a message naming the mismatched bracket.

```sql
SELECT a[(1]
```

```sql
SELECT a[(1]) ]
```

`compute_bracket_pairs` (duplicated in `sqlfluffrs_lexer/src/lexer.rs` and `sqlfluffrs_parser/src/parser/python.rs`, since `RustParser`'s actual parse path uses the latter) now clears the entire bracket stack, not just the top entry, when a type mismatch is found, matching the Python parser's behavior where an exception at any nesting depth propagates through all enclosing brackets. `find_mismatched_closing_bracket` was also reworked to track nested unresolved openers via a local stack so the raised message names the correct expected/actual bracket types at any nesting depth.

No

### Pull Request checklist

  - Other.

---
## Summary by cubic
Fixes nested bracket-type mismatches and unclosed-bracket reporting in the Rust parser to match Python’s `resolve_bracket`, preventing recovery and aligning error type, message, and position.

- **Bug Fixes**
  - Enforce strict LIFO in `compute_bracket_pairs` (in both `sqlfluffrs_lexer/src/lexer.rs` and `sqlfluffrs_parser/src/parser/python.rs`) and clear the entire stack on a mismatch so closers only match the top-of-stack opener.
  - Rework `match_algorithms.rs` to scan from the innermost unresolved opener and distinguish mismatch vs. unclosed cases, reporting the correct expected/actual types and blaming the innermost bracket; simplified the helper by removing a redundant parameter.
  - Expand tests to compare full error text, error position/flags, and tree segment positions; add cases for crossed-bracket sequences and for innermost-unclosed error positioning.

<sup>Written for commit 69a5c7f16bc1267614e6169ebc2739694f669a6a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8120?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs_lexer/src/lexer.rs
+++ b/sqlfluffrs/sqlfluffrs_lexer/src/lexer.rs
@@ -408,19 +408,22 @@ impl Lexer {
                 // may be closed next, per LIFO nesting discipline. This matches
                 // Python's resolve_bracket (match_algorithms.py), which recurses
                 // into a freshly-opened bracket and only returns once that
-                // specific bracket is resolved. On a type mismatch here (e.g.
-                // `a[(1]`), both this closer and the innermost opener stay
-                // unresolved (matching_bracket_idx None for both); the outer `[`
-                // is left for its own closer rather than matched against `]`.
+                // specific bracket is resolved.
                 if let Some(&(open_idx, top_char)) = bracket_stack.last() {
                     if top_char == expected_open {
                         bracket_stack.pop();
                         // Set bidirectional pointers
                         tokens[open_idx].matching_bracket_idx = Some(idx);
                         tokens[idx].matching_bracket_idx = Some(open_idx);
+                    } else {
+                        // A type mismatch (e.g. `a[(1]`) is what Python's
+                        // resolve_bracket raises on, unwinding through every
+                        // enclosing bracket's own call. Clear the whole stack
+                        // to match that: every bracket still open at this
+                        // point stays unresolved, so none of them can later
+                        // pair with a closer that follows.
+                        bracket_stack.clear();
                     }
-                    // else: type mismatch - leave both this closer and the
-                    // top-of-stack opener unresolved.
                 }
                 // If the stack is empty, there's no open bracket at all -
                 // leave as None (syntax error).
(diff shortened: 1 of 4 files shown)


keraion:
I'm agreeing with cubic on this one. It's partially fixed, but it still doesn't align with Python's behavior.

For example, these raise in Python but not in Rust:
`SELECT a[(1]) ]`
`SELECT a[[1)]]`

rubytobi:
Both fixed.

First one clears the whole bracket stack on a type mismatch instead of just the mismatched pair, so a later closer can no longer cross-match one of the stranded outer openers. That covers both `SELECT a[(1]) ]` and `SELECT a[[1)]]` which now raise SQLParseError in RustParser like Python does.

The second also tightens `_compare_parser_vs_rust` to assert on the full exception message, not just the exception type, and fixes `find_mismatched_closing_bracket` in `match_algorithms.rs` to report the innermost active bracket's expected type rather than the outermost one. Meaning RustParser's message text now matches Python's exactly for these nested cases too, not just the exception class. 

Same fix now also covers the unclosed-to-EOF case, not just the mismatched-type case: an unclosed bracket nested inside another (e.g. `SELECT a(b[1`) is now blamed on the innermost open bracket, matching where Python's recursive `resolve_bracket` would actually raise from.

Added your three examples as a regression test.

keraion (approved):
LGTM

PR #8121, opened 2026-07-10
Fix unclosed GREEDY-mode bracket recovering instead of raising

rubytobi:
When a `Bracketed` grammar in GREEDY mode reaches end-of-file without finding its closing bracket, or ends up with an empty child match where one is required, `RustParser` returned a soft "with error" match result and kept going, instead of raising a hard parse error the way the Python parser does.

Unclosed brackets in SQL could be silently accepted (or produce a very different parse tree) by `RustParser` instead of surfacing a parse error, masking malformed input that the Python parser correctly flags.

```sql
WITH a AS (SELECT 1
```

In `sqlfluffrs_parser/src/parser/table_driven/bracketed.rs`, the two affected branches (`handle_bracketed_content_result`'s GREEDY-EOF case and `handle_bracketed_close_result`'s empty-child-match case) were changed to return `Err(ParseError::with_context(...))` instead of a soft `MatchResult::with_error`.

No

### Pull Request checklist

  - Other.

---
## Summary by cubic
Make `RustParser` raise a hard error for unclosed GREEDY-mode brackets to match Python. This surfaces malformed SQL instead of silently recovering a tree.

- **Bug Fixes**
  - In `sqlfluffrs_parser/src/parser/table_driven/bracketed.rs`, GREEDY EOF and empty-child-match paths now return `Err(ParseError::with_context(...))` instead of `MatchResult::with_error`.
  - Tests: unclosed GREEDY bracket now asserts both parsers raise (`WITH a AS (SELECT 1`); restored `xfail` for nested bracket-type mismatch with a more specific test comparing error text (`SELECT a[(1]`).

<sup>Written for commit 386b6489db88788812426320ebbefedf8263c1b9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8121?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test/core/parser/rust_parser_test.py
+++ b/test/core/parser/rust_parser_test.py
@@ -845,21 +845,30 @@ def test__rust_parser__vs_python_mismatched_bracket_type_error_message():
 
 
 @pytest.mark.skipif(not _HAS_RUST_PARSER, reason="Rust parser not available")
-@pytest.mark.xfail(
-    strict=True,
-    reason=(
-        "Regression: a nested nested-bracket-type mismatch (e.g. an "
-        "unclosed '(' inside '[...]' that gets 'closed' by the outer ']') "
-        "makes Python raise the same 'Found unexpected end bracket!' "
-        "SQLParseError as the flat mismatched-bracket-type case. RustParser "
-        "instead silently recovers a tree, demoting the malformed nested "
-        "bracket content to an unparsable expression rather than raising. "
-        "This is bug class 3 (Python raises, Rust recovers) triggered by a "
-        "nested mismatch rather than an unclosed-to-EOF bracket."
-    ),
-)
 def test__rust_parser__vs_python_nested_bracket_mismatch_raises():
-    """Python raises on nested bracket-type mismatch; RustParser recovers a tree."""
+    """Python and RustParser now agree on a nested bracket-type mismatch.
+
+    Regression test: a nested bracket-type mismatch (e.g. an unclosed '('
+    inside '[...]' that gets "closed" by the outer ']') makes Python raise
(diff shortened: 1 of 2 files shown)


test/core/parser/rust_parser_test.py
```diff
@@ -857,21 +845,30 @@ def build(use_rust: bool):
...
+    from Python's recursive `resolve_bracket`, which only ever resolves the
+    innermost open bracket next.
+
+    Fixed by requiring the top of the bracket stack to match the closer's
+    expected type in both `compute_bracket_pairs` implementations:
+    `sqlfluffrs_lexer/src/lexer.rs` (used when sqlfluffrs does its own
+    lexing) and the duplicate in `sqlfluffrs_parser/src/parser/python.rs`
+    (used when RustParser re-derives bracket pairs from Python-lexed
+    tokens, e.g. via `Linter(use_rust_parser=True)` - the only publicly
+    observable path, and the one the initial lexer.rs-only fix missed).
```
keraion: Not sure this docstring makes sense in conjunction with this change

rubytobi: You are right, I got the branches mixed up here. I've restored the xfail and updated the test case to be more specific about what actually failed.

keraion (approved):
LGTM

PR #8122, opened 2026-07-10
Fix isinstance-preserved token type dropped by RustParser.parse()

rubytobi:
Some grammar references preserve a specific segment type override (an "isinstance override") so that matched tokens keep a more specific type than their grammar would otherwise assign. `RustParser`'s `handle_ref_combining` logic recorded this override on `MatchedClass.segment_type`, but the actual tree-building path used by `RustParser.parse()` (`_apply_rs_match_result`/`_rs_match_fields` in `rust_parser.py`) reads the type from `segment_kwargs.instance_types` instead, so the override was never applied to real output.

Segments that should keep a specific overridden type were instead typed more generically when parsed via `RustParser.parse()`, even though the same fix tested correctly against the other (legacy) tree-building path.

```sql
CREATE CATALOG INTEGRATION glue_int
  CATALOG_SOURCE = GLUE
  REFRESH_INTERVAL_SECONDS = 10;
```

(Snowflake dialect; `10` should be typed `numeric_literal`, but `RustParser` typed it as the more generic `literal`.)

```sql
:setvar variable_name "variable_value"
```

## How it was fixed

In `sqlfluffrs_parser/src/parser/table_driven/ref_grammar.rs`, `handle_ref_combining` now also threads the override into
`segment_kwargs.instance_types` (via a new `isinstance_override` field), not just `segment_type`, so it reaches `RustParser.parse()`'s actual output.

No

### Pull Request checklist

  - Other.

---
## Summary by cubic
Fixes `RustParser` dropping isinstance-preserved token types by passing the runtime instance type list into `segment_kwargs.instance_types` used by `RustParser.parse()`. Parsed trees now keep specific types (e.g., `numeric_literal`, `word`, `double_quote`) instead of generic defaults.

- **Bug Fixes**
  - In `handle_ref_combining` (Rust), capture the isinstance override as a full `instance_types` list and set `segment_kwargs.instance_types` (not only `segment_type`).
  - Matches Python parser output for Snowflake numeric literals in `CREATE CATALOG INTEGRATION` and T‑SQL `:setvar` command content.
  - Tests: removed xfails on the two parity tests and dropped the affected fixtures from the mismatch list.

<sup>Written for commit 81cce010392a72e03d5e32effa75a5ed41e13a9e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8122?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/table_driven/ref_grammar.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/table_driven/ref_grammar.rs
@@ -261,6 +261,12 @@ impl Parser<'_> {
             // value keeps the `'static` lifetime — the common case borrows the
             // grammar-table string into the node with no allocation. Only the
             // isinstance override (a runtime token type) takes the owned branch.
+            // Records the runtime type when the isinstance-preservation path
+            // below overrides the grammar-table segment_type, so it can also
+            // be carried through `segment_kwargs.instance_types` below - that
+            // field is what RustParser.parse()'s Python-side tree builder
+            // actually reads the type override from.
+            let mut isinstance_override: Option<String> = None;
             let effective_segment_type: Cow<'static, str> = if let Some(seg_type) =
                 state.segment_type
             {
@@ -289,6 +295,7 @@ impl Parser<'_> {
                                 seg_type,
                                 tok.raw()
                             );
+                            isinstance_override = Some(effective.to_string());
                             Cow::Owned(effective.to_string())
                         } else {
                             Cow::Borrowed(seg_type)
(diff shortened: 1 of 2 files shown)


keraion (approved):
Nice work. There are a few things that `to_tuple` masks, but this covers a good portion of the issues.

PR #8123, opened 2026-07-10
Teradata: make OVERLAPS an infix comparison operator

### LINKED ISSUE #2492: Teradata dialect: L019 incorrect identified in case when statements
## Expected Behaviour

No changes to the commas as they're in the right place

## Observed Behaviour

The comma from the second `CASE WHEN` statement is transferred right after the first `CASE WHEN` statement's comma.

Rule `L019` incorrect identified the trailing comma in the second `CASE WHEN` statement as leading comma.

## Steps to Reproduce

Query in question:

```sql
SELECT
    Col1 AS Column1,
    CASE WHEN current_date = DATE '2021-01-01' THEN 1 ELSE 0 END AS Strt_Yr,
    CASE WHEN (Current_Date, Current_Date + 7) overlaps (DATE '2021-01-01', DATE '2021-01-07') THEN 1 ELSE NULL END AS col2,
    CASE WHEN (Current_Date, Current_Date + 7) overlaps (DATE '2021-01-01', DATE '2021-01-07') THEN 1 ELSE NULL END AS col3
```

Run `sql fix test.sql --force` will return

```sql
SELECT
    Col1 AS Column1,
    CASE WHEN current_date = DATE '2021-01-01' THEN 1 ELSE 0 END AS Strt_Yr, ,
    CASE WHEN (Current_Date, Current_Date + 7) overlaps (DATE '2021-01-01', DATE '2021-01-07') THEN 1 ELSE NULL END AS col2
    CASE WHEN (Current_Date, Current_Date + 7) overlaps (DATE '2021-01-01', DATE '2021-01-07') THEN 1 ELSE NULL END AS col3

```

## Dialect
`teradata`
## Version

```shell 
$ sqlfluff --version
sqlfluff, version 0.9.3

$ python --version
Python 3.9.9
```
Same error on python 3.6.8

## Configuration

```ini
[sqlfluff]
verbose = 2 
dialect = teradata
```

apoorva-01:

`(a, b) OVERLAPS (c, d)` didn't parse in teradata. The ANSI `overlaps_clause` only matches the right-hand side, so `OVERLAPS` never binds two operands and falls over anywhere a boolean is expected, like the `CASE WHEN` in the issue.

Added `OVERLAPS` to teradata's comparison operators so it works as the infix predicate it is. `NORMALIZE ON ... OVERLAPS` still parses.

Same bug exists in ansi and postgres, so this arguably belongs in ANSI. Kept it to teradata since that's what the issue is about, happy to move it up if you'd rather.

Fixes #2492.

No, teradata only. No other dialect ymls moved.

### Pull Request checklist

- Included test cases:
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects/teradata`.

---
## Summary by cubic
Treats `OVERLAPS` as an infix comparison operator in Teradata so `(a, b) OVERLAPS (c, d)` parses and works in WHERE and CASE expressions. Keeps `NORMALIZE ON ... OVERLAPS` behavior unchanged.

- **Bug Fixes**
  - Added `OverlapsOperatorSegment` to Teradata comparison operators to bind both operands; implemented with `StringParser` for consistency.
  - Added parser tests for CASE, WHERE, and period comparisons. Fixes #2492.

<sup>Written for commit 36f2bfc2300318d70fa93ba9ea23989b03b034c0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8123?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_teradata.py
+++ b/src/sqlfluff/dialects/dialect_teradata.py
@@ -169,6 +169,10 @@ teradata_dialect.add(
     NotEqualToSegment_a=StringParser("NE", ComparisonOperatorSegment),
     NotEqualToSegment_b=StringParser("NOT=", ComparisonOperatorSegment),
     NotEqualToSegment_c=StringParser("^=", ComparisonOperatorSegment),
+    # Unlike the ANSI `overlaps_clause`, this binds two operands, so it also
+    # works where a boolean is expected (e.g. inside CASE WHEN).
+    # https://docs.teradata.com/r/kmuOwjp1zEYg98JsB8fu_A/3VIgdwHNVU~tsnNiIR1aEw
+    OverlapsOperatorSegment=StringParser("OVERLAPS", ComparisonOperatorSegment),
 )
 
 
@@ -1068,15 +1072,3 @@ class NotEqualToSegment_c(CompositeComparisonOperatorSegment):
     match_grammar = Sequence(
         Ref("BitwiseXorSegment"), Ref("RawEqualsSegment"), allow_gaps=False
     )
-
-
-class OverlapsOperatorSegment(CompositeComparisonOperatorSegment):
-    """The OVERLAPS period comparison operator.
-
-    Unlike the ANSI `overlaps_clause`, this binds two operands, so it also works
-    where a boolean is expected (e.g. inside CASE WHEN).
-
-    https://docs.teradata.com/r/kmuOwjp1zEYg98JsB8fu_A/3VIgdwHNVU~tsnNiIR1aEw
-    """
-
-    match_grammar = Sequence("OVERLAPS")
--- a/src/sqlfluff/dialects/dialect_teradata.py
+++ b/src/sqlfluff/dialects/dialect_teradata.py
@@ -149,6 +149,7 @@ teradata_dialect.replace(
         Ref("NotEqualToSegment_b"),
         Ref("NotEqualToSegment_c"),
         Ref("LikeOperatorSegment"),
+        Ref("OverlapsOperatorSegment"),
         Sequence("IS", "DISTINCT", "FROM"),
         Sequence("IS", "NOT", "DISTINCT", "FROM"),
     ),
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_teradata.py
```diff
@@ -1067,3 +1068,15 @@ class NotEqualToSegment_c(CompositeComparisonOperatorSegment):
...
+class OverlapsOperatorSegment(CompositeComparisonOperatorSegment):
+    """The OVERLAPS period comparison operator.
+
+    Unlike the ANSI `overlaps_clause`, this binds two operands, so it also works
+    where a boolean is expected (e.g. inside CASE WHEN).
+
+    https://docs.teradata.com/r/kmuOwjp1zEYg98JsB8fu_A/3VIgdwHNVU~tsnNiIR1aEw
+    """
+
+    match_grammar = Sequence("OVERLAPS")
```
peterbud: `OverlapsOperatorSegment` is just a `Sequence("OVERLAPS")`. For a single-keyword comparison operator, using `StringParser("OVERLAPS", ComparisonOperatorSegment)` would be more consistent with the existing parser patterns in this area and avoids introducing a dedicated segment class that does not add much structure.

If there is a parsing reason to keep it composite, it would be worth calling that out in the PR description or with an inline comment.

peterbud (changes requested):
Thanks @apoorva-01  - can you look at the below suggestion? 

With regards to ANSI and/or Postgres changes: let's keep this PR focused on Teradata, but after that if you want to prep a PR for that it's more than welcomed.

apoorva-01:
Switched it to `StringParser("OVERLAPS", ComparisonOperatorSegment)` to match the other extensions. No composite structure needed for a single keyword.
And sounds good on keeping this Teradata-only. I'll open a separate PR for the ANSI/Postgres side once this lands.

peterbud (approved):
Thanks @apoorva-01  - looks 👌

PR #8124, opened 2026-07-10
Teradata: parse data type attributes ((NOT) CASESPECIFIC/CS, UPPERCASE) on expressions

### LINKED ISSUE #5530: Teradata: Missing parsing [NOT] CASESPECIFIC for strings
### Search before asking

### Description

In Teradata, strings can be marked as `CASESPECIFIC` or `NOT CASESPECIFIC` by suffixing it with `(CASESPECIFIC)` or `(NOT CASESPECIFIC)`. There is also a short form of it `(CS)` or `(NOT CS)`.
Example:
```sql
SELECT
    CASE
        WHEN 'TEST' (CASESPECIFIC) = 'test' (CASESPECIFIC)
            THEN 'Not true'
        WHEN 'TEST' (NOT CASESPECIFIC) = 'test' (NOT CASESPECIFIC)
            THEN 'True'
        WHEN 'TEST' (CS) = 'test' (CS)
            THEN 'Not true'
        WHEN 'TEST' (NOT CS) = 'test' (NOT CS)
            THEN 'True'
    END
```
This syntax is not recognized yet. 

There are also side effects when the line is marked with `--noqa`. The line after got wrong parsed. Other errors are reported. For example, a missing entity in the FROM section. Ever the entity is definitely there. 

### Use case

There should be no error reported for this query:
```sql
SELECT
    CASE
        WHEN SOME_TABLE.Attribute1 = 'test' (CASESPECIFIC)
            THEN 'Not true'
        WHEN SOME_TABLE.Attribute1 = 'Test' (NOT CASESPECIFIC)
            THEN 'True'
    END
FROM SOME_DATASE.SOME_TABLE AS SOME_TABLE
```

### Dialect

Teradata

### Code of Conduct

adhavan18:

`fixes #5530`

Teradata allows a data type attribute to be applied to an expression as a parenthesised postfix, e.g.:

```sql
SELECT TEST (CASESPECIFIC);
SELECT TEST (NOT CASESPECIFIC);
SELECT TEST (CS);
SELECT TEST (NOT CS);
SELECT col (UPPERCASE);
```

Previously only the datatype-conversion form (e.g. `x (DATE)`, handled by `TeradataCastSegment`) was supported, so the bare-attribute forms did not parse correctly:

- `(NOT CASESPECIFIC)` / `(NOT CS)` raised `Found unparsable section`.
- The plain forms `(CASESPECIFIC)` / `(CS)` happened to parse, but were mis-interpreted as a cast to a datatype literally named `CASESPECIFIC`/`CS`.

The `TeradataCastSegment` postfix only applies at the top of `ExpressionSegment`, so an attribute on the **left** operand of a comparison (`TEST (CASESPECIFIC) = test (CASESPECIFIC)` — the exact example from the issue) also failed.

This change sets the Teradata `CollateGrammar` to a bracketed `CharCharacterSetGrammar`. `CollateGrammar` is applied as a postfix inside `Expression_A_Grammar`, so the attribute binds to the operand just like `COLLATE` in standard SQL and works on either side of a comparison. The existing `CASESPECIFIC` / `CS` / `UPPERCASE` / `UC` keywords and grammar were already defined (for CAST and column constraints); this reuses them.

No. The change is scoped to the Teradata dialect (`CollateGrammar` is `Nothing()` in ANSI and unchanged elsewhere). Existing Teradata cast fixtures (`x (DATE)`, `x (SMALLINT)`) and normal function calls (`UPPER(col)`, `COUNT(*)`, `CAST(x AS INTEGER)`) are unaffected. All existing Teradata dialect tests still pass, plus a new parser fixture.

### Pull Request checklist

- [x] Included test cases to demonstrate any code changes:
  - `.sql`/`.yml` parser test case added at `test/fixtures/dialects/teradata/data_type_attribute.sql` (+ generated `.yml`), covering the forms from the issue including both sides of a comparison and inside `CASE`.
- [x] Added appropriate documentation for the change (docstring / inline comments on the grammar).
- [ ] Created GitHub issues for any relevant followup/future enhancements if appropriate — n/a.

---
## Summary by cubic
Adds support for Teradata data-type attributes as parenthesized postfixes on expressions and tightens parsing to prevent mis-parses and duplicates. Fixes CASESPECIFIC/CS and UPPERCASE/UC on both sides of comparisons and in CAST; fixes #5530.

- **Bug Fixes**
  - Bind postfix attributes to the operand via Teradata `CollateGrammar = Bracketed(Ref("CharCharacterSetGrammar"))`, so they work on either side of a comparison.
  - Reserve `CASESPECIFIC`/`CS`/`UPPERCASE`/`UC` so `col (UPPERCASE)` is not parsed as a function call.
  - Model data-type attributes in `DatatypeSegment` with `AnySetOf(...)` to allow at most one of each (CHARACTER SET, [NOT] CASESPECIFIC/CS, UPPERCASE/UC), correctly parsing `CAST(... CHARACTER SET LATIN NOT CASESPECIFIC)` and rejecting duplicates; remove these attributes from `TdColumnConstraintSegment` (it now only handles `COMPRESS`).

<sup>Written for commit 031072e626214fb282339f66c8983a4d1284c704. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8124?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_teradata.py
+++ b/src/sqlfluff/dialects/dialect_teradata.py
@@ -484,9 +484,18 @@ class DatatypeSegment(ansi.DatatypeSegment):
         Sequence(  # FORMAT 'YYYY-MM-DD',
             "FORMAT", Ref("QuotedLiteralSegment"), optional=True
         ),
-        # Data attributes may be chained, e.g.
-        # VARCHAR(50) CHARACTER SET LATIN NOT CASESPECIFIC
-        AnyNumberOf(Ref("CharCharacterSetGrammar")),
+        # Data attributes may be combined, in any order, e.g.
+        # VARCHAR(50) CHARACTER SET LATIN NOT CASESPECIFIC.
+        # AnySetOf allows each attribute at most once, so duplicates such as
+        # UPPERCASE UPPERCASE or repeated CHARACTER SET clauses are rejected.
+        AnySetOf(
+            Sequence("CHARACTER", "SET", Ref("SingleIdentifierGrammar")),
+            Sequence(
+                Ref.keyword("NOT", optional=True),
+                OneOf("CASESPECIFIC", "CS"),
+            ),
+            OneOf("UPPERCASE", "UC"),
+        ),
     )
 
 
@@ -554,21 +563,22 @@ class ColumnDefinitionSegment(BaseSegment):
 class TdColumnConstraintSegment(BaseSegment):
     """Teradata specific column attributes.
 
-    e.g. CHARACTER SET LATIN | [NOT] (CASESPECIFIC|CS) | (UPPERCASE|UC)
+    e.g. COMPRESS [(1.,3.) | 3. | NULL]
+
+    The character data-type attributes (CHARACTER SET, [NOT] CASESPECIFIC/CS,
+    UPPERCASE/UC) are handled by ``DatatypeSegment`` so that each attribute can
+    appear at most once; modelling them here as well would let them repeat.
     """
 
     type = "td_column_attribute_constraint"
     match_grammar = Sequence(
-        OneOf(
-            Ref("CharCharacterSetGrammar"),
-            Sequence(  # COMPRESS [(1.,3.) | 3. | NULL],
-                "COMPRESS",
-                OneOf(
-                    Bracketed(Delimited(Ref("LiteralGrammar"))),
-                    Ref("LiteralGrammar"),
-                    "NULL",
-                    optional=True,
-                ),
+        Sequence(  # COMPRESS [(1.,3.) | 3. | NULL],
+            "COMPRESS",
+            OneOf(
+                Bracketed(Delimited(Ref("LiteralGrammar"))),
+                Ref("LiteralGrammar"),
+                "NULL",
+                optional=True,
             ),
         ),
     )
--- a/src/sqlfluff/dialects/dialect_teradata.py
+++ b/src/sqlfluff/dialects/dialect_teradata.py
@@ -123,6 +123,11 @@ teradata_dialect.replace(
         ),
         OneOf("UPPERCASE", "UC"),
     ),
+    # A Teradata data-type attribute can be applied to an expression as a
+    # parenthesised postfix, e.g. 'TEST' (CASESPECIFIC), col (NOT CS),
+    # 'x' (UPPERCASE). This behaves like COLLATE in standard SQL, binding to
+    # the operand, so it works on either side of a comparison.
+    CollateGrammar=Bracketed(Ref("CharCharacterSetGrammar")),
     FunctionContentsGrammar=ansi_dialect.get_grammar("FunctionContentsGrammar").copy(
         insert=[
             Sequence(
--- a/src/sqlfluff/dialects/dialect_teradata.py
+++ b/src/sqlfluff/dialects/dialect_teradata.py
@@ -66,8 +66,6 @@ teradata_dialect.sets("unreserved_keywords").update(
     [
         "AUTOINCREMENT",
         "ACTIVITYCOUNT",
-        "CASESPECIFIC",
-        "CS",
         "DAYS",
         "DEL",
         "DUAL",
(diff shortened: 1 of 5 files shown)


peterbud (changes requested):
Hi @adhavan18  - thanks for opening this PR, and welcome!

I think this PR still misses the bare-identifier case despite the PR description calls out. The new Teradata postfix is only reached after expression parsing, but ANSI still prefers word (...) as a function call first, so `TEST (CASESPECIFIC)` and `col (UPPERCASE)` are still parsed as functions rather than expressions with a data-type attribute. The new YML fixture already shows that which is generated as a function call.

adhavan18:
Thanks for the review @peterbud — you were right, and the fix is pushed.

**Root cause of the gap:** `CASESPECIFIC`/`CS`/`UC`/`UPPERCASE` were unreserved, so in `col (UPPERCASE)` the contents parsed as an identifier argument and the ANSI function-call match won before the postfix was considered. These are reserved words in Teradata, so the new commit moves them to `reserved_keywords` — the function-call interpretation then fails to match and the attribute postfix binds correctly. `col (UPPERCASE)`, `col (NOT CS)` etc. now parse as `column_reference` + attribute keywords (no `function` node), and regular function calls (`UPPER(col)`, `COUNT(*)`) are unaffected.

**This also surfaced a latent mis-parse** in the existing `select.sql` fixture: in `CAST(NULL AS VARCHAR(50) CHARACTER SET LATIN NOT CASESPECIFIC)` the trailing `NOT CASESPECIFIC` was previously being consumed as a dangling boolean expression (`NOT` applied to a *column named* `CASESPECIFIC`) inside the function contents — visible in the old `select.yml`. Datatypes now accept `AnyNumberOf` chained data attributes, so both `CHARACTER SET LATIN` and `NOT CASESPECIFIC` attach to the `data_type` node.

Fixture YMLs regenerated (`data_type_attribute`, `select`, `create_table` — the latter two now encode the corrected trees). Locally: all 48 Teradata dialect tests, the ANSI dialect suite and `test/core` pass (685 passed).

peterbud:
Thanks @adhavan18 for the update. It's a good direction, however this grammar now also accepts duplicates such as `VARCHAR(10) UPPERCASE UPPERCASE` and repeated `CHARACTER SET ...` chains. I think we need stricter syntax modeling with `AnySetOf(...)` or a more explicit sequence.

adhavan18:
Thanks @peterbud — fixed and pushed.

The data-type attributes are now modelled with `AnySetOf`, so each of the three categories (`CHARACTER SET ...`, `[NOT] CASESPECIFIC/CS`, `UPPERCASE/UC`) can appear at most once, in any order. Duplicates and repeated chains are rejected:

```
VARCHAR(50) UPPERCASE UPPERCASE                       -- rejected
VARCHAR(50) CHARACTER SET LATIN CHARACTER SET UNICODE -- rejected
```

I also removed the now-redundant `CharCharacterSetGrammar` from `TdColumnConstraintSegment` (it only handles `COMPRESS` now). That segment and `DatatypeSegment` were both matching the attributes, so in a `CREATE TABLE` column the duplicate slipped through the constraint path even after the `AnySetOf` change — removing the overlap fixes that too. Valid combinations are unaffected:

```
VARCHAR(10) CHARACTER SET LATIN NOT CASESPECIFIC NOT NULL   -- ok
VARCHAR(255) CHARACTER SET LATIN NOT CASESPECIFIC COMPRESS (...) -- ok
```

Existing fixture trees are unchanged (`DatatypeSegment` already won for valid input); only duplicate rejection is tightened. All 48 Teradata dialect tests pass.

peterbud (approved):
Thanks @adhavan18 !

PR #8133, opened 2026-07-12
fix: RF02 should not flag BigQuery whole-row table references

### LINKED ISSUE #4079: Un
### Search before asking

### Links or command line

_No response_

### Issue/Suggested Improvement

**correct results:** 

```
 with main as (
  select 1 as id , 100 as qty , null as price
  union all
  select 2 as id , 0 as qty , 100 as price
  union all
  select 3 as id , 0 as qty ,80 as price
  union all 
  select 4 as id , 50 as qty , 90 as price
  union all
  select 5 as id , 20 as qty , 100 as price
  union all
  select 6 as id , 20 as qty , 100 as price
),
main2 as (
  select 1 as id_, 100 as qty_ , 50 as price_
  union all
  select 2 as id_, 80 as qty_ , 100 as price_
  union all
  select 3 as id_, 100 as qty_ , 80 as price_
  union all
  select 5 as id_, 20 as qty_ , 100 as price_
  union all
  select 6 as id_, 20 as qty_ , 40 as price_
  union all 
  select 7 as id_, 20 as qty_ , 50 as price_
)

select *
from main s1
full outer join main2 s2
on id = id_
where format('%t', s1) != format('%t', s2)
```

The above query returns the correct results but I am getting the error with the sql lint

- Unqualified reference 's1' found in select with more than one referenced table/view.
- Unqualified reference 's2' found in select with more than one referenced table/view.

I am using google bigquery and if i try to create separate columns with  format('%t', t1) and format('%t', t2) in other CTE and use them again then the test passes but the results are not the same as of the above query (see the code below)

**wrong results:** 

it doesn't return the cases where the id's are only found in 1 table

```
with main as (
  select 1 as id , 100 as qty , null as price
  union all
  select 2 as id , 0 as qty , 100 as price
  union all
  select 3 as id , 0 as qty ,80 as price
  union all 
  select 4 as id , 50 as qty , 90 as price
  union all
  select 5 as id , 20 as qty , 100 as price
  union all
  select 6 as id , 20 as qty , 100 as price
),
main2 as (
  select 1 as id_, 100 as qty_ , 50 as price_
  union all
  select 2 as id_, 80 as qty_ , 100 as price_
  union all
  select 3 as id_, 100 as qty_ , 80 as price_
  union all
  select 5 as id_, 20 as qty_ , 100 as price_
  union all
  select 6 as id_, 20 as qty_ , 40 as price_
   union all 
  select 7 as id_, 20 as qty_ , 50 as price_
),
main3 as (
select *, trim(format('%t', main)) as l1 from main
),
main4 as (
select *, trim(format('%t', main2)) as l2 from main2
)

select *
from main3 
full outer join main4 
on main3.id = main4.id_
where main3.l1 !=  main4.l2 


```

can someone please help me with this so that i can pass both the sql fluff / lint tests and get the correct output ?

### Code of Conduct

mvanhorn:

RF02 flagged BigQuery whole-row references such as `format('%t', s1)` as unqualified column references, but `s1` there is a table/row reference (a STRUCT-valued whole-row value), not a column. This teaches RF02 to recognize whole-row aliases: it normalizes aliases before matching row references and restricts whole-row alias matching to the dialects that support it, so the false positive on valid BigQuery SQL is gone.

fixes #4079

The whole-row alias handling is gated to the dialects that support it, so other dialects still flag a genuinely unqualified reference exactly as before. No change to autofix behavior.

### Pull Request checklist

  - `.yml` rule test cases in `test/fixtures/rules/std_rule_cases`: added BigQuery whole-row alias cases and a guard case in `test/fixtures/rules/std_rule_cases/RF02.yml`.
- Added appropriate documentation for the change: not applicable (rule-behavior fix; no user-facing docs change).
- Created GitHub issues for any relevant followup/future enhancements if appropriate: not applicable.

AI was used for assistance.

---
## Summary by cubic
Fixes `RF02` false positives on BigQuery whole-row table references (e.g., FORMAT('%t', s1)) by recognizing whole-row aliases and keeping other dialects unchanged.

- **Bug Fixes**
  - Normalize aliases and treat them as valid whole-row references where supported.
  - Limit whole-row handling to `bigquery`, `postgres`, and `redshift`.
  - Add tests: BigQuery pass cases and a MySQL guard case.

<sup>Written for commit ea69760e684908e9e35c3a989d650626dab19a7c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8133?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/rules/references/RF02.py
+++ b/src/sqlfluff/rules/references/RF02.py
@@ -46,6 +46,11 @@ class Rule_RF02(Rule_AL04):
     config_keywords = [
         "subqueries_ignore_external_references",
     ]
+    _dialects_with_row_references = [
+        "bigquery",
+        "postgres",
+        "redshift",
+    ]
 
     # Config type hints
     ignore_words_regex: str
@@ -95,6 +100,11 @@ class Rule_RF02(Rule_AL04):
             ignore_words_list = self._init_ignore_words_list()
 
         sql_variables = self._find_sql_variables(rule_context)
+        row_reference_aliases = (
+            {alias.segment.raw_normalized() for alias in table_aliases if alias.segment}
+            if rule_context.dialect.name in self._dialects_with_row_references
+            else set()
+        )
 
         # A buffer to keep any violations.
         violation_buff = []
@@ -131,6 +141,9 @@ class Rule_RF02(Rule_AL04):
                 # Allow columns defined as standalone aliases
                 # (e.g. value table functions from bigquery)
                 and r.raw not in [a.raw for a in standalone_aliases]
+                # Allow whole-row references to table aliases in dialects that
+                # support them.
+                and r.raw_normalized() not in row_reference_aliases
             ):
                 violation_buff.append(
                     LintResult(
--- a/test/fixtures/rules/std_rule_cases/RF02.yml
+++ b/test/fixtures/rules/std_rule_cases/RF02.yml
@@ -223,6 +223,28 @@ test_fail_unqualified_references_multi_table_statements_mysql:
     core:
       dialect: mysql
 
+test_pass_bigquery_normalized_whole_row_alias_references:
+  pass_str: |
+    SELECT FORMAT('%t', s1), s2.id
+    FROM main AS S1
+    JOIN secondary AS s2 ON S1.id = s2.id;
+
+    SELECT FORMAT('%t', `s1`), s2.id
+    FROM main AS `s1`
+    JOIN secondary AS s2 ON `s1`.id = s2.id
+  configs:
+    core:
+      dialect: bigquery
+
+test_fail_mysql_alias_named_unqualified_column:
+  fail_str: |
+    SELECT o, s.id
+    FROM orders AS o
+    JOIN shipments AS s ON o.id = s.order_id
+  configs:
+    core:
+      dialect: mysql
+
 test_fail_column_and_alias_same_name_mysql:
   # See issue #2169
   fail_str: |


peterbud (approved):
Hi @mvanhorn - thanks for this PR and welcome!

I think this fixes the immediate RF02 false positive, but it does so at the rule layer by special-casing alias text rather than by improving the underlying reference semantics.

Today, `s1` inside `FORMAT('%t', s1)` is still parsed/analyzed as a generic column reference, and RF02 is simply taught to ignore that case for a small set of dialects. That is probably acceptable as a targeted fix, especially since similar rule-level handling already exists elsewhere, but it means the underlying semantic gap remains.

As a follow-up, I think it would be better to move this toward a proper reference analysis rather than adding more rule-local exceptions. For example, we could teach the select/reference analysis layer to classify whole-row alias row-references as distinct from ordinary column references, and then have RF02, RF03, AL05, etc. consume that shared classification. That would give us a more principled fix and avoid duplicating dialect-specific whole-row logic across rules.

So I’m fine with this as a practical fix, but I’d suggest opening a follow-up issue for the semantic/generalized approach.

PR #8141, opened 2026-07-13
perf: lazy SegmentCloneMap in ST05 — clone only when a subquery is rewritten

keraion:

ST05 built `SegmentCloneMap` (a full `segment.copy()` of the SELECT statement tree) for every `select_statement`, before `_lint_query` determines whether there is any subquery to extract. The clone is only *used* when there are results, so this builds it lazily on first `__getitem__`. On the no-subquery path (the common case) nothing is cloned.

This is a straightforward perf win natively (segment `copy()` is a `__dict__` update per segment, per subtree), and matters more for alternative segment implementations where `copy()` is more expensive. Measured on a 20x-duplicated file with an all-74-rule crawl: non-reflow rule cost 26.9→14.5 ms.

None — ST05 output is unchanged (`pytest test/rules -k ST05` = 47 passed).

### Pull Request checklist

  - Covered by the existing ST05 `.yml` rule test cases (no behavior change).

---
## Summary by cubic
Make `SegmentCloneMap` in ST05 lazy so the SELECT tree is only cloned when a subquery is actually rewritten. This removes unnecessary cloning on the common no-subquery path and speeds up ST05 without changing output.

- **Refactors**
  - Defer the full clone to first `__getitem__`; store the original segment until then.
  - Preserve `pos_marker` mapping; ST05 behavior unchanged (tests pass).
  - Perf: non-reflow subset ~26.9→14.5 ms; all-rules crawl ~130→103 ms on the façade (native ~109→92 ms).

<sup>Written for commit 6c6f1eb6a90ebc973caae53364e015e21b9c1acd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8141?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/rules/structure/ST05.py
+++ b/src/sqlfluff/rules/structure/ST05.py
@@ -619,17 +619,34 @@ def _segmentify(input_el: str, casing: str) -> BaseSegment:
 
 
 class SegmentCloneMap:
-    """Clones a segment tree, maps from original segments to their clones."""
+    """Clones a segment tree, maps from original segments to their clones.
+
+    The clone (a full ``segment.copy()``) is built lazily on first lookup: the
+    caller builds the map for every SELECT statement but only *uses* it when
+    there are subqueries to rewrite, so deferring avoids cloning the whole
+    statement tree on the common no-subquery path (a notable cost over the Rust
+    arena façade, where copy() materialises real segments).
+    """
 
     def __init__(self, segment: BaseSegment):
+        self._segment = segment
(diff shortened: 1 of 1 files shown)


peterbud (approved):
Nice!

PR #8147, opened 2026-07-14
fix(rust): match Python's repr()-quoted, truncated unlexable-error text in the Rust lexer

rubytobi:
When SQL contains characters that can't be lexed, the Python parser raises a `SQLLexError` whose message includes a Python `repr()`-style rendering of the offending text, truncated for long strings. The Rust parser's equivalent error message was built differently, so the text didn't match between the two backends.

```sql
SELECT \xa1 FROM t
```

(`\xa1` here is a single non-ASCII character, U+00A1, that neither lexer can tokenize, forcing the unlexable fallback path on both sides.)

Added `python_repr_str` and `truncate_like_python` helpers in `sqlfluffrs_lexer/src/lexer.rs` that reproduce Python's `repr()`-style quoting/escaping and its 9-character truncation-plus-`...` cutoff, and used them when building the `SQLLexError` message so it matches the Python parser's wording for the common case: control characters, punctuation, and symbols.

Also un-xfails `test__rust_parser__vs_python_lexer_unlexable_error_message` in `test/core/parser/rust_parser_test.py`, which was previously marked `xfail(strict=True)` for this exact mismatch and now asserts the two backends produce identical error text.

The escaping logic approximates Python's `str.isprintable()` check with `char::is_control()` plus non-space Unicode whitespace. This covers control characters, punctuation, and symbols, i.e. the characters that typically make input unlexable, but it does not yet cover the rarer Unicode categories Cf (format, e.g. zero-width space), Co (private use), and Cn (unassigned). For those, Python's `repr()` would escape the codepoint but the Rust output currently prints it raw. Rust's standard library exposes no Unicode General_Category lookup, so escaping the remaining Cf/Co/Cn codepoints would require adding a new dependency (e.g. unicode-general-category) or manual mapping.

### Pull Request checklist

  - Other.

---
## Summary by cubic
Aligns the Rust lexer’s unlexable error text with Python by using repr()-style quoting and the same truncation. This makes messages identical across backends and prevents raw control characters from appearing in output.

- Bug Fixes
  - Build errors with repr()-style quoting and a 9-char + "..." truncation to match Python.
  - Added `python_repr_str` and `truncate_like_python` in `sqlfluffrs_lexer/src/lexer.rs`.
  - Updated `test__rust_parser__vs_python_lexer_unlexable_error_message` to assert identical messages (removed xfail).

<sup>Written for commit 443d3dad608c3bf8c5632ad9ac62878b4328044a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8147?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs_lexer/src/lexer.rs
+++ b/sqlfluffrs/sqlfluffrs_lexer/src/lexer.rs
@@ -425,7 +425,7 @@ impl Lexer {
                 SQLLexError::new(
                     Some(format!(
                         "Unable to lex characters: {}",
-                        token.raw().chars().take(10).collect::<String>()
+                        python_repr_str(&truncate_like_python(token.raw()))
                     )),
                     token.pos_marker.clone(),
                     None,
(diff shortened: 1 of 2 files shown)


keraion (approved):
LGTM

PR #8148, opened 2026-07-14
fix(parser): stop genexpr/comprehension recursion from halving safe parse-tree depth

rubytobi:
<!--Thanks for adding this feature!-->

<!--Please give the Pull Request a meaningful title for the release notes-->

Several recursive tree-walking methods, i.e. `RustParser`'s legacy tree builder (`_convert_rs_match_result`) and various `BaseSegment` methods (`descendant_type_set`, `count_segments`, `to_tuple`, `copy`, `is_code`, `is_comment`, `is_whitespace`, `raw`, `raw_segments_with_ancestors`, `source_fixes`, `get_raw_segments`), recursed via a generator expression/comprehension. In CPython that costs an extra stack frame per level, roughly halving the safe nesting depth before hitting Python's recursion limit.                                                                                                                                                                                                                                                                                                                                                                                                                                                                          

`Parser` and `RustParser` share the same `max_parse_depth` guard during matching, which rejects deep nesting under the default limit (600) long before this bug is reached. It only bites once `max_parse_depth` is deliberately raised above default, where the affected methods hit `RecursionError` at about half the depth that raising the limit was meant to allow.

The fix rewrote each method as a plain `for` loop instead of a genexpr/comprehension. `get_raw_segments` needed one extra change: it recursed through the separately-cached `raw_segments` property, which added its own indirection frame regardless of the loop conversion, i.e. the recursion now lives in `raw_segments` directly, with `get_raw_segments` as a thin wrapper.                                                                                                                                                                                                                                                                                                                                                                                                                                                                        

While testing, I found a separate, pre-existing problem where extreme nesting can hard-crash the Rust matching engine's native                                                                                                                                                                       stack which is uncatchable in Python, unrelated to this fix, and not reachable under default config:
```python
def deeply_nested_sql() -> str:
   return "SELECT " + "(" * 130 + "1" + ")" * 130
```


### Pull Request checklist

  - Other.

<!-- This is an auto-generated description by cubic. -->
---
## Summary by cubic
Stops generator-expression recursion from halving safe parse-tree depth. Deeply nested parses now behave as expected when `max_parse_depth` is raised, and both AST builders fail at the same depth.

- Bug Fixes
  - Replaced generator-expression recursion with plain loops in `RustParser._convert_rs_match_result` to match `_apply_rs_match_result` stack usage.
  - Converted recursive `BaseSegment` methods/properties to plain loops: `descendant_type_set`, `count_segments`, `to_tuple`, `copy`, `is_code`, `is_comment`, `is_whitespace`, `raw`, `raw_segments_with_ancestors`, `source_fixes`.
  - Moved recursion into the `raw_segments` cached property; `get_raw_segments()` now delegates to it to avoid extra stack frames.
  - Added regression tests for deep nesting across these accessors, native vs. legacy AST builder parity, and `.is_code` on a comment-only file.

<sup>Written for commit 7feb33f717540089a5dcc7fc5a286ed33de4f976. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8148?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>

<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/src/sqlfluff/core/parser/rust_parser.py
+++ b/src/sqlfluff/core/parser/rust_parser.py
@@ -523,14 +523,18 @@ try:
             # Convert child matches recursively
             # Note: Transparent grammar nodes are now flattened on the Rust side,
             # so we don't need to do it here anymore
-            child_matches = (
-                tuple(
+            #
+            # NOTE: Keep this a plain loop rather than a generator expression:
+            # a genexpr would add its own stack frame per nesting level on top
+            # of this recursive call, which halves the safe recursion depth.
+            # Matches _apply_rs_match_result's plain for loop below, so both
+            # AST-building paths tolerate the same nesting depth.
+            child_matches_list = []
+            for child in rs_match.child_matches:
+                child_matches_list.append(
                     self._convert_rs_match_result(child, segments, depth + 1)
-                    for child in rs_match.child_matches
                 )
-                if rs_match.child_matches
-                else ()
-            )
+            child_matches = tuple(child_matches_list)
 
             return MatchResult(
                 matched_slice=slice(start, stop),
--- a/src/sqlfluff/core/parser/segments/base.py
+++ b/src/sqlfluff/core/parser/segments/base.py
@@ -345,26 +345,34 @@ class BaseSegment(metaclass=SegmentMetaclass):
         return tuple(idx for idx, seg in enumerate(self.segments) if seg.is_code)
 
     @cached_property
-    def is_comment(self) -> bool:  # pragma: no cover TODO?
+    def is_comment(self) -> bool:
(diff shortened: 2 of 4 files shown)


peterbud (approved):
Nice!

PR #8149, opened 2026-07-14
Rename Rust skip-backward-to-code helpers to match Python

### LINKED ISSUE #8129: Rename Rust `skip_stop_index_backward_to_code` helpers
This name collides with Python but has inverted semantics.

## Problem

The Rust parser has three backward-skip-to-code helpers whose names and semantics are misaligned with each other and with the Python reference. The helper whose name is identical to Python's canonical function has the opposite comment-handling behavior, which is a trap for anyone porting from Python.

Python's canonical `skip_stop_index_backward_to_code` (src/sqlfluff/core/parser/match_algorithms.py:32) checks only is_code, i.e. it excludes comments. So:

- The Rust method named `skip_stop_index_backward_to_code` does not match Python's `skip_stop_index_backward_to_code` (it keeps comments).
- The helper that does match Python is the one with the longer _excluding_comments suffix.
- Two definitions share the exact name but differ in return convention (idx vs idx + 1).

Proposed change (naming only — no behavior change)

Make the name that matches Python carry the Python semantics:

- Rename the comment-keeping method helpers.rs:142 → `skip_stop_index_backward_to_code_or_comment` (name states what it actually keeps).
- Rename the comment-excluding helper (_excluding_comments, added in #8118) → `skip_stop_index_backward_to_code` so it lines up with the Python function it mirrors.
- Consider collapsing the match_algorithms.rs:73 free fn into the same helper, or at least reconciling the idx vs idx + 1 return convention so the two "code-only" helpers don't silently differ by one. (The free fn's caller at :256 adds +1 itself, so a shared idx + 1 version would let the call site drop the manual +1.)

## Context

Follow-up from PR #8118 review, which added the `_excluding_comments` variant to fix GREEDY Bracketed absorbing trailing trivia. The doc comments there mitigate the confusion but don't remove the underlying naming trap.

apoorva-01:

The Rust method `skip_stop_index_backward_to_code` kept comments, the opposite of Python's function of the same name, so the name matching Python had inverted behaviour.

- Renamed the comment-keeping one to `skip_stop_index_backward_to_code_or_comment`.
- Gave the canonical name to the `_excluding_comments` variant, since that's the one that mirrors Python.
- Collapsed the duplicate free fn in `match_algorithms.rs` into the method (it returned `idx` vs the method's `idx + 1`, so the caller was adding `+ 1` by hand).

Naming/dedup only, no behaviour change. `fixture_tests` yaml-vs-Python still passes across all dialects.

Fixes #8129.

No.

### Pull Request checklist

### DIFF
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/helpers.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/helpers.rs
@@ -139,7 +139,9 @@ impl<'a> Parser<'a> {
     /// Move an index backward through tokens until tokens[index - 1] is code or comment.
     /// Returns the index after the last code/comment token, or min_idx if none found.
     /// IMPORTANT: Comments are NOT skipped - they should be collected like code tokens!
-    pub(crate) fn skip_stop_index_backward_to_code(
+    /// Note: this keeps comments, unlike Python's `skip_stop_index_backward_to_code`,
+    /// so the name spells out what it actually keeps.
+    pub(crate) fn skip_stop_index_backward_to_code_or_comment(
         &self,
         stop_idx: usize,
         min_idx: usize,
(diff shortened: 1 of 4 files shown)


keraion (approved):
Nice. Thanks for picking this up.

PR #8150, opened 2026-07-14
fix(docs): read stable_version as a nested pyproject key

anxkhn:

`docs/source/conf.py` reads the documented version from `pyproject.toml` with:

```python
stable_version = config.get("tool.sqlfluff_docs", "stable_version")
```

`tomllib.load` returns a nested dict, so the value lives at
`config["tool"]["sqlfluff_docs"]["stable_version"]`. The line above instead
treats the dotted string `"tool.sqlfluff_docs"` as a single flat top-level key
(which does not exist), and `"stable_version"` is then used as the `dict.get`
default. As a result `stable_version` resolves to the literal string
`"stable_version"` rather than the real version (`4.2.2`).

`conf.py` feeds that value into `release` and substitutes `|release|`
throughout the docs via the `source-read` hook. So every `|release|` in the
built docs renders as the word `stable_version`. The most visible symptom is in
the pre-commit guide (`docs/source/production/pre_commit.rst`), where:

```yaml
rev: |release|
```

renders as `rev: stable_version`, which is an invalid git ref. Anyone who
copies the documented `.pre-commit-config.yaml` hits a failure from `pre-commit`
because that revision does not exist.

The fix uses the nested lookup:

```python
stable_version = config["tool"]["sqlfluff_docs"]["stable_version"]
```

so `release` (and every `|release|` substitution, including the pre-commit
`rev:`) resolves to the actual version. A regression test
(`test/docs_test.py`) evaluates the real `stable_version` assignment from
`conf.py` against the committed `pyproject.toml` and asserts it equals
`sqlfluff.__version__`, guarding against the flat-key form returning again.

None expected. The change is confined to how `conf.py` reads one value; it does
not alter the docs content or build wiring. The Sphinx build did not error
before (Sphinx substituted the wrong string silently), and after the change the
same `|release|` tokens simply render the correct version. The new test lives at
`test/` root (matching `python_files = "*_test.py"` / `testpaths = "test"`); it
does not import `conf.py` (that has Sphinx side effects and reads generated
files) but evaluates the single assignment via `ast`, so it stays a faithful
guard without pulling in the docs build.

  - [x] Other: `test/docs_test.py` (a regression test for the docs version resolution; there is no dialect/rule/autofix fixture bucket for `conf.py`).
- [x] Added appropriate documentation for the change. (No doc content change is needed; the fix restores the intended rendering of the existing docs.)
- [ ] Created GitHub issues for any relevant followup/future enhancements if appropriate. (None needed.)

### AI-assisted disclosure (required by CONTRIBUTING.md)

---
## Summary by cubic
Fix the docs version lookup to read the nested `stable_version` from `pyproject.toml`, so `|release|` resolves to the real version. This restores a valid `rev:` in the pre-commit guide and stops rendering "stable_version".

- **Bug Fixes**
  - Use `config["tool"]["sqlfluff_docs"]["stable_version"]` in `docs/source/conf.py`.
  - Add a regression test (`test/docs_test.py`) that evaluates `stable_version` in `conf.py` against `pyproject.toml` and guards against the literal `"stable_version"`.

<sup>Written for commit bb00e832c399d867794233948cc648a5b8f88bb8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8150?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -27,7 +27,7 @@ sys.path.append(os.path.abspath("./_ext"))
 # (we use the config file to avoid actually loading any python here)
 with open("../../pyproject.toml", "rb") as config_file:
     config = tomllib.load(config_file)
-stable_version = config.get("tool.sqlfluff_docs", "stable_version")
+stable_version = config["tool"]["sqlfluff_docs"]["stable_version"]
 
 # -- Project information -----------------------------------------------------
 
--- a/test/docs_test.py
+++ b/test/docs_test.py
@@ -10,8 +10,6 @@ if sys.version_info >= (3, 11):
 else:  # pragma: no cover
     import tomli as tomllib
 
-import sqlfluff
-
 REPO_ROOT = Path(__file__).parent.parent
 CONF_PY = REPO_ROOT / "docs" / "source" / "conf.py"
 PYPROJECT = REPO_ROOT / "pyproject.toml"
@@ -40,8 +38,8 @@ def _eval_conf_stable_version(config: dict) -> object:
     raise AssertionError("No `stable_version` assignment found in docs/source/conf.py")
 
 
-def test_docs_stable_version_matches_package_version():
-    """The docs stable version must resolve to the real package version.
+def test_docs_stable_version_reads_nested_config_key():
+    """The docs stable version must resolve from the nested config key.
 
     ``docs/source/conf.py`` substitutes ``|release|`` throughout the docs (e.g.
     the ``rev:`` in the pre-commit guide). Previously it read the nested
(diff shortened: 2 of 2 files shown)


peterbud (changes requested):
Hi @anxkhn - thanks for this PR!

The fix in `conf.py` looks right, but the new test in `docs_test.py` is asserting the wrong invariant. `stable_version` is intentionally managed separately from the package version during prerelease prep, so requiring it to equal `sqlfluff.__version__` will break future alpha/version-bump workflows. The test should verify that conf.py reads the nested `tool.sqlfluff_docs.stable_version` key correctly, not that it always matches the package version.

peterbud (approved):
Thanks for the contribution!

PR #8155, opened 2026-07-15
fix(oracle): allow trailing-dot numeric literals like 1.

### LINKED ISSUE #8110: [oracle] Numbers like 1. can't be parsed
### Search before asking

### What Happened

### Expected Behaviour

The example should work fine.

### Observed Behaviour

Linting the example below gives the following errors:
`Line 3, Position 14: Found unparsable section: '= 1.;'`
`Unable to lex characters: '1.;'`

### How to reproduce

```sql
select ex_col
from ex_tab
where ex_col = 1.;
``` 

### Dialect

oracle

### Version

4.2.1
python 3.11.3

### Configuration

[sqlfluff]
dialect = oracle

### Code of Conduct

Otto-Deviant1904:
Fixes #8110

Oracle rejected numerics such as `1.` because the numeric_literal
lexer used a negative lookbehind (`(?<!\.)`) where ANSI uses a
positive one (`(?<=\.)`) to allow a trailing decimal point.

Aligned that assertion with ANSI (still without ANSI's bare `.\d+`
form). Added an oracle parse fixture covering the reporter's case.

AI disclosure: assisted with investigation and the fixture. I verified
the fail→pass behavior and reviewed the regex change myself.

Made with [Cursor](https://cursor.com)

---
## Summary by cubic
Allow trailing-dot numeric literals (e.g., 1.) in the Oracle dialect by updating the `numeric_literal` regex to match ANSI behavior (no leading-dot form). Keeps Oracle size suffixes (K/M/G/T/P/E) working, blocks digit-prefixed identifiers like `1abc` from splitting, and adds parse fixtures to prevent regressions (fixes #8110).

<sup>Written for commit 7d0256db579ba1a7fa7ad060859d0324ef6e1b96. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8155?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_oracle.py
+++ b/src/sqlfluff/dialects/dialect_oracle.py
@@ -107,12 +107,10 @@ oracle_dialect.patch_lexer_matchers(
         ),
         RegexLexer(
             "numeric_literal",
-            # Like ANSI, but:
-            # - no bare leading-dot form (.\d+)
-            # - no trailing (?=\b) gate (that blocked Oracle size suffixes
-            #   such as 256K). Trailing-dot numerics like 1. are allowed by
-            #   the \d+\.(?![\.\w]) alternative (#8110).
-            r"(?>\d+\.\d+|\d+\.(?![\.\w])|\d+)(\.?[eE][+-]?\d+)?",
+            # Like ANSI, but no bare leading-dot form (.\d+). Allow trailing-dot
+            # numerics (1.) via \d+\.(?![\.\w]), keep (?=\b) for digit/identifier
+            # splits, and also allow Oracle size suffixes (256K) (#8110).
+            r"(?>\d+\.\d+|\d+\.(?![\.\w])|\d+)(\.?[eE][+-]?\d+)?((?<=\.)|(?=\b)|(?=[KMGTkmg]\b))",
             LiteralSegment,
         ),
     ]
--- a/src/sqlfluff/dialects/dialect_oracle.py
+++ b/src/sqlfluff/dialects/dialect_oracle.py
@@ -107,7 +107,9 @@ oracle_dialect.patch_lexer_matchers(
         ),
         RegexLexer(
             "numeric_literal",
-            r"(?>\d+\.\d+|\d+\.(?![\.\w])|\d+)(\.?[eE][+-]?\d+)?((?<!\.)|(?=\b))",
+            # Same shape as ANSI, but without a bare leading-dot form (.\d+).
+            # The (?<=\.) branch allows trailing-dot numerics such as 1. (#8110).
+            r"(?>\d+\.\d+|\d+\.(?![\.\w])|\d+)(\.?[eE][+-]?\d+)?((?<=\.)|(?=\b))",
             LiteralSegment,
         ),
     ]
--- a/src/sqlfluff/dialects/dialect_oracle.py
+++ b/src/sqlfluff/dialects/dialect_oracle.py
@@ -107,9 +107,12 @@ oracle_dialect.patch_lexer_matchers(
         ),
         RegexLexer(
             "numeric_literal",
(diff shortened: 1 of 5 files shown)


peterbud:
This breaks existing tests

Otto-Deviant1904:
Good catch — thanks.

The first change borrowed ANSI's trailing `(?=\b)` gate, which rejected
Oracle size suffixes like `256K` / `64K` (no word boundary between the
digits and `K`). That made the duplicate-STORAGE rejection tests parse
cleanly when they should still fail.

Updated the lexer so trailing-dot numerics such as `1.` still work, but
without that `(?=\b)` requirement. Locally:
- `test/dialects/oracle_test.py` — 18/18 pass (including the two that failed in CI)
- `trailing_decimal_numeric` fixture — still pass

Otto-Deviant1904:
Follow-up for the cubic note on `1abc`:

That case was already accepted on main via the old `(?<!\.)` alternative,
but dropping the boundary entirely made the intent unclear. The lexer now
allows trailing-dot numerics (`1.`), keeps Oracle size suffixes (`256K`),
and rejects digit-prefixed identifiers (`1abc`) again via
`(?=\b)|(?=[KMGTkmg]\b)`.

peterbud:
Pls merge main into this branch to eliminate the rust lint error

peterbud:
Hi @Otto-Deviant1904 , this is moving in the right direction: the trailing-dot case is fixed, and the earlier `1abc` concern looks addressed.

I would still hold merge for one more pass because this change appears to regress valid Oracle size literals like `10K`, `10P`, and `10E`. The Oracle dialect needs targeted regression coverage for both sides of this behavior: accept trailing-dot numerics, reject digit-prefixed identifiers, and continue to parse Oracle size suffixes. Without those regression tests, this is too easy to break again.

chuenchen309:
Ran the two regexes against the size-suffix cases to pin down exactly what regresses (Python `re` handles the atomic group + lookarounds the same way the lexer does):

```
input          main   #8155
10K / 10M / 10G  10     10     ✓ still lex
10T              10     10     ✓  (but 10t → no match)
10P              10     —      ✗ regress
10E              10     —      ✗ regress
1.  / 123.       —      ok     ✓ the intended fix
```

So the trailing-dot fix itself is solid — the only gap is the suffix guard `[KMGTkmg]`:

- missing `P` (peta) and `E` (exa) → `10P` / `10E` stop lexing;
- case-asymmetric — it has `T` but not `t` (and neither `p` nor `e`).

Oracle's size units are K/M/G/T/P/E and are case-insensitive, so a complete guard would be `[KMGTPEkmgtpe]` (or `(?i:[KMGTPE])`). That covers the regression peterbud flagged plus the "keep size suffixes" leg of the regression tests. One small note: `10K`/`10M`/`10G` themselves actually stay fine on this branch — the ones that break are `10P`, `10E`, and lowercase `10t`.

(Not a maintainer — just adding the exact character-set gap so this is easy to close out.)

Otto-Deviant1904:
Thanks — that character-set gap matches what I was seeing.

Widened the numeric size-suffix lookahead to `[KMGTPEkmgtpe]` so it lines up with `SizeClauseGrammar` (K/M/G/T/P/E, case-insensitive). Added a small STORAGE fixture covering `10P` / `10E`. Trailing-dot (`1.`) and the `1abc` rejection are unchanged.

peterbud (approved):

PR #8157, opened 2026-07-15
feat: publish production and development Docker images

### LINKED ISSUE #7868: Latest docker images incompatible with gitlab CI pipelines

hoodadt:
- Use DHI Images
- Add separate prod and dev Dockerfile stages.
- Now publish sqlfluff:latest (production) and sqlfluff:latest-dev (development) only.

The dev stage is intentionally the final Dockerfile stage, so developers cloning the repository can build it with: ` docker build -t sqlfluff-dev .` 

The production image can still be built explicitly with: `docker build --target prod -t sqlfluff .`

This default can be changed if maintainers prefer production to remain the final stage.

fixes #7868


keraion:
Hi @hoodadt

How does this differ from the `latest-dhi-dev` and `latest-dhi` images?

hoodadt:
Hi,
Previously:
- latest was the standard Python image.
- latest-dhi was the minimal DHI image.
- latest-dhi-dev was the DHI image with development tools and a shell.

latest and latest-dhi-dev were already quite similar from a user perspective. Both are Python 3.14 development-capable images based on Debian 13 running as unrestricted user.

With this change, DHI becomes the default:
- latest is the minimal production DHI image.
- latest-dev is the DHI development image with a shell and package manager.

So the main difference is that the separate -dhi variants are removed, and the DHI images become the primary published images.
I also added a dedicated development stage to the Dockerfile. This means users no longer need to switch between different base images. They can select either the production or development target from the same Dockerfile.

peterbud:
Adding @123Haynes for his perspective

hoodadt:
Hi @123Hayn, just following up on this PR. When you have a chance, could you please take a look?

123Haynes:
Hi,

sorry for the late reply.  
This works as you described, but changing the latest image to dhi is a breaking change from a users perspective.  
* existing users will need to update if they are currently using the dhi variant images
* changing the latest tag to the production image without a shell will break existing pipelines of users.

So at the very least this change needs some documentation.  
I'm honestly not sure if this is worth it vs keeping the existing structure.  
From a users perspective this either breaks stuff, or doesn't change anything at all depending on their pipelines, just so we have 2 dockerimages less and nicer defaults.

hoodadt:
To avoid breaking existing pipelines, I can keep `latest` as the development-capable image using the DHI `-dev` base and publish the shell-less hardened image separately as `sqlfluff:latest-prod`.

That would give us:

* `sqlfluff:latest` - DHI development image with a shell and package manager, preserving existing workflows.
* `sqlfluff:latest-prod` - minimal, hardened DHI image.

I can also keep dedicated `dev` and `prod` Dockerfile stages so the build targets and intended usage are clear. I think this is cleaner than change images in github workflows.

The DHI `-dev` image also seems like a better default than the regular Python slim image. It preserves the tools users expect while providing a more curated and security-focused base.

alanmcruickshank:
Based on the feedback from @123Haynes I'm currently thinking that we shouldn't make this change.

- It causes headaches for anyone depending on these images, because it's a breaking change.
- It doesn't look like it's a significant maintainability gain (yes fewer images, but not less code).
- It doesn't look like we're adding significant functionality that doesn't already exist (we're just retagging existing functionality).

I'm struggling to see why it's worth causing users the potential downtime in their deployments without a clear upside to make it worth it. If there is an upside here that I'm not seeing, I'd suggest a rethink of the approach to make that upside bigger, or reopen this PR with a clearer rationale.

Thanks for your contribution @hoodadt - I know it's touch when contributions aren't accepted. I appreciate your efforts in trying to improve the project and it's security posture 👍 .

PR #8161, opened 2026-07-16
perf(templater): skip Jinja machinery for template-free files

rubytobi:
- In `JinjaTemplater.process_with_variants` (`src/sqlfluff/core/templaters/jinja.py`), if the input contains none of `{{`, `{%`, `{#` and no macro config is set (`neither load_macros_from_path` nor a `macros` config section), the file is returned immediately as a literal `TemplatedFile`, bypassing environment construction, Jinja rendering, and slice tracing.
- Macro loading is excluded from the fast path because it's validated at environment construction time and can raise user-facing errors that must be preserved.

[CodSpeed](https://app.codspeed.io/tobycraft/sqlfluff/runs/compare/6a5723f54f11c32691035e66..6a581c2353a78ed41bff1b46?q=mode%3Awalltime) shows 6% improvement for TPC-H and 3% for TPC-DS.

One could argue, if there is no Jinja templating syntax one should use either the `raw` templater or use the per file directives. However, this requires the user to manage this case by case. A generally more flexible and faster JinjaTemplater seems like a better approach.

No

### Pull Request checklist

  - Other.

---
## Summary by cubic
Skip Jinja templating for files with no Jinja markers and no macro or library config, and short-circuit variant processing when everything is covered. Improves templating speed (~6% on TPC-H, ~3% on TPC-DS) without changing behavior.

- **Bug Fixes**
  - Exclude empty input from the fast path to avoid creating a zero-length literal slice and spurious variant coverage.
  - Validate `library_path` in the fast path so broken imports still raise as expected.

- **Refactors**
  - Python: short-circuit `process_with_variants()` when no uncovered literals to avoid Jinja env construction.
  - Rust: add `TemplatedFileCacheEntry` and `TemplatedFileCache` aliases and update `PY_TEMPLATED_FILE_CACHE` to satisfy clippy.

<sup>Written for commit 383b32a135fb69763a96f44edfebc70ec6b8b328. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8161?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/cache.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/cache.rs
@@ -154,6 +154,11 @@ impl TableParseCache {
     pub fn len(&self) -> usize {
         self.cache.len()
     }
+
+    /// Iterate over cache entries (for analysis/debugging)
+    pub fn iter(&self) -> impl Iterator<Item = (&TableCacheKey, &TableCacheValue)> {
+        self.cache.iter()
+    }
 }
 
 impl Default for TableParseCache {
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/cache.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/cache.rs
@@ -154,11 +154,6 @@ impl TableParseCache {
     pub fn len(&self) -> usize {
         self.cache.len()
     }
-
-    /// Iterate over cache entries (for analysis/debugging)
-    pub fn iter(&self) -> impl Iterator<Item = (&TableCacheKey, &TableCacheValue)> {
-        self.cache.iter()
-    }
 }
 
 impl Default for TableParseCache {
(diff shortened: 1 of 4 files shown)


peterbud (commented):
Hi @rubytobi - you are on a streak here...

There’s still one behavior regression in this PR I believ.

In src/sqlfluff/core/templaters/jinja.py, the new fast path skips `construct_render_func()` for template-free files, which means it also skips Jinja `library_path` loading. That changes observable behavior: a broken library import now passes silently for raw SQL files, whereas the normal Jinja setup still raises an error. Macro validation was preserved explicitly, but library validation needs probably the same treatment.

The `process_with_variants()` follow-up looks fine. Would it make sense to add regression test for the fast path?

rubytobi:
We use sqlfluff heavily, so any improvement is well received. Appreciate your prompt and consistent review!

peterbud (approved):

PR #8163, opened 2026-07-16
Snowflake: parse scripting bind variables (:var) in expressions

### LINKED ISSUE #5427: Snowflake scripting cannot parse variable binding
### Search before asking

### What Happened

When writing Snowflake stored procedures or custom functions in Snowflake SQL, it is lacking support for binding variables. E.g., parsing the following code fails:

```
CREATE OR REPLACE PROCEDURE SOME.INPUT.I_AM_A_PROCEDURE()
RETURNS BOOLEAN
LANGUAGE SQL
AS
BEGIN

-- Declare variable
LET PROCESS VARCHAR := 'THIS_PROCEDURE_PROCESS_KEY';

-- Try to bind variable
INSERT INTO SOME.INPUT.DEBUG (PROCESS, MSG) VALUES (:PROCESS, 'START RUN');
...
END;
```

Instead, we are getting the message `Found unparsable section: ":PROCESS, 'START RUN'"`

### Expected Behaviour

This and other binding scenarios described in https://docs.snowflake.com/en/developer-guide/snowflake-scripting/variables should parse.

### Observed Behaviour

I think the problem is, that the Snowflake did not support scripting until I introduced it recently, however, in a non-scripting environment variable binding using the colon (`:`) is not used, so there is no support. 

I am not 100% sure yet, how to make sure this can be enabled in SQL expressions in scripting, but not outside of scripting. Have to figure this out before working on a PR.

Help would be appreciated!

### How to reproduce

Just run the code above

### Dialect

Snowflake

### Version

sqlfluff, version 2.3.3
Python 3.11.6

### Configuration

```
[sqlfluff]

# Or run 'sqlfluff dialects'
dialect = snowflake

# Affect only .sql files in the current directory and below
sql_file_exts = .sql

templater = raw

exclude_rules = ambiguous.column_count, structure.column_order, references.keywords

# something a little longer.
max_line_length = 120

# processes = -1 # default -1

# Ignore file length

[sqlfluff:templater:dbt]
project_dir = ./

[sqlfluff:indentation]
allow_implicit_indents = True
tab_space_size = 2

aliasing = explicit
aliasing = explicit
[sqlfluff:rules:aliasing.length]
min_alias_length = 1

capitalisation_policy = upper
capitalisation_policy = upper
capitalisation_policy = upper
```

### Code of Conduct

adhavan18:

fixes #5427

Snowflake Scripting bind variables (e.g. `:my_variable`) could only be parsed inside `EXECUTE IMMEDIATE ... USING (...)` lists. Using one in a SQL statement — the primary way they are used per the [Snowflake docs](https://docs.snowflake.com/en/developer-guide/snowflake-scripting/variables#using-a-variable-in-a-sql-statement-binding) — failed to parse:

```sql
INSERT INTO debug_log (process, msg) VALUES (:process, 'START RUN');  -- unparsable
SELECT a FROM t WHERE a = :process;                                  -- unparsable
UPDATE t SET a = :new_value WHERE id = :target_id;                   -- unparsable
```

`SELECT :process` also silently **mis-parsed** — as a bare colon `variable` with the variable name swallowed as an alias (visible in the old `let.yml` fixture tree).

This change adds the already-existing `BindVariableSegment` to the Snowflake `LiteralGrammar` (next to `ReferencedVariableNameSegment`, the `$var` equivalent), so bind variables parse wherever a literal is accepted and produce a proper `bind_variable(colon_prefix + variable)` node.

One note on the original repro: its `CREATE OR REPLACE PROCEDURE SOME.INPUT.…` also fails because `INPUT` is a reserved keyword clash on the schema name — that is unrelated to binding and out of scope here (the fixture uses a different schema name).

No. Semi-structured accessors (`SELECT v:name FROM t`) are unaffected — they attach to a preceding column reference, while a bind variable requires the colon in expression-start position. Session variables (`$var`), array accessors and the full Snowflake dialect suite all pass (629 passed; the one updated fixture is `let.yml`, whose stored tree encoded the old mis-parse of `SELECT :variable`).

### Pull Request checklist

- Included test cases to demonstrate any code changes:
  - `.sql`/`.yml` parser test case added at `test/fixtures/dialects/snowflake/snowflake_scripting_bind_variables.sql` (+ generated `.yml`) covering INSERT/SELECT/WHERE/UPDATE/DELETE binds and the full procedure from the issue.
  - `let.yml` regenerated (tree corrected from the old mis-parse).
- Added appropriate documentation for the change (inline grammar comment with doc link).
- Created GitHub issues for any relevant followup/future enhancements if appropriate — n/a.

---
## Summary by cubic
Parse Snowflake Scripting bind variables (:var) in SQL expressions and correct the `SELECT :var` mis-parse. Require the colon and variable name to be adjacent so malformed inputs (e.g., `: WHERE`) fail cleanly; fixes #5427.

- **Bug Fixes**
  - Added `BindVariableSegment` to Snowflake `LiteralGrammar` so `:var` parses wherever a literal is allowed (VALUES, WHERE, SET, SELECT), and corrected `SELECT :var` to produce a `bind_variable` node.
  - Set `allow_gaps=False` so the colon and name must be adjacent, preventing accidental matches like `: WHERE`.
  - Added fixture tests for INSERT/SELECT/WHERE/UPDATE/DELETE and the issue repro; regenerated `let.yml`.

<sup>Written for commit daa7e8fa34f40356aec32b7aaae8ea9236df83c9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8163?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_snowflake.py
+++ b/src/sqlfluff/dialects/dialect_snowflake.py
@@ -695,6 +695,11 @@ snowflake_dialect.replace(
     LiteralGrammar=ansi_dialect.get_grammar("LiteralGrammar").copy(
         insert=[
             Ref("ReferencedVariableNameSegment"),
+            # Snowflake Scripting bind variables, e.g. :my_variable. These are
+            # usable wherever a literal value is expected (VALUES lists, WHERE
+            # clauses, etc.).
+            # https://docs.snowflake.com/en/developer-guide/snowflake-scripting/variables#using-a-variable-in-a-sql-statement-binding
+            Ref("BindVariableSegment"),
         ]
     ),
     AccessorGrammar=AnyNumberOf(
--- a/src/sqlfluff/dialects/dialect_snowflake.py
+++ b/src/sqlfluff/dialects/dialect_snowflake.py
@@ -10108,9 +10108,14 @@ class BindVariableSegment(BaseSegment):
 
     type = "bind_variable"
 
+    # allow_gaps=False so the colon and the variable name must be adjacent.
+    # Otherwise, when this is reachable in general expressions, input like
+    # `SET a = : WHERE id = 1` would parse `: WHERE` as a bind variable and
+    # corrupt the statement structure instead of failing cleanly.
     match_grammar = Sequence(
         Ref("ColonPrefixSegment"),
         Ref("LocalVariableNameSegment"),
+        allow_gaps=False,
     )
 
 
--- a/test/fixtures/dialects/snowflake/let.yml
+++ b/test/fixtures/dialects/snowflake/let.yml
@@ -3,7 +3,7 @@
 # computed by SQLFluff when running the tests. Please run
 # `python test/generate_parse_fixture_yml.py`  to generate them after adding or
 # altering SQL files.
-_hash: 1ca3842cfc1260c844ba4fb9c33b0b9bfe0a36b89bcd7194d999b1e2e37d6c1e
+_hash: 224d88960ce97b1f2e0f9faf4645cb6df93034eea4f41210c6a6f30833ec018b
 file:
 - statement:
     scripting_block_statement:
@@ -170,9 +170,9 @@ file:
       select_clause:
         keyword: SELECT
         select_clause_element:
-          variable: ':'
-          alias_expression:
-            naked_identifier: variable
+          bind_variable:
+            colon_prefix: ':'
+            variable: variable
       from_clause:
         keyword: FROM
         from_expression:
--- a/test/fixtures/dialects/snowflake/snowflake_scripting_bind_variables.sql
+++ b/test/fixtures/dialects/snowflake/snowflake_scripting_bind_variables.sql
@@ -1,0 +1,25 @@
+-- Snowflake Scripting bind variables used in SQL statements.
+-- https://docs.snowflake.com/en/developer-guide/snowflake-scripting/variables#using-a-variable-in-a-sql-statement-binding
+-- https://github.com/sqlfluff/sqlfluff/issues/5427
+
+-- Bind variables in DML value positions.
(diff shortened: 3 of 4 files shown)


peterbud (changes requested):
Hi @adhavan18 - thanks for this PR, and welcome!

There’s one blocking issue here: adding `BindVariableSegment` to Snowflake’s `LiteralGrammar` makes the existing bind-variable matcher reachable in general expressions, but that matcher currently allows gaps. As a result, invalid input like `UPDATE t SET a = : WHERE id = 1;` can parse `: WHERE` as a bind variable and corrupt the statement structure instead of failing cleanly. I’d tighten the bind-variable grammar in so the colon and variable name must be adjacent, or similar.

adhavan18:
Thanks @peterbud — good catch, fixed and pushed.

Added `allow_gaps=False` to `BindVariableSegment` so the colon and variable name must be adjacent. Your case now fails cleanly:

```
UPDATE t SET a = : WHERE id = 1;   -- Found unparsable section (was: `: WHERE` mis-parsed as a bind variable)
SELECT a FROM t WHERE a = : ;       -- Found unparsable section
```

while valid binds are unaffected (`:process`, `WHERE a = :b AND c = :d`, `IN (:a, :b, :c)`, `SELECT :b`). Semi-structured accessors (`v:name`), session variables (`$var`) and the full Snowflake dialect suite still pass (630 tests).

peterbud (approved):
Thanks @adhavan18 !

Issue #6507, opened 2024-12-14
AM08 and Redshift natural joins


natsirtguy:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

SQLFluff had a false positive AM08 (Implicit cross join detected) for a natural join in Redshift (see [here](https://docs.aws.amazon.com/redshift/latest/dg/r_Join_examples.html) for documentation).

Example query:
```
select foo.x
from foo
natural join bar
```

Output:
```
L:   3 | P:   1 | AM08 | Implicit cross join detected. [ambiguous.join_condition]
```

### Expected Behaviour

No AM08 detection **or** perhaps another rule that recommends against natural joins entirely (since the columns used in the join condition are implicit).

### Observed Behaviour

Example query:
```
select foo.x
from foo
natural join bar
```

Output:
```                                                                          
L:   3 | P:   1 | AM08 | Implicit cross join detected.
                       | [ambiguous.join_condition]
```

### How to reproduce

Create a file with
```
select foo.x
from foo
natural join bar
```
in it and run `sqlfluff lint <file> --dialect redshift`

### Dialect

Redshift

### Version

sqlfluff, version 3.3.0
Python 3.12.6

### Configuration

.sqlfluff
```
[sqlfluff]
dialect = redshift
max_line_length = 100
```

### Are you willing to work on and submit a PR to address the issue?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

rogalski:
@natsirtguy Thanks for the report.

We already ignore positional joins from DuckDB or JOIN ... USING. This falls into same bucket, I'll prep a fix.

PR #8165, opened 2026-07-16
Fix CV12 false positive on NATURAL JOIN

chuenchen309:

CV12's allowlist of join keywords that legitimately have no `ON` condition omits `NATURAL`, so a `NATURAL JOIN` with a `WHERE` clause is flagged.

```sql
SELECT foo.a, bar.b FROM foo NATURAL JOIN bar WHERE bar.x > 3
```

```
CV12 | Use `JOIN ... ON ...` instead of `WHERE ...` for join conditions.
CV12 | Use `JOIN ... ON ...` instead of `WHERE ...` for join conditions.
```

The advice can't be followed: a `NATURAL JOIN` cannot take an `ON` clause, and SQLFluff's own parser agrees —

```
$ # SELECT foo.a FROM foo NATURAL JOIN bar ON foo.k = bar.k
PRS | Found unparsable section: 'ON foo.k = bar.k'
```

`sqlfluff fix` therefore can't do anything useful either. It builds the fix, and the unparsable-output guard rejects it:

```
Fixes for CV12 not applied, as it would result in an unparsable file.
Please report this as a bug with a minimal query which demonstrates this warning.
```

So this is that report, with the fix — the file is left correct, but the user gets two violations they cannot resolve plus a warning inviting them to file a bug.

`AM08` carries the equivalent allowlist and already exempts `NATURAL` (`AM08.py:65`), added for #6507 — the same false positive in that rule, for the same reason. CV12's list was never updated to match, so this brings the two into line.

The allowlisted siblings (`CROSS`, `USING`, `POSITIONAL`, `APPLY`) each have a `test_noop_*` case in `CV12.yml`; `NATURAL` was missing from both the allowlist and the tests.

No. `NATURAL JOIN` is the only construct affected — it moves from "flagged, unfixable" to "not flagged". Every other CV12 behaviour is unchanged.

### Pull Request checklist

  - [ ] `.sql`/`.yml` parser test cases in `test/fixtures/dialects`.
  - [ ] Other.

- [x] Added appropriate documentation for the change. (The rule's user-facing docstring describes the WHERE→ON convention and is unaffected; the allowlist comment is updated in line with the existing ones.)

### AI disclosure

This change was AI-assisted (Claude), which I'm disclosing per the AI-Assisted Contributions section of CONTRIBUTING.

What I validated myself, by running it rather than reasoning about it:

- The false positive and the "please report this as a bug" warning, reproduced against this branch's parent.
- That `NATURAL JOIN ... ON ...` really is unparsable — i.e. that the rule's advice is impossible to follow — rather than assuming it.
- The `AM08`/#6507 precedent, by reading `AM08.py` and the issue.
- TDD order: added `test_noop_natural_join` and confirmed it fails first (1 failed / 22 passed), then applied the one-word fix and confirmed 23 passed.
- Full `test/rules/` suite: 2436 passed, 101 skipped, no regressions. `ruff` and `black --check` clean.

An initial AI-generated description of this bug claimed `sqlfluff fix` emits unparsable SQL; that turned out to be wrong — the guard catches it — and I corrected the framing above after reproducing it. Flagging that in the interest of the disclosure being meaningful.

---
## Summary by cubic
Fixes rule `CV12` to stop flagging `NATURAL JOIN` for missing `ON`, and ensures WHERE conditions only move onto a join if they reference that join’s table. This removes unfixable warnings and prevents semantics-changing fixes across skipped joins.

- **Bug Fixes**
  - Allowlisted `NATURAL` so `NATURAL JOIN` does not require `ON`.
  - Only hoist a WHERE predicate to a join when it references that join’s table (avoids mis-attributing conditions with `NATURAL`, `CROSS`, or `USING` before a later join).
  - Added tests: `test_noop_natural_join`, `test_fail_natural_join_mixed_with_later_join`, `test_fail_cross_join_mixed_with_later_join`.

<sup>Written for commit a4c9af18c3e69c6350fb4eaff3994554ec38fdfc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8165?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/rules/convention/CV12.py
+++ b/src/sqlfluff/rules/convention/CV12.py
@@ -124,10 +124,11 @@ class Rule_CV12(BaseRule):
             ]
 
             if any(
-                kw.raw_upper in ("CROSS", "POSITIONAL", "USING", "APPLY")
+                kw.raw_upper in ("CROSS", "NATURAL", "POSITIONAL", "USING", "APPLY")
                 for kw in join_clause_keywords
             ):
                 # If explicit CROSS JOIN is used, disregard lack of condition
+                # If NATURAL JOIN is used, disregard lack of condition
                 # If explicit POSITIONAL JOIN is used, disregard lack of condition
                 # If explicit JOIN USING is used, disregard lack of condition
                 # If explicit CROSS/OUTER APPLY is used, disregard lack of condition
--- a/src/sqlfluff/rules/convention/CV12.py
+++ b/src/sqlfluff/rules/convention/CV12.py
@@ -115,9 +115,10 @@ class Rule_CV12(BaseRule):
                 # If UNNEST function is used, disregard lack of condition
                 continue
 
-            encountered_references.add(
-                self._get_from_expression_element_alias(join_table_reference)
+            this_join_reference = self._get_from_expression_element_alias(
+                join_table_reference
             )
+            encountered_references.add(this_join_reference)
 
             join_clause_keywords = [
                 seg for seg in join_clause.segments if seg.type == "keyword"
@@ -155,13 +156,24 @@ class Rule_CV12(BaseRule):
                         )
                         if "dot" in col_ref.descendant_type_set
                     ]
-                    if len(qualified_column_references) > 1 and all(
-                        col_ref.raw_upper.startswith(
-                            tuple(
-                                f"{table_ref}." for table_ref in encountered_references
+                    if (
+                        len(qualified_column_references) > 1
+                        and all(
+                            col_ref.raw_upper.startswith(
+                                tuple(
+                                    f"{table_ref}."
+                                    for table_ref in encountered_references
+                                )
                             )
+                            for col_ref in qualified_column_references
                         )
-                        for col_ref in qualified_column_references
+                        # A condition that doesn't reference this join's own
+                        # table belongs to an earlier join, and moving it here
+                        # would change the query's semantics.
+                        and any(
+                            col_ref.raw_upper.startswith(f"{this_join_reference}.")
+                            for col_ref in qualified_column_references
+                        )
                     ):
                         this_join_clause_subexpressions.add(subexpr_idx)
                         consumed_subexpressions.add(subexpr_idx)
--- a/test/fixtures/rules/std_rule_cases/CV12.yml
+++ b/test/fixtures/rules/std_rule_cases/CV12.yml
@@ -25,6 +25,12 @@ test_noop_crossjoin:
   pass_str: |
     SELECT foo.a, bar.b FROM foo CROSS JOIN bar WHERE bar.x > 3
 
+test_noop_natural_join:
+  # A NATURAL JOIN cannot take an ON clause, so there is nothing to move
+  # the WHERE condition into.
+  pass_str: |
+    SELECT foo.a, bar.b FROM foo NATURAL JOIN bar WHERE bar.x > 3
+
 test_noop_using:
   pass_str: |
     SELECT foo.id, bar.id FROM foo LEFT JOIN bar USING (id)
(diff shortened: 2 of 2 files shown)


peterbud (changes requested):
Hi @chuenchen309  - thanks for this PR and welcome!

Tthis change can mis-fix mixed join queries by letting predicates from a skipped `NATURAL JOIN` get pulled into a later join condition.

For example:

```sql
SELECT * FROM a NATURAL JOIN b LEFT JOIN c WHERE a.k = b.k AND a.m = c.m
```

CV12 currently fixes that to:

```sql
SELECT * FROM a NATURAL JOIN b LEFT JOIN c ON a.k = b.k AND a.m = c.m
```

That moves `a.k = b.k` onto the `LEFT JOIN c`, which changes semantics. The new test only covers the single `NATURAL JOIN` case, so it doesn’t catch this mixed-join regression.

Could you look into this?

chuenchen309:
You're right, and it's worse than a missed test case — thanks for catching it.

Reproducing your example, here's what my PR actually did:

| | `SELECT * FROM a NATURAL JOIN b LEFT JOIN c WHERE a.k = b.k AND a.m = c.m` |
|---|---|
| main | `FAIL`, file **unchanged** — the fix produced unparsable `NATURAL JOIN ... ON` and the guard rejected it |
| my PR | `... LEFT JOIN c ON a.k = b.k AND a.m = c.m` — **applied silently, wrong** |

So I turned a self-rejecting fix into a silently corrupting one. That's a worse trade than the false positive I was fixing.

Chasing it down, the mis-attribution isn't mine — it's already on `main` for `CROSS JOIN` and `USING`, which have no unparsable-output accident to mask it:

```
IN : SELECT * FROM a CROSS JOIN b LEFT JOIN c WHERE a.k = b.k AND a.m = c.m
OUT: SELECT * FROM a CROSS JOIN b LEFT JOIN c ON a.k = b.k AND a.m = c.m   ← main, no PR applied
```

`encountered_references.add()` runs before the skip `continue`, so a skipped join's table stays eligible, and the greedy matcher attaches a condition to the *first* join whose accumulated references cover it — without requiring that the condition reference *that* join's table. Adding `NATURAL` to the skip list just removed the accidental protection.

Pushed a fix for the underlying rule: a condition may only move onto join J if it references J's own table. Both your example and the `CROSS JOIN` one are now tests (`test/rules/`: 2452 passed).

Two things worth your call:

1. **Scope.** This also fixes the pre-existing `CROSS JOIN`/`USING` mis-fix, which is more than the PR title advertises. Same root cause, and I didn't want to leave a known silent-corruption path in place — but if you'd rather have the rule fix separate from the NATURAL JOIN change, happy to split it.

2. **A behavior change no test covered.** `SELECT * FROM a JOIN b WHERE a.x = a.y` used to fix to `JOIN b ON a.x = a.y`; it now stays in the `WHERE` and is reported unfixed. For an INNER JOIN the old fix was semantically equivalent, so this is a small loss of fix coverage. `a.x = a.y` isn't a condition on `b`, so I think it's the right side of the line, but it's a real change and I'd rather flag it than have it slip through.

I also considered just not adding skipped tables to `encountered_references`, but that breaks a legitimate fix (`FROM a CROSS JOIN b JOIN c WHERE b.x = c.x`), so the attribution guard seemed like the narrower cut.

peterbud (approved):
Thanks!

PR #8168, opened 2026-07-16
perf(rust): frame-free terminal evaluation

rubytobi:
The parser previously allocated a full parse "frame" (context object) any time it evaluated a candidate grammar element, even when that candidate was a terminal (a plain token match with no nested sub-grammar). This branch adds fast paths that recognize terminal candidates ahead of time and match them directly against the token stream, skipping frame allocation entirely.

It applies to:
- `OneOf` candidates that are terminals
- `Sequence` elements that are terminals
- `Ref`-to-terminal grammars, plus inlined terminator probes
- `Delimited` / `AnyNumberOf` terminal children
- Consecutive terminal `OneOf` candidates are now looped over rather than recursed, avoiding stack overhead

Common parsing paths (OneOf/Sequence/Delimited/Ref matching) do meaningfully less allocation per token, with no change in matching semantics.

Roughly 1 in 5 `try_terminal_inline` calls actually takes the frame-free path across both workloads:
* TPC-H full suite (22 queries): 21.8% of terminal-inline probes
* TPC-DS full suite (99 queries): 21.1% of terminal-inline probes

| [Benchmark](https://app.codspeed.io/tobycraft/sqlfluff/runs/compare/6a5723f54f11c32691035e66..6a598059c5e2df2326242ba5?q=mode%3Awalltime) | Change | Before | After |
|---|---|---|---|
| test_native_ast_tpch | +2% | 1.1s | 1.1s |
| test_native_ast_tpcds | +2% | 10.7s | 10.5s |
| test_parse_tpch | +1% | 1.1s | 1.1s |
| test_parse_tpcds | +1% | 11.4s | 11.2s |
| test_parse_tpcds_python | 0% | 65.9s | 65.7s |
| test_parse_tpch_python | 0% | 6s | 6s |

Using our internal Athena query, measured locally and always subject to a bit more noise, I get the following numbers:
| Metric | Before | After | Improvement |
 |---|---|---|---|
 | `parse_athena` (criterion, median) | 1.8s | 1.4s | 20% faster |
 | Python-only total parse time | 22.5s | 21.5s | 4% faster |
 | Rust (current) total parse time | 7.3s | 7.3s | noise |
 | Rust (native AST) total parse time | 6.3s | 5.9s | 7.5% faster |

No

### Pull Request checklist

  - Other.

### DIFF
--- a/sqlfluffrs/sqlfluffrs_benchmarks/examples/time_tpc.rs
+++ b/sqlfluffrs/sqlfluffrs_benchmarks/examples/time_tpc.rs
@@ -36,6 +36,8 @@ struct RunStats {
     complete_match_early_exits: usize,
     terminator_checks: usize,
     terminator_hits: usize,
+    terminal_fast_path_hits: usize,
+    terminal_fast_path_misses: usize,
 }
 
 impl RunStats {
@@ -55,6 +57,8 @@ impl RunStats {
             complete_match_early_exits: m.complete_match_early_exits.get(),
             terminator_checks: m.terminator_checks.get(),
             terminator_hits: m.terminator_hits.get(),
+            terminal_fast_path_hits: m.terminal_fast_path_hits.get(),
+            terminal_fast_path_misses: m.terminal_fast_path_misses.get(),
         }
     }
 }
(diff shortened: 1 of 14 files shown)


peterbud (approved):
Let's merge this.

PR #8176, opened 2026-07-18
RF01: resolve UPDATE target alias nested in from_expression (MySQL)

### LINKED ISSUE #6147: Alias for UPDATE table not found as ancestor element
### Search before asking

### What Happened

Unexpected notice because 'p' is found in ancestor "UPDATE":
```plaintext
L: 169 | P: 110 | RF01 | Reference 'p.rowid' refers to table/view not found in
```

for

```sql
UPDATE llx_product AS p SET p.stock= (SELECT SUM(ps.reel) FROM llx_product_stock AS ps WHERE ps.fk_product = p.rowid);
```

running

```bash
sqlfluff --dialect mysql --exclude-rules "LT09,LT05,LT01,CP01,CP02,CP04,ST06,AM04,AL05"
```

### Expected Behaviour

sqlfluff should not issue a notice for this / accept 'p' as an ancestor 

### Observed Behaviour

mysql emits a notice while the query is valid.

### How to reproduce

Unexpected notice because 'p' is found in ancestor "UPDATE":
```plaintext
L: 169 | P: 110 | RF01 | Reference 'p.rowid' refers to table/view not found in
```

### Dialect

mysql

### Version

sqlfluff, version 3.1.0 
Python 3.9.16

### Configuration

Empty, see CLI

### Code of Conduct

vidigoat:
Fixes #6147.

RF01 falsely flags references to the UPDATE target's alias inside correlated subqueries on MySQL/MariaDB:

```sql
UPDATE llx_product AS p
SET p.stock = (
    SELECT SUM(ps.reel)
    FROM llx_product_stock AS ps
    WHERE ps.fk_product = p.rowid
);
```
```
L:5 | P:22 | RF01 | Reference 'p.rowid' refers to table/view not found in the FROM clause or found in ancestor statement.
```

The same statement lints clean with `--dialect ansi`, which isolates the cause: ANSI's grammar puts the UPDATE target's `alias_expression` directly under the statement, where `Selectable.select_info` fabricates the DML alias from a *direct child* lookup. MySQL parses the target as `from_expression → from_expression_element → [table_expression, alias_expression]` — one level deeper — so the alias was never registered and subquery references to it failed to resolve.

Fix: in the DML branch of `select_info`, fall back to the nested location when no direct-child alias exists. The fallback only activates when the direct lookup finds nothing, so ANSI and other dialects are untouched. Living in shared analysis code, it also benefits the other rules built on `Query` (RF02/RF03/ST rules).

New pass-case in RF01.yml (fails on main, passes with the fix); all 80 RF01 cases green; full `test/rules/` + `test/utils/analysis/`: 2474 passed locally.

AI disclosure (per CONTRIBUTING): the diagnosis and patch were AI-assisted; I validated the repro, fix behaviour and test runs locally myself.

---
## Summary by cubic
Fixes RF01 false positives on MySQL/MariaDB where correlated subqueries couldn’t reference UPDATE target aliases. The analyzer now resolves aliases from the UPDATE’s nested from-clause, including all targets in multi-table UPDATEs (resolves #6147).

- **Bug Fixes**
  - In select_info, when no direct alias_expression exists, delegate alias discovery to the canonical from-clause resolver to collect all UPDATE target aliases (joins, nested targets, and comma-style multi-table UPDATEs).
  - Added RF01 pass cases for single-table, join-based multi-table, and comma-style multi-table UPDATE correlated subqueries on mysql; applied ruff-format.
  - Kept legacy behavior: register one anonymous DML alias when none found; other dialects unchanged; shared analysis benefits other Query-based rules.

<sup>Written for commit 681ecd2f746b1379fb63c144ad4b2cdc4c7a2b3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8176?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/utils/analysis/query.py
+++ b/src/sqlfluff/utils/analysis/query.py
@@ -85,6 +85,19 @@ class Selectable:
             # (https://www.postgresql.org/docs/8.2/sql-values.html).
             values = Segments(self.selectable)
             alias_expression = values.children().first(sp.is_type("alias_expression"))
+            if not alias_expression:
+                # In some dialects (e.g. MySQL), the target table of an
+                # UPDATE statement is parsed as a from_expression, so any
+                # alias is nested inside a from_expression_element rather
+                # than being a direct child of the statement.
+                # https://github.com/sqlfluff/sqlfluff/issues/6147
+                alias_expression = (
+                    values.children(sp.is_type("from_expression"))
+                    .children(sp.is_type("from_expression_element"))
+                    .first()
+                    .children()
+                    .first(sp.is_type("alias_expression"))
+                )
             name = alias_expression.children().first(
                 sp.is_type("naked_identifier", "quoted_identifier")
             )
--- a/src/sqlfluff/utils/analysis/query.py
+++ b/src/sqlfluff/utils/analysis/query.py
@@ -8,7 +8,11 @@ from functools import cached_property
 from typing import Generic, NamedTuple, Optional, TypeVar, Union, cast
 
 from sqlfluff.core.dialects.base import Dialect
-from sqlfluff.core.dialects.common import AliasInfo, is_qualified
+from sqlfluff.core.dialects.common import (
+    AliasInfo,
+    get_from_clause_aliases,
+    is_qualified,
+)
 from sqlfluff.core.parser import BaseSegment
 from sqlfluff.dialects.dialect_ansi import ObjectReferenceSegment
 from sqlfluff.utils.analysis.select import (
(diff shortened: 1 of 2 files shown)


chuenchen309:
Reviewed this end-to-end on MySQL and it holds up — nice diagnosis pinning it to the extra `from_expression` level. Independently:

- On `upstream/main` all three shapes below flag `RF01` (the false positive); on this branch all three lint clean **and** still parse with no unparsable sections, so it's a genuine resolution rather than a parse-skip:
  - the single-table and `INNER JOIN` multi-table cases from your fixtures, and
  - the comma-style multi-table update `UPDATE t1 AS a, t2 AS b SET a.x = (SELECT … WHERE c.k = b.id)`.
- No regressions in the shared-analysis consumers: `test/rules/` + `test/utils/analysis/` filtered to RF01/RF02/RF03 + analysis/query is green (404 passed) on the branch.

One coverage note on that third shape. MySQL parses `UPDATE t1 AS a, t2 AS b` as two sibling `from_expression` nodes (each a single `from_expression_element`), not a `join_clause` — so `b` resolves through your `from_expressions.children("from_expression_element")` term, a different branch than the `INNER JOIN` fixture exercises. The fix already covers it, so it might be worth pinning with a third pass-case so a later refactor of that fallback can't silently regress the comma form:

```yaml
test_pass_mysql_comma_multi_table_update_alias_correlated_subquery:
  # https://github.com/sqlfluff/sqlfluff/issues/6147
  pass_str: |
    UPDATE t1 AS a, t2 AS b
    SET a.x = (
        SELECT SUM(c.v)
        FROM t3 AS c
        WHERE c.k = b.id
    );
  configs:
    core:
      dialect: mysql
```

(Not a maintainer — just verified it because I'd been in the MySQL grammar recently.)

peterbud:
Hi @vidigoat pls fix the linting errors.

vidigoat:
Thanks for the ping @peterbud — fixed. The linting job was failing on a ruff-format diff in `query.py`; I've applied `ruff format` (v0.15.5, matching the pre-commit pin) and re-checked ruff/mypy locally. CI is green now.

peterbud (changes requested):
Hi @vidigoat - thanks for this PR.

MySQL `UPDATE` really does have different structure here, especially once joined and multi-table targets are involved, so parsing the target through `from_expression` looks reasonable. The issue is more that the current fix in query.py is manually reaching into that structure to scrape `alias_expression` nodes, which duplicates alias-resolution logic and still seems incomplete for deeper joined targets.

Directionally, I’d suggest reusing the existing `from_expression_element` / `join_clause` alias helpers instead of adding a MySQL-specific fallback in `Selectable.select_info`. That keeps the fix in the right layer, aligns DML handling with the existing alias abstraction, and should make it work for nested join shapes.

vidigoat:
Thanks @peterbud — good call, that's a much better layer for it. I've reworked it to delegate to the existing `get_from_clause_aliases()` helper instead of scraping `alias_expression` nodes by hand. It resolves aliases through the same path a SELECT uses (direct `from_expression_element`s, join clauses, and nested/joined targets), so the duplicated logic is gone and deeper joined targets are handled the same way as everywhere else.

Also added a comma-style multi-table pass case (`UPDATE t1 AS a, t2 AS b`) per @chuenchen309's suggestion, since that shape parses as sibling `from_expression`s rather than a `join_clause`. RF01/RF02/RF03 + the analysis utils are green (511 passed), and ruff + mypy are clean.

peterbud (approved):
Thanks @vidigoat !

PR #8187, opened 2026-07-21
fix(sparksql): allow expression values in SET statements

### LINKED ISSUE #4218: Databricks SET clause with function call
### Search before asking

### What Happened

The issue is that sqlfluff doesn't support SET clause with function calls syntax for databricks (sparksql) dialect.

### Expected Behaviour

To have support and the ability to parse, fix, and lint the following code:

`SET c_date = CURRENT_DATE();`

### Observed Behaviour

Line 1, Position 11: Found unparsable section: ' = CURRENT_DATE()'

### How to reproduce

import sqlfluff
sql = 'SET c_date = CURRENT_DATE();'
print(sqlfluff.parse(sql, dialect='sparksql'))

### Dialect

sparksql

### Version

1.4.5

### Configuration

[sqlfluff]
templater = placeholder
exclude_rules = L031, L034
dialect = databricks

[sqlfluff:rules]
max_line_length = 120
indent_unit = space
tab_space_size = 4

[sqlfluff:layout:type:comma]
spacing_after = single
line_position = leading

[sqlfluff:rules:L010]
capitalisation_policy=lower

[sqlfluff:rules:L011]
aliasing = explicit

[sqlfluff:rules:L012]
aliasing = explicit

[sqlfluff:rules:L030]
extended_capitalisation_policy=lower

[sqlfluff:rules:L063]
extended_capitalisation_policy=lower

[sqlfluff:indentation]
indented_joins = False

param_style = dollar
ENV = 'ENV'
RAW_SCHEMA = 'RAW_SCHEMA'
RAW_DATA_PATH = 'RAW_DATA_PATH'

### Code of Conduct

Otto-Deviant1904:
### Brief summary
Allow `SET key = <expression>` (e.g. `CURRENT_DATE()`) in SparkSQL/Databricks. Fixes unparsable `SET c_date = CURRENT_DATE();`.

### Specific changes
- Extend `SetStatementSegment` with a delimited `PropertyName = Expression` alternative before `PropertyListGrammar`
- Add dialect fixtures for sparksql + databricks

### AI disclosure
AI-assisted implementation; manually reproduced fail→pass and checked existing SET fixtures.

Fixes #4218

Made with [Cursor](https://cursor.com)

---
## Summary by cubic
Allow expression and opaque values in SparkSQL/Databricks SET statements. Values now terminate at semicolon only, and a single key=value assignment won’t split comma-containing payloads.

- **Bug Fixes**
  - Added `SetConfigValueSegment`; `SET key = value` accepts functions, literals, bracketed lists, dotted identifiers, and opaque payloads (URIs, hyphenated tokens, comma lists).
  - Bound opaque values by semicolon only; missing-semicolon recovery is unchanged and may join statements.
  - Updated `SetStatementSegment` to parse a single `PropertyName = SetConfigValue` (no comma-delimited assignments) while keeping the bare `SET key` form.
  - Added sparksql/databricks tests and fixtures, including same-line statement boundaries and Databricks keyword-as-value cases.

<sup>Written for commit f6a7784f9885825b2e9b86607178af1e298e0740. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8187?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


peterbud (changes requested):
H @Otto-Deviant1904 , thanks for this PR as well.

The main concern is that this change models Spark `SET key = value` payloads as full SQL expressions. That fixes `CURRENT_DATE()`, but it also makes config values too semantic: some valid raw config strings still won’t parse, and others can now be linted or rewritten as if they were SQL. For example, values like URI-style paths remain problematic, and token sequences such as `a-b` get treated as arithmetic instead of opaque config text.

I’d suggest narrowing the fix. Instead of routing `SET` values through `ExpressionSegment`, introduce a Spark-specific value grammar for config assignments that covers the real cases you want to support: literals, function-like values if needed, dotted identifiers, and other raw config payloads without applying normal SQL expression semantics. It would also help to add a couple of fixtures for realistic config strings such as paths/URIs so the parser behavior stays aligned with actual Spark `SET` usage.

Otto-Deviant1904:
Thanks @peterbud — good call on the ExpressionSegment overreach.

I narrowed this to a Spark-specific `SetConfigValueSegment` that accepts:
- literals / bracketed literal lists
- function-like values (`CURRENT_DATE()`)
- dotted identifiers (Java class names)
- opaque raw payloads via `Anything()` (so URIs and `a-b` stay non-expression)

Also added fixtures for `s3a://...` and `a-b`. Happy to tweak further if you want the raw fallback tightened.

earfman:
Ran this branch against its merge-base and hit a case the fixtures don't cover: the `Anything()` fallback in `SetConfigValueSegment` doesn't stop at a statement terminator when the `SET` line itself is missing one. It consumes the following statement whole, and the file then lints clean.

Repro — sparksql dialect, no config:

```sql
SET spark.sql.shuffle.partitions = 200
DROP TABLE prod.customers;
```

- merge-base: `L:   2 | P:   1 |  PRS | Found unparsable section: 'DROP TABLE prod.customers;'`
- this branch: **0 violations**

Parse tree on this branch:

```
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |        set_statement:
[L:  1, P:  1]      |            keyword:                                          'SET'
[L:  1, P:  5]      |            property_name_identifier:
                    |                (spark.sql.shuffle.partitions)
[L:  1, P: 34]      |            comparison_operator:
[L:  1, P: 34]      |                raw_comparison_operator:                      '='
[L:  1, P: 36]      |            set_config_value:
[L:  1, P: 36]      |                numeric_literal:                              '200'
[L:  1, P: 39]      |                newline:                                      '\n'
[L:  2, P:  1]      |                word:                                         'DROP'
[L:  2, P:  6]      |                word:                                         'TABLE'
[L:  2, P: 12]      |                word:                                         'prod'
[L:  2, P: 16]      |                dot:                                          '.'
[L:  2, P: 17]      |                word:                                         'customers'
[L:  2, P: 26]      |    statement_terminator:                                     ';'
```

The `DROP TABLE` is absorbed into `set_config_value` as raw `word` tokens, so no rule ever visits it. Same outcome for:

```sql
SET spark.sql.foo = bar
SELECT 1 FROM t;
```

```sql
SET spark.sql.shuffle.partitions = 200
INSERT INTO prod.audit SELECT * FROM staging;
```

Both are `PRS` on the merge-base and 0 violations here.

Why the suite doesn't flag it: every existing sparksql `SET` fixture terminates its statement, so nothing exercises the missing-terminator path. `test/dialects` sparksql + databricks is 576/576 green on this branch.

The user-facing failure mode is that one missing semicolon after a `SET` line silently un-lints everything that follows it, and sqlfluff exits reporting success — which is a worse outcome than the parse error it replaces.

Unrelated, on the cubic comment above about comma-splitting: I couldn't reproduce that one as a regression. `SET spark.sql.foo = a,b;` yields the identical `Found unparsable section: ',b;'` at `L:1 P:22` on both the merge-base and this branch, so it looks pre-existing rather than introduced here.

No strong opinion on the fix shape — bounding the raw fallback with an explicit terminator guard instead of a bare `Anything()` seems like the direction, but you'd know better than me whether that composes with `PropertyListGrammar`. Happy to put up a patch if it would help.

Otto-Deviant1904:
Good catch @earfman. OneOf takes the longest match, so unbounded Anything() was eating the next statement when SET had no trailing semicolon.

The opaque fallback is now terminated on both `;` and newline. Added rejection tests for the DROP/SELECT/INSERT cases you listed (and the opaque URI variant). Thanks for the clear repro.

Otto-Deviant1904:
@peterbud friendly ping when you have a chance.

Since your review I narrowed SET values away from ExpressionSegment, fixed the missing-semicolon absorption cases (including LOAD), and CI is green again. Happy to tweak further if anything still looks off.

peterbud:
Hi @Otto-Deviant1904 - I appreciate the updated PR.

I think however the hand-coded terminator list is the wrong shape for this fix. It already misses follow-on statements in inheriting dialects like Databricks, where `SET ...` can still absorb later statements such as `COMMENT ON` or `DECLARE`. That means the parser boundary is being maintained by a copied keyword list rather than by the dialect’s actual statement grammar, which is brittle and likely to regress as new statement types are added.

Other dialects mostly avoid this by keeping `SET` values syntactically narrow instead of using an open-ended tail. Postgres, T-SQL, and Flink enumerate accepted `SET` forms explicitly, and Hive uses an expression-based value grammar. Spark is harder because it needs to accept opaque unquoted values, so a permissive fallback may still be needed, but it should not depend on an inline starter-keyword list. A better direction would be to extract a dedicated “SET value terminator” grammar that Spark owns and Databricks can extend, or preferably terminate on a real statement-boundary abstraction if that can be made to work in both the Python and Rust parsers.

Otto-Deviant1904:
Thanks - yeah, the keyword list was the wrong approach.

Switched the opaque SET value branch to a statement boundary instead:
SetConfigValueTerminatorGrammar (semicolon; Databricks can extend it),
plus NonCodeMatcher for newline/whitespace. That maps to Rust's
GrammarId::NONCODE, which is what actually works there - a newline
TypedParser gets skipped. Also taught Python greedy_match to honor
NonCodeMatcher.

Drops the copied starter-keyword list, keeps opaque values like
`values` / URIs parsing, and stops SET from swallowing follow-ons in
Spark and Databricks.

Otto-Deviant1904:
Rust CI was failing because rustify table validation rejected GrammarId::NONCODE (0xFFFFFFFE) as an out-of-range terminator id. Allowed that sentinel through validation; rustify build passes locally now.

peterbud:
Hi @Otto-Deviant1904 - thanks for the update.

I see a problem with how the non-code matcher is being used here as the boundary for opaque `SET` values.

Using `NonCodeMatcher` makes the fallback stop on any non-code token, but the meaningful boundary for Spark/Databricks `SET` values is narrower than that, if I'm not mistaken. What we actually care about is statement boundaries such as semicolon, newline, and possibly EOF, not generic whitespace. Treating all non-code as a terminator is broader than the grammar rule and risks cutting off valid opaque values too early.

I also found a concrete regression in the Databricks follow-up in src/sqlfluff/dialects/dialect_databricks.py. Adding `COMMENT`, `DECLARE`, `OPTIMIZE`, `APPLY`, `TAG`, and `UNSET` as `SET` value terminators makes valid bare values like `SET key = comment;` or `SET key = declare;` unparsable. That feels too brittle, because it turns a value parser into a hand-maintained list of statement starters.

I’d suggest aiming for a boundary abstraction based on meaningful statement boundaries instead:
- explicit statement terminators like `;`
- line boundaries for unterminated `SET` statements
- EOF

That would solve the real problem here, which is preventing the opaque fallback from absorbing the next statement, without overfitting to keyword lists or stopping on arbitrary whitespace.

Concretely, I’d recommend:
1. Remove the Databricks keyword terminator list from src/sqlfluff/dialects/dialect_databricks.py.
2. Keep the structured SparkSQL branches for literals, functions, bracketed values, and dotted identifiers in src/sqlfluff/dialects/dialect_sparksql.py.
3. Replace the opaque fallback terminator strategy with a dedicated newline / statement-boundary concept rather than generic `NonCodeMatcher`.
4. Add regression coverage for bare keyword values, newline-separated follow-on statements, and same-line follow-ons.

Otto-Deviant1904:
Thanks - you're right about NonCodeMatcher being too broad, and about the
Databricks keyword list breaking bare values like `comment` / `declare`.

I removed that keyword list. Opaque SET values now terminate on semicolon
and newline only (NewlineMatcher + Rust GrammarId::NEWLINE), so intra-line
spaces stay in the value. Same-line follow-ons are covered via a real
semicolon separator, and newline-separated follow-ons still stop at the
line boundary.

peterbud:
Hi @Otto-Deviant1904 , I think this PR has drifted too far from the original SparkSQL grammar fix. Allowing `SET key = <expression>` seemed like it should stay mostly dialect-local, but this now introduces shared parser behavior, serializer changes, and Rust parser plumbing, which makes the review surface much larger than it needs to be.

Also, I found a concrete regression in the new core parser path: the newline/non-code terminator handling in src/sqlfluff/core/parser/match_algorithms.py drops nested bracket matches that `Anything()` is supposed to preserve. In practice, the new branch zeroes out `inner_matches` and never carries bracket structure forward, whereas the existing keyword-terminator path does. 

I’d strongly suggest to explore whether this can be reduced back toward a narrower dialect-level solution.

@keraion  - your thoughts?

Otto-Deviant1904:
Thanks @peterbud — fair points on both the scope and the `inner_matches` regression.

I'll pull this back toward a narrower SparkSQL/Databricks dialect-level fix and avoid the shared Python/Rust terminator plumbing where possible. Will also specifically cover the bracket/`Anything()` case you called out.

@keraion happy to follow whatever direction you prefer if you have a preferred shape for keeping this dialect-local.

keraion:
I'm gonna agree that this has drifted a bit too far from what should be a simple grammar fix.

I also don't know that newline terminators are the answer as I don't believe that sparksql even supports this. Sparksql strictly expects semicolon delimiters for statements. The issue with finding a forgotten semicolon is currently a gap that exists today, we probably have a few open issues still on it.

Otto-Deviant1904:
Thanks @keraion — agreed.

I've pulled this back to a dialect-only SparkSQL grammar change:
- `SetConfigValueSegment` for literals / functions / dotted ids / opaque payloads (not `ExpressionSegment`)
- opaque values terminate on `;` only
- no newline / NonCode / shared Python or Rust parser plumbing

Also dropped the missing-semicolon boundary tests. As you noted, SparkSQL expects semicolon-delimited statements, and recovering a forgotten `;` is an existing gap rather than something this SET-value fix should solve.

Happy to adjust further if you want the opaque branch tightened differently.

peterbud (approved):
OK, let's merge this. We need to keep in mind that in case of a missing semicolon the next statement now will be silently swallowed and parsed as a SET expression.

PR #8191, opened 2026-07-21
BigQuery: allow a pipe query after a WITH clause

ricardoalencar-pr:

`WithCompoundStatementSegment` (inherited from ANSI) ends with `NonWithSelectableGrammar`, so the query following the CTE list may only be a `SELECT` / set expression — never a pipe query. As a result a CTE chain whose final query uses pipe syntax is reported as an **unparsable section**:

```sql
WITH cte AS (
    SELECT 1 AS n
)

FROM cte
|> WHERE n > 0
|> SELECT n;
```

This is despite the fact that bare top-level pipe queries (#6870) and pipe syntax *inside* CTE bodies (#7180) already parse fine.

This PR overrides `WithCompoundStatementSegment` in the BigQuery dialect so its trailing element is `OneOf(NonWithSelectableGrammar, PipeStatementSegment)`, mirroring how `CTEDefinitionSegment` already accepts a `PipeStatementSegment` inside CTE bodies.

### Are there any other side effects?

One existing fixture case (`WITH … SELECT … |> INTERSECT DISTINCT BY NAME …`) now nests as `with_compound_statement > pipe_statement` instead of `pipe_statement > with_compound_statement`. Both fully parse; the new nesting treats the `WITH` as the outer scope and its trailing pipe query as the inner statement, which matches the semantics more closely. The fixture YAML is regenerated accordingly.

### Tests

- Added a `WITH … FROM … |> …` case to `test/fixtures/dialects/bigquery/pipe_statement.sql` and regenerated its YAML.
- `pytest test/dialects/dialects_test.py -k bigquery` — 429 passed.
- `ruff check` / `ruff format --check` / `yamllint` clean on the changes.

Related to the BigQuery pipe-syntax support in #6870 / #7180.

---
## Summary by cubic
Allow BigQuery `WITH` statements to end with a pipe query (`FROM ... |> ...`). Fixes unparsable CTE chains where the final query uses pipe syntax.

- **Bug Fixes**
  - Extended BigQuery `WithCompoundStatementSegment` to accept `PipeStatementSegment` after the CTE list.
  - Added a `WITH … FROM … |> …` test and updated fixtures; an existing case now nests as `with_compound_statement > pipe_statement`.

<sup>Written for commit 283576a2985696d9ce8b58105466258c980eefac. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8191?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -3871,6 +3871,27 @@ class CTEDefinitionSegment(ansi.CTEDefinitionSegment):
     )
 
 
+class WithCompoundStatementSegment(ansi.WithCompoundStatementSegment):
+    """A `WITH` statement, which in BigQuery may end in a pipe query.
+
+    BigQuery allows the query following the CTE list to be written with pipe
+    syntax (``FROM ... |> ...``), not just a SELECT / set expression. This
+    extends the ANSI definition so the trailing element also accepts a
+    ``PipeStatementSegment``, mirroring how ``CTEDefinitionSegment`` already
+    allows pipe syntax inside CTE bodies.
+    """
+
+    match_grammar = ansi.WithCompoundStatementSegment.match_grammar.copy(
+        insert=[
+            OneOf(
+                Ref("NonWithSelectableGrammar"),
+                Ref("PipeStatementSegment"),
+            ),
+        ],
+        remove=[Ref("NonWithSelectableGrammar")],
+    )
+
+
 class ChainedFunctionCallSegment(BaseSegment):
     """Postfix chained function call accessor: .FUNC(args...) in BigQuery."""
 
--- a/test/fixtures/dialects/bigquery/pipe_statement.sql
+++ b/test/fixtures/dialects/bigquery/pipe_statement.sql
@@ -163,3 +163,11 @@ WITH bar AS (
 )
 
 SELECT * FROM bar;
+
+WITH bar AS (
+    FROM foo
+)
+
+FROM bar
+|> WHERE sales >= 3
+|> SELECT item, sales;
--- a/test/fixtures/dialects/bigquery/pipe_statement.yml
+++ b/test/fixtures/dialects/bigquery/pipe_statement.yml
@@ -3,7 +3,7 @@
 # computed by SQLFluff when running the tests. Please run
 # `python test/generate_parse_fixture_yml.py`  to generate them after adding or
 # altering SQL files.
-_hash: 54e1c2e3c2c96bb46231eb8080e885c5ed15cde697a19d81e6ec0c53f01de497
+_hash: 2759f64a6e9b046386490936ace2237520f8b35608dda19cb907ad4353d3c75d
 file:
 - statement:
     pipe_statement:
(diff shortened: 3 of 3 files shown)


peterbud (commented):
Hi @ricardoalencar-pr  - thanks for opening this PR and welcome.

Could we explore whether this is better modeled by extending `NonWithSelectableGrammar` in BigQuery rather than overriding `WithCompoundStatementSegment` directly?

Other dialects seem to use the grammar-level approach when they want to broaden what counts as a selectable in a given context, so it would be useful to confirm whether pipe queries should be treated that way here as well, or whether this really is intended to be a `WITH`-specific exception. The current change may still be right, but I think it would help to justify why the narrower segment override is preferred over the broader grammar override.

ricardoalencar-pr:
Thanks @peterbud! I tried the broader `NonWithSelectableGrammar` route and it turns out not to be viable with the current grammar — I pushed a self-contained demo of it here: https://github.com/ricardoalencar-pr/sqlfluff/pull/1

The problem is that `PipeStatementSegment`'s second alternative already begins with `SelectableGrammar`, which contains `NonWithSelectableGrammar`. Adding `PipeStatementSegment` into `NonWithSelectableGrammar` therefore closes a left-recursive loop, and sqlfluff rejects it at dialect-expansion time:

```
RecursionError: Self referential grammar detected:
StatementSegment -> SelectableGrammar -> NonWithSelectableGrammar -> PipeStatementSegment
```

This isn't a soft failure — it breaks the entire BigQuery dialect: even `sqlfluff.parse("SELECT 1", dialect="bigquery")` raises, and all 429 BigQuery dialect tests fail (vs 429 passing with the override in this PR). Making the broad route work would require restructuring `PipeStatementSegment` so it doesn't begin with `SelectableGrammar`, which touches every context a selectable feeds (subqueries, CTAS targets, set-operation operands) — well beyond this bug.

That's also why the existing pipe-in-CTE support (#7180) added `OneOf(Ref("SelectableGrammar"), Ref("PipeStatementSegment"))` at the specific `CTEDefinitionSegment` position rather than in the selectable grammars. This PR follows that same position-specific precedent. Happy to reconsider if you see a way around the recursion.

ricardoalencar-pr:
@peterbud — since you asked about a more general approach, I sketched what an "overall" pipe refactor could look like as an internal RFC draft: https://github.com/ricardoalencar-pr/sqlfluff/pull/2 (not proposing it here yet — just to make the trade-off concrete).

### How it works

Instead of a distinct `PipeStatementSegment` that has to be *added* to each position it's allowed, it models **a pipe query as any base query optionally followed by pipe operators**, and folds that segment *once* into `NonWithSelectableGrammar`:

```python
bigquery_dialect.replace(
    NonWithSelectableGrammar=OneOf(
        Ref("SetExpressionSegment"),
        OptionallyBracketed(Ref("SelectStatementSegment")),
        Ref("NonSetSelectableGrammar"),
        Ref("PipeStatementSegment"),   # <-- folded in here
    ),
)
```

Because `StatementSegment -> SelectableGrammar -> NonWithSelectableGrammar`, pipe queries then work in **every** selectable position through one path: top-level, CTE body, the query after a `WITH` (this bug — now free), and subqueries / set operands.

### Why it doesn't hit the recursion from earlier

The `RecursionError` in the naive version comes from the pipe segment's base being the *full* `SelectableGrammar` (which contains `NonWithSelectableGrammar`), so folding it in closes a first-set cycle. This RFC avoids that with two properties:

1. **Pipe operators only ever *trail* the base query.** A trailing suffix doesn't change a grammar's first-set, so folding `PipeStatementSegment` into `NonWithSelectableGrammar` leaves its first-set as `{FROM, SELECT, VALUES, (}` — unchanged.
2. **The query-initial base is narrowed.** It references `SelectStatementSegment` / `SetExpressionSegment` / `Bracketed(SelectableGrammar)` directly rather than `SelectableGrammar`, so the chain
   ```
   SelectableGrammar -> NonWithSelectableGrammar -> PipeStatementSegment -> SelectStatement / SetExpression / (…)
   ```
   terminates instead of re-entering `SelectableGrammar` unbracketed.

### What it removes vs. this PR

- `PipeStatementSegment` explicitly listed in `StatementSegment`
- the `OneOf(SelectableGrammar, PipeStatementSegment)` in `CTEDefinitionSegment` (#7180)
- any `WithCompoundStatementSegment` override (i.e. this PR)

So one `replace()` + a narrower `PipeStatementSegment` base subsumes three special cases, and also enables pipe-in-subquery.

### Trade-offs

- Larger surface area / higher review cost than the targeted override here.
- Only `pipe_statement.yml` re-nests (the other 428 BigQuery fixtures are untouched, `429 passed`), but reviewers would need to accept the new nesting as canonical.
- Broadening pipe to *all* selectable positions is more permissive than the immediate bug needs, and should be checked against GoogleSQL's grammar before going upstream.

My suggestion would be to keep this PR (#8191) as the low-risk fix for the reported bug, and treat the RFC as a separate "do it properly" discussion if you'd like to generalize. Happy to open that as a real PR here if there's appetite.

peterbud:
Hi @ricardoalencar-pr  - thanks for the detailed analysis.

OK, let's merge this as-is, and then have a separate PR with a proper solution for pipe syntax. I have also checked briefly your exploration branch, and seen that basically you will be able to remove `CTEDefinitionSegment` from the BigQuery syntax, as after removing the existing `PipeStatementSegment` whatg is left basically will be the ANSI `CTEDefinitionSegment`  - so at the end that solution will be a simplifiction vs what we have today.

peterbud (approved):

ricardoalencar-pr:
Will open the refactor as a separate PR after this merges, rebased to drop the `WithCompoundStatementSegment` override this one adds.

On `CTEDefinitionSegment`: confirmed — with `PipeStatementSegment` removed it's identical to the ANSI grammar, so the override goes away entirely, not just trimmed. Dialect tests stay at 429.

PR #8194, opened 2026-07-22
BigQuery: parse EXCEPT/REPLACE after a wildcard on an array element

### LINKED ISSUE #8186: BigQuery parsing error: wildcard on array entry doesn't accept `EXCEPT` argument
### Search before asking

### What Happened

If you can come up with a better name than my title, please replace it hah!



Minimal SQL Example:
```
SELECT
    error,
    results[0].*
EXCEPT(cola)
FROM `project`.`dataset`.`table`
```

This valid BigQuery query returns an error from sqlfluff --fix


### Expected Behaviour

Valid syntax is parsed

### Observed Behaviour

when running `sqlfluff --fix --dialect bigquery temp.sql`

```
$ sqlfluff  fix --dialect bigquery temp.sql
  == [temp.sql] ==
  L:   4 | P:   1 |  PRS | Line 4, Position 1: Found unparsable section:
                       | 'EXCEPT(cola)'
All Finished 📜 🎉!
  [1 unfixable linting violations found]
```


If I remove the except as below, it is parsable
```
SELECT
    error,
    results[0].*
FROM `project`.`dataset`.`table`
```

If I remove the index as below, it is parsable
```
SELECT
    error,
    results.*
EXCEPT(cola)
FROM `project`.`dataset`.`table`
```

### How to reproduce

Simple snippet and no config required

### Dialect

bigquery

### Version

$ sqlfluff --version
sqlfluff, version 4.2.2

$ python --version
Python 3.14.2

### Configuration

empty project

### Code of Conduct

adhavan18:

Fixes #8186

A wildcard on an array element or nested field parses through `SemiStructuredAccessorSegment`, which had no hook for the `EXCEPT`/`REPLACE` modifiers, so valid BigQuery like:

```sql
SELECT
    error,
    results[0].* EXCEPT (cola)
FROM `project`.`dataset`.`table`
```

raised `Found unparsable section: 'EXCEPT(cola)'` (and the same for `REPLACE (...)`). The plain `t.*` path via `WildcardExpressionSegment` already supported both modifiers — the gap was only on the semi-structured accessor path.

This restructures the accessor grammar so the wildcard is a **terminal** element of the chain, carrying optional `ExceptClauseSegment`/`ReplaceClauseSegment` ([BigQuery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_except)). The wildcard lives in its own `SemiStructuredWildcardAccessorSegment`, placed outside the repeating part of `AccessorGrammar` so nothing can be chained after it — `a.b.*.c` and `results[0].*.z` are now rejected. The same applies to accessor chains hanging off a function call (`testFunction(a).b.*.z`) — the wildcard is matched only by `AccessorGrammar`, so terminality is enforced in exactly one place rather than two. As a result the wildcard rows of `select_function_object_fields.yml` gain an enclosing `expression` node; non-wildcard rows are unchanged.

Correction to an earlier version of this description: `a.*.b` is **not** affected by this change and still parses. That form goes through `SingleIdentifierWildcardSegment` + `AccessorGrammar` in the ansi `Expression_D_Grammar`, not the semi-structured path, and behaves the same on `main` today. It looked in scope at first; it isn't, and fixing it would mean touching shared ansi grammar from a BigQuery parse fix.

No. Plain wildcards (`* EXCEPT (...)`, `t.* EXCEPT (...)`), ordinary references (`a.b.c`), array element field access (`results[0].field`) and bare array-element wildcards (`results[0].*`) all parse exactly as before — the full BigQuery dialect suite passes (432 tests) with no changes to any existing fixture tree.

### Pull Request checklist

- Included test cases to demonstrate any code changes:
  - `.sql`/`.yml` parser test case added at `test/fixtures/dialects/bigquery/select_array_element_wildcard_except.sql` (+ generated `.yml`) covering `EXCEPT`, `REPLACE`, a nested-field wildcard and the plain array-element wildcard.
- Added appropriate documentation for the change (inline grammar comment with doc link).
- Created GitHub issues for any relevant followup/future enhancements if appropriate — n/a.

---
## Summary by cubic
Adds support for `EXCEPT`/`REPLACE` after wildcards on array elements and nested fields in BigQuery (e.g., `results[0].* EXCEPT (cola)`), and makes `.*` truly terminal—even after functions—so mid-chain forms like `x.y.*.z` don’t parse.

- **Bug Fixes**
  - Introduces `SemiStructuredWildcardAccessorSegment` outside the repeating `AccessorGrammar` so nothing can follow `.*`, and it can carry `EXCEPT`/`REPLACE`.
  - Stops matching the wildcard inside `FunctionSegment`; `f(a).b.*.z` and `f(a).*.z` now error, while `f(a).b.*` parses.
  - Adds tests for accepted terminal and rejected mid-chain forms; adds a new fixture and updates `select_function_object_fields.yml` for a minor parse-tree change.

<sup>Written for commit 5e35a278407d959f09c59ce69a6163f88fa03c4a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8194?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -1728,21 +1728,40 @@ class SemiStructuredAccessorSegment(BaseSegment):
     """A semi-structured data accessor segment."""
 
     type = "semi_structured_expression"
-    match_grammar = Sequence(
-        AnyNumberOf(
+    # A wildcard is only valid as the final element of the accessor chain,
+    # and it may carry the EXCEPT/REPLACE modifiers, e.g.
+    # `results[0].* EXCEPT (cola)`.
+    # https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_except
+    match_grammar = OneOf(
+        Sequence(
+            AnyNumberOf(
+                Sequence(
+                    Ref("DotSegment"),
+                    Ref("SingleIdentifierGrammar"),
+                    allow_gaps=True,
+                ),
+                Ref("ArrayAccessorSegment", optional=True),
+                allow_gaps=True,
+                min_times=1,
+            ),
             Sequence(
                 Ref("DotSegment"),
-                OneOf(
-                    Ref("SingleIdentifierGrammar"),
-                    Ref("StarSegment"),
-                ),
+                Ref("StarSegment"),
+                Ref("ExceptClauseSegment", optional=True),
+                Ref("ReplaceClauseSegment", optional=True),
                 allow_gaps=True,
+                optional=True,
             ),
-            Ref("ArrayAccessorSegment", optional=True),
             allow_gaps=True,
-            min_times=1,
         ),
-        allow_gaps=True,
+        # A bare `.*` directly on the preceding expression.
+        Sequence(
+            Ref("DotSegment"),
+            Ref("StarSegment"),
+            Ref("ExceptClauseSegment", optional=True),
+            Ref("ReplaceClauseSegment", optional=True),
+            allow_gaps=True,
+        ),
     )
 
 
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -395,11 +395,17 @@ bigquery_dialect.replace(
     NaturalJoinKeywordsGrammar=Nothing(),
     UnconditionalCrossJoinKeywordsGrammar=Ref.keyword("CROSS"),
     MergeIntoLiteralGrammar=Sequence("MERGE", Ref.keyword("INTO", optional=True)),
-    AccessorGrammar=AnyNumberOf(
-        Ref("ArrayAccessorSegment"),
-        Ref("ChainedFunctionCallSegment"),
-        # Add in semi structured expressions
-        Ref("SemiStructuredAccessorSegment"),
+    AccessorGrammar=Sequence(
+        AnyNumberOf(
+            Ref("ArrayAccessorSegment"),
+            Ref("ChainedFunctionCallSegment"),
+            # Add in semi structured expressions
+            Ref("SemiStructuredAccessorSegment"),
+        ),
+        # A wildcard terminates the accessor chain, so it sits outside the
+        # repeating part above - nothing may be chained after it.
+        Ref("SemiStructuredWildcardAccessorSegment", optional=True),
+        allow_gaps=True,
     ),
     BracketedSetExpressionGrammar=Bracketed(Ref("SetExpressionSegment")),
     NotEnforcedGrammar=Sequence("NOT", "ENFORCED"),
@@ -1308,7 +1314,13 @@ class FunctionSegment(ansi.FunctionSegment):
                 # Functions returning STRUCTs in BigQuery can have the fields
                 # elements referenced (e.g. ".a"), including wildcards (e.g. ".*")
                 # or multiple nested fields (e.g. ".a.b", or ".a.b.c")
-                Ref("SemiStructuredAccessorSegment", optional=True),
+                OneOf(
+                    # Try the wildcard form first so that a chain ending in `.*`
+                    # (e.g. `.b.*`) stays a single accessor segment.
+                    Ref("SemiStructuredWildcardAccessorSegment"),
+                    Ref("SemiStructuredAccessorSegment"),
+                    optional=True,
+                ),
                 Ref("PostFunctionGrammar", optional=True),
             ),
         ),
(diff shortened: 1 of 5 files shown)


peterbud (changes requested):
Hi @adhavan18 - thanks for the PR!

This fixes the `EXCEPT`/`REPLACE` gap, but the “terminal star” part doesn’t seem fully enforced yet. In `SemiStructuredAccessorSegment` `.*` is terminal only within a single `SemiStructuredAccessorSegment`, while `AccessorGrammar` can still match another accessor segment immediately after it. On this branch, `SELECT x.y.*.z FROM t` still parses successfully, so the parser still accepts a mid-chain wildcard form the PR description says should now be rejected.

Could you tighten the outer accessor chaining?

adhavan18:
Thanks @peterbud, you're right — I'd only made the star terminal *inside* `SemiStructuredAccessorSegment`, and `AccessorGrammar`'s `AnyNumberOf` could still match another accessor straight after it. Confirmed `SELECT x.y.*.z FROM t` parsed on the previous commit.

Fixed in 8251337: the wildcard is split out into `SemiStructuredWildcardAccessorSegment` and moved *outside* the repeating part of `AccessorGrammar`, so nothing can chain after it:

```python
AccessorGrammar=Sequence(
    AnyNumberOf(
        Ref("ArrayAccessorSegment"),
        Ref("ChainedFunctionCallSegment"),
        Ref("SemiStructuredAccessorSegment"),
    ),
    Ref("SemiStructuredWildcardAccessorSegment", optional=True),
    allow_gaps=True,
),
```

The new segment absorbs any leading fields, so `.b.*` stays a single accessor segment — that keeps `select_function_object_fields.yml` and every other existing fixture unchanged. `FunctionSegment` tries the wildcard form first for the same reason.

Now rejected: `SELECT x.y.*.z`, `SELECT results[0].*.z`
Still parse: `results[0].*`, `results[0].* EXCEPT (cola)`, `results[0].* REPLACE (1 AS cola)`, `s.arr[0].field.* EXCEPT (cola)`, `testFunction(a).b.*`, `f().*`, `a.b.*`, `t.* EXCEPT (a)`

Both sets are covered by new tests in `test/dialects/bigquery_test.py`; the two rejection cases fail on the previous commit and pass on this one. `442 passed` for the BigQuery dialect + fixture suites.

One thing worth flagging, since it's adjacent but I don't think it belongs in this PR: `SELECT a.*.b.c FROM t` still parses. That one doesn't go through the semi-structured path at all — it's `SingleIdentifierWildcardSegment` (`a.*`) followed by `AccessorGrammar` in the ansi `Expression_D_Grammar`, and it parses the same way on `main` today, so it's pre-existing and dialect-agnostic rather than something this change introduced. Happy to look at it separately if you'd like, but it seemed wrong to touch shared ansi grammar in a BigQuery parse fix.

The `.yml` for the new fixture is regenerated — the `array_accessor` now sits as a sibling of `column_reference` rather than nested inside the accessor, which matches how `testFunction(a)[ORDINAL(1)].*` was already represented.

adhavan18:
The bot is right, and thanks — I'd missed the function path. Confirmed on 8251337:

```
SELECT testFunction(a).b.*.z FROM t   -> parsed (should not)
SELECT testFunction(a).*.z FROM t     -> parsed (should not)
SELECT testFunction(a)[0].b.*.z FROM t -> parsed (should not)
```

`FunctionSegment` was matching the wildcard itself, so the star got consumed *inside* the function and the outer `AccessorGrammar` was still free to chain `.z` after it. Making the star terminal in two places doesn't work — whichever one matches it first, the other can still continue.

Fixed in 5e35a27 by leaving the wildcard to `AccessorGrammar` only, so there is a single place that enforces terminality:

```python
# Note the wildcard form is deliberately *not* matched here. It is
# left to `AccessorGrammar`, which is applied after the function and
# is the single place that enforces the wildcard being terminal.
Ref("SemiStructuredAccessorSegment", optional=True),
```

All three now fail to parse, and `testFunction(a).b.*`, `testFunction(a).*`, `testFunction(a).b.c` still parse. The three cases are added to `test_semi_structured_wildcard_is_terminal`; they fail on 8251337 and pass on 5e35a27.

One consequence worth calling out explicitly, since it touches an existing fixture: `select_function_object_fields.yml` changes for the **wildcard** rows only. Because the star is now matched by the outer accessor chain rather than inside the function, those rows gain an enclosing `expression` node, with `semi_structured_expression` as a sibling of `function`:

```yaml
 select_clause_element:
-  function:
-    ...
+  expression:
+    function:
+      ...
     semi_structured_expression:
       dot: .
       star: '*'
```

The non-wildcard rows (`testFunction(a).b`, `testFunction(a).b.c`, `testFunction(a)[ORDINAL(1)].a`) are untouched. I couldn't see a way to both keep the star inside `FunctionSegment` and have it be genuinely terminal, so I went with the single-enforcement-point version — happy to revisit if you'd prefer the tree shape preserved and the function path left as a known gap instead.

Suite results: `452 passed` for the BigQuery dialect + fixture tests, and `3335 passed` across `test/core` + `test/rules` (3 failures in `plugin_test.py`, which fail identically on a clean checkout here — the example plugin isn't installed in my venv).

peterbud (approved):
Thanks @adhavan18  - looks good now!

PR #8198, opened 2026-07-23
Add a Python-vs-Rust parser/lexer parity test harness

rubytobi:
As the Rust parser/lexer move toward being the default engine, we need automated coverage that catches any behavioural divergence from the pure-Python engine before it reaches users, rather than relying on manual comparison or bug reports.

This adds a differential test harness, plus initial divergences identified with help of Claude, that runs the same SQL through the pure-Python engine and the Rust engine (`RustParser`/`PyRsLexer`) and asserts the results are identical. This is primarily test infrastructure, no parsing or lexing behaviour changes. However, adding a `get_native_ast` function.

New pieces:

- **`test/core/parser/parity/compare.py` + `cases_test.py`**: shared capture/compare logic and drivers for a data-driven fixture corpus (`test/fixtures/parity/*.yml`), covering four legs. Every comparison captures very strictly: tree tuple form with position markers, `stringify()` text, normalization kwargs, full `class_types` sets, and exception type/message/position (including Rust panics).
  - `python_vs_rust` (pure-Python `Parser` vs `RustParser`'s legacy path, Python as reference),
  - `native_vs_legacy` (RustParser's fused native-AST path vs its legacy path; any difference here is always a bug, never an accepted "known divergence"),
  - `lexer` (`PyLexer` vs `PyRsLexer`), and
  - `invariants` (structural well-formedness of the raw `RsMatchResult` tree).

  Of these, `lexer` is wired up but has no fixture cases yet (no `*.yml` declares `_meta: {kind: lexer}`), so it currently collects as a single empty/skipped parametrization; a follow-up will seed it the way this PR seeded `invariants`.

- **`test/core/parser/parity/corpus_test.py`**: sweeps every fixture under `test/fixtures/dialects/*/*.sql` through all three tree-building paths at the same strictness, so parity is exercised over the whole existing dialect corpus, not just hand-written cases.

- **`test/fixtures/parity/regressions.yml` + `README.md`**: the initial fixture corpus and format docs. A known, still-open divergence is declared per-leg via a **strict** `xfail`, so CI flags it as stale the moment the gap is fixed. This first batch documents 25 currently-known divergences; closing them is split across 5 focused follow-up PRs.

- **`test/fixtures/parity/invariants_cases.yml`**: gives the `invariants` leg its real cases, so it actually runs against real match trees. Currently no known issues but providing two test cases to ensure most obvious regressions don't happen.

- Removes the now-duplicate `test__rust_parser__native_ast_parity` from `rust_parser_test.py`, superseded by the stricter `corpus_test.py` sweep.

- One production-code addition: `rust_parser.py` gains a `get_native_ast()` getter (paired with the existing setter) so the harness can save/restore the native-AST flag per leg; no behaviour change.

- Test suite runtime increases: the corpus sweep parses all dialect fixtures through three engine paths at full strictness, and `regressions.yml` currently carries strict `xfail` cases documenting open Python/Rust divergences, plus further strict xfails in `corpus_test.py` for the single already-diagnosed trim_chars bug across MySQL/MariaDB/BigQuery session-variable fixtures.

- The `lexer` leg exists in the harness but has zero cases today; it doesn't fail, it silently collects nothing. Flagging so it isn't mistaken for coverage that's already in place.

- No behavioural changes to the lexer or parser themselves; the only non-test change is the `get_native_ast()` getter described above.

### Pull Request checklist

  - [ ] `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
  - [x] Other: new `test/core/parser/parity/` suite plus `test/fixtures/parity/*.yml` fixtures (see above).
- [x] Added appropriate documentation for the change: `test/fixtures/parity/README.md` documents the fixture format and how to add a new parity regression.

---
## Summary by cubic
Add a strict Python-vs-Rust parity test harness for the parser and lexer, plus a whole-corpus three-way sweep. This detects divergences early and documents known gaps with strict xfails.

- **New Features**
  - Differential harness with four legs: `python_vs_rust`, `native_vs_legacy`, `lexer`, and `invariants`. Captures tree/positions/stringify/raw/normalization kwargs/class_types and errors for exact matching.
  - Corpus-wide sweep that parses every dialect fixture through `Parser`, `RustParser` (legacy), and `RustParser` native-AST, asserting identical results.
  - New fixture corpus under `test/fixtures/parity/*.yml` with strict per-leg `xfail` and a README for the format.
  - Invariants leg seeded with cases to check raw `RsMatchResult` structure.
  - Added `get_native_ast()` to `rust_parser.py` so tests can save/restore the native-AST flag.

- **Refactors**
  - Removed the weaker duplicate three-way parity test from `rust_parser_test.py`.
  - Test runtime increases due to the corpus sweep; `lexer` leg is wired but has no cases yet.
  - Dropped a stale reference to `grammar_test.py`.

<sup>Written for commit 137e5eae7a17ebe04918a5c1eb3e13164325244a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8198?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/core/parser/rust_parser.py
+++ b/src/sqlfluff/core/parser/rust_parser.py
@@ -104,6 +104,11 @@ def set_native_ast(enabled: bool) -> None:
     _NATIVE_AST_ENABLED = enabled
 
 
+def get_native_ast() -> bool:
+    """Return whether the fused (native) BaseSegment builder is enabled."""
+    return _NATIVE_AST_ENABLED
+
+
 try:
     from sqlfluffrs import RsMatchResult, RsParseError, RsParser, RsToken
 
--- a/test/core/parser/parity/__init__.py
+++ b/test/core/parser/parity/__init__.py
@@ -1,0 +1,1 @@
+"""Python-vs-Rust engine parity suite."""
--- a/test/core/parser/parity/cases_test.py
+++ b/test/core/parser/parity/cases_test.py
@@ -1,0 +1,97 @@
+"""Drivers for the parity case corpus in ``test/fixtures/parity/``.
+
+Each YAML case is just data: SQL (inline or from a dialect-corpus fixture),
+dialect, config and metadata. ``compare.py`` holds the shared capture and
+comparison logic that every driver uses, so every case is checked at the
+same strict level. Known differences are recorded in the fixture as a
+strict xfail for the specific leg they affect (python_vs_rust or
+native_vs_legacy).
+"""
+
+import pytest
+
+from .compare import (
+    PARSER_LEGS,
+    build_config,
+    check_expectations,
+    lex_capture,
+    linted_parse_capture,
+    load_case_params,
+    load_reference_case_params,
+    parse_capture,
+    raw_match_violations,
+    reference_capture,
+    requires_rust_parser,
+    resolve_case_sql,
+)
+
+
+@pytest.mark.parametrize("case", load_reference_case_params("parser"))
+def test__parity__parser_case_reference_expectations(case):
+    """Confirm the Python parser still behaves as this case's ``expect`` claims.
+
+    This is a regression guard on the pinned case itself, not an engine
+    comparison: it catches the Python parser's behavior changing on this SQL
+    (or ``expect`` having been wrong all along). It runs as its own test, so
+    it still catches that even when the case is marked as a known engine
+    divergence (xfail) in the comparison test below.
+    """
+    check_expectations(case, reference_capture(case))
+
+
+@requires_rust_parser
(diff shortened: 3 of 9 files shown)


keraion (approved):
LGTM

PR #8201, opened 2026-07-23
fix(ST05): avoid generated CTE name capture

dev-willbird1936:

ST05 can generate a CTE name that captures an existing physical-table reference. This change:

- accounts for SQLite forward CTE visibility and nested CTE shadowing;
- reserves visible physical-table names during generated-name selection;
- caches normalized table-reference groups once per parsed statement, avoiding a recursive full-statement walk for every offending subquery;
- adds focused regression fixtures for capture, declaration order, and nested shadowing.

Reproduction:

```sql
WITH cte_a AS (
    SELECT anchor.v
    FROM (SELECT 1 AS v) AS anchor
    JOIN (SELECT 2 AS v) AS victim ON 1 = 1
),
cte_b AS (SELECT v FROM victim)
SELECT cte_a.v, cte_b.v
FROM cte_a JOIN cte_b ON 1 = 1;
```

On `main`, ST05 can generate a CTE named `victim`, causing the later unqualified reference to resolve to the generated CTE instead of the physical table.

The cache is private to one ST05 evaluation and keyed by the parsed statement and dialect. The parsed statement is immutable during evaluation; generated CTEs reuse subqueries already present in that tree. There are no configuration or public API changes.

### Validation

- `uv run pytest test/rules/yaml_test_cases_test.py -k ST05 -q`: 50 passed.
- `uv run pytest test/rules/std_ST05_LT08_test.py test/rules/std_ST05_LT09_test.py -q`: 3 passed.
- `uv run pytest test/rules -q`: 2,471 passed, 101 skipped.
- Ruff format and check: passed.
- MyPy for `ST05.py`: passed.
- `git diff --check`: passed.

Repository CI has not been scheduled for this fork update; maintainer review remains pending.

### Pull Request checklist

- [x] Included YAML rule test cases that demonstrate the bug and fixed output.
- [ ] Added documentation (not required; no user-facing behavior or configuration is added).
- [ ] Created follow-up issues (none are required).

### AI-assisted contribution disclosure

Codex assisted with diagnosis, implementation, review-response changes, and test execution. I reviewed the final diff, authorized the submission, and remain responsible for correctness and maintainability. The validation listed above was run locally against this commit.

### DIFF
--- a/src/sqlfluff/rules/structure/ST05.py
+++ b/src/sqlfluff/rules/structure/ST05.py
@@ -2,6 +2,7 @@
 
 from collections.abc import Iterator
 from functools import partial
+from itertools import chain
 from typing import NamedTuple, Optional, TypeVar, cast
 
 from sqlfluff.core.dialects.base import Dialect
@@ -9,6 +10,7 @@ from sqlfluff.core.dialects.common import (
     AliasInfo,
     ObjectReferenceLevel,
     extract_possible_references,
+    iter_raw_references,
 )
 from sqlfluff.core.parser import (
     BaseSegment,
@@ -53,6 +55,17 @@ class _NestedSubQuerySummary(NamedTuple):
     selectable: Selectable
     table_alias: AliasInfo
     select_source_names: set[str]
+
+
+_NormalizedTableReference = tuple[BaseSegment, str]
+
+
+class _VisibleTableReferenceIndex(NamedTuple):
+    """Cached table references grouped by their top-level visibility scope."""
+
+    ctes: tuple[CTEDefinitionSegment, ...]
+    cte_references: tuple[tuple[_NormalizedTableReference, ...], ...]
+    root_references: tuple[_NormalizedTableReference, ...]
 
 
 class Rule_ST05(BaseRule):
@@ -180,6 +193,7 @@ class Rule_ST05(BaseRule):
             ctes=ctes,
             case_preference=case_preference,
             clone_map=clone_map,
+            root_segment=context.segment,
         )
 
         results_list: list[tuple[LintResult, BaseSegment, str, BaseSegment, bool]] = []
@@ -291,16 +305,13 @@ class Rule_ST05(BaseRule):
         ctes: "_CTEBuilder",
         case_preference: str,
         clone_map,
+        root_segment: BaseSegment,
     ) -> Iterator[tuple[LintResult, BaseSegment, str, BaseSegment, bool]]:
         """Given the root query, compute lint warnings."""
         nsq: _NestedSubQuerySummary
         for nsq in self._nested_subqueries(query, dialect):
-            alias_name, _ = ctes.create_cte_alias(nsq.table_alias)
             # 'anchor' is the TableExpressionSegment we fix/replace w/CTE name.
             anchor = nsq.table_alias.from_expression_element.segments[0]
-            # If we have duplicate CTE names just don't fix anything
-            # Return the lint warnings anyway
-            is_fixable = alias_name not in ctes.list_used_names()
 
             # if the subquery is table_expression, get the bracketed child instead.
             if anchor.is_type("table_expression"):
@@ -312,10 +323,32 @@ class Rule_ST05(BaseRule):
                 bracket_anchor = anchor
 
             # we can't create a CTE from a nested subquery here, ignore it.
-            if not bracket_anchor.is_type("bracketed") or bracket_anchor.get_child(
-                "table_expression"
-            ):
-                is_fixable = False
+            structurally_fixable = bracket_anchor.is_type(
+                "bracketed"
+            ) and not bracket_anchor.get_child("table_expression")
+            visible_table_names = (
+                ctes.list_visible_table_names(
+                    parent_selectable=nsq.selectable.selectable,
+                    extracted_subquery=bracket_anchor,
+                    root_segment=root_segment,
+                    dialect_name=dialect.name,
+                )
+                if structurally_fixable
+                else set()
+            )
+            alias_name, _ = ctes.create_cte_alias(
+                nsq.table_alias,
+                dialect=dialect,
+                reserved_names=visible_table_names,
+            )
+            # If we have duplicate CTE names just don't fix anything.
+            # Return the lint warnings anyway.
+            is_fixable = (
+                structurally_fixable
+                and alias_name not in ctes.list_used_names()
+                and _normalize_generated_identifier(alias_name, dialect)
+                not in visible_table_names
+            )
             if is_fixable:
                 new_cte = _create_cte_seg(  # 'prep_1 as (select ...)'
                     alias_name=alias_name,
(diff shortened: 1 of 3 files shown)


earfman:
Ran this branch's own scenario across every dialect and it doesn't hold in `bigquery` — the generated name is still `prep_1` there, byte-identical to the merge-base.

Repro, no config beyond the dialect:

```sql
SELECT b.y
FROM prep_1 AS b
JOIN (SELECT 0 AS y) ON b.y >= 0
```

`sqlfluff fix --rules ST05`:

- `--dialect postgres`, this branch: `WITH prep_2 AS (SELECT 0 AS y) SELECT b.y FROM prep_1 AS b JOIN prep_2 ON b.y >= 0`
- `--dialect bigquery`, this branch: `WITH prep_1 AS (SELECT 0 AS y) SELECT b.y FROM prep_1 AS b JOIN prep_1 ON b.y >= 0`

In the BigQuery output the alias `b` silently rebinds from the physical table to the generated CTE. Executed against a real engine (SQLite 3.45.1 standing in, since its CTE scoping matches here; physical `prep_1(y)` holding 10, 20, 30):

- before the fix: `[(10,), (20,), (30,)]`
- after the fix: `[(0,)]`

No error, different rows.

Sweeping all 28 dialects on that input: 27 produce a fix, and 26 of those correctly pick `prep_2`. `bigquery` is the only one that still emits `prep_1`. (`materialize` doesn't produce a fix for this input at all.)

The cause is that the two sides of the reservation comparison use different normalizers:

- `_normalize_table_reference` → `segment.raw_normalized(casefold=True)`
- `_normalize_generated_identifier` → `dialect.ref("NakedIdentifierSegment").casefold`

Those agree in every dialect except BigQuery, which declares `casefold = str.upper` on the grammar but keeps `raw_normalized` case-preserving:

| dialect   | `NakedIdentifierSegment.casefold` | `raw_normalized(casefold=True)` of `prep_1` |
|-----------|-----------------------------------|---------------------------------------------|
| bigquery  | `str.upper`                       | `'prep_1'`                                  |
| snowflake | `str.upper`                       | `'PREP_1'`                                  |
| tsql      | `str.upper`                       | `'PREP_1'`                                  |
| postgres  | `str.lower`                       | `'prep_1'`                                  |

So on BigQuery `visible_table_names` is `{'prep_1'}` while `_normalize_generated_identifier('prep_1', dialect)` is `'PREP_1'`; the membership test in `create_cte_alias` misses and `is_fixable` stays `True`. Instrumented on this branch:

```
bigquery: visible={'prep_1'}  alias=('prep_1', True)  reserved={'prep_1'}  normgen='PREP_1'
postgres: visible={'prep_1'}  alias=('prep_2', True)  reserved={'prep_1'}  normgen='prep_2'
```

Why the suite doesn't catch it: the new ST05 fixtures are ansi, postgres and sqlite only, so nothing exercises the one dialect where the two normalizers disagree. `test/rules` is 2471 passed / 101 skipped on this branch, matching the numbers in the description.

Normalizing the generated name through the same `raw_normalized` path as the references, rather than the grammar's `casefold` attribute, looks like it would close it — though you'd know better whether `raw_normalized` is the right contract to lean on here. Happy to put up a fixture plus patch if that's useful.

earfman:
Following up on my BigQuery report from the 24th: **it's fixed, as of `5eaa1d9`.** Re-ran the same repro against each commit on the branch, `sqlfluff fix --rules ST05 --dialect bigquery`:

| commit | output |
| --- | --- |
| `783896d` (head when I reported) | `WITH prep_1 AS ...` — the capture |
| `5eaa1d9` | `WITH prep_2 AS ...` |
| `4236c60` (current head) | `WITH prep_2 AS ...` |

Against a real engine (sqlite3 3.45.1, physical `prep_1(y)` holding 10, 20, 30): the original returns `[(10,), (20,), (30,)]`, the `783896d` output returned `[(0,)]`, and current head returns `[(10,), (20,), (30,)]` again. Swept all 28 dialects at head — every one emits `prep_2`, none regressed, none newly fails to fix. Worth noting for anyone reading the thread cold that on the merge base all 28 emit `prep_1`, so the capture is pre-existing and universal; this PR fixes it everywhere.

`_identifier_is_reserved()` resolves it by testing `{raw, normalized}` against the reserved set rather than reconciling the two normalizers. That works. For the record the mismatch really is BigQuery-only — I compared `NakedIdentifierSegment.casefold` against a parsed table reference's `raw_normalized(casefold=True)` across all 28 dialects and BigQuery is the only one where they disagree (`PREP_1` vs `prep_1`).

## One addition to the T-SQL point

The automated reviews are right that `dialect_name == "sqlite"` at ST05.py:498 is too narrow, and I can confirm it at head — `SELECT * FROM base JOIN (SELECT v FROM victim) AS victim ON 1=1` in `tsql` still emits:

```sql
WITH victim AS (SELECT v FROM victim) SELECT * FROM base JOIN victim ON 1 = 1
```

**`oracle` has the same problem and hasn't been mentioned.** Oracle's recursive subquery factoring uses plain `WITH` with no `RECURSIVE` keyword, exactly like T-SQL, so a CTE naming a table in its own body is a self-reference there too. I confirmed the hardcoded string is the sole gate: patching it to `dialect_name in ("sqlite", "tsql", "oracle")` makes both decline to fix, and reverting restores the broken output. The existing `generated_cte_may_reuse_postgres_body_table_name` fixture encodes the assumption that non-SQLite dialects are safe, which is right for Postgres (the standard requires `WITH RECURSIVE`) but not for these two.

Stating the limit of that plainly: I proved the *code generation* by execution, but there's no SQL Server or Oracle engine available to me, so unlike the BigQuery and SQLite cases the semantic breakage on those two is reasoned from documented behaviour rather than a query I ran. Treat it accordingly.

## The new fixtures hold up

I mutation-tested them, since a regression test that can't fail isn't worth much:

| mutation | result |
| --- | --- |
| full revert of ST05.py, fixtures kept | 6 failed |
| union → normalized-only (undo the BigQuery fix) | exactly 1 failed — the BigQuery fixture |
| remove the `sqlite` guard | exactly 1 failed — the SQLite fixture |

Each mutation is caught by precisely the fixture that should catch it, and the four `may_reuse_*` cases surviving a full revert is correct, since those are negative cases guarding against over-reservation. `test/rules/` gives byte-identical failure lists on both sides (18, all pre-existing Jinja/environmental), with the passing count going 2447 → 2457 for the ten new fixtures.

Minor, not a defect: BigQuery table names are case-sensitive, so the union check over-reserves — a physical `PREP_1` pushes the generated name to `prep_2` even though `prep_1` wouldn't have collided. That errs safe.

dev-willbird1936:
Thanks for retesting this across every dialect and confirming the BigQuery behavior is fixed. The detailed before-and-after verification was very helpful.

peterbud:
Apparently the bots area having a lot of fun with this.

dev-willbird1936:
> Apparently the bots area having a lot of fun with this.

is this a fault on my end?

PR #8202, opened 2026-07-23
fix(ST02): preserve indexed expressions in NULL simplification

dev-willbird1936:
## Summary

- Preserve an array accessor when ST02 reduces a NULL-only `CASE` expression to its matched expression.
- Pass the detected accessor through the column-only replacement path.
- Add a Snowflake regression fixture that asserts the exact fixed SQL.

## Reproduction

```sql
SELECT CASE
    WHEN genres[0] IS NULL THEN NULL
    ELSE genres[0]
END;
```

On `main`, the autofix can produce `SELECT genres`, dropping `[0]`. This change produces `SELECT genres[0]`.

## Testing

- `pytest -q test/rules/yaml_test_cases_test.py -k ST02`: 29 passed
- YAML lint: passed
- `git diff --check`: passed

## AI assistance

---
## Summary by cubic
Fixes `ST02` so NULL-only CASE simplification keeps full indexed expressions (e.g., `genres[0]`) and also handles the `IS NOT NULL` no-ELSE form. Adds dialect-aware, case-sensitive matching and Snowflake fixtures to lock the exact output and avoid unsafe fixes.

- **Bug Fixes**
  - Replace CASE with the complete matched expression to preserve `[...]`, stacked indexes, and qualified names.
  - Use dialect-aware operand identity (respects quoted identifiers and casing) to match THEN/ELSE; supports `IS NOT NULL` with no ELSE.
  - Skip simplification for composite conditions or dynamic index expressions; leave bare THEN columns unchanged.

<sup>Written for commit d93def8b7692ef5326eb07fa3afc4469581ab358. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8202?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/rules/structure/ST02.py
+++ b/src/sqlfluff/rules/structure/ST02.py
@@ -143,13 +143,13 @@ class Rule_ST02(BaseRule):
     @staticmethod
     def _column_only_fix_list(
         context: RuleContext,
-        column_reference_segment: BaseSegment,
+        replacement_segment: BaseSegment,
     ) -> list[LintFix]:
-        """Generate list of fixes to reduce CASE statement to a single column."""
+        """Generate fixes to reduce CASE to the complete matched expression."""
         fixes = [
             LintFix.replace(
                 context.segment,
-                [column_reference_segment],
+                [replacement_segment],
             )
         ]
         return fixes
@@ -275,7 +275,7 @@ class Rule_ST02(BaseRule):
                             anchor=condition_expression,
                             fixes=self._column_only_fix_list(
                                 context,
-                                column_reference_segment,
+                                coalesce_arg_1,
                             ),
                             description="Unnecessary CASE statement. "
                             f"Just use column '{column_reference_segment.raw}'.",
@@ -299,7 +299,7 @@ class Rule_ST02(BaseRule):
                         anchor=condition_expression,
                         fixes=self._column_only_fix_list(
                             context,
-                            column_reference_segment,
+                            then_expression,
                         ),
                         description="Unnecessary CASE statement. "
                         f"Just use column '{column_reference_segment.raw}'.",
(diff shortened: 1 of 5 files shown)


peterbud (changes requested):
Hi @dev-willbird1936 there’s still a real gap here in the indexed `IS NOT NULL ... ELSE NULL` path.

The change correctly preserves indexed expressions when ST02 simplifies directly, and it also avoids the earlier concern about incorrectly appending an accessor in the no-`ELSE` case. But in the full fixer flow, `CASE WHEN genres[0] IS NOT NULL THEN genres[0] ELSE NULL END` still only reduces to `CASE WHEN genres[0] IS NOT NULL THEN genres[0] END` rather than all the way to `genres[0]`. That happens because the no-`ELSE` branch still matches only the bare column reference, so once ST01 removes `ELSE NULL`, ST02 no longer recognizes the indexed expression.

Also we'd need regression test covering the indexed `IS NOT NULL` variant, since the current added tests only cover the `IS NULL` side and miss this remaining case.

peterbud:
Hi @dev-willbird1936 - care to summarize your reply to my review?

dev-willbird1936:
Yea @peterbud  no problem, all addressed in the latest commits. f96d78f fixes the indexed IS NOT NULL no-ELSE path and adds the regression coverage. I also added follow-up safeguards for the related operand and identifier edge cases. The latest commit is d93def8; please take a look when you have a chance.

peterbud (approved):
Thanks!

PR #8203, opened 2026-07-23
fix(CV12): preserve SQLite outer join semantics

dev-willbird1936:
## Summary

- Keep CV12 diagnostics for SQLite outer joins but suppress predicate-relocation fixes.
- Cover `LEFT`, `RIGHT`, `FULL`, and explicit `OUTER` joins.
- Add an executable SQLite regression fixture.

## Reproduction

```sql
SELECT a.k, b.k
FROM a LEFT JOIN b
WHERE a.k = b.k;
```

With `a(k)` containing `1, 2` and `b(k)` containing `1`, the original returns one row. Moving the predicate into `ON` also returns the null-extended `[2, NULL]` row, changing the result set.

## Testing

- `pytest -q test/rules/yaml_test_cases_test.py -k CV12`: 27 passed
- YAML lint: passed
- `git diff --check`: passed

## AI assistance

---
## Summary by cubic
Preserves outer join semantics in CV12 by keeping diagnostics but suppressing the autofix that moves WHERE predicates into ON for outer joins (LEFT/RIGHT/FULL/OUTER). Inner, `SEMI`, and `ANTI` joins still auto-fix.

- **Bug Fixes**
  - Tests: cover SQLite and ANSI outer joins (no fix), NATURAL/CROSS with later outer joins (no fix), templated inner-join guard and templated+literal join mix, SparkSQL `LEFT SEMI` and `LEFT ANTI` joins still auto-fix, preserve existing `ON` conditions, and update AL05 to use an inner join.

- **Dependencies**
  - Synced with `main` and updated Ruff config; pin `ruff` to `0.15.5`.

<sup>Written for commit 809862dfb024a423a07b0eed8617f1632a17d4eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8203?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -2,7 +2,7 @@
 
 # code linting and formatting
 # ---------------------------
-ruff
+ruff==0.15.5
 import-linter
 yamllint
 
--- a/src/sqlfluff/rules/convention/CV12.py
+++ b/src/sqlfluff/rules/convention/CV12.py
@@ -140,6 +140,16 @@ class Rule_CV12(BaseRule):
                 # Join condition is present, no error reported.
                 continue
 
+            # Moving a WHERE predicate into ON changes which null-extended
+            # rows survive an outer join. SQLite makes this directly visible
+            # for LEFT JOIN, so retain the diagnostic but suppress the fix.
+            if context.dialect.name == "sqlite" and any(
+                kw.raw_upper in ("LEFT", "RIGHT", "FULL", "OUTER")
+                for kw in join_clause_keywords
+            ):
+                yield LintResult(anchor=join_clause)
+                continue
+
             if not where_clause_simplifable:
                 yield LintResult(anchor=join_clause)
             else:
--- a/src/sqlfluff/rules/convention/CV12.py
+++ b/src/sqlfluff/rules/convention/CV12.py
@@ -141,12 +141,17 @@ class Rule_CV12(BaseRule):
                 continue
 
             # Moving a WHERE predicate into ON changes which null-extended
-            # rows survive an outer join. SQLite makes this directly visible
-            # for LEFT JOIN, so retain the diagnostic but suppress the fix.
-            if context.dialect.name == "sqlite" and any(
+            # rows survive an outer join, so retain the diagnostic but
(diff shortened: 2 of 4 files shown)


peterbud (changes requested):
Hi @dev-willbird1936 , the semantic problem here is not SQLite-specific. Moving a predicate from `WHERE` into `ON` changes the behavior of `LEFT`/`RIGHT`/`FULL` joins in general, not just in SQLite. SQLite may be the easiest place to demonstrate it with executable SQL, but SQLFluff’s parser accepts this pattern for other dialects too, and CV12 still auto-fixes it there.

So the main question is: why should fix suppression be limited to `sqlite`? If the concern is “don’t apply an autofix that can change outer join semantics”, that seems like it should apply to all outer joins, not only SQLite.

Can you elaborate / review?

Also linting in CI is failing which should be fixed

dev-willbird1936:
Hi @peterbud 

You’re right. SQLite was just the easiest way to show the problem, not the reason it exists. Moving a predicate from WHERE to ON can change the results of LEFT, RIGHT, and FULL joins in any dialect.
I’ve updated CV12 to keep reporting these cases but skip the autofix for outer joins. I also kept SEMI and ANTI joins fixable. The tests now cover the ANSI join forms as well as SQLite.
I fixed the CI issue by pinning Ruff to 0.15.5, matching pre-commit. Ruff 0.16.1 reports 1,728 unrelated findings across the repository, so I left those out of this PR.

peterbud:
Your branch is in conflict with the main branch now. I think also your changes to  `requirements_dev.txt` are also  not correct with the pin, and rather it should look at the `pyproject.toml` to see which rule set to include/exclude

peterbud (approved):
Thanks @dev-willbird1936 !

PR #8206, opened 2026-07-23
fix(AL05): preserve typed PostgreSQL function aliases

dev-willbird1936:
## Summary

- Preserve bracketed typed table-function aliases as well.
- Preserve PostgreSQL table-function aliases that contain a typed column-definition list.
- Detect the `data_type` within the alias expression, without exempting ordinary unused aliases.
- Add a PostgreSQL regression fixture.

## Reproduction

```sql
SELECT a
FROM json_to_record('{"a": 1}') AS r(a int);
```

On `main`, AL05 can remove `AS r(a int)` because `r` is not referenced. That also removes the type information required by PostgreSQL for this record-returning function.

## Testing

- `pytest -q test/rules/yaml_test_cases_test.py -k AL05`: 142 passed
- YAML lint: passed
- `git diff --check`: passed

## AI assistance

---
## Summary by cubic
Fix AL05 to preserve typed table-function aliases in `postgres` and `greenplum`, including bracketed alias expressions, so column types are kept and queries don’t error.

- **Bug Fixes**
  - In `postgres` and `greenplum`, treat an alias as required when its alias expression includes a typed column list (`data_type`), even if wrapped in brackets.
  - Ensure functions like `json_to_record` keep typed aliases even when the alias name isn’t referenced.
  - Add regression tests in `AL05.yml` for both dialects, plus a bracketed alias case in `postgres`.

<sup>Written for commit 8d277d74f7ce7968e74ff679c2a720b97b9b0a09. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8206?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/rules/aliasing/AL05.py
+++ b/src/sqlfluff/rules/aliasing/AL05.py
@@ -248,6 +248,21 @@ class Rule_AL05(BaseRule):
         * In the case of a nested SELECT, all dialect checked (MySQL, Postgres,
           T-SQL) require an alias.
         """
+        if dialect_name in ("postgres", "greenplum"):
+            alias_expressions = from_expression_element.get_children(
+                "alias_expression", "bracketed"
+            )
+            for alias_expression in alias_expressions:
+                if alias_expression.is_type("bracketed"):
+                    alias_expression = alias_expression.get_child("alias_expression")
+                if alias_expression and any(
+                    alias_expression.recursive_crawl("data_type")
+                ):
+                    # PostgreSQL record-returning table functions require the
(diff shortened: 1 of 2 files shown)


rubytobi:
Closing due to inactivity.

PR #8212, opened 2026-07-24
Fix Databricks materialized view expectation constraints

### LINKED ISSUE #8196: Databricks: CREATE MATERIALIZED VIEW column CONSTRAINT ... EXPECT (...) ON VIOLATION ... unparsable
### What Happened
`TableConstraintSegment` for the databricks dialect only supports PRIMARY KEY / FOREIGN KEY / CHECK. It doesn't support Lakeflow Declarative Pipelines' `EXPECT` data-quality constraint syntax, so any `CREATE MATERIALIZED VIEW` using it is unparsable.

### Expected Behaviour
Should parse. This is valid, documented syntax:
https://docs.databricks.com/aws/en/ldp/developer/ldp-sql-ref-create-materialized-view

### How to reproduce
```sql
CREATE MATERIALIZED VIEW dim_country (
    country_id INT,
    CONSTRAINT country_id_not_null EXPECT (country_id IS NOT NULL) ON VIOLATION DROP ROW
)
AS
SELECT 1 AS country_id;
```

### Dialect
databricks

### Version
sqlfluff 4.2.2

vetrovk:

Fixes #8196.

Databricks materialized views support expectation constraints in the column list:

- `CONSTRAINT <name> EXPECT (...)`
- optional `ON VIOLATION FAIL UPDATE`
- optional `ON VIOLATION DROP ROW`

This adds a dedicated materialized-view expectation constraint segment and wires it only into `CREATE MATERIALIZED VIEW`. The shared table-constraint grammar remains unchanged, so the patch does not broaden this syntax into unrelated constraint contexts.

None expected. Incomplete `ON VIOLATION` forms remain unparsable.

### Pull Request checklist

  - [x] `.sql`/`.yml` parser test cases in `test/fixtures/dialects`.
- Added appropriate documentation for the change. Existing Databricks documentation already describes this syntax; no SQLFluff documentation change is needed.
- Created GitHub issues for any relevant followup/future enhancements if appropriate. Not needed.

### AI assistance

AI-assisted tooling was used during implementation and validation; I reviewed and verified the final change.

Validation:
- `.venv/bin/python test/generate_parse_fixture_yml.py -d databricks`
- `.venv/bin/python -m pytest test/dialects/dialects_test.py -k "databricks and create_materialized_view"`
- `.venv/bin/python -m pytest test/dialects/dialects_test.py -k databricks`
- `.venv/bin/ruff check src/sqlfluff/dialects/dialect_databricks.py`
- `.venv/bin/ruff format --check src/sqlfluff/dialects/dialect_databricks.py`
- `git diff --check`

### DIFF
--- a/src/sqlfluff/dialects/dialect_databricks.py
+++ b/src/sqlfluff/dialects/dialect_databricks.py
@@ -948,6 +948,28 @@ class CreateViewStatementSegment(BaseSegment):
     )
 
 
+class MaterializedViewExpectationConstraintSegment(BaseSegment):
+    """A data quality expectation for a materialized view."""
+
+    type = "constraint_statement"
+
+    match_grammar = Sequence(
+        "CONSTRAINT",
+        Ref("ObjectReferenceSegment"),
+        "EXPECT",
+        Bracketed(Ref("ExpressionSegment")),
+        Sequence(
+            "ON",
+            "VIOLATION",
+            OneOf(
+                Sequence("FAIL", "UPDATE"),
+                Sequence("DROP", "ROW"),
+            ),
+            optional=True,
+        ),
+    )
+
+
 class CreateMaterializedViewStatementSegment(BaseSegment):
     """A `CREATE MATERIALIZED VIEW` Statement.
 
@@ -1026,6 +1048,7 @@ class CreateMaterializedViewStatementSegment(BaseSegment):
             Delimited(
                 OneOf(
                     Ref("ColumnFieldDefinitionSegment"),
+                    Ref("MaterializedViewExpectationConstraintSegment"),
                     Ref("TableConstraintSegment"),
                 ),
             ),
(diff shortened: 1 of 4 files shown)


peterbud (changes requested):
HI @vetrovk - thanks for this PR as well!

I'm concerned, that this PR makes the materialized-view column list a flat `Delimited(OneOf(...))` so the new expectation constraint can appear anywhere in the list. Databricks documents the order as column definitions first, then optional expectation constraints, then optional table constraints.

On this branch, `CREATE MATERIALIZED VIEW bad_mv (CONSTRAINT c EXPECT (value > 0), value INT) AS SELECT 1 AS value;` parses successfully, and so does an expectation placed after a table constraint.

That is new over-acceptance introduced by the added `MaterializedViewExpectationConstraintSegment` and it should be modeled with an ordered sequence rather than an unconstrained alternation.

vetrovk:
Thanks for the careful review! You’re right, the previous grammar was too permissive because the flat Delimited(OneOf(...)) allowed the sections to appear in any order.

I’ve updated the grammar to enforce the documented Databricks ordering:
* column definitions
* optional EXPECT constraints
* optional table constraints

I also added regression coverage for both invalid ordering cases you pointed out, so they now fail to parse.

peterbud (approved):
Thanks @vetrovk for the quick update and great contribution!

PR #8213, opened 2026-07-25
Fix CV10 corrupting string literals that use quote-doubling escapes

### LINKED ISSUE #8179: CV10 corrupts string literals that use quote-doubling escapes (e.g. 'O''Brien' → "O''Brien")
### What Happened

`CV10` (`convention.quoted_literals`) silently changes the **value** of a string literal when it rewrites the quote style, if that literal uses SQL's quote-doubling escape (`''` inside `'…'`, or `""` inside `"…"`).

The rule's normalizer (`_normalize_preferred_quoted_literal_style`) was adapted from Black's Python string handling and understands only **backslash** escaping. SQL escapes an embedded quote by **doubling** it, so when CV10 converts a literal to the other quote char it copies the doubled quotes through unchanged, and they stop being an escape in the new context.

This fires under the **default** config: `preferred_quoted_literal_style = consistent` derives the target quote char from the file's first literal, so any file whose first string is double-quoted will convert later `'…''…'` literals and corrupt them.

### Expected Behaviour

`sqlfluff fix` must never change what a query returns. The literal `'O''Brien'` (value `O'Brien`) should either be left alone or re-encoded to an equivalent double-quoted form — not turned into a literal with a different value.

### Observed Behaviour

The value silently changes. `'O''Brien'` → `"O''Brien"`: inside a MySQL double-quoted string `''` is **not** an escape, so the value becomes `O''Brien` (two apostrophes). Confirmed in both directions.

### How to reproduce

```sql
-- t.sql  (default full ruleset, so CV10 runs)
select "x", 'O''Brien' from t
```
```
sqlfluff fix t.sql --dialect mysql -f
```
```sql
-- result
select
    "x",
    "O''Brien"     -- value changed from O'Brien to O''Brien
from t
```
Reverse direction too: `"say ""hi"""` (value `say "hi"`) → `'say ""hi""'` (value `say ""hi""`).

**Control** — the backslash-escape path is fine: `'it\'s'` → `"it's"` (correct). Only the quote-doubling path corrupts, which is why the existing `CV10.yml` fixtures (bigquery, backslash-only) never caught it. Affected dialects are the ones in CV10's list that treat `''`/`""` as escapes (mysql/mariadb; the databricks/hive/sparksql grammars are worth checking too).

### Dialect / Version

`mysql` (and `mariadb`); sqlfluff 4.2.2 (also present on `main` @ 203599e).

### Possible fix / scope question

The raw-string branch already bails out (`return s`) when it can't convert safely. The minimal safe fix is the same: **skip the fix when the literal body contains an unescaped `orig_quote`** (i.e. a bare `'` in a `'…'` string signals quote-doubling this logic can't re-encode). A fuller fix would re-encode the doubling for the target quote char, but that depends on per-dialect escape semantics — I'd rather let you pick the scope. Happy to open a PR for whichever direction you prefer.

---

*Disclosure: this bug was found and this report written by an AI agent (Claude Code), running autonomously on this account; every repro above was executed against 4.2.2, not taken on trust.*

Sanjays2402:
Fixes #8179.

`CV10`'s quote normalizer was adapted from Black and understands only backslash escaping. SQL escapes an embedded quote by doubling it, so when the rule converted a literal to the other quote char it copied the doubled quotes through unchanged — where they are no longer an escape. `'O''Brien'` became `"O''Brien"`, silently changing the value the query returns.

This takes the minimal option from the issue: skip the fix when the literal body contains an unescaped `orig_quote`, mirroring the bail-out the raw-string branch already uses for cases it can't convert safely.

Literals that use quote doubling are now left as-is rather than being restyled, so `consistent`/`single_quotes`/`double_quotes` will no longer flag them. Backslash-escaped literals and all other conversions are unaffected. Re-encoding the doubling for the target quote char would be the fuller fix, but that depends on per-dialect escape semantics.

### Pull Request checklist

  - Other: `test/rules/std_CV10_test.py` — both directions, failing before this change and passing after.
- Added appropriate documentation for the change. N/A — no user-facing documentation change.
- Created GitHub issues for any relevant followup/future enhancements if appropriate. N/A

---
## Summary by cubic
Prevent CV10 from corrupting SQL string literals that use quote-doubling by skipping conversion when the literal contains an unescaped original quote. This preserves the literal’s value while keeping other conversions intact.

- **Bug Fixes**
  - Skip conversion if the body has an unescaped original quote (indicates quote-doubling).
  - Add tests covering both single- and double-quote directions.
  - Literals using quote-doubling are left as-is; `consistent`/`single_quotes`/`double_quotes` no longer flag them. Backslash-escaped cases are unchanged.

<sup>Written for commit 70984cc7b9f035195a679b32cea91a783053f9a6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8213?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/rules/convention/CV10.py
+++ b/src/sqlfluff/rules/convention/CV10.py
@@ -255,6 +255,19 @@ class Rule_CV10(BaseRule):
         escaped_orig_quote = regex.compile(rf"([^\\]|^)\\((?:\\\\)*){orig_quote}")
         body = s[first_quote_pos + len(orig_quote) : -len(orig_quote)]
 
+        unescaped_orig_quote = regex.compile(rf"(([^\\]|^)(\\\\)*){orig_quote}")
+        if unescaped_orig_quote.search(body):
+            # SQL escapes an embedded quote by doubling it (e.g. 'O''Brien'), which
+            # this backslash-oriented logic cannot re-encode for the new quote char.
+            # Copying the doubled quotes through unchanged would silently alter the
+            # value of the literal, so leave it alone.
+            self.logger.debug(
+                "Body %s contains an unescaped orig_quote, which suggests "
+                "quote doubling. Converting safely is impossible so skipping.",
+                body,
+            )
+            return s
+
         if "r" in prefix.lower():
             if unescaped_new_quote.search(body):
                 self.logger.debug(
--- a/test/rules/std_CV10_test.py
+++ b/test/rules/std_CV10_test.py
@@ -1,0 +1,25 @@
+"""Tests the python routines within CV10."""
+
+import sqlfluff
+
+
+def test__rules__std_CV10_quote_doubling_not_converted() -> None:
+    """CV10 leaves literals that escape a quote by doubling it alone.
+
+    Rewriting them would copy the doubled quotes into the other quote style,
+    where they are not an escape, silently changing the value of the literal.
+    The leading literal sets the consistent style, so the second one would
+    otherwise be converted.
+    """
+    sql = "SELECT \"x\", 'O''Brien' FROM t\n"
+    result = sqlfluff.fix(sql, rules=["CV10"], dialect="mysql")
+
+    assert "'O''Brien'" in result
+
+
+def test__rules__std_CV10_quote_doubling_not_converted_double_quotes() -> None:
+    """The same holds in the double-quoted direction."""
+    sql = 'SELECT \'x\', "say ""hi""" FROM t\n'
+    result = sqlfluff.fix(sql, rules=["CV10"], dialect="mysql")
+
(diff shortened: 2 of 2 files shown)


Sanjays2402:
Closing this — opened in error on my side, and I don't want to add review noise. Apologies for the churn.

PR #8216, opened 2026-07-25
fix(dialects/postgres): parse MERGE ... WHEN NOT MATCHED BY SOURCE/TARGET

hdimer:

PostgreSQL 17 extended `MERGE` with a `WHEN NOT MATCHED BY SOURCE` clause, and made `BY TARGET` an accepted synonym for the plain `WHEN NOT MATCHED`. The `postgres` dialect inherits the ANSI `MERGE` grammar, which knows neither, so valid Postgres 17 statements are reported as an unparsable section:

```sql
MERGE INTO wines w
USING wine_stock_changes s
ON s.winename = w.winename
WHEN NOT MATCHED BY TARGET AND s.stock_delta > 0 THEN
    INSERT VALUES (s.winename, s.stock_delta)
WHEN MATCHED AND w.stock + s.stock_delta > 0 THEN
    UPDATE SET stock = w.stock + s.stock_delta
WHEN NOT MATCHED BY SOURCE THEN
    DELETE;
```

This change:

- Overrides `MergeMatchSegment` in the `postgres` dialect to accept a `WHEN NOT MATCHED BY SOURCE` clause and the optional `BY TARGET` qualifier on `WHEN NOT MATCHED`, mirroring the existing `sparksql` and `bigquery` implementations. A `NOT MATCHED BY SOURCE` clause combines with `UPDATE`/`DELETE` (like a matched clause), consistent with those dialects.
- Registers `SOURCE` and `TARGET` as unreserved keywords. They remain usable as ordinary identifiers (e.g. a table aliased `AS source`), since naked identifiers only exclude reserved keywords.
- Adds a `postgres` parser fixture covering the new clauses (example adapted from the PG17 docs).

Ref: https://www.postgresql.org/docs/17/sql-merge.html

None that I could find. All existing `postgres` and `ansi` dialect fixtures still pass, and `ansi` continues to (correctly) reject `WHEN NOT MATCHED BY SOURCE`.

Two related constructs remain out of scope, as they are pre-existing gaps in the shared MERGE grammar rather than anything introduced here (both fail on plain clauses too, independent of this change): `THEN DO NOTHING`, and a leading `WITH ...` before `MERGE`. Happy to follow up on those separately if wanted.

### Pull Request checklist

- [x] Included `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (`postgres/merge_when_not_matched_by_source.sql` + auto-generated `.yml`).
- [x] Added grammar-level documentation (docstrings referencing the PG17 MERGE docs).
- [ ] Created GitHub issues for followup: not created yet; can file for `DO NOTHING` / `WITH ... MERGE` if desired.

---
## Summary by cubic
Adds PostgreSQL 17 `MERGE` parsing to the `postgres` dialect, including `WHEN NOT MATCHED BY SOURCE` and optional `BY TARGET`, and prevents these from leaking into `redshift`. This fixes false parse errors in Postgres 17 and keeps Redshift aligned with its documented `MERGE`.

- **New Features**
  - Extend `postgres` `MERGE` to support `WHEN NOT MATCHED BY SOURCE` and `BY TARGET`, allowing `UPDATE`/`DELETE`/`DO NOTHING`.
  - Register `SOURCE` and `TARGET` as unreserved keywords.
  - Add parser fixtures for the new clauses.

- **Bug Fixes**
  - Reset `redshift` `MERGE` clauses to ANSI so it does not inherit Postgres-specific `BY SOURCE`/`BY TARGET` or `DO NOTHING`.

<sup>Written for commit 187f42db784b654358a9d20792360ad3c87acb42. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_postgres.py
+++ b/src/sqlfluff/dialects/dialect_postgres.py
@@ -381,6 +381,13 @@ postgres_dialect.sets("unreserved_keywords").difference_update(
     get_keywords(postgres_keywords, "not-keyword")
 )
 
+# `SOURCE` and `TARGET` are not PostgreSQL keywords in general, but the
+# `MERGE ... WHEN NOT MATCHED BY {SOURCE,TARGET}` clauses (added in PostgreSQL
+# 17) need them. Register them as unreserved so they remain usable as ordinary
+# identifiers (e.g. a table aliased `AS source`) while the merge grammar can
+# still match them.
+postgres_dialect.sets("unreserved_keywords").update(["SOURCE", "TARGET"])
+
 # Add datetime units
 postgres_dialect.sets("datetime_units").update(
     [
@@ -5556,10 +5563,11 @@ class MergeMatchedClauseSegment(ansi.MergeMatchedClauseSegment):
 
 
 class MergeNotMatchedClauseSegment(ansi.MergeNotMatchedClauseSegment):
-    """The `WHEN NOT MATCHED` clause within a `MERGE` statement.
+    """The `WHEN NOT MATCHED [BY TARGET]` clause within a `MERGE` statement.
 
-    Overriding ANSI to allow `DO NOTHING`, which Postgres accepts as a merge
-    action alongside `INSERT`.
+    Overriding ANSI to allow the optional `BY TARGET` qualifier (a PostgreSQL 17
+    synonym for the plain `WHEN NOT MATCHED`) and `DO NOTHING`, which Postgres
+    accepts as a merge action alongside `INSERT`.
     https://www.postgresql.org/docs/current/sql-merge.html
     """
 
@@ -5568,6 +5576,7 @@ class MergeNotMatchedClauseSegment(ansi.MergeNotMatchedClauseSegment):
         "WHEN",
         "NOT",
         "MATCHED",
+        Sequence("BY", "TARGET", optional=True),
         Sequence("AND", Ref("ExpressionSegment"), optional=True),
         "THEN",
         Indent,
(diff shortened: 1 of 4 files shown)


hdimer:
Fair close, and thanks for building AgentScan. To be upfront: this was AI-assisted (I use Claude Code) and I should have said so here. The change itself is real PostgreSQL 17 support (MERGE ... WHEN NOT MATCHED BY SOURCE / BY TARGET), which I've reviewed and tested; it mirrors the existing sparksql and bigquery grammar. If a disclosed resubmission would be welcome I'll happily reopen with the disclosure up front and if you'd rather just take the change directly, please do.

peterbud:
Hi @hdimer - we have just introduced AgentScan to the repo, and in fact that evaluated your contribution as  "organic":
https://github.com/sqlfluff/sqlfluff/actions/runs/30166919740

We are OK to have AI assisted PRs, if it is disclosed - see our CONTRIBUTING.md

This PR was closed by accident as @keraion I believe included the term in his reply:
> ...resolve https://github.com/sqlfluff/sqlfluff/pull/8216 

So when that PR was closed, Github closed this PR as well. 

I'm re-opening it.

peterbud (changes requested):
Hi @hdimer  - thanks for opening this PR, and welcome.

I have quickly reviewed the PR, and there is one blocking issue: this change appears to widen Postgres 17 MERGE syntax into Redshift through dialect inheritance.

Redshift is copied from the Postgres dialect and still reuses the shared MERGE match grammar, so adding WHEN NOT MATCHED BY SOURCE / BY TARGET support in Postgres now makes that syntax parse successfully under the redshift dialect as well. That looks incorrect based on Redshift’s documented MERGE syntax, which only supports WHEN MATCHED, WHEN NOT MATCHED, and REMOVE DUPLICATES.

Could we scope this more tightly so the new grammar only applies to Postgres, or add an explicit Redshift override to preserve its existing MERGE behavior?

Also check out the recently merged #8223 and make sure there is no conflict.

hdimer:
Thanks @peterbud — good catches on both. Fixed:

- **Redshift:** added overrides resetting `MergeMatchSegment`, `MergeMatchedClauseSegment`, and `MergeNotMatchedClauseSegment` to the ANSI grammar, so Redshift keeps its documented `WHEN MATCHED` / `WHEN NOT MATCHED` / `REMOVE DUPLICATES` behavior and no longer accepts `BY SOURCE` / `BY TARGET`. This also strips the `DO NOTHING` that #8223 had similarly leaked into Redshift via the same inheritance. The existing redshift merge fixture is unchanged.
- **#8223 conflict:** rebased onto `main` and merged both features into a single `WHEN NOT MATCHED` clause (`[BY TARGET] ... THEN {INSERT | DO NOTHING}`), and added `DO NOTHING` to the new `BY SOURCE` clause for consistency.

Verified the full matrix (postgres accepts, redshift rejects) and the suite is green.

peterbud (approved):
Thanks @hdimer !

PR #8217, opened 2026-07-25
fix(ci): collapse agent-scan into one pull_request_target workflow

keraion:
The analyze/moderate split existed because `pull_request` runs from forks get a read-only token, forcing a `workflow_run` hand-off with artifact plumbing and PR-number resolution. That resolution used `listPullRequestsAssociatedWithCommit`, which can't see fork-only commits — so moderation failed on exactly the PRs it was built for ([example run](https://github.com/sqlfluff/sqlfluff/actions/runs/30172181698) failing to resolve #8216). It also meant first-time contributors — the accounts AgentScan targets — needed a maintainer to approve CI before the scan ran.

`pull_request_target` (the action's own documented trigger) runs the base-branch workflow definition with a write token and no approval gate, so classification and moderation happen in one job and the PR number comes straight from the payload. Safe because the workflow never checks out or executes PR code — noted in the file for future editors. Moderation behavior (label, sentinel comment, auto-close) is unchanged.

The first commit fixes the fork-PR resolution in the two-workflow shape (kept for history); the second collapses the pair into a single workflow.

Both verified in a fork: the fixed fallback resolved a live PR by `owner:branch`, and the collapsed workflow ran under `pull_request_target` with no approval hold and no-op'd cleanly on an organic classification.

Written with Claude Code.


keraion:
@peterbud I had a first approach that kept the split, but Claude made a good point that these don't really need to be split out and we want the action to run even for first time contributors. This version does use `pull_request_target` but we strictly only run items that are from main closing the security issue there.

peterbud:
Hi @keraion - thanks for looking into this. 

I wanted to avoid `pull_request_target` at all costs, as I have read too much about exploits using that trigger - so that's the reason for the original split workflow solution. But seeing that it does not work at all due to limitations of the Github API.... I agree maybe that's the only viable solution.

peterbud (approved):

PR #8219, opened 2026-07-25
BigQuery: only parse date parts in date part functions

### LINKED ISSUE #5015: CP01: SQLFluff 0.9.1 capitalizes columns named year when used as parameters in DATE_ADD
### Search before asking

### What Happened

When run on a SQL file containing a line like REGEXP_EXTRACT(year, r"[0-9]{4}"), SQLFluff issued the CP01 linting error instead of the RF04 linting error.

### Expected Behaviour

When run on a SQL file containing a line like REGEXP_EXTRACT(year, r"[0-9]{4}"), SQLFluff should understand that year is not a parameter but a variable name and not issue the CP01 error.  RF04 would be issued.

### Observed Behaviour

I was linting BigQuery SQL with the line REGEXP_EXTRACT(year, r"[0-9]{4}") and unexpectedly got the CP01 error.  I believe this to be an error since year is the _variable name_ in this context, not a keyword.

### How to reproduce

Run the following to reproduce on online.sqlfluff.com:
- Set the dialect to BigQuery
- Run the following SQL:
```
SELECT REGEXP_EXTRACT(year, r"[0-9]{4}")
FROM
    table
```

### Dialect

BigQuery

### Version

sqlfluff, version 2.0.0 (but I think it's still broken in the newest version, see the online.sqlfluff.com example above)

### Configuration

```
[sqlfluff]
dialect = bigquery
templater = dbt
exclude_rules = LT01,LT02,AL01,CP02,LT05,RF04,AL07,ST06

[sqlfluff:layout:type:comma]
line_position = leading

capitalisation_policy = upper

capitalisation_policy = upper
```

### Code of Conduct

pratyushsinghal7:

Fixes #5015

In the BigQuery dialect, `FunctionContentsExpressionGrammar` included `DatetimeUnitSegment` and `DatePartWeekSegment`, so *any* function argument matching a date part keyword (`year`, `day`, etc.) was parsed as a `date_part` rather than a column reference. This caused CP01 (keyword capitalisation) to fire on column names like `year`:

```sql
SELECT REGEXP_EXTRACT(year, r"[0-9]{4}") FROM some_table
-- L: 1 | P: 23 | CP01 | Keywords must be consistently upper case.
```

Functions which genuinely take date part arguments (`EXTRACT`, `DATE_TRUNC`, `DATE_DIFF`, `LAST_DAY`, ...) are already routed through `DateTimeFunctionContentsSegment` via the `date_part_function_name` set, so the generic grammar doesn't need to match date parts. This change removes the two references from the generic grammar.

Two existing fixtures change as a result, both for the better:
- `parameters.yml`: `parse_date("%Y%m", year)` now parses `year` as an identifier — the fixture's own comment says `-- this should parse year as an identifier`.
- `select_safe_function.yml`: `SAFE.DATEADD(DAY, ...)` now parses `DAY` as a column reference (`DATEADD` isn't a BigQuery date part function).

Added a parser fixture covering the reported case plus checks that `EXTRACT`/`DATE_TRUNC`/`DATE_DIFF` still parse date parts.

Disclosure: this change was made with AI assistance (Devin), reviewed and tested before submission.

Arguments named after date parts in non-date-part functions now parse as column references instead of `date_part` segments, which may shift results for rules that treat keywords/identifiers differently (as intended here).

### Pull Request checklist

  - Other.
- Created GitHub issues for any relevant followups/issues I identified.

---
## Summary by cubic
Fix BigQuery parsing so date part keywords are only recognized in date-part functions, including `SAFE.`-qualified ones (Fixes #5015). Prevents CP01 false positives on columns named like year or day in non-date-part functions.

- **Bug Fixes**
  - Route `SAFE.`-qualified date-part functions through `DateTimeFunctionContentsSegment` and match each argument as a date part or expression to prefer date parts where valid.
  - Non-date-part functions now treat columns named like date parts as identifiers (e.g., `REGEXP_EXTRACT(year, ...)`, `SAFE.DATEADD(DAY, ...)`).
  - Updated fixtures and added parser tests, including coverage for `SAFE.DATE_TRUNC`, `SAFE.DATETIME_TRUNC`, `SAFE.TIMESTAMP_TRUNC`, `SAFE.DATE_DIFF`, and `SAFE.LAST_DAY`.

<sup>Written for commit 1ccd442fdd408ac2e8ebb5623f310081ec03803b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8219?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -345,8 +345,6 @@ bigquery_dialect.replace(
         )
     ),
     FunctionContentsExpressionGrammar=OneOf(
-        Ref("DatetimeUnitSegment"),
-        Ref("DatePartWeekSegment"),
         Sequence(
             Ref("ExpressionSegment"),
             Sequence(OneOf("IGNORE", "RESPECT"), "NULLS", optional=True),
--- a/src/sqlfluff/dialects/dialect_bigquery.py
+++ b/src/sqlfluff/dialects/dialect_bigquery.py
@@ -1188,15 +1188,39 @@ class FunctionNameSegment(ansi.FunctionNameSegment):
     )
 
 
+class DatePartFunctionNameSegment(ansi.DatePartFunctionNameSegment):
+    """Date part function name, allowing an optional SAFE. prefix.
+
+    BigQuery allows date part functions to be SAFE-qualified (e.g.
+    ``SAFE.DATE_TRUNC(some_date, YEAR)``) and the date part argument
+    should still parse as a date part rather than a column reference.
+    https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-reference#safe_prefix
+    """
+
+    match_grammar: Matchable = Sequence(
+        Sequence("SAFE", Ref("DotSegment"), optional=True),
+        Ref("DatePartFunctionName"),
+    )
+
+
 class DateTimeFunctionContentsSegment(ansi.DateTimeFunctionContentsSegment):
-    """Datetime function contents segment."""
+    """Datetime function contents segment.
+
+    Each argument is either a date part or an expression: matching
+    them one argument at a time (rather than via the generic
+    ``FunctionContentsGrammar``, which contains its own delimited list)
+    ensures date parts are always preferred over column references for
+    every argument.
+    """
 
     match_grammar = Sequence(
         Bracketed(
             Delimited(
-                Ref("DatetimeUnitSegment"),
-                Ref("DatePartWeekSegment"),
-                Ref("FunctionContentsGrammar"),
+                OneOf(
+                    Ref("DatetimeUnitSegment"),
+                    Ref("DatePartWeekSegment"),
+                    Ref("FunctionContentsExpressionGrammar"),
+                ),
             ),
         )
     )
--- a/test/fixtures/dialects/bigquery/parameters.yml
+++ b/test/fixtures/dialects/bigquery/parameters.yml
@@ -3,7 +3,7 @@
 # computed by SQLFluff when running the tests. Please run
 # `python test/generate_parse_fixture_yml.py`  to generate them after adding or
 # altering SQL files.
-_hash: 8c54cfecb9e82789d21a903c1597a099d07f8d290574d9b70d891d49a51329a5
+_hash: 7577d3b4ebea648adc117297efba03042cef3fc329305a836f645f6e5a3a1176
 file:
 - statement:
     select_statement:
@@ -92,12 +92,14 @@ file:
               function_name_identifier: parse_date
             function_contents:
               bracketed:
-                start_bracket: (
-                expression:
+              - start_bracket: (
+              - expression:
                   quoted_literal: '"%Y%m"'
-                comma: ','
-                date_part: year
-                end_bracket: )
+              - comma: ','
+              - expression:
+                  column_reference:
+                    naked_identifier: year
+              - end_bracket: )
 - statement_terminator: ;
 - statement:
     select_statement:
--- a/test/fixtures/dialects/bigquery/select_function_datepart_column.sql
+++ b/test/fixtures/dialects/bigquery/select_function_datepart_column.sql
@@ -1,0 +1,12 @@
+-- Columns named after date parts should parse as column references in
+-- functions which don't take date part arguments.
+-- https://github.com/sqlfluff/sqlfluff/issues/5015
+SELECT REGEXP_EXTRACT(year, r"[0-9]{4}")
+FROM some_table;
+
+-- Date part functions should still parse date parts.
+SELECT
+    EXTRACT(YEAR FROM some_date),
+    DATE_TRUNC(some_date, WEEK(MONDAY)),
+    DATE_DIFF(date_a, date_b, DAY)
+FROM some_table;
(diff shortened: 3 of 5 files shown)


peterbud (changes requested):
Hi @pratyushsinghal7  - thanks for this PR.

This however introduces a regression for SAFE-qualified date-part functions. The generic argument grammar no longer recognizes date parts, and the special date-part branch appears to only catch bare function names, so something like `SAFE.DATE_TRUNC(some_date, YEAR)` now parses `YEAR` as a column reference instead of a date part.

I’d suggest either routing SAFE-qualified date-part functions through the same specialized path, and adding coverage for `SAFE.DATE_TRUNC`/similar cases before merging.

adhavan18:
Not my PR, but I had a BigQuery checkout handy so I reproduced @peterbud's finding to pin down how wide it is. It is not only `DATE_TRUNC`: every SAFE-qualified date-part function loses the date part.

Applied this branch on top of `a36f5f2b1` and parsed each form, checking whether the final argument lands as a `date_part` or as a `column_reference`:

| statement | main | this branch |
| --- | --- | --- |
| `DATE_TRUNC(d, YEAR)` | date_part | date_part |
| `SAFE.DATE_TRUNC(d, YEAR)` | date_part | **column_reference** |
| `DATE_DIFF(a, b, DAY)` | date_part | date_part |
| `SAFE.DATE_DIFF(a, b, DAY)` | date_part | **column_reference** |
| `TIMESTAMP_TRUNC(ts, HOUR)` | date_part | date_part |
| `SAFE.TIMESTAMP_TRUNC(ts, HOUR)` | date_part | **column_reference** |
| `DATETIME_TRUNC(dt, MINUTE)` | date_part | date_part |
| `SAFE.DATETIME_TRUNC(dt, MINUTE)` | date_part | **column_reference** |
| `LAST_DAY(d, MONTH)` | date_part | date_part |
| `SAFE.LAST_DAY(d, MONTH)` | date_part | **column_reference** |
| `DATE_ADD(d, INTERVAL 1 DAY)` | date_part | date_part |
| `SAFE.DATE_ADD(d, INTERVAL 1 DAY)` | date_part | date_part |

Everything parses either way, which is what makes it easy to miss: nothing goes unparsable, the date part just silently becomes a column reference, so rules that resolve references see a column that does not exist.

Two things that might save you a round trip:

- The `INTERVAL` form is unaffected, so whatever you do for the bare names probably only needs to cover the `(value, DATE_PART)` shape.
- `SAFE.` is not the only prefix in the dialect. Worth checking whether the same specialized path needs to handle other qualified forms before settling on the fix.

Happy to re-run this matrix against your next push if that helps.

pratyushsinghal7:
@peterbud Done — `DatePartFunctionNameSegment` now accepts an optional `SAFE.` prefix, so SAFE-qualified date part functions route through the same specialized `DateTimeFunctionContentsSegment` path, and the fixture now covers `SAFE.DATE_TRUNC` / `SAFE.DATETIME_TRUNC` / `SAFE.TIMESTAMP_TRUNC` / `SAFE.DATE_DIFF` / `SAFE.LAST_DAY`.

While fixing this I found the earlier CI failures (`rules_suite-rust`, parity corpus) were caused by `DateTimeFunctionContentsSegment` delegating to the generic `FunctionContentsGrammar`: its nested delimited list could consume the whole argument list in the Rust parser, so date parts silently became column references once they were removed from the generic grammar (previously masked because the generic grammar parsed date parts too). It now matches arguments one at a time as `OneOf(DatetimeUnitSegment, DatePartWeekSegment, FunctionContentsExpressionGrammar)`, which leaves the Python parse trees unchanged and restores Python/Rust parity — the full parity corpus, dialect fixture suite and rule yaml suites pass locally with the Rust parser built from this branch.

@adhavan18 thanks for the matrix — re-ran all twelve forms on this branch and they all parse the final argument as `date_part` now (the `INTERVAL` form was already unaffected). Branch is also rebased on latest main.

peterbud:
Hi @pratyushsinghal7, this still looks incorrect for date-part functions with expression arguments. The change fixes the non-date-part case like `REGEXP_EXTRACT(year, ...)`, but the new `DateTimeFunctionContentsSegment` now prefers date parts for every positional argument in date-part functions, not just the slots where a date part is actually valid.

That means identifiers such as `year`, `month`, or `day` can still be misparsed as date parts when they appear in ordinary expression positions. For example:

```sql
SELECT LAST_DAY(year, MONTH) FROM some_table;
```

In this case, `year` should parse as a column reference in the first argument, but it is still recognized as a date part, which then surfaces as a CP01 false positive.

So the current change appears to move the ambiguity rather than resolve it. The added tests only cover date-part functions using neutral identifiers like `some_date` and `date_a`, so they miss this regression. I think this needs either argument-position-aware parsing for the affected functions, or coverage that proves identifiers named like date parts still parse correctly in expression slots.

peterbud:
Closing due to inactivity.

PR #8226, opened 2026-07-26
rust(fix): Resolve two codegen-determinism hazards (id-cache collisions and unsorted currency-symbol regex)

rubytobi:

`utils/build_parsers.py` builds the Rust dialect tables by caching flattened grammar objects in a dict keyed by Python's `id()`. If a cached grammar got garbage collected, CPython could hand its freed memory address to a later, unrelated grammar, which then silently inherited the stale cache entry and wired the wrong subgrammar into the generated tables. This was observed for real with oracle's `AttributeIndicatorSegment` picking up a dead `ModuloSegment` entry. The fix pins every grammar placed in the cache, not just the synthetic ones, so no `id()` can be reused while its entry is live.

Separately, tsql's money-literal lexer pattern built its character class from a set of currency symbols with an unsorted `"".join(...)`, so the generated regex string's byte order depended on the interpreter's hash seed, unlike every other `dialect.sets(...)`-derived pattern in the codebase, which already sorts first. That made the checked-in generated Rust lexer output non-reproducible across rebuilds. The join is now sorted.

Both are regression-guarded with new tests in `test/core/parser/parity/special_cases_test.py`, added to the parity test suite for cases that need Python-side instrumentation (object-lifetime and subprocess hash-seed probes) rather than a plain SQL-plus-config case.

None expected. Both fixes only affect the codegen path (`utils/build_parsers.py`, `utils/build_lexers.py` regex assembly) and should make generated output deterministic without changing what it matches.

### Pull Request checklist

  - Other.

---
## Summary by cubic
Fix two codegen nondeterminism issues so the generated Rust parser/lexer output is reproducible. No behavior change to what the parser or lexer matches.

- **Bug Fixes**
  - Parser codegen: In `utils/build_parsers.py`, `TableBuilder` now pins every cached grammar to prevent `id()` reuse after GC, replacing `_synthetic_grammars` with `_cached_grammars` and appending in `flatten_grammar`.
  - Lexer codegen (T-SQL): Sort `currency_symbols` when building the money-literal regex in `src/sqlfluff/dialects/dialect_tsql.py` to make the pattern stable across hash seeds.
  - Tests: Add `test/core/parser/parity/special_cases_test.py` to guard object-lifetime and hash-seed stability; update `test/fixtures/parity/README.md`.

<sup>Written for commit 4de8bae62d9777e452549f69ffe3d092840ad2ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8226?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_tsql.py
+++ b/src/sqlfluff/dialects/dialect_tsql.py
@@ -263,8 +263,10 @@ tsql_dialect.insert_lexer_matchers(
             (
                 r"([xX]'([\da-fA-F][\da-fA-F])+'"
                 r"|0[xX][\da-fA-F]*"
-                r"|[+-]*[" + "".join(tsql_dialect.sets("currency_symbols")) + r"]"
-                r"[" + "".join(tsql_dialect.sets("currency_symbols")) + r"+-]*"
+                r"|[+-]*["
+                + "".join(sorted(tsql_dialect.sets("currency_symbols")))
+                + r"]"
+                r"[" + "".join(sorted(tsql_dialect.sets("currency_symbols"))) + r"+-]*"
                 r"(?>\d+\.\d+|\d+\.(?![\.\w])|\.\d+|\d+))"
             ),
             LiteralSegment,
--- a/test/core/parser/parity/special_cases_test.py
+++ b/test/core/parser/parity/special_cases_test.py
@@ -18,12 +18,6 @@ import pytest
 _REPO_ROOT = Path(__file__).resolve().parents[4]
 
 
-@pytest.mark.xfail(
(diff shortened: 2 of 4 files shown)


peterbud (approved):
Thanks @rubytobi !

PR #8228, opened 2026-07-26
feat(clickhouse): add support for ```TRUNCATE DATABASE```

AndrewKurg:
### Brief summary of the change made
fixes #8227. Adds support for ClickHouse [```TRUNCATE DATABASE```](https://clickhouse.com/docs/reference/statements/truncate#truncate-database) statement.
I am also planning to add support for [```TRUNCATE ALL TABLES```](https://clickhouse.com/docs/reference/statements/truncate#truncate-all-tables) in a separate PR. 

### Are there any other side effects of this change that we should be aware of?
There should not be any

### Pull Request checklist
- [x] Please confirm you have completed any of the necessary steps below.

- Included test cases to demonstrate any code changes, which may be one or more of the following:
  - `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (note YML files can be auto generated with `tox -e generate-fixture-yml`).
- Added appropriate documentation for the change.

---
## Summary by cubic
Add parsing support for ClickHouse TRUNCATE DATABASE in the ClickHouse dialect so these statements parse and lint correctly, including optional IF EXISTS and ON CLUSTER. Fixes #8227.

- **New Features**
  - Implemented TruncateDatabaseStatementSegment and registered it in StatementSegment.
  - Added fixtures and parse tests for base, IF EXISTS, and ON CLUSTER variants.

<sup>Written for commit 5ac02565865a7c080e3a5a1acdd66395e4aa3c60. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8228?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #8227, opened 2026-07-26
ClickHouse: Missing syntax support for TRUNCATE DATABASE


AndrewKurg:
### Search before asking

- [x] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

sqlfluff can't parse ```TRUNCATE DATABASE``` using clickhouse dialect

### Expected Behaviour

The statement is parsed correctly

### Observed Behaviour

```
[L:  1, P:  1]      |file:
[L:  1, P:  1]      |    statement:
[L:  1, P:  1]      |        truncate_table:
[L:  1, P:  1]      |            keyword:                                          'truncate'
[L:  1, P:  9]      |            whitespace:                                       ' '
[L:  1, P: 10]      |            table_reference:
[L:  1, P: 10]      |                naked_identifier:                             'database'
[L:  1, P: 18]      |    whitespace:                                               ' '
[L:  1, P: 19]      |    unparsable:                                               !! Expected: 'Nothing else in FileSegment.'
[L:  1, P: 19]      |        word:                                                 'test'
[L:  1, P: 23]      |        semicolon:                                            ';'
[L:  1, P: 24]      |    newline:                                                  '\n'
[L:  2, P:  1]      |    [META] end_of_file:

==== parsing violations ====
L:   1 | P:  19 |  PRS | Line 1, Position 19: Found unparsable section: 'test;'

```

### How to reproduce

```
sqlfluff parse --dialect=clickhouse -
```
```
truncate database test;
```

### Dialect

clickhouse

### Version

sqlfluff, version 4.2.2
Python 3.11.15

### Configuration

default

### Are you willing to work on and submit a PR to address the issue?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/sqlfluff/sqlfluff/blob/main/CODE_OF_CONDUCT.md)

PR #8231, opened 2026-07-27
ClickHouse: Support TRUNCATE DATABASE statement

sejalr28:
## Summary

This PR adds parser support for the `TRUNCATE DATABASE` statement in the ClickHouse dialect.

Previously, SQLFluff only parsed `TRUNCATE TABLE` statements. This change extends the grammar to recognize `TRUNCATE DATABASE` using the existing `DatabaseReferenceSegment`.

## Changes

- Added support for `TRUNCATE DATABASE <database_name>`
- Added a regression test in `truncate_table.sql`
- Updated the corresponding parse fixture

## Testing

- Regenerated ClickHouse parse fixtures
- Ran ClickHouse dialect tests (`156 passed`)

---
## Summary by cubic
Adds parser support for TRUNCATE DATABASE in the ClickHouse dialect. SQLFluff now parses TRUNCATE DATABASE statements (e.g., TRUNCATE DATABASE test) alongside TRUNCATE TABLE, with a regression test and updated fixtures.

<sup>Written for commit 76e2d903d408d1f61fc24439393c84cf9630d143. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8231?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


adhavan18 (commented):
Heads-up @sejalr28 — this overlaps with #8228, which adds the same ClickHouse `TRUNCATE DATABASE` support and was opened about half a day earlier against #8227. Worth linking the two so a maintainer only has to review one.

Since the two take different approaches, I checked both locally against `main` to see how they actually behave:

| statement | `main` | this PR (#8231) | #8228 |
|---|---|---|---|
| `TRUNCATE DATABASE test;` | unparsable | parses | parses |
| `TRUNCATE DATABASE IF EXISTS test;` | unparsable | **unparsable** | parses |
| `TRUNCATE DATABASE test ON CLUSTER clstr;` | unparsable | parses | parses |
| `TRUNCATE TABLE DATABASE test;` | unparsable | **parses** | unparsable |

Two things fall out of extending `TruncateStatementSegment` rather than adding a separate segment:

1. **`IF EXISTS` ends up in the wrong position.** The existing sequence is `TRUNCATE [TABLE] [IF EXISTS] <ref> [ON CLUSTER] [SYNC]`, so `IF EXISTS` has to come *before* the `OneOf`. But [the docs](https://clickhouse.com/docs/sql-reference/statements/truncate) put it after the keyword — `TRUNCATE DATABASE [IF EXISTS] db` — so that form doesn't parse here.

2. **`TRUNCATE TABLE DATABASE test;` now parses**, because `TABLE` is optional and independent of the new `DATABASE` branch. That's accepted on this branch but rejected on `main`, so it's a small loosening rather than an addition.

The resulting parse tree also types a database statement as `truncate_table`, which may be awkward for rules that key off segment type.

Not a maintainer, just another contributor who's been in the dialect files lately — take or leave. But given #8228 already handles `IF EXISTS`/`ON CLUSTER` and is linked to the issue, it might be least effort for everyone to consolidate there. Either way, the `TRUNCATE TABLE DATABASE` case is worth a fixture wherever this lands.

peterbud:
Hi @sejalr28  - I agree that this is already being addressed in #8228

peterbud:
Closing in favor of #8228

PR #8243, opened 2026-07-28
fix(rust): match Rust parser bracket matching to Python's behaviour on crossed, unclosed, and dialect-specific brackets

rubytobi:

This closes several bracket matching gaps between the Rust parser and Python's reference implementation, all previously tracked as `xfail` cases in the parity test suite.

- The `Anything()` grammar's bracket-recursion path (`match_bracket_recursively`) now raises a parse error instead of silently producing a fabricated tree when it hits a crossed bracket (a closing bracket of the wrong type) or runs out of input before finding its closing bracket, matching Python's `resolve_bracket`.
- Bracket recognition in that same path, and in the parser's top-level bracket handling, is no longer limited to a hardcoded set of `(`, `[`, `{`. It's now driven by each dialect's actual bracket pairs (generated from the dialect definitions), so dialect-specific brackets such as Snowflake's exclude bracket `{-`/`-}` are recognised, typed, and structurally preserved the same way Python does.
- `Bracketed` content matching now distinguishes a failed *optional* content element (which Python lets the sequence continue past) from a failed *required* one (which should stop the sequence there). This fixes cases where a required element's failure was previously masked, including one where a STRICT bracket would incorrectly succeed with a partial match if the failure happened to land exactly at the closing bracket.
- The error messages produced for these content-failure cases (naming the expected grammar/token and whether it is "to start sequence" or "after X") are now derived the same way Python derives them, rather than falling back to a generic message.

Supporting changes: dialect bracket pairs are generated into the Rust dialect tables (`build_dialects.py` / `build_lexers.py`) alongside the existing lexer generation, and several previously `xfail`'d parity fixtures in `test/fixtures/parity/regressions.yml` are now passing and documented as such.

Bracket-opener detection for the `Anything()` grammar path is now a lookup against the dialect's bracket set rather than a fixed ASCII check, so any dialect that defines additional bracket-like pairs will pick them up automatically through this path.

### Pull Request checklist

  - Other.

---
## Summary by cubic
Aligns Rust bracket behavior with Python across all dialects. Uses each dialect’s `bracket_pairs`, raises on crossed/unclosed brackets, and fixes `Bracketed` content handling and error messages.

- **Bug Fixes**
  - Dialect-aware brackets end-to-end: lexer precomputes `matching_bracket_idx` using each dialect’s `bracket_pairs`; parser/`greedy_match` detect opens/closes via `Dialect::get_bracket_pairs` (e.g., Snowflake `{-`/`-}`); stray dialect closers now abort terminator search.
  - `Anything()` recursion now errors on crossed or unclosed brackets instead of fabricating a match.
  - `Bracketed` content: stop on required failures; STRICT no longer accepts partial matches at the close; optional failures produce “Nothing here.”; error text mirrors Python (“to start sequence”/“after X” with correct found-token fallbacks), without rejecting valid empty parens.
  - Reset stale `matching_bracket_idx` before recomputing bracket pairs to avoid carrying incorrect links between runs.

- **Refactors**
  - Introduced `BracketPairEntry` and `BracketPairSet` with shared lookups (`compute_matching_indices`, `is_open`/`is_close`). Generate per‑dialect `bracket_pairs` in `utils/build_lexers.py`, expose via `Dialect::get_bracket_pairs`; lexer has an ASCII default with `persists` flags and can be overridden via `Lexer::with_bracket_pairs`; the Python bridge now uses it.
  - Bracket leaf types and persistence now match Python; tests and the harness updated to supply dialect bracket pairs.

<sup>Written for commit 5033156f9819ed1b0afad3eb924e37975cca2ec5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8243?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/python.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/python.rs
@@ -662,12 +662,16 @@ fn compute_bracket_pairs(tokens: &mut [Token], bracket_pairs: &[(&str, &str, &st
         let raw = tokens[idx].raw();
 
         // Check if this is an opening bracket
-        if let Some(pair_idx) = bracket_pairs.iter().position(|(open, _, _, _, _)| *open == raw) {
+        if let Some(pair_idx) = bracket_pairs
+            .iter()
+            .position(|(open, _, _, _, _)| *open == raw)
+        {
             bracket_stack.push((idx, pair_idx));
         }
         // Check if this is a closing bracket
-        else if let Some(expected_idx) =
-            bracket_pairs.iter().position(|(_, close, _, _, _)| *close == raw)
+        else if let Some(expected_idx) = bracket_pairs
+            .iter()
+            .position(|(_, close, _, _, _)| *close == raw)
         {
             // Only the most-recently-opened bracket (the top of the stack) may
             // be closed next, per LIFO nesting discipline. See the sibling
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/python.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/python.rs
@@ -514,7 +514,7 @@ impl PyParser {
         // Compute bracket pairs for the tokens.
         // When tokens come from Python (lexed by Python lexer), matching_bracket_idx is None.
         // We need to compute it for the Bracketed grammar to work correctly.
-        compute_bracket_pairs(&mut rust_tokens);
+        compute_bracket_pairs(&mut rust_tokens, self.dialect.get_bracket_pairs());
 
         // Create parser
         let mut parser = Parser::new_with_max_parse_depth(
@@ -555,7 +555,7 @@ impl PyParser {
         let mut rust_tokens: Vec<Token> = tokens.into_iter().map(|t| t.into()).collect();
 
         // Compute bracket pairs for the tokens.
-        compute_bracket_pairs(&mut rust_tokens);
+        compute_bracket_pairs(&mut rust_tokens, self.dialect.get_bracket_pairs());
 
         // Create parser
         let mut parser = Parser::new_with_max_parse_depth(
@@ -601,7 +601,7 @@ impl PyParser {
         let mut rust_tokens: Vec<Token> = tokens.into_iter().map(|t| t.into()).collect();
 
         // Compute bracket pairs for the tokens.
-        compute_bracket_pairs(&mut rust_tokens);
+        compute_bracket_pairs(&mut rust_tokens, self.dialect.get_bracket_pairs());
 
         // Create parser with grammar counting enabled
         let mut parser = Parser::new_with_max_parse_depth(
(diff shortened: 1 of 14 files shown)


sqlfluffrs/sqlfluffrs_parser/src/parser/python.rs
```diff
@@ -647,9 +647,9 @@ impl PyParser {
 ///
 /// This is necessary when tokens come from Python (via PyToken -> Token conversion)
 /// because the Python lexer doesn't compute these indices.
-fn compute_bracket_pairs(tokens: &mut [Token]) {
-    // Stack to track opening brackets: (index, bracket_char)
-    let mut bracket_stack: Vec<(usize, char)> = Vec::new();
+fn compute_bracket_pairs(tokens: &mut [Token], bracket_pairs: &[(&str, &str, &str, &str, bool)]) {
```
keraion: This will need to be brought in, but I think this pattern can be reused in a few places.
```suggestion
fn compute_bracket_pairs(tokens: &mut [Token], bracket_pairs: &[BracketPairEntry]) {
```

sqlfluffrs/sqlfluffrs_types/src/matcher.rs
```diff
@@ -12,6 +12,11 @@ use crate::{token::CaseFold, PositionMarker, RegexModeGroup, Token, TokenConfig}
 /// Function pointer type for token generation: one of `Token::{kind}_token`.
 pub type TokenGenerator = fn(String, PositionMarker, TokenConfig) -> Token;
 
+/// A single bracket pair: (open raw text, close raw text, start-bracket
+/// segment type, end-bracket segment type, persists). See
+/// `utils/build_lexers.py::generate_bracket_pairs` for how these are derived.
+pub type BracketPairEntry = (&'static str, &'static str, &'static str, &'static str, bool);
```
keraion: There are a lot of `(_, close, _, _, _)` or `(open, _, _, _, _)` around. Let's stick with a named struct instead
```suggestion
pub struct BracketPairEntry {
    pub open: &'static str,
    pub close: &'static str,
    pub start_type: &'static str,
    pub end_type: &'static str,
    pub persists: bool,
}
```

keraion (commented):
Overall a good start. I think we can probably simplify some things. It might also be worth having a `struct BracketPairSet(Vec<BracketPairEntry>)` with some of the reused methods.

Additionally the comments feel maybe a bit verbose and could probably be trimmed down to help understanding.

sqlfluffrs/sqlfluffrs_types/src/matcher.rs
```diff
@@ -12,6 +12,84 @@ use crate::{token::CaseFold, PositionMarker, RegexModeGroup, Token, TokenConfig}
...
+
+    pub fn iter(&self) -> impl Iterator<Item = &BracketPairEntry> {
+        self.0.iter()
+    }
+
+    /// Pre-compute `matching_bracket_idx` for every bracket token via LIFO
+    /// nesting order (Python parity: `resolve_bracket`, match_algorithms.py).
+    /// Only code tokens are considered, so bracket characters inside a
+    /// comment can't pair with real brackets in the SQL.
+    pub fn compute_matching_indices(&self, tokens: &mut [Token]) {
```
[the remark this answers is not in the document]
rubytobi: Not really applicable. Python's bracket matching is also raw-text based (`StringParser` checks `raw_upper == template`, no type check), so no parity gap here. 
The reason angle brackets don't collide is that BigQuery/Trino/SparkSQL keep `<`/`>` out of the default `bracket_pairs` set entirely, using a separate `angle_bracket_pairs` set only in specific grammar contexts. No dialect's `BracketPairSet` includes `<`/`>`, so nothing can currently mismatch. If we add parsing angle-bracket types later, the fix is a separate scoped bracket set based on context, probably not touching this function.

keraion (approved):
Ok, these changes help with reading. Thanks!

PR #8244, opened 2026-07-28
fix(rust): Parser parity across Ref errors, arena-tree panics, regex case folding, and indentation config

rubytobi:

Fixes four Python/Rust parser parity gaps, three found by the new parity test harness and one extra observation, where the Rust parser (`RustParser`) previously diverged silently from the pure-Python parser instead of matching its behaviour or errors:

1. Dangling `Ref` grammars now raise, instead of silently failing the branch. When a grammar references a keyword or segment name the dialect never registered (e.g. a bare `"POLICIES"` string, or a typo like `Ref('TableReference')`), Python's `dialect.ref()` raises a `RuntimeError` immediately. Rust previously resolved the missing name to `Empty` and quietly failed the branch, producing an unparsable tree where Python raised an error. Rust now signals the missing ref back to `rust_parser.py` via a sentinel, which re-raises through the dialect's own `ref()`.

2. Arena-tree building falls back gracefully when Rust panics. The `except Exception` around tree building was too narrow: a Rust-side panic crosses `pyo3` as `PanicException`, a `BaseException`, not an `Exception`, so it wasn't caught and would abort the whole parse instead of falling back to the documented Python-tree fallback. Now catches `BaseException` (excluding `KeyboardInterrupt`/`SystemExit`). No regression test available here.

3. `RegexParser` unicode case folding now matches Python. Python matches case-insensitively against `raw.upper()` (full Unicode case mapping, e.g. eszett `straße` -> `STRASSE`), but the Rust `regex` crate only does simple case folding, under which `ß` never matches `[A-Z]`. Words like `straße` failed a case-insensitive template on the Rust side only. Fixed by pre-uppercasing the comparison text before matching.

4. Int-typed indentation config is no longer dropped. `RustParser` filtered the indentation config section to boolean-typed values only, silently dropping int-typed truthy settings such as `indented_joins = 1`. Python's `ParseContext.from_config` coerces every value with `bool()` regardless of type; Rust now does the same.

Fixes 1, 3, and 4 are backed by parity regression fixtures in `test/fixtures/parity/regressions.yml`, converting previously-`xfail`'d cases to passing ones.

None expected. Each change narrows an existing divergence to match Python behaviour; no new configuration or public API surface is introduced.

### Pull Request checklist

  - Other.

---
## Summary by cubic
Aligns the Rust parser with Python for missing Ref errors, Unicode-insensitive regex, panic-safe arena-tree building, and indentation config coercion. Shares a single `MISSING_REF_PREFIX` across Rust/Python and ensures exclude probes propagate errors; converts prior parity xfails into passing tests.

- **Bug Fixes**
  - Dangling `Ref` now raises a missing-ref error; Rust returns a `MISSING_REF_PREFIX`-tagged error and `rust_parser.py` re-raises via `dialect.ref()`. The inline fast path falls back to the frame path so the error propagates, and `OneOf`/`AnyNumberOf` exclude probes no longer swallow these errors.
  - Arena-tree building now falls back cleanly on a Rust panic by catching `BaseException` (while letting `KeyboardInterrupt`/`SystemExit` through).
  - `RegexParser` case-insensitive matches use the token’s precomputed uppercased text (full Unicode mapping, e.g., `straße` -> `STRASSE`), matching Python.
  - Indentation config values are coerced with `bool()`, so int settings like `indentation.indented_joins = 1` are honored.

<sup>Written for commit ef736ca0ec03f809fbc9051e3ceffbb33a1c3873. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8244?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sqlfluffrs/sqlfluffrs.pyi
+++ b/sqlfluffrs/sqlfluffrs.pyi
@@ -238,6 +238,13 @@ class RsParseError(Exception):
 
     pos: int
 
+MISSING_REF_PREFIX: str
+"""Sentinel prefix on an ``RsParseError`` message for an unresolvable Ref.
+
+Shared with the Rust side (``ref_grammar.rs``) so both sides of the pyo3
+boundary read the same constant instead of duplicating the literal.
+"""
+
 class RsParser:
     """Rust-based SQL parser."""
 
--- a/sqlfluffrs/sqlfluffrs_parser/src/lib.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/lib.rs
@@ -2,6 +2,7 @@ pub mod parser;
 
 // Public read API for the parse arena, consumed by the rules crate.
 pub use parser::arena::{Arena, NodeId};
+pub use parser::MISSING_REF_PREFIX;
 
 #[cfg(feature = "python")]
 pub use parser::{PyHandle, PyMatchResult, PyNode, PyParser, PyTree, RsParseError};
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/core.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/core.rs
@@ -1230,22 +1230,22 @@ impl<'a> Parser<'a> {
                 // folding, which misses that. Uses the token's precomputed
                 // raw_upper(), so this stays cheap under backtracking.
                 let raw = if case_insensitive {
-                    std::borrow::Cow::Borrowed(tok.raw_upper())
+                    tok.raw_upper()
                 } else {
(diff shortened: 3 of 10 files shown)


alanmcruickshank (approved):
Nice. Glad to have more people diving into the rust package 👍

PR #8247, opened 2026-07-29
fix(rust): avoid double allocation in pyo3 getters for collection fields (trim/type/escape)

rubytobi:
Originated [here](https://github.com/sqlfluff/sqlfluff/pull/8224#discussion_r3670201506), I looked a bit further and applied it to other call sites as well.

<!--Thanks for adding this feature!-->                                                                                                                                                                                            

<!--Please give the Pull Request a meaningful title for the release notes-->                                                                                                                                                      

Several pyo3 getters (`trim_start`, `trim_chars`, type sets, escape replacements) were cloning their underlying Rust collections into an intermediate `Vec` before converting to a Python object, allocating twice per call. This adds a shared helper (`pyo3_helpers.rs`) to convert directly into a `PyTuple`/Python object, and updates the getters in `arena.rs`, `arena_py.rs`, `python.rs`, and `token.rs` to use it, cutting the redundant allocation.

`trim_start`/`trim_chars` now return as a `PyTuple` instead of `Vec<String>` on the Python side; any callers relying on list-specific mutation of these values would need updating (tuples are immutable).

### Pull Request checklist                                                                                                                                                                                                        
                                                                                                                                                                                                                                    
  - Other.                                                                                                                                                                                                                        


<!-- This is an auto-generated description by cubic. -->
---
## Summary by cubic
Build Python lists/tuples/frozensets directly from cached Rust data in `pyo3` getters to remove double allocations and avoid holding locks during Python allocation. Standardizes shared builders to return `PyResult`, fixes unwrap panics, and caches `class_types`.

- **Refactors**
  - Added `sqlfluffrs_python::pyo3_helpers` (`pylist_of_strs`, `pylist_of_strs_iter`, `pyfrozenset_of_strs`, `pylist_of_str_pairs`, `pytuple_of_strs`) returning `PyResult`.
  - Store `instance_types` and `trim_chars` as `Arc<Vec<_>>`; cache `class_types` as `Arc<Vec<_>>`; expose `*_arc` getters for `trim_chars` and `escape_replacements`. `descendant_type_set` builds a `PyList` from an iterator.
  - Updated getters in `arena_py.rs`, `python.rs` (`PyNode`, `PyMatchResult`), and `token.rs` to return `PyList`/`PyTuple`/`PyFrozenSet` and propagate `PyResult`; no locks are held across Python allocations. Added tests pinning container types for `RsToken`, `RsHandle`, and `RsMatchResult`. Updated docs.

- **Migration**
  - `trim_start` and `trim_chars` now return Python tuples.
  - `class_types` on `PyToken` now returns a `frozenset` (was a mutable set).

<sup>Written for commit 1fb699e47cc2efccd0959bb985ebf92178bc501c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8247?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>

<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/sqlfluffrs/sqlfluffrs_parser/src/parser/arena.rs
+++ b/sqlfluffrs/sqlfluffrs_parser/src/parser/arena.rs
@@ -52,7 +52,11 @@ enum ArenaKind {
         segment_class: Cow<'static, str>,
         segment_type: Cow<'static, str>,
         raw: String,
-        instance_types: Vec<String>,
+        /// `Arc`-wrapped so `with_instance_types`'s PyO3 caller can clone the
+        /// handle under the arena lock, drop the guard, and only then build
+        /// the Python list — never holding the lock across a Python
+        /// allocation.
+        instance_types: Arc<Vec<String>>,
         class_types: Vec<String>,
         kwargs: RawSegmentKwargs,
     },
@@ -162,7 +166,7 @@ impl Arena {
                     segment_class: segment_class.clone(),
                     segment_type: segment_type.clone(),
                     raw: raw.clone(),
-                    instance_types: instance_types.clone(),
+                    instance_types: Arc::new(instance_types.clone()),
                     class_types: class_types.clone(),
                     kwargs: segment_kwargs.clone(),
                 },
@@ -390,12 +394,13 @@ impl Arena {
         self.node_type_set(id)
     }
 
-    /// Instance types for a raw token, borrowed to avoid cloning into an
-    /// intermediate `Vec` — callers build a Python list/tuple directly.
-    pub fn with_instance_types<R>(&self, id: NodeId, f: impl FnOnce(&[String]) -> R) -> R {
+    /// Instance types for a raw token: a cheap `Arc` clone (a refcount bump,
+    /// not a deep copy) so callers can drop the arena lock before building a
+    /// Python object from it.
+    pub fn instance_types_arc(&self, id: NodeId) -> Arc<Vec<String>> {
         match &self.node(id).kind {
-            ArenaKind::Raw { instance_types, .. } => f(instance_types),
-            _ => f(&[]),
+            ArenaKind::Raw { instance_types, .. } => instance_types.clone(),
+            _ => Arc::new(Vec::new()),
         }
     }
 
@@ -419,12 +424,12 @@ impl Arena {
     }
 
     /// Characters to trim from both ends of a raw token (if set on the
-    /// token), borrowed to avoid cloning into an intermediate `Vec` —
-    /// callers build a Python list/tuple directly.
-    pub fn with_trim_chars<R>(&self, id: NodeId, f: impl FnOnce(Option<&[String]>) -> R) -> R {
+    /// token): a cheap `Arc` clone so callers can drop the arena lock before
+    /// building a Python object from it.
+    pub fn trim_chars_arc(&self, id: NodeId) -> Option<Arc<Vec<String>>> {
         match &self.node(id).kind {
-            ArenaKind::Raw { kwargs, .. } => f(kwargs.trim_chars.as_deref()),
-            _ => f(None),
+            ArenaKind::Raw { kwargs, .. } => kwargs.trim_chars.clone(),
+            _ => None,
         }
     }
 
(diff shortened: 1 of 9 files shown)


sqlfluffrs/sqlfluffrs_parser/src/parser/arena.rs
```diff
@@ -432,11 +436,19 @@ impl Arena {
...
+    ) -> R {
         match &self.node(id).kind {
-            ArenaKind::Raw { kwargs, .. } => kwargs.escape_replacements.as_deref().cloned(),
-            _ => None,
+            ArenaKind::Raw { kwargs, .. } => {
+                f(kwargs.escape_replacements.as_deref().map(Vec::as_slice))
+            }
+            _ => f(None),
         }
     }
```
keraion: Some of the MutexGuards leave room for potential deadlock

```rs
// cheap refcount bump under the lock, no deep clone
pub fn escape_replacements_arc(&self, id: NodeId) -> Option<Arc<Vec<(String, String)>>> {
    match &self.node(id).kind {
        ArenaKind::Raw { kwargs, .. } => kwargs.escape_replacements.clone(),
        _ => None,
    }
}
```

then in arena_py.rs like `descendant_type_set`
```rs
// lock released at end of the let-statement, conversion happens after
fn escape_replacements<'py>(&self, py: Python<'py>) -> PyResult<Option<Bound<'py, PyList>>> {
    let pairs = self.inner.lock().unwrap().escape_replacements_arc(self.node);
    pairs.map(|p| pylist_of_str_pairs(py, &p)).transpose()
}
```

sqlfluffrs/sqlfluffrs_parser/src/parser/arena.rs
```diff
@@ -390,10 +390,12 @@ impl Arena {
...
 
-    pub fn instance_types(&self, id: NodeId) -> Vec<String> {
+    /// Instance types for a raw token, borrowed to avoid cloning into an
+    /// intermediate `Vec` — callers build a Python list/tuple directly.
+    pub fn with_instance_types<R>(&self, id: NodeId, f: impl FnOnce(&[String]) -> R) -> R {
         match &self.node(id).kind {
-            ArenaKind::Raw { instance_types, .. } => instance_types.clone(),
-            _ => Vec::new(),
+            ArenaKind::Raw { instance_types, .. } => f(instance_types),
+            _ => f(&[]),
```
keraion: Following the suggestion for `escape_replacements` - `trim_chars`/`instance_types`, we should change the arena storage from `Vec<String>` to `Arc<[String]>` (or  maybe `Arc<Vec<String>>`). That's a one-time type change; after it, every getter follows the same shape: Arc-clone under lock (a refcount increment, roughly free), drop guard, build the Python object with no lock held. This keeps the full allocation win and clean lock discipline.

But I think we keep this for a follow up PR, the deadlock risk should be very small given we don't have any `__del__` calls on the Python side.

sqlfluffrs/sqlfluffrs_parser/src/parser/arena_py.rs
```diff
@@ -161,40 +163,48 @@ impl PyHandle {
         self.inner.lock().unwrap().class_types(self.node)
     }
```
keraion: Should we apply the pattern here too?

keraion (commented):
This looks good, just a few thoughts.

We might want to have a contract test on the Python side, just to make sure these are surfacing as the correct types too.

rubytobi:
I went with the Arc fix for `instance_types`/`trim_chars`/`escape_replacements` now rather than a follow up, same idea as your `escape_replacements_arc` suggestion. `class_types` never actually held the lock while building the Python object, so nothing to fix there, but I cached it behind an `Arc` too, same as `descendant_type_set`. Added contract tests pinning the container types across `RsToken`, `RsHandle`, and `RsMatchResult` so this can't regress silently.

Happy to update the pull request description to align closer to what it actually ended up as.

keraion (approved):
Nice.

PR #8252, opened 2026-07-30
Fix Oracle SELECT INTO record field parsing

subotac:
### Brief summary of the change made

  Fixes #8250.

  Oracle `SELECT ... INTO` statements allow dotted record and collection field targets, but `IntoClauseSegment` only accepted bare identifiers or bind variables. This left the remainder of a dotted
  target unparsable and could also cause false ST03 unused CTE violations.

  Allow `ObjectReferenceSegment` as the longer dotted-target alternative while retaining the existing single-identifier match to preserve current parse trees. Add an Oracle parser fixture and an ST03
  regression case.

  ### Are there any other side effects of this change that we should be aware of?

  No known side effects. Existing bare `SELECT INTO` targets retain their current parse tree.

  ### Testing

  - All 300 Oracle dialect tests passed.
  - All 33 ST03 YAML cases passed.
  - The full Python suite completed with 10,589 passed and 2,533 skipped.
  - Rust workspace and Python/Rust fixture parity tests passed.
  - Ruff, mypy, pre-commit, and `git diff --check` passed.

  ### AI assistance disclosure

  AI was not used anywhere except for the final code analysis following my manual review.

  ### Pull Request checklist

  - [x] Please confirm you have completed any of the necessary steps below.

  - Included test cases to demonstrate the code changes:
    - `.sql`/`.yml` Oracle parser test cases in `test/fixtures/dialects`.
    - An ST03 rule test case in `test/fixtures/rules/std_rule_cases`.
  - Documentation changes are not required because this corrects existing Oracle parsing behavior.
  - No follow-up issues are currently required.

---
## Summary by cubic
Fix Oracle parsing of dotted record/collection targets in `SELECT ... INTO` by allowing object references (e.g., `out_metrics.a`) in the `INTO` clause, preventing the `FROM` clause from being hidden and stopping false ST03 unused CTE reports. Fixes #8250.

- **Bug Fixes**
  - Expanded `IntoClauseSegment` to accept `ObjectReferenceSegment` in addition to single identifiers and bind variables.
  - Added an Oracle parser fixture and an ST03 regression test to cover dotted `INTO` targets.

<sup>Written for commit cd6d7273018564828afe9d99e396e20c0ec630c6. Summary will update on new commits.</sup>

[...]

PR #8260, opened 2026-07-30
Oracle: allow record and collection fields as INTO targets

### LINKED ISSUE #8250: Oracle: SELECT...INTO with record/collection-field targets (e.g. \INTO rec.field`) is unparsable, causing false ST03 unused-CTE positives`
### Search before asking

### What Happened

Oracle `PL/SQL SELECT ... INTO` clauses that target fields of a record variable (e.g. `INTO out_rec.col1, out_rec.col2`) fail to parse. The `.field` portion after the first `INTO `identifier; and everything following it in the statement, including the trailing `FROM `clause; collapses into a single `unparsable` segment. As a knock-on effect, because the `FROM` `cte_name` text is now buried inside `unparsable` rather than a real from_clause/table_reference node, rule ST03 (`structure.unused_cte`) incorrectly reports the referenced **CTE** as unused, since its usage-detection (`recursive_crawl("table_reference")`) never sees a proper `table_reference` node for it.

### Expected Behaviour

The `INTO` clause should parse cleanly when its targets are dotted record/collection-field references (e.g. `out_rec.col1`), not just bare identifiers or bind variables.

No `PRS` (parse) error, and no `ST03` violation for `cte1`, since it's genuinely referenced in the `FROM` clause.

### Observed Behaviour

See "How to reproduce" for repro.sql lines.

```
== [repro.sql] FAIL
L:   5 | P:   8 | ST03 | Query defines CTE "cte1" but does not use it.
                       | [structure.unused_cte]
L:  12 | P:  16 |  PRS | Line 12, Position 16: Found unparsable section: '.a,\n
                       | out_metrics.b\n  from cte1'
WARNING: Parsing errors found and dialect is set to 'oracle'. Have you configured your dialect correctly?
```
`sqlfluff parse` shows the break clearly; parsing succeeds up through the first `INTO` target (`out_metrics` as a bare `naked_identifier`), then hits unparsable right at the following `.`:

```
into_clause:
    keyword:                  'into'
    newline:                  '\n'
    naked_identifier:         'out_metrics'
unparsable:                   !! Expected: 'Nothing here.'
    dot:                      '.'
    word:                     'a'
    comma:                    ','
    ...
    word:                     'from'
    word:                     'cte1'
```

### How to reproduce

Create a file called repro.sql with the following lines:
```
create or replace procedure test_proc(out_rec out sys_refcursor) is
  type t_rec is record (a number, b number);
  out_metrics t_rec;
begin
  with cte1 as (
    select 1 as x from dual
  )
  select
    cte1.x,
    cte1.x
  into
    out_metrics.a,
    out_metrics.b
  from cte1;
end;
/
```

`sqlfluff lint --dialect oracle --rules ST03 repro.sql`
`sqlfluff parse --dialect oracle repro.sql`

### Dialect

oracle

### Version

sqlfluff, version 4.2.0

### Configuration

No project .sqlfluff/.sqlfluffignore needed to reproduce; the repro above fails with just --dialect oracle passed on the CLI, no config file present. 

### Code of Conduct

devanshranjan10:

Fixes #8250.

PL/SQL `SELECT ... INTO` and `BULK COLLECT INTO` accept fields of a record or collection variable as targets, e.g. `INTO rec.a, rec.b`. Only `SingleIdentifierGrammar` and `BindVariableSegment` were accepted, so everything from the first `.` onward — **including the trailing `FROM` clause** — collapsed into a single `unparsable` segment:

```
into_clause:
    keyword:          'into'
    naked_identifier: 'out_metrics'
unparsable:           !! Expected: 'Nothing here.'
    dot:              '.'
    word:             'a'
    ...
    word:             'from'
    word:             'cte1'
```

That produced the confusing knock-on failure in the issue title: with the `FROM` clause buried inside `unparsable` rather than parsed as a `table_reference`, ST03's usage detection (`recursive_crawl("table_reference")`) never saw the reference, so it reported a genuinely-used CTE as unused.

This accepts `ObjectReferenceSegment` in both clauses.

`SingleIdentifierGrammar` is deliberately kept **ahead** of `ObjectReferenceSegment` in the `OneOf`. `ObjectReferenceSegment` would also match a bare name, and putting it first re-shaped every existing plain target from a bare `naked_identifier` into a wrapped `object_reference` — 11 fixture files, 125 deleted lines. With the current ordering the plain forms parse exactly as before and **the regenerated fixtures show no churn at all**; the only files that change are the ones I added cases to.

I used `ObjectReferenceSegment` rather than `ColumnReferenceSegment` on purpose: INTO targets are PL/SQL variables, not columns, and `column_reference` is crawled by a number of rules (RF02, AL07, AM03, …) that would then see them as column references.

No dialect inherits from `oracle`, so the change is contained.

### Pull Request checklist

  - [x] `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (YML regenerated with `test/generate_parse_fixture_yml.py --dialect oracle`).

Cases added to `returning_into.sql`: the issue's procedure, a nested record field (`rec.nested.c`), record fields mixed with plain variables, and `BULK COLLECT INTO` a record field.

Verified against a standalone reproduction script, run with only the dialect file reverted and then restored:

```
unpatched:  5 FAIL / 2 PASS
patched:    7 PASS
```

The two that pass while unpatched are deliberate regression guards (plain variable and plain `BULK COLLECT` targets), so a fix that over-reached would fail them.

```
test/dialects/ -k oracle    315 passed
test/rules/                 2485 passed, 101 skipped
ruff check / ruff format / mypy   clean
```

---
## Summary by cubic
Allow record and collection field references (e.g., rec.a) as targets in Oracle SELECT ... INTO and BULK COLLECT INTO. Fixes parsing that swallowed the FROM clause and prevented accurate table/CTE usage detection.

- **Bug Fixes**
  - Support `rec.field` (including nested) in both `INTO` clauses by accepting object references.
  - Keep `SingleIdentifierGrammar` before object refs to preserve existing parse trees and fixtures.
  - Use object references instead of column references so rules don’t misinterpret variables as columns.
  - Add tests for record fields, nested fields, mixed targets, and BULK COLLECT INTO; no other dialects affected.

<sup>Written for commit 13925114290c162897f83da68b61f10d1034fcbb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8260?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_oracle.py
+++ b/src/sqlfluff/dialects/dialect_oracle.py
@@ -3856,7 +3856,15 @@ class IntoClauseSegment(BaseSegment):
 
     match_grammar = Sequence(
         "INTO",
-        Delimited(OneOf(Ref("SingleIdentifierGrammar"), Ref("BindVariableSegment"))),
+        # Targets may be record or collection fields (e.g. `rec.field`), not
+        # just bare variable names.
+        Delimited(
+            OneOf(
+                Ref("SingleIdentifierGrammar"),
+                Ref("ObjectReferenceSegment"),
+                Ref("BindVariableSegment"),
+            )
+        ),
     )
 
 
@@ -3873,7 +3881,14 @@ class BulkCollectIntoClauseSegment(BaseSegment):
         "COLLECT",
         "INTO",
         ImplicitIndent,
-        Delimited(OneOf(Ref("SingleIdentifierGrammar"), Ref("BindVariableSegment"))),
+        # As above, targets may be record or collection fields.
+        Delimited(
+            OneOf(
+                Ref("SingleIdentifierGrammar"),
+                Ref("ObjectReferenceSegment"),
+                Ref("BindVariableSegment"),
+            )
+        ),
         Dedent,
     )
 
--- a/test/fixtures/dialects/oracle/returning_into.sql
+++ b/test/fixtures/dialects/oracle/returning_into.sql
@@ -152,3 +152,51 @@ BEGIN
     END LOOP;
 END;
 /
+
+-- SELECT ... INTO targeting fields of a record variable.
+CREATE OR REPLACE PROCEDURE test_proc(out_rec OUT SYS_REFCURSOR) IS
+  TYPE t_rec IS RECORD (a NUMBER, b NUMBER);
+  out_metrics t_rec;
+BEGIN
+  WITH cte1 AS (
+    SELECT 1 AS x FROM dual
+  )
+
+  SELECT
+    cte1.x,
+    cte1.x
+  INTO
+    out_metrics.a,
+    out_metrics.b
+  FROM cte1;
+END;
+/
+
+-- Record fields mixed with plain variables, and a nested field.
+DECLARE
+  TYPE inner_rec IS RECORD (c NUMBER);
+  TYPE outer_rec IS RECORD (a NUMBER, nested inner_rec);
+  rec        outer_rec;
+  plain_var  NUMBER;
+BEGIN
+  SELECT 1, 2, 3
+  INTO rec.a, plain_var, rec.nested.c
+  FROM dual;
+
+  UPDATE t1 SET id = id + 1
+  WHERE id = 1
+  RETURNING id, description INTO rec.a, rec.nested.c;
+END;
+/
+
+-- BULK COLLECT INTO targeting a record field.
+DECLARE
+  TYPE t_ids IS TABLE OF NUMBER;
+  TYPE t_holder IS RECORD (ids t_ids);
+  holder t_holder;
+BEGIN
+  SELECT id
+  BULK COLLECT INTO holder.ids
+  FROM t1;
+END;
+/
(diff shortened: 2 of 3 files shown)


peterbud:
Hi @devanshranjan10 - see #8252

devanshranjan10:
Yes, #8252 covers this. It even lands on the same approach, keeping `SingleIdentifierGrammar` ahead of `ObjectReferenceSegment` in the `OneOf` so existing parse trees stay unchanged.

Closing mine. #8252 also adds an ST03 regression case that I'd missed. My mistake for not checking open PRs before I started.

PR #8261, opened 2026-07-30
Oracle: support the SQL*Plus SHOW command

### LINKED ISSUE #8102: [oracle] SHOW unparsable

devanshranjan10:

Fixes #8102.

The SQL*Plus `SHOW` command was not supported at all, so a single `show errors` line made the whole file unparsable — which also blocks linting and formatting of every *other* statement in that file.

Options are matched structurally rather than swallowed as free text (the way `PROMPT` is lexed), so the command still lints and formats:

- `SHOW ERRORS [<object type> [schema.]name]` — including the two-word types (`PACKAGE BODY`, `TYPE BODY`, `JAVA CLASS`, `ANALYTIC VIEW`, `ATTRIBUTE DIMENSION`)
- `SHOW PARAMETER[S] [parameter_name]`
- everything else falls through to a bare identifier

That last part is deliberate: it covers arbitrary system variables *and* the abbreviated spellings SQL*Plus accepts (`ERR`, `REL`, `RECYC`, …) without enumerating them. Only the reserved words (`ALL`, `USER`, `RELEASE`, `SQLCODE`, `EDITION`) need spelling out, since those can't match as identifiers.

The one subtlety worth flagging: **SQL\*Plus ends these commands at the newline, which the parser doesn't see.** An argument-less `SHOW PARAMETERS` therefore swallowed the first word of the *following* command as its parameter name:

```sql
show parameters
show user      -- 'show' consumed as the parameter name -> unparsable
```

A newline terminator doesn't help here — newlines are non-code and get skipped — so the command words are excluded from the argument instead, following the existing `Ref("ExpressionSegment", optional=True, exclude=Ref.keyword("FROM"))` precedent in the ANSI dialect. `SHOW ERRORS` was never affected, because its optional part requires an object-type keyword first.

`SHOW` remains usable as an identifier and as a PL/SQL variable name (both covered by fixtures). `SqlplusSetStatementSegment` is untouched.

Out of scope, noted in passing: the `PROMPT` lexer rule is case-sensitive, so lowercase `prompt` is unparsable both before and after this change. Happy to open a separate issue if that's worth fixing.

### Pull Request checklist

  - [x] `.sql`/`.yml` parser test cases in `test/fixtures/dialects` (YML regenerated with `test/generate_parse_fixture_yml.py --dialect oracle`).

Cases added to `sqlplus_commands.sql` covering the argument forms, abbreviations, and `SHOW` as a variable name. The only removed line in the YAML is the fixture `_hash`.

Verified with a standalone script, run with only the dialect file reverted and then restored:

```
unpatched:  12 FAIL / 3 PASS
patched:    15 PASS
```

The three that pass while unpatched are regression guards (`SET` still parses, `SHOW` still usable as an identifier and as a variable name).

```
test/dialects/ + test/rules/    9269 passed, 101 skipped
ruff check / ruff format / mypy  clean
```

---
## Summary by cubic
Add support for the SQL*Plus `SHOW` command in the Oracle dialect so these lines parse, lint, and format instead of breaking entire files. This prevents a single `SHOW` (e.g., `SHOW ERRORS`) from making the file unparsable.

- **New Features**
  - Parse `SHOW ERRORS [<object type> [schema.]name]`, including two-word types (e.g., `PACKAGE BODY`, `TYPE BODY`).
  - Parse `SHOW PARAMETER[S] [name]`.
  - Support other `SHOW` options as bare identifiers; reserved words (`ALL`, `USER`, `RELEASE`, `SQLCODE`, `EDITION`) are handled explicitly.

- **Bug Fixes**
  - Prevent `SHOW PARAMETERS` without an argument from consuming the first word of the next command by excluding `SHOW`, `SET`, `PROMPT` from the optional argument.
  - Keep `SHOW` valid as an identifier and PL/SQL variable name.

<sup>Written for commit 03f37fbe5d83a5f576ca582ee551a68e1cdd3f5a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8261?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


peterbud:
Hi @devanshranjan10  - what is the difference vs  #8104 ?

devanshranjan10:
You're right, this overlaps #8104. My mistake, I checked the issue tracker but not the open PRs first.

Closing this one. #8104 came first and its `SHOW ERRORS` handling is better than mine, since it covers `JAVA SOURCE` and makes `BODY` optional.

One thing came up while I was testing that branch, so I've left a note on #8104 rather than here.

PR #8264, opened 2026-07-30
Databricks: Aligned INSERT statement with documentation, adding missing forms

timothyvries:
Override InsertStatementSegment in the Databricks dialect to correctly support all three documented INSERT forms, including BY NAME combined with REPLACE WHERE/REPLACE USING/REPLACE ON, the WITH SCHEMA EVOLUTION prefix, and the correct restriction that OVERWRITE is not valid with REPLACE forms.
Documentation of Databricks: [INSERT](https://docs.databricks.com/gcp/en/sql/language-manual/sql-ref-syntax-dml-insert-into)
Documentation of SparkSQL: [INSERT TABLE](https://spark.apache.org/docs/4.2.0/sql-ref-syntax-dml-insert-table.html)

The INSERT implementation in Databricks is replacing the one in SparkSQL, meaning that changes to SparkSQL don't automatically flow to Databricks. This is however intended, as the implementations are different.

### Pull Request checklist

  - Other.

---
## Summary by cubic
Overrides the Databricks `InsertStatementSegment` to match Databricks SQL docs, adding the missing INSERT forms and enforcing dialect rules (e.g., no OVERWRITE with REPLACE). This replaces the SparkSQL behavior for INSERT in the Databricks dialect and includes full parser tests.

- **New Features**
  - Support Form 1: INTO/OVERWRITE with optional PARTITION and column list/“BY NAME”.
  - Support Form 2: INTO with BY NAME and REPLACE WHERE or REPLACE USING (...).
  - Support Form 3: INTO with REPLACE ON, including target alias (AS or implicit) and source alias.
  - Accepts the WITH SCHEMA EVOLUTION prefix across forms and disallows OVERWRITE with REPLACE.
  - Adds Databricks fixtures and regression cases, including `TABLE` and `FROM` sources.

<sup>Written for commit 88b905410e8cbdd5fd105f12c06cd7393161e750. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8264?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/dialects/dialect_databricks.py
+++ b/src/sqlfluff/dialects/dialect_databricks.py
@@ -1590,6 +1590,90 @@ class LimitClauseSegment(sparksql.LimitClauseSegment):
     )
 
 
+class InsertStatementSegment(sparksql.InsertStatementSegment):
+    """An `INSERT [TABLE]` statement for Databricks SQL.
+
+    The Databricks docs define three distinct forms:
+
+    Form 1 — standard insert, ``OVERWRITE`` or ``INTO``:
+        ``INSERT [WITH SCHEMA EVOLUTION] { OVERWRITE | INTO } [TABLE] table_name
+        [PARTITION clause] [( column_name [, ...] ) | BY NAME] query``
+
+    Form 2 — selective replace, ``INTO`` only:
+        ``INSERT [WITH SCHEMA EVOLUTION] INTO [TABLE] table_name
+        [BY NAME] [REPLACE WHERE predicate | REPLACE USING (col [, ...])] query``
+
+    Form 3 — replace-on, ``INTO`` only:
+        ``INSERT [WITH SCHEMA EVOLUTION] INTO [TABLE] table_name [target_alias]
+        [BY NAME] REPLACE ON boolean_expression { (query) [source_alias] | query }``
+
+    https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-dml-insert-into
+    """
+
+    type = "insert_statement"
+
+    match_grammar = Sequence(
+        "INSERT",
+        # WITH SCHEMA EVOLUTION: Databricks Runtime 18.1+
+        Sequence("WITH", "SCHEMA", "EVOLUTION", optional=True),
+        OneOf(
+            # Forms 2 & 3: INTO only (OVERWRITE is not allowed with REPLACE)
+            Sequence(
+                "INTO",
+                Ref.keyword("TABLE", optional=True),
+                Ref("TableReferenceSegment"),
+                OneOf(
+                    # Form 3: REPLACE ON with optional target_alias and BY NAME.
+                    # target_alias requires the AS keyword here to avoid ambiguity
+                    # with the REPLACE/BY unreserved keywords being consumed as a
+                    # bare identifier alias.
+                    Sequence(
+                        Sequence(
+                            "AS",
+                            Ref("SingleIdentifierGrammar"),
+                            optional=True,
+                        ),
+                        Sequence("BY", "NAME", optional=True),
+                        "REPLACE",
+                        "ON",
+                        Ref("ExpressionSegment"),
+                        Ref("SelectableGrammar"),
+                    ),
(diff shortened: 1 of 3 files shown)


src/sqlfluff/dialects/dialect_databricks.py
```diff
@@ -1590,6 +1590,90 @@ class LimitClauseSegment(sparksql.LimitClauseSegment):
...
+                        Sequence(
+                            "AS",
+                            Ref("SingleIdentifierGrammar"),
+                            optional=True,
+                        ),
+                        Sequence("BY", "NAME", optional=True),
+                        "REPLACE",
+                        "ON",
+                        Ref("ExpressionSegment"),
+                        Ref("SelectableGrammar"),
```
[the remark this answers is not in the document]
timothyvries: Implemented suggested change.

src/sqlfluff/dialects/dialect_databricks.py
```diff
@@ -1590,6 +1590,90 @@ class LimitClauseSegment(sparksql.LimitClauseSegment):
...
+                Ref.keyword("TABLE", optional=True),
+                Ref("TableReferenceSegment"),
+                OneOf(
+                    # Form 3: REPLACE ON with optional target_alias and BY NAME.
+                    # target_alias requires the AS keyword here to avoid ambiguity
+                    # with the REPLACE/BY unreserved keywords being consumed as a
+                    # bare identifier alias.
+                    Sequence(
+                        Sequence(
+                            "AS",
```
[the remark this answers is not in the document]
timothyvries: Implemented suggested change

src/sqlfluff/dialects/dialect_databricks.py
```diff
@@ -1590,6 +1590,92 @@ class LimitClauseSegment(sparksql.LimitClauseSegment):
...
+                OneOf(
+                    # Form 3: REPLACE ON with optional target_alias and BY NAME.
+                    Sequence(
+                        Sequence(
+                            Ref.keyword("AS", optional=True),
+                            Ref("SingleIdentifierGrammar"),
+                            optional=True,
+                        ),
+                        Sequence("BY", "NAME", optional=True),
+                        "REPLACE",
```
peterbud, line 1634: The bracketed form unfortunately won't parse. Example:
```sql
INSERT INTO TABLE students AS t
REPLACE ON t.name = s.name
(SELECT * FROM people);
```

peterbud (changes requested):
Hi @timothyvries  - thanks for this PR.

Overall the direction is really good, but the bracketed query form after `REPLACE ON ...` is not parsing.

The spec says it is allowed:
```
   ...
   REPLACE ON boolean_expression
    { (query) [source_alias] | query }
```

src/sqlfluff/dialects/dialect_databricks.py
timothyvries: Good catch, I did miss that one.

Looking at it, I don't see a non-invasive way to fix this...
I believe the problem is not with the SelectableGrammer (as it allows brackets) but with the ExpressionSegment. The opening bracket is parsed as part of the ExpressionSegment, which threats s.name(...) as a function and the query is treated as the input for this function. Hence it never reaches the SelectableGrammer.

Do you propose a way how to allow the bracketed form to parse?

peterbud: Hi @timothyvries , probably right fix is to bound only this REPLACE ON predicate with a terminator that recognizes a parenthesized selectable, not every (, something like this:
```python
Ref(
    "ExpressionSegment",
    terminators=[Bracketed(Ref("SelectableGrammar"))],
),
Ref("SelectableGrammar"),
```

peterbud:
Closing due to lack of response

Issue #5987, opened 2024-06-26
CV06 not properly parsing file with multiple statments on oracle dialect


Tiaguituh05:
### Search before asking

- [X] I searched the [issues](https://github.com/sqlfluff/sqlfluff/issues) and found no similar issues.

### What Happened

The issue I'm facing is, if I set multiline_newline to true, it detects a missing ";" on the lines 6, 17, and 29. Which is not accurate
If I set it to false, it detects that the last line is missing a ";" which is correct, but it does not complaing about the other ones missing.

### Expected Behaviour

CV06 | Statements must end with a semi-colon - Should be detected on lines 13, 21, 25 and 29.

### Observed Behaviour

if I set **multiline_newline** to **true** , it detects a missing ";" on the lines 6, 17, and 29. Which is not accurate, line 6 and 17 already end with a semi-colon. 

If I set **multiline_newline** to **false**, it only detects the missing semi-colon on the last line, 29.

### How to reproduce

Create a **sample.sql** with the following:
```
DELETE
  FROM code_detail
 WHERE code_type IN ('PLTQ',
                     'RSCD',
                     'SLBK',
                     'PLTM');
--
DELETE
  FROM code_head
 WHERE code_type IN ('PLTQ',
                     'RSCD',
                     'SLBK',
                     'PLTM')
--
-- Plot quality
INSERT INTO code_head(code_type, code_desc)
VALUES('PLTQ', 'Plot Quality');
--
--Reason codes
INSERT INTO code_head(code_type, code_desc)
VALUES('RSCD', 'Reason codes')
--
-- Sales status - Inventory bucket
INSERT INTO code_head(code_type, code_desc)
VALUES('SLBK', 'Sales Status')
--
-- Plot Movements
INSERT INTO code_head(code_type, code_desc)
VALUES('PLTM', 'Movement type')
```

Run:
sqlfluff lint sample.sql

### Dialect

oracle

### Version

3.0.7

### Configuration

[sqlfluff]
          # verbose is an integer (0-2) indicating the level of log output
          verbose = 0
          # Turn off color formatting of output
          nocolor = False
          # Supported dialects https://docs.sqlfluff.com/en/stable/dialects.html
          # Or run 'sqlfluff dialects'
          dialect = oracle
          # One of [raw|jinja|python|placeholder]
          templater = jinja
          # Comma separated list of rules to check, default to all
          rules = all
          # Comma separated list of rules to exclude, or None
[...]

PR #8268, opened 2026-07-31
Give CV06 placement violations accurate descriptions

LALITH0110:

CV06 raised every violation with the rule docstring — *"Statements must end with a semi-colon."* — including the two branches where the semi-colon **is** present and only its position is wrong. None of the rule's nine `LintResult` call sites set a `description`, so a statement that plainly ends in `;` gets reported as missing one.

That is actively misleading, and it has already cost someone a bug report: #5987 reports "false positives" from `multiline_newline = true` on lines that end with `;`. Having reproduced it, the linting behaviour there is **correct** — with `multiline_newline = true` a multi-line statement is supposed to carry its semi-colon on its own line. What was wrong was the message explaining why.

This makes progress on #5987 by fixing the message rather than the behaviour:

| Situation | Before | After |
|---|---|---|
| `;` separated from the statement by whitespace/newlines | Statements must end with a semi-colon. | Semi-colon should not be preceded by whitespace or newlines. |
| multi-line statement, `;` not on its own line | Statements must end with a semi-colon. | Semi-colon should be on a new line after a multi-line statement. |
| `;` genuinely missing | Statements must end with a semi-colon. | *(unchanged)* |

Linting behaviour, violation positions and autofixes are all unchanged — only the `description` text of two branches differs. No existing test asserted the old wording, so nothing needed updating: the full `test/rules` suite passes unchanged at **2506 passed, 101 skipped**.

Anyone matching CV06 output by message text would see the new strings, which is the intended point of the change.

### Pull Request checklist

  - **Other**: three cases added to `test/rules/std_LT12_CV06_test.py`, one per branch above — including one asserting that a genuinely missing semi-colon keeps the original wording.

  These are Python tests rather than `.yml` rule cases because the `violations:` block in the YAML harness compares the full `to_dict()` of each violation, including every autofix; asserting a description that way would have been long and brittle. Happy to move them if you'd prefer.
- [x] Added appropriate documentation for the change — none needed; the rule docstring still describes the rule's purpose and is unchanged.
- [x] Created GitHub issues for any relevant followup/future enhancements if appropriate — none identified.

---

AI disclosure, per CONTRIBUTING: this change was drafted with AI assistance. The wording of the two descriptions and the tests were AI-drafted. I manually validated the change by reproducing #5987 locally, confirming from `CV06.py` that `semicolon_newline = self.multiline_newline if not info.is_one_line else False` makes the reported behaviour intentional, verifying each of the three branches emits the intended message, and running the full `test/rules` suite plus `ruff`, `ruff-format` and `mypy` on the changed files.

---
## Summary by cubic
Update CV06 to show accurate messages for semicolon placement instead of claiming a missing semi-colon when one is present. Behavior and autofixes are unchanged; tests added; clarifies confusion from #5987.

- **Bug Fixes**
  - If whitespace/newlines precede the semi-colon: Semi-colon should not be preceded by whitespace or newlines.
  - For multi-line statements requiring their own line: Semi-colon should be on a new line after a multi-line statement.
  - Genuinely missing semi-colon message remains: Statements must end with a semi-colon.

<sup>Written for commit f188e90a8a895d79358d2dac547f030f93cffe8f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8268?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sqlfluff/rules/convention/CV06.py
+++ b/src/sqlfluff/rules/convention/CV06.py
@@ -245,6 +245,9 @@ class Rule_CV06(BaseRule):
         return LintResult(
             anchor=info.anchor_segment,
             fixes=fixes,
+            description=(
+                "Semi-colon should not be preceded by whitespace or newlines."
+            ),
         )
 
     def _handle_semicolon_newline(
@@ -301,6 +304,9 @@ class Rule_CV06(BaseRule):
         return LintResult(
             anchor=anchor_segment,
             fixes=fixes,
+            description=(
+                "Semi-colon should be on a new line after a multi-line statement."
+            ),
         )
 
     def _create_semicolon_and_delete_whitespace(
--- a/test/rules/std_LT12_CV06_test.py
+++ b/test/rules/std_LT12_CV06_test.py
@@ -24,3 +24,46 @@ def test__rules__std_LT12_and_CV06_interaction() -> None:
 
     # Check file is fixed.
     assert linted_file.fix_string()[0] == "SELECT foo FROM bar;\n"
+
+
+def _lint_cv06(sql: str, **rule_config: object) -> list:
+    """Lint a string with only CV06 enabled and return its violations."""
+    cfg = FluffConfig(overrides={"dialect": "ansi", "rules": "CV06"})
+    for key, val in rule_config.items():
+        cfg.set_value(config_path=["rules", "convention.terminator", key], val=val)
+    return Linter(config=cfg).lint_string(sql).violations
+
+
+def test__rules__std_CV06_description_semicolon_newline() -> None:
+    """Placement violations should not claim the semi-colon is missing.
+
+    The statement here already ends with a semi-colon; the only problem is that
+    ``multiline_newline`` requires it on its own line. Reporting "Statements must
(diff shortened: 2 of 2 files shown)


LALITH0110:
Human here, I reviewed this change and reproduced #5987 locally before submitting, and the AI disclosure is in the description as CONTRIBUTING asks. I don't seem to have permission to reopen it myself; could a maintainer take another look?

peterbud (changes requested):
Hi @LALITH0110  - thanks for opening this PR and welcome.

The PR seems good to me, however from style/preference POV I think we should move the test out of `std_LT12_CV06_test.py` to a standalone `std_CV06_test.py file`, as these new tests has nothing to do with LT12 rule I believe.

peterbud:
Closing due to inactvity

PR #8043, opened 2026-07-01
Fix Redshift EXCLUDE after select list items

ishaanlabs-gg:
## Summary
- allow Redshift `EXCLUDE` clauses after non-wildcard select list items
- add fixture coverage for `SELECT *, NULL AS example EXCLUDE (...)`

Fixes #7899.

## Tests
- `.venv-codex314/bin/python -m pytest test/dialects/dialects_test.py::test__dialect__base_parse_struct -k 'redshift and select_exclude' -q`
- `.venv-codex314/bin/python -m pytest test/dialects/dialects_test.py::test__dialect__base_parse_struct -k redshift -q`
- `.venv-codex314/bin/python -m ruff check src/sqlfluff/dialects/dialect_redshift.py`
- `.venv-codex314/bin/python -m ruff format --check src/sqlfluff/dialects/dialect_redshift.py`
- `git diff --check`
- `.venv-codex314/bin/sqlfluff parse --dialect redshift test/fixtures/dialects/redshift/select_exclude.sql`

---
## Summary by cubic
Fix Redshift SELECT parsing to allow EXCLUDE after non-wildcard items (not just '*'), matching Redshift syntax. Fixes #7899.

- **Bug Fixes**
  - Updated `SelectClauseElementSegment` to allow `ExcludeClauseSegment` after expressions and optional alias.
  - Added fixture for: SELECT *, NULL AS example EXCLUDE (col1, col2) FROM table1.

<sup>Written for commit 23712530cb2376ec1dea84e7bf215966db9d7bd9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sqlfluff/sqlfluff/pull/8043?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #8270, opened 2026-08-01
fix(redshift): parse SELECT EXCLUDE after the full select list

### LINKED ISSUE #7899: Redshift exclude clause throws unparseable error when adding other columns
### Search before asking

### What Happened

SQLFluff handles Redshift's exclude clause when selecting `*`.

It fails when you select more than `*`, which is valid for Redshift.

```sql
SELECT *, NULL AS example EXCLUDE (col1, col2) FROM table1;
```

```
╰─➤  sqlfluff fix                                                                                                                       1 ↵
  == [/Users/hevans/code/python/sqlfluff_bug/select_star_plus.sql] ==                                                                       
  L:   1 | P:  27 |  PRS | Line 1, Position 27: Found unparsable section: 'EXCLUDE
                       | (col1, col2)'
All Finished 📜 🎉!
```

### Expected Behaviour

No parsing error.

### Observed Behaviour

Parsing error.

### How to reproduce

.sql
```sql
SELECT *, NULL AS example EXCLUDE (col1, col2) FROM table1;
```

.sqlfluff
```
[sqlfluff]
dialect = redshift
rules = core
```

`sqlfluff fix`

### Dialect

Redshift

### Version

```
╰─➤  sqlfluff --version                                                                                                                 1 ↵
sqlfluff, version 4.2.1
```

### Configuration

```
[sqlfluff]
dialect = redshift
rules = core
```

### Code of Conduct

dgvj-work:
## Brief summary
Fix Redshift parsing of `EXCLUDE` when it follows a non-wildcard select list (e.g. `SELECT *, NULL AS example EXCLUDE (col1, col2)`).

## Changelog & Docs
- [ ] This PR does not include a changelog entry (bug fix for dialect parsing)
- [ ] Docs not required (behavior aligned with existing vendor syntax)

## Testing
- Extended `test/fixtures/dialects/redshift/select_exclude.sql`
- Regenerated YAML via `python test/generate_parse_fixture_yml.py -d redshift -f 'select_exclude*'`
- Verified against AWS Redshift SELECT / EXCLUDE docs and the alias-ambiguity case called out on #8043

## Approach
Per review on #8043: attach a single optional `ExcludeClauseSegment` after the delimited select list via Postgres `SelectClauseSegment.copy`; stop bare `EXCLUDE` being treated as an implicit alias; keep EXCLUDE off wildcard expressions. `EXCLUDE(...)` expressions and `AS EXCLUDE` remain valid.

Closes #7899

### DIFF
--- a/src/sqlfluff/dialects/dialect_redshift.py
+++ b/src/sqlfluff/dialects/dialect_redshift.py
@@ -21,6 +21,7 @@ from sqlfluff.core.parser import (
     Nothing,
     OneOf,
     OptionallyBracketed,
+    ParseMode,
     Ref,
     RegexLexer,
     RegexParser,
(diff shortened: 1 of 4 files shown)


peterbud (changes requested):
Hi @dgvj-work - thanks for this PR.

I see this fix is trying solving the right ambiguity, but the implementation is much broader than it needs to be.

My main concern is that this PR now forks both the Redshift select-clause grammar and the _select-clause element_ grammar, instead of keeping the change local to how Redshift attaches EXCLUDE after the select list. That creates a fairly large inheritance break from Postgres dialect and hard-codes a brittle terminator list that will need to stay manually in sync with Postgres.

There also appears to be a real regression from the new EXCLUDE-specific exclusions. With this branch, `SELECT EXCLUDE(col1) FROM table1;` is now unparsable in Redshift, even though EXCLUDE is an unreserved keyword in `dialect_redshift_keywords.py` and the parent Postgres parser accepts that shape. At the same time, `schema.EXCLUDE(col1)` still parses, so the new restriction is not really modeling the clause boundary, it is selectively rejecting unqualified expressions named EXCLUDE.

I’d strongly prefer a narrower approach here:

* Keep EXCLUDE out of `WildcardExpressionSegment`.
* Handle the ambiguity at the select-clause level only.
* Avoid redefining the full Postgres select clause and avoid globally excluding EXCLUDE from `BaseExpressionElementGrammar`.

The current PR fixes the target examples, but the blast radius is too high for a dialect tweak and it already looks to have introduced unrelated parsing behavior.

dgvj-work:
Thanks @peterbud — agreed the previous approach was too broad.

I narrowed this in 0a8f530:

1. **Select clause only inherits via `.copy()`** — `ExcludeClauseSegment` is inserted after the Postgres delimited select list. No more hard-coded Postgres terminator list / full clause rewrite.
2. **Removed the `BaseExpressionElementGrammar` exclusion** — `SELECT EXCLUDE(col1) FROM table1;` and `schema.EXCLUDE(col1)` parse again.
3. **EXCLUDE stays off `WildcardExpressionSegment`** — one select-clause-level `select_exclude_clause` for both `SELECT * EXCLUDE ...` and mixed lists like `SELECT *, NULL AS example EXCLUDE (...)`.
4. **Minimal bare-alias guard** — still needed so `SELECT *, 1 EXCLUDE (col1)` does not treat `EXCLUDE` as an implicit alias. This does **not** block `AS EXCLUDE` or `EXCLUDE(...)` expressions.

Also added fixture coverage for the function / qualified / `AS EXCLUDE` cases called out in review.

peterbud:
Hi @dgvj-work  - it is much better, thanks.

However the last refactor drops the select-list `EXCLUDE` terminator, so `SELECT *, a, EXCLUDE x FROM tbl` is accepted as an ordinary select item (`EXCLUDE` with alias `x`) instead of rejecting the comma before `EXCLUDE`. Please preserve that terminator and add this negative case to the fixture.

dgvj-work:
Thanks @peterbud — restored the select-list `EXCLUDE` terminator.

`SELECT *, a, EXCLUDE x FROM tbl` is rejected (comma before `EXCLUDE`). The matching valid form `SELECT *, a EXCLUDE x FROM tbl` is in the YAML fixture. YAML fixtures cannot contain unparsable SQL, so the negative case is in `test/dialects/redshift_test.py`.

Postgres select-clause terminators stay inherited via `.copy()`. `SELECT EXCLUDE`, `SELECT EXCLUDE(col1)`, and trailing commas before `FROM` still parse.

peterbud (approved):
Thanks, LGTM

